PR #9827, opened 2026-03-13
feat(server): add counter animation to marketing stats

pepicrft:
## Summary

Add a smooth counting animation for statistics on marketing pages. When the stats section scrolls into view, numbers animate from 0 to the target value with an ease-out cubic easing over 2 seconds.

This creates a more engaging experience compared to numbers just appearing statically (which was noticed as a "jump").

## Changes

- **New hook:** `CounterAnimation` - uses IntersectionObserver to trigger animation when visible
- Uses locale-aware number formatting (e.g., `17,637,304`)
- Configurable duration via `data-counter-duration` attribute

## Updated Pages

- Cache (artifacts served in last 24 hours)
- Build Insights (builds analyzed)
- Flaky Tests (tests detected)
- Selective Testing (test runs)
- Test Insights (test cases processed)

## Testing

Visit any of the marketing pages and scroll to the stats section to see the animation.

### DIFF
--- a/server/assets/marketing/js/hooks/counter-animation.js
+++ b/server/assets/marketing/js/hooks/counter-animation.js
@@ -13,7 +13,7 @@ const CounterAnimation = {
           }
         });
       },
-      { threshold: 0.1 }
+      { threshold: 0.1 },
     );
 
     this.observer.observe(this.el);
@@ -55,9 +55,7 @@ const CounterAnimation = {
       // Ease-out cubic for smooth deceleration
       const easedProgress = 1 - Math.pow(1 - progress, 3);
 
-      this.currentValue = Math.floor(
-        startValue + (endValue - startValue) * easedProgress
-      );
+      this.currentValue = Math.floor(startValue + (endValue - startValue) * easedProgress);
       this.el.textContent = this.currentValue.toLocaleString();
 
       if (progress < 1) {


pepicrft:
Tested the following two scenarios:
- From 0 to the current value when the page is first loaded
- From the current value to a new one when values are inserted (I inserted them while on the page).

server/assets/marketing/js/hooks/counter-animation.js
```diff
@@ -0,0 +1,95 @@
+const CounterAnimation = {
+  mounted() {
+    this.currentValue = 0;
+    this.targetValue = parseInt(this.el.dataset.counterTarget, 10) || 0;
+    this.duration = parseInt(this.el.dataset.counterDuration, 10) || 2000;
+    this.locale = this.el.dataset.counterLocale || "en-US";
```
fortmarek: we're not using `counterLocale`?

server/assets/marketing/js/hooks/counter-animation.js
```diff
@@ -0,0 +1,95 @@
...
+
+    if (newTarget !== this.targetValue) {
+      const oldValue = this.targetValue;
+      this.targetValue = newTarget;
+
+      // If element is visible, animate to new value
+      if (this.isElementVisible()) {
+        this.animateCounter(oldValue, newTarget);
+      } else {
+        // Update immediately if not visible, will animate when scrolled into view
```
fortmarek: hmm ... why? feels like a premature optimization

server/assets/marketing/js/hooks/counter-animation.js
```diff
@@ -0,0 +1,95 @@
...
+    this.currentValue = 0;
+    this.targetValue = parseInt(this.el.dataset.counterTarget, 10) || 0;
+    this.duration = parseInt(this.el.dataset.counterDuration, 10) || 2000;
+    this.locale = this.el.dataset.counterLocale || "en-US";
+    this.animationId = null;
+
+    this.observer = new IntersectionObserver(
+      (entries) => {
+        entries.forEach((entry) => {
+          if (entry.isIntersecting && this.currentValue !== this.targetValue) {
```
fortmarek: feels somewhat unnecessary given this observer should be created just once on mounted and we always start the `currentValue` from 0?

pepicrft: You are right. I'll drop it.

server/assets/marketing/js/hooks/counter-animation.js
pepicrft: I updated it to use currentLocale.

server/assets/marketing/js/hooks/counter-animation.js
pepicrft: I'll drop it.

PR #9832, opened 2026-03-13
feat(server): add SSO enforcement option for organizations

fortmarek:
Screenshot 2026-03-13 at 14 47 13

## Summary

- Adds `sso_enforced` boolean to organizations, configurable in the SSO settings tab
- When enforced, members must verify their identity with the org's SSO provider (Google/Okta) before accessing organization resources in the dashboard
- Follows the Fly.io model: normal login is never blocked, but accessing an SSO-enforced org redirects to the SSO provider for verification
- Enforcement is checked via a `require_sso_authentication` function plug in the router pipeline, which compares the session's `auth_provider` (set at login time) against the org's required SSO provider
- After SSO verification, users are redirected back to the page they originally tried to access
- The admin enabling enforcement must have already authenticated via the SSO provider to prevent lockout scenarios
- API responses and OpenAPI schema updated to include `sso_enforced` field

## Design notes

**Session-based auth provider tracking:** At login time, the session records how the user authenticated (`auth_provider: :password | :google | :okta`). The `require_sso_authentication` plug checks this value against the org's `sso_provider`. This ensures that even if a user has a prior OAuth identity from a past session, they must re-authenticate via SSO in the current session.

**Function plug in the router pipeline:** SSO enforcement runs as a function plug (like `require_authenticated_user`) in the HTTP request pipeline, which allows it to write `oauth_return_to` to the session and redirect directly to the SSO provider — no intermediate controller needed.

## Test plan

- [x] Enable SSO (Google or Okta) on an organization
- [x] Toggle "Enforce SSO" on — verify it rejects if admin hasn't authenticated via SSO
- [x] With enforcement enabled, log in with email/password — verify redirect to SSO provider when accessing the org
- [x] Log in via the matching SSO provider — verify org resources are accessible without redirect
- [x] After SSO verification, verify user returns to the originally requested page
- [x] Verify personal accounts and non-enforced orgs are unaffected
- [x] Verify disabling SSO also clears enforcement
- [x] Verify API returns `sso_enforced` in organization responses

### DIFF
--- a/server/lib/tuist_web/authentication.ex
+++ b/server/lib/tuist_web/authentication.ex
@@ -133,14 +133,14 @@ defmodule TuistWeb.Authentication do
   def log_in_user(conn, user, params \\ %{}) do
     token = Accounts.generate_user_session_token(user)
     user_return_to = get_session(conn, :user_return_to)
-    auth_provider = Map.get(params, :auth_provider, :password)
+    auth_method = Map.get(params, :auth_method, :password)
 
     Analytics.user_authenticate(user)
 
     conn
     |> renew_session()
     |> put_token_in_session(token)
-    |> put_session(:auth_provider, auth_provider)
+    |> put_session(:auth_method, auth_method)
     |> maybe_write_remember_me_cookie(token, params)
     |> redirect(to: user_return_to || signed_in_path(user))
     |> halt()
@@ -329,16 +329,10 @@ defmodule TuistWeb.Authentication do
          organization_id when not is_nil(organization_id) <- account.organization_id,
          {:ok, organization} <- Accounts.get_organization_by_id(organization_id),
          true <- organization.sso_enforced and not is_nil(organization.sso_provider),
-         auth_provider = get_session(conn, :auth_provider),
-         false <- auth_provider == organization.sso_provider do
-      return_to =
-        case conn.query_string do
-          "" -> conn.request_path
-          qs -> "#{conn.request_path}?#{qs}"
-        end
-
+         auth_method = get_session(conn, :auth_method),
+         false <- auth_method == organization.sso_provider do
       conn
-      |> put_session(:oauth_return_to, return_to)
+      |> put_session(:oauth_return_to, current_path(conn))
       |> redirect(to: sso_provider_path(organization))
       |> halt()
     else
--- a/server/lib/tuist_web/live/sso_settings_live.ex
+++ b/server/lib/tuist_web/live/sso_settings_live.ex
@@ -29,7 +29,7 @@ defmodule TuistWeb.SSOSettingsLive do
       |> assign(selected_tab: "sso")
       |> assign(organization: organization)
       |> assign(sso_enabled: sso_enabled)
-      |> assign(sso_enforced: organization.sso_enforced || false)
+      |> assign(sso_enforced: organization.sso_enforced)
       |> assign(flash_message: nil)
       |> assign_form_from_organization(organization)
       |> assign_saved_state()
--- a/server/test/tuist_web/authentication_test.exs
+++ b/server/test/tuist_web/authentication_test.exs
@@ -80,14 +80,14 @@ defmodule TuistWeb.AuthenticationTest do
       assert redirected_to(conn) == "/hello"
     end
 
-    test "stores auth_provider as :password by default", %{conn: conn, user: user} do
+    test "stores auth_method as :password by default", %{conn: conn, user: user} do
       conn = Authentication.log_in_user(conn, user)
-      assert get_session(conn, :auth_provider) == :password
+      assert get_session(conn, :auth_method) == :password
     end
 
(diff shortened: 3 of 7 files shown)


server/lib/tuist_web/live/sso_settings_live.ex
```diff
@@ -29,6 +29,7 @@ defmodule TuistWeb.SSOSettingsLive do
       |> assign(selected_tab: "sso")
       |> assign(organization: organization)
       |> assign(sso_enabled: sso_enabled)
+      |> assign(sso_enforced: organization.sso_enforced || false)
```
cschmatzler: row is `null: false`, so we shouldn't need the `|| false` fallback

server/lib/tuist_web/authentication.ex
```diff
@@ -322,6 +324,35 @@ defmodule TuistWeb.Authentication do
...
+         organization_id when not is_nil(organization_id) <- account.organization_id,
+         {:ok, organization} <- Accounts.get_organization_by_id(organization_id),
+         true <- organization.sso_enforced and not is_nil(organization.sso_provider),
+         auth_provider = get_session(conn, :auth_provider),
+         false <- auth_provider == organization.sso_provider do
+      return_to =
+        case conn.query_string do
+          "" -> conn.request_path
+          qs -> "#{conn.request_path}?#{qs}"
+        end
```
pepicrft: Doesn't conn include an attribute that contains both the path and the query string already concatenated?

server/test/tuist_web/controllers/api/organizations_controller_test.exs
```diff
@@ -422,6 +422,47 @@ defmodule TuistWeb.API.OrganizationsControllerTest do
...
+
+      AccountsFixtures.organization_fixture(name: "enforced-org", creator: user)
+
+      # When
+      conn =
+        conn
+        |> put_req_header("content-type", "application/json")
+        |> put(~p"/api/organizations/enforced-org",
+          sso_provider: "google",
+          sso_organization_id: "tuist.io",
```
pepicrft:
```suggestion
          sso_organization_id: "tuist.dev",
```

server/test/tuist_web/controllers/api/organizations_controller_test.exs
```diff
@@ -422,6 +422,47 @@ defmodule TuistWeb.API.OrganizationsControllerTest do
       assert response["name"] == "tuist-org"
       assert response["sso_provider"] == "google"
       assert response["sso_organization_id"] == "tuist.io"
+      assert response["sso_enforced"] == false
+    end
+
+    test "updates an organization with SSO enforcement enabled",
```
pepicrft: Do we want to include the new table attributes in list and show endpoints?

server/test/tuist_web/authentication_test.exs
```diff
@@ -79,6 +80,16 @@ defmodule TuistWeb.AuthenticationTest do
       assert redirected_to(conn) == "/hello"
     end
 
+    test "stores auth_provider as :password by default", %{conn: conn, user: user} do
+      conn = Authentication.log_in_user(conn, user)
+      assert get_session(conn, :auth_provider) == :password
```
pepicrft: I suggest using `:auth_method` as a name since password is not a provider.

server/test/tuist_web/authentication_test.exs
```diff
@@ -503,6 +514,106 @@ defmodule TuistWeb.AuthenticationTest do
...
+        AccountsFixtures.organization_fixture(
+          creator: user,
+          sso_provider: :google,
+          sso_organization_id: "example.com",
+          preload: [:account]
+        )
+
+      Accounts.update_organization(organization, %{sso_enforced: true})
+
+      request_path = "/#{organization.account.name}/settings"
```
pepicrft: What about using `sigil_p` for these things to get those checked at compile-time?

server/lib/tuist_web/authentication.ex
```diff
@@ -322,6 +324,35 @@ defmodule TuistWeb.Authentication do
     end
   end
 
+  def require_sso_authentication(%{params: %{"account_handle" => account_handle}} = conn, _opts) do
```
pepicrft, line 327: Do we need `_opts` here? It doesn't seem to be used.

server/test/tuist_web/authentication_test.exs
fortmarek: I don't see a strong need for that in tests

server/lib/tuist_web/authentication.ex
fortmarek: yes, we do because that's how the phoenix pipeline calls this method

server/test/tuist_web/controllers/api/organizations_controller_test.exs
fortmarek: I would skip it for now

PR #9839, opened 2026-03-13
fix(server): redirect OAuth failures to dedicated error page

fortmarek:
## Summary
- When OAuth authentication fails (e.g., Okta returns `access_denied`), users were silently redirected to the landing page with no visible error
- Added a dedicated `/users/auth/error` page that displays the specific error message
- For Okta `access_denied` errors, the message tells users to ask their org admin to assign them to the Tuist application
- Other OAuth errors show the provider's error message directly

## Test plan
- [x] Existing auth controller tests updated and passing
- [ ] Manually trigger an Okta `access_denied` error and verify the error page renders with the correct message
- [ ] Verify the "Back to log in" button navigates to `/users/log_in`

### DIFF
--- a/server/lib/tuist_web/controllers/auth_controller.ex
+++ b/server/lib/tuist_web/controllers/auth_controller.ex
@@ -9,6 +9,8 @@ defmodule TuistWeb.AuthController do
   alias Tuist.Accounts.Organization
   alias Tuist.OAuth.Okta
   alias TuistWeb.Authentication
+  alias TuistWeb.Errors.UnauthorizedError
+  alias Ueberauth.Failure.Error
 
   require Logger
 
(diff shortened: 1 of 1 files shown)


server/lib/tuist_web/controllers/auth_controller.ex
```diff
@@ -282,6 +279,41 @@ defmodule TuistWeb.AuthController do
     redirect(conn, to: ~p"/users/log_in")
   end
 
+  defp redirect_to_auth_error(conn, %Ueberauth.Failure{} = failure) do
+    redirect_to_auth_error(conn, oauth_failure_message(failure))
+  end
+
+  defp redirect_to_auth_error(conn, message) do
```
pepicrft: Since we already have an error handling system in place that takes care of mapping the error to a status code and error page with populated content, what if we align with that approach and throw an error?

fortmarek: sure, can do that

PR #9842, opened 2026-03-13
feat(cache): distributed KV

cschmatzler:
Implements the RFC: https://community.tuist.dev/t/distributed-key-value-store-for-cache-nodes/939

As we scale cache nodes horizontally, keeping KV metadata in per-node SQLite causes same-region hit/miss flip-flopping and poor cross-region visibility. This PR adds an optional distributed KV mode that keeps the request hot path local while asynchronously replicating the hot working set through a shared Postgres store.

### What changed

- Added `KEY_VALUE_MODE=distributed`, a shared Postgres repo, and the migrations/configuration needed to run distributed KV without changing the default local mode.
- Added async replication workers:
  - a shipper that batches local writes and access bumps from SQLite to Postgres
  - a poller that materializes shared updates back into each node's local SQLite hot cache
- Added distributed cleanup coordination with leases, tombstones, cutoff-aware disk/S3 deletion, and tombstone purging so `tuist cache clean --remote` remains race-safe across nodes.
- Removed `key_value_entry_hashes` and decoupled KV eviction from artifact reference tracking. Local KV eviction now only removes KV metadata, while artifact lifetime is handled by disk eviction, orphan cleanup, and S3 lifecycle policies.
- Updated the self-hosting guide with the new distributed KV configuration and added focused test coverage around replication, cleanup, eviction, and distributed-mode configuration.

### DIFF
--- a/cache/priv/key_value_repo/migrations/20260312110000_drop_key_value_entry_hashes.exs
+++ b/cache/priv/key_value_repo/migrations/20260312110000_drop_key_value_entry_hashes.exs
@@ -1,9 +1,0 @@
-defmodule Cache.KeyValueRepo.Migrations.DropKeyValueEntryHashes do
-  use Ecto.Migration
-
-  def change do
-    drop_if_exists(index(:key_value_entry_hashes, [:account_handle, :project_handle, :cas_hash]))
-    drop_if_exists(unique_index(:key_value_entry_hashes, [:key_value_entry_id, :cas_hash]))
-    drop_if_exists(table(:key_value_entry_hashes))
-  end
-end


docs/docs/en/guides/cache/self-host.md
```diff
@@ -90,6 +91,23 @@ STORAGE_DIR=/storage
 # Optional dedicated KV SQLite database path.
 # Defaults to /data/key_value.sqlite.
 KEY_VALUE_DATABASE_PATH=/data/key_value.sqlite
+
+# Optional distributed KV mode. Defaults to local for single-node deployments.
```
cschmatzler: I'm undecided as to whether properly document this in self hosting in a full "Multi-regional horizontally scaled deployment" section _or_ just not add it at all and just making local mode the only self-hosted option.

cache/lib/cache/key_value_entry.ex
```diff
@@ -9,19 +9,31 @@ defmodule Cache.KeyValueEntry do
...
     |> validate_required([
       :key,
       :json_payload,
       :last_accessed_at
     ])
     |> unique_constraint(:key)
   end
+
+  def scope_from_key(key) when is_binary(key) do
+    ["keyvalue", account_handle, project_handle, cas_id] = String.split(key, ":", parts: 4)
```
fortmarek, line 31: **Critical: `scope_from_key/1` will crash on non-keyvalue keys**

The pattern match `["keyvalue", account_handle, project_handle, cas_id] = String.split(key, ":", parts: 4)` raises a `MatchError` if the key does not have the `keyvalue:` prefix or has fewer than 4 colon-separated segments.

The shipper calls this on every pending row via `prepare_pending_row/1`. If a non-keyvalue entry ever gets a `replication_enqueued_at` token (plausible since `write_batch` in `key_value_buffer.ex` sets it unconditionally), the entire shipper flush crashes.

**Suggestion:** guard this with a `case` or add a fallback clause, or ensure at the write layer that `replication_enqueued_at` is only set for keyvalue-scoped keys.

cache/lib/cache/key_value_replication_shipper.ex
```diff
@@ -0,0 +1,214 @@
...
+    schedule_flush(Config.distributed_kv_ship_interval_ms())
+    {:noreply, state}
+  end
+
+  defp flush_pending_rows do
+    pending_rows = KeyValueEntries.list_pending_replication()
+    :telemetry.execute(@pending_rows_event, %{count: length(pending_rows)}, %{})
+
+    started_at = System.monotonic_time(:millisecond)
+    :ok = ship_entries(pending_rows)
```
fortmarek, line 56: **High: No error handling in `flush_pending_rows/0`**

`:ok = ship_entries(pending_rows)` will crash the GenServer on any Postgres failure (connection timeout, network error). Since this is called from both `handle_info(:flush, ...)` and `handle_call(:flush, ...)`, a failure crashes the GenServer and potentially the caller.

The pending rows will be retried on restart (at-least-once is preserved), but crash-loop churn could be noisy and affect other operations.

**Suggestion:** wrap the flush in try/rescue and log the error, or use `{:error, reason}` returns instead of the `:ok =` assertion.

cache/lib/cache/key_value_entries.ex
```diff
@@ -77,67 +66,172 @@ defmodule Cache.KeyValueEntries do
...
+
+    count
+  end
+
+  def delete_project_entries_before(account_handle, project_handle, cutoff) do
+    prefix = "keyvalue:#{account_handle}:#{project_handle}:"
+
+    query =
+      from(entry in KeyValueEntry,
+        where: fragment("instr(?, ?) = 1", entry.key, ^prefix),
```
fortmarek: **High: `instr()` cannot use a SQLite index — full table scan**

`fragment("instr(?, ?) = 1", entry.key, ^prefix)` forces a sequential scan of the entire `key_value_entries` table. With the configured max of 25 GiB, this could be very slow.

`instr()` avoids SQL wildcard injection (good, and there is a test for it), but the performance trade-off is significant.

**Suggestion:** consider using `LIKE` with proper escaping of `%` and `_` characters, or add dedicated indexed `account_handle` / `project_handle` columns to the SQLite schema (as you already have in the Postgres schema) to enable equality-based lookups.

cache/lib/cache/key_value_entries.ex
```diff
@@ -77,67 +66,172 @@ defmodule Cache.KeyValueEntries do
...
+        from(entry in KeyValueEntry,
+          where: entry.key == ^key,
+          where: entry.source_updated_at <= ^cutoff
+        )
+      )
+
+    count
+  end
+
+  def delete_project_entries_before(account_handle, project_handle, cutoff) do
```
fortmarek, line 184: **High: Query re-evaluation between DELETE and SELECT**

The `query` binding is an Ecto query that gets evaluated twice: once for the DELETE and once for the follow-up SELECT of remaining candidates. Between these two evaluations, rows could be inserted or have their `replication_enqueued_at` change, leading to inconsistent results.

**Suggestion:** capture the candidate keys as a concrete list from step 1 and use `where: entry.key in ^candidate_keys` for the DELETE, so both operations work on the same set of rows.

cache/lib/cache/key_value_replication_poller.ex
```diff
@@ -0,0 +1,258 @@
...
+            })
+
+          if status in [:inserted, :payload_updated] do
+            Cachex.del(:cache_keyvalue_store, row.key)
+          end
+
+          KeyValueAccessTracker.mark_shared_lineage(row.key)
+          {materialized_acc + 1, deleted_acc}
+        else
+          deleted_rows = KeyValueEntries.delete_local_entry_if_not_pending(row.key)
```
fortmarek, line 124: **High: Tombstone + pending local write can resurrect deleted data**

`delete_local_entry_if_not_pending` only deletes if `replication_enqueued_at IS NULL`. If a local write happened between the tombstone being created and the poller processing it, the local entry survives deletion. The shipper then ships this entry to Postgres, potentially resurrecting a key that was meant to be cleaned up.

The upsert conflict resolution in `upsert_shared_entries` partially handles this via the `EXCLUDED.source_updated_at > e.deleted_at` comparison, but edge cases remain when timestamps are equal (e.g., `deleted_at == source_updated_at`), since the CASE expression uses strict `>`.

**Suggestion:** document this edge case, and consider whether `deleted_at` should use microsecond precision matching `source_updated_at` rather than the truncated `cleanup_started_at`.

cache/lib/cache/key_value_entry.ex
cschmatzler: > or ensure at the write layer that replication_enqueued_at is only set for keyvalue-scoped keys.

this is already the case; the `key_value_entries` table cannot contain anything that is not prefixed with `keyvalue:`; and if it somehow does contain something it shouldn't, I'd like to know via Sentry.

cache/lib/cache/key_value_replication_shipper.ex
cschmatzler: > will crash the GenServer on any Postgres failure (connection timeout, network error). 

This matches the behavior we have on the server with Postgres/Clickhouse failures reporting `DBConnection.ConnectionError` via Sentry and not specifically handling it.

cache/lib/cache/key_value_replication_poller.ex
cschmatzler: Also got this from GPT's code review, but the timestamp being the same at millisecond level seems unlikely enough that adding more complicated logic or more precise timestamps doesn't seem to have significant value to me.  

In addition,
>  If a local write happened between the tombstone being created and the poller processing it

This means that there was a write _after_ `tuist cache clean` was called, so preserving the entry also seems correct in terms of business logic. Anything after the `clean` comment should remain.

cache/priv/distributed_kv_repo/migrations/20260312113000_create_distributed_kv_project_cleanups.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Cache.DistributedKV.Repo.Migrations.CreateProjectCleanups do
+  use Ecto.Migration
+
+  def change do
+    create table(:project_cleanups, primary_key: false) do
```
fortmarek: **Suggestion: rename `project_cleanups` → `projects`**

This table feels like it's encoding one operation's state as a dedicated schema. It's really just project-scoped metadata with a lease. A `projects` table would be a more natural home — the current columns map cleanly:

| Current | Renamed |
|---|---|
| `cleanup_started_at` | `last_cleanup_at` |
| `lease_expires_at` | `cleanup_lease_expires_at` |

This gives you a natural place to add future per-project state (e.g., per-project replication config, eviction policy overrides, entry count estimates) without creating yet another single-purpose table each time. The code change is minimal — `Cleanup` module functions just update the schema alias and field names.

cache/lib/cache/config.ex
```diff
@@ -130,6 +132,61 @@ defmodule Cache.Config do
...
+  def distributed_kv_database_url do
+    Application.get_env(:cache, Repo)[:url]
+  end
+
+  def distributed_kv_pool_size do
+    Application.get_env(:cache, Repo)[:pool_size] || 5
+  end
+
+  def distributed_kv_database_timeout_ms do
+    Application.get_env(:cache, :distributed_kv_database_timeout_ms, 10_000)
```
fortmarek: the defaults here are duplicated at cache/config/runtime.exs. Is there a way for us to have a single definition of these variables?

cache/lib/cache/disk.ex
```diff
@@ -93,14 +95,33 @@ defmodule Cache.Disk do
...
       {:ok, _} -> :ok
       {:error, reason, _} -> {:error, reason}
     end
   end
 
+  @doc """
+  Deletes project artifacts only when their current mtime is strictly before the
+  cutoff second.
+  """
+  def delete_project_before(account_handle, project_handle, cutoff, opts \\ []) do
```
fortmarek: **Nit: ambiguous method name**

`delete_project_before` reads like "delete the project before [doing something]". Something like `delete_project_files_before` or `delete_project_artifacts_before_cutoff` would make it clearer that you're deleting *files* older than a *timestamp*.

cache/lib/cache/config.ex
cschmatzler: yeah they should just live in runtime.exs and we remove it here, since it's never nil

cache/lib/cache/key_value_replication_poller.ex
fortmarek: ok, I hope it's clear which comments come from the LLM and which not. They felt real-ish, but happy to skip those next time around. Thanks for the explanation!

cschmatzler: Yup, it was clear, but I still wanted to address it with a message instead of just… resolving it uncommented 😀

docs/docs/en/guides/cache/self-host.md
fortmarek: > just not add it at all and just making local mode the only self-hosted option.

I'm leaning to start with that and if a global mode is something that comes up for teams that self-host, we can revisit this decision.

pepicrft: Yeah, I agree with @fortmarek.

PR #9856, opened 2026-03-16
fix(server): preserve mobile menu state during live counter updates and fix counter width stability

pepicrft:
## Summary

- The live counter on marketing feature pages triggers LiveView DOM patches every 5 seconds. The MobileMenu and MobileMenuDropdown hooks were resetting their open/closed state on each patch, collapsing any open mobile menu. Both hooks now preserve state across patches and restore it without animation. Also consolidated the duplicate MobileMenu export that lived in mobile-menu-dropdown.js into mobile-menu.js.
- The counter digits were shifting width as values changed because DM Sans lacks tabular figure support. Switched the counter font-family to Inter (`var(--noora-font-body)`) which supports `font-variant-numeric: tabular-nums` for equal-width digits.


server/assets/marketing/js/hooks/mobile-menu-dropdown.js
```diff
@@ -12,8 +12,10 @@ export const MobileMenuDropdown = {
   },
 
   updated() {
+    // Preserve open state across LiveView patches
+    const wasOpen = this.isOpen || false;
     this.cleanup();
-    this.initDropdown();
+    this.initDropdown(wasOpen);
```
fortmarek: running `initDropdown` on `updated` is probably not the best pattern. The current solution seems more like a plaster, I would consider having an `updateDropdown` method instead.

PR #9864, opened 2026-03-16
fix(server): avoid GitHub API rate limits when installing Render CLI

fortmarek:
## Summary
- Adds `render-oss/cli` to `mise.toml` so it's managed as a locked tool with pinned CDN URLs and checksums in `mise.lock`
- Replaces ad-hoc `mise exec "github:render-oss/cli@2.6.1"` calls in deployment workflows with `mise exec -- cli_v2.6.1`, which uses the locked URLs
- Removes `MISE_GITHUB_ATTESTATIONS: 0` env vars that were workarounds for the same issue

This fixes the 403 rate limit errors during deployment ([failed run](https://github.com/tuist/tuist/actions/runs/23150572534/job/67250925182#step:8:20)) caused by mise hitting the GitHub API to resolve release assets.

## Test plan
- [ ] Trigger a canary deployment and verify the Render CLI installs without API rate limit errors
- [ ] Verify deployment completes successfully

### DIFF
--- a/.github/workflows/server-canary-deployment.yml
+++ b/.github/workflows/server-canary-deployment.yml
@@ -55,7 +55,5 @@ jobs:
           VERSION="${{ steps.version.outputs.version }}"
           sed -i 's/version: "[^"]*"/version: "'$VERSION'"/' mix.exs
       - name: Trigger canary deploy with Render CLI
-        env:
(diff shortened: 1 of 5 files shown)


cschmatzler:
can also use the opportunity to update it; https://github.com/render-oss/cli/releases/tag/v2.14.0

PR #9871, opened 2026-03-16
fix(processor): limit concurrent file reads to prevent emfile errors

fortmarek:
## Summary
- Cap CAS metadata task group concurrency to 50 in `analyzeCacheableTasks` and `analyzeCASOutputs` using a `concurrentMap`/`concurrentCompactMap` helper (same pattern as the CLI's `Array+ExecutionContext`)
- Bump container `nofile` ulimit from default (~1024) to 65536

## Context
Sentry issue [PROCESSOR-1](https://tuist.sentry.io/issues/PROCESSOR-1): `MatchError` on `{:error, :emfile}` from `:zip.unzip`. The root cause is unbounded `withTaskGroup` concurrency in the NIF's CAS metadata analysis — for large builds with hundreds of cache keys/CAS outputs, every key spawned a concurrent task reading files from disk, exhausting file descriptors.

## Test plan
- [x] All 12 Swift NIF tests pass
- [x] All 7 Elixir processor tests pass
- [ ] Deploy to staging and process a large build to verify no `:emfile` errors

### DIFF
--- a/processor/config/deploy.yml
+++ b/processor/config/deploy.yml
@@ -8,6 +8,7 @@ servers:
       publish: 4002:4002
       health-cmd: "curl -f http://localhost:4002/health || exit 1"
       health-interval: 10s
+      ulimit: "nofile=65536:65536"
 
 ssh:
   user: github-actions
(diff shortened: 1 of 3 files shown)


processor/config/deploy.yml
```diff
@@ -8,6 +8,7 @@ servers:
       publish: 4002:4002
       health-cmd: "curl -f http://localhost:4002/health || exit 1"
       health-interval: 10s
+      ulimit: "nofile=65536:65536"
```
cschmatzler, line 11: cache also increases this at kernel level; not sure if needed here (probably not since 65k is below the default) but it's a knob we can tweak in the future

fortmarek: good to know, let's start with this for now 🤞

PR #9888, opened 2026-03-17
Post: Turning AI agents into doctors for your project's health

pepicrft:


server/lib/tuist/marketing/mdex_converter.ex
```diff
@@ -34,12 +35,40 @@ defmodule Tuist.Marketing.MDExConverter do
...
-    [markdown: body]
+  def convert(_path, body, _attrs, _opts), do: body
+
+  def compile_markdown(markdown, path, live?) do
+    document = build_document(markdown)
+    html = MDEx.to_html!(document)
+
+    template =
+      if live? do
+        compile_heex_template(html, path)
```
pepicrft, line 46: We don't compile all of them to templates because some of them have syntax that cause issues with the parsing. So I built it to be opt-in where posts can indicate which ones have live components.

server/priv/marketing/blog/2026/03/18/doctor.md
```diff
@@ -0,0 +1,99 @@
...
+
+![Illustration of a developer overwhelmed by build and performance questions while staring at a computer showing a build in progress.](/marketing/images/blog/2026/03/18/mcp/david.jpg)
+
+## Every platform team has a doctor problem
+
+David's story is not unique. Developers becoming doctors of their product teams' projects, tools, and generated artifacts is extremely common. In fact, we at Tuist were starting to become doctors of our own customers' data. They would come to us asking for solutions to problems that were not directly related to Tuist itself. Helping them felt like the right thing to do, but we soon realized we needed a different approach if we wanted to scale that kind of support. They say that if you listen to and observe your customers closely enough, you can learn a lot about what they need. Turns out we had the best example right in front of us. We just had to connect the dots.
```
fortmarek:
```suggestion
David's story is not unique. Developers becoming doctors of their product teams' projects, tools, and generated artifacts is extremely common. In fact, we at Tuist were starting to become doctors of our own users' data. They would come to us asking for solutions to problems that were not directly related to Tuist itself. Helping them felt like the right thing to do, but we soon realized we needed a different approach if we wanted to scale that kind of support. They say that if you listen to and observe your users closely enough, you can learn a lot about what they need. Turns out we had the best example right in front of us. We just had to connect the dots.
```

In general, I'd prefer framing users > customers. Some users may be customers, but not necessarily so.

server/priv/marketing/blog/2026/03/18/doctor.md
```diff
@@ -0,0 +1,99 @@
...
+
+Think of it as getting a second opinion. The agent can traverse the bundle tree, identify the heaviest nodes, flag duplicated resources, and suggest concrete actions. The interfaces are designed so that coding agents can go as deep and broad as needed without exhausting the context window.
+
+## Optimizing cache usage
+
+Something we keep telling our customers about the cache is that part of the responsibility for making it effective is on us. We ensure low latency and high-bandwidth access from any environment. But the second part is on you, since you need to ensure your project's graph is designed so that tasks with side effects are minimal and tasks are not a parallelization bottleneck. Since Tuist collects cache data from your builds, you can also use that data to optimize your cache usage:
```
fortmarek:
```suggestion
Something we keep telling our users about the cache is that part of the responsibility for making it effective is on us. We ensure low latency and high-bandwidth access from any environment. But the second part is on you, since you need to ensure your project's graph is designed so that tasks with side effects are minimal and tasks are not a parallelization bottleneck. Since Tuist collects cache data from your builds, you can also use that data to optimize your cache usage:
```

fortmarek: > But the second part is on you, since you need to ensure your project's graph is designed so that tasks with side effects are minimal and tasks are not a parallelization bottleneck.

Tasks are mentioned twice in quick succession. Maybe instead of the parallelization bottleneck example, we can mention that the hashes are stable across environments?

PR #9898, opened 2026-03-18
Update homepage hero copy

pepicrft:
Updates the homepage hero section to better reflect Tuist's positioning as a platform team service. The headline now reads "Your mobile platform team, as a service" and the subtitle mentions both Xcode and Gradle support instead of just iOS.

### DIFF
--- a/server/lib/tuist_web/marketing/controllers/marketing_html/home.html.heex
+++ b/server/lib/tuist_web/marketing/controllers/marketing_html/home.html.heex
@@ -22,7 +22,7 @@
         <p>
           {dgettext(
             "marketing",
-            "We focus on optimizing your setup, so your team can focus on shipping"
+            "We virtually join your team to optimize your setup, so you can focus on shipping"
           )}
         </p>
       </div>
--- a/server/lib/tuist_web/marketing/controllers/marketing_controller.ex
+++ b/server/lib/tuist_web/marketing/controllers/marketing_controller.ex
@@ -40,7 +40,7 @@ defmodule TuistWeb.Marketing.MarketingController do
       :head_description,
       dgettext(
         "marketing",
-        "We focus on optimizing your setup, so your team can focus on shipping"
+        "We virtually join your team to optimize your setup, so you can focus on shipping"
       )
     )
     |> assign(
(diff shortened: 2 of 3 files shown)


server/lib/tuist_web/marketing/controllers/marketing_html/home.html.heex
```diff
@@ -11,18 +11,18 @@
     <section data-part="hero">
       <header data-part="title">
         <div data-part="title">
-          <h1>{dgettext("marketing", "Make mobile")}</h1>
+          <h1>{dgettext("marketing", "Your mobile platform team,")}</h1>
```
fortmarek, line 14:
```suggestion
          <h1>{dgettext("marketing", "Your virtual platform team,")}</h1>
```

I like the framing of the virtual platform team, so I wonder if we should use that?

server/lib/tuist_web/marketing/controllers/marketing_html/home.html.heex
```diff
@@ -11,18 +11,18 @@
...
+          <h1>{dgettext("marketing", "as a service")}</h1>
         </div>
       </header>
 
       <div data-part="subtitle">
         <p>
           {dgettext(
             "marketing",
-            "The same iOS tooling that powers apps with millions of users, delivered as a service for your team"
+            "The mobile tooling that powers Xcode and Gradle projects with millions of users, delivered as a service for your team"
```
fortmarek: we have twice "as a service", that doesn't read well

server/lib/tuist_web/marketing/controllers/marketing_html/home.html.heex
pepicrft: I added the "mobile" to narrow the scope and capture the markets we are focusing on now. I think "as a service" captures a bit the "virtual" nature of the solution.

server/lib/tuist_web/marketing/controllers/marketing_html/home.html.heex
pepicrft: Since we say "as a service" above, we can introduce the concept of "being virtual" here. Thoughts?

fortmarek: let's go with it!

PR #9900, opened 2026-03-18
perf(server): optimize test_case_runs ClickHouse query performance

fortmarek:
## Summary

The `test_case_runs` table (ReplacingMergeTree) has `ORDER BY (test_run_id, test_module_run_id, id)` — chosen for deduplication, not query patterns. The PrimaryKey is never used in 9 out of 10 common query patterns (`Condition: true` in EXPLAIN). Projections were added to compensate, but **projections don't work with ReplacingMergeTree** ([ClickHouse issue #46968](https://github.com/ClickHouse/ClickHouse/issues/46968), by-design). All queries rely on bloom filter skip indices, reading 14K–37.6M rows with p50 latencies of 350ms–3.1s.

This PR fixes the root cause with three changes:

### 1. Reorder `test_case_runs` table

**New ORDER BY: `(project_id, test_case_id, ran_at, id)`**

Since `id` is unique per row, any ORDER BY containing `id` correctly deduplicates in ReplacingMergeTree. The remaining columns are chosen to match the dominant query patterns:

- `project_id` — present in nearly every query as a filter
- `test_case_id` — used by flaky test queries, test case detail pages
- `ran_at` — used by time-range filters, branch/CI queries

The migration follows the established pattern from `20260313120000_convert_build_runs_to_replacing_merge_tree.exs`:
1. Drop dependent MVs (`test_case_runs_by_inserted_at`, `test_case_runs_daily_stats`)
2. Introspect columns via `system.columns`
3. Create `test_case_runs_new` with new ORDER BY and explicit skip indices (no projections — useless on ReplacingMergeTree)
4. `INSERT INTO ... SELECT * FROM` with 20-minute timeout
5. Atomic swap via `RENAME TABLE`
6. Recreate both MVs with `POPULATE`
7. `SETTINGS allow_nullable_key = 1` since `test_case_id` is `Nullable(UUID)` (0 actual nulls)

### 2. Dashboard count materialized view

The 4 dashboard functions (`total_test_case_run_count`, `flaky_test_case_run_count`, `last_24h_*` variants) run `count(*) FINAL` over the entire table — 37.6M rows, p50=474ms, called 26K times/day.

New `test_case_runs_dashboard_count` AggregatingMergeTree MV pre-computes daily counts by `is_flaky`:
- `ORDER BY (day, is_flaky)` with `countState()` aggregate
- Dashboard queries now use `countMerge(count)` over ~365 rows/year instead of scanning millions

### 3. Pass `project_id` to point queries

`get_test_case_run_by_id` now accepts an optional `:project_id` keyword. When provided, it adds `WHERE project_id = ?` for PrimaryKey prefix match instead of relying on bloom filters. Callers that already have `project_id` and verify it post-fetch now pass it:

- `TestCaseRunLive.mount/3` — has `project.id` from socket assigns
- `TestCaseRunsController.show/2` — has `selected_project.id`
- `TestCaseRunAttachmentsController.index/2` and `create/2` — has `project.id`

MCP tools (`get_test_case_run.ex`, `list_test_case_run_attachments.ex`) are unchanged — they don't have `project_id` upfront and use the bloom filter fallback via UUIDv7 partition hint.

`get_existing_ci_runs_for_commit` also takes `project_id` now (the caller `check_cross_run_flakiness` has `test.project_id`).

## EXPLAIN indexes comparison (before vs after)

Measured locally after running migrations. The key metric is whether the **PrimaryKey** is used (`Condition: true` = full scan vs an actual condition = binary search).

### Q1: `WHERE id = ?` — get_test_case_run_by_id without project_id (19K/day, p50=3.1s prod)

No change for this fallback path — bloom filter on `idx_id` still handles it:

```
PrimaryKey Condition: (id in [...])  → bloom filter → 0 granules  [same before and after]
```

### Q2: `WHERE git_commit_sha + is_ci + status` — cross-run flakiness (3.6K/day, p50=488ms prod)

**Before:** `PrimaryKey Condition: true` (full scan), then bloom filters
**After:** `PrimaryKey Condition: true` (unchanged without project_id), but Q7 below shows the improvement with project_id

### Q3: `WHERE test_case_id + is_flaky` — flaky test lookup (4/day, p50=3s prod)

**Before:** Used projection `proj_by_test_case_id` (which doesn't actually work on ReplacingMergeTree — returns stale data)
```
ReadFromPreparedSource (proj_by_test_case_id)   ← projection, broken on RMT
```
**After:** PrimaryKey binary search on `test_case_id`:
```
PrimaryKey Keys: test_case_id
  Condition: (test_case_id in [...])
  Parts: 0/14, Granules: 0/14                   ← binary search, correct results
  Search Algorithm: generic exclusion search
```

### Q4: `WHERE project_id + branch + is_ci + ran_at` (4.9K/day, p50=349ms prod)

**Before:** Used projection `proj_test_case_runs_by_project_ran_at` (broken on RMT)
```
ReadFromPreparedSource (proj_test_case_runs_by_project_ran_at)  ← broken projection
```
**After:** PrimaryKey binary search on `project_id` + `ran_at`:
```
PrimaryKey Keys: project_id, ran_at
  Condition: and((ran_at in ['1735689600', +Inf)), (project_id in [1227, 1227]))
  Parts: 0/14, Granules: 0/14                   ← binary search
  Search Algorithm: generic exclusion search
```

### Q5: `count(*) FINAL` — dashboard count (26K/day, p50=474ms prod)

**Before:** Full table scan, 37.6M rows:
```
PrimaryKey Condition: true
  Parts: 17/17, Granules: 17/17                  ← reads everything
```
**After:** Reads from dashboard count MV (~1 granule):
```
ReadFromMergeTree (test_case_runs_dashboard_count)
  PrimaryKey Condition: true
  Parts: 1/1, Granules: 1/1                      ← ~365 rows/year, 1 granule
```

### Q6: `WHERE id + project_id` — get_test_case_run_by_id with project_id (new default)

**Before:** PrimaryKey only on `id` (last in ORDER BY), bloom filter fallback:
```
PrimaryKey Keys: id
  Condition: (id in [...])
  Parts: 16/17, Granules: 16/17                  ← nearly full scan before bloom
```
**After:** PrimaryKey binary search on `project_id` prefix + `id`:
```
PrimaryKey Keys: project_id, id
  Condition: and((project_id in [1227, 1227]), (id in [...]))
  Parts: 0/14, Granules: 0/14                    ← binary search, 0 granules
  Search Algorithm: generic exclusion search
```

### Q7: `WHERE project_id + git_commit_sha + is_ci + status` — get_existing_ci_runs_for_commit with project_id (new)

**Before:** `PrimaryKey Condition: true` (full scan), bloom filters only
**After:** PrimaryKey binary search on `project_id`:
```
PrimaryKey Keys: project_id
  Condition: (project_id in [1227, 1227])
  Parts: 0/14, Granules: 0/14                    ← binary search
  Search Algorithm: binary search
```

## Summary table

| Query | Before PrimaryKey | After PrimaryKey |
|---|---|---|
| `WHERE id = ?` (no project_id) | `id` exclusion search + bloom | Same (fallback path) |
| `WHERE id + project_id` (19K/day) | `id` exclusion → 16/17 parts | `project_id, id` → **0/14 parts** |
| `WHERE test_case_id + is_flaky` | Broken projection | `test_case_id` → **0/14 parts** |
| `WHERE project_id + branch + ran_at` | Broken projection | `project_id, ran_at` → **0/14 parts** |
| `count(*) FINAL` (26K/day) | 17/17 granules (full scan) | MV: **1/1 granule** (~365 rows) |
| `WHERE project_id + commit + ci` | `Condition: true` (full scan) | `project_id` binary → **0/14 parts** |

## Test plan

- [x] Migrations run successfully locally
- [x] `SELECT sorting_key FROM system.tables WHERE name = 'test_case_runs'` returns `project_id, test_case_id, ran_at, id`
- [x] EXPLAIN indexes confirms PrimaryKey usage on all key query patterns (see above)
- [ ] Run tests: `cd server && mix test test/tuist/tests_test.exs`
- [ ] Start dev server, verify dashboard counts load correctly
- [ ] Verify test case run detail page works
- [ ] Deploy to staging and compare p50/p99 latencies vs baseline

### DIFF
--- a/server/lib/tuist/tests.ex
+++ b/server/lib/tuist/tests.ex
@@ -536,8 +536,13 @@ defmodule Tuist.Tests do
     query =
       case Keyword.get(opts, :project_id) do
         nil ->
-          month = uuidv7_to_yyyymm(id)
-          where(query, [tcr], fragment("toYYYYMM(?)", tcr.inserted_at) == ^month)
+          case uuidv7_to_yyyymm(id) do
+            {:ok, month} ->
+              where(query, [tcr], fragment("toYYYYMM(?)", tcr.inserted_at) == ^month)
+
+            :error ->
+              query
+          end
 
         project_id ->
           where(query, [tcr], tcr.project_id == ^project_id)
@@ -563,8 +568,13 @@ defmodule Tuist.Tests do
   defp uuidv7_to_yyyymm(uuid_string) do
     hex = uuid_string |> String.replace("-", "") |> String.slice(0, 12)
     timestamp_ms = String.to_integer(hex, 16)
-    {:ok, datetime} = DateTime.from_unix(timestamp_ms, :millisecond)
-    datetime.year * 100 + datetime.month
+
+    case DateTime.from_unix(timestamp_ms, :millisecond) do
+      {:ok, datetime} -> {:ok, datetime.year * 100 + datetime.month}
+      _ -> :error
+    end
+  rescue
+    _ -> :error
   end
 
   defp create_test_modules(test, test_modules) do


server/priv/ingest_repo/migrations/20260319120001_create_dashboard_count_mv.exs
```diff
@@ -0,0 +1,38 @@
+defmodule Tuist.IngestRepo.Migrations.CreateDashboardCountMv do
```
pepicrft: Considering experiences we had in the past with failing ClickHouse migrations. What if we separate these two migrations in two separate PRs.

fortmarek: yeah, can do

server/lib/tuist/tests.ex
```diff
@@ -562,13 +563,8 @@ defmodule Tuist.Tests do
   defp uuidv7_to_yyyymm(uuid_string) do
     hex = uuid_string |> String.replace("-", "") |> String.slice(0, 12)
```
pepicrft, line 569: I believe we have callers that don't do any validation, which means we can end up in this code path with a `uuid_string` that's not a `uuid_string`. I'd design this to account for that scenario, and let the error bubble up instead of raising, which can manifest for example in a failing MCP tool call (I believe some tools are calling `get_test_case_run_by_id` without any validation.

server/priv/ingest_repo/migrations/20260319120000_reorder_test_case_runs.exs
```diff
@@ -0,0 +1,151 @@
...
+          INDEX idx_git_commit_sha (git_commit_sha) TYPE bloom_filter GRANULARITY 1,
+          INDEX idx_test_run_id (test_run_id) TYPE bloom_filter GRANULARITY 4,
+          INDEX idx_is_ci (is_ci) TYPE set(2) GRANULARITY 1,
+          INDEX idx_status (status) TYPE set(3) GRANULARITY 1,
+          INDEX idx_git_branch (git_branch) TYPE bloom_filter GRANULARITY 1,
+          INDEX idx_is_flaky (is_flaky) TYPE set(2) GRANULARITY 1\
+        """
+
+        IngestRepo.query!("""
+        CREATE TABLE test_case_runs_new (
```
pepicrft, line 46: The PR description says that the projections were intentionally dropped since they are ineffective on `ReplacingMergeTree` but have you benchmarked it against the current projections? `proj_by_branch_ci`,  `proj_by_id`, `proj_by_project_flaky`?

fortmarek: Some of them are being used, but rarely so (highlighted by the fact that most queries are slow against the same table). We can bring some of them back if needed, but the better ordering should help.

PR #9901, opened 2026-03-18
perf(server): use dashboard count MV for global test case run counts

fortmarek:
## Summary

Follow-up to #9900. Uses the `test_case_runs_dashboard_count` AggregatingMergeTree MV (created in the parent PR's migration) to replace expensive `count(*) FINAL` full-table scans in the 4 dashboard count functions.

- `total_test_case_run_count/0` → `countMerge(count)` from MV
- `flaky_test_case_run_count/0` → `countMerge(count) WHERE is_flaky = true`
- `last_24h_test_case_run_count/0` → `countMerge(count) WHERE day >= yesterday`
- `last_24h_flaky_test_case_run_count/0` → `countMerge(count) WHERE day >= yesterday AND is_flaky = true`

### Before
- Queries scan 37.6M rows with `count(*) FINAL` (p50=474ms, 26K calls/day)
- EXPLAIN: `PrimaryKey Condition: true`, all parts/granules read

### After
- Queries read ~365 pre-aggregated rows from the MV (~1 granule)
- EXPLAIN: `ReadFromMergeTree (test_case_runs_dashboard_count)`, Parts: 1/1, Granules: 1/1

## Test plan

- [ ] Ensure parent PR #9900 migration has run (creates the MV)
- [ ] Run tests: `cd server && mix test test/tuist/tests_test.exs`
- [ ] Start dev server, verify dashboard counts load correctly


server/lib/tuist/tests.ex
```diff
@@ -68,21 +67,23 @@ defmodule Tuist.Tests do
   end
 
   def last_24h_test_case_run_count do
-    twenty_four_hours_ago = DateTime.add(DateTime.utc_now(), -24, :hour)
+    yesterday = Date.add(Date.utc_today(), -1)
 
     ClickHouseRepo.one(
-      from(t in TestCaseRun, hints: ["FINAL"], where: t.inserted_at >= ^twenty_four_hours_ago, select: count())
+      from(d in TestCaseRunDashboardCount,
+        where: d.day >= ^yesterday,
```
pepicrft, line 74: If I'm not mistaken, I think this changes the behavior from a rolling 24-window to whole-day buckets. Is this intentional?

server/lib/tuist/tests.ex
```diff
@@ -49,15 +50,13 @@ defmodule Tuist.Tests do
   end
 
   def total_test_case_run_count do
-    TestCaseRun
-    |> from(hints: ["FINAL"], select: count())
-    |> ClickHouseRepo.one() || 0
+    ClickHouseRepo.one(from(d in TestCaseRunDashboardCount, select: fragment("countMerge(count)"))) || 0
```
pepicrft, line 53: These counts used to run against `test_case_runs FINAL` deduping the logical rows. If this is now being read from the MW built from the raw inserts into the `ReplacingMergeTree`, can this double-count those rows compared to the old behavior?

server/lib/tuist/tests.ex
fortmarek: I don't have a better idea to preserve the current logic, would rather make this slightly inaccurate, while not straining our clickhouse instance

Issue #6048, opened 2024-03-07
Support for `.xcstrings` catalog


danieleformichelli:
### What problem or need do you have?

They were added in Xcode 15, see https://github.com/tuist/tuist/discussions/5221#discussioncomment-6103709

### macOS version

14.4

### Tuist version

4.5.0

### Xcode version

15.3

muukii:
what's the problem actually? Xcode will produce strings or stringsdict automatically if we put a xcstrings.

leszko11:
Hey @muukii 👋🏼
Are you sure about that? I have never used strings catalogs, but after a quick test on `ios_app_with_frameworks_and_resources` I see that whenever I migrate `.strings`to a strings catalog, it doesn’t exist in a project anymore.
I think the point of this issue is to include strings catalog file in a generated workspace, because now it’s not included after generation. 
However, swiftgen does not support strings catalog yet according to [this issue](https://github.com/SwiftGen/SwiftGen/issues/1065). What should we do about it?

muukii:
In my case, using different name to `Localizable` because I still have classic style strings file.
Specify table name as `StringCatalog` to load. Not sure it's making some effects.

![CleanShot 2024-03-14 at 13 06 23@2x](https://github.com/tuist/tuist/assets/1888355/a2a448cd-669a-4491-b135-43afcd4cb8f1)
![CleanShot 2024-03-14 at 13 07 34@2x](https://github.com/tuist/tuist/assets/1888355/00aefab1-c435-4bd6-8de3-ffa30828a3e9)

leszko11:
Yes, that's the possible solution, but in the context of tuist capabilities, it looks a bit like a workaround to hold both of these. I think 

Just out of curiosity, are `StringCatalog.strings` and `StringCatalog.xcstrings` update each other, or if you want to add a new key, you need add it in both files?

leszko11:
A little update on this one.

As I mentioned before _SwiftGen_ does not support `.xcstrings` files yet. 
I've tried to implement the solution mentioned [here](https://github.com/SwiftGen/SwiftGen/issues/1065#issuecomment-1773334930), however, flow `.xcstrings` -> `.strings` -> `.swift` does not fit into current resource synthesizer generator, because it requires running it twice. In addition, it doesn't support plurals(`.stringsdict`)

[...]

PR #9907, opened 2026-03-18
fix(cli): prevent xcstrings stale extraction state in static targets

pepicrft:
## Context

When static frameworks have `.xcstrings` files with a companion resource bundle, building from Xcode marks all strings as `"extractionState": "stale"`. This was reported in #6048.

## Root cause

Tuist was adding `.xcstrings` to the main target's **Sources build phase** (same as `.xcassets`) for symbol generation. This triggered Xcode's string extraction, which scans the Sources phase for string references. Since the strings use `bundle: .module` (pointing to the companion bundle), Xcode couldn't match them and marked everything stale.

Key insight from studying `swift-build` and `swift-package-manager`: unlike `.xcassets`, `.xcstrings` files should NOT be in the Sources build phase. Xcode handles string catalog compilation and symbol generation from the Resources phase via `xcstringstool`.

## Fix

- Remove `"xcstrings"` from `codeGeneratingResourceExtensions` so they're no longer added to the Sources build phase
- Keep `.xcstrings` in the main target's **Resources phase** so Xcode's string catalog editor can correctly associate string references in Swift code with the catalog entries
- The `.xcstrings` also stays in the companion bundle target's Resources for runtime compilation into `.strings` files

## Verification

Tested with a reproduction project (static framework + xcstrings + companion bundle):
- Before: building from Xcode adds `"extractionState": "stale"` to all strings
- After: xcstrings files remain unmodified after build
- Strings are correctly compiled into the bundle and available at runtime

### DIFF
--- a/cli/Tests/TuistGeneratorAcceptanceTests/GenerateAcceptanceTests.swift
+++ b/cli/Tests/TuistGeneratorAcceptanceTests/GenerateAcceptanceTests.swift
@@ -1213,14 +1213,15 @@ struct GenerateAcceptanceTestiOSAppWithStaticFrameworkWithXcstrings {
         let fixturePath = try fixtureDirectory()
         try await run(GenerateCommand.self)
 
+        let fileSystem = FileSystem()
         let xcstringsPath = fixturePath.appending(
             components: "StaticFramework", "Resources", "Localizable.xcstrings"
         )
-        let contentBeforeBuild = try Data(contentsOf: URL(fileURLWithPath: xcstringsPath.pathString))
+        let contentBeforeBuild = try await fileSystem.readFile(at: xcstringsPath)
 
         try await run(BuildCommand.self)
 
-        let contentAfterBuild = try Data(contentsOf: URL(fileURLWithPath: xcstringsPath.pathString))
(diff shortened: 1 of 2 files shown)


cli/Tests/TuistGeneratorAcceptanceTests/GenerateAcceptanceTests.swift
```diff
@@ -1207,6 +1207,38 @@ struct GeneratediOSStaticLibraryWithStringResources {
...
+struct GenerateAcceptanceTestiOSAppWithStaticFrameworkWithXcstrings {
+    @Test(.withFixture("generated_ios_app_with_static_framework_with_xcstrings"), .inTemporaryDirectory)
+    func ios_app_with_static_framework_with_xcstrings() async throws {
+        let fixturePath = try fixtureDirectory()
+        try await run(GenerateCommand.self)
+
+        let xcstringsPath = fixturePath.appending(
+            components: "StaticFramework", "Resources", "Localizable.xcstrings"
+        )
+        let contentBeforeBuild = try Data(contentsOf: URL(fileURLWithPath: xcstringsPath.pathString))
```
fortmarek: nit: can use fileSystem instead

PR #9908, opened 2026-03-18
fix(server): make /cache/ac endpoint fire-and-forget to eliminate DB reads

fortmarek:
## Summary

- The `/cache/ac` endpoint was doing a synchronous DB SELECT (`get_cache_action_item`) inside a Cachex transaction lock on every request, just to distinguish `201` vs `200` responses
- In CI bursts with 100+ unique artifact hashes, this caused 100+ concurrent DB reads, exhausting the connection pool — observed as 60s Cloudflare 502s in production
- The inserts were already batched through the Broadway pipeline (`insert_all` in batches of 20); the SELECT was entirely redundant since the CLI doesn't act differently on 201 vs 200
- Now the endpoint is truly fire-and-forget: push to pipeline, return 201, done — `ON CONFLICT DO NOTHING` handles duplicate hashes safely

## Test plan

- [ ] Existing cache controller tests pass (26 tests, 0 failures)
- [ ] Removed the now-irrelevant "returns ok if already exists" and "returns 201 for CLI 4.28.0" tests; added "returns created even if already exists" covering the new unified behavior

### DIFF
--- a/server/lib/tuist_web/controllers/api/cache_controller.ex
+++ b/server/lib/tuist_web/controllers/api/cache_controller.ex
@@ -354,58 +354,20 @@ defmodule TuistWeb.API.CacheController do
         %{assigns: %{selected_project: selected_project}, body_params: %{hash: hash}} = conn,
         _params
       ) do
-    cache_action_item =
-      Tuist.KeyValueStore.get_or_update(
-        [
-          Atom.to_string(__MODULE__),
-          "upload_cache_action_item",
-          selected_project.id,
-          hash
-        ],
-        [
-          ttl: Map.get(conn.assigns, :cache_ttl, to_timeout(minute: 1)),
-          cache: Map.get(conn.assigns, :cache, :tuist),
-          locking: true
-        ],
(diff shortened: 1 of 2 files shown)


server/lib/tuist_web/controllers/api/cache_controller.ex
```diff
@@ -354,58 +354,20 @@ defmodule TuistWeb.API.CacheController do
...
-             }}
-          )
-
-        conn
-        |> put_status(:created)
-        |> json(%{
-          hash: hash
-        })
-
-      conn |> get_req_header("x-tuist-cli-version") |> List.first() == "4.28.0" ->
```
cschmatzler, line 397: are we good with yeeting the version-specific handler?

pepicrft (approved):

server/lib/tuist_web/controllers/api/cache_controller.ex
fortmarek: I think so given how old this cli version is

PR #9916, opened 2026-03-18
feat(cli): resolve inspect bundle app names like share

pepicrft:
## Summary
- make `tuist inspect bundle` accept an app name on macOS and resolve it from built products the same way `tuist share` does
- keep the existing explicit bundle-path behavior intact
- add CLI options and env keys for configuration, platforms, and derived data path, plus service coverage for app-name resolution

## Example
```sh
tuist xcodebuild build -scheme abb -workspace Loyalty.xcworkspace -configuration Debug -sdk iphonesimulator -- -derivedDataPath ./DerivedData
tuist inspect bundle abb --platforms ios --configuration Debug --derived-data-path ./DerivedData
tuist share abb --derived-data-path ./DerivedData
```

## Validation
- added focused service coverage for resolving an app name to a built `.app`
- `swift test --filter InspectBundleCommandServiceTests` is currently blocked by a pre-existing SwiftPM graph issue in the repo: conflicting `FileSystem` products declared by two packages

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -103,6 +103,7 @@ var tuistServerDependencies: [Target.Dependency] = [
     "TuistOpener",
     "TuistUniqueIDGenerator",
     "TuistSupport",
+    "TuistSimulator",
     fileSystemDependency,
     mockableDependency,
     pathDependency,
@@ -373,7 +374,7 @@ tuistAuthCommandDependencies.append(contentsOf: ["TuistLoader", "TuistSupport"])
 tuistServerDependencies.append(contentsOf: [
     "TuistCore", "TuistProcess", "TuistCI",
     "TuistAutomation", "TuistXCActivityLog",
-    "TuistXCResultService", "TuistSimulator",
+    "TuistXCResultService",
     xcodeGraphDependency,
 ])
 tuistHTTPDependencies.append(contentsOf: ["TuistSupport", "TuistHAR"])


cli/Sources/TuistXcodeBuildProducts/BuiltAppBundleLocator.swift
```diff
@@ -0,0 +1,126 @@
+import FileSystem
+import Foundation
+import Mockable
+import Path
+import TuistSimulator
+import TuistSupport
+import XcodeGraph
+
+public struct BuiltAppBundle: Equatable, Sendable {
```
fortmarek: Any reason why not to use the `AppBundle` from `TuistAutomation`?

cli/Sources/TuistInspectCommand/Commands/InspectBundleCommand.swift
```diff
@@ -15,12 +16,33 @@ public struct InspectBundleCommand: AsyncParsableCommand {
...
+        envKey: .inspectBundlePlatform
+    )
+    var platforms: [XcodeGraph.Platform] = []
+
+    @Option(
+        help: "The derived data path to find the app in. When absent, the system-configured one.",
+        completion: .directory,
+        envKey: .inspectBundleDerivedDataPath
+    )
+    var derivedDataPath: String?
```
fortmarek: these options don't make sense on Linux. I would make sure they appear just on macOS.

cli/Sources/TuistXcodeBuildProducts/XcodeProjectBuildDirectoryLocator.swift
```diff
@@ -2,19 +2,9 @@ import Foundation
...
-    /// Locates the build output directory for `xcodebuild` command.
-    ///
-    /// For example: `~/Library/Developer/Xcode/DerivedData/PROJECT_NAME/Build/Products/CONFIG_NAME`
-    ///
-    /// - Parameters:
-    ///   - destinationType: The destination platform for the built scheme.
-    ///   - projectPath: The path of the Xcode project or workspace.
-    ///   - derivedDataPath: The path of the derived data
-    ///   - configuration: The configuration name, i.e. `Release`, `Debug`, or something custom.
     func locate(
```
fortmarek, line 17: why did this comment go away? is that no longer accurate?

cli/Sources/TuistInspectCommand/Services/InspectBundleCommandService.swift
```diff
@@ -104,6 +152,47 @@ public struct InspectBundleCommandService {
...
+        _ bundle: String,
+        path: AbsolutePath,
+        configuration: String?,
+        platforms: [Platform],
+        derivedDataPath: String?
+    ) async throws -> AbsolutePath {
+        let currentWorkingDirectory = try await Environment.current.currentWorkingDirectory()
+        let explicitPath = try AbsolutePath(validating: bundle, relativeTo: currentWorkingDirectory)
+
+        if try await fileSystem.exists(explicitPath) || looksLikeBundlePath(explicitPath) {
```
fortmarek: I don't fully follow this condition. If we pass an explicit path, why do additionally check `looksLikeBundlePath`? I think if an explicit path is passed, we should just try to use it (and throw if it doesn't exist)

cli/Sources/TuistKit/Utils/AppBundleTargetResolver.swift
```diff
@@ -0,0 +1,216 @@
...
+        self.fileSystem = fileSystem
+    }
+
+    public func resolve(
+        app: String?,
+        path: AbsolutePath,
+        configuration: String?,
+        platforms: [Platform],
+        derivedDataPath: AbsolutePath?
+    ) async throws -> ResolvedAppBundleTarget {
```
fortmarek, line 100: I would expect `AppBundlePathResolver` to resolve a path, but it resolves the target.

cli/Sources/TuistInspectCommand/Services/InspectBundleCommandService.swift
```diff
@@ -104,6 +152,47 @@ public struct InspectBundleCommandService {
...
+            return explicitPath
+        }
+
+        let resolvedDerivedDataPath = try derivedDataPath.map {
+            try AbsolutePath(validating: $0, relativeTo: fileHandler.currentPath)
+        }
+
+        let resolved = try await appBundlePathResolver.resolve(
```
fortmarek: `let resolved` is such an odd naming that makes the code quite hard to follow ...

cli/Sources/TuistInspectCommand/Services/InspectBundleCommandService.swift
```diff
@@ -104,6 +152,47 @@ public struct InspectBundleCommandService {
...
+            app: bundle,
+            path: path,
+            configuration: configuration,
+            platforms: platforms,
+            derivedDataPath: resolvedDerivedDataPath
+        )
+
+        return try await builtAppBundleLocator.locateBuiltAppBundlePath(
```
fortmarek: instead of `buildAppBundleLocator`, I'd consider having a single service for `BuildProductService` with a method of `appBundlePath(...)`. I find the naming a bit non-conventional.

cli/Sources/TuistXcodeBuildProducts/BuiltAppBundleLocator.swift
pepicrft: Not really. I consolidated both of them into a single symbol.

cli/Sources/TuistInspectCommand/Commands/InspectBundleCommand.swift
pepicrft: I'll use compiler directives and exclude them when compiling for other platforms.

cli/Sources/TuistXcodeBuildProducts/XcodeProjectBuildDirectoryLocator.swift
pepicrft: I brought this back.

cli/Sources/TuistInspectCommand/Services/InspectBundleCommandService.swift
pepicrft: I dropped the exists check, it's now just `looksLikeBundlePath`.

On using the path directly: the tricky bit is that bundle is overloaded. It can be either a target name (e.g. App) or an explicit path (e.g. `./build/App.app`). We need some heuristic to disambiguate the two before deciding whether to resolve via the build products locator or treat it as a path.

The extension check (.app, .xcarchive, .ipa, .aab, .apk) is the signal: target names won't carry those extensions, so if we see one we know the user meant a path. If we relied on exists alone, a target name that happened to collide with a file/dir in cwd would silently get routed to "path mode" and then fail with a confusing error. With the extension heuristic, a missing App.app cleanly throws a "path doesn't exist" error, and a bare App always goes through target resolution.

cli/Sources/TuistKit/Utils/AppBundleTargetResolver.swift
pepicrft: I renamed this to `AppBundleTargetResolver` and `AppBundleTargetResolving` to align with the function signature.

cli/Sources/TuistInspectCommand/Services/InspectBundleCommandService.swift
pepicrft: Renamed to `appBundleTarget`

cli/Sources/TuistInspectCommand/Services/InspectBundleCommandService.swift
pepicrft: Renamed `BuiltAppBundleLocator` to `BuildProductService` with an `appBundlePath(...)` method (in `TuistXcodeBuildProducts`). The inspect/share services now call `BuildProductService.appBundlePath(...)` with the resolved target.

pepicrft:
@fortmarek this is ready for another review.

PR #9930, opened 2026-03-19
chore(server): add shared credo rule to disallow import/alias/require inside function bodies

fortmarek:
## Summary
- Adds a custom Credo check `DisallowDirectivesInFunction` that flags `import`, `alias`, and `require` statements inside `def`/`defp` function bodies
- Directives inside `quote` blocks (Phoenix macro helpers) are correctly excluded
- Lives in `tuist_common/credo/checks/` so it's shared across server, cache, and processor
- Adds `.credo.exs` to processor (it had credo as a dependency but no config)

## Example

These will be flagged:

```elixir
def my_func do
  import Ecto.Query      # ❌
  alias Tuist.Projects   # ❌
  require Logger         # ❌
  ...
end
```

This is fine:

```elixir
defmodule MyModule do
  import Ecto.Query      # ✅ Module level
  alias Tuist.Projects   # ✅
  require Logger         # ✅

  def my_func do
    ...
  end
end
```

## Test plan

Verified e2e by creating a test file with all three directives inside function bodies:

```
$ mix credo --only Credo.Checks.DisallowDirectivesInFunction lib/test_directives_check.ex

  Warnings - please take a look
┃
┃ [W] → Move `alias` to the module level instead of inside a function body.
┃       lib/test_directives_check.ex:16:5 #(TestDirectivesModule.bad_alias)
┃ [W] → Move `require` to the module level instead of inside a function body.
┃       lib/test_directives_check.ex:21:5 #(TestDirectivesModule.bad_require)
┃ [W] → Move `import` to the module level instead of inside a function body.
┃       lib/test_directives_check.ex:11:5 #(TestDirectivesModule.bad_import)

5 mods/funs, found 3 warnings.
```

Also verified:
- [x] Full `mix credo` passes on server (1033 files), cache (138 files), and processor (16 files) with 0 issues
- [x] Phoenix macro helpers in `lib/tuist_web.ex` with directives inside `quote` blocks are not flagged


server/.credo.exs
```diff
@@ -24,6 +25,7 @@ alias Credo.Checks.TimestampsType
           {TimestampsType, files: %{included: ["priv/repo/migrations/"]}, allowed_type: :timestamptz},
           {TimestampsType, files: %{included: ["lib/"]}, allowed_type: :utc_datetime},
           {DisallowSpec, []},
+          {DisallowImportInFunction, files: %{included: ["lib/"]}},
```
pepicrft: Can we reuse this in `cache/`?

PR #9932, opened 2026-03-19
fix(server): resolve noora assets from source and fix trivy version

pepicrft:
Just noticed serving the app failed in a clean clone because the Noora assets hadn't been built. Since we are not doing any JS/CSS transformation at build-time other than just bundling when doing a release of the server, I thought we could just let the browser resolve the CSS and JS graph and skip the need to build anything in development.

> [!IMPORTANT]
> I checked that the release bundle tasks bundle `noora.js` and `noora.css` continue to bundle noora successfully.

## How to test it locally

Just serve the app. The browser should be able to load the JS and the CSS without any errors in the console. 

## Summary
- In dev, esbuild now aliases `noora` directly to the source entry point (`noora/js/index.js`) instead of the pre-built bundle (`noora/priv/static/noora.js`)
- Removes the separate `esbuild_noora` JS watcher since esbuild resolves the source graph inline
- Adds `pnpm` pre-build of noora in `assets.deploy` for production where the bundled output is still needed

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -38,10 +38,12 @@ COPY server/pnpm-lock.yaml /app/pnpm-lock.yaml
 WORKDIR /app
 RUN --mount=type=cache,id=pnpm,target=/pnpm/store pnpm install --frozen-lockfile
 
-# Install noora dependencies for source resolution by esbuild
-COPY noora/package.json noora/pnpm-lock.yaml /noora/
+# Build noora static assets (JS + CSS) from source
+COPY noora /noora
 WORKDIR /noora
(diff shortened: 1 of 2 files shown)


server/mise.toml
```diff
@@ -6,7 +6,7 @@ TUIST_SECRET_KEY_BASE = "rpux4+W/oBey0drSFOctyIbppOG2A9VUUuTMC1WaM8xZgYxA6gg4yry
   node = "24.11.0"
   sops = "3.9.3"
   age = "1.2.1"
-  "ubi:aquasecurity/trivy" = "0.63.0"
+  "ubi:aquasecurity/trivy" = "0.69.3"
```
pepicrft, line 9: The version 0.63.0 doesn't exist.

PR #9935, opened 2026-03-19
feat(server): add shared HTTP failure observability

pepicrft:
## Summary
- add a shared `TuistCommon.HTTP.TransportPromExPlugin` so `server` and `cache` consume the same native Bandit and Thousand Island transport metrics
- keep the metric surface narrow with low-cardinality counters for request timeouts, request failures, connection drops, and recv/send errors
- bound route cardinality to `phoenix_route || "unknown"` so incidents do not create per-path time series

## Verification
- `cd tuist_common && mix test test/tuist_common/http/transport_prom_ex_plugin_test.exs`
- `cd server && mix compile`
- `cd cache && mix compile`

## Notes
- `server`, `cache`, and `tuist_common` now point at the same `prom_ex` source so the shared plugin can live in `tuist_common` without mixed-source dependency conflicts. We are still pinned to the fork because the Finch plugin work and its follow-up crash fixes live on [`pepicrft/prom_ex`](https://github.com/pepicrft/prom_ex/tree/finch) and do not appear to be in upstream yet.
- `Body read timeout` detection is based on Bandit’s current emitted error message in [`bandit/http1/socket.ex#L335`](https://github.com/pepicrft/tuist4/blob/main/server/deps/bandit/lib/bandit/http1/socket.ex#L335) and [`bandit/http1/socket.ex#L344`](https://github.com/pepicrft/tuist4/blob/main/server/deps/bandit/lib/bandit/http1/socket.ex#L344), which Bandit forwards into telemetry metadata via [`bandit/pipeline.ex#L224`](https://github.com/pepicrft/tuist4/blob/main/server/deps/bandit/lib/bandit/pipeline.ex#L224).

> [!NOTE]
> This PR intentionally keeps transport observability out of Sentry by default. The new Bandit and Thousand Island metrics/logs are meant to cover connection drops, body read timeouts, and recv/send failures, where trend detection and request/connection correlation matter more than exception capture. Sentry remains the better fit for unexpected request exceptions and stack-trace-driven debugging.

### DIFF
--- a/tuist_common/test/tuist_common/http/transport_logger_test.exs
+++ b/tuist_common/test/tuist_common/http/transport_logger_test.exs
@@ -1,11 +1,24 @@
 defmodule TuistCommon.HTTP.TransportLoggerTest do
-  use ExUnit.Case, async: true
+  use ExUnit.Case, async: false
 
   import ExUnit.CaptureLog
 
   alias TuistCommon.HTTP.TransportLogger
 
-  test "logs Bandit body read timeouts with route and connection context" do
+  setup do
+    handler_suffix = System.unique_integer([:positive])
+    TransportLogger.attach(handler_suffix)
+
+    on_exit(fn ->
+      TransportLogger.detach(handler_suffix)
+    end)
+
+    %{handler_suffix: handler_suffix}
+  end
+
+  test "logs Bandit body read timeouts with route and connection context", %{
+    handler_suffix: _handler_suffix
+  } do
     log =
       capture_log(
         [
@@ -13,7 +26,7 @@ defmodule TuistCommon.HTTP.TransportLoggerTest do
           metadata: [:request_id, :method, :route, :connection_span_context]
         ],
         fn ->
-          TransportLogger.handle_event(
+          :telemetry.execute(
             [:bandit, :request, :stop],
             %{duration: System.convert_time_unit(2, :millisecond, :native), req_body_bytes: 128},
             %{
@@ -25,8 +38,7 @@ defmodule TuistCommon.HTTP.TransportLoggerTest do
                 private: %{phoenix_route: "/upload"},
                 resp_headers: [{"x-request-id", "req_123"}]
               }
-            },
-            nil
+            }
           )
         end
       )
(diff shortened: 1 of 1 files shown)


cache/mix.exs
```diff
@@ -54,7 +54,7 @@ defmodule Cache.MixProject do
       {:phoenix_pubsub, "~> 2.1"},
       {:plug, "~> 1.18"},
       {:plug_cowboy, "~> 2.7"},
-      {:prom_ex, "~> 1.10"},
+      {:prom_ex, git: "https://github.com/pepicrft/prom_ex", branch: "finch"},
```
pepicrft, line 57: I'm aligning it with the version that we use in `server/mix.exs`.

tuist_common/lib/tuist_common/http/transport.ex
```diff
@@ -0,0 +1,150 @@
...
+      true -> nil
+    end
+  end
+
+  def bandit_request_metadata(metadata) do
+    conn = metadata[:conn]
+
+    %{
+      method: (conn && conn.method) || "unknown",
+      route: (conn && conn.private[:phoenix_route]) || "unknown"
```
pepicrft, line 55: We `phoenix_route` over the request path to reduce the cardinality.

tuist_common/lib/tuist_common/http/transport_logger.ex
```diff
@@ -0,0 +1,72 @@
+defmodule TuistCommon.HTTP.TransportLogger do
```
pepicrft, line 1: I added logging so we can go deeper debugging if we notice spikes in Bandit or ThousandIsland errors. The logs include the request_id, so we can trace it if needed.

tuist_common/test/tuist_common/http/transport_logger_test.exs
```diff
@@ -0,0 +1,82 @@
+defmodule TuistCommon.HTTP.TransportLoggerTest do
+  use ExUnit.Case, async: false
```
pepicrft, line 2: I went with an integration test here sending real telemetry events, which means we can't run this in parallel with other test. So we are trading the ability to parallelize with having a more realistic test scenario, which I think it's wroth it.

cache/mix.exs
fortmarek: any luck on getting our changes merged?

pepicrft: I haven't done it yet, but yeah, I'll look into pushing that upstream.

pepicrft:
Tested this in staging.

PR #9936, opened 2026-03-19
fix(cli): resolve swift package graph namespaces

pepicrft:
## Summary

Fix Swift package graph resolution in Tuist for packages whose dependency/product references use different names across:
- workspace package refs
- SwiftPM dependency identities
- manifest package names

This is scoped to package graph resolution only.
It does not include FileSystem API changes or unrelated filesystem cleanups.

## Problem

The `FileSystem` branch that depends on `swift-file-system` compiles in its own SwiftPM context, but Tuist failed while generating or resolving the generated Xcode workspace graph.

The root issue was that Tuist resolved external package products with inconsistent package namespaces, which broke cases like:
- duplicate product names across different packages
- package dependency identities that differ from workspace package ref names
- unqualified external product lookups when the product name is unique

## Changes

- normalize target product dependency package names in `SwiftPackageManagerGraphLoader`
- resolve external products by package-qualified key when needed in `PackageInfoMapper`
- preserve unqualified lookup for unique external product names
- add regressions covering duplicate package/product namespaces and mixed naming schemes

## Validation

- `../FileSystem` builds directly with `swift build --replace-scm-with-registry`
- with this Tuist fix applied, `tuist generate --no-open` succeeds while pointing Tuist at the `FileSystem` branch that depends on `swift-file-system`
- generated graph includes the expected distinct package projects, including `swift-file-system` and `coenttb.swift-async-algorithms-fork`

### DIFF
--- a/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
+++ b/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
@@ -102,6 +102,8 @@ public struct SwiftPackageManagerGraphLoader: SwiftPackageManagerGraphLoading {
             (
                 id: String,
                 name: String,
+                identityName: String,
+                manifestName: String,
                 folder: AbsolutePath,
                 targetToArtifactPaths: [String: AbsolutePath],
                 info: PackageInfo,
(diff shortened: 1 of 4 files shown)


pepicrft:
Closing for now. The investigation showed the issue is in Swift package graph resolution when different packages expose the same product name across different package namespaces, but I’m not moving this draft forward in its current form.

PR #9937, opened 2026-03-19
fix(server): fix broken marketing links

pepicrft:
## Summary
- Footer "Features" links now point to marketing feature pages where available (`/cache`, `/build-insights`, `/selective-testing`, `/flaky-tests`, `/test-insights`, `/previews`), and use correctly localized docs URLs for Generated projects, Registry, and Bundle analysis
- Home page and banner "Get started" buttons now link to `/users/register` instead of docs
- Fix `localized_href` for `docs.tuist.dev` URLs — it was incorrectly prepending `/docs` to paths (producing `/en/docs/guides/...` instead of `/en/guides/...`)
- `Marketing.Stats.get_stats/0` gracefully returns zeroes when the GenServer isn't running (dev/test) instead of crashing

### DIFF
--- a/server/priv/gettext/dashboard_account.pot
+++ b/server/priv/gettext/dashboard_account.pot
@@ -730,7 +730,7 @@ msgstr ""
 msgid "United States"
 msgstr ""
 
-#: lib/tuist/vcs.ex:493
+#: lib/tuist/vcs.ex:500
 #, elixir-autogen, elixir-format
 msgid "Unknown"
 msgstr ""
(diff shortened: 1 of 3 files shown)


server/lib/tuist/marketing/stats.ex
```diff
@@ -14,8 +14,20 @@ defmodule Tuist.Marketing.Stats do
     GenServer.start_link(__MODULE__, %{}, name: __MODULE__)
   end
 
+  @default_stats %{
```
pepicrft, line 17: Since we don't run this in development, the feature pages failed to render trying to get state from a non-running GenServer.

PR #9939, opened 2026-03-19
feat(server): buffer ClickHouse inserts for test and cache tables

fortmarek:
## Summary

- Adds write buffering for 8 high-volume ClickHouse tables (`test_case_runs`, `test_module_runs`, `test_suite_runs`, `test_cases`, `test_case_failures`, `test_case_run_repetitions`, `test_case_events`, `cas_events`) that were generating ~75K concurrent INSERT queries
- Each table gets a dedicated `Buffer` GenServer following the existing `Build.Buffer` pattern — rows accumulate in memory and flush periodically (every 5s or when buffer reaches max size)
- Fixes a double-resolution bug in `Bufferable.compile_time_prepare` where `Ecto.UUID` types were passed through `Ecto.Type.type()` before `remap_type/3` (which calls it again internally)
- Adds `sync_writes` config flag so buffers flush synchronously in test environment

## Test plan
- [x] `mix test test/tuist/tests_test.exs` — 164 tests, 0 failures
- [x] `mix test test/tuist/cache_test.exs test/tuist/tests/` — 183 tests, 0 failures
- [x] `mix test test/tuist/tests/workers/ test/tuist_web/live/flaky_tests_live_test.exs test/tuist_web/live/quarantined_tests_live_test.exs test/tuist_web/controllers/webhooks/cache_controller_test.exs` — 18 tests, 0 failures
- [x] `mix credo --strict` — no issues on changed files
- [ ] Monitor ClickHouse query count after deploy to staging

### DIFF
--- a/server/lib/tuist/ingestion/bufferable.ex
+++ b/server/lib/tuist/ingestion/bufferable.ex
@@ -39,69 +39,7 @@ defmodule Tuist.Ingestion.Bufferable do
 
       defmodule Buffer do
         @moduledoc false
-
-        alias Tuist.Ingestion.Buffer
-
-        @parent unquote(parent)
-
-        defp opts do
-          case :persistent_term.get({__MODULE__, :opts}, nil) do
-            nil ->
-              computed = Tuist.Ingestion.Bufferable.compile_time_prepare(@parent)
-              :persistent_term.put({__MODULE__, :opts}, computed)
-              computed
-
-            cached ->
-              cached
-          end
-        end
-
-        def child_spec(child_opts) do
-          %{header: header, insert_sql: insert_sql, insert_opts: insert_opts} = opts()
-
-          child_opts =
-            Keyword.merge(child_opts,
-              name: __MODULE__,
-              header: header,
-              insert_sql: insert_sql,
-              insert_opts: insert_opts
-            )
-
-          Buffer.child_spec(child_opts)
-        end
-
-        def insert(row) do
-          %{fields: fields, encoding_types: encoding_types} = opts()
-
-          row_binary =
-            [Enum.map(fields, fn field -> Map.fetch!(row, field) end)]
-            |> Ch.RowBinary._encode_rows(encoding_types)
-            |> IO.iodata_to_binary()
-
-          :ok = Buffer.insert(__MODULE__, row_binary)
-          {:ok, row}
-        end
-
-        def insert_all([]), do: {0, nil}
-
-        def insert_all(rows) do
-          %{fields: fields, encoding_types: encoding_types} = opts()
-
-          row_binary =
-            rows
-            |> Enum.map(fn row ->
-              Enum.map(fields, fn field -> Map.fetch!(row, field) end)
-            end)
-            |> Ch.RowBinary._encode_rows(encoding_types)
-            |> IO.iodata_to_binary()
-
-          :ok = Buffer.insert(__MODULE__, row_binary)
-          {length(rows), nil}
-        end
-
-        def flush do
-          Buffer.flush(__MODULE__)
-        end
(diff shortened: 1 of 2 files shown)


server/lib/tuist/ingestion/buffer.ex
```diff
@@ -110,4 +126,8 @@ defmodule Tuist.Ingestion.Buffer do
   defp default_max_buffer_size do
     Keyword.fetch!(Application.get_env(:tuist, IngestRepo), :max_buffer_size)
   end
+
+  defp sync_writes? do
```
pepicrft, line 130: Can we expect `Application.get_env(:tuist, IngestRepo)` to always return a keyword list? If not, I'd then adjust the logic to account for that.

server/lib/tuist/gradle/buffer/task.ex
```diff
@@ -1,39 +1,4 @@
...
-      |> IO.iodata_to_binary()
-
-    :ok = Buffer.insert(__MODULE__, row_binary)
-    {:ok, task}
-  end
-
-  def flush do
-    Buffer.flush(__MODULE__)
-  end
+  use Tuist.Ingestion.SchemaBuffer, schema: Tuist.Gradle.Task
```
pepicrft: What if instead of having `_buffer.ex` modules, you add this to the schema:

```exs
defmodule Tuist.Gradle.Task
  use Tuist.Ingestion.Bufferable
  # It defines Tuist.Gradle.Task.Buffer
end
```

Also, I noticed the file location, `/gradle/buffer/task` doesn't align with the symbol name `Tuist.Gradle.Task.Buffer`. I'd revisit all the schemas.

server/lib/tuist/tests.ex
```diff
@@ -606,10 +606,7 @@ defmodule Tuist.Tests do
         inserted_at: NaiveDateTime.utc_now()
       }
 
-      {:ok, _module_run} =
-        %TestModuleRun{}
-        |> TestModuleRun.create_changeset(module_run_attrs)
```
pepicrft, line 611: Are we fine with skipping the changeset validations here?

server/lib/tuist/ingestion/buffer.ex
```diff
@@ -68,6 +72,18 @@ defmodule Tuist.Ingestion.Buffer do
     {:noreply, %{state | buffer: [], buffer_size: 0, timer: timer}}
   end
 
+  @impl true
+  def handle_call({:insert_and_flush, row_binary}, _from, state) do
```
pepicrft, line 76: The other flush paths cancel the timer, but this one does not. Is that intentional?

fortmarek: there's no timer to cancel if the inserts are synchronous

server/lib/tuist/ingestion/buffer.ex
fortmarek: Made `sync_writes?` defensive — now handles non-keyword-list config gracefully.

server/lib/tuist/gradle/buffer/task.ex
fortmarek: `use Bufferable` now auto-generates the `Buffer` submodule inside the schema via `__before_compile__`. All 15 separate buffer files and the `SchemaBuffer` macro are deleted. The Buffer opts are computed lazily at first use via `persistent_term` (needed because schema functions aren't callable during `__before_compile__`).

server/lib/tuist/tests.ex
fortmarek: Good catch — restored changeset validation. Now validates via `apply_action!/2` before buffering, which raises on invalid data (matching the original crash behavior).

PR #9946, opened 2026-03-20
feat(cli): add --cache-profile option to cache warm with profile-driven exclusions

### LINKED ISSUE #9945: Cache warm should respect CacheProfile.exceptTargetQueries
## Problem

When using local Swift packages via `Tuist/Package.swift` (`.package(path:)`), Tuist treats them as external dependencies. Running `tuist cache --external-only` attempts to build and cache these local packages as XCFrameworks, which is unnecessary and often fails with:

```
error: at least one framework or library must be specified
```

`CacheProfile` already supports `exceptTargetQueries` for excluding targets during **generation** (via `TargetReplacementDecider`), but cache **warm** ignores these exclusions entirely. Users cannot leverage their existing profile configuration to skip unnecessary targets during caching.

## Reproducible Sample Project

https://github.com/gnejfejf2/tuist-cache-profile-sample

```
$ tuist cache --external-only
Targets to be cached: Alamofire, Alamofire_Alamofire, CommonKit
                                                      ^^^^^^^^^ should be excluded
```

## Expected Behavior

Cache warm should read the default cache profile from config and apply its `exceptTargetQueries` to exclude targets, just as generation already does via `TargetReplacementDecider`.

For example, given:
```swift
// Tuist.swift
let tuist = Tuist(
    project: .tuist(
        cacheOptions: .options(
            profiles: .profiles(
                [
                    "development": .profile(
                        .onlyExternal,
                        except: ["CommonKit"]
                    ),
                ],
                default: "development"
            )
        )
    )
)
```

Running `tuist cache --external-only` should skip `CommonKit` during cache warm.

## Current Behavior

`CacheWarmCommandService.cacheableTargets()` does not reference the cache profile's `exceptTargetQueries` at all. The only exclusion mechanism is the `externalOnly` boolean, which excludes all internal targets but has no way to exclude specific external targets.

## Why This Matters

Many teams use local SPM packages for shared infrastructure -- design systems, networking layers, common utilities. These are actively developed source code that changes frequently. Developers need source-level debugging and code navigation, not binary XCFrameworks.

Without `except` being respected during cache warm, teams endure unnecessary compilation of these local packages every time `tuist cache` runs.

## Related

- PR #9946

gnejfejf2:
## Summary

Adds a `--cache-profile` option to `tuist cache` that allows users to specify which cache profile to use during cache warming. When specified, the profile's configuration takes priority:

- **Profile's `base`** determines external-only behavior (overrides `--external-only` flag)
- **Profile's `exceptTargetQueries`** excludes specified targets from caching
- Both `.named()` and `.tagged()` queries are supported

When `--cache-profile` is not specified, existing behavior is preserved.

### Problem

When using local Swift packages via `Tuist/Package.swift` (`.package(path:)`), Tuist treats them as external dependencies. Running `tuist cache --external-only` caches these local packages as XCFrameworks unnecessarily.

`CacheProfile` supports `exceptTargetQueries` for excluding targets during **generation** (via `TargetReplacementDecider`), but cache **warm** has no way to leverage profile configuration.

Reproducible sample project: https://github.com/gnejfejf2/tuist-cache-profile-sample

```
$ tuist cache --external-only
Targets to be cached: Alamofire, Alamofire_Alamofire, CommonKit
                                                      ^^^^^^^^^ should be excluded
```

### Changes

- `CacheWarmCommand.swift`: Added `--cache-profile` CLI option
- `CacheServicing` protocol: Added `cacheProfile: String?` parameter
- `CacheWarmCommandService.swift`:
  - Added `ProfileResolution` struct and `resolveProfile()` method
  - When `--cache-profile` is specified, profile's `base` overrides `--external-only`:
    - `.onlyExternal` → external-only mode
    - `.allPossible` → cache all targets
    - `.none` → no caching
  - Profile's `exceptTargetQueries` are resolved against the graph and excluded from caching

### Usage

```bash
# Profile takes priority -- profile's base determines behavior,
# exceptTargetQueries excludes specified targets
tuist cache --cache-profile development

# Can combine with --external-only, but profile overrides it when specified
tuist cache --external-only --cache-profile development

# Without --cache-profile, existing behavior is preserved
tuist cache --external-only
```

### Design decisions

- **Explicit opt-in via CLI flag** rather than automatic profile resolution, respecting that cache profiles were originally a consumer-side control
- **Profile base overrides --external-only** because the profile already encodes this information, avoiding conflicting signals
- **No breaking changes to default behavior** -- existing `tuist cache` commands work exactly as before

## Test plan

- [x] Verified with [sample project](https://github.com/gnejfejf2/tuist-cache-profile-sample) that CommonKit appears in cache targets without profile (the bug)
- [x] Verify with EE build that `--cache-profile development` excludes CommonKit

Resolves #9945

### DIFF
--- a/cli/Sources/TuistCacheCommand/CacheWarmCommand.swift
+++ b/cli/Sources/TuistCacheCommand/CacheWarmCommand.swift
@@ -44,6 +44,12 @@
         var targets: [String] = []
 
         @Flag(
+            help: "If passed, the command doesn't cache the targets passed in the `--targets` argument, but only their dependencies. Deprecated: use `--cache-profile only-external` instead.",
+            envKey: .cacheExternalOnly
+        )
+        var externalOnly: Bool = false
+
+        @Flag(
             name: .long,
             help: "When passed, it generates the project and skips warming the cache. This is useful for debugging purposes.",
             envKey: .cacheGenerateOnly
@@ -74,10 +80,18 @@
                 return
             }
 
+            if externalOnly {
+                AlertController.current.warning(.alert(
+                    "The \(.command("--external-only")) flag is deprecated.",
+                    takeaway: "Use \(.command("--cache-profile only-external")) instead."
+                ))
+            }
+
             try await Extension.cacheService.run(
                 path: path,
                 configuration: configuration,
                 targetsToBinaryCache: Set(targets),
+                externalOnly: externalOnly,
                 generateOnly: generateOnly,
                 cacheProfile: cacheProfile
             )
--- a/cli/Sources/TuistEnvKey/EnvKey.swift
+++ b/cli/Sources/TuistEnvKey/EnvKey.swift
@@ -439,6 +439,7 @@ public enum EnvKey: String, CaseIterable {
 
     // CACHE
 
+    case cacheExternalOnly = "TUIST_CACHE_EXTERNAL_ONLY"
     case cacheProfile = "TUIST_CACHE_PROFILE"
     case cacheGenerateOnly = "TUIST_CACHE_GENERATE_ONLY"
     case cachePrintHashes = "TUIST_CACHE_PRINT_HASHES"
--- a/cli/Sources/TuistExtension/Extension.swift
+++ b/cli/Sources/TuistExtension/Extension.swift
@@ -12,6 +12,7 @@ public protocol CacheServicing {
         path: String?,
         configuration: String?,
         targetsToBinaryCache: Set<String>,
+        externalOnly: Bool,
         generateOnly: Bool,
         cacheProfile: String?
     ) async throws
@@ -31,6 +32,7 @@ public struct EmptyCacheService: CacheServicing {
         path _: String?,
         configuration _: String?,
         targetsToBinaryCache _: Set<String>,
+        externalOnly _: Bool,
         generateOnly _: Bool,
         cacheProfile _: String?
     ) async throws {
--- a/cli/Sources/TuistKit/Commands/Cache/CacheWarmCommandService.swift
+++ b/cli/Sources/TuistKit/Commands/Cache/CacheWarmCommandService.swift
@@ -90,6 +90,7 @@ import XcodeGraph
             path directory: String?,
             configuration: String?,
             targetsToBinaryCache: Set<String>,
+            externalOnly: Bool,
             generateOnly: Bool,
             cacheProfile: String?
         ) async throws {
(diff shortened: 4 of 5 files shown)


pepicrft:
Thanks, @Garfeild, for the contribution. I understand the motivation behind this, but I don't think the execution is the best one. Cache profiles are a tool to _select_ the targets that should be binaries on generation. You are reusing a concept, taking a subset of its parameters to solve your particular need, which is not the right way to approach the problem.
If we plan to reuse the concept of cache profiles, then we should not change the meaning. Since the current API is flexible enough, I'd say, let's apply the same filtering as we do in generation, but then apply a `tuist:` tag that you can use to tweak your profile:

```swift
CacheProfile(
      base: .onlyExternal,
      targetQueries: ["..."],       // ← also include this local target
      exceptTargetQueries: ["tuist:package:local"]   // ← but exclude this
  )
```

Before you make any changes, let's also wait for @fortmarek's input.

fortmarek:
Agree with @pepicrft. The CLI `tuist cache` and `tuist generate` integration of profiles should be as aligned as possible. While at it, we should probably consider deprecating `--external-only` flag in favor of using a predefined profile, similar to how we deprecated `tuist generate --no-binary-cache`

gnejfejf2:
@fortmarek Thank you so much for the detailed feedback! I completely agree with aligning the cache warm behavior with generate's profile filtering. I'll rework the implementation to:

1. Apply the same profile-based filtering logic that generation uses
2. Deprecate and remove the `--external-only` flag in favor of using cache profiles exclusively

Will push the updated changes soon!

gnejfejf2:
Thank you so much for the feedback\! I've reworked the implementation to align cache warm with generate's profile filtering pipeline. The `--external-only` flag has been removed entirely in favor of using `--cache-profile` exclusively, which now uses the same `resolveCacheProfile()` → `CacheProfileTargetReplacementDecider` flow as generation. Would love your review on the updated approach\!

fortmarek:
## Changes made

Pushed [27c9aba](https://github.com/tuist/tuist/pull/9946/commits/27c9aba3d6) to address review feedback:

1. **Restored `--external-only` as deprecated** — the flag still works but shows a warning pointing users to `--cache-profile only-external`. The `TUIST_CACHE_EXTERNAL_ONLY` env var is also preserved.

2. **Fixed default behavior** — when neither `--cache-profile` nor `--external-only` is specified, defaults to `.allPossible` (cache everything), preserving the original `tuist cache` behavior. Previously the PR used `resolveCacheProfile` which falls through to `.onlyExternal`, silently changing the default.

3. **Added clarifying comments** explaining why `includedTargets: []` is passed to the profile resolver (because `--targets` filtering is applied separately).

4. **Restored `externalOnly` in `CacheServicing` protocol** to avoid breaking out-of-tree conformances.

## E2E testing

Built the `tuist` executable with EE (`TUIST_EE=1 tuist generate && xcodebuild build -scheme tuist`) and tested against the `examples/xcode/generated_ios_app_with_cache_profiles` fixture, which has:
- External dep: Alamofire
- `ExpensiveModule` (named in profile)
- `TaggedModule` (tagged `cacheable`, matched by `tag:cacheable`)
- `NonCacheableModule` (not in profile)
- `development` profile: `.onlyExternal` base + `ExpensiveModule` + `tag:cacheable`

### Results

| Command | Targets cached | Expected? |
|---------|---------------|-----------|
| `tuist cache --generate-only` | Alamofire, ExpensiveModule, NonCacheableModule, TaggedModule | Yes -- caches everything (old default preserved) |
| `tuist cache --external-only --generate-only` | Alamofire | Yes -- only external + deprecation warning shown |
| `tuist cache --cache-profile development --generate-only` | Alamofire, ExpensiveModule, TaggedModule | Yes -- NonCacheableModule correctly excluded by profile |
| `tuist cache --cache-profile only-external --generate-only` | Alamofire | Yes -- equivalent to deprecated `--external-only` |

PR #9952, opened 2026-03-20
feat(server): update organization creation flow

asmitbm:
Resolves TUI-535

- Update the onboarding flow to require organization creation before project creation: sign up → confirm → choose username → create organization → create project → dashboard

Screenshot 2026-03-20 at 12 31 32
Screenshot 2026-03-20 at 12 31 53

- Update CreateOrganizationLive with new copy, hint text for name field, and conditional Cancel button (hidden for new users, visible for existing users)

- Update CreateProjectLive to only show organization accounts in the dropdown, remove "Create organization" button, and reorder fields (select org → name → build system)

- Redirect post-confirmation and OAuth signup to `/organizations/new` for users without organizations

- Enable password visibility toggle (eye icon) on registration, login, and reset password pages by setting `type="password"` on text inputs

- Remove bottom border from Previews card in marketing page

### DIFF
--- a/server/priv/gettext/dashboard.pot
+++ b/server/priv/gettext/dashboard.pot
@@ -258,7 +258,6 @@ msgstr ""
 msgid "Server error"
 msgstr ""
 
-#: lib/tuist_web/components/app_layout_components.ex:353
 #: lib/tuist_web/components/layouts/account.html.heex:69
 #, elixir-autogen, elixir-format
 msgid "Settings"
@@ -503,7 +502,7 @@ msgstr ""
 msgid "Metrics"
 msgstr ""
 
-#: lib/tuist_web/controllers/auth_controller.ex:295
+#: lib/tuist_web/controllers/auth_controller.ex:293
 #, elixir-autogen, elixir-format
 msgid "Failed to authenticate."
 msgstr ""
@@ -523,7 +522,7 @@ msgstr ""
 msgid "Authentication failed"
 msgstr ""
 
-#: lib/tuist_web/controllers/auth_controller.ex:286
+#: lib/tuist_web/controllers/auth_controller.ex:284
 #, elixir-autogen, elixir-format
 msgid "Your identity provider denied access. Please ask your organization admin to assign you to the Tuist application."
 msgstr ""
--- a/server/priv/gettext/dashboard_auth.pot
+++ b/server/priv/gettext/dashboard_auth.pot
@@ -16,17 +16,17 @@ msgstr ""
 msgid "Account confirmed!"
 msgstr ""
 
-#: lib/tuist_web/live/user_registration_live.ex:294
+#: lib/tuist_web/live/user_registration_live.ex:295
 #, elixir-autogen, elixir-format
 msgid "Account created successfully! Please log in."
 msgstr ""
 
-#: lib/tuist_web/live/user_registration_live.ex:306
+#: lib/tuist_web/live/user_registration_live.ex:307
 #, elixir-autogen, elixir-format
 msgid "Account name is already taken"
 msgstr ""
 
-#: lib/tuist_web/live/user_registration_live.ex:187
+#: lib/tuist_web/live/user_registration_live.ex:188
 #, elixir-autogen, elixir-format
 msgid "Already have an account?"
 msgstr ""
@@ -56,7 +56,7 @@ msgstr ""
 msgid "Check your email"
 msgstr ""
 
-#: lib/tuist_web/live/user_registration_live.ex:220
+#: lib/tuist_web/live/user_registration_live.ex:221
 #, elixir-autogen, elixir-format
 msgid "Check your inbox for a confirmation email and click the link to continue."
 msgstr ""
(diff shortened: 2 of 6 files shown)


server/mise.lock
fortmarek, line 1: let's not update `mise.lock`

server/lib/tuist_web/live/create_project_live.ex
```diff
@@ -19,27 +19,28 @@ defmodule TuistWeb.CreateProjectLive do
...
     selected_account =
       case Map.get(params, "account_id") do
         account_id when is_binary(account_id) ->
           account_id = String.to_integer(account_id)
 
-          if Enum.any?(all_accounts, &(&1.id == account_id)),
+          if Enum.any?(organization_accounts, &(&1.id == account_id)),
             do: account_id,
-            else: current_user.account.id
+            else: List.first(organization_accounts) && List.first(organization_accounts).id
```
fortmarek: should we try to use the latest accessed project (and its organization) instead of the first org account from the list? And if there's no accessed project, _then_ we can use the first org account from the list?

server/lib/tuist_web/live/create_project_live.ex
```diff
@@ -19,27 +19,28 @@ defmodule TuistWeb.CreateProjectLive do
...
 
         _ ->
-          current_user.account.id
+          case organization_accounts do
+            [first | _] -> first.id
+            [] -> nil
```
fortmarek: and we're getting an id _again_ although we try to get [here](https://github.com/tuist/tuist/pull/9952/changes#diff-34bacc228d9d905604b206cb1bea95eefe5d9b2bf54b6ca24201f1ea9f971de4R29) as well. Let's create a helper for a preselected organization account

fortmarek: what happens when we return nil from this method?

server/lib/tuist_web/live/create_project_live.ex
```diff
@@ -91,41 +112,11 @@ defmodule TuistWeb.CreateProjectLive do
...
                 variant="primary"
                 label={dgettext("dashboard_projects", "Continue")}
                 type="submit"
               />
-              <.line_divider text={dgettext("dashboard_projects", "Or set up a new organization")} />
-              <.button
-                label={dgettext("dashboard_projects", "Create organization")}
-                variant="secondary"
-                navigate={~p"/organizations/new"}
-              />
```
fortmarek, line 128: why are removing this? shouldn't it still be possible to create an org from the create project page?

mise.lock
fortmarek, line 1: let's revert these changes as well

server/lib/tuist_web/live/create_organization_live.ex
```diff
@@ -33,33 +44,41 @@ defmodule TuistWeb.CreateOrganizationLive do
...
                 id="organization-name"
                 field={@form[:name]}
                 type="basic"
                 label={dgettext("dashboard_account", "Name")}
-                show_required
-                required
+                hint={
+                  dgettext(
+                    "dashboard_account",
+                    "What's the name of your company or team? You can change this later in the settings"
```
fortmarek, line 69: this seems to be taken from supabase, but looks a bit odd in our UI imho. What do you think about skipping this?

Image

Image

server/lib/tuist_web/live/create_organization_live.ex
```diff
@@ -33,33 +44,41 @@ defmodule TuistWeb.CreateOrganizationLive do
...
                 <.button
                   type="submit"
                   variant="primary"
-                  label={dgettext("dashboard_account", "Create organization")}
+                  label={dgettext("dashboard_account", "Continue")}
                 />
                 <.button
-                  navigate={~p"/projects/new"}
+                  :if={@has_organizations}
+                  navigate={~p"/#{@current_user.account.name}/projects"}
```
fortmarek: but this leads to user account, which is what we're trying to get rid of. We should imho redirect to either the last accessed organization or the last accessed project.

fortmarek:
In general, actually, I wonder if we should have a different design for when creating a project/org as part of sign up or outside of it. For example, ` By continuing, you agree to Tuist's Terms of Service and Privacy Policy ` also doesn't make a ton of sense if I already have an account, no?
image

asmitbm:
> Left a couple of mostly minor comments 😌
> 
> A bit unrelated, but what do you think about retaining the headerbar during the creation flow, like supabase does?

I’ll be redesigning these sections as part of a new header and design pattern, so I’ll update these pages once that’s finalized.

server/lib/tuist_web/live/create_project_live.ex
asmitbm: From a UX perspective, this flow may introduce unnecessary cognitive load. We’re first asking users to create an organization and then a project, so including an option to create an organization within the “Create Project” modal feels redundant and could lead to confusion.

server/lib/tuist_web/live/create_organization_live.ex
asmitbm: Do you mean the hint text below the Name field?

server/lib/tuist_web/live/create_organization_live.ex
asmitbm: Got it!

asmitbm:
> In general, actually, I wonder if we should have a different design for when creating a project/org as part of sign up or outside of it. For example, `By continuing, you agree to Tuist's Terms of Service and Privacy Policy` also doesn't make a ton of sense if I already have an account, no? 

Yupp I will see if I can hide this when the user is authenticated

server/lib/tuist_web/live/create_project_live.ex
asmitbm: When selected_account is nil (no organization accounts), the user would see an empty select dropdown and submitting would fail. I've added a redirect to /organizations/new when the user has no organizations, so they create one first before reaching project creation.

PR #9953, opened 2026-03-20
fix(cli): fix static framework resource bundle crash when using xcstrings

pepicrft:
## Summary

- Fixes a runtime crash (`fatalError: unable to find bundle named X_X`) when a static framework uses `.xcstrings` resources
- The regression was introduced in #9907 which retained xcstrings in the static framework's `resources` to prevent stale string extraction. This caused `Target.containsResources` to return `true`, making `GraphTraverser.resourceBundleDependencies()` skip the framework during traversal — so the companion resource bundle was never propagated to the app target and never embedded at runtime.
- The fix removes the xcstrings retention from the main target's resources. The xcstrings files are still compiled in the companion bundle target's Resources phase, and are still excluded from the Sources phase (preserving the stale extraction fix from #9907).

## Test plan

- [ ] Existing acceptance test `ios_app_with_static_framework_with_xcstrings` validates both: no stale extraction AND the bundle is embedded with compiled `Localizable.strings`
- [ ] Unit test updated to verify xcstrings is not retained in the static framework's resources
- [ ] Verify the sample project from the bug report no longer crashes at runtime

### DIFF
--- a/cli/Tests/TuistGeneratorAcceptanceTests/GenerateAcceptanceTests.swift
+++ b/cli/Tests/TuistGeneratorAcceptanceTests/GenerateAcceptanceTests.swift
@@ -98,15 +98,15 @@ struct GenerateAcceptanceTestiOSAppWithTests {
     @Test(.withFixture("generated_ios_app_with_tests"), .inTemporaryDirectory)
     func focused_targets() async throws {
         let fixturePath = try fixtureDirectory()
-        let xcodeprojPath = try await TuistAcceptanceTest.xcodeprojPath(in: fixturePath)
 
-        func generatedTargets() throws -> [String] {
-            try XcodeProj(pathString: xcodeprojPath.pathString).pbxproj.nativeTargets.map(\.name).sorted()
+        func generatedTargets() async throws -> [String] {
+            let xcodeprojPath = try await TuistAcceptanceTest.xcodeprojPath(in: fixturePath)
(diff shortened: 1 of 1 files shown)


cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
```diff
@@ -91,20 +91,15 @@ public struct ResourcesProjectMapper: ProjectMapping { // swiftlint:disable:this
...
+            // a companion resource bundle (bundle: .module). The xcstrings files are compiled
+            // in the companion bundle target's Resources phase instead.
             let codeGeneratingResourceExtensions: Set<String> = ["xcassets"]
             for resource in target.resources.resources {
                 if let ext = resource.path.extension, codeGeneratingResourceExtensions.contains(ext) {
                     modifiedTarget.sources.append(SourceFile(path: resource.path))
                 }
             }
-            // Keep xcstrings in the main target's resources so Xcode's string catalog editor
-            // can match string references in the target's Swift sources. Other resources are
```
fortmarek, line 104: will this again break the catalog editor?

pepicrft: I added an acceptance test covering the scenario, so I don't expect that to be the case.

Issue #7164, opened 2024-12-15
Build failed when using swift macros as external dependencies


mo5tone:
### What happened?

My app and modules have some swift macros as external dependencies. 
Everything works well when run through Xcode.
But it failed when run by `xcodebuild` no matter on CI nodes or my machine.

### How do we reproduce it?

1. ```
    cd fixtures/multiplatform_app_with_macros_and_embedded_watchos_app \
    && tuist install \
    && tuist generate
    ```
2. `⌘ + U` in Xcode and it will build and test successfully
3. Open terminal and run `xcodebuild -workspace AppWithWatchApp.xcworkspace -scheme AppWithWatchApp-Workspace -configuration Debug -scmProvider system -disableAutomaticPackageResolution -sdk 'iphonesimulator' -destination 'platform=iOS Simulator,id=theUUIDofYourSimulator' -enableCodeCoverage YES -skipPackagePluginValidation -skipMacroValidation -retry-tests-on-failure -test-iterations 11 clean test`
4. The tests build failed with below errors:
`Multiple commands produce '/yourDerivedData/AppWithWatchApp-ddfnoigpczzvcocxhnuttxpcsxlc/Build/Products/Debug-iphonesimulator/CasePathsMacros'`

    `Multiple commands produce '/yourDerivedData/AppWithWatchApp-ddfnoigpczzvcocxhnuttxpcsxlc/Build/Products/Debug-iphonesimulator/ModuleAMacros'`
5. Same errors when changed to other swift macros such [MetaCodable](https://github.com/SwiftyLab/MetaCodable) and [swift-composable-architecture](https://github.com/pointfreeco/swift-composable-architecture)

### Error log

`Multiple commands produce '/yourDerivedData/AppWithWatchApp-ddfnoigpczzvcocxhnuttxpcsxlc/Build/Products/Debug-iphonesimulator/CasePathsMacros'`

`Multiple commands produce '/yourDerivedData/AppWithWatchApp-ddfnoigpczzvcocxhnuttxpcsxlc/Build/Products/Debug-iphonesimulator/ModuleAMacros'`

### macOS version

14.6.1

### Tuist version

4.37.0

### Xcode version

15.4

mo5tone:
Same errors when try with Xcode 16.2 on macOS 14.6.1.

mo5tone:
[...]

Issue #5987, opened 2024-02-24
Macros cannot be expanded when using cache


JaroVoltix:
### What happened?

When I use tuist cache macros work, however they cannot be expanded

### How do we reproduce it?

1. Download attached project [framework_with_native_swift_macro.zip](https://github.com/tuist/tuist/files/14392646/framework_with_native_swift_macro.zip)

2. tuist install
3. tuist generate 
4. open Framework.swift 
5. Right click on @Reducer -> Expand macro option is visible
6. tuist cache
7. tuist generate
8. Right click on @Reducer

Expected: Expand macro option is visible

Actual: Expand macro option is not visible

### Error log

-

### macOS version

14.1

### Tuist version

4.3.4

### Xcode version

15.2

pepicrft:
Thanks for reporting this @JaroVoltix. We added it to the list of issues to look into.

mihaicris-adoreme:
Hi. I ran into this issue as well and I wonder if it will be addressed soon. I started using Mockable, https://github.com/Kolos65/Mockable.git, and if I cache this dependency I cannot expand the macros. 

I also found a defect which is more serious, not sure if it is Tuist or Mockable, but when I use Mockable in a SPM package to create mocks, and try to import the Mock  generated classes from the SPM module in another target (main app for example)  the classes are not found, even if they are generated with public access control. I will try to make a project sample and create another issue.

pepicrft:
I haven't been able to looked into it yet. My hunch is that it's a bug in Xcode's editor when the Swift Macro is provided as a compiled executable (to confirm). If that's the case, I'll have to file a radar and provide an example project for Apple to look into the issue.

pepicrft:
@mihaicris-adoreme could it be that you are missing the MOCKING compiler directive that that macro requires? We are using it in Tuist, and we haven't encountered any issue.

lu15gv:
We have the same problem and had to pause macro integration as we prefer to use cache.
@pepicrft were you able to see if it was a problem with the xcode editor?

hiltonc:
This is still relevant with Tuist 4.78.0.

PR #9962, opened 2026-03-20
fix(cli): always copy macro executable on incremental builds

### LINKED ISSUE #9958: Incremental builds silently reuse stale macro expansions for first-party macros (two-layer .macro + .framework pattern)
## Summary

First-party Swift macro targets in Tuist-generated Xcode workspaces do not properly invalidate macro expansion caches during incremental builds when *only the plugin implementation changes*. The test suite passes with stale output. `xcodebuild clean` works around it; `xcodebuild test` (incremental) silently reuses the old expansion.

**Minimal reproducer:** https://github.com/ffittschen/tuist-macro-staleness

## Environment

- Tuist: 4.162.0
- Xcode: 26.3.0
- Swift: 6.0
- macOS: 15 (Apple Silicon)

## Root Cause (hypothesis)

Tuist represents a first-party macro using the two-layer pattern:

1. **`product: .macro` (macOS)** — `InjectVersionImpl` — the compiler plugin binary
2. **`product: .framework` (iOS)** — `InjectVersion` — a thin public API wrapper that re-exports the macro declaration

Dynamic frameworks emit a `.swiftinterface` file. Xcode uses `.swiftinterface` as the dependency-tracking boundary for incremental builds. When `InjectVersion`'s *public interface* doesn't change (the macro declaration is identical), Xcode skips recompiling `Consumer` and reuses its stale macro-expanded output — even though `InjectVersionImpl`'s plugin binary changed.

```
InjectVersionImpl (plugin binary changes) ─┐
                                            │  .swiftinterface does NOT change
InjectVersion (.swiftinterface) ────────────┤
                                            │  Xcode skips Consumer recompile
Consumer (macro expansion site) ────────────┘  → stale expansion reused silently
```

The stale incremental build completes in ~0.001 s — Xcode never recompiles `Consumer` at all.

## Reproduction Steps

See [README](https://github.com/ffittschen/tuist-macro-staleness#reproduction-steps) for exact commands. The short version:

```sh
# 1. Setup
git clone https://github.com/ffittschen/tuist-macro-staleness
cd tuist-macro-staleness
mise trust && mise install
tuist install && tuist generate --no-open

# 2. Baseline clean build — PASSES (version_1) ✅
xcodebuild clean test -workspace MacroStaleness.xcworkspace \
  -scheme MacroStaleness-Workspace \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro'

# 3. Mutate only the plugin implementation (no other files touched)
sed -i '' '/"version_1"/s/version_1/version_2/' \
  Sources/MacroImpl/InjectVersionMacro.swift

# 4. Incremental build — should FAIL, silently PASSES ❌ (the bug)
xcodebuild test -workspace MacroStaleness.xcworkspace \
  -scheme MacroStaleness-Workspace \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro'

# 5. Clean build — FAILS correctly ✅
xcodebuild clean test -workspace MacroStaleness.xcworkspace \
  -scheme MacroStaleness-Workspace \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro'
# Expectation failed: (Widget.generatedValue → "version_2") == "version_1"
```

## Impact

When a developer changes a macro's generated output — adding a new synthesised member, fixing a generated expression, changing generated code structure — the change is silently invisible to all consuming targets in the next incremental build. The consuming targets continue compiling against the old expansion without error or warning. The developer only sees the effect of their change after a clean build, which may not happen for a long time in a typical iterative workflow. This makes macro implementations particularly hazardous to iterate on: a change that appears to have no effect may in fact be silently discarded by the build system.

## Question

Is this fixable on the Tuist side, or is it a fundamental limitation of Xcode's incremental build system?

A few angles I'm wondering about:

1. **`product: .staticFramework` instead of `.framework`** — Static frameworks don't emit `.swiftinterface`, so Xcode can't use the interface boundary to skip recompilation. Would Tuist switching the interface wrapper to a static framework be a viable fix, and are there downsides (e.g. duplicate symbol risks)?
2. **Explicit dependency on the plugin binary** — Could the `InjectVersion` target declare an explicit file-level dependency on the `InjectVersionImpl` binary, so any change to the plugin binary forces a recompile of the interface wrapper (and by extension, all consumers)?
3. **Known Xcode limitation** — Is this a known gap in Xcode's incremental build system for macro plugin binaries where no Tuist-side mitigation exists?

Related issues: #5987, #7164

ffittschen:
Resolves #9958

## Summary

After the [response from dosubot](https://github.com/tuist/tuist/issues/9958#issuecomment-4099557778) I did a little more digging, since I remembered that there was some build phase related to Macros generated by Tuist. It turned out that the "Copy Swift Macro executable into $BUILT_PRODUCT_DIR" build phase has a `! -f` guard that skips copying when the destination file already exists. On incremental builds this means the macro binary is never updated after the initial clean build, causing Xcode to silently use stale macro expansions.

**Before (broken):**
```bash
if [[ -f "$source" && ! -f "$destination" ]]; then
    cp "$source" "$destination"
fi
```

**After (fixed):**
```bash
if [[ -f "$source" ]]; then
    cp -f "$source" "$destination"
fi
```

The script phase already declares `inputPaths` (the macOS-built macro binary) and `outputPaths` (the platform-specific copy). With the guard removed, `cp -f` updates the destination's mtime on each copy, so Xcode can correctly skip the script when the macro binary hasn't been rebuilt. The old `! -f` guard prevented the destination from ever being updated, which meant Xcode re-ran the script on every incremental build anyway — it just did nothing.

I verified the fix by building Tuist from source, regenerating the reproducer, and confirming that incremental builds now correctly detect macro implementation changes.

Related: #5987, #7164

### How to test locally

- Verifiable with [reproducer project](https://github.com/ffittschen/tuist-macro-staleness):
  - `xcodebuild clean test` → passes (baseline)
  - Mutate macro `"version_1"` → `"version_2"`
  - `xcodebuild test` (incremental) → now correctly **fails** (was silently passing before)
- Existing Tuist test suite passes

### DIFF
--- a/cli/Tests/TuistGeneratorTests/Generator/BuildPhaseGeneratorTests.swift
+++ b/cli/Tests/TuistGeneratorTests/Generator/BuildPhaseGeneratorTests.swift
@@ -2021,7 +2021,7 @@ struct BuildPhaseGeneratorTests {
         #expect(buildPhase != nil)
 
         let expectedScript =
-            "if [[ -f \"$BUILD_DIR/$CONFIGURATION/macro\" && ! -f \"$BUILD_DIR/Debug$EFFECTIVE_PLATFORM_NAME/macro\" ]]; then\n    mkdir -p \"$BUILD_DIR/Debug$EFFECTIVE_PLATFORM_NAME/\"\n    cp \"$BUILD_DIR/$CONFIGURATION/macro\" \"$BUILD_DIR/Debug$EFFECTIVE_PLATFORM_NAME/macro\"\nfi"
+            "if [[ -f \"$BUILD_DIR/$CONFIGURATION/macro\" ]]; then\n    mkdir -p \"$BUILD_DIR/Debug$EFFECTIVE_PLATFORM_NAME/\"\n    cp -f \"$BUILD_DIR/$CONFIGURATION/macro\" \"$BUILD_DIR/Debug$EFFECTIVE_PLATFORM_NAME/macro\"\nfi"
         #expect(buildPhase?.shellScript?.contains(expectedScript) == true)
         #expect(buildPhase?.inputPaths.contains("$BUILD_DIR/$CONFIGURATION/\(macroExecutable.productName)") == true)
         #expect(


pepicrft:
Thanks @ffittschen. The changes make sense.
I'm taking the opportunity to fail a regression in one of our acceptance tests.

PR #9966, opened 2026-03-21
fix(cli): support .embedInCode SPM resource rule in PackageInfo decoder

### LINKED ISSUE #9965: SPM `.embedInCode` resource rule causes decoding error when generating project
## What happened?

When a local SPM package declares a target with the `.embedInCode(_:)` resource rule (available since PackageDescription 5.9), Tuist fails to decode the package descriptor produced by `swift package dump-package` and the project generation fails.

The root cause is that `PackageInfo.Target.Resource.Rule` only handles `process` and `copy`. The `embedInCode` rule is entirely missing from the enum and its `init(from:)` decoder.

SPM 6.1+ serializes the rule in `swift package dump-package` output as a JSON object:

```json
{
  "rule": {
    "embedInCode": {}
  }
}
```

Neither the Xcode14 enum format decoder nor the single-string fallback matches this format, so the decoder throws a `DecodingError.dataCorrupted` with the message `"Invalid content for Resource decoder"`.

**Expected behavior:** Tuist should recognize `.embedInCode` as a valid resource rule and map it appropriately (e.g. to `.embedInCode` in the generated Xcode target, or at minimum treat it as a known rule to avoid crashing).

## How do we reproduce it?

1. Create a local SPM package (`LocalSwiftPackage/Package.swift`) with a target using `.embedInCode`:

```swift
// swift-tools-version: 6.1
import PackageDescription

let package = Package(
    name: "LocalSwiftPackage",
    products: [
        .library(name: "LocalSwiftPackage", targets: ["LocalSwiftPackage"]),
    ],
    targets: [
        .target(
            name: "LocalSwiftPackage",
            dependencies: [],
            resources: [.embedInCode("Resources")]
        ),
    ]
)
```

2. Reference it in `Tuist/Package.swift`:

```swift
// swift-tools-version: 6.0
@preconcurrency import PackageDescription

#if TUIST
    import struct ProjectDescription.PackageSettings

    let packageSettings = PackageSettings(productTypes: [:])
#endif

let package = Package(
    name: "App",
    dependencies: [
        .package(path: "../LocalSwiftPackage"),
    ]
)
```

3. Reference the product in `Project.swift`:

```swift
import ProjectDescription

let project = Project(
    name: "App",
    targets: [
        .target(
            name: "App",
            destinations: .iOS,
            product: .app,
            bundleId: "dev.tuist.App",
            sources: ["App/Sources/**"],
            dependencies: [
                .external(name: "LocalSwiftPackage"),
            ]
        ),
    ]
)
```

4. Run `tuist generate`
5. Observe the decoding error

A fixture reproducing the issue has been added at `examples/xcode/generated_ios_app_with_spm_embed_in_code_resources` (see linked PR).

## Error log

```sh
error: dataCorrupted(Swift.DecodingError.Context(
  codingPath: [..., CodingKeys(stringValue: "resources"), Index 0, CodingKeys(stringValue: "rule")],
  debugDescription: "Invalid content for Resource decoder",
  underlyingError: nil
))
```

The error originates in `PackageInfo.Target.Resource.Rule.init(from:)` and surfaces via `PackageInfoLoader` when Tuist calls `swift package dump-package` and decodes the result.

## macOS version

26.3.1

## Tuist version

4.162.1

## Xcode version

26.3 (17C529)

stefanomondino:
## Summary

- Add `embedInCode` case to `PackageInfo.Target.Resource.Rule` enum and its `Codable` decoder to handle the `{"embedInCode": {}}` JSON format produced by `swift package dump-package` (PackageDescription 5.9+)
- Handle `.embedInCode` in `PackageInfoMapper` by treating it like `.process` (file is included in the generated target's resources)
- Add unit test for JSON decoding of the `embedInCode` rule format
- Add unit test for resource mapping when a target uses `.embedInCode`
- Add fixture `generated_ios_app_with_spm_embed_in_code_resources` to reproduce the issue

Closes #9965

## Test plan

- [x] `TuistLoaderTests/PackageInfoMapperTests/map_whenHasEmbedInCodeResources` — new test, passes
- [x] `XcodeGraphTests/PackageInfoTests/resourceRule_decodesEmbedInCode` — new test, passes
- [x] Full `PackageInfoMapperTests` suite (117 tests) — all pass

### DIFF
--- a/cli/Sources/XcodeGraph/Sources/XcodeGraph/PackageInfo.swift
+++ b/cli/Sources/XcodeGraph/Sources/XcodeGraph/PackageInfo.swift
@@ -442,6 +442,7 @@ public struct PackageInfo: Equatable, Hashable, Codable { // swiftlint:disable:t
             public enum Rule: String, Codable, Hashable {
                 case process
                 case copy
+                case embedInCode
 
                 public init(from decoder: Decoder) throws {
                     // Xcode 14 format


pepicrft:
@stefanomondino thanks a lot for putting this together. Before I review, would you mind sharing more about what prompted this work? (the motivation/need)

stefanomondino:
@pepicrft sure thing! We recently received a closed source framework (that I can't share unfortunately) that had this embedInCode definition in their SPM Package.swift file.
We were unable to integrate it in our codebase unless we manually (with a script) changed it to `.process` right before generating the project.
The `embedInCode` definition is a [valid one](https://developer.apple.com/documentation/packagedescription/resource/embedincode(_:)) from SPM but Tuist is not mapping it properly. Probably it's a less-used one and nobody ever experienced that.

But we couldn't ask the framework maintainers to change their SPM because it's formally correct and works with regular SPM.

I've included a fixture in this PR with the failing use case and the fix. `tuist generate` fails generating the dump for the SPM during the encoding phase.
Shouldn't impact too many people at the moment, and as far as I've understood "process" and "embedInCode" should be almost equivalent

cli/Tests/TuistGeneratorAcceptanceTests/GenerateAcceptanceTests.swift
```diff
@@ -1358,6 +1358,15 @@ struct GenerateAcceptanceTesAppWithLocalSPMModuleWithRemoteDependencies {
     }
 }
 
+struct GenerateAcceptanceTestAppWithLocalSPMEmbedInCodeResources {
+    @Test(.withFixture("generated_ios_app_with_spm_embed_in_code_resources"), .inTemporaryDirectory)
+    func app_with_local_spm_embed_in_code_resources() async throws {
+        try await run(InstallCommand.self)
+        try await run(GenerateCommand.self)
+        try await run(BuildCommand.self)
```
pepicrft: I think to properly test it e2e, we'd need to launch the app to exercise the access to the embedded resource?

stefanomondino:
@pepicrft and on top of that, it seems like it's something that has changed in how `swift package dump-package` generates the JSON dump because it wasn't happening when changing the package's declaration to swift 6.0 instead of 6.1.
So it seems like it's the combination of `embedInCode` and swift `// swift-tools-version: 6.1` in the target package to cause the issue. I'm not 100% sure about this but the fixture example in this PR should be enough to trigger the error (without the fix, on current Tuist version on main) and solve it (with the fix in this PR).
Also, the CI/CD errors seems to be unrelated / flaky (problems with registry / alamofire tests which I'm sure you are already aware of)

pepicrft:
You can ignore the failing acceptance test, it's unrelated to your changes.
It's likely that the schema is bound to the version of the Swift tools that you are using.

stefanomondino:
@pepicrft not sure about what you mean here.

> It's likely that the schema is bound to the version of the Swift tools that you are using.

For sure it is, but isn't that a legit configuration that Tuist is supposed to support?

And, most important: does the included example fixture here only fails on my end?

Issue #9597, opened 2026-02-25
CoreData Models for static frameworks are copied to staticFramework and not Copied to the Main App bundle


ebubedimobi:
### What happened?

**Bug Report: CoreData Model Resources Not Copied to Main App Bundle When Using Static Frameworks**

i have been using 4.34.3 for a while now, and decided to upgrade to the latest version 4.152.0 as at time of creating report

**Description**
When a framework is linked as a staticFramework and contains either:
- Resources with *.xcdatamodeld files, or
- A coreDataModel

The resources are incorrectly copied to the staticFramework instead of the main app bundle. This causes a crash in the Swift Bundle accessor since staticFrameworks cannot hold their own resources.

**Steps to Reproduce for 4.152.0**
1. Create a staticFramework that includes a CoreData model (.xcdatamodeld)
2. Link this framework to an app target
3. Build the project
4. Observe that CoreData resources are copied to the static framework itself instead of the main app bundle

**Expected Behavior**
Resources should be copied to the final app product without any duplicates in "Frameworks/" path (same behavior as in Tuist 4.34.3)

**Actual Behavior**
Resources are copied to the staticFramework and static framework is seen in "Frameworks/" path, causing a runtime crash when accessing the bundle.

**Regression Information (as far as i could go)**
- Works correctly in: Tuist 4.34.3
- Tuist 4.122.1 - works correctly partially, coredata resources for staticFrameworks are copied to main app product but are still duplicated in the "Frameworks/" along with the target itself path
- Tuist 4.122.2 - all staticFrameworks resources where embedded inside themselves in framework path(wrong behaviour) - https://github.com/tuist/tuist/pull/9081/changes#diff-fd82aca7a4046536b40e4923717b57aa7d9b94763a1ec03805753747fdd6a38a
- Version 4.146.0 tried to fix this reverting a commit - https://github.com/tuist/tuist/commit/fbeb6cecbacb523fef629c941dfbf9e23d63592d but this didn't fix for coreData Models. The fix resolved issues for other resources, but CoreData models remain affected from version 4.146.0 onward.

### How do we reproduce it?

1. Create a staticFramework target that includes a coreDataModel(.xcdatamodeld)
2. Link this framework to an app target
3. Build the project and check the build phases - the resources are not copied to the main app bundle but copied to the staticFramework
4. Run the app - it crashes at runtime when trying to access the bundle with Swift Bundle Accessor

[...]

PR #9970, opened 2026-03-22
fix(cli): fix static framework core data bundle crash

AppleFlash:
Resolves https://github.com/tuist/tuist/issues/9597

When a static framework target contains CoreData models, Tuist incorrectly embeds the framework as a dynamic one and misplaces the resource bundle. This causes a crash at runtime:
`Fatal error: unable to find bundle named Framework_Framework`.

### Root cause

`ResourcesProjectMapper` delegates resources (including `.xcdatamodeld`) to a companion bundle target, but didn't mark the original target as having done so. As a result, `GraphTraverser` still treated the static framework as if it contained resources and added it to the `Embedded Frameworks` build phase. This turns the static framework effectively dynamic — evidenced by the binary inspection on device:
```
$ file App.app/Frameworks/Framework.framework/Framework
Mach-O universal binary with 1 architecture: [arm64:Mach-O 64-bit dynamically linked shared library arm64]
```
And an empty `__TEXT` segment in the linked binary:
```
Section
  sectname __text
   segname __TEXT
      addr 0x00000000000002e8
      size 0x0000000000000000
    offset 744
     align 2^0 (1)
    reloff 0
    nreloc 0
     flags 0x80000400
```

The bundle ends up inside `Frameworks/Framework.framework` rather than next to the app binary, so `TuistBundle+Framework.swift` cannot locate it at runtime, causing a crash: `Fatal error: unable to find bundle named Framework_Framework`.

### Fix

- Tag the source target with `tuist:resourcesDelegatedToBundle` in `ResourcesProjectMapper` once its resources are moved to the companion bundle.
- In `GraphTraverser.canHostResources` and `isEmbeddableDependencyTarget`, skip static frameworks that carry this tag so they are no longer embedded and the bundle is placed correctly.

### How to test locally

1. Open the example project:
```
cd examples/xcode/generated_ios_app_with_coredata_in_static_framework
tuist generate
```
2. Verify in `App → Build Phases`:
   - `Embedded Frameworks` is **empty**
   - `Dependencies` phase is present and contains `Framework_Framework.bundle`
3. Run the app on a simulator — no crash should occur.
4. Run `ResourcesProjectMapperTests` in `TuistGeneratorTests`.
5. Run `ios_app_with_coredata_in_static_framework` in `TuistGeneratorAcceptanceTests`.

### DIFF
--- a/cli/Sources/TuistCore/Graph/GraphTraverser.swift
+++ b/cli/Sources/TuistCore/Graph/GraphTraverser.swift
@@ -272,7 +272,9 @@ public class GraphTraverser: GraphTraversing {
         guard let target = graph.projects[path]?.targets[name] else { return [] }
 
         func canHostResources(target: Target) -> Bool {
-            target.supportsResources && (target.product != .staticFramework || target.containsResources)
+            let isStatic = target.product == .staticFramework
+            let isDelegatedToBundle = target.metadata.tags.contains("tuist:resourcesDelegatedToBundle")
+            return target.supportsResources && (!isStatic || (target.containsResources && !isDelegatedToBundle))
         }
 
         guard canHostResources(target: target) else { return [] }
@@ -1501,7 +1503,9 @@ public class GraphTraverser: GraphTraversing {
 
     private func isEmbeddableDependencyTarget(dependency: GraphDependency) -> Bool {
         testTarget(dependency: dependency) {
-            $0.product.isDynamic || $0.product == .staticFramework && $0.containsResources
+            let isStatic = $0.product == .staticFramework
+            let isDelegatedToBundle = $0.metadata.tags.contains("tuist:resourcesDelegatedToBundle")
+            return $0.product.isDynamic || isStatic && $0.containsResources && !isDelegatedToBundle
         }
     }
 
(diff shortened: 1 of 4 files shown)


AppleFlash:
@fortmarek @ivan-gaydamakin @pepicrft may be you could review this changes?

pepicrft:
@AppleFlash I'll take over this branch if that's fine and finish it.

pepicrft:
@AppleFlash could you check if the changes [here](https://github.com/tuist/tuist/pull/10003) fix your issue. Tuist should not treat static frameworks with resources as an embeddable thing, and if it does this is likely a leftover from where we tried to make them work (Apple said they were supported). If it does, I suggest we merge that one instead.

AppleFlash:
> @AppleFlash could you check if the changes [here](https://github.com/tuist/tuist/pull/10003) fix your issue. Tuist should not treat static frameworks with resources as an embeddable thing, and if it does this is likely a leftover from where we tried to make them work (Apple said they were supported). If it does, I suggest we merge that one instead.

@pepicrft I've checked, it works perfectly. Thanks!

AppleFlash:
Fixed [here](https://github.com/tuist/tuist/pull/10003)

PR #9978, opened 2026-03-23
feat(cli,server): run quarantined tests instead of skipping them

fortmarek:
Screenshot 2026-03-23 at 17 51 59

## Summary

- **Quarantined tests now run** instead of being skipped (`-skip-testing`/`excludeTestsMatching`). This allows collecting test data to determine when quarantined tests are fixed, enabling automatic unquarantine.
- **Exit code is overridden to 0** when only quarantined tests fail (CI passes). Non-quarantined failures still cause non-zero exit.
- **Dashboard reflects quarantine-aware status**: test run shows "Passed" when only quarantined tests fail, failures card hides quarantined failures, and quarantine badges are shown on test case runs.

### CLI (`tuist test`)
- Fetch quarantined tests but don't add them to skip list — they run normally
- After test failure, classify failures as quarantined vs real using `classifyFailures()`
- If only quarantined tests fail: log summary, return exit 0
- If mixed: log both counts, re-throw (non-zero exit)

### CLI (`tuist xcodebuild test`)
- Add quarantine support (fetch list, parse xcresult, classify failures)
- Same exit code override behavior as `tuist test`

### Gradle Plugin
- Replace `excludeTestsMatching` with `ignoreFailures = true`
- Add `doLast` block that classifies failures and throws `GradleException` only for non-quarantined failures
- Add `matchesQuarantinePattern()` for pattern matching

### Server Dashboard
- Override displayed test run status to "Passed" when all failures are from quarantined tests
- Filter quarantined failures from Failures card and tab
- Adjust "Failed test cases" metric to exclude quarantined
- Add "Quarantined" badge on test case runs (test run detail + test case run detail)
- Show pass/fail status indicator on quarantined tests dashboard

### Exit Code Rules
| Scenario | Exit Code |
|----------|-----------|
| All tests pass | 0 |
| Only quarantined tests fail | 0 |
| Any non-quarantined test fails | non-zero |
| Network failure fetching quarantine list | treat all failures as real |

RFC: https://community.tuist.dev/t/adjust-quarantine-behavior/945

## Test plan

- [x] Updated existing `test_run_fetches_quarantined_tests_and_runs_them` test to verify quarantined tests are NOT skipped
- [x] Added `matchesQuarantinePattern` tests for Gradle plugin
- [ ] Manual: `tuist test` with quarantined tests — verify they run and exit 0 when only quarantined fail
- [ ] Manual: verify dashboard shows "Passed" status and hides quarantined failures
- [ ] Manual: verify Gradle plugin with quarantined tests

### DIFF
--- a/cli/Tests/TuistKitTests/Services/XcodeBuildTestCommandServiceTests.swift
+++ b/cli/Tests/TuistKitTests/Services/XcodeBuildTestCommandServiceTests.swift
@@ -185,6 +185,11 @@ struct XcodeBuildTestCommandServiceTests {
                 .loadConfig(path: .any)
                 .willReturn(.test(fullHandle: "tuist/tuist"))
 
+            xcResultService.reset()
+            given(xcResultService)
+                .parse(path: .any, rootDirectory: .any)
+                .willReturn(TestSummary(testPlanName: nil, status: .passed, duration: 0, testModules: []))
+
             given(xcodeBuildArgumentParser)
                 .parse(.any)
                 .willReturn(


cli/Sources/TuistKit/Services/XcodeBuild/XcodeBuildTestCommandService.swift
```diff
@@ -205,21 +244,27 @@ struct XcodeBuildTestCommandService {
         return arguments[valueIndex]
     }
 
-    private func inspectResultBundleIfNeeded(
-        resultBundlePath: AbsolutePath?,
+    private func rootDirectory() async -> AbsolutePath? {
+        guard let workingDirectory = try? await Environment.current.currentWorkingDirectory() else {
```
pepicrft, line 248: I got a bit confused by this, how is this directory used? I'd have expected the directory returned by the root directory locator to be the default one.

cli/Sources/TuistKit/Services/TestService.swift
```diff
@@ -883,7 +893,13 @@ public struct TestService { // swiftlint:disable:this type_body_length
                 "built"
             }
 
-        AlertController.current.success(.alert("The project tests \(verb) successfully"))
+        if !quarantinedTests.isEmpty {
+            AlertController.current.success(
```
pepicrft, line 1033: Do we log the names somewhere? Users might want to see this information output somewhere.

cli/Tests/TuistKitTests/Services/TestQuarantineServiceTests.swift
```diff
@@ -0,0 +1,377 @@
...
+            serverEnvironmentService: serverEnvironmentService
+        )
+    }
+
+    // MARK: - quarantinedTests
+
+    @Test(.withMockedDependencies())
+    func quarantinedTests_returnsEmpty_whenSkipQuarantineIsTrue() async throws {
+        let config = Tuist.test(fullHandle: "org/project")
+        let result = await subject.quarantinedTests(config: config, skipQuarantine: true)
```
pepicrft, line 32: If I understood the semantics right, does this make sense?
```suggestion
        let result = await subject.quarantinedTests(config: config, skipQuarantined: true)
```

cli/Tests/TuistKitTests/Services/TestQuarantineServiceTests.swift
```diff
@@ -0,0 +1,377 @@
...
+            )
+            .willReturn(response)
+
+        let result = await subject.quarantinedTests(config: config, skipQuarantine: false)
+
+        let expected0 = try TestIdentifier(target: "AppTests", class: "FooSuite", method: "testFoo()")
+        let expected1 = try TestIdentifier(target: "CoreTests", class: nil, method: "testBar()")
+        #expect(result.count == 2)
+        #expect(result[0] == expected0)
+        #expect(result[1] == expected1)
```
pepicrft: What about comparing lists instead?

cli/Tests/TuistKitTests/Services/TestQuarantineServiceTests.swift
```diff
@@ -0,0 +1,377 @@
...
+                    ),
+                ]),
+            ]
+        )
+
+        #expect(subject.onlyQuarantinedTestsFailed(testSummary: summary) == false)
+    }
+}
+
+extension Components.Schemas.TestCase {
```
pepicrft, line 346: What about making these extensions `fileprivate`?

cli/Sources/TuistKit/Services/TestService.swift
fortmarek: We don't log them, but they are available in the dashboard's run detail, which I think should be enough.

And the full list should be available in the verbose logs.

cli/Tests/TuistKitTests/Services/TestQuarantineServiceTests.swift
fortmarek: I'd keep the current name which is derived from the flag `--skip-quarantine` which we already have in `tuist test`. I can see the reasoning behind this, but would rather keep aligned with the current flag name.

pepicrft: Sounds good

fortmarek:
Regarding [the rootDirectory comment](https://github.com/tuist/tuist/pull/9978#discussion_r2976753876): This was addressed — replaced the custom git-based `rootDirectory()` with `RootDirectoryLocator.locate(from:)` which is the standard utility.

cli/Tests/TuistKitTests/Services/TestQuarantineServiceTests.swift
fortmarek: done!

cli/Tests/TuistKitTests/Services/TestQuarantineServiceTests.swift
fortmarek: done!

fortmarek: It's passed to `xcResultService.parse(path:rootDirectory:)` which uses it to resolve failure file paths relative to the project root. The method already delegates to `RootDirectoryLocator.locate(from:)` — it's just a small helper to avoid repeating the `currentWorkingDirectory` + `locate` boilerplate at the two call sites.

cli/Sources/TuistKit/Services/XcodeBuild/XcodeBuildTestCommandService.swift
fortmarek: It's passed to `xcResultService.parse(path:rootDirectory:)` which uses it to resolve failure file paths relative to the project root. The method already delegates to `RootDirectoryLocator.locate(from:)` — it's just a small helper to avoid repeating the `currentWorkingDirectory` + `locate` boilerplate at the two call sites.

PR #9979, opened 2026-03-23
fix(server): share clone-local dev instance scoping with cache

pepicrft:
This PR enables running multiple instances of the main and cache servers concurrently from multiple directories, something we found ourselves doing in the agentic world. 
The solution I ended up going with was configuring `mise` source a script that lazy-creates a state file with a number, which then uses to construct ports and names used to configure things at runtime. The configurable pieces that I identified (let me know if I missed any) are:

- ClickHouse DB name
- Postgres DB name
- Minio console and dashboard port
- Server port
- Cache server port

### DIFF
--- a/.github/workflows/gradle-cache-acceptance.yml
+++ b/.github/workflows/gradle-cache-acceptance.yml
@@ -121,8 +121,9 @@ jobs:
         working-directory: server
         run: |
           MIX_ENV=dev mix phx.server &
+          server_url="${TUIST_SERVER_URL:-http://localhost:8080}"
           for i in {1..30}; do
-            if curl -sf http://localhost:8080 >/dev/null 2>&1; then
+            if curl -sf "$server_url" >/dev/null 2>&1; then
               echo "Main server is ready"
               break
             fi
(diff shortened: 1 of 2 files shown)


server/README.md
```diff
@@ -36,8 +36,6 @@ Pass: tuistrocks
 #### To run additional features
 1. Clone the repository: `https://github.com/tuist/tuist.git`.
 1. Go to `tuist/examples/xcode/generated_ios_app_with_frameworks`.
-1. Change the url in `Tuist/Config.swift` to `http://localhost:8080`.
+1. Change the url in `Tuist/Config.swift` to the clone-specific local URL, for example `http://localhost:8523`.
```
cschmatzler, line 39: Can we instead, via mise, also set `TUIST_URL` for the CLI? Not sure how much friction that is moving the `_.source` bit one level higher to the repo root, but would be fewer manual steps (that can be forgotten by me :D) for the engineer checking it out.

pepicrft: Something we can do is make configuration options overridable via flags when invoking the CLI:
```
tuist ..... --config.url $TUIST_SERVER_URL
```

Defaulting to getting the value from the environment can lead to confusion, since you might expect the production environment instead.

PR #9982, opened 2026-03-23
feat(server): add custom OAuth2 SSO support

pepicrft:
One of our soon-to-be customers has a custom SSO solution that they'd like to integrate Tuist with, and it implements the OAuth2 interface, so I'm adding an option to configure a custom solution.

## Summary
- add customer-managed custom OAuth2 SSO support alongside existing Google and Okta flows
- add dashboard configuration for custom OAuth2 provider URL, endpoints, and client credentials
- add migration, auth flow wiring, gettext templates, and focused server-side tests

I adjusted the design that we have already for Okta, just replacing the field names:
image

### DIFF
--- a/server/lib/tuist/oauth2/sso_client.ex
+++ b/server/lib/tuist/oauth2/sso_client.ex
@@ -26,7 +26,6 @@ defmodule Tuist.OAuth2.SSOClient do
                {"accept", "application/json"},
                {"authorization", "Basic " <> Base.encode64("#{client_id}:#{client_secret}")}
              ],
-             finch: Tuist.Finch,
              connect_options: SSRFGuard.connect_options(hostname),
              decode_body: false
            ) do
@@ -49,7 +48,6 @@ defmodule Tuist.OAuth2.SSOClient do
                {"authorization", "Bearer #{access_token}"},
                {"accept", "application/json"}
              ],
-             finch: Tuist.Finch,
              connect_options: SSRFGuard.connect_options(hostname),
              decode_body: false
            ) do


server/lib/tuist_web/live/sso_settings_live.html.heex
```diff
@@ -55,6 +55,7 @@
           >
             <:item value="google" label="Google" />
             <:item value="okta" label="Okta" />
+            <:item value="custom_oauth2" label={dgettext("dashboard_account", "Custom OAuth2")} />
```
fortmarek:
```suggestion
            <:item value="custom_oauth2" label={dgettext("dashboard_account", "OAuth2")} />
```

I'd probably call this just OAuth2

server/lib/tuist/accounts/custom_oauth2.ex
```diff
@@ -0,0 +1,40 @@
...
+
+  def config_for_organization(_organization) do
+    {:error, :custom_oauth2_not_configured}
+  end
+
+  def normalize_site(site) when is_binary(site) do
+    site
+    |> String.trim()
+    |> String.trim_trailing("/")
+  end
```
fortmarek: should the normalization be done when storing the values in the db instead?

server/lib/tuist/accounts/oauth2_identity.ex
```diff
@@ -7,7 +7,7 @@ defmodule Tuist.Accounts.Oauth2Identity do
   import Ecto.Changeset
 
   schema "oauth2_identities" do
-    field :provider, Ecto.Enum, values: [github: 0, okta: 1, google: 2, apple: 3]
+    field :provider, Ecto.Enum, values: [github: 0, okta: 1, google: 2, apple: 3, custom_oauth2: 4]
```
fortmarek: I get the reasoning by `custom_oauth2` but I think I'd go with just `oauth2: 4`, although I understand the other providers are also using oauth2 under the hood. Don't feel strongly about this, though.

server/lib/tuist/accounts/organization.ex
```diff
@@ -8,13 +8,25 @@ defmodule Tuist.Accounts.Organization do
...
     field :sso_organization_id, :string
     field :sso_enforced, :boolean, default: false
     field :okta_client_id, :string
-    field :okta_encrypted_client_secret, Tuist.Vault.Binary
+    field :okta_encrypted_client_secret, Binary
+    field :custom_oauth2_client_id, :string
+    field :custom_oauth2_encrypted_client_secret, Binary
+    field :custom_oauth2_authorize_url, :string
+    field :custom_oauth2_token_url, :string
+    field :custom_oauth2_user_info_url, :string
```
fortmarek: what do you think about migrating away from `okta_` fields and having just `oauth2_` fields? In the end, the Okta integration maps 1-1 to these `oauth2` values, it's just that we can derive some of them from less.

server/lib/tuist/accounts/custom_oauth2.ex
pepicrft: Yes. I'll move it to save/update time.

server/lib/tuist/accounts/organization.ex
pepicrft: I like the idea. I just did that.

server/lib/tuist/accounts.ex
```diff
@@ -1618,69 +1602,26 @@ defmodule Tuist.Accounts do
...
-  def okta_organization_for_user_email(email) do
+  def sso_organization_for_user_email(email) do
     with {:ok, user} <- get_user_by_email(email),
-         organization when not is_nil(organization) <- user_okta_organization(user) do
+         organization when not is_nil(organization) <- user_sso_organization(user) do
       {:ok, organization}
     else
       _ ->
-        # If user doesn't exist or has no SSO organization, try domain-based matching
-        okta_organization_for_email_domain(email)
```
cschmatzler, line 1636: This removes  the fallback domain-based matching entirely from what I can tell.

server/lib/tuist/accounts/organization.ex
```diff
@@ -45,13 +63,103 @@ defmodule Tuist.Accounts.Organization do
...
+            invalid_url_error(field, url)
+          end)
+        end)
+
+      _ ->
+        changeset
+    end
+  end
+
+  defp valid_oauth2_url?(url) do
```
cschmatzler: This needs some sort of blocking of private addresses; without, this can be pointed to the server itself or any other Render-or-other-private-infrastructure address and make unauthorized requests from our server.

server/lib/tuist_web/live/sso_settings_live.ex
```diff
@@ -150,19 +152,11 @@ defmodule TuistWeb.SSOSettingsLive do
...
 
-    attrs = %{sso_organization_id: domain, okta_client_id: client_id, sso_enforced: socket.assigns.sso_enforced}
-
-    attrs =
-      if client_secret == "",
-        do: attrs,
-        else: Map.put(attrs, :okta_client_secret, client_secret)
-
-    {:ok, updated_organization} = Accounts.update_okta_configuration(organization.id, attrs)
+    {:ok, updated_organization} = Accounts.update_sso_configuration(organization.id, sso_provider, attrs)
```
cschmatzler: This should also handle `{:error, changeset}`; entering non-valid values crashes the page right now.
```
[error] GenServer #PID<0.3072.0> terminating
** (MatchError) no match of right hand side value:

    {:error,
     #Ecto.Changeset<
       action: :update,
       changes: %{
         sso_provider: :oauth2,
         sso_organization_id: "not-a-url",
         oauth2_client_id: "not-a-url",
         oauth2_authorize_url: "not-a-url",
         oauth2_token_url: "not-a-url",
         oauth2_user_info_url: "not-a-url",
         oauth2_encrypted_client_secret: "not-a-url"
       },
       errors: [
         oauth2_user_info_url: {"must be a valid URL", []},
         oauth2_token_url: {"must be a valid URL", []},
         oauth2_authorize_url: {"must be a valid URL", []},
         sso_organization_id: {"must be a valid URL", []}
       ],
       data: #Tuist.Accounts.Organization<>,
       valid?: false,
       ...
     >}

    (tuist 0.1.0) lib/tuist_web/live/sso_settings_live.ex:159: TuistWeb.SSOSettingsLive.save_oauth2_sso/2
```

server/lib/tuist/accounts.ex
pepicrft: Yes. I brought it back.

server/lib/tuist/accounts/organization.ex
pepicrft: Good point. I extended the validation to include checks against private URLs.

server/lib/tuist/url.ex
```diff
@@ -0,0 +1,43 @@
...
+  def public_url?(_), do: false
+
+  defp private_host?(host) do
+    case :inet.parse_address(String.to_charlist(host)) do
+      {:ok, ip} -> private_ip?(ip)
+      {:error, _} -> private_hostname?(host)
+    end
+  end
+
+  defp private_ip?({127, _, _, _}), do: true
```
pepicrft, line 61: Iterating on this one.

server/lib/tuist_web/controllers/auth_controller.ex
```diff
@@ -25,64 +30,80 @@ defmodule TuistWeb.AuthController do
...
+        route_provider = get_session(conn, :sso_route_provider) || :oauth2
 
+        with {:ok, %Organization{} = organization} <- Accounts.get_organization_by_id(organization_id),
+             {:ok, config} <- Accounts.oauth2_config_for_organization(organization),
+             {:ok, auth} <- sso_auth(conn, config, organization.sso_provider, route_provider) do
+          conn
+          |> delete_session(:sso_organization_id)
+          |> delete_session(:sso_state)
+          |> delete_session(:sso_route_provider)
+          |> complete_oauth_callback(auth)
```
cschmatzler: Passing the synthesized auth from the configured OAuth2 endpoints into complete_oauth_callback/2 also enables its existing email-based auto-linking behavior for `:oauth2`. Because org admins can now choose arbitrary authorize/token/userinfo URLs, a malicious admin could return `pedro@tuist.dev` from /userinfo and get logged in as that Tuist user. For custom OAuth2 providers, we need something stronger than just linking by email.

server/lib/tuist/url.ex
```diff
@@ -0,0 +1,75 @@
+defmodule Tuist.URL do
+  @moduledoc ~S"""
+  URL validation utilities including SSRF protection.
+  """
+
+  def public_url?(url) when is_binary(url) do
+    case URI.parse(url) do
+      %URI{scheme: scheme, host: host, query: nil, fragment: nil}
+      when scheme in ["http", "https"] and is_binary(host) and host != "" ->
```
cschmatzler: imo this should only allow HTTPS unless in dev? and test?

server/lib/tuist_web/live/user_okta_login_live.ex
```diff
@@ -27,7 +27,7 @@ defmodule TuistWeb.UserOktaLoginLive do
   def handle_event("submit", %{"user" => %{"email" => email}}, socket) do
     email = String.trim(email)
 
-    case Accounts.okta_organization_for_user_email(email) do
+    case Accounts.sso_organization_for_user_email(email) do
```
cschmatzler: This can return organizations with Google or custom OAuth2; we should check for provider here and redirect if `organization.sso_provider` is not Okta.

server/lib/tuist/url.ex
pepicrft: Agree. I adjusted it to be https-only in production-like environments.

server/lib/tuist_web/live/user_okta_login_live.ex
pepicrft: Good catch. I added logic to handle the other scenarios.

server/lib/tuist_web/controllers/auth_controller.ex
```diff
@@ -294,21 +298,104 @@ defmodule TuistWeb.AuthController do
...
+    |> Client.put_serializer("application/json", JSON)
+  end
 
-    case params["login_hint"] do
-      login_hint when is_binary(login_hint) ->
-        default_oauth2_params = [scope: "openid email profile"]
-        oauth2_params = Keyword.put(default_oauth2_params, :login_hint, login_hint)
-        Keyword.put(strategy_options, :oauth2_params, oauth2_params)
+  defp sso_auth(conn, config, sso_provider, route_provider) do
+    params = [code: conn.params["code"]]
```
cschmatzler: [P1] This assumes every callback includes a `code`. When the provider redirects back with `error=access_denied` and no code, `AuthCodeBasicAuth.get_token/3` raises on the missing parameter and this request becomes a 500 instead of following the existing authentication failure path. Please handle `error` or missing `code` before the token exchange and translate it into `UnauthorizedError`.

server/lib/tuist/accounts/oauth2_identity.ex
```diff
@@ -7,7 +7,7 @@ defmodule Tuist.Accounts.Oauth2Identity do
   import Ecto.Changeset
 
   schema "oauth2_identities" do
-    field :provider, Ecto.Enum, values: [github: 0, okta: 1, google: 2, apple: 3]
+    field :provider, Ecto.Enum, values: [github: 0, okta: 1, google: 2, apple: 3, oauth2: 4]
```
cschmatzler, line 10: [P1] `:oauth2` is now a single shared identity namespace, but we still key rows only by `{provider, id_in_provider}`. OIDC `sub` or `id` is only unique per issuer, so two different customer-configured OAuth2 providers can legally return the same value here. When that happens, the second login will match the first row in `get_oauth2_identity/2` and sign the user into the wrong account before any organization or email checks run.

server/lib/tuist_web/controllers/auth_controller.ex
```diff
@@ -294,21 +298,104 @@ defmodule TuistWeb.AuthController do
...
-      _ ->
-        strategy_options
+  defp validate_sso_urls(config) do
+    urls =
+      Enum.filter(
+        [config.site, config.authorize_url, config.token_url, config.user_info_url],
+        &String.starts_with?(&1, "http")
+      )
+
+    if Enum.all?(urls, &Tuist.URL.public_url?/1) do
```
cschmatzler, line 359: [P1] This only validates the hostname before the request is sent. The actual token and userinfo fetch will resolve DNS again when the HTTP client connects, so an attacker-controlled IdP hostname can answer with a public IP during validation and rebind to loopback, RFC1918, or link-local space at request time. Because these OAuth2 URLs are admin-configurable and fetched server-side, we need connect-time IP enforcement rather than a one-off preflight lookup.

server/lib/tuist/accounts.ex
```diff
@@ -1626,69 +1610,88 @@ defmodule Tuist.Accounts do
...
-  defp user_okta_organization(user) do
+  defp sso_organization_for_email_domain(email) do
+    case String.split(email, "@") do
+      [_username, domain] ->
+        okta_domain = String.replace(domain, ".com", ".okta.com")
+        url_domain = "https://#{domain}"
+
+        case Repo.one(
+               from(o in Organization,
+                 where: not is_nil(o.sso_provider),
```
cschmatzler: [P1] This broadens the domain fallback from Okta-only to any non-`nil` `sso_provider`, which now includes Google organizations. `/users/log_in/sso` only knows how to route `:okta` and `:oauth2`, so entering an email for a Google SSO org now hits a `CaseClauseError` in `SSOLoginLive` instead of redirecting cleanly. This lookup needs to stay limited to providers with an interactive SSO route, or the caller needs to handle `:google` explicitly.

server/lib/tuist/url.ex
```diff
@@ -0,0 +1,79 @@
...
+  defp private_ip?({0, 0, 0, 0}), do: true
+  defp private_ip?({169, 254, _, _}), do: true
+  defp private_ip?({100, second, _, _}) when second >= 64 and second <= 127, do: true
+  defp private_ip?({198, second, _, _}) when second >= 18 and second <= 19, do: true
+
+  # IPv6 private/reserved ranges
+  defp private_ip?({0, 0, 0, 0, 0, 0, 0, 1}), do: true
+  defp private_ip?({0, 0, 0, 0, 0, 0, 0, 0}), do: true
+  defp private_ip?({first, _, _, _, _, _, _, _}) when first >= 0xFC00 and first <= 0xFDFF, do: true
+  defp private_ip?({0xFE80, _, _, _, _, _, _, _}), do: true
```
cschmatzler: [P2] Link-local IPv6 is `fe80::/10`, not just `fe80::/16`. With this match we still accept literals such as `https://[fe90::1]/userinfo` or `https://[feb0::1]/userinfo`, which are local addresses and should be blocked by the SSRF guard.

server/test/tuist_web/live/user_okta_login_live_test.exs
```diff
@@ -69,5 +71,29 @@ defmodule TuistWeb.UserOktaLoginLiveTest do
...
+          sso_provider: :oauth2,
+          sso_organization_id: "https://auth.example.com",
+          oauth2_client_id: "client-id",
+          oauth2_client_secret: "client-secret",
+          oauth2_authorize_url: "https://auth.example.com/oauth2/authorize",
+          oauth2_token_url: "https://auth.example.com/oauth2/token",
+          oauth2_user_info_url: "https://auth.example.com/oauth2/userinfo"
+        )
+
+      {:ok, lv, _html} = live(conn, ~p"/users/log_in/okta")
```
fortmarek: should this page be called differently than okta if it's for all oauth2 flows?

server/priv/repo/migrations/20260324103000_migrate_okta_to_custom_oauth2_fields.exs
```diff
@@ -0,0 +1,38 @@
...
+        oauth2_token_url = 'https://' || sso_organization_id || '/oauth2/v1/token',
+        oauth2_user_info_url = 'https://' || sso_organization_id || '/oauth2/v1/userinfo'
+    WHERE sso_provider = 1
+      AND okta_client_id IS NOT NULL
+    """
+
+    alter table(:organizations) do
+      remove :okta_client_id
+      remove :okta_encrypted_client_secret
+    end
```
fortmarek: I'd consider splitting destructive operations into a different PR once we validate our changes

fortmarek: I'd consider doing destructive operations in a separate PR

server/lib/tuist/oauth2/auth_code_basic_auth.ex
```diff
@@ -0,0 +1,41 @@
...
+    |> put_param(:redirect_uri, client.redirect_uri)
+    |> merge_params(params)
+  end
+
+  @impl true
+  def get_token(client, params, headers) do
+    {code, params} = Keyword.pop(params, :code, client.params["code"])
+
+    if !code do
+      raise OAuth2.Error, reason: "Missing required key `code` for #{inspect(__MODULE__)}"
```
fortmarek: where is this handled? I assume this can be raised when the code passed by the client/sso provider is missing, so we probably don't want this to bubble up to Sentry?

server/lib/tuist_web/live/sso_login_live.ex
```diff
@@ -24,10 +24,18 @@ defmodule TuistWeb.SSOLoginLive do
...
-    case Accounts.okta_organization_for_user_email(email) do
+    case Accounts.sso_organization_for_user_email(email) do
       {:ok, organization} ->
         encoded_email = URI.encode_www_form(email)
-        redirect_url = "/users/auth/okta?organization_id=#{organization.id}&login_hint=#{encoded_email}"
+
+        provider_path =
+          case organization.sso_provider do
+            :okta -> "okta"
+            :oauth2 -> "custom_oauth2"
```
fortmarek:
```suggestion
            :oauth2 -> "oauth2"
```

server/lib/tuist_web/live/sso_settings_live.ex
```diff
@@ -150,26 +151,58 @@ defmodule TuistWeb.SSOSettingsLive do
...
-        do: attrs,
-        else: Map.put(attrs, :okta_client_secret, client_secret)
+      {:error, changeset} ->
+        message =
+          changeset
+          |> Ecto.Changeset.traverse_errors(fn {msg, _opts} -> msg end)
+          |> Enum.map_join(", ", fn {field, msgs} -> "#{field} #{Enum.join(msgs, ", ")}" end)
 
-    {:ok, updated_organization} = Accounts.update_okta_configuration(organization.id, attrs)
+        {:noreply, assign(socket, flash_message: {"error", message})}
```
fortmarek: the flash message appears under the "Single Sign-On" title, but I think it should appear close to the "Save changes" button. As right now if you click on save changes, the flash message out of the view, which is odd.

Also, I'd suggest making the error message a bit more user friendly instead of `sso_organization_id must be a valid URL, oauth2_authorize_url must be a valid URL, oauth2_token_url must be a valid URL, oauth2_user_info_url must be a valid URL`

fortmarek: fwiw, it would be great if we could show the error per the input field instead of a single flash message, but we can leave that as a possible future improvement

server/lib/tuist_web/router.ex
```diff
@@ -726,6 +726,8 @@ defmodule TuistWeb.Router do
     pipe_through :browser_app
     get "/okta", AuthController, :okta_request
     get "/okta/callback", AuthController, :okta_callback
+    get "/custom_oauth2", AuthController, :oauth2_request
+    get "/custom_oauth2/callback", AuthController, :oauth2_callback
```
fortmarek:
```suggestion
    get "/oauth2", AuthController, :oauth2_request
    get "/oauth2/callback", AuthController, :oauth2_callback
```

fortmarek (approved):
Looks good! left a couple of minor comments

server/lib/tuist_web/controllers/auth_controller.ex
pepicrft: Good catch. I added a validate_sso_callback_params! check right after the state validation that catches both the error param case and the missing code case, and raises a clean 401 instead of letting it blow up as a 500. Two tests covering both branches.

server/lib/tuist/accounts/oauth2_identity.ex
pepicrft: You're right, this was a real cross-tenant takeover. Fixed it by scoping the oauth2_identities uniqueness key to include the issuer for okta/oauth2 providers. The lookup now takes provider_organization_id as a third arg, and the DB has partial unique indexes (global for github/google/apple, per-issuer for okta/oauth2). Added a regression test that proves two customers with the same sub from different IdPs can't hijack each other.

server/lib/tuist_web/controllers/auth_controller.ex
pepicrft: Good catch. I fixed this by pinning the URL to the resolved IP before the request fires, so there's no hostname left for a second DNS lookup. The SSO flow now goes through Req/Finch with Mint's hostname: connect option for SNI and cert verification against the original host. 17 SSRFGuard tests and 8 SSOClient tests cover the behavior.

server/lib/tuist/accounts.ex
pepicrft: You're right. The domain fallback query was matching Google orgs too, which would crash SSOLoginLive. I narrowed it to only match :okta and :oauth2, same filter that user_sso_organization already used. Added a test proving Google SSO orgs are skipped.

server/lib/tuist/url.ex
pepicrft: Good catch, fixed. Widened the guard to cover fe80:: through febf:: (the full /10 range). Added tests for fe90::1 and febf::1 to make sure the edges are covered.

server/test/tuist_web/live/user_okta_login_live_test.exs
pepicrft: Actually, you're right. I removed UserOktaLoginLive and the /okta route entirely. Everything goes through /users/log_in/sso now, and the login page button says "SSO" instead of "Okta". Also renamed /users/auth/custom_oauth2 to /users/auth/oauth2 since we dropped the "custom" terminology.

server/lib/tuist/oauth2/auth_code_basic_auth.ex
pepicrft: This file is gone now. I replaced the whole OAuth2.Client path with direct Req/Finch calls. The missing code case is handled earlier by validate_sso_callback_params! which checks for the code param before we ever try the token exchange.

server/lib/tuist_web/live/sso_login_live.ex
pepicrft: The route is /users/auth/custom_oauth2 (not /users/auth/oauth2), so the path string needs to stay as "custom_oauth2" to match. I could rename the route too if you think "custom_oauth2" is confusing, but that would also mean updating the callback URL in every existing Okta/OAuth2 app config.

pepicrft: Done, renamed the route from custom_oauth2 to oauth2 everywhere.

server/lib/tuist_web/router.ex
pepicrft: Done.

server/lib/tuist_web/live/sso_settings_live.ex
pepicrft: Fixed. Errors now show inline next to each field instead of a flash banner at the top. I mapped changeset fields to form fields so things like sso_organization_id show up on the right input depending on the provider.

server/priv/repo/migrations/20260324103000_migrate_okta_to_custom_oauth2_fields.exs
pepicrft: Good call. I split it so the data gets copied but the old columns stay. If something goes wrong we can revert the code and the okta_client_id / okta_encrypted_client_secret columns still have the original data. I'll drop them in a follow-up PR once we validate.

PR #9996, opened 2026-03-24
Add command-default cache profile inheritance

bryansum:
## Summary
- add a shared `.commandDefault` base cache profile for custom cache profiles
- resolve custom profiles against an explicit command default so `tuist test` and `tuist generate` can share the same `profiles` model without changing their existing fallback behavior
- preserve the legacy `tuist test` default cache behavior while letting focused `tuist generate` keep using its generate-specific fallback

## Validation
- `mise run cli:lint --fix`
- `xcodebuild build -workspace /Users/bks/src/tuist/Tuist.xcworkspace -scheme ProjectDescription -destination 'platform=macOS' CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `xcodebuild test -workspace /Users/bks/src/tuist/Tuist.xcworkspace -scheme TuistUnitTests -destination 'platform=macOS' -only-testing:TuistKitTests -only-testing:TuistLoaderTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""` currently fails in the `Mockable` dependency's script phase because it copies `MockableMacro` onto itself on macOS, so the scoped tests do not get to execute

### DIFF
--- a/cli/Sources/TuistCache/TargetReplacementDecider.swift
+++ b/cli/Sources/TuistCache/TargetReplacementDecider.swift
@@ -63,7 +63,7 @@ public struct CacheProfileTargetReplacementDecider: TargetReplacementDeciding {
         switch project.type {
         case .external:
             switch resolvedBase {
-            case .none:
+            case .commandDefault, .none:
                 return false
             case .onlyExternal, .allPossible:
                 return true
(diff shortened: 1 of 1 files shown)


pepicrft:
@bryansum I appreciate you putting this together, but before we do any code review, we'd need to understand what the motivation behind it is.

PR #9998, opened 2026-03-24
fix(cli): sort concurrentMap results in content hashers for determinism

fortmarek:
## Summary
- `ContentHashing` protocol is not `Sendable`, so Swift resolves `concurrentMap` to the `withThrowingTaskGroup` overload which returns results in **completion order** rather than insertion order. This caused non-deterministic cache hashes for targets with 2+ input paths in script build phases — the same target would produce different hashes across consecutive runs with no code changes.
- Sort all `concurrentMap` results across the hasher codebase: `TargetScriptsContentHasher`, `ResourcesContentHasher`, `CoreDataModelsContentHasher`, `CopyFilesContentHasher`, and `TargetContentHasher` (buildable folders).
- Also hash embedded script text in `TargetScriptsContentHasher` — previously only script file paths were hashed via `script.path` (which is `nil` for `.embedded` scripts), meaning changes to inline script content would not invalidate the cache.

## Context
Reported by a user who observed that their module had to rebuild on every `tuist cache` invocation despite no code changes. Root cause: the SwiftLint and Sourcery script build phases each had 2 `inputPaths` (`mise.toml` + another file), so the two content hashes were appended in non-deterministic completion order, producing a different `targetScripts` hash each run.

## Test plan
- [x] Updated `TargetScriptsContentHasherTests` expected orderings to match sorted output
- [x] Added `test_hash_targetAction_embeddedScriptTextIsHashed` test
- [x] Added `test_hash_targetAction_differentEmbeddedScriptProducesDifferentHash` test
- [ ] Run full test suite

### DIFF
--- a/cli/Tests/TuistHasherTests/ResourcesContentHasherTests.swift
+++ b/cli/Tests/TuistHasherTests/ResourcesContentHasherTests.swift
@@ -108,47 +108,11 @@ final class ResourcesContentHasherTests: TuistUnitTestCase {
         let got = try await subject.hash(identifier: "resources", resources: resourceFileElements)
 
         // Then
+        // Children are sorted by hash for determinism
         XCTAssertEqual(got, MerkleNode(
-            hash: "b310dc618b8893abc76c9a4b9628e04e",
+            hash: "6c5f573450d379898ffa415402ad349c",
             identifier: "resources",
             children: [
-                MerkleNode(
-                    hash: "e297e7925b4d0d805f87eb8842081511",
-                    identifier: resource1.pathString,
-                    children: [
-                        MerkleNode(
-                            hash: "4a47a0db6e60853dedfcfdf08a5ca249",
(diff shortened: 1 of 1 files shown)


cli/Sources/TuistHasher/CopyFilesContentHasher.swift
```diff
@@ -46,7 +46,7 @@ public struct CopyFilesContentHasher: CopyFilesContentHashing {
                     identifier: file.path.pathString,
                     children: fileChildren
                 )
-            }
+            }.sorted(by: { $0.hash < $1.hash })
```
pepicrft, line 49: Does sorting the resulting hashes have the same effect as ensuring the hashes order is the same as the inputs they were generated from?

fortmarek: Yes, it does 😌

PR #10002, opened 2026-03-24
feat(server): show git branch as Noora badge in dashboard navbar

pepicrft:
## Summary

- Display the current git branch name as a Noora badge (with git-branch icon) in the dashboard headerbar, only in development mode
- Remove the previous approach of prefixing page titles with the branch name
- Use `attention` color with `light-fill` style for clear visibility

image

### DIFF
--- a/server/lib/tuist/environment.ex
+++ b/server/lib/tuist/environment.ex
@@ -13,8 +13,6 @@ defmodule Tuist.Environment do
           {branch, 0} -> String.trim(branch)
           _ -> nil
         end
-      else
-        nil
       end
   end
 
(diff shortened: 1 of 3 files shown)


pepicrft:
> I'm not sure if I would put the branch name into the page title because then we'd rarely see the actual page title teams see.
> 
> I wonder if instead we can include this information in a footer/header/somewhere in the dashboard pages?

Like this?

image

server/lib/tuist/environment.ex
```diff
@@ -6,6 +6,17 @@ defmodule Tuist.Environment do
     @env
   end
 
+  def git_branch do
```
fortmarek: what do you think about making this more generic and be something like `server_version_identifier`? I was thinking it would be also useful for self-hosting teams to know which server version they are on (while we would still default to `nil` for prod)

pepicrft: I like the idea!

PR #10006, opened 2026-03-24
feat(docs): add test sharding documentation

fortmarek:
image

## Summary

- Add documentation for test sharding covering Xcode projects, generated projects (via `tuist test`), and Gradle
- Each platform page documents the two-phase workflow (build + test), options tables, and a GitHub Actions CI example
- Add sidebar navigation entry with icon, placed after Flaky tests
- Add i18n string for the sidebar label
- Instructions for other CI providers (GitLab CI, CircleCI, Buildkite, Codemagic) will come in a follow-up PR once native matrix output support is added for those providers

## Test plan

- [ ] Verify docs build passes (`mise exec -- pnpm run build` in `docs/`)
- [ ] Check sidebar navigation renders correctly
- [ ] Review GitHub Actions examples for accuracy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10009, opened 2026-03-24
feat(cli,gradle): add native shard matrix output for all CI providers

fortmarek:
## Summary

- Add native shard matrix output for **all supported CI providers** (previously only GitHub Actions)
- Use `CIController` for CI provider detection in `ShardMatrixOutputService` instead of manual env var checks
- Add test sharding documentation with CI examples for every provider

### CI provider integrations

| Provider | Output format | Mechanism |
|----------|--------------|-----------|
| **GitHub Actions** | `GITHUB_OUTPUT` matrix variable | Existing — unchanged |
| **GitLab CI** | `.tuist-shard-child-pipeline.yml` | Child pipeline trigger with `.tuist-shard` job template |
| **CircleCI** | `.tuist-shard-continuation.json` | Continuation orb parameters |
| **Buildkite** | `.tuist-shard-pipeline.yml` | `buildkite-agent pipeline upload` |
| **Codemagic** | `CM_ENV` variables | `TUIST_SHARD_MATRIX` + `TUIST_SHARD_COUNT` |
| **Bitrise** | `BITRISE_DEPLOY_DIR` artifact | `.tuist-shard-matrix.json` as build artifact |
| **Other** | `.tuist-shard-matrix.json` | Fallback JSON file |

### Files changed

- `cli/Sources/TuistKit/Services/Sharding/ShardMatrixOutputService.swift` — refactored to use `CIController`, added per-provider output methods
- `gradle/src/main/kotlin/dev/tuist/gradle/TuistTestSharding.kt` — added `detectCIProvider()` and per-provider output in `writeShardMatrixOutput()`
- `docs/docs/en/guides/features/test-sharding/` — new docs with full CI examples for all 6 providers
- `docs/.vitepress/` — sidebar navigation, icon, and i18n string for test sharding

Builds on top of #10006.

## Test plan

- [ ] Verify CLI compiles
- [ ] Verify Gradle plugin compiles
- [ ] Verify docs build passes (`mise run docs:build`)
- [ ] Review generated output format for each CI provider

### DIFF
--- a/cli/Tests/TuistKitTests/Services/Sharding/ShardMatrixOutputServiceTests.swift
+++ b/cli/Tests/TuistKitTests/Services/Sharding/ShardMatrixOutputServiceTests.swift
@@ -19,15 +19,9 @@ struct ShardMatrixOutputServiceTests {
         subject = ShardMatrixOutputService(fileSystem: fileSystem, ciController: ciController)
     }
 
-    private func mockedCWD() async throws -> AbsolutePath {
-        let cwd = try await Environment.current.currentWorkingDirectory()
-        try await fileSystem.makeDirectory(at: cwd)
-        return cwd
-    }
-
     @Test(.withMockedEnvironment())
     func output_github_writesMatrixToGitHubOutput() async throws {
-        let cwd = try await mockedCWD()
+        let cwd = try await Environment.current.currentWorkingDirectory()
         let githubOutputPath = cwd.appending(component: "github_output")
         try await fileSystem.writeText("", at: githubOutputPath, encoding: .utf8)
 
@@ -42,7 +36,7 @@ struct ShardMatrixOutputServiceTests {
 
     @Test(.withMockedEnvironment())
     func output_gitlab_writesChildPipelineYML() async throws {
-        let cwd = try await mockedCWD()
+        let cwd = try await Environment.current.currentWorkingDirectory()
 
         given(ciController).ciInfo().willReturn(.test(provider: .gitlab))
 
@@ -66,7 +60,7 @@ struct ShardMatrixOutputServiceTests {
 
     @Test(.withMockedEnvironment())
     func output_circleci_writesContinuationJSON() async throws {
-        let cwd = try await mockedCWD()
+        let cwd = try await Environment.current.currentWorkingDirectory()
 
         given(ciController).ciInfo().willReturn(.test(provider: .circleci))
 
@@ -83,7 +77,7 @@ struct ShardMatrixOutputServiceTests {
 
     @Test(.withMockedEnvironment())
     func output_buildkite_writesPipelineYML() async throws {
-        let cwd = try await mockedCWD()
+        let cwd = try await Environment.current.currentWorkingDirectory()
 
         given(ciController).ciInfo().willReturn(.test(provider: .buildkite))
 
@@ -106,7 +100,7 @@ struct ShardMatrixOutputServiceTests {
 
     @Test(.withMockedEnvironment())
     func output_codemagic_writesToCMEnv() async throws {
-        let cwd = try await mockedCWD()
+        let cwd = try await Environment.current.currentWorkingDirectory()
         let cmEnvPath = cwd.appending(component: "CM_ENV")
         try await fileSystem.writeText("", at: cmEnvPath, encoding: .utf8)
 
@@ -121,7 +115,7 @@ struct ShardMatrixOutputServiceTests {
 
     @Test(.withMockedEnvironment())
     func output_bitrise_writesToDeployDir() async throws {
-        let cwd = try await mockedCWD()
+        let cwd = try await Environment.current.currentWorkingDirectory()
         let deployDir = cwd.appending(component: "deploy")
         try await fileSystem.makeDirectory(at: deployDir)
 
@@ -136,7 +130,7 @@ struct ShardMatrixOutputServiceTests {
 
     @Test(.withMockedEnvironment())
     func output_noCI_writesFallbackJSON() async throws {
-        let cwd = try await mockedCWD()
+        let cwd = try await Environment.current.currentWorkingDirectory()
 
         given(ciController).ciInfo().willReturn(nil)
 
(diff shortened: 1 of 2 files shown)


cli/Sources/TuistKit/Services/Sharding/ShardMatrixOutputService.swift
```diff
@@ -26,29 +32,127 @@ public struct ShardMatrixOutputService: ShardMatrixOutputServicing {
...
+            at: outputPath,
+            encoding: .utf8,
+            options: [.overwrite]
+        )
+        Logger.current.debug("GitHub Actions matrix output written.")
+    }
+
+    private func writeGitLabCIOutput(indices: [Int]) async throws {
+        let currentDirectory = try await Environment.current.currentWorkingDirectory()
+        let outputPath = currentDirectory.appending(component: ".tuist-shard-child-pipeline.yml")
```
pepicrft, line 72: Does GitLab expect the file relative to the working directory?

pepicrft: Same regarding the other `write` functions. Would it make sense to make this configurable via args?

cli/Sources/TuistKit/Services/Sharding/ShardMatrixOutputService.swift
```diff
@@ -26,29 +32,127 @@ public struct ShardMatrixOutputService: ShardMatrixOutputServicing {
...
+        if try await fileSystem.exists(outputPath) {
+            try await fileSystem.remove(outputPath)
+        }
+        var yaml = ""
+        for index in indices {
+            yaml += "shard-\(index):\n"
+            yaml += "  extends: .tuist-shard\n"
+            yaml += "  variables:\n"
+            yaml += "    TUIST_SHARD_INDEX: \"\(index)\"\n\n"
+        }
```
pepicrft, line 82: Don't we have a YAML encoder that we can use for this instead?

cli/Tests/TuistKitTests/Services/Sharding/ShardMatrixOutputServiceTests.swift
```diff
@@ -0,0 +1,189 @@
...
+        subject = ShardMatrixOutputService(fileSystem: fileSystem, ciController: ciController)
+    }
+
+    private func mockedCWD() async throws -> AbsolutePath {
+        let cwd = try await Environment.current.currentWorkingDirectory()
+        if !(try await fileSystem.exists(cwd)) {
+            try await fileSystem.makeDirectory(at: cwd)
+        }
+        return cwd
+    }
```
pepicrft: I think mockedEnvironment already creates a temporary directory and points cwd to it, no? If so, I think we can assume existence and clean this up.

docs/docs/en/guides/features/test-sharding/generated-projects.md
```diff
@@ -116,3 +124,267 @@ jobs:
       - run: tuist auth login
       - run: tuist test
 ```
+
+### GitLab CI {#gitlab-ci}
```
pepicrft, line 128: Once we move the docs to Elixir, I think this could be a component where we fill in the specifics of every build system. Otherwise, we end up with a lot of duplication across build systems and potential inconssitencies.

gradle/src/test/kotlin/dev/tuist/gradle/TuistTestShardingTest.kt
```diff
@@ -140,6 +168,118 @@ class DeriveReferenceTest {
...
+            File(deployDir, ".tuist-shard-matrix.json").readText()
+        )
+    }
+
+    @Test
+    fun `writeShardMatrixOutput writes fallback json when no CI detected`(@TempDir tempDir: File) {
+        val task = createTask(tempDir)
+        task.writeShardMatrixOutput(listOf(0, 1), "test-ref", shardPlan(), envWith())
+
+        assertEquals(
```
pepicrft: Using string equality in tests feels brittle. I recommend instead to serialize that into some data structure (without having to type & validate) and then assert against the actual content. I'd do this in all the asserts here.

cli/Sources/TuistKit/Services/Sharding/ShardMatrixOutputService.swift
fortmarek: > Same regarding the other write functions. Would it make sense to make this configurable via args?

Wouldn't make this configurable until such a need comes up.

fortmarek: Yes — GitLab's `trigger: include: artifact:` expects the artifact path relative to the project root, which is the working directory where the build job runs.

cli/Sources/TuistKit/Services/Sharding/ShardMatrixOutputService.swift
fortmarek: No Yams or YAML encoder dependency in the CLI. The YAML output is simple enough (just shard index entries) that string building is straightforward here. Adding a YAML library dependency for this felt like overkill.

cli/Tests/TuistKitTests/Services/Sharding/ShardMatrixOutputServiceTests.swift
fortmarek: Fixed — MockEnvironment now creates the CWD directory in its init using FileSystem, so the test helper is gone entirely.

docs/docs/en/guides/features/test-sharding/generated-projects.md
fortmarek: Agreed — once the docs move to Elixir, this should become a shared component with per-build-system slots. For now the duplication is manageable across 3 pages.

gradle/src/test/kotlin/dev/tuist/gradle/TuistTestShardingTest.kt
fortmarek: Updated — JSON outputs are now parsed with Gson and asserted on data structures. YAML outputs (GitLab, Buildkite) use content-based assertions since there's no YAML parser in test dependencies.

PR #10013, opened 2026-03-25
feat(server,cli,skills): add selective testing observability via MCP, API, CLI, and skills

pepicrft:
### API endpoint
- New `GET /:test_run_id/selective-testing-targets` endpoint with `hit_status` filter (miss/local/remote) and pagination

### CLI command
- New `tuist test selection list <test-run-id>` command with `--hit-status` filter, pagination, and `--json` output

### MCP tool
- New `list_xcode_selective_testing_targets` MCP tool: per-target selective testing hit/miss status and hash

### MCP prompt
- New `analyze_selective_testing` prompt: guided workflow for analyzing effectiveness, comparing runs, and diagnosing regressions

### Skill
- New `analyze-selective-testing` skill: CLI-based workflow using `tuist test selection list` for diagnosing selective testing issues

### Housekeeping
- Removes non-deterministic server URL and logo URL from OpenAPI spec
- Fixes `ListBundlesService` and `CreateTestService` argument order after OpenAPI regeneration

> [!NOTE]
> The API route uses `selective-testing-targets` without an `xcode-` prefix, consistent with existing drill-down routes like `module-cache-targets` and `cache-tasks`. The `xcode_` prefix is used in MCP tool names and response field names where build-system disambiguation is needed.

## Test plan

- [x] Server MCP test tools tests pass (16 tests)
- [x] Server tests controller API tests pass (15 tests)
- [x] Full workspace builds successfully
- [ ] Verify on staging

### DIFF
--- a/cli/Sources/TuistServer/Services/ListBundlesService.swift
+++ b/cli/Sources/TuistServer/Services/ListBundlesService.swift
@@ -64,8 +64,8 @@ public struct ListBundlesService: ListBundlesServicing {
                 ),
                 query: .init(
                     page: page,
-                    git_branch: gitBranch,
-                    page_size: pageSize
+                    page_size: pageSize,
+                    git_branch: gitBranch
                 )
             )
         )


server/lib/tuist_web/controllers/api/selective_testing_targets_controller.ex
fortmarek, line 1: This controller is not tested at all

server/lib/tuist_web/controllers/api/selective_testing_targets_controller.ex
```diff
@@ -0,0 +1,155 @@
...
+
+  def index(
+        %{assigns: %{selected_project: selected_project}, params: %{test_run_id: test_run_id} = params} = conn,
+        _params
+      ) do
+    case Tests.get_test(test_run_id) do
+      {:ok, %{project_id: project_id} = run}
+      when project_id == selected_project.id ->
+        page = Map.get(params, :page, 1)
+        page_size = Map.get(params, :page_size, 20)
```
fortmarek: the defaults are already in the openapi spec

cli/Sources/TuistTestCommand/TestSelectionListCommand.swift
```diff
@@ -0,0 +1,65 @@
...
+        help: "The path to the directory or a subdirectory of the project.",
+        completion: .directory
+    )
+    var path: String?
+
+    @Option(
+        name: .long,
+        help: "Filter by hit status (miss, local, or remote)."
+    )
+    var hitStatus: String?
```
fortmarek: you can use an expressible by argument kind protocol conformance instead of using `String?` for better help and we can save ourselves the later conversion.

pepicrft:
@fortmarek what do you think about this consolidation & alignment?

- **Aligned CLI hierarchy**: `tuist test selection list` → `tuist test xcode target list` (mirrors `tuist build xcode target list`)
- **Updated API route**: `/selective-testing-targets` → `/targets`
- **Renamed MCP tool**: `list_xcode_selective_testing_targets` → `list_xcode_test_targets`
- **Renamed service types**: `ListSelectiveTestingTargetsService` → `ListTestTargetsService`

The command hierarchy now follows the same pattern:
```
tuist build xcode target list <build-id>
tuist test  xcode target list <test-run-id>
```

server/lib/tuist_web/controllers/api/selective_testing_targets_controller.ex
pepicrft: Done — removed the redundant defaults since `CastAndValidate` applies them from the OpenAPI spec.

cli/Sources/TuistTestCommand/TestSelectionListCommand.swift
pepicrft: Done — added a `SelectiveTestingHitStatus` enum with `ExpressibleByArgument` conformance. The typed enum flows through the command and service layers, converting to `rawValue` only at the API call boundary.

server/lib/tuist_web/controllers/api/selective_testing_targets_controller.ex
pepicrft: Added a controller test in 476d0edbfb — covers empty list, data serialization, hit_status filtering, pagination, 404 (not found / wrong project), and 403 (unauthorized).

fortmarek:
> what do you think about this consolidation & alignment?

I like it 🙌

PR #10019, opened 2026-03-25
fix(cli): propagate file write errors instead of silently swallowing them

inju2403:
`FileHandler.write(_:path:atomically:)` declares `throws` in its signature but wraps the actual write operation in an empty `catch {}` block, silently swallowing all errors. This means callers expect the method to throw
  on failure, but it never does — file write failures (e.g., disk full, permission denied, invalid path) go completely unnoticed, potentially leading to missing or corrupted project files.

  This PR removes the unnecessary `do-catch` block so that errors from `String.write(to:atomically:encoding:)` are properly propagated to callers. Since all call sites already handle this method with `try` (due to the
  existing `throws` signature), no additional changes are needed.

  ### How to test locally

  1. Set a file path to a read-only location:
     ```swift
     let readOnlyPath = AbsolutePath("/read-only-path/test.txt")
     try fileHandler.write("hello", path: readOnlyPath, atomically: true)
  2. Before this fix: No error is thrown, caller assumes success
  3. After this fix: A proper error is thrown and can be caught by the caller

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -277,11 +277,13 @@ jobs:
           path: .build
           key: ${{ steps.cache-restore.outputs.cache-primary-key }}
 
-  cli-generator-acceptance-tests:
-    name: Generator Acceptance Tests
+  cli-generator-acceptance-tests-build:
+    name: Build Generator Acceptance Tests
     runs-on: macos-26
     timeout-minutes: 60
     if: github.event.pull_request.draft == false || github.event_name != 'pull_request'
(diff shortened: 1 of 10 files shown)


fortmarek:
Hey @inju2403 👋 

Thanks a lot for this PR, however, `FileHandler` is a deprecated component that we shouldn't be using anymore in the first place. I'm [working on](https://github.com/tuist/tuist/pull/10040) fully removing this component from the codebase, which supersedes this PR. Thanks again for the contribution, though!

PR #10020, opened 2026-03-25
feat(server): expire stale in-progress test runs after 6 hours

fortmarek:
## Summary

Sharded test runs stay `in_progress` until all shards report. If a CI run is cancelled before all shards finish, the test run stays stuck forever.

Adds an hourly Oban worker (`ExpireStaleTestRunsWorker`) that marks `in_progress` test runs older than 6 hours as `failure`.

### Future improvement

We could add a GitHub `workflow_run` webhook handler that reacts to `conclusion: "cancelled"` events. The webhook payload contains `id` and `run_attempt`, which map directly to the shard reference (`github-{runId}-{attemptNumber}`). This would give near-instant cleanup instead of waiting up to 6 hours. The hourly worker would remain as a safety net for non-GitHub CI providers or missed webhooks.

## Test plan

- [x] Unit test: stale runs are marked as failure
- [x] Unit test: recent runs are not affected
- [x] Unit test: non-in_progress runs are not affected
- [ ] Server CI passes
- [ ] Worker runs hourly in prod/stag/can

### DIFF
--- a/server/test/tuist/tests/workers/expire_stale_test_runs_worker_test.exs
+++ b/server/test/tuist/tests/workers/expire_stale_test_runs_worker_test.exs
@@ -5,6 +5,8 @@ defmodule Tuist.Tests.Workers.ExpireStaleTestRunsWorkerTest do
   alias Tuist.Tests
   alias Tuist.Tests.Workers.ExpireStaleTestRunsWorker
 
+  setup :verify_on_exit!
+
   describe "perform/1" do
     test "calls Tests.expire_stale_in_progress_test_runs/0" do
       expect(Tests, :expire_stale_in_progress_test_runs, fn -> :ok end)


server/test/tuist/tests/workers/expire_stale_test_runs_worker_test.exs
```diff
@@ -0,0 +1,15 @@
+defmodule Tuist.Tests.Workers.ExpireStaleTestRunsWorkerTest do
+  use TuistTestSupport.Cases.DataCase, async: true
+  use Mimic
+
+  alias Tuist.Tests
+  alias Tuist.Tests.Workers.ExpireStaleTestRunsWorker
+
+  describe "perform/1" do
+    test "calls Tests.expire_stale_in_progress_test_runs/0" do
+      expect(Tests, :expire_stale_in_progress_test_runs, fn -> :ok end)
```
pepicrft, line 12: Should we verify on exit that this function has been called?

PR #10021, opened 2026-03-25
perf(cli): use dictionary lookup for target resolution in PackageInfoMapper

inju2403:
In `PackageInfoMapper.map(packageInfo:...)`, target lookups within the `while` loop used `Array.first(where:)` and `Array.contains(where:)`, both O(n) per call. For packages with many targets (e.g., Firebase with ~100
  targets), this results in O(n*m) complexity.
  >
  > This PR pre-builds a `[String: PackageInfo.Target]` dictionary before the loop, replacing:
  > - `packageInfo.targets.first(where: { $0.name == target })!` → `targetsByName[target]` (O(1), force unwrap removed)
  > - `packageInfo.targets.contains(where: { $0.name == name })` → `targetsByName[name] != nil` (O(1))
  >
  > This is safe because SPM guarantees unique target names within a package, so `Dictionary(uniqueKeysWithValues:)` will not crash.
  >
  > Overall complexity improves from O(n*m) to O(n+m). No behavioral changes — the resulting `targetToProducts` mapping is identical.

  ### How to test locally

  1. Run existing `PackageInfoMapper` tests to verify no regressions
  2. Add a large SPM dependency with many targets (e.g., `firebase-ios-sdk`) to a Tuist project
  3. Run `tuist generate` and verify the generated project is identical to before this change

### DIFF
--- a/cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift
+++ b/cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift
@@ -302,6 +302,7 @@ public struct PackageInfoMapper: PackageInfoMapping {
             uniquingKeysWith: { userDefined, _ in userDefined }
         )
 
+        let targetsByName = Dictionary(uniqueKeysWithValues: packageInfo.targets.map { ($0.name, $0) })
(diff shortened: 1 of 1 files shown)


pepicrft:
Thank you @inju2403 🙏🏼 
@all-contributors add @inju2403 for performance

PR #10032, opened 2026-03-25
feat(server,cli,gradle): link build runs to shard plans

fortmarek:
## Summary

- Adds `build_run_id` and `gradle_build_id` columns to the `shard_plans` ClickHouse table
- CLI passes the `buildRunId` from `RunMetadataStorage` when creating a shard plan (both `tuist xcodebuild build --shard-*` and `tuist test --shard-*`)
- Gradle plugin passes `gradleBuildId` from the `tuistBuildInsights` shared service when creating a shard plan
- Test run detail page resolves the associated build through the shard plan for sharded runs, so the "Build: {scheme}" link appears correctly
- Build run detail page can find sharded test runs via shard plan lookup (fallback in `get_latest_test_by_build_run_id` and `get_latest_test_by_gradle_build_id`)

### Context

When shards were introduced, the connection between build runs and test runs was broken. With sharding, the build happens once (producing `.xctestproducts` / test products), then N shard machines only run tests — they don't build. The test run's `build_run_id` was either set from the first-reporting shard's test-only activity log or left nil, losing the link to the actual build.

The fix stores the build ID on the `shard_plans` table since the build machine knows its build ID at the time it creates the shard plan. This correctly models the 1:1 relationship: one build → one shard plan → N shard runs → one test run.

## Test plan

- [x] Server compiles cleanly (`mix compile --warnings-as-errors`)
- [x] Server tests pass (`mix test test/tuist/shards_test.exs`, `test/tuist/tests_test.exs`, `test/tuist_web/controllers/api/shards_controller_test.exs`, `test/tuist_web/live/test_run_live_test.exs`)
- [x] CLI builds (`xcodebuild build -scheme Tuist-Workspace`)
- [x] Gradle plugin compiles and tests pass (`./gradlew compileKotlin`, `./gradlew test --tests "*TuistTestSharding*"`)
- [ ] Verify on staging: sharded test run shows associated build link
- [ ] Verify on staging: build run page shows link to sharded test run


cli/Sources/TuistServer/OpenAPI/server.yml
```diff
@@ -18614,6 +18643,6 @@ security:
   - authorization: []
   - cookie: []
 servers:
-  - url: http://localhost:8080
+  - url: http://localhost:8440
```
pepicrft: I remember fixing this in a PR. Also, there's other order non-deterministic likely coming from the underlying tool.

PR #10034, opened 2026-03-25
perf(cli): use Set for target name lookups in SchemeLinter

inju2403:
### Description

  Optimize target name lookups in `SchemeLinter` by using `Set` instead of `Array` for `contains()` checks.

  **Before:**
  ```swift
  let targetNames = targets.map(\.name) // Array
  !targetNames.contains(target.name)    // O(n) per call

  After:
  let targetNames = Set(targets.map(\.name)) // Set
  !targetNames.contains(target.name)          // O(1) per call

  Both lintExpandVariableTarget and lintCodeCoverageTargets iterate over schemes and check target name membership inside loops. Using a Set reduces each contains() call from O(n) to O(1).

  How to test locally

  ```

### DIFF
--- a/cli/Sources/TuistGenerator/Linter/SchemeLinter.swift
+++ b/cli/Sources/TuistGenerator/Linter/SchemeLinter.swift
@@ -92,7 +92,7 @@ struct SchemeLinter: SchemeLinting {
     }
 
     private func lintExpandVariableTarget(schemes: [Scheme], targets: [Target]) -> [LintingIssue] {
-        let targetNames = targets.map(\.name)
+        let targetNames = Set(targets.map(\.name))
         var issues: [LintingIssue] = []
 
         for scheme in schemes {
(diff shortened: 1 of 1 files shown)


inju2403:
Hi @pepicrft @fortmarek — adding measurements to help unblock this. (Sorry for the late follow-up.)

Since `SchemeLinter` runs once per generate and the lookup scales with `targets.count × schemes.count`, the `[String].contains` → `Set<String>.contains` change is meaningful on larger projects.

**Setup**: Isolated Swift -O microbenchmark on macOS (M-series). Builds an array of target names, then performs N random `contains` lookups. Measures BEFORE (`[String]`) vs AFTER (`Set<String>`).

  | targets | lookups | BEFORE (ms) | AFTER (ms) | speedup |
  | 100 | 1,000 | 0.229 | 0.049 | 4.7× |
  | 500 | 5,000 | 4.438 | 0.199 | 22.3× |
  | 1,000 | 10,000 | 15.123 | 0.390 | 38.8× |
  | 2,000 | 20,000 | 66.787 | 0.859 | 77.7× |
  | 5,000 | 50,000 | 385.728 | 2.285 | 168.8× |

Expected from O(n·m) → O(n+m). Small projects see little impact; large monorepos save ~100 ms+ per generate.

Happy to add real-world `hyperfine` numbers if helpful — let me know.

PR #10035, opened 2026-03-25
perf(cli): precompute Xcode version key set in DefaultSettingsProvider

inju2403:
### Description

  Optimize `createFilter` in `DefaultSettingsProvider` by precomputing the set of higher Xcode version keys once, instead of recomputing it on every closure invocation.

  **Before:**
  ```swift
  return { key, _ in
      !newXcodeKeys
          .filter { $0.key > xcodeVersion }
          .values.flatMap { $0 }.contains(key) &&
          !excludedKeys.contains(key)
  }

  Every time the returned closure is called, it filters the dictionary, flattens the values, and performs a linear search — all O(n) per call.

  After:
  let higherVersionKeys = Set(newXcodeKeys.filter { $0.key > xcodeVersion }.values.flatMap { $0 })
  return { key, _ in
      !higherVersionKeys.contains(key) &&
          !excludedKeys.contains(key)
  }

  The Set is built once before the closure is returned. Each subsequent contains() call is O(1).

  How to test locally

  ```

### DIFF
--- a/cli/Sources/TuistGenerator/Settings/DefaultSettingsProvider.swift
+++ b/cli/Sources/TuistGenerator/Settings/DefaultSettingsProvider.swift
@@ -237,11 +237,9 @@ public struct DefaultSettingsProvider: DefaultSettingsProviding {
             return { key, _ in essentialKeys.contains(key) && !excludedKeys.contains(key) }
         case let .recommended(excludedKeys):
             let xcodeVersion = try await XcodeController.current.selectedVersion()
+            let higherVersionKeys = Set(newXcodeKeys.filter { $0.key > xcodeVersion }.values.flatMap { $0 })
             return { key, _ in
-                // Filter keys that are from higher Xcode version than current (otherwise return true)
-                !newXcodeKeys
-                    .filter { $0.key > xcodeVersion }
-                    .values.flatMap { $0 }.contains(key) &&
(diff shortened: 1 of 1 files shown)


inju2403:
Hi @pepicrft @fortmarek — measurements for context. (Sorry for the late follow-up.)

The closure returned by `recommended` is invoked once per settings key per target, so even a small per-call cost compounds. Hoisting `Set(... filter ... flatMap ...)` out of the closure removes that per-call work.

**Setup**: Isolated Swift -O microbenchmark. Builds `[Int: Set<String>]` (proxy for `[Version: Set<String>]`) and invokes the predicate N times.

  | versions | keys/v | calls | BEFORE (ms) | AFTER (ms) | speedup |
  | --- | --- | --- | --- | --- | --- |
  | 10 | 5 | 100 | 0.119 | 0.008 | 14.1× |
  | 10 | 10 | 1,000 | 2.000 | 0.061 | 33.0× |
  | 10 | 20 | 5,000 | 9.806 | 0.308 | 31.9× |
  | 15 | 30 | 10,000 | 31.795 | 0.457 | 69.6× |
  | 20 | 50 | 50,000 | 264.177 | 2.184 | 121.0× |

Same observable behavior; just one filter+flatMap+Set construction up-front instead of per call.

Let me know if a real-world `tuist generate` measurement on a fixture would help.

PR #10036, opened 2026-03-25
perf(cli): use Set for header path lookups in TargetContentHasher

inju2403:
### Description

  Optimize header path lookups in `TargetContentHasher` by using `Set` instead of `Array` for `publicHeaders` and `privateHeaders`.

  **Before:**
  ```swift
  let publicHeaders = buildableFolder.exceptions.flatMap(\.publicHeaders)   // Array
  let privateHeaders = buildableFolder.exceptions.flatMap(\.privateHeaders) // Array

  return try await buildableFiles.concurrentMap { buildableFile in
      if publicHeaders.contains(buildableFile.path) { ... }  // O(n) per file
      if privateHeaders.contains(buildableFile.path) { ... }  // O(n) per file
  }

  After:
  let publicHeaders = Set(buildableFolder.exceptions.flatMap(\.publicHeaders))   // Set
  let privateHeaders = Set(buildableFolder.exceptions.flatMap(\.privateHeaders)) // Set

  return try await buildableFiles.concurrentMap { buildableFile in
      if publicHeaders.contains(buildableFile.path) { ... }  // O(1) per file
      if privateHeaders.contains(buildableFile.path) { ... }  // O(1) per file
  }

  The contains() calls run inside concurrentMap over every buildable file, making this a hot path during cache hashing. Converting to Set reduces each lookup from O(n) to O(1).

  How to test locally

  ```

### DIFF
--- a/cli/Sources/TuistHasher/TargetContentHasher.swift
+++ b/cli/Sources/TuistHasher/TargetContentHasher.swift
@@ -223,8 +223,8 @@ public struct TargetContentHasher: TargetContentHashing {
             .map { ($0, $0.resolvedFiles.sorted(by: { $0.path < $1.path })) }
             .concurrentFlatMap {
                 (buildableFolder: BuildableFolder, buildableFiles: [BuildableFolderFile]) in
-                let publicHeaders = buildableFolder.exceptions.flatMap(\.publicHeaders)
-                let privateHeaders = buildableFolder.exceptions.flatMap(\.privateHeaders)
+                let publicHeaders = Set(buildableFolder.exceptions.flatMap(\.publicHeaders))
+                let privateHeaders = Set(buildableFolder.exceptions.flatMap(\.privateHeaders))
 
                 return try await buildableFiles.concurrentMap { buildableFile in
                     let fileHash = try await contentHasher.hash(path: buildableFile.path)


inju2403:
Hi @pepicrft @fortmarek — adding measurements. (Sorry for the late follow-up.)

`TargetContentHasher` calls `publicHeaders.contains(...)` and `privateHeaders.contains(...)` once per `BuildableFile`, so the lookup scales with `headers × files`.

**Setup**: Isolated Swift -O microbenchmark with string paths matching typical lengths. BEFORE = two `[String].contains`, AFTER = two `Set<String>.contains` (set construction included in the measurement).

  | headers | files | BEFORE (ms) | AFTER (ms) | speedup |
  | 10 | 120 | 0.022 | 0.024 | 0.9× |
  | 50 | 600 | 0.509 | 0.108 | 4.7× |
  | 100 | 1,200 | 1.932 | 0.178 | 10.9× |
  | 200 | 2,400 | 7.024 | 0.343 | 20.5× |
  | 500 | 6,000 | 42.384 | 0.952 | 44.5× |

One honest note: at very small scale (≤10 headers) the Set-construction overhead makes it a slight regression. The real-world buildable folders in most projects sit well above that threshold, but happy to discuss if you'd prefer a guard for tiny cases.

Let me know if a real-world `tuist generate` run on a fixture would help.

PR #10037, opened 2026-03-25
feat(docs): add CLI command reference documentation

pepicrft:
## Summary

- Adds CLI command reference pages to the Elixir/Phoenix docs, fetching `tuist.spec.json` from the latest GitHub release and caching it in a GenServer (refreshes hourly)
- Renders per-command pages with arguments, flags, environment variables, and usage examples
- Adds a "CLI" tab to the docs top navigation alongside Guides, References, and Resources
- Loads static CLI docs (debugging, directories, shell-completions) that were previously excluded
- Preserves sidebar scroll position across LiveView navigations
- Updates nav tab selection client-side to stay in sync during live navigation


server/lib/tuist/docs/cli.ex
```diff
@@ -0,0 +1,112 @@
+defmodule Tuist.Docs.CLI do
```
pepicrft, line 1: I'm adding tests to this one.

server/test/tuist/docs/cli_test.exs
```diff
@@ -0,0 +1,177 @@
+defmodule Tuist.Docs.CLITest do
+  use ExUnit.Case, async: false
```
pepicrft: I'm making some changes so this one is `async: true`.

pepicrft:
All comments addressed @fortmarek

PR #10038, opened 2026-03-25
perf(cli): use min() instead of sorted().first in GraphLoader

inju2403:
### Description

  Replace `.sorted().first` with `.min()` in `GraphLoader` for finding the minimum platform.

  **Before:**
  ```swift
  platforms.sorted().first  // O(n log n) — sorts entire array, then takes first element

  After:
  platforms.min()  // O(n) — single pass to find minimum

  Both .sdk and .xctest cases only need the smallest platform value, not a fully sorted array. .min() achieves the same result in a single linear pass.

  How to test locally

  This is a pure refactor with no behavioral change — .sorted().first and .min() return the same value. The change only affects algorithmic complexity, not logic.
  ```

### DIFF
--- a/cli/Sources/TuistCore/Graph/GraphLoader.swift
+++ b/cli/Sources/TuistCore/Graph/GraphLoader.swift
@@ -175,7 +175,7 @@ public struct GraphLoader: GraphLoading {
             )
 
         case let .sdk(name, status, _):
-            return try platforms.sorted().first.map { platform in
+            return try platforms.min().map { platform in
                 try loadSDK(
                     name: name,
                     platform: platform,
@@ -197,7 +197,7 @@ public struct GraphLoader: GraphLoading {
             }
 
         case .xctest:
-            return try platforms.sorted().first.map { platform in
+            return try platforms.min().map { platform in
                 try loadXCTestSDK(platform: platform)
             }
         }


inju2403:
Hi @pepicrft @fortmarek — measurements + a quick note on scope. (Sorry for the late follow-up.)

I want to be upfront: `platforms` here is a `Set<Platform>` with realistic cardinality of 1–5, so absolute wall-clock savings are in the microsecond range — this is more about algorithmic correctness (O(n) vs O(n log n)) than measurable speedup in real builds.

**Setup**: Isolated Swift -O microbenchmark sweeping `Set<Int>` sizes.

  | N | BEFORE (ms/call) | AFTER (ms/call) | speedup | note |
  | 5 | 0.00020 | 0.00001 | 30.7× | real-world platform-set scale |
  | 50 | 0.00076 | 0.00004 | 18.9× | |
  | 500 | 0.01380 | 0.00054 | 25.6× | |
  | 5,000 | 0.25680 | 0.00500 | 51.3× | |
  | 50,000 | 3.14590 | 0.05680 | 55.4× | |

So the change is essentially a correctness/idiom cleanup — `min()` is what we want here semantically. Totally understand if this isn't worth the review cost — happy to close if you'd prefer to keep the bar at measurable wins.

PR #10039, opened 2026-03-25
perf(cli): replace string concatenation loop with joined(separator:) …

inju2403:
### Description

  Replace string concatenation in a loop with `map().joined(separator:)` in `ContentHasher.hash(_:)`.

  **Before:**
  ```swift
  let sortedKeys = dictionary.keys.sorted { $0 < $1 }
  var dictString = ""
  for (counter, key) in sortedKeys.enumerated() {
      let value: String = dictionary[key]!
      dictString += "\(key):\(value)"
      let isLastKey = counter == (sortedKeys.count - 1)
      if !isLastKey {
          dictString += "-"
      }
  }
  return try hash(dictString)

  Each += allocates a new String, copying all previous characters. With N keys this results in O(n²) total character copies.

  After:
  let dictString = dictionary.keys.sorted()
      .map { "\($0):\(dictionary[$0]!)" }
      .joined(separator: "-")
  return try hash(dictString)

  joined(separator:) pre-calculates the total length and allocates once, reducing the complexity to O(n). This method is in ContentHasher, a hot path called for every target and file during cache hashing.

  How to test locally

  This is a pure refactor with no behavioral change — the output string format (key1:value1-key2:value2-...) is identical. The change only affects allocation performance, not logic.
  ```

### DIFF
--- a/cli/Sources/TuistCore/ContentHashing/ContentHasher.swift
+++ b/cli/Sources/TuistCore/ContentHashing/ContentHasher.swift
@@ -41,16 +41,9 @@ public struct ContentHasher: ContentHashing {
     }
 
     public func hash(_ dictionary: [String: String]) throws -> String {
-        let sortedKeys = dictionary.keys.sorted { $0 < $1 }
-        var dictString = ""
-        for (counter, key) in sortedKeys.enumerated() {
-            let value: String = dictionary[key]!
-            dictString += "\(key):\(value)"
-            let isLastKey = counter == (sortedKeys.count - 1)
-            if !isLastKey {
-                dictString += "-"
-            }
-        }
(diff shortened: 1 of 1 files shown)


inju2403:
After running an isolated microbenchmark, I see no meaningful performance difference between the `+=` loop and `joined(separator:)` here — Swift's string `+=` is already well-optimized for this pattern (results were ~1.0× across N=10 to 5,000 entries).

Closing this since I can't honestly claim a perf win. If the cleaner `map(...).joined(...)` form is desirable on readability grounds alone, I'd be happy to re-open with that framing — but otherwise it doesn't belong in the perf series.

Thanks!

PR #10045, opened 2026-03-26
fix(cache): ignore unexpected SQLiteBuffer messages

cschmatzler:
## Summary
- ignore unexpected `handle_info/2` messages in `Cache.SQLiteBuffer` instead of crashing
- log the unexpected message so we can still investigate its source later
- add a regression test for the `{[:alias | ref], :dropped}` message shape seen in production

Fixes CACHE-2T, CACHE-2S, and some random 502s that popped up.

## Testing
- mise run format --check
- mise run credo
- mise run test

### DIFF
--- a/cache/test/cache/sqlite_buffer_test.exs
+++ b/cache/test/cache/sqlite_buffer_test.exs
@@ -3,6 +3,7 @@ defmodule Cache.SQLiteBufferTest do
   use Mimic
 
   import Ecto.Query
+  import ExUnit.CaptureLog
 
   alias Cache.CacheArtifact
   alias Cache.CacheArtifactsBuffer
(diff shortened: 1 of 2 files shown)


fortmarek (approved):
Once we figure out the source behind the unexpected message, we might want to revert this, so _new_ unexpected messages are still thrown.

cache/test/cache/sqlite_buffer_test.exs
```diff
@@ -227,6 +228,31 @@ defmodule Cache.SQLiteBufferTest do
     assert record.json_payload == payload
   end
 
+  test "unexpected info messages are logged and ignored" do
+    key = "keyvalue:unexpected-message:account:project"
+    payload = Jason.encode!(%{entries: [%{"value" => "unexpected"}]})
```
pepicrft, line 233: What about using JSON? It should be available in the version that we are using of Elixir

cschmatzler: We use Jason as our [Phoenix JSON library](https://github.com/tuist/tuist/blob/ac821a08b5de4013e6cbda642cb672f9c6969c2a/cache/config/config.exs#L114), so I think the rest of the codebase should match that.

fortmarek: @pepicrft probably another credo rule that we can enforce across our whole monorepo since that's generally our preference. We can keep using `Jason` for now until we migrate the usage in the whole `cache` codebase.

cschmatzler: Yeah I'm not against migrating everything to JSON, sorry. Would just see it outside of the scope of this PR since right now this is consistent and would touch other things.

PR #10052, opened 2026-03-26
refactor(cache): Jason -> JSON

cschmatzler:
title.


cache/mix.exs
cschmatzler, line 1: fyi `mix.lock` is unchanged since it's still a transitive dependency of other stuff

PR #10055, opened 2026-03-26
feat(helm): add self-hosted Helm chart with observability

pepicrft:
> [!NOTE]
> **For reviewers**
> It'd be great if you could also run this locally and run an e2e workflow against the server, for example caching and usage of binaries. You can run `mise run helm:up --license "YOUR_KEY" --observability` (get the license from keygen). The cluster will also use the cache node running in it. Note there's also an observability component in case we want to observe traces, metrics, and logs of those running services.

## Summary

- Add an umbrella Helm chart (`infra/helm/tuist/`) for deploying Tuist server and cache on Kubernetes
- Infrastructure dependencies (PostgreSQL, ClickHouse, object storage) default to embedded and can be individually switched to external mode
- Optional observability stack (Grafana, Prometheus, Loki, Tempo, OTel Collector) with working Logs, Traces, and Metrics drilldowns
- Server automatically wires the cache endpoint via `TUIST_CACHE_ENDPOINTS` env var
- Add `TUIST_CACHE_ENDPOINTS` env var support to `Tuist.Environment`
- Add git-cliff config and release workflow job for publishing the chart as an OCI artifact to `ghcr.io/tuist/charts`
- Add Kubernetes/Helm section to self-hosting documentation
- Add mise tasks (`helm:build`, `helm:up`, `helm:down`, `helm:status`) for local development with usage annotations
- Images default to `ghcr.io/tuist/tuist` and `ghcr.io/tuist/cache` with tag falling back to `Chart.appVersion`

## Test plan

- [x] `helm lint` and `helm template` pass
- [x] Deploy to local kind cluster with embedded dependencies
- [x] Server starts and serves requests
- [x] Cache node connects to server and handles artifact uploads
- [x] Observability stack: metrics, logs, and traces visible in Grafana
- [x] Grafana drilldowns work for Logs, Traces, and Metrics
- [x] `mise run helm:up` builds images, deploys, and port-forwards automatically
- [x] `mise run helm:down` tears down cluster and kills port-forwards
- [x] License tests pass (`mix test test/tuist/license_test.exs`)

### DIFF
--- a/server/priv/docs/en/guides/server/self-host/install.md
+++ b/server/priv/docs/en/guides/server/self-host/install.md
@@ -375,12 +375,6 @@ Tuist provides an official Helm chart for deploying on Kubernetes. The chart pac
 ### Installing the chart {#installing-the-chart}
 
 ```bash
-helm install tuist oci://ghcr.io/tuist/charts/tuist
-```
-
-You will need to provide your license key:
-
-```bash
 helm install tuist oci://ghcr.io/tuist/charts/tuist \
   --set server.license.key="YOUR_LICENSE_KEY"
 ```
@@ -388,14 +382,17 @@ helm install tuist oci://ghcr.io/tuist/charts/tuist \
 To pin a specific chart version:
 
 ```bash
-helm install tuist oci://ghcr.io/tuist/charts/tuist --version 0.1.0
+helm install tuist oci://ghcr.io/tuist/charts/tuist \
+  --set server.license.key="YOUR_LICENSE_KEY" \
+  --version 0.1.0
 ```
 
 ### Infrastructure dependencies {#helm-infrastructure-dependencies}
 
-The chart manages three infrastructure dependencies: `postgresql`, `clickhouse`, and `objectStorage`. Each defaults to **embedded** mode, meaning the chart deploys them inside the cluster. To point at your own external instances instead, set the dependency's `mode` to `external` and fill in the connection details. For example, to use an external PostgreSQL database:
+The chart manages three infrastructure dependencies: `postgresql`, `clickhouse`, and `objectStorage`. Each defaults to **embedded** mode, meaning the chart deploys them inside the cluster. To point at your own external instances instead, set the dependency's `mode` to `external` and fill in the connection details in your `values.yaml`. For example, to use an external PostgreSQL database:
 
 ```yaml
+# values.yaml
 postgresql:
   mode: external
   external:
@@ -410,9 +407,10 @@ The same pattern applies to `clickhouse` and `objectStorage`. See the `external`
 
 ### Observability {#helm-observability}
 
-The chart includes an optional observability stack (OpenTelemetry Collector, Prometheus, Grafana, Loki, and Tempo). It is **disabled by default**. To enable it:
+The chart includes an optional observability stack (OpenTelemetry Collector, Prometheus, Grafana, Loki, and Tempo). It is **disabled by default**. To enable it in your `values.yaml`:
 
 ```yaml
+# values.yaml
 observability:
   enabled: true
 ```
@@ -420,6 +418,7 @@ observability:
 When enabled, Grafana is available with Logs, Traces, and Metrics drilldowns pre-configured. For external observability, keep this disabled and configure the endpoints via `server.extraEnv`:
 
 ```yaml
+# values.yaml
 server:
   extraEnv:
     - name: TUIST_OTEL_EXPORTER_OTLP_ENDPOINT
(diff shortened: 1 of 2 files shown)


infra/helm/tuist/templates/observability.yaml
```diff
@@ -0,0 +1,367 @@
+{{- if .Values.observability.enabled }}
+apiVersion: v1
```
pepicrft, line 2: I included this one as an opt-in, since it can be useful if we end up using this one for ephemeral environments to give our agents access to observability data.

infra/helm/tuist/templates/app-secrets.yaml
```diff
@@ -0,0 +1,20 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "tuist.componentName" (dict "root" . "component" "app-secrets") }}
+  labels:
+    {{- include "tuist.labels" . | nindent 4 }}
+type: Opaque
+stringData:
+  server-secret-key-base: {{ randAlphaNum 64 | quote }}
```
cschmatzler: Generating `server-secret-key-base`, `cache-secret-key-base`, and `cache-guardian-secret-key` with `randAlphaNum` means every `helm upgrade` will emit brand-new secrets, which rotates credentials on every release and invalidates state that depends on those keys. In practice this will log users out and can break cache authentication after a routine upgrade. These values need to come from stable chart inputs (or be preserved from the existing Secret) so upgrades do not rewrite them each time.

pepicrft: I updated it to use Helm's `lookup`. On first hit, values are generated, and then upgrades reuse the values. They can also set the values manually.

server/priv/docs/en/guides/server/self-host/install.md
```diff
@@ -368,6 +368,93 @@ docker compose down -v
...
+```bash
+helm install tuist oci://ghcr.io/tuist/charts/tuist
+```
+
+You will need to provide your license key:
+
+```bash
+helm install tuist oci://ghcr.io/tuist/charts/tuist \
+  --set server.license.key="YOUR_LICENSE_KEY"
+```
```
fortmarek: I'd combine these two statements together instead of having an installation step without the license key when it's required

server/priv/docs/en/guides/server/self-host/install.md
```diff
@@ -368,6 +368,93 @@ docker compose down -v
...
+
+To pin a specific chart version:
+
+```bash
+helm install tuist oci://ghcr.io/tuist/charts/tuist --version 0.1.0
```
fortmarek: this should also include the license key, no? otherwise, it won't work

server/priv/docs/en/guides/server/self-host/install.md
```diff
@@ -368,6 +368,93 @@ docker compose down -v
...
+```
+
+### Infrastructure dependencies {#helm-infrastructure-dependencies}
+
+The chart manages three infrastructure dependencies: `postgresql`, `clickhouse`, and `objectStorage`. Each defaults to **embedded** mode, meaning the chart deploys them inside the cluster. To point at your own external instances instead, set the dependency's `mode` to `external` and fill in the connection details. For example, to use an external PostgreSQL database:
+
+```yaml
```
fortmarek, line 394: this doesn't say what yaml this refers to

server/priv/docs/en/guides/server/self-host/install.md
```diff
@@ -368,6 +368,93 @@ docker compose down -v
...
+postgresql:
```
fortmarek, line 396: should we recommend to host `postgres` and clickhouse externally? otherwise, the data won't survive redeployments, or will it?

server/priv/docs/en/guides/server/self-host/install.md
```diff
@@ -368,6 +368,93 @@ docker compose down -v
...
+
+> [!TIP]
+> **Testing locally with kind**
+>
+> You can spin up the full stack on your machine using [kind](https://kind.sigs.k8s.io/) and the included mise tasks. This builds the Docker images locally, creates a kind cluster, installs the chart, and port-forwards all services automatically:
+>
+> ```bash
+> # Deploy everything (builds images from source)
+> mise run helm:up --license "YOUR_LICENSE_KEY"
+>
```
fortmarek: this is for when wanting to run things from source, right? Then I think this should be in the debugging or contributors section, it doesn't feel like it belongs to self-hosting

server/priv/docs/en/guides/server/self-host/install.md
pepicrft: That's right. I'll reflect that in the example.

server/priv/docs/en/guides/server/self-host/install.md
pepicrft: It should be `values.yaml`. I'll make it obvious in the example.

server/priv/docs/en/guides/server/self-host/install.md
pepicrft: It should if configured to be persisted across upgrades. Let me see if the configuration is right for that.

pepicrft: Yes, I confirmed. All components with persistence are configured to use `StatefulSets` with `volumeClaimTemplates` which means the data survives pod restarts, upgrades, and redeployments.

server/priv/docs/en/guides/server/self-host/install.md
pepicrft: That's right. I moved them to the contributors' section.

Issue #5925, opened 2024-02-13
TargetScript output files are ignored if the files don't exist at generate time


dcramps:
### What happened?

If a `TargetScript` specifies an output file it is not added to the script's output files in the generated project unless it already exists at the time `tuist generate` is run. Similarly, a file that does not exist at generate time is not added to the `Compile Sources` phase.

My use case here is that I have a target script that runs on build, specifying its output file. The file is added to the `Compile Sources` phase, and all is well: the script generates the file, so it exists by the time sources are compiled. With Tuist, this is not possible. In my own project I have hacked around it by using `FileManager` to create the file prior to returning the script, but that is super fragile and can break due to project caching.

A `TargetScript` requiring output files to exist to be included is definitely a bug, and in Tuist 4 is even more of a bug as files that exist are not included in output files for some reason.

For the `Compile Sources` part, I believe Tuist should, or should have the option to, include files in `Compile Sources` that do not exist, as it is perfectly reasonable for a script to generate those missing files.

Sample projects:
[tuist3.zip](https://github.com/tuist/tuist/files/14270300/tuist3.zip)
[tuist4.zip](https://github.com/tuist/tuist/files/14270301/tuist4.zip)

## Tuist 3
1. Run `tuist generate` on the attached sample `tuist3.zip`
2. Open the `TuistScriptOutputFiles` target, and expand the `Touch File.swift for TuistScriptOutputFiles` phase under Build Phases
3. Note that `Output Files` is empty
4. Expand `Compile Sources` under Build Phases
5. Note that `File.swift` is not included
6. Run `tuist generate` again
7. Note that `Output Files` is no longer empty
8. Note that `File.swift` is now included in `Compile Sources` phase

## Tuist 4
1. Run `tuist generate` on the attached sample `tuist4.zip`
2. Open the `TuistScriptOutputFiles` target, and expand the `Touch File.swift for TuistScriptOutputFiles` phase under Build Phases
3. Note that `Output Files` is empty
4. Expand `Compile Sources` under Build Phases
5. Note that `File.swift` is not included
6. Run `tuist generate` again
7. Note that `Output Files` is **still empty**
8. Note that `File.swift` is now included in `Compile Sources` phase

### Error log

No errors

### macOS version

14.3.1

### Tuist version

3.42.3

### Xcode version

15.2

abhishek818:
@fortmarek can i get this assigned?

abhishek818:
[...]

PR #10058, opened 2026-03-26
fix: include non-existent output paths in TargetScript build phases /claim #5925

Nithinfgs:
TargetScript.outputPaths are currently resolved using the same helper (absolutePaths) that is used for input paths. That helper performs filesystem glob expansion and therefore only returns files that already exist on disk.

However, output files declared in a build-phase script are usually generated during the build, meaning they often do not exist at tuist generate time. Because of this, the glob step silently drops non-existent output paths and they never appear in the generated Xcode build phase.

Solution

Introduce a dedicated helper absoluteOutputPaths() that resolves output paths without glob expansion.

Unlike input paths, output paths are resolved directly through generatorPaths.resolve(path:), ensuring that:

output files are included even if they do not yet exist
paths containing build variables (e.g. $(DERIVED_FILE_DIR)) are preserved
existing behaviour for input paths remains unchanged
Tests

Added regression tests in TargetActionManifestMapperTests to verify:

non-existent output paths are included
existing output paths continue to work
paths containing build variables are preserved
input path behaviour (glob-based filtering) remains unchanged
mixed existing/non-existing output paths are handled correctly
Result

All declared TargetScript.outputPaths are now correctly propagated to the generated Xcode build phase regardless of whether the files exist at generation time. 

/claim #5925

### DIFF
--- a/cli/Sources/TuistLoader/Models+ManifestMappers/TargetAction+ManifestMapper.swift
+++ b/cli/Sources/TuistLoader/Models+ManifestMappers/TargetAction+ManifestMapper.swift
@@ -1,4 +1,3 @@
-import FileSystem
 import Foundation
 import Path
 import ProjectDescription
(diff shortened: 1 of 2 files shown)


Nithinfgs:
Opened PR #10058 implementing this fix.

PR #10061, opened 2026-03-26
fix(server): render changelog entries with mdex

pepicrft:
## Summary
- compile changelog entry bodies through `MDExConverter` instead of returning raw markdown
- bound the frontmatter/body split to avoid truncating entries that contain `---` in the body
- add focused parser coverage for markdown rendering and thematic breaks

## Verification
- `mix run -e ...` exercised `Tuist.Marketing.Changelog.EntryParser.parse/2` and confirmed the generated HTML contains `<strong>`, `<hr>`, and the expected `` output
- `mix test test/tuist/marketing/changelog_entry_parser_test.exs` is currently blocked locally because the test DB migrations fail with `ERROR 42P01 (undefined_table) relation "users" does not exist`

### DIFF
--- a/server/assets/shared/js/hooks/code-copy.js
+++ b/server/assets/shared/js/hooks/code-copy.js
@@ -35,6 +35,9 @@ const CodeCopy = {
   mounted() {
     setupCodeCopy(this.el);
   },
+  updated() {
+    setupCodeCopy(this.el);
+  },
 };
 
 export { CodeCopy, setupCodeCopy, flashCopyCheck };


server/config/dev.exs
```diff
@@ -109,7 +109,7 @@ config :esbuild,
     args:
       ~w(apidocs.js --bundle --target=es2017 --outfile=../../priv/static/apidocs/assets/bundle.js --external:/fonts/* --external:/images/*),
     cd: Path.expand("../assets/apidocs", __DIR__),
-    env: %{"NODE_PATH" => deps_path}
+    env: %{"NODE_PATH" => "#{deps_path}:#{node_modules_path}"}
```
pepicrft, line 112: It failed to resolve the dependencies locally.

PR #10065, opened 2026-03-26
fix(cli): mark generated bundle accessors as nonisolated

### LINKED ISSUE #9759: Bundle generated accessors can be non isolated
### Why is this needed?

Autogenerated code to access frameworks resource bundles has default actor isolation.

In case your project default actor isolation is other than "nonisolated", this can introduce unnecessary complexity. These are static immutable variables that could be defined as nonisolated.

This improvement aims to introduce the nonisolated keywords to the generated code.

### Steps to address the need

[This repository](https://github.com/jrcmramos-bumble/Tuist-Bundle-Generated-Accessors) contains a project where we can see this behaviour.

1) Create a project with an app target and framwork dependency.
2) Define "Main Actor" as the default isolation.
3) Access `Foundation.Bundle.module` from a nonisolated context.
4) Generate the code and compile

DimaMishchenko:
Resolves #9759

Generated bundle accessors should remain callable from nonisolated code even when a module opts into Swift 6 default `MainActor` isolation.
Updated the generated accessors to emit `nonisolated`.

### How to test locally
* with local `tuist` version run `tuist generate` and inspect generated `TuistBundle+*.swift` files
    * good examples inside repo are `examples/xcode/generated_ios_app_with_framework_and_resources` and `examples/xcode/generated_ios_static_library_with_string_resources`  
    * #9759 also provides a nice reproducible example for testing

### DIFF
--- a/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
+++ b/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
@@ -299,7 +299,7 @@ public struct ResourcesProjectMapper: ProjectMapping { // swiftlint:disable:this
         extern "C" {
         #endif
 
-        NSBundle* \(targetName)_SWIFTPM_MODULE_BUNDLE(void);
+        NSBundle* \(targetName)_SWIFTPM_MODULE_BUNDLE(void) NS_SWIFT_NONISOLATED;
 
         #define SWIFTPM_MODULE_BUNDLE \(targetName)_SWIFTPM_MODULE_BUNDLE()
 
@@ -370,7 +370,7 @@ public struct ResourcesProjectMapper: ProjectMapping { // swiftlint:disable:this
         // MARK: - Objective-C Bundle Accessor
         @objc
         public final class \(target.productName.toValidSwiftIdentifier())Resources: NSObject {
-        @objc public class var bundle: Bundle {
+        @objc public nonisolated class var bundle: Bundle {
             return .module
         }
         }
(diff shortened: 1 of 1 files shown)


DimaMishchenko:
I didn’t add tests for this yet.
The current `ResourcesProjectMapper` tests are mostly higher-level scenarios, and don't test for specific generated code. 
If it makes sense, I can add targeted checks similar to this one:
```swift
#expect(file.contents?.contains("nonisolated static let module = Bundle(for: BundleFinder.self)".data(using: .utf8)!) == true)
```

cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
```diff
@@ -299,7 +299,7 @@ public struct ResourcesProjectMapper: ProjectMapping { // swiftlint:disable:this
         extern "C" {
         #endif
 
-        NSBundle* \(targetName)_SWIFTPM_MODULE_BUNDLE(void);
+        NSBundle* \(targetName)_SWIFTPM_MODULE_BUNDLE(void) NS_SWIFT_NONISOLATED;
```
pepicrft, line 302: Can these changes break current users in any way?

pepicrft:
I did a pass to this using `codex` to confirm there are no risks of introducing breaking changes, and the conclusion is that it should not introduce breaking changes. I also validated it with some projects locally.
@fortmarek it'd be great if you could do another one to have an extra confidence.

cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
DimaMishchenko: Hey @pepicrft, it shouldn't.

Obj-C behavior will stay as it is, but `NS_SWIFT_NONISOLATED` will mark generated Swift interfaces as `nonisolated`.
To be honest, I was unable to figure out a Tuist fixture that would generate Obj-C accessors, so I just tested a small snippet with `NS_SWIFT_NONISOLATED` separately to confirm it does the thing.

And the Swift changes are non-breaking; they're mainly for projects that are now adopting approachable concurrency and opting into `MainActor` isolation by default.

DimaMishchenko:
Hey @fortmarek, when you get a chance, I’d appreciate if you could review this one 🙂

ebubedimobi:
Goodday, any idea when we can expect this fix? really need this

PR #10073, opened 2026-03-27
feat(scripts): add registry management scripts

cschmatzler:
Adds two scripts:
```
󰘧 · scripts/purge-registry-package.sh zendesk/sdk_logger_ios
[purge-registry-package.sh] Checking required commands
[purge-registry-package.sh] Checking rclone remote configuration
[purge-registry-package.sh] Found rclone remote "tigris"
[purge-registry-package.sh] Verifying read access to tigris:tuist-registry
[purge-registry-package.sh] Package input: zendesk/sdk_logger_ios
[purge-registry-package.sh] Normalized storage package: zendesk/sdk_logger_ios
[purge-registry-package.sh] Purging package artifacts at tigris:tuist-registry/registry/swift/zendesk/sdk_logger_ios
[purge-registry-package.sh] Purging package metadata at tigris:tuist-registry/registry/metadata/zendesk/sdk_logger_ios
[purge-registry-package.sh] Registry purge completed
```

which also optionally takes a version as `scripts/purge-registry-package.sh zendesk/sdk_logger_ios 0.11.0`

and
```
󰘧 · scripts/sync-registry-package.sh zendesk/sdk_logger_ios 0.11.0
[sync-registry-package.sh] Package input: zendesk/sdk_logger_ios
[sync-registry-package.sh] Using scope "zendesk", name "sdk_logger_ios", tag "0.11.0"
[sync-registry-package.sh] Randomly selected production cache node from cache/config/deploy.production.yml: cache-eu-central.tuist.dev
[sync-registry-package.sh] Connecting to cache-eu-central.tuist.dev
[sync-registry-package.remote:cache-eu-central] Looking for the running cache container
[sync-registry-package.remote:cache-eu-central] Using container cache-web-production-27d1501d58bf94a032c015715d46dec14083dbca_uncommitted_0f445be52d470843
[sync-registry-package.remote:cache-eu-central] Enqueueing Cache.Registry.ReleaseWorker for zendesk/sdk_logger_ios@0.11.0
Enqueued Oban job 9353899
[sync-registry-package.remote:cache-eu-central] Worker enqueue request finished
[sync-registry-package.sh] Registry sync request completed
```

### DIFF
--- a/mise/tasks/registry/purge.sh
+++ b/mise/tasks/registry/purge.sh
@@ -14,9 +14,15 @@ readonly LOCAL_REGISTRY_ROOT="/cas/registry/swift"
 cleanup_paths=()
 
 cleanup() {
-    for path in "${cleanup_paths[@]:-}"; do
-        [[ -n "$path" && -e "$path" ]] && rm -rf "$path"
+    local path
+
+    for path in "${cleanup_paths[@]}"; do
+        if [[ -n "$path" && -e "$path" ]]; then
+            rm -rf "$path"
+        fi
     done
+
+    return 0
 }
 
 trap cleanup EXIT


mise/tasks/registry/purge.sh
```diff
@@ -0,0 +1,290 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
pepicrft, line 6: Let's make this a `mise` task instead so we can run `mise run` and discover it through auto-completion.

scripts/purge-registry-package.sh
```diff
@@ -0,0 +1,290 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+readonly SCRIPT_NAME="$(basename "$0")"
+readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+readonly PROJECT_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
```
pepicrft: Once we move this to mise, the root is `MISE_PROJECT_ROOT`.

mise/tasks/registry/purge.sh
```diff
@@ -0,0 +1,290 @@
...
+
+require_command() {
+    local command_name="$1"
+
+    if ! command -v "$command_name" >/dev/null 2>&1; then
+        fail "Missing required command: ${command_name}"
+    fi
+}
+
+print_rclone_setup_instructions() {
```
pepicrft, line 36: Once we get all of this to Mise and use `#USAGE` comments, a big chunk of this goes away.

scripts/sync-registry-package.sh
```diff
@@ -0,0 +1,181 @@
...
+    local version
+    local normalized_scope
+    local repository_full_handle
+    local selected_host
+
+    if [[ "$#" -ne 2 ]]; then
+        usage
+    fi
+
+    require_command ssh
```
pepicrft: Can't we assume its presence in `macos`?

scripts/purge-registry-package.sh
```diff
@@ -0,0 +1,290 @@
...
+  rclone lsf ${REGISTRY_ROOT}/registry/metadata/ | head
+EOF
+}
+
+check_rclone_configuration() {
+    local remotes
+
+    log "Checking required commands"
+    require_command rclone
+    require_command jq
```
pepicrft: What if we add them to `mise.toml`, which is our source of truth for the tools that are needed to work on this repository? If we want to avoid installing them on `mise install`, you can use `mise x jq@version -- jq` and lazily install run them. I think it's a better experience than telling the contributor _you are missing x_

scripts/purge-registry-package.sh
```diff
@@ -0,0 +1,290 @@
...
+
+print_rclone_setup_instructions() {
+    cat >&2 <<EOF
+[${SCRIPT_NAME}] ERROR: rclone remote "${RCLONE_REMOTE}" is not configured.
+
+Set it up with the production registry Tigris credentials from 1Password/Kamal secrets, for example:
+  export S3_ACCESS_KEY_ID='...'
+  export S3_SECRET_ACCESS_KEY='...'
+  export S3_HOST='fly.storage.tigris.dev'
+  export S3_REGION='auto'
```
fortmarek: what do you think about using the 1p cli to get the secrets directly?

scripts/AGENTS.md
```diff
@@ -0,0 +1,9 @@
+# Scripts
```
pepicrft: If we move this to a Mise task, the description can be added via annotations, and the comments in the script can contain the context. I think we can drop this.

scripts/sync-registry-package.sh
```diff
@@ -0,0 +1,181 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+readonly SCRIPT_NAME="$(basename "$0")"
```
pepicrft: Similar comments regarding Mise and using `#usage` apply here.

cschmatzler:
> I think this is a great first step. I do have to wonder whether it would make sense for some of these scripts to be scripts directly in given services (such as resyncing being mostly implemented directly in the cache node) 

I thought about this, but then also thought about future scripts that cross-pollinate between cache and server and then it becomes a less clear cut. I can move it if you have a stronger preference than I do, or we can merge and move it later if we feel like the root has too many tasks, it's just a `mv` away :)

Also you added this comment before the whole mise-ification, so let me know how you feel about it now 😄

scripts/purge-registry-package.sh
cschmatzler: Not a bad idea, but the `rclone config` is a one-time command and I'm wondering whether we should have per-user access keys for this for audit logging purposes (hello Vanta); at least I'm using a `cschmatzler` one for it.  
But yeah, given that it's a one-time "go to Tigris, create a key and run a command" kind of workflow, I feel like the need for automating it is quite low compared to something that needs the key every single run.

PR #10074, opened 2026-03-27
fix(cli): skip static products linting when warning targets is .none

Hoontopia:
When `staticSideEffectsWarningTargets` is set to `.none`, the `StaticProductsGraphLinter` still performs the full DFS traversal of the dependency graph before discarding all results. This adds an early return to skip the traversal entirely, matching the user's intent of "no warnings."

On a large project (~6,700 targets), this reduces lint time from ~36s to ~10s.

### How to test locally

1. Set `staticSideEffectsWarningTargets: .none` in `Tuist.swift`
2. Run `tuist generate` and observe that no static side-effect warnings are emitted
3. Run the test:
```
  -only-testing TuistGeneratorTests/StaticProductsGraphLinterTests \
```


fortmarek (approved):
Thanks! looks good

PR #10084, opened 2026-03-27
feat(server): generate docs OG images with Carta and BrowseServo

pepicrft:
## Summary
- Adds `carta ~> 0.2.0` and `browse_chrome ~> 0.4.0` for HTML-to-image rendering using headless Chrome
- Creates `Tuist.Docs.OgImage` Phoenix component matching the [Figma design](https://www.figma.com/design/WfWG28o26G3reZkdIKcjtc/New-Website-images?node-id=352-4324&m=dev) (gradient background, DM Sans font, text shadows, Tuist | Docs branding, category label)
- Adds `mix docs.gen.og_images` task that generates 1920x1080 JPEG images for all docs pages across all locales, using a pool of 4 concurrent Chrome instances
- Wires up `head_image` and `head_twitter_card` assigns in DocsLive so generated images appear in `og:image` meta tags
- Integrates into the Dockerfile: installs Chromium in the builder stage and runs the generation step (cached by Docker layer cache when source doesn't change)

> [!NOTE]
> The OG image generation takes ~2 minutes. In Docker, the Chromium install adds ~20s and generation runs in the builder stage. The layer is cached by Docker BuildKit as long as `server/lib/` and `docs/` don't change. The alternative would have been runtime with object-storage caching, but I think there's some room to optimize this further before we consider the runtime model.

## Example
image

## Test plan
- [x] Run `mix docs.gen.og_images` and verify images in `priv/static/docs/images/og/generated/{locale}/`
- [x] Verify generated images match the Figma design
- [x] Docker build succeeds (`docker build -f server/Dockerfile --target builder .`)
- [x] Verify OG meta tags render correctly on docs pages

### DIFF
--- a/server/lib/mix/tasks/docs/gen/og_images.ex
+++ b/server/lib/mix/tasks/docs/gen/og_images.ex
@@ -6,6 +6,7 @@ defmodule Mix.Tasks.Docs.Gen.OgImages do
   use Boundary, classify_to: Tuist.Mix
 
   alias Tuist.Docs
+  alias Tuist.Docs.OgImage
   alias Tuist.Docs.Sidebar
 
   @pool Tuist.Docs.OgImagePool
@@ -24,7 +25,7 @@ defmodule Mix.Tasks.Docs.Gen.OgImages do
     logo_path =
       :tuist |> Application.app_dir("priv") |> Path.join("static/docs/images/logo.webp")
 
-    pool_size = System.schedulers_online() |> max(4)
+    pool_size = max(System.schedulers_online(), 4)
 
     {:ok, _} = Browse.start_link(@pool, implementation: BrowseChrome.Browser, pool_size: pool_size)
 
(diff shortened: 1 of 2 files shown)


server/mix.exs
```diff
@@ -155,7 +155,11 @@ defmodule Tuist.MixProject do
       {:opentelemetry_broadway, "~> 0.3"},
       {:loki_logger_handler, "~> 0.2"},
       {:processor, path: "../processor", runtime: false},
-      {:tidewave, "~> 0.5", only: :dev}
+      {:tidewave, "~> 0.5", only: :dev},
+      {:carta, "~> 0.2.0"},
+      {:browse_chrome, "~> 0.4.0"},
```
pepicrft, line 160: I tried with servo, but there are some CSS features that we need that are not yet supported unfortunately.

PR #10089, opened 2026-03-27
fix(cli): preserve external static xcframework deps for cached dynamics

pepicrft:
## Summary
- update `TuistCacheEE` to only prune static xcframework deps that came from replaced target dependencies
- keep external precompiled static xcframeworks like `GoogleMaps.xcframework` behind cached dynamic frameworks so module visibility and search-path generation still work
- add regression coverage for hostless tests and shared app/test consumers of cached dynamic frameworks

## Private dependency
- `tuist/TuistCacheEE#54`

## Test plan
- [x] `TUIST_EE=1 tuist generate --no-open`
- [x] `swiftformat cli/TuistCacheEE/Sources/Generator/CacheGraphMutator.swift cli/Tests/TuistCacheEETests/Generator/CacheGraphMutatorTests.swift`
- [x] `TUIST_EE=1 xcodebuild test -workspace Tuist.xcworkspace -scheme TuistCacheEEUnitTests -only-testing:TuistCacheEETests/CacheGraphMutatorTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- [ ] validate against the original GoogleMaps reproduction in the affected repo

## Notes
This keeps the previously added hostless-unit-test pruning for cached static target deps, and preserves the behavior that re-adds those cached static deps to non-test consumers. The new distinction is that external static xcframeworks are no longer treated as if they were baked into the cached dynamic artifact.


pepicrft:
Simple version:

This fixes one cache-replacement bug.

For hostless unit tests, we should prune cached static target dependencies from cached dynamic xcframeworks because those symbols are already baked into the dynamic binary. But we must not prune external static xcframeworks like `GoogleMaps.xcframework`, because downstream targets still need their module visibility and search paths.

Why this is safe:
- It only changes the cache mutator path.
- It preserves the previous behavior for cached static target dependencies.
- It preserves the shared app/test consumer behavior.
- It covers the explicit hostless-import case.
- Focused `CacheGraphMutatorTests` and `GraphTraverserTests` pass on this branch.

pepicrft:
Added one focused regression test for the reported graph:

`Feature -> Library (cached dynamic) -> GoogleMaps (external static xcframework)` and `FeatureTests -> Feature + Library`.

The assertion is simple: `GoogleMaps` must stay behind `Library.xcframework` and must not be re-added directly to `Feature`.

PR #10091, opened 2026-03-27
feat(server): make chart bars clickable to navigate to detail pages

pepicrft:
## Summary
- Add `url` field to chart data points across all dashboard bar charts (Xcode builds, Xcode cache, Gradle builds, Gradle cache, tests, module cache)
- Add click handler and pointer cursor in the Noora Chart.js hook so clicking a bar navigates to the detail page
- Generic implementation: any chart can opt-in by including a `url` field in its data points

### DIFF
--- a/noora/js/Chart.js
+++ b/noora/js/Chart.js
@@ -114,9 +114,12 @@ export default {
     );
     this.chart.setOption(option);
 
-    const hasClickableData = option.series && option.series.some(
-      (s) => s.data && s.data.some((d) => d && typeof d === "object" && d.url)
-    );
+    const hasClickableData =
+      option.series &&
+      option.series.some(
+        (s) =>
+          s.data && s.data.some((d) => d && typeof d === "object" && d.url),
+      );
 
     if (hasClickableData) {
       this.chart.on("click", (params) => {


noora/js/Chart.js
```diff
@@ -114,6 +114,31 @@ export default {
     );
     this.chart.setOption(option);
 
+    const clickUrlsEl = this.el.querySelector("[data-part='click-urls']");
```
fortmarek: this pattern is a bit odd – we're already passing options from phoenix to the JS chart, why can't we take a similar approach for the URLs instead of using a hidden html element with a data part?

pepicrft: This was an attempt to fix the wrong issue. Now bars include the right attribute that's used from the click event to determine where to navigate to.

server/lib/tuist_web/live/gradle_cache_live.ex
```diff
@@ -203,19 +203,20 @@ defmodule TuistWeb.GradleCacheLive do
...
-  defp assign_recent_builds(%{assigns: %{selected_project: project}} = socket, _params) do
+  defp assign_recent_builds(%{assigns: %{selected_project: project, selected_account: account}} = socket, _params) do
     {builds, _meta} = Gradle.list_builds(project.id, %{page_size: @recent_builds_page_size})
     builds = Repo.preload(builds, :built_by_account)
 
+    reversed_builds = Enum.reverse(builds)
+
     recent_builds_chart_data =
-      builds
-      |> Enum.reverse()
```
cschmatzler, line 212: This can still use the pipe instead of assigning `reversed_builds`, no?

PR #10093, opened 2026-03-27
feat(noora,server): add search and sorting to Ran by filter

fortmarek:
https://github.com/user-attachments/assets/d7e24504-4b1c-4f1c-993d-03dba6b8c371

## Summary
- Sort "Ran by" filter options alphabetically (case-insensitive) while keeping "CI" first
- Add a searchable text input to the option dropdown for filters marked as `searchable: true`
- Add `searchable` field to `Noora.Filter.Filter` struct (defaults to `false`)
- New `NooraDropdownSearch` JS hook handles client-side item filtering, keyboard isolation, and auto-reset on close

## Test plan
- [ ] Verify "Ran by" dropdown options are sorted alphabetically on all dashboard pages (test runs, cache runs, generate runs, xcode build runs, gradle build runs, test case)
- [ ] Verify the search input appears in the "Ran by" dropdown and filters items as you type
- [ ] Verify typing in search does not trigger dropdown typeahead navigation
- [ ] Verify search clears when dropdown closes and reopens
- [ ] Verify "CI" always appears first in the list
- [ ] Verify other option-type filters without `searchable: true` are unaffected

### DIFF
--- a/server/lib/tuist_web/live/test_case_live.ex
+++ b/server/lib/tuist_web/live/test_case_live.ex
@@ -93,9 +93,7 @@ defmodule TuistWeb.TestCaseLive do
     if Accounts.organization?(project.account) do
       {:ok, organization} = Accounts.get_organization_by_id(project.account.organization_id)
 
-      users =
-        organization
-        |> Accounts.get_organization_members()
+      users = Accounts.get_organization_members(organization)
 
       filters ++
         [
--- a/server/lib/tuist_web/live/cache_runs_live.ex
+++ b/server/lib/tuist_web/live/cache_runs_live.ex
@@ -85,9 +85,7 @@ defmodule TuistWeb.CacheRunsLive do
       if Accounts.organization?(project.account) do
         {:ok, organization} = Accounts.get_organization_by_id(project.account.organization_id)
 
-        users =
-          organization
-          |> Accounts.get_organization_members()
+        users = Accounts.get_organization_members(organization)
 
         [
           %Filter.Filter{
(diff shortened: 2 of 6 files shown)


noora/js/Dropdown.js
```diff
@@ -167,6 +169,44 @@ export default {
     window.addEventListener("phx:close-dropdown", this.handleCloseDropdown);
   },
 
+  setupSearch() {
+    const searchInput = this.el.querySelector('[data-part="search-input"]');
+    if (!searchInput) return;
+
+    searchInput.addEventListener("input", () => {
```
fortmarek, line 176: this is a pretty basic client-side search. I can see us wanting to eventually hold the state in the phoenix layer which would allow us to also search for values not in the list (for very long lists), but it felt adding unnecessary complexity at this stage. I can look into that if you feel otherwise, however

cschmatzler: Think this is a good trade-off for the type of lists we currently have. If we ever have something that is 1000+ items we can still add a `search_type: client | server` bit or replace this, but this is faster and easier for the specific use case we're having.

server/lib/tuist_web/live/test_case_live.ex
```diff
@@ -92,7 +92,11 @@ defmodule TuistWeb.TestCaseLive do
 
     if Accounts.organization?(project.account) do
       {:ok, organization} = Accounts.get_organization_by_id(project.account.organization_id)
-      users = Accounts.get_organization_members(organization)
+
+      users =
+        organization
+        |> Accounts.get_organization_members()
+        |> Enum.sort_by(fn user -> user.account.name end)
```
cschmatzler: Instead of sprinkling this in every live view, is there any downside to just `sort_by` in the domain function and leave the mount handlers untouched?

fortmarek: agree and done 🙂

PR #10094, opened 2026-03-27
feat(cli,server,xcode_processor): add remote processing mode for tuist inspect test

fortmarek:
## Summary

A new Elixir/Phoenix service (`xcode_processor/`) for server-side xcresult processing on macOS. This offloads test result parsing from the CLI to dedicated Scaleway Mac minis, following the same architecture as the existing Linux `processor/` service for xcactivitylog files.

### What's included

**Xcode Processor Service** (`xcode_processor/`)
- Elixir/Phoenix webhook service that receives xcresult archives, downloads from S3, and parses via a Swift NIF
- Swift NIF calls `xcresulttool` + custom parsing logic, returns structured JSON
- Prometheus metrics, Sentry error tracking, Grafana dashboard

**CLI changes**
- `tuist inspect test --mode remote`: uploads raw xcresult for server-side processing instead of parsing locally. Default is `local` for now; will switch to `remote` once validated e2e against tuist/tuist.
- CLI generates test run ID, uploads xcresult first, then creates test run with \`status: "processing"\` (same pattern as builds)
- Shared \`XCResultParser\` package: models and parsing logic deduplicated between CLI and NIF

**Server changes**
- \`ProcessXcresultWorker\`: Oban worker that sends webhooks to the xcode processor, replaces test run with parsed data
- Test run "processing" status with UI support (processing badge, refresh button, illustration)
- Client-generated test run ID support in the create test API

**Infrastructure**
- nix-darwin platform config for Scaleway Mac minis (launchd services, sops-nix secrets, Caddy TLS, Grafana Alloy)
- Automated provisioning: \`mise run provision <name>\` (creates machine, installs Nix, generates age key, sets up SSH, stores secrets in 1Password)
- Automated config management: \`mise run darwin-apply <host> <hostname>\`
- Automated deploys: \`mise run deploy <host> <env>\` and GitHub Actions workflows
- Three environments deployed: staging, canary, production

## Provisioning a new machine

```bash
cd xcode_processor

# 1. Provision (creates machine, installs Nix, generates age key, updates sops, sets up SSH)
mise run provision xcode-processor-<name>

# 2. Add host config + flake entry + DNS record, commit and push

# 3. Apply nix-darwin config (creates services, decrypts secrets, sets up Caddy)
mise run darwin-apply <ip> xcode-processor-<name>

# 4. Deploy
mise run deploy <ip> <environment>
```

## Current infrastructure

| Environment | URL | IP |
|---|---|---|
| Production | https://xcode-processor.tuist.dev | 62.210.195.94 |
| Canary | https://xcode-processor-canary.tuist.dev | 51.159.120.166 |
| Staging | https://xcode-processor-staging.tuist.dev | 51.159.120.11 |

## Test plan
- [x] NIF parses real xcresult bundles (2600+ test cases)
- [x] Full e2e: CLI upload -> server webhook -> xcode_processor parse -> test run created
- [x] Staging e2e verified
- [x] CI deploys working (staging workflow)
- [x] Grafana dashboard showing metrics
- [x] Sentry error tracking configured
- [x] Processing UI (badge, refresh, illustration)
- [x] Elixir tests (server worker, xcode_processor webhook/auth/processor)
- [x] Swift tests (IPSCrashReportParser)

### DIFF
--- a/xcode_processor/lib/xcode_processor/xcresult_nif.ex
+++ b/xcode_processor/lib/xcode_processor/xcresult_nif.ex
@@ -12,20 +12,29 @@ defmodule XcodeProcessor.XCResultNIF do
 
   @on_load :load_nif
 
+  @nif_loaded :nif_load_status
+
   def load_nif do
     nif_path = ~c"#{:code.priv_dir(:xcode_processor)}/native/xcresult_nif"
 
     case :erlang.load_nif(nif_path, 0) do
       :ok ->
+        :persistent_term.put(@nif_loaded, true)
         :ok
 
       {:error, {:reload, _}} ->
+        :persistent_term.put(@nif_loaded, true)
         :ok
 
       {:error, reason} ->
+        :persistent_term.put(@nif_loaded, false)
         Logger.error("Failed to load xcresult NIF from #{nif_path}: #{inspect(reason)}")
-        {:error, reason}
+        :ok
     end
+  end
+
+  def nif_loaded? do
+    :persistent_term.get(@nif_loaded, false)
   end
 
   @doc """
--- a/xcode_processor/lib/xcode_processor_web/controllers/health_controller.ex
+++ b/xcode_processor/lib/xcode_processor_web/controllers/health_controller.ex
@@ -2,10 +2,19 @@ defmodule XcodeProcessorWeb.HealthController do
   use XcodeProcessorWeb, :controller
 
   def check(conn, _params) do
-    json(conn, %{
-      status: "ok",
-      version: System.get_env("DEPLOY_ENV", "dev"),
-      git_sha: System.get_env("GIT_SHA", "unknown")
-    })
+    if XcodeProcessor.XCResultNIF.nif_loaded?() do
+      json(conn, %{
+        status: "ok",
+        version: System.get_env("DEPLOY_ENV", "dev"),
+        git_sha: System.get_env("GIT_SHA", "unknown")
+      })
+    else
+      conn
+      |> put_status(503)
+      |> json(%{
+        status: "error",
+        reason: "xcresult NIF not loaded"
+      })
+    end
   end
 end


cli/Sources/TuistInspectCommand/Commands/InspectTestCommand.swift
```diff
@@ -35,14 +40,22 @@
         )
         var resultBundlePath: String?
 
+        @Option(
+            name: .long,
+            help: "Processing mode: 'local' parses the xcresult on this machine, 'remote' uploads it for server-side processing.",
+            envKey: .inspectTestMode
+        )
+        var mode: ProcessingMode = .local
```
fortmarek, line 48: I'll update the mode once I test things e2e against `tuist/tuist`

xcode_processor/mise/tasks/provision.sh
fortmarek, line 1: this is sort of on the best effort basis ... it's unlikely we will need to run this again anytime soon, but figured it'd be good to keep around just in case.

cschmatzler: Even if we never use this again, it's a great historical reference.

xcode_processor/native/xcresult_nif/Sources/XCResultParser/TestCase.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-public struct TestCase {
+public struct TestCase: Encodable, Sendable {
     public let name: String
     public let testSuite: String?
```
cschmatzler, line 5: This sets the test suite as `testSuite`, but `Tuist.Test:1022` reads `test_suite_name` instead, so I _think_ suit grouping is broken because of the name mismatch.

xcode_processor/native/xcresult_nif/Sources/XCResultParser/TestCaseFailure.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
+import Path
+
+public struct TestCaseFailure: Encodable, Sendable {
+    public enum IssueType: String, Encodable, Sendable {
+        case errorThrown = "Thrown Error"
+        case assertionFailure = "Assertion Failure"
+        case issueRecorded = "Issue Recorded"
```
cschmatzler: `TuistWeb.Helpers.FailureMessage` expects these to be in snake-case.

server/lib/tuist_web/controllers/api/tests_controller.ex
```diff
@@ -450,18 +455,41 @@ defmodule TuistWeb.API.TestsController do
...
+            project_id: selected_project.id,
+            is_ci: test_run.is_ci || false,
+            git_branch: test_run.git_branch,
+            git_commit_sha: test_run.git_commit_sha,
+            git_ref: test_run.git_ref,
+            macos_version: test_run.macos_version,
+            xcode_version: test_run.xcode_version,
+            model_identifier: test_run.model_identifier,
+            scheme: test_run.scheme
+          }
```
cschmatzler, line 478: This omits CI fields like `ci_run_id`, `ci_host` and `ci_provider`. As far as I can see, we're not adding them anywhere else.

xcode_processor/lib/xcode_processor/xcresult_nif.ex
```diff
@@ -0,0 +1,57 @@
...
+    case :erlang.load_nif(nif_path, 0) do
+      :ok ->
+        :ok
+
+      {:error, {:reload, _}} ->
+        :ok
+
+      {:error, reason} ->
+        Logger.warning("Failed to load xcresult NIF from #{nif_path}: #{inspect(reason)}")
+        :ok
```
cschmatzler: What happens to processing if we silently fail here and the service comes up healthy without the NIF working? I wonder if it'd be better to fail fast here instead of accepting processing requests and marking them failed.

server/lib/tuist/tests/test.ex
```diff
@@ -95,7 +95,7 @@ defmodule Tuist.Tests.Test do
       :ran_at,
       :build_system
     ])
-    |> validate_inclusion(:status, ["success", "failure", "skipped", "in_progress"])
+    |> validate_inclusion(:status, ["success", "failure", "skipped", "in_progress", "processing", "failed_processing"])
```
cschmatzler, line 98: These need to be included in the OpenApiSpex schemas.

xcode_processor/mise/tasks/provision.sh
fortmarek: Agreed, keeping it.

xcode_processor/native/xcresult_nif/Sources/XCResultParser/TestCase.swift
fortmarek: Good catch. Fixed the CodingKey to encode as `test_suite_name` in 5c3a052.

xcode_processor/native/xcresult_nif/Sources/XCResultParser/TestCaseFailure.swift
fortmarek: Fixed -- changed the enum raw values to `error_thrown`, `assertion_failure`, `issue_recorded` to match the server in 5c3a052.

server/lib/tuist_web/controllers/api/tests_controller.ex
fortmarek: Added `ci_run_id`, `ci_host`, and `ci_provider` to the worker job args and `base_attrs` in 5c3a052.

xcode_processor/lib/xcode_processor/xcresult_nif.ex
fortmarek: Good point. Changed to return `{:error, reason}` so the application fails to start when the NIF can't load, instead of silently accepting requests. Fixed in 5c3a052.

server/lib/tuist/tests/test.ex
fortmarek: Added `failed_processing` to all three OpenApiSpex status enums in tests_controller.ex in 5c3a052.

xcode_processor/lib/xcode_processor/xcresult_nif.ex
fortmarek: Updated approach: `load_nif` returns `:ok` so the module is available (needed for Mimic in tests and for the fallback `:nif_not_loaded` error in `parse_nif/2`), but we now track load status via `persistent_term` and expose it through the health endpoint -- returns 503 when the NIF isn't loaded, so load balancers won't route traffic to a broken instance. See 472bc39.

PR #10105, opened 2026-03-30
feat(cli): add --no-lint flag to tuist generate

Hoontopia:
Adds a `--no-lint` flag to `tuist generate` that skips graph linting entirely.

Graph linting is a validation-only step that does not affect the generated project. On large projects (~6,700 targets), linting can take ~36s. This flag allows developers to skip it during local development while keeping it enabled in CI.

Follows the existing `--no-open` and `--no-binary-cache` flag patterns.

Also available via environment variable: `TUIST_GENERATE_LINT=false`.

### How to test locally

1. Run `tuist generate --no-lint` and verify no lint warnings are emitted
2. Run `tuist generate` (without flag) and verify lint warnings still appear
3. Run tests:
```
  -only-testing TuistKitTests/GenerateServiceTests \
```

### DIFF
--- a/cli/Sources/TuistGenerateCommand/GenerateRunCommand.swift
+++ b/cli/Sources/TuistGenerateCommand/GenerateRunCommand.swift
@@ -65,6 +65,12 @@
         )
         var cacheProfile: CacheProfileType?
 
+        @Flag(
+            help: "Run graph linting during generation.",
+            envKey: .generateLint
+        )
+        var lint: Bool = true
+
         @Option(
             name: .shortAndLong,
             help: "Configuration to generate for."
@@ -86,6 +92,7 @@
                 path: path,
                 includedTargets: Set(includedTargets),
                 noOpen: !open,
+                skipLint: !lint,
                 configuration: configuration,
                 ignoreBinaryCache: !binaryCache,
                 cacheProfile: cacheProfile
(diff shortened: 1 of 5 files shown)


cli/Sources/TuistGenerateCommand/GenerateRunCommand.swift
```diff
@@ -65,6 +65,12 @@
         )
         var cacheProfile: CacheProfileType?
 
+        @Flag(
+            help: "Skip graph linting during generation.",
+            envKey: .generateLint
+        )
+        var lint: Bool = true
```
pepicrft, line 72: Can we use Swift Argument Parser’s inverted Boolean flag support here? For this codebase, the `@Flag(..., envKey:)` overload already applies `inversion: .prefixedNo`, so this can be expressed as a `--lint` / `--no-lint` pair instead of documenting the flag as `--skip-lint`. I’d update the help to describe the positive value, for example `Run graph linting during generation.`. That keeps this aligned with `--open` / `--no-open` and avoids a help entry where `--lint` reads as if it skips linting.

Hoontopia: I'll change the help text to `Run graph linting during generation.`
so `--lint` / `--no-lint` reads correctly instead of looking like a skip flag.

Hoontopia: Pushed. `--lint` / `--no-lint` now describes the positive value.

PR #10106, opened 2026-03-30
feat(cache): add k6 load testing suite with load-test PR label

cschmatzler:
- Fix `cache-staging-deploy.yml` to actually use the `commit_sha` input
- Add a k6 TypeScript harness under `cache/k6` covering:
  - KV read/write
  - Xcode CAS read/write
  - module cache exists/read/write
  - Gradle cache read/write
- Use staging-safe traffic profiles and payload sizes:
  - Rates: KV read `512/s`, KV write `256/s`, Xcode read `512/s`, Xcode write `64/s`, module exists `64/s`, module read `128/s`, module write `4/s`, Gradle read `128/s`, Gradle write `4/s`
  - Sizes: `10KB` / `256KB` / `2MB` for Xcode, `10MB` / `25MB` / `50MB` for module and Gradle
  - Each scenario runs sequentially for `2m30s` with warmup, steady-state, and a `1.5x` burst phase
- Trigger cache load tests by applying the `load-test` label to a same-repo PR
- `cache-load-test.yml` now:
  - deploys `main` to staging first
  - runs baseline tests in `eu-central` and `us-east`
  - captures and reuses baseline seed data for the PR run
  - deploys the PR branch to staging
  - compares PR vs `main`, posts/updates a sticky PR comment, removes the trigger label, and restores `main` afterward
- Regression gates:
  - p50 latency must stay within `10%` of baseline
  - throughput must stay above `80%` of baseline
  - error rate must not increase by more than `10` percentage points

### DIFF
--- a/cache/k6/src/main.ts
+++ b/cache/k6/src/main.ts
@@ -47,6 +47,7 @@ interface ScenarioConfig {
   duration: number;
   expectedLatencyMs: number;
   startTime: string;
+  gracefulStopSeconds: number;
   preAllocatedVUs?: number;
   maxVUs?: number;
 }
@@ -62,6 +63,7 @@ function makeScenario(cfg: ScenarioConfig): any {
     rate: cfg.rate,
     timeUnit: '1s',
     duration: `${cfg.duration}s`,
+    gracefulStop: `${cfg.gracefulStopSeconds}s`,
     preAllocatedVUs,
     maxVUs,
     startTime: cfg.startTime,
@@ -73,16 +75,19 @@ function makeScenario(cfg: ScenarioConfig): any {
 const scenarios: Record<string, any> = {};
 let offset = 0;
 for (const plan of SCENARIO_PLANS) {
+  const gracefulStopSeconds = Math.max(Math.ceil((plan.expectedLatencyMs * 2) / 1000), 1);
+
   scenarios[plan.key] = makeScenario({
     exec: plan.exec,
     rate: plan.rate,
     duration: plan.duration,
     expectedLatencyMs: plan.expectedLatencyMs,
     startTime: `${offset}s`,
+    gracefulStopSeconds,
     preAllocatedVUs: plan.preAllocatedVUs,
     maxVUs: plan.maxVUs,
   });
-  offset += plan.duration;
+  offset += plan.duration + gracefulStopSeconds;
 }
 
 export const options: Partial<Options> = {
--- a/cache/k6/src/payloads.ts
+++ b/cache/k6/src/payloads.ts
@@ -10,12 +10,7 @@ for (const bucket of [...XCODE_SIZES, ...LARGE_SIZES]) {
 const cache: Record<string, ArrayBuffer> = {};
 
 function buildPayload(size: number): ArrayBuffer {
-  const buf = new ArrayBuffer(size);
-  const view = new Uint8Array(buf);
-  for (let i = 0; i < size; i++) {
-    view[i] = i & 0xff;
-  }
-  return buf;
+  return new ArrayBuffer(size);
 }
 
 export function getPayloadForSize(size: number): ArrayBuffer {


.github/workflows/cache-load-test.yml
```diff
@@ -0,0 +1,590 @@
...
+          echo "pr_ref=${{ github.event.pull_request.head.ref }}" >> "$GITHUB_OUTPUT"
+          echo "pr_sha=${{ github.event.pull_request.head.sha }}" >> "$GITHUB_OUTPUT"
+          MAIN_SHA=$(gh api "/repos/${{ github.repository }}/commits/main" --jq '.sha')
+          echo "main_sha=$MAIN_SHA" >> "$GITHUB_OUTPUT"
+
+      - name: Remove label
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh api "/repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/labels/load-test" -X DELETE || true
```
fortmarek, line 44: can already see the rate limits 🙈 😅

.github/workflows/cache-load-test.yml
```diff
@@ -0,0 +1,590 @@
...
+          AUTH_PASSWORD: tuistrocks
+          ACCOUNT_HANDLE: tuist
+          PROJECT_HANDLE: tuist
+        run: |
+          set -euo pipefail
+
+          AUTH_RESPONSE=$(curl -fsS "$SERVER_URL/api/auth" \
+            -H 'Content-Type: application/json' \
+            -d "{\"email\":\"$AUTH_EMAIL\",\"password\":\"$AUTH_PASSWORD\"}")
+          ACCESS_TOKEN=$(printf '%s' "$AUTH_RESPONSE" | grep -o '"access_token":"[^"]*"' | head -1 | cut -d '"' -f4)
```
fortmarek: I'd consider using an account token instead. We can use the cli to create one or we can precreate one as part of the seed data (and make sure it's part of the current staging)

.github/workflows/cache-load-test.yml
```diff
@@ -0,0 +1,590 @@
...
+        with:
+          ssh-private-key: ${{ secrets.CACHE_SSH_PRIVATE_KEY }}
+      - name: Deploy main with Kamal
+        env:
+          OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.CACHE_OP_SERVICE_ACCOUNT_TOKEN }}
+          SECRETS: ${{ secrets.CACHE_KAMAL_SECRETS }}
+        run: mise run deploy "staging"
+
+  baseline:
+    name: Baseline (${{ matrix.region }})
```
fortmarek: I do wonder if running the benchmark itself (not the deploy) could be a mise task in a single script, so you could run it locally if needed (or have an agent run it)

.github/workflows/cache-load-test.yml
```diff
@@ -0,0 +1,590 @@
...
+          MAIN_SHA=$(gh api "/repos/${{ github.repository }}/commits/main" --jq '.sha')
+          echo "main_sha=$MAIN_SHA" >> "$GITHUB_OUTPUT"
+
+      - name: Remove label
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh api "/repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/labels/load-test" -X DELETE || true
+
+  deploy-main:
```
fortmarek, line 46: this is something I'd maybe consider running nightly off our working hours, so we continuously measure roughly the load times instead of doing it on-demand.

Running this continuously would also help with this script still working if we don't run for a couple weeks and suddenly we get back to doing perf work in this area.

.github/workflows/cache-load-test.yml
```diff
@@ -0,0 +1,590 @@
...
+        uses: webfactory/ssh-agent@v0.9.1
+        with:
+          ssh-private-key: ${{ secrets.CACHE_SSH_PRIVATE_KEY }}
+      - name: Deploy with Kamal
+        env:
+          OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.CACHE_OP_SERVICE_ACCOUNT_TOKEN }}
+          SECRETS: ${{ secrets.CACHE_KAMAL_SECRETS }}
+        run: mise run deploy "staging"
+
+  test:
```
fortmarek, line 106: this is basically the same as `baseline` with small adjustments here and there. All the more reason to extract the testing into a script that we could reuse between the baseline and the PR (and then we'd just store the artifacts under a different name based on the sha/branch/whatever)

.github/workflows/cache-load-test.yml
```diff
@@ -0,0 +1,590 @@
...
+          P50_ABS_THRESHOLD="10"
+          THROUGHPUT_THRESHOLD="0.97"
+          ERROR_RATE_THRESHOLD="0"
+          VERDICT="pass"
+          REGRESSIONS=""
+          SUMMARY=""
+
+          add_failure() {
+            local region="$1" metric="$2" check_name="$3" baseline="$4" pr="$5" delta="$6"
+            REGRESSIONS="${REGRESSIONS}| ${region} | ${metric} | ${check_name} | ${baseline} | ${pr} | ${delta} | ❌ |
```
fortmarek: writing a bash is certainly easier with LLMs, but I do wonder if for parts of this we should have picked a higher-level programming language that'd be easier to follow. Don't feel strongly about it.

fortmarek: I don't think it's necessarily a need right now, but I can see having a need for an internal service where we could also run our own GitHub app which would allow us to be more structured about this, such as by storing the results into a db, reusing elixir code to post comments, etc. But that's certainly out of scope.

.github/workflows/cache-load-test.yml
```diff
@@ -0,0 +1,590 @@
...
+        working-directory: cache/k6
+        run: |
+          set -euo pipefail
+          k6 run src/main.ts \
+            -e CACHE_HOST=${{ matrix.cache_host }} \
+            -e CACHE_AUTH_TOKEN="$CACHE_AUTH_TOKEN" \
+            -e SERVER_URL=https://staging.tuist.dev \
+            -e REGION=${{ matrix.region }} \
+            -e RUN_ID="lt-baseline-${{ github.run_id }}-${{ matrix.region }}" \
+            -e COMMIT_SHA=${{ needs.resolve.outputs.main_sha }} \
```
fortmarek: k6 supports exporting the data to grafana via prometheus (--out experimental-prometheus-rw). If we run this nightly, we could track the data on `main` in a graph over time. But consider this a possible future improvement.

cache/k6/src/lib/seed.ts
fortmarek, line 1: can we reuse the seeding we already have in elixir instead?

.github/workflows/cache-load-test.yml
cschmatzler: Yup it should be.

cache/k6/src/lib/seed.ts
cschmatzler: Elixir seeds don't generate files and also don't upload those files to S3 for the full end-to-end testing bit, and uploading files to S3 within the seed script feels... weird to me? 
Don't feel incredibly strongly about it, though.

.github/workflows/cache-load-test.yml
cschmatzler: Ohhhhhhhhh.
I'll put it on the list :)

.github/workflows/cache-load-test.yml
cschmatzler: let's close our eyes and pretend they dont exist

PR #10112, opened 2026-03-30
fix(cli): strip dSYMs and compress shard bundle before upload

fortmarek:
## Summary

- Strip `.dSYM` and `.swiftmodule` directories from the `.xctestproducts` bundle before uploading for sharding -- these are only needed for symbolication/compilation, not for running tests
- Replace ZIPFoundation/ditto with AppleArchive (LZFSE compression) for both upload and download
- No new external dependencies -- AppleArchive is a built-in macOS framework

### Why AppleArchive over ZIP?

We benchmarked multiple approaches on a real-world 28 GB `.xctestproducts` bundle stripped to 7.2 GB:

| Method | Compress | Decompress | Size |
|---|---|---|---|
| ditto (deflate zip) | 3m 01s | 14m 35s | 2.1 GB |
| ZIPFoundation (deflate) | 2m 40s | **FAILS** | 2.09 GB |
| tar + zstd -1 | 24s | 34s | 2.0 GB |
| **AppleArchive (LZFSE)** | **20s** | **10s** | **1.99 GB** |

- **ZIPFoundation** crashes on extraction with `uncontainedSymlink` because `.xctestproducts` bundles contain relative symlinks that traverse parent directories (`../../Binaries/0/Debug-iphonesimulator`). This is the bug referenced in the original `ditto` comment.
- **LZFSE** is hardware-accelerated on Apple Silicon and AppleArchive is internally multi-threaded (observed 219% CPU utilization), explaining the large speed advantage.
- **tar + zstd** was fast but requires shelling out to `tar` and adding a custom archive format. AppleArchive achieves the same performance with a pure Swift API.
- Since both upload (CI Mac build machine) and download (CI Mac shard runners) are macOS, the Apple-only format is not a limitation.

### Context

Qonto's shard downloads were timing out because the `bundle.zip` was ~28 GB uncompressed. The CLI's `timeoutIntervalForResource` is 300s (5 min), requiring ~100 MB/s sustained throughput -- unrealistic for most CI runners downloading from Tigris.

Net effect: **28 GB uncompressed zip** becomes **1.99 GB compressed archive** (14x smaller), with compression taking 20s instead of 3 minutes.

## Test plan

- [ ] Verify shard upload/download cycle works end-to-end on a project with dSYMs
- [ ] Verify test execution succeeds after downloading the stripped bundle (no missing dSYMs needed at runtime)
- [ ] Verify symlinks are preserved through the archive round-trip

Depends on tuist/FileSystem#303 (for `ZipOptions`, used by other callers of `zipFileOrDirectoryContent`)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -893,16 +893,6 @@ var targets: [Target] = [
         path: "cli/Tests/TuistCASTests"
     ),
     .testTarget(
-        name: "TuistAppleArchiverTests",
-        dependencies: [
-            "TuistAppleArchiver",
-            fileSystemDependency,
-            .product(name: "FileSystemTesting", package: "tuist.FileSystem"),
-            pathDependency,
-        ],
-        path: "cli/Tests/TuistAppleArchiverTests"
-    ),
-    .testTarget(
         name: "TuistOIDCTests",
         dependencies: [
             "TuistOIDC",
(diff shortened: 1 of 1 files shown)


fortmarek:
> Would it make sense to consider the same archiving/unarchiving solution for our module cache?

Yes, it would. In fact, I think anything potentially going over >100 MB is worth compressing as compression will be always faster than download/upload.

Issue #7551, opened 2025-05-05
Copy Headers delayed by unsandboxed script phases warning


TamarMilchtaich:
### What happened?

Hi,

We have a static library with some headers, and build phase script which has an output file but no inputs.
When building this target, there's a warning:

```
warning: tasks in 'Copy Headers' are delayed by unsandboxed script phases; set ENABLE_USER_SCRIPT_SANDBOXING=YES to enable sandboxing (in target 'MyStaticLibrary' from project 'MyStaticLibrary')
```

Currently with Tuist, the pre- scripts are always before the Headers phase, which means this we will always get such warning for those intentionally unsandboxed script.

**Suggested solution**: since it's not possible to just suppress the warning, to solve this warning it makes sense to allow the script to be run after Headers build phase. There are some cases were the script must be ran before it (e.g. it generates headers, which is something we might want to support), but we can make it optional for the user. There might be some other workarounds like adding a dummy input file that changes every time, but this is a bit hacky, and I think it makes sense to provide some structural way to not have this warning.

### How do we reproduce it?

[ios_app_with_script_warning.zip](https://github.com/user-attachments/files/20037164/ios_app_with_script_warning.zip)

1. Download the project and unzip
2. Run `tuist generate`
3. Build the target, and observe the warning

### Error log

```
warning: tasks in 'Copy Headers' are delayed by unsandboxed script phases; set ENABLE_USER_SCRIPT_SANDBOXING=YES to enable sandboxing (in target 'MyStaticLibrary' from project 'MyStaticLibrary')
```

### macOS version

15.1

### Tuist version

4.46.1

### Xcode version

16.1

pepicrft:
> to solve this warning it makes sense to allow the script to be run after Headers build phase.

What would you suggest as an API here?

PR #10116, opened 2026-03-30
fix(cli): declare symlink target in macro copy script input paths for sandbox compatibility

zippi-MD:
### Problem

When `ENABLE_USER_SCRIPT_SANDBOXING=YES` and a target depends on a Swift Macro, the "Copy Swift Macro executable into $BUILT_PRODUCT_DIR" script phase fails during **archive builds** with:

```
error: Sandbox: cp deny(1) file-read-data .../IntermediateBuildFilesPath/UninstalledProducts/macosx/<MacroExecutable>
cp: .../BuildProductsPath/Release/<MacroExecutable>: Operation not permitted
```

### Root cause

During archive builds, Xcode places macOS tool products (like Swift Macro executables) in `$OBJROOT/UninstalledProducts/macosx/` and creates a **symlink** from `$BUILD_DIR/$CONFIGURATION/` pointing to that location.

The script phase declared only `$BUILD_DIR/$CONFIGURATION/<executable>` as an input path. When user script sandboxing is enabled, Xcode generates a sandbox profile that allows reads only from declared input paths. The `cp` command follows the symlink to the real file at `$OBJROOT/UninstalledProducts/macosx/<executable>`, but that path was not declared as an input, so the sandbox blocks the read.

This does not occur during regular builds because `$BUILD_DIR/$CONFIGURATION/<executable>` is a real file (not a symlink) in that context.

### Fix

Add `$(OBJROOT)/UninstalledProducts/macosx/<executable>` as an additional input path to the script phase. This tells the sandbox to also permit reading from the symlink target during archive builds. During regular builds this path does not exist and is harmlessly ignored.

### How to test locally

1. Create a project with a target that depends on a Swift Macro executable
2. Set `ENABLE_USER_SCRIPT_SANDBOXING=YES` on the target
3. Run `xcodebuild archive` for the scheme containing that target
4. Without this fix: archive fails with sandbox `deny(1) file-read-data` on the macro executable
5. With this fix: archive succeeds


waltflanagan (approved):
Nice work!  This should fix #7551

cli/Sources/TuistGenerator/Generator/BuildPhaseGenerator.swift
```diff
@@ -507,7 +507,12 @@ struct BuildPhaseGenerator: BuildPhaseGenerating {
         \(copyLines.joined(separator: "\n"))
         """
 
-        copySwiftMacrosBuildPhase.inputPaths = executableNames.map { "$BUILD_DIR/$CONFIGURATION/\($0)" }
+        copySwiftMacrosBuildPhase.inputPaths = executableNames.flatMap {
```
pepicrft, line 510: I'd add a unit test to prevent regressions down the line.

zippi-MD: Agree!! I updated `generateLinks_generatesAShellScriptBuildPhase_when_targetIsAMacroFramework` in `BuildPhaseGeneratorTests` to assert that the build phase includes both input paths

PR #10117, opened 2026-03-30
feat(server,cli): add paginated test case events endpoint

justine-acorns:
## Summary

- Add `GET /api/projects/:account/:project/tests/test-cases/:id/events` paginated API endpoint exposing test case lifecycle events (e.g. `first_run`, `marked_flaky`, `quarantined`, `unquarantined`)
- Add `list_test_case_events` MCP tool for AI assistant access to event history
- Add `tuist test case events <identifier>` CLI subcommand with text and JSON output formats

## Motivation

Per review feedback, rather than adding individual `flaky_since`/`quarantined_since` timestamp fields to the show endpoint, expose the full event history. This approach:
- Scales to future event types without new fields or ClickHouse queries
- Provides richer context (actor attribution, full timeline)
- Reuses the existing `test_case_events` ClickHouse table and `Tests.list_test_case_events/2` function

## Test plan

- [x] Server tests pass (36 tests — controller + MCP tool)
- [x] CLI tests pass (4 tests — TestCaseEventsCommandServiceTests)
- [ ] CI passes


pepicrft (approved):
Thanks a lot @justine-acorns 🙏🏼. Would you mind adding the new attributes to [this MCP tool too](https://github.com/tuist/tuist/blob/main/server/lib/tuist/mcp/components/tools/list_test_cases.ex) for consistency.

justine-acorns:
> Thanks a lot @justine-acorns 🙏🏼. Would you mind adding the new attributes to [this MCP tool too](https://github.com/tuist/tuist/blob/main/server/lib/tuist/mcp/components/tools/list_test_cases.ex) for consistency.

I only added it to `get_test_case`, since I only added the new fields to `tuist test case show` and not `tuist test case list`. I can add it to `list_test_cases` too, though, if you think it should be there! It will require batching requests for all of the test IDs to get the status events.

fortmarek (commented):
Thanks for the PR!

However, I think `flaky_since` and `quarantined_since` as fields on the show endpoint aren't the best fit here. A couple of concerns:

1. They conflate the status-change event with the current state. A test can be marked flaky, unmarked, then marked again. flaky_since only gives you the latest marked_flaky timestamp, hiding the history. The same applies to quarantine as you lose unquarantined/re-quarantined transitions.

2. They don't scale well. Each new "since" field means another ClickHouse query bolted onto the show response. If we later want to surface unmarked_flaky or unquarantined timestamps, we'd keep adding nullable fields.

3. The data already exists in a richer form. We have a `test_case_events` table in ClickHouse with full event history (marked_flaky, unmarked_flaky, quarantined, unquarantined, first_run), including actor attribution. The LiveView already renders this as a timeline, we're just missing an API endpoint for this.

### Suggested alternative

Instead of adding these fields, I'd expose a paginated test case events endpoint:

`GET /api/projects/:account/:project/tests/test-cases/:id/events`

Server-side implementation:
- Add a show_events action to TestCasesController (or a new TestCaseEventsController)
- Reuse Tests.list_test_case_events/2 which already does paginated ClickHouse queries with Flop (default page size 20, max 100)
- Return event_type, inserted_at (as unix timestamp), and optionally actor info
- Define an OpenAPI schema for the response

CLI-side implementation:
- Add a `tuist test case events <id>` subcommand (or include events in the show output)
- The client can derive any "since" value it needs from the event list. e.g., the most recent marked_flaky event gives you the flaky-since date, but you also get the full timeline

This keeps the show endpoint focused on current state and metrics, and gives the CLI/API the full picture through a dedicated resource.

fortmarek:
cc @pepicrft would like your input on the proposed direction

pepicrft:
Very onboard with the direction. I'd also expose the MCP tool to get the events, for example `get_test_case_events`.

cli/Tests/TuistTestCommandTests/Services/TestCaseEventsCommandServiceTests.swift
```diff
@@ -0,0 +1,196 @@
...
+            testCaseIdentifier: "tc-1",
+            path: nil,
+            page: nil,
+            pageSize: nil,
+            json: true
+        )
+
+        // Then
+        #expect(ui().contains("marked_flaky"))
+        #expect(ui().contains("first_run"))
```
pepicrft: What if we snapshot-test here? I believe we added a library from Pointfree for that. If we inline the snapshot, we can see the exact output, which we can easily review.

pepicrft: If we agree, I'd apply this to the other tests here.

server/lib/tuist/mcp/components/tools/list_test_case_events.ex
```diff
@@ -0,0 +1,75 @@
...
+  @moduledoc """
+  List events for a test case (e.g. marked_flaky, quarantined, first_run). The account_handle and project_handle can be extracted from a Tuist dashboard URL: https://tuist.dev/{account_handle}/{project_handle}.
+  """
+
+  use Tuist.MCP.Tool,
+    name: "list_test_case_events",
+    authorize: [action: :read, category: :test],
+    schema: %{
+      "type" => "object",
+      "properties" => %{
```
pepicrft, line 10: We are a bit inconsistent in this pattern across tools, but I think we should support passing a test case URL tooo, as we do with `list_xcode_build_targets`. That way, you can pass the test case URL to the agent, and they'll figure out how to use this tool.

server/lib/tuist_web/controllers/api/test_cases_controller.ex
```diff
@@ -271,6 +271,140 @@ defmodule TuistWeb.API.TestCasesController do
...
+                     nullable: true,
+                     description: "The user who triggered the event, or null for system events.",
+                     properties: %{
+                       id: %Schema{type: :integer, description: "The actor's account ID."},
+                       name: %Schema{type: :string, description: "The actor's account handle."}
+                     },
+                     required: [:id, :name]
+                   }
+                 },
+                 required: [:event_type, :inserted_at]
```
pepicrft, line 346: Is  `actor` optional?

server/test/tuist/mcp/components/tools/test_tools_test.exs
```diff
@@ -258,6 +259,81 @@ defmodule Tuist.MCP.Components.Tools.TestToolsTest do
...
+      result = JSON.decode!(text)
+      assert length(result["events"]) == 2
+
+      first_event = hd(result["events"])
+      assert first_event["event_type"] == "marked_flaky"
+      assert first_event["actor"]["name"] == "alice"
+
+      second_event = Enum.at(result["events"], 1)
+      assert second_event["event_type"] == "first_run"
+      assert second_event["actor"] == nil
```
pepicrft: For some reason, coding agents default to asserting against every individual element, but I think we should assert the whole response against the value that we expect. I'd apply this in all the tests here.

pepicrft (approved):
Thanks a lot @justine-acorns 🙏🏼

server/lib/tuist_web/controllers/api/test_cases_controller.ex
justine-acorns: Yes, `actor` is nullable here, because It's Nullable in the Clickhouse schema for events! System events (like the first run event for ex) have a null actor ID

pepicrft:
Let's wait for @fortmarek's review

cli/Tests/TuistTestCommandTests/Services/TestCaseEventsCommandServiceTests.swift
justine-acorns: So adding `SnapshotTesting` here causes the linux versions of the tests to fail. What's your preference here? Given that all of the other tests in `TuistTestCommandTests` are not using SnapshotTesting, would it make sense to just revert back to the `#expect`s and let your team decide how you want to go about integrating SnapshotTesting into this test target?

Or if you have an alternative approach, I'm happy to make that change.

fortmarek (approved):
This is great 👏 

Thanks a lot for your patience here 🙂

pepicrft (approved):

PR #10120, opened 2026-03-31
fix(cli): embed App Intents metadata in cached xcframeworks

fortmarek:
## Summary

- When caching static frameworks containing App Intents, the `.appintents` metadata bundle produced by `appintentsmetadataprocessor` was not being included in the cached xcframework
- After creating each xcframework, we now check for a sibling `.appintents` bundle and copy its `Metadata.appintents/` into each framework slice
- This fixes intermittent `AppIntentsSSUTraining` failures when the app target builds with cached xcframeworks

### Root cause

For static frameworks, Apple's build system writes App Intents metadata to `<ProductName>.appintents/` as a **sibling** to the `.framework` in the build products directory (not inside it). This is controlled by `$(ProductResourcesDir)`, an internal Xcode build system variable that resolves differently for static frameworks vs dynamic frameworks/apps. It cannot be overridden via build settings.

During `tuist cache warm`, `copyDerivedDataArtifacts` correctly copies both the `.framework` and `.appintents` sibling to the staging directory (via glob `["*"]`). However, `buildXCFrameworks` only passes the `.framework` to `xcodebuild -create-xcframework`, so the `.appintents` metadata is lost.

### What the fix does

After creating each xcframework, we look for a sibling `.appintents` bundle in the staging directory and copy its `Metadata.appintents/` subdirectory into each framework slice. This is specifically `Metadata.appintents/` (not the outer `<Name>.appintents/` wrapper) because Xcode's `APP_INTENTS_METADATA_PATH` defaults to `$(ProductResourcesDir)/Metadata.appintents` -- that's the path Xcode searches for inside linked frameworks.

The metadata is architecture-neutral (`IsArchitectureNeutral = YES` in Apple's xcspec), so a single source copy works for all slices.

### Build products layout (before fix)

```
Build/Products/Debug-iphonesimulator/
├── IntentsFramework.framework/          # binary, headers, modules
└── IntentsFramework.appintents/         # sibling -- NOT included in xcframework
    └── Metadata.appintents/
        ├── extract.actionsdata
        ├── extract.packagedata
        └── version.json
```

### Cached xcframework layout (after fix)

```
IntentsFramework.xcframework/
├── ios-arm64/
│   └── IntentsFramework.framework/
│       ├── IntentsFramework
│       ├── Metadata.appintents/         # <-- embedded by this PR
│       └── ...
└── ios-arm64_x86_64-simulator/
    └── IntentsFramework.framework/
        ├── IntentsFramework
        ├── Metadata.appintents/         # <-- embedded by this PR
        └── ...
```

## Repro

[appintents-cache-repro.zip](https://github.com/user-attachments/files/26370121/appintents-cache-repro.zip)

Setup:
- `IntentsFramework` -- static framework with `OpenSearchIntent` (AppIntent) and `AppIntentsPackage` conformance
- `AppIntentsCacheRepro` -- app target with `AppShortcutsProvider` referencing `OpenSearchIntent`

### Steps to reproduce (before fix)

1. `tuist generate --no-open`
2. `tuist cache warm`
3. `tuist generate --no-open` (now uses cached xcframework)
4. `xcodebuild clean build -workspace AppIntentsCacheRepro.xcworkspace -scheme AppIntentsCacheRepro -destination "generic/platform=iOS Simulator" CODE_SIGNING_ALLOWED=NO`
5. `AppIntentsSSUTraining` fails intermittently because `Metadata.appintents` is missing from the cached xcframework

### Steps to verify the fix (e2e)

1. Build the Tuist CLI from this branch
2. `tuist cache warm` on the repro project
3. Verify metadata is embedded: `find ~/.cache/tuist/Binaries -name "Metadata.appintents" -type d` should show paths inside `.xcframework/*/*.framework/`
4. `tuist generate --no-open`
5. `xcodebuild clean build -workspace AppIntentsCacheRepro.xcworkspace -scheme AppIntentsCacheRepro -destination "generic/platform=iOS Simulator" CODE_SIGNING_ALLOWED=NO`
6. Build succeeds with `AppIntentsSSUTraining` completing: `Training 'Search on ${+applicationName}' for en`

## Test plan

- [x] Verified end-to-end with the repro project
- [x] `tuist cache warm` embeds `Metadata.appintents/` in both simulator and device xcframework slices
- [x] `tuist generate` + clean app build succeeds with `AppIntentsSSUTraining` passing
- [ ] CI passes

### DIFF
--- a/cli/Sources/TuistKit/Commands/Cache/CacheWarmCommandService.swift
+++ b/cli/Sources/TuistKit/Commands/Cache/CacheWarmCommandService.swift
@@ -503,7 +503,12 @@ import XcodeGraph
                 // as a sibling to the .framework in the build products directory. We copy the
                 // Metadata.appintents directory into each framework slice of the xcframework so
                 // the app target's AppIntentsSSUTraining phase can discover it.
-                let appIntentsBundleName = "\(cacheableTarget.0.target.productName).appintents"
+                // We derive the .appintents name from productNameWithExtension (which respects
+                // PRODUCT_NAME build setting overrides) to stay consistent with how the framework
+                // artifact itself is resolved.
+                let resolvedProductName = cacheableTarget.0.target.productNameWithExtension
+                    .replacingOccurrences(of: ".framework", with: "")
+                let appIntentsBundleName = "\(resolvedProductName).appintents"
                 for artifactDir in Set(platformBinaryArtifacts) {
                     let metadataSource = artifactDir.appending(
                         components: [appIntentsBundleName, "Metadata.appintents"]


cli/Sources/TuistKit/Commands/Cache/CacheWarmCommandService.swift
```diff
@@ -498,6 +498,30 @@ import XcodeGraph
                     output: xcframeworkPath
                 )
 
+                // Embed App Intents metadata if available.
+                // For static frameworks, appintentsmetadataprocessor writes a .appintents bundle
+                // as a sibling to the .framework in the build products directory. We copy the
+                // Metadata.appintents directory into each framework slice of the xcframework so
+                // the app target's AppIntentsSSUTraining phase can discover it.
+                let appIntentsBundleName = "\(cacheableTarget.0.target.productName).appintents"
```
cschmatzler: > The new lookup builds the bundle path from cacheableTarget.0.target.productName, but the framework artifact itself is resolved with productNameWithExtension, which already handles a static PRODUCT_NAME override. For targets renamed via build settings, Xcode emits <resolved-product>.appintents/Metadata.appintents, so this code silently misses the bundle and the cached xcframework is still missing App Intents metadata. That means cached builds keep failing in projects that customize the product name.

fortmarek: Good catch! Fixed in 057b3d7 — now deriving the `.appintents` bundle name from `productNameWithExtension` (which resolves `PRODUCT_NAME` build setting overrides) instead of `productName`. This keeps it consistent with how the framework artifact itself is resolved on line 480.

PR #10127, opened 2026-03-31
feat(server,cli): add parameterized test argument support

fortmarek:
Screenshot 2026-04-08 at 19 01 35

## Summary

- Adds `test_case_run_arguments` table to support parameterized tests (Swift Testing `@Test(arguments:)`)
- Each argument variant gets its own failures, repetitions, and attachments, preserving per-argument grouping from xcresult
- Full-stack implementation: ClickHouse migrations, Elixir schemas, API, CLI parsing, dashboard UI
- Fixes Swift Testing identifier normalization (backtick-quoted names + parameter signatures)
- Non-parameterized tests continue to work unchanged

### Why no Gradle plugin changes?

Gradle drops the method name from `@ParameterizedTest` invocations in `afterTest` descriptors ([gradle/gradle#23324](https://github.com/gradle/gradle/issues/23324)), reporting names like `[1] value=hello` instead of `testMethod[1] value=hello`. Without the method name prefix, we cannot safely group invocations back into their parent test method — multiple `@ParameterizedTest` methods with the same index would be incorrectly merged. Until Gradle fixes this, each parameterized invocation appears as a separate test case, which matches Gradle's own reporting behavior.

## Test plan
- [ ] Upload an xcresult with parameterized Swift Testing tests and verify per-argument grouping in the dashboard
- [ ] Upload a non-parameterized test result and verify existing behavior is unchanged
- [ ] Verify collapsible argument sections work on both test run and test case run pages
- [ ] Run `mix test` for server-side tests
- [ ] Build CLI with xcodebuild and verify it compiles

### DIFF
--- a/xcode_processor/native/xcresult_nif/Sources/XCResultParser/TestCase.swift
+++ b/xcode_processor/native/xcresult_nif/Sources/XCResultParser/TestCase.swift
@@ -14,7 +14,7 @@ public struct TestCase: Encodable, Sendable {
     public var isQuarantined: Bool
 
     enum CodingKeys: String, CodingKey {
-        case name, module, duration, status, failures, repetitions, attachments
+        case name, module, duration, status, failures, repetitions, arguments, attachments
         case testSuite = "test_suite_name"
         case crashReport = "crash_report"
         case isQuarantined = "is_quarantined"
--- a/cli/Sources/TuistServer/OpenAPI/server.yml
+++ b/cli/Sources/TuistServer/OpenAPI/server.yml
@@ -1904,6 +1904,12 @@ components:
           type: integer
           x-struct:
           x-validate:
+        test_case_run_argument_id:
+          description: The UUID of the argument this attachment belongs to (for parameterized tests).
+          nullable: true
+          type: string
+          x-struct:
+          x-validate:
         test_case_run_id:
           description: The UUID of the test case run.
           type: string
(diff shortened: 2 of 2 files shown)


xcode_processor/native/xcresult_nif/Sources/XCResultParser/TestCase.swift
```diff
@@ -8,6 +8,7 @@ public struct TestCase: Encodable, Sendable {
     public let status: TestStatus
     public let failures: [TestCaseFailure]
     public let repetitions: [TestCaseRepetition]
+    public let arguments: [TestCaseArgument]
```
pepicrft, line 11: Adding `arguments` here is not enough for the async xcresult path. `XCResultNIF` serializes `TestCase` with `JSONEncoder`, but `CodingKeys` still omits `arguments`, so `uploadResultBundle` and xcode-processor processing drop parameterized argument data before the server persists it.

server/lib/tuist/tests.ex
```diff
@@ -1174,20 +1171,92 @@ defmodule Tuist.Tests do
...
+        TestCaseRunArgument.Buffer.insert_all(all_arguments)
+      end
     end)
 
     create_first_run_events(test_case_runs, new_test_case_ids)
 
     {test_case_ids_with_flaky_run, test_case_runs}
   end
 
+  defp build_argument_records(arguments, test_case_run_id) do
```
pepicrft, line 1185: I think we are still losing per-argument attachments in the result-bundle path. `build_argument_records/2` creates argument rows, failures, and repetitions, but attachments tagged with `argument_name` still fall through to `build_attachments/3`, which only inserts root attachments. That means async xcresult uploads cannot populate `test_case_run_argument_id`.

cli/Sources/TuistServer/OpenAPI/Types.swift
```diff
@@ -4815,6 +4815,10 @@ public enum Components {
             ///
             /// - Remark: Generated from `#/components/schemas/TestCaseRunAttachmentParams/repetition_number`.
             public var repetition_number: Swift.Int?
+            /// The UUID of the argument this attachment belongs to (for parameterized tests).
```
pepicrft, line 4818: This looks like generated client code without the matching source-of-truth spec update. `server.yml` still omits `test_case_run_argument_id`, and `RunsTest.test_case_runs[].arguments` too, but `Types.swift` now includes both. Could we regenerate and commit `server.yml` as well so the checked-in contract stays in sync?

server/lib/tuist_web/controllers/api/test_case_run_attachments_controller.ex
```diff
@@ -193,9 +198,15 @@ defmodule TuistWeb.API.TestCaseRunAttachmentsController do
...
         },
         fn attrs ->
-          case Map.get(body_params, :repetition_number) do
+          attrs =
+            case Map.get(body_params, :repetition_number) do
+              nil -> attrs
+              repetition_number -> Map.put(attrs, :repetition_number, repetition_number)
+            end
+
+          case Map.get(body_params, :test_case_run_argument_id) do
```
pepicrft, line 207: Should we validate that `test_case_run_argument_id` belongs to the provided `test_case_run_id` before inserting? Right now any UUID from the request body is persisted if the caller is authorized for the project.

fortmarek:
Addressed review feedback in 1c8ee56:

1. **CodingKeys**: Added `arguments` to `TestCase.CodingKeys` so JSONEncoder includes argument data in the async xcresult path.

2. **Per-argument attachments in result-bundle path**: Known limitation. The async xcresult path (`build_attachments`) doesn't have per-argument context. The sync CLI path handles this correctly via `test_case_run_argument_id` on the attachment upload API. Follow-up for async path.

3. **server.yml**: Regenerated and committed. `Types.swift` was already in sync.

4. **argument_id validation**: The ID comes from the server's own `createTest` response, so it's trusted in the normal flow. This is consistent with how `test_case_run_id` and `repetition_number` aren't cross-validated either. Happy to add validation if the risk warrants it.

PR #10141, opened 2026-04-01
chore(server): Reduce PromEx HTTP cardinality in server and cache

pepicrft:
## Summary
- replace the default Phoenix PromEx plugin in `server` and `cache` with a shared Tuist plugin that can drop redundant HTTP labels per service
- `server`: keep `method`, normalized `path`, exact `status`, and scrape `instance`; drop `host`, `controller`, and `action`
- `cache`: keep `method`, normalized `path`, `status_class`, and scrape `instance`; drop `host`, `controller`, and `action`
- lower non-production cache scrape cadence in `cache/platform/alloy.nix`
- disable unix exporter self-metrics on cache nodes

## Why
- `prometheus.scrape.tuist` churn is being driven by Phoenix HTTP label combinations. In this plugin, `path` is already the normalized Phoenix route template, so it captures the endpoint identity without needing `controller` and `action` as separate labels.
- `host` is redundant with scrape `instance` for both services. On `server` it mainly reflects request domains; on `cache` it is not the cache node identity.
- `prometheus.scrape.cache_promex` gets its biggest direct win from replacing exact status codes with `status_class`.

## Metric Examples
These label changes apply to the Phoenix HTTP metric families, including:
- `tuist_prom_ex_phoenix_http_request_duration_milliseconds_bucket`
- `tuist_prom_ex_phoenix_http_response_size_bytes_bucket`
- `tuist_prom_ex_phoenix_http_requests_total`
- `cache_prom_ex_phoenix_http_request_duration_milliseconds_bucket`
- `cache_prom_ex_phoenix_http_response_size_bytes_bucket`
- `cache_prom_ex_phoenix_http_requests_total`

Server before:
```promql
tuist_prom_ex_phoenix_http_request_duration_milliseconds_bucket{status="200",method="GET",path="/:account_handle/:project_handle/runs/:id",controller="TuistWeb.RunController",action="show",host="tuist.dev",instance="production",le="500"}
```

Server after:
```promql
tuist_prom_ex_phoenix_http_request_duration_milliseconds_bucket{status="200",method="GET",path="/:account_handle/:project_handle/runs/:id",instance="production",le="500"}
```

Server alternatives retained:
- `instance` remains the deployment identity instead of `host`
- `path` remains the normalized route template instead of `controller` + `action`

Cache before:
```promql
cache_prom_ex_phoenix_http_request_duration_milliseconds_bucket{status="401",method="GET",path="/api/cache/cas/:account_handle/:project_handle/:hash",controller="CacheWeb.ArtifactsController",action="show",host="cache_upstream",instance="fra1-cache-production-1",le="500"}
```

Cache after:
```promql
cache_prom_ex_phoenix_http_request_duration_milliseconds_bucket{status_class="4xx",method="GET",path="/api/cache/cas/:account_handle/:project_handle/:hash",instance="fra1-cache-production-1",le="500"}
```

Cache alternatives retained:
- `instance` remains the cache node identity instead of `host`
- `path` remains the normalized route template instead of `controller` + `action`
- `status_class` replaces exact `status` while preserving operational groupings

## Validation
- `mix test test/tuist_common/prom_ex_phoenix_plugin_test.exs` in `tuist_common/`
- `mix compile` in `server/`
- `mix compile` in `cache/`

### DIFF
--- a/tuist_common/lib/tuist_common/prom_ex_phoenix_plugin.ex
+++ b/tuist_common/lib/tuist_common/prom_ex_phoenix_plugin.ex
@@ -1,455 +1,442 @@
-if Code.ensure_loaded?(Phoenix) do
-  defmodule TuistCommon.PromExPhoenixPlugin do
-    @moduledoc """
-    Shared Phoenix PromEx plugin for Tuist services.
+defmodule TuistCommon.PromExPhoenixPlugin do
+  @moduledoc """
+  Shared Phoenix PromEx plugin for Tuist services.
 
-    It mirrors `PromEx.Plugins.Phoenix` but excludes the `host` label from HTTP
-    metrics by default to avoid high-cardinality series from arbitrary request
-    hosts and preview/custom domains.
-    """
+  It mirrors `PromEx.Plugins.Phoenix` but excludes the `host` label from HTTP
+  metrics by default to avoid high-cardinality series from arbitrary request
+  hosts and preview/custom domains.
+  """
 
-    use PromEx.Plugin
+  use PromEx.Plugin
 
-    require Logger
+  require Logger
 
-    alias Phoenix.Socket
-    alias Plug.Conn
-    alias PromEx.Utils
+  alias Phoenix.Socket
+  alias Plug.Conn
+  alias PromEx.Utils
 
-    @stop_event [:prom_ex, :plugin, :phoenix, :stop]
-    @init_event [:phoenix, :endpoint, :init]
+  @stop_event [:prom_ex, :plugin, :phoenix, :stop]
+  @init_event [:phoenix, :endpoint, :init]
 
-    @impl true
-    def event_metrics(opts) do
-      otp_app = Keyword.fetch!(opts, :otp_app)
-      metric_prefix = Keyword.get(opts, :metric_prefix, PromEx.metric_prefix(otp_app, :phoenix))
-      phoenix_event_prefixes = fetch_event_prefixes!(opts)
-      duration_unit = Keyword.get(opts, :duration_unit, :millisecond)
-      normalize_event_name = Keyword.get(opts, :normalize_event_name, fn event -> event end)
+  @impl true
+  def event_metrics(opts) do
+    otp_app = Keyword.fetch!(opts, :otp_app)
+    metric_prefix = Keyword.get(opts, :metric_prefix, PromEx.metric_prefix(otp_app, :phoenix))
+    phoenix_event_prefixes = fetch_event_prefixes!(opts)
+    duration_unit = Keyword.get(opts, :duration_unit, :millisecond)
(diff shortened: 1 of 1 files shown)


pepicrft:
@cschmatzler would the loss of precision with the `status` to `status_class` have an impact in our dashboards, alerts, or ability to dig into an issue?
Also, are we using the data from the unix exporter?

@tuist/engineering alternatively, Grafana has some features to drop metrics adaptively in case we want to keep some of the precision without our costs going up that much.

tuist_common/lib/tuist_common/prom_ex_phoenix_plugin.ex
```diff
@@ -0,0 +1,455 @@
+if Code.ensure_loaded?(Phoenix) do
+  defmodule TuistCommon.PromExPhoenixPlugin do
+    @moduledoc """
+    Shared Phoenix PromEx plugin for Tuist services.
+
+    It mirrors `PromEx.Plugins.Phoenix` but excludes the `host` label from HTTP
+    metrics by default to avoid high-cardinality series from arbitrary request
```
fortmarek: it'd be great if the promex plugin allowed to configure this. I wonder if doing a fork, making that configurable, and putting up a PR would be a better approach than copy-pasting it here.

tuist_common/lib/tuist_common/prom_ex_phoenix_plugin.ex
```diff
@@ -0,0 +1,455 @@
+if Code.ensure_loaded?(Phoenix) do
```
cschmatzler: I think for our own vendored version we can strip this check.

tuist_common/lib/tuist_common/prom_ex_phoenix_plugin.ex
pepicrft: I'll open a PR against the trunk repository making these things configurable.

tuist_common/lib/tuist_common/prom_ex_phoenix_plugin.ex
pepicrft: Removed in `2d0bf3abf3`. Since this is our vendored copy and both consumers already depend on Phoenix, the plugin now assumes Phoenix is present and no longer carries the fallback wrapper.

cache/lib/cache/prom_ex.ex
```diff
@@ -12,9 +12,11 @@ defmodule Cache.PromEx do
   def plugins do
     [
       PromEx.Plugins.Beam,
-      {PromEx.Plugins.Phoenix,
+      {TuistCommon.PromExPhoenixPlugin,
```
pepicrft, line 15: I opened a [PR](https://github.com/akoutmos/prom_ex/pull/284) against `:prom_ex` to land this there instead of maintaining our copy of the plugin.

PR #10144, opened 2026-04-01
feat(cli): support shared volumes for test shard distribution

fortmarek:
## Summary

Adds shared volume support to test sharding, allowing the `.xctestproducts` bundle to be passed between build and test runners via a shared filesystem instead of uploading/downloading through S3. This eliminates the bundle transfer overhead (which can be significant for large test bundles).

### Changes

**CLI (`tuist xcodebuild`):**
- `build-for-testing`: adds `--shard-skip-upload` flag to skip S3 upload while still creating the shard plan on the server
- `test-without-building`: when `-testProductsPath` is passed alongside `--shard-index`, uses the local path instead of downloading from S3. Writes a filtered `.xctestrun` (as `.tuist-shard-<index>.xctestrun`) next to the original so `__TESTROOT__` paths resolve correctly, then passes only `-xctestrun` to xcodebuild (xcodebuild rejects `-testProductsPath` and `-xctestrun` together)

**CLI (`tuist test`):**
- Adds `--shard-skip-upload` flag (build phase)
- Adds `--shard-index` as a CLI flag (previously only available via `TUIST_SHARD_INDEX` env var)
- When `-testProductsPath` is in passthrough args during shard execution, uses the local path instead of downloading from S3

**Docs:**
- Adds "Shared volumes" section to both Xcode and generated projects test sharding pages
- Includes Namespace runners example with unique paths per run and cleanup

### Usage

```yaml
# GitHub Actions with Namespace volumes
env:
  TEST_PRODUCTS_PATH: /Volumes/test-products/${{ github.run_id }}/MyScheme.xctestproducts

jobs:
  build:
    runs-on: namespace-profile-default-macos
    volumes:
      - name: test-products
        path: /Volumes/test-products
    steps:
      - run: |
          tuist xcodebuild build-for-testing \
            --shard-total 5 --shard-skip-upload \
            -scheme MyScheme \
            -testProductsPath $TEST_PRODUCTS_PATH

  test:
    needs: build
    runs-on: namespace-profile-default-macos
    volumes:
      - name: test-products
        path: /Volumes/test-products
    env:
      TUIST_SHARD_INDEX: ${{ matrix.shard }}
    steps:
      - run: |
          tuist xcodebuild test \
            -scheme MyScheme \
            -testProductsPath $TEST_PRODUCTS_PATH
      - if: always()
        run: rm -rf /Volumes/test-products/${{ github.run_id }}
```

## Test plan

- [x] Unit tests for ShardService volume mode (skip download, filter xctestrun, error on missing xctestrun)
- [x] Acceptance test for `tuist xcodebuild` shard flow with local test products (hits canary, passes locally)
- [x] Acceptance test for `tuist test` shard flow with local test products (hits canary, passes locally)
- [x] Lint passes

### DIFF
--- a/cli/Sources/TuistKit/Services/Sharding/ShardService.swift
+++ b/cli/Sources/TuistKit/Services/Sharding/ShardService.swift
@@ -120,7 +120,7 @@ public struct ShardService: ShardServicing {
         guard let xcTestRunSourcePath = try await fileSystem
             .glob(directory: resolvedTestProductsPath, include: ["**/*.xctestrun"])
             .collect()
-            .first
+            .first(where: { !$0.basename.contains(".tuist-shard-") })
         else {
             throw ShardServiceError.xcTestRunNotFound(resolvedTestProductsPath)
         }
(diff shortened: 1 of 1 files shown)


server/priv/docs/en/guides/features/test-sharding/generated-projects.md
fortmarek, line 1: this and the xcode page for sharding would benefit from our tab component, worth setting that up once we do another pass of the docs content cc @pepicrft

pepicrft: Yeah, it'll be much better. I'll take a look at it.

PR #10145, opened 2026-04-01
feat(cli): add --inspect-mode flag for remote xcresult processing

fortmarek:
## Summary
- Adds `--inspect-mode {local|remote}` flag to `tuist test`, `tuist xcodebuild test`, and `tuist xcodebuild test-without-building` (env: `TUIST_INSPECT_TEST_MODE`, shared with `tuist inspect test --mode`)
- **Local mode** (default): parses the xcresult locally and uploads the summary (existing behavior)
- **Remote mode**: uploads the raw xcresult for server-side processing via xcode_processor, skipping expensive local parsing (action logs, attachments, crash reports, durations)
- Embeds `quarantined_tests.json` into the xcresult directory before remote upload so the xcode_processor applies the same quarantine list the CLI had at test time
- Adds lightweight `parseTestStatuses` method that only extracts test name/module/status for quarantine and selective testing checks, skipping full xcresult parsing
- Extracts shared remote upload logic into `UploadResultBundleService.uploadResultBundle` (renamed from `uploadResultBundleRemotely`), old local upload renamed to `uploadTestSummary`
- Moves `onlyQuarantinedTestsFailed(testStatuses:quarantinedTests:)` into `TestQuarantineService` for reuse across commands

### CLI changes
- `TestRunCommand`, `XcodeBuildTestCommand`, `XcodeBuildTestWithoutBuildingCommand`: new `--inspect-mode` option
- `TestService`, `XcodeBuildTestCommandService`: mode-aware upload and quarantine handling
- `UploadResultBundleService`: writes quarantine JSON into xcresult, handles remote upload
- `InspectTestCommandService`: simplified to use shared `uploadResultBundle`
- `TestResultStatuses`: new lightweight struct for quarantine/selective testing checks

### xcode_processor changes
- Reads `quarantined_tests.json` from extracted xcresult
- Applies `is_quarantined` marking to parsed test cases matching the quarantine list

## Feedback requested

**Flag naming**: The flag is currently `--inspect-mode` (shared env var `TUIST_INSPECT_TEST_MODE` with `tuist inspect test --mode`). However, on `tuist test` the name "inspect" may be confusing since "inspect" is a separate command. Alternatives to consider:
- `--processing-mode` -- describes what it controls (where parsing happens)
- `--upload-mode` -- describes the upload behavior
- Keep `--inspect-mode` -- consistent with the `tuist inspect test` command

Would appreciate feedback on which name feels most intuitive for users.

## Test plan
- [ ] `tuist test --inspect-mode local` behaves identically to current behavior
- [ ] `tuist test --inspect-mode remote` uploads raw xcresult, server processes it
- [ ] `TUIST_INSPECT_TEST_MODE=remote tuist test` works via env var
- [ ] `tuist xcodebuild test --inspect-mode remote` works
- [ ] Quarantine suppression works in remote mode (only-quarantined failures don't fail the run)
- [ ] `quarantined_tests.json` is written into xcresult and read by xcode_processor
- [ ] Unit tests pass for TestResultStatuses, TestQuarantineService, UploadResultBundleService, XCResultProcessor

### DIFF
--- a/cli/Tests/TuistKitTests/Services/TestServiceTests.swift
+++ b/cli/Tests/TuistKitTests/Services/TestServiceTests.swift
@@ -1386,8 +1386,12 @@ final class TestServiceTests: TuistUnitTestCase {
                     ]
                 )
             )
+        xcResultService.reset()
         given(xcResultService)
-            .parseTestStatuses(path: .value(xcresultPath))
+            .parse(path: .any, rootDirectory: .any)
+            .willReturn(nil)
+        given(xcResultService)
+            .parseTestStatuses(path: .any)
             .willReturn(
                 TestResultStatuses(testCases: [
                     .init(name: "testA", testSuite: nil, module: "FrameworkATests", status: .failed),


cli/Sources/TuistKit/Services/InspectResultBundleService.swift
```diff
@@ -210,6 +283,21 @@ public struct UploadResultBundleService: UploadResultBundleServicing {
...
+        to resultBundlePath: AbsolutePath
+    ) async throws {
+        let entries = quarantinedTests.map { test in
+            QuarantinedTestEntry(
+                target: test.target,
+                class: test.class,
+                method: test.method
+            )
+        }
+        let filePath = resultBundlePath.appending(component: "quarantined_tests.json")
```
fortmarek, line 297: the xcode processor needs to know about which tests were quarantined when the client was running the tests. Instead of passing that through the API, decided to include that in the `.xcresult`, so the xcode processor has all information it needs in the `.xcresult` artifact itself.

cli/Sources/TuistKit/Commands/XcodeBuild/XcodeBuildTestWithoutBuildingCommand.swift
```diff
@@ -33,12 +33,20 @@ public struct XcodeBuildTestWithoutBuildingCommand: AsyncParsableCommand, Tracka
     @Option(name: .long, help: "The zero-based shard index to execute.")
     var shardIndex: Int?
 
+    @Option(
```
pepicrft, line 36: What about surfacing this somewhere in the docs. As a user, it's not obvious when I'd opt into server-side processing and what the implications of that decision are.

xcode_processor/test/xcode_processor/xcresult_processor_test.exs
```diff
@@ -78,6 +78,95 @@ defmodule XcodeProcessor.XCResultProcessorTest do
...
     end
 
+    test "applies quarantine marking from quarantined_tests.json" do
+      temp_dir =
+        Path.join(
+          System.tmp_dir!(),
+          "xcresult_quarantine_test_#{:erlang.unique_integer([:positive])}"
+        )
+
+      File.mkdir_p!(temp_dir)
```
pepicrft: Maybe for a Credo rule, but we should lean on using `ExUnit`'s `tmp_dir` tag so they take care of the lifecycle for us.

cli/Sources/TuistKit/Commands/XcodeBuild/XcodeBuildTestWithoutBuildingCommand.swift
fortmarek: > What about surfacing this somewhere in the docs. As a user, it's not obvious when I'd opt into server-side processing and what the implications of that decision are.

Agree, this is coming along with switching the default as this will be a requirement for teams using self-hosting (unless they want to run the xcode processor node)

PR #10148, opened 2026-04-01
fix: circular dependency when SPM target wraps external product of same name

### LINKED ISSUE #8319: AgoraRtcKit can't install as found circular dependencies
### What happened?

When using AgoraRtcKit (https://github.com/AgoraIO/AgoraRtcEngine_iOS/blob/main/Package.swift)

`tuist generate` command prints this error.

```
Found circular dependency between targets: /Users/hiroshi.kimura/.ghq/github.com/eure/swift-playground/projects/AgoraPlay/AgoraPlay/Tuist/.build/checkouts/AgoraRtcEngine_iOS:AgoraInfra_iOS -> /Users/hiroshi.kimura/.ghq/github.com/eure/swift-playground/projects/AgoraPlay/AgoraPlay/Tuist/.build/checkouts/AgoraRtcEngine_iOS:AgoraInfra_iOS
```

Here is a reproduction project.
`tuist install` then
`tuist generate` fails
[AgoraPlay.zip](https://github.com/user-attachments/files/22561934/AgoraPlay.zip)

### How do we reproduce it?

Use attached zip

`tuist install` then
`tuist generate` fails

### Error log

```
Found circular dependency between targets: /Users/hiroshi.kimura/.ghq/github.com/eure/swift-playground/projects/AgoraPlay/AgoraPlay/Tuist/.build/checkouts/AgoraRtcEngine_iOS:AgoraInfra_iOS -> /Users/hiroshi.kimura/.ghq/github.com/eure/swift-playground/projects/AgoraPlay/AgoraPlay/Tuist/.build/checkouts/AgoraRtcEngine_iOS:AgoraInfra_iOS
```

### macOS version

26

### Tuist version

4.79.7

### Xcode version

26.0.0

vijaytholpadi:
## Summary
- When an SPM package target has the same name as an external `.product` dependency, Tuist incorrectly resolves it as a local target reference, creating a self-referential loop
- The `.product`'s `package:` parameter was being discarded (`_`), so `mapDependency` couldn't distinguish local vs external dependencies
- Passes `package:` through as a `targetPackage` parameter; when it differs from the current package name, resolves as `.external` instead of `.target`

Fixes #8319

## Changes
**`cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift`**
- Capture `package` instead of `_` in the `.product` case
- Add `targetPackage: String? = nil` parameter to `mapDependency` (backward compatible)
- Add early return for cross-package products to resolve as `.external`

## Test plan
- [x] Added unit test `map_whenProductDependencyHasSameNameAsLocalTarget_mapsToExternalDependency`
- [x] Test passes locally via `xcodebuild test-without-building`
- [ ] CI passes


cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift
```diff
@@ -731,6 +735,17 @@ public struct PackageInfoMapper: PackageInfoMapping {
         } catch {
             return nil
         }
+        // If this is a .product dependency that explicitly references a different package
```
fortmarek, line 738: Ran this with agentic review and got the following:
```
 2. The fix is narrower than the underlying bug
  The root issue is that .product(name:, package:) with a package referencing any other package is being
  matched against local targets. The current fix only handles the case sourceTargetName == name. The
  following still produces an incorrect resolution:
  - Source target Bar depends on .product(name: "Foo", package: "ExternalFoo")
  - The local package also defines a target Foo (unrelated)
  - Result: still resolves as a local .target(name: "Foo") instead of an external dependency.

  A more correct fix would be: whenever targetPackage != nil && targetPackage != packageInfo.name, resolve
   as .external. That's the SPM semantic — .product(package:) is explicitly cross-package. The narrow
  scope is defensible for a bug fix, but it's worth either (a) widening the fix or (b) calling out the
  limitation in the PR description and filing a follow-up
```

is this a false positive? wdyt?

vijaytholpadi: This is actually not a false positive, but would require updating a tuist-specific behavior we assert in our tests.

The broader fix (`targetPackage != nil && targetPackage != packageInfo.name` resolving as `.external`) is semantically more correct since `.product(package:)` in SPM explicitly means a cross-package dependency.

The reason I narrowed it was that the existing test `map_whenHasModuleAliases` failed with the broader approach. That test has a `.product(name: "Product", package: "Package2")` where the local package also has a target named "Product", and it expects the dependency to resolve as `.target(name: "Package2Product")` (local, with alias applied) rather than `.external`. This is a Tuist-specific behavior where all packages get mapped into one workspace project, so cross-package products are treated as local targets.

If you're okay with changing that behavior so that cross-package `.product` dependencies always resolve as `.external` (which aligns with SPM semantics), I can widen the fix and update the affected test accordingly. Just want to make sure that change is intentional since it touches how module aliases interact with cross-package resolution.

fortmarek: thanks for the comment @vijaytholpadi! Let's keep the changes simple for now and we can make the larger changes proposed here if needed.

PR #10151, opened 2026-04-01
feat(cli): add defaultSwiftVersion generation option and respect package-declared Swift versions

pepicrft:
## Summary
- Adds a `defaultSwiftVersion` generation option (defaults to `"5"`) so users can control the `SWIFT_VERSION` build setting applied to local targets.
- External SPM packages now always get their `SWIFT_VERSION` resolved correctly by `PackageInfoMapper`, matching SPM's behavior:
  - If the package declares `swiftLanguageVersions` → uses the highest compatible version
  - If not declared → derives from the package's `swift-tools-version` (e.g. `5.x` → `"5"`, `6.x` → `"6"`)
- `DefaultSettingsProvider` no longer injects `SWIFT_VERSION` into remote/package targets at all, since `PackageInfoMapper` always handles it.

## Usage

```swift
// Tuist.swift
let tuist = Tuist(
    generationOptions: .options(defaultSwiftVersion: "6")
)
```

## Context
Reported by users in Slack: Tuist injects `SWIFT_VERSION = "5.0"` as an inline build setting on all targets (including external packages). Since inline settings take priority over xcconfig values, this silently overrides user-configured Swift versions.

For external packages, SPM derives the Swift language version from the package's `swift-tools-version` when `swiftLanguageVersions` is not explicitly declared (e.g. `// swift-tools-version: 5.9` implies Swift 5 language mode). Tuist was not doing this — it was overriding all packages with its own default. Now Tuist matches SPM's behavior.

## Test plan
- [x] Added test: remote targets no longer get a default `SWIFT_VERSION` injected by `DefaultSettingsProvider`
- [x] Updated `PackageInfoMapper` tests to verify `SWIFT_VERSION` is always set from tools-version fallback
- [x] Existing tests for explicit `swiftLanguageVersions` still pass
- [ ] Verify with a project that has external SPM dependencies (with and without `swiftLanguageVersions` declared)

### DIFF
--- a/cli/Tests/TuistGeneratorTests/Generator/ConfigGeneratorTests.swift
+++ b/cli/Tests/TuistGeneratorTests/Generator/ConfigGeneratorTests.swift
@@ -65,7 +65,7 @@ struct ConfigGeneratorTests {
             "INFOPLIST_FILE": "$(SRCROOT)/Info.plist",
             "PRODUCT_BUNDLE_IDENTIFIER": "com.test.bundle_id",
             "CODE_SIGN_ENTITLEMENTS": "$(SRCROOT)/Test.entitlements",
-            "SWIFT_VERSION": "5.0",
+            "SWIFT_VERSION": "5",
         ]
 
         let debugSettings: SettingsDictionary = [
--- a/cli/Tests/TuistKitTests/Commands/DumpServiceIntegrationTests.swift
+++ b/cli/Tests/TuistKitTests/Commands/DumpServiceIntegrationTests.swift
@@ -227,6 +227,7 @@ final class DumpServiceTests: TuistTestCase {
 
                     ],
                     "buildInsightsDisabled": false,
+                    "defaultSwiftVersion": "5",
                     "disablePackageVersionLocking": false,
                     "disableSandbox": true,
                     "enableCaching": false,


cli/Sources/ProjectDescription/ConfigGenerationOptions.swift
```diff
@@ -119,7 +123,8 @@ extension Tuist {
             enableCaching: Bool = false,
             registryEnabled: Bool = false,
             additionalPackageResolutionArguments: [String] = [],
-            warningsAsErrors: WarningsAsErrors = .none
+            warningsAsErrors: WarningsAsErrors = .none,
+            defaultSwiftVersion: String = "5"
```
pepicrft, line 130: I decided to keep the current default to prevent this change from breaking.

cli/Sources/ProjectDescription/ConfigGenerationOptions.swift
```diff
@@ -107,6 +107,10 @@ extension Tuist {
         /// Controls which generation warnings are promoted to errors, causing `tuist generate` to fail.
         public var warningsAsErrors: WarningsAsErrors
 
+        /// The default `SWIFT_VERSION` build setting applied to targets that don't specify one.
+        /// For example, set to `"6"` to opt into Swift 6 language mode by default.
```
fortmarek: the version naming is a bit confusing since it's really a lang mode, maybe worth expanding on here

cli/Sources/TuistConfig/TuistGeneratedProjectOptions.swift
```diff
@@ -49,7 +49,8 @@ public struct TuistGeneratedProjectOptions: Equatable, Hashable {
                 includeGenerateScheme: false,
                 enableCaching: false,
                 registryEnabled: false,
-                warningsAsErrors: .none
+                warningsAsErrors: .none,
+                defaultSwiftVersion: "5"
```
fortmarek: this is repeated in many places, maybe worth putting into constants

PR #9670, opened 2026-03-02
feat(server): add mcp OAuth scope for read-only MCP access

pepicrft:
## Summary

This adds a new `mcp` OAuth scope group that grants read-only access to all project resources. When MCP clients (like Claude Desktop) connect via OAuth, the token is now scoped to read-only permissions instead of having full user access. The scope is registered as a Boruta public scope so dynamic OAuth clients can request it, and it's advertised in the well-known metadata endpoints.

The authorize controller now filters OAuth scopes, allowing `mcp` through while stripping unknown ones. The TokenGenerator generates a scoped `AuthenticatedAccount` JWT (with `all_projects: true`) instead of a full-permission User JWT when the `mcp` scope is present.

## Test plan

- [x] TokenGenerator tests for scoped MCP token generation
- [x] Authorize controller tests for scope filtering
- [x] Well-known metadata tests for scopes_supported
- [ ] Manual test with Claude Desktop against staging

PR #10159, opened 2026-04-01
fix(server): restore backward compatibility for iOS app project listing

fortmarek:
## Summary
- Restores the deprecated `token` field (as empty string) in the project API response for backward compatibility with older iOS app clients (v0.24.3 and earlier)
- Changes the OAuth token generator to mint User tokens instead of AuthenticatedAccount tokens, so users can see projects from all their organizations in the app

## Root cause

Two server-side regressions broke the Tuist iOS app's project listing:

1. **Missing `token` field** (7ed51ea0c1 "Merge commit from fork" / "remove deprecated project token from project responses"): The `token` field was removed from the project API response, but older app clients declare it as a required `Codable` field. When the server omits it, `JSONDecoder` throws `keyNotFound` and the app shows an error instead of the project list.

2. **OAuth tokens resolving as AuthenticatedAccount** (tuist/tuist#9670): The token generator was changed to mint `AuthenticatedAccount` JWTs with `type: "account"`. This causes `list_accessible_projects` to only return projects under the token's single account, hiding all organization projects. The iOS app (and MCP) need User-level access to see projects across all orgs. See also tuist/tuist#10156 which fixes the same issue scoped to MCP.

## Test plan
- [x] Unit tests for `token` field backward compatibility in `projects_controller_test.exs`
- [x] Unit tests verifying OAuth tokens resolve to `User` and can access org projects in `token_generator_test.exs`
- [x] E2E tested by building and running the `app@0.24.3` (latest App Store release) on an iOS simulator against a local server, confirming the project list works after the fix

### DIFF
--- a/server/lib/tuist/oauth/token_generator.ex
+++ b/server/lib/tuist/oauth/token_generator.ex
@@ -49,6 +49,7 @@ defmodule Tuist.OAuth.TokenGenerator do
   end
 
   @default_user_scopes [
+    "project:admin:read",
     "project:cache:read",
     "project:cache:write",
     "project:previews:read",
--- a/server/lib/tuist/authorization.ex
+++ b/server/lib/tuist/authorization.ex
@@ -267,6 +267,9 @@ defmodule Tuist.Authorization do
 
       desc("Allows users with ops access to read any projects.")
       allow([:authenticated_as_user, :ops_access])
+
+      desc("Allows an account token with project:admin:read scope to read projects.")
+      allow([:authenticated_as_account, scopes_permit: "project:admin:read"])
     end
 
     action :update do
@@ -332,6 +335,9 @@ defmodule Tuist.Authorization do
 
       desc("Allows users with ops access to read any projects.")
       allow([:authenticated_as_user, :ops_access])
+
+      desc("Allows an account token with project:admin:read scope to read projects.")
+      allow([:authenticated_as_account, scopes_permit: "project:admin:read"])
     end
   end
 
--- a/server/lib/tuist/authorization/checks.ex
+++ b/server/lib/tuist/authorization/checks.ex
@@ -18,6 +18,7 @@ defmodule Tuist.Authorization.Checks do
       "project:runs:write"
     ],
     "mcp" => [
+      "project:admin:read",
       "project:cache:read",
       "project:previews:read",
       "project:bundles:read",
--- a/server/test/tuist/oauth/token_generator_test.exs
+++ b/server/test/tuist/oauth/token_generator_test.exs
@@ -145,6 +145,7 @@ defmodule Tuist.OAuth.TokenGeneratorTest do
       {:ok, claims} = Tuist.Guardian.decode_and_verify(jwt_token)
 
       assert claims["scopes"] == [
+               "project:admin:read",
                "project:cache:read",
                "project:cache:write",
                "project:previews:read",
@@ -172,6 +173,7 @@ defmodule Tuist.OAuth.TokenGeneratorTest do
       {:ok, claims} = Tuist.Guardian.decode_and_verify(jwt_token)
 
       assert claims["scopes"] == [
+               "project:admin:read",
                "project:cache:read",
                "project:cache:write",
                "project:previews:read",


cschmatzler (approved):

server/lib/tuist/oauth/token_generator.ex
```diff
@@ -15,7 +14,7 @@ defmodule Tuist.OAuth.TokenGenerator do
   alias Tuist.Repo
 
   @impl TokenGenerator
-  def generate(token_type, %Boruta.Ecto.Token{sub: sub, client_id: client_id, scope: scope}) do
+  def generate(token_type, %Boruta.Ecto.Token{sub: sub, client_id: client_id}) do
```
cschmatzler: [P1] This drops the Boruta `scope` from the JWT and no longer marks OAuth tokens as `"type" => "account"`, so they decode as `%User{}` instead of `%AuthenticatedAccount{}`. The published OAuth metadata still advertises the `mcp` scope, but that scope is only enforced in the `AuthenticatedAccount`-only `scopes_permit/3` checks. In practice, an access token minted for `scope=mcp` now authorizes through the normal user-role branches instead of the read-only scope gates, so it can perform writes on any project the user belongs to.

server/lib/tuist_web/api/schemas/project.ex
```diff
@@ -29,6 +29,11 @@ defmodule TuistWeb.API.Schemas.Project do
         description:
           "The URL of the connected git repository, such as https://github.com/tuist/tuist or https://github.com/tuist/tuist.git"
       },
+      token: %Schema{
```
cschmatzler, line 32: [P3] This reintroduces `Project.token` in the server schema, but the generated OpenAPI artifacts were not updated in the same PR. `cli/Sources/TuistServer/OpenAPI/server.yml` and `Types.swift` still describe `Components.Schemas.Project` without this field, so the checked-in API contract is now out of sync with the actual response shape. Please regenerate the spec/client with `mise run generate-api-cli-code` alongside this schema change.

server/lib/tuist/oauth/token_generator.ex
fortmarek: Good catch — you're right that dropping scopes entirely was a security issue.

Fixed in 9f93ce47b5: OAuth tokens now remain as scoped `AuthenticatedAccount` JWTs (preserving `scope` from the Boruta grant and the `type: "account"` marker), so `scopes_permit/3` enforcement stays intact. To solve the cross-org project listing, the token now includes a `user_id` claim which gets resolved into a `user` field on `AuthenticatedAccount`. This lets `list_accessible_projects` and `project_access_permitted` delegate to user-level org membership checks without bypassing scope gates.

server/lib/tuist_web/api/schemas/project.ex
fortmarek: Done — ran `mise run generate-api-cli-code` and the regenerated `server.yml` and `Types.swift` are now included in 9f93ce47b5.

server/lib/tuist/accounts/authenticated_account.ex
```diff
@@ -9,7 +9,8 @@ defmodule Tuist.Accounts.AuthenticatedAccount do
   - `project_ids` - When all_projects is false, list of project IDs the token can access
   - `token_id` - The ID of the token used for authentication
   - `created_by_account_id` - The ID of the account that created this token
+  - `user` - The user who authorized the token (set for OAuth tokens to enable cross-org project listing)
   """
   @enforce_keys [:account, :scopes]
-  defstruct [:account, :scopes, :all_projects, :project_ids, :token_id, :created_by_account_id]
+  defstruct [:account, :scopes, :all_projects, :project_ids, :token_id, :created_by_account_id, :user]
```
pepicrft: We are mixing two concerns here:
1. The authorization subject: `account, `scopes`, `all_projects`, `project_ids`
2. The authorization actor: `created_by_account_id` and `user`

The consumer using `.user` doesn't know that the user it represents. I'd make it obvious in the name, for example `grated_by` or `issued_by`. Also, why do we have `created_by_account_id` and `user`?

server/lib/tuist/guardian.ex
```diff
@@ -33,12 +33,15 @@ defmodule Tuist.Guardian do
   def resource_from_claims(%{"sub" => id, "type" => "account"} = claims) do
```
pepicrft, line 33: I'd test that this logic is including the `user` in the returned data structure.

server/lib/tuist/accounts/authenticated_account.ex
pepicrft: I think the authorization is getting a bit harder to reason about from the data structures. We don't need to address this as part of the PR, but I'd consider bringing some explicitness around these authorization models:

- Token-scoped access: We determine the access based on the information in the token (scopes, projects...)
- User-identity access: We dtermine the access based on the relationship between the user and the objects they are trying to access

```exs
%Subject{actor: %User{}, grant: %Grant.User{ account: %Account() }}

%Subject{actor: %User{}, grant: %Grant.AccountToken{
  account: %Account{},
  scopes: ["project:cache:write", ...],
  project_ids: [1, 2, 3] | :all,
  token_id: 123
}}
```

fortmarek: Good call on the naming. Renamed `user` to `issued_by` in c28f9287ed to make the intent clear.

Re: the overlap with `created_by_account_id` -- they represent different things and never coexist on the same token:
- `created_by_account_id`: which account created a manually-issued account token (set in `authentication.ex` when resolving from the DB, used for preview/build attribution)
- `issued_by`: which user authorized an OAuth grant (set in `guardian.ex` from the `user_id` JWT claim, used for cross-org membership checks)

+1 on the broader `Subject`/`Grant` refactor idea. Separating the authorization actor from the token-scoped grant would make these flows much easier to reason about. Happy to tackle that as a follow-up.

server/lib/tuist/guardian.ex
fortmarek: Added in 6ead950ea3 -- two tests in `guardian_test.exs`:
- `issued_by` is populated when `user_id` claim is present
- `issued_by` is `nil` when `user_id` claim is absent

server/lib/tuist/accounts/authenticated_account.ex
fortmarek: In other words, `created_by_account_id` should stay for now until we can distinguish between the oauth and our long-lived account token

cli/Sources/TuistServer/Utilities/ServerCredentialsStore.swift
```diff
@@ -144,8 +144,9 @@ public enum ServerCredentialsStoreBackend: Sendable {
         public func read(serverURL: URL) async throws -> ServerCredentials? {
             switch backend {
             case .keychain:
-                let refreshToken = try keychain(serverURL: serverURL).get(serverURL.absoluteString + "_refresh_token")!
-                let accessToken = try keychain(serverURL: serverURL).get(serverURL.absoluteString + "_access_token")!
+                guard let accessToken = try keychain(serverURL: serverURL).get(serverURL.absoluteString + "_access_token")
```
cschmatzler, line 147: [P2] This turns a partially-written keychain state into a silent “logged out” result. `store()` persists the refresh token and access token as separate entries, so an interrupted write can leave only the refresh token behind; before this change that corruption surfaced as a failure, but now `read()` returns `nil` and `fetchTokenFromStore()` stops before it ever attempts a refresh. That forces an unnecessary re-login in exactly the case where the refresh token could have recovered the session.

server/lib/tuist/oauth/token_generator.ex
```diff
@@ -25,13 +27,19 @@ defmodule Tuist.OAuth.TokenGenerator do
           :refresh_token -> client.refresh_token_ttl
         end
 
+      scopes = parse_scopes(scope)
+
       claims = %{
+        "type" => "account",
```
cschmatzler, line 33: [P1] These OAuth tokens now decode as `%AuthenticatedAccount{}` instead of `%User{}`, but the project CRUD endpoints are still wired for users only: `ProjectsController.create/show/update/delete` call `Authentication.current_user(conn)`, and the `:project` / `:projects` policies only have `authenticated_as_user` branches. In practice `/api/projects` still works because it uses `authenticated_subject`, while `GET/POST/PUT/DELETE /api/projects...` starts rejecting OAuth clients (and the deprecated `name`-only create path can even crash when it dereferences `user.account.name` on `nil`).

fortmarek:
> There are many policy rules that don't have an `authenticated_as_account`, which means they'll fail to authorize. I believe we should update those, no?

Good point. Added `authenticated_as_account` with `scopes_permit: "project:admin:read"` to both `project:read` and `projects:read` in 454abd254b. Also added `project:admin:read` to the `mcp` scope group and default OAuth scopes, so app and MCP tokens can read project details out of the box.

The remaining rules without `authenticated_as_account` are intentionally admin-only operations (account/org management, billing, invitations, members) that shouldn't be accessible via scoped tokens. A broader audit of all rules would fit well in the `Subject`/`Grant` refactor discussed above.

Issue #6456, opened 2024-06-28
Generate Objective C resources for internal targets


fortmarek:
### What problem or need do you have?

We currently don't generate resource bundle accessors for internal targets that purely include Objective C sources.
We used to do that but had to revert that in [this PR](https://github.com/tuist/tuist/pull/6388).

In other words, an objc accessor file should be generated for `Framework1` of the following project:
[ios_app_with_framework_and_disabled_resources.zip](https://github.com/user-attachments/files/16027873/ios_app_with_framework_and_disabled_resources.zip)

### Potential solution

As described in the [PR](https://github.com/tuist/tuist/pull/6388), before we can [delete the line](https://github.com/tuist/tuist/blob/main/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift#L84) guarding the objc resource accessors only to external projects, we will need to figure out how we can generate the Objc resource accessor without overriding the `GCC_PREFIX_HEADER`.

### macOS version

xx

### Tuist version

4.19.0

### Xcode version

xx

varshith257:
@fortmarek Though I am new to this project, I would like to tackle this issue under your guidance and marking my strong commitment with an attempt :)
/attempt #6456

Harry-jain:
/attempt #6456

Hi maintainers 👋
I'd like to work on this issue. Please let me know if there are any specific constraints or guidance I should keep in mind while implementing the fix.

Athx9:
I am interested in this bounty. Can you please assign it to me or let me know if I can start working on it?

Nithinfgs:
/attempt #6456

Analyze the current implementation in ResourcesProjectMapper.swift (around line ~84) where the guard prevents generating ObjC resource bundle accessors for internal targets.
Remove or modify the guard condition so that ObjC resource bundle accessors can also be generated for internal targets.
Update the ObjC accessor generation logic to avoid using GCC_PREFIX_HEADER.
Modify the generated .h accessor file to directly import the required framework (e.g., #import <Foundation/Foundation.h>) so the header is self-contained.
Ensure the generated accessor works without overriding any user-defined prefix headers.
Verify the fix by generating a project with an internal target and confirming that the ObjC resource accessor is produced correctly and compiles without prefix header conflicts.
Add or update tests if applicable to ensure resource bundle accessors are generated correctly for internal targets.
Run the full build and test suite to confirm no regressions before submitting the PR.

yemsy26:
/attempt #6456
[...]

PR #10162, opened 2026-04-01
Fix ObjC internal resources generation (Issue #6456)

yemsy26:
This PR fixes the generator to emit Objective-C accessors for internal targets.

### The Fix
Historically, generating accessors for internal targets was reverted because Tuist forcefully overrode `GCC_PREFIX_HEADER`, which collided with heavily customized prefix headers on large projects (e.g. Etsy).
This solution takes a purely additive approach following SPM's standard:
1. We inject the generated header via an `-include` flag added cleanly to `OTHER_CFLAGS`.
2. We wrap the `Foundation` import inside the generated header in `#ifdef __OBJC__` to protect C/C++ clang phases from breaking.

**Note:** This PR is pushed as a DRAFT. GitHub Actions will validate the Xcode compatibility.

/claim #6456


yemsy26:
🚀 Solution Summary & Verification

I've validated that the previous approach failed because it overwrote the GCC_PREFIX_HEADER. This new implementation:

Uses an additive approach: It appends the generated header via -include in OTHER_CFLAGS, preserving any existing user PCH settings.

Includes a Clang Guard: Wrapped the generated header content in #ifdef __OBJC__ to ensure pure C/C++ targets don't break during compilation.

Ready for review and for the CI to verify the Xcode build. Thanks!

PR #10167, opened 2026-04-02
feat(server): adopt dashboard locale from browser

pepicrft:
> [!NOTE]
> In a follow-up PR I'll add the option for users to override the language preference for their account.

## Summary
This PR makes the authenticated dashboard adopt the browser language reported by the client.

## What Changed
- added a shared locale normalizer for browser locale strings used by the dashboard
- passed the browser language preferences from the app LiveView client
- set the app request and LiveView locale from the normalized browser locale
- updated the dashboard app layout to reflect the active locale in the HTML `lang` attribute
- reused the same normalization path for marketing `Accept-Language` handling so both sides interpret locale codes consistently
- added focused tests for locale normalization and translated dashboard labels

## Why
The dashboard was translation-aware in its templates and LiveView labels, but the app pipeline never set a locale from the authenticated user's browser preferences. The browser and server were also not guaranteed to use the same locale code format, especially for Chinese variants.

## Impact
Authenticated users now see the dashboard in the language that best matches their browser locale when that locale is supported. This lays the groundwork for a later user setting that can override the browser preference.

## Validation
- `mix test test/tuist_web/locale_test.exs test/tuist_web/live/layout_live_test.exs`

### DIFF
--- a/L10N.md
+++ b/L10N.md
@@ -1,5 +1,4 @@
 ---
-model: "anthropic:claude-sonnet-4-6"
 source_language: "en"
 ---
 
--- a/translate.exs
+++ b/translate.exs
@@ -519,7 +519,8 @@ defmodule L10n.CLI do
     * `--force`, `-f` — Re-translate all files, ignoring lock state
     * `--locale`, `-l` — Translate only the specified locale (e.g., `--locale es`)
     * `--model`, `-m` — Override the LLM model from L10N.md (e.g., `--model openai:gpt-4.1`)
-    * `--concurrency`, `-c` — Max parallel translations per .pot file (default: 4)
+    * `--concurrency`, `-c` — Max parallel translations per locale within one source file (default: 4)
+    * `--pot-concurrency` — Max parallel source files being translated at once (default: 1)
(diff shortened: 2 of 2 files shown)


server/test/tuist_web/live/layout_live_test.exs
```diff
@@ -164,6 +165,18 @@ defmodule TuistWeb.LayoutLiveTest do
...
+    @tag user_role: :user
+    test "uses the session locale for translated dashboard labels", %{
+      params: params,
+      session: session
+    } do
+      {:cont, socket} = Locale.on_mount(:assign_locale, %{}, Map.put(session, "locale", "es"), %LiveView.Socket{})
+
+      {:cont, socket} = LayoutLive.on_mount(:project, params, session, socket)
+
+      assert Enum.any?(Enum.at(socket.assigns.breadcrumbs, 0).items, &(&1.label == "Crear organización"))
```
pepicrft, line 178: Codex decided to add an integration test, and I think it's a good idea. I don't expect that piece of text to change, so I think it's fine.

fortmarek (approved):

PR #10168, opened 2026-04-02
feat(cli): fix App Intents metadata for cached xcframeworks

pepicrft:
## Summary
- inject a deterministic App Intents metadata bridge for cached static XCFrameworks
- move transitive static XCFramework metadata discovery into `GraphTraverser`
- cover the behavior with traverser, mapper, and factory tests

## Why
Cached static XCFrameworks can already contain `Metadata.appintents`, but the generated project does not feed those paths into Xcode's App Intents metadata file lists. As a result, Xcode can miss actions during App Intents metadata processing and fail App Shortcuts validation.

> [!NOTE]
> `Metadata.appintents` is the metadata bundle generated for App Intents symbols and shortcuts. During the build, Xcode's App Intents tools read the current target's extracted metadata together with dependency metadata listed in `DependencyMetadataFileList` and `DependencyStaticMetadataFileList`. Static frameworks also need the sibling `<Name>.appintents/Metadata.appintents` layout that Xcode expects when processing static dependency metadata.

## Upstream context
This appears to be an underlying Xcode or `swift-build` limitation rather than a Tuist-specific modeling issue.

Relevant open-source references:
- `swift-build` populates App Intents dependency metadata from transitive build-target products: https://github.com/swiftlang/swift-build/blob/main/Sources/SWBApplePlatform/AppIntentsMetadataTaskProducer.swift
- `swift-build` passes those file lists to `appintentsmetadataprocessor` via `--metadata-file-list` and `--static-metadata-file-list`: https://github.com/swiftlang/swift-build/blob/main/Sources/SWBApplePlatform/Specs/AppIntentsMetadata.xcspec
- XCFramework handling is planned separately as a copy step: https://github.com/swiftlang/swift-build/blob/main/Sources/SWBTaskConstruction/TaskProducers/WorkspaceTaskProducers/XCFrameworkTaskProducer.swift

I did not find equivalent App Intents metadata plumbing in `swift-package-manager`; the relevant behavior appears to live below SwiftPM in `swift-build`.

## Implementation
For runnable targets that depend on static XCFrameworks with App Intents metadata, Tuist now generates a narrow pre-build script that:
- derives the affected framework names from each XCFramework `Info.plist`
- rewrites `DependencyMetadataFileList` and `DependencyStaticMetadataFileList` with the exact expected paths
- materializes the sibling `.appintents` bundle Xcode expects

The graph-walk logic now lives in `GraphTraverser`, so the mapper is reduced to filtering existing metadata directories and generating the script.

No cache version bump is included. The fix is in project generation, and the validated repro artifact already contains the metadata.

## Validation
- added `TuistCoreTests/GraphTraverserTests` coverage for transitive static XCFramework App Intents metadata discovery
- added `TuistKitTests/StaticXCFrameworkAppIntentsMetadataGraphMapperTests`
- updated `TuistKitTests/GraphMapperFactoryTests`
- verified the cached repro fails without the injected file-list bridge and succeeds with it
- ran targeted `TuistUnitTests` for traverser, mapper, and factory coverage
- verified the rebased CLI builds with `xcodebuild build -workspace Tuist.xcworkspace -scheme tuist ...`

### DIFF
--- a/cli/Sources/TuistKit/Mappers/Graph/StaticXCFrameworkAppIntentsMetadataGraphMapper.swift
+++ b/cli/Sources/TuistKit/Mappers/Graph/StaticXCFrameworkAppIntentsMetadataGraphMapper.swift
@@ -48,8 +48,8 @@ public struct StaticXCFrameworkAppIntentsMetadataGraphMapper: GraphMapping {
         let dependenciesScript = dependencies.map { dependency in
             """
             framework_name='\(dependency.frameworkName)'
-            framework_metadata="${BUILT_PRODUCTS_DIR}/${framework_name}.framework/Metadata.appintents"
-            static_metadata="${BUILT_PRODUCTS_DIR}/${framework_name}.appintents/Metadata.appintents"
+            framework_metadata="\(dependency.frameworkMetadataPath)"
+            static_metadata="\(dependency.staticMetadataPath)"
 
             if [ -d "$framework_metadata" ] && [ ! -d "$static_metadata" ]; then
                 mkdir -p "$static_metadata"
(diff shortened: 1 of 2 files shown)


cli/Sources/TuistKit/Mappers/Graph/StaticXCFrameworkAppIntentsMetadataGraphMapper.swift
```diff
@@ -0,0 +1,110 @@
...
+        let script = """
+        METADATA_FILE="\(Constants.metadataFile)"
+        STATIC_METADATA_FILE="\(Constants.staticMetadataFile)"
+
+        : > "$METADATA_FILE"
+        : > "$STATIC_METADATA_FILE"
+
+        \(dependenciesScript)
+        """
+        return TargetScript(
```
fortmarek, line 79: should the script have inputs/outputs? otherwise, it will be run on every build which may not be desirable.

If we do add script outputs, we might need to ensure this is not run twice in the same dependency tree as that could lead to a mulitple products error

PR #10169, opened 2026-04-02
feat(cli): support self-managed shard archives

pepicrft:
## Summary

This change adds a first-class self-managed shard artifact flow for test sharding.

- adds `--shard-archive-path` and `TUIST_TEST_SHARD_ARCHIVE_PATH` to `tuist test` and `tuist xcodebuild`
- writes the optimized `.xctestproducts` archive locally during build-for-testing instead of uploading when an archive path is provided
- extracts a local shard archive during shard execution instead of downloading from remote storage when an archive path is provided
- keeps the existing remote upload/download flow as the default
- updates the sharding docs for both generated projects and `xcodebuild`

## Why

Teams that already use their CI provider's native artifact upload and download steps still need Tuist's archive and extraction behavior. Before this change, they had to duplicate Tuist's internal packing logic in custom scripts.

This PR makes the archive boundary explicit while leaving transport ownership flexible.

## Impact

Users can now let Tuist own the shard archive format and optimizations while their CI system owns artifact transport.

This covers both:

- `tuist test`
- `tuist xcodebuild build-for-testing` / `tuist xcodebuild test` / `tuist xcodebuild test-without-building`

## Validation

- `tuist generate --no-open`
- `mise run cli:lint --fix`
- added unit coverage for:
  - archive creation and remote-upload bypass
  - local archive extraction and `.xctestrun` filtering
  - command/service forwarding
  - command environment variable parsing

## Test blocker

Targeted `xcodebuild test` is currently blocked by an unrelated dependency compile failure in `mcp-swift-sdk` under Swift concurrency checks:

- `.../NetworkTransport.swift:581`
- `.../NetworkTransport.swift:812`

That failure happens before the modified TuistKit test targets run, so this PR documents the blocker rather than claiming a green targeted test run.

### DIFF
--- a/cli/Sources/TuistKit/Services/Sharding/ShardService.swift
+++ b/cli/Sources/TuistKit/Services/Sharding/ShardService.swift
@@ -107,8 +107,9 @@ public struct ShardService: ShardServicing {
             resolvedTestProductsPath = testProductsPath
             Logger.current.debug("Using local test products at \(testProductsPath.pathString)")
         } else if let testProductsArchivePath {
-            resolvedTestProductsPath = try await fileSystem.makeTemporaryDirectory(prefix: "tuist-shard-unzip")
-            try await appleArchiver.decompress(archive: testProductsArchivePath, to: resolvedTestProductsPath)
+            let extractedTestProductsPath = try await fileSystem.makeTemporaryDirectory(prefix: "tuist-shard-unzip")
+            try await appleArchiver.decompress(archive: testProductsArchivePath, to: extractedTestProductsPath)
+            resolvedTestProductsPath = try await normalizeExtractedTestProductsPath(extractedTestProductsPath)
             Logger.current.debug("Extracted local shard archive to \(resolvedTestProductsPath.pathString)")
         } else {
             guard let downloadURL = URL(string: shard.download_url) else {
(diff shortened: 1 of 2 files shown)


server/priv/docs/en/guides/features/test-sharding/generated-projects.md
```diff
@@ -416,6 +418,29 @@ tuist test -- -testProductsPath /path/to/shared/volume/$UNIQUE_ID/MyScheme.xctes
...
+2. Upload that archive using your CI's native artifact step.
+
+3. In each **test phase** job, download the archive and pass the same path back to Tuist:
+
+```sh
+tuist test \
+  --shard-archive-path /tmp/shards/${UNIQUE_ID}/bundle.aar
+```
+
+When `--shard-archive-path` is set, Tuist skips remote test-products transfer and uses the local archive instead. If you also pass `--shard-skip-upload`, the archive path takes precedence.
```
fortmarek, line 442: I'd use these new variables in the namespace example below (since the volume flow doesn't actually exist, AI hallucinated that one)

fortmarek: Same for the xcode page

PR #10176, opened 2026-04-02
fix(docs): Clarify account token account handle in auth docs

pepicrft:
## Summary
Clarify that the `account` argument in account token commands must be the handle of the account that owns the target projects.

## What Changed
- Replaced `my-account` with the more neutral placeholder `account`
- Added an admonition under the first example explaining that the account can be an organization account, not necessarily the user's personal account
- Updated the related `create`, `list`, and `revoke` examples to use the same placeholder consistently

## Why
The previous example could be read as "use your own account" rather than "use the account that owns the projects you need to access," which is the important distinction for shared organization setups.

## Validation
- Reviewed the rendered markdown structure locally via diff
- No build or test run; documentation-only change

### DIFF
--- a/server/priv/docs/en/guides/server/authentication.md
+++ b/server/priv/docs/en/guides/server/authentication.md
@@ -80,14 +80,14 @@ For CI environments that don't support OIDC, or when you need fine-grained contr
 ### Creating an account token {#creating-an-account-token}
 
 ```bash
-tuist account tokens create account \
+tuist account tokens create account-handle \
   --scopes project:cache:read project:cache:write \
   --name ci-cache-token \
   --expires 1y
 ```
 
 > [!IMPORTANT]
-> Replace `account` with the handle of the account that owns the projects you want the token to access. This can be an organization account, not necessarily your personal account.
+> Replace `account-handle` with the handle of the account that owns the projects you want the token to access. This can be an organization account, not necessarily your personal account.
 
 The command accepts the following options:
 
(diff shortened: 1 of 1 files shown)


server/priv/docs/en/guides/server/authentication.md
```diff
@@ -80,12 +80,15 @@ For CI environments that don't support OIDC, or when you need fine-grained contr
 ### Creating an account token {#creating-an-account-token}
 
 ```bash
-tuist account tokens create my-account \
+tuist account tokens create account \
```
ivan-delivery: (suggestion/nitpick) Can we use `account-handle` here, as this is what I see from the CLI.
```
Usage: tuist account tokens list <account-handle> [--path <path>] [--json]
```
I was confused by my assumption that `member == account`, and not knowing that organisation is also an `account-handle` in fact.

server/priv/docs/en/guides/server/authentication.md
```diff
@@ -80,12 +80,15 @@ For CI environments that don't support OIDC, or when you need fine-grained contr
...
   --scopes project:cache:read project:cache:write \
   --name ci-cache-token \
   --expires 1y
 ```
 
+> [!IMPORTANT]
+> Replace `account` with the handle of the account that owns the projects you want the token to access. This can be an organization account, not necessarily your personal account.
```
ivan-delivery: 👍

PR #10199, opened 2026-04-08
feat(server): page level filters

asmitbm:
Resolves TUI-559

Move filters to page level:

Screenshot 2026-04-08 at 11 09 37

Modified pages:

- Builds


fortmarek (changes requested):
This should _not_ be just a UI change. We need to go through each page and make sure the filters actually apply to all the content where it makes sense.

For example, the `Environment: Local` filter should also apply to `Configuration insights` and `Recent builds`. If it does not, then I think the page-level filters look odd:

Image

Also, noticed the CSS for checkmarks needs to be updated, so only the currently selected value actually has a visible one:

Image

Given the scope of the changes will increase, I'd probably suggest to ship one or two pages with page-level filters _first_ before doing everything at once.

asmitbm:
Changes:

* Remove page level filters from all pages except builds page
* Fix dropdown selector to show just the selected choice
* Fix date picker component to use right shadows and colors

noora/css/dropdown.css
```diff
@@ -324,6 +324,10 @@
     }
   }
 
+  &:not([data-selected]) > [data-part="right-icon"] {
```
fortmarek: This globally hides `right-icon` for any dropdown item without `data-selected`, which breaks the two places that use `:right_icon` for a semantic (non-selection) icon:

- [`docs/components/toc.html.heex:29`](https://github.com/tuist/tuist/blob/main/server/lib/tuist_web/docs/components/toc.html.heex#L29) — the "View as Markdown" external-link icon disappears.
- [`docs/components/page_layout.html.heex:95`](https://github.com/tuist/tuist/blob/main/server/lib/tuist_web/docs/components/page_layout.html.heex#L95) — same.

Suggestion: scope this rule to dropdowns that actually represent a selection group (e.g. add a `[data-selectable]` attribute on the parent and key off that), or render `<.check />` conditionally in the templates instead of always emitting it. Either way, keep the rule out of the shared dropdown stylesheet.

server/lib/tuist_web/live/xcode_builds_live.ex
```diff
@@ -336,8 +336,8 @@ defmodule TuistWeb.XcodeBuildsLive do
             }
           end)
 
-        %{successful_count: successful_builds_count, failed_count: failed_builds_count} =
-          Builds.recent_build_status_counts(project.id, limit: 40)
+        successful_builds_count = Enum.count(recent_builds, &(&1.status == "success"))
+        failed_builds_count = Enum.count(recent_builds, &(&1.status == "failure"))
```
fortmarek, line 340: This replaces the dedicated `Builds.recent_build_status_counts/2` ClickHouse aggregation with in-memory counting of the already-fetched `recent_builds`. Two notes:

1. **Behavior change (arguably an improvement):** counts now respect the active filters, which is more consistent with the list that's rendered. Worth confirming that's intentional — the old function ignored filters entirely.
2. **Test coverage:** `test/tuist/builds_test.exs` still exercises `recent_build_status_counts/2`, but there's no LiveView-level test that asserts the status counts update when a filter (environment/scheme/tag) changes. Easy to regress.

server/lib/tuist_web/live/gradle_builds_live.ex
```diff
@@ -238,8 +240,18 @@ defmodule TuistWeb.GradleBuildsLive do
...
+           socket
+       ) do
+    filters =
+      case analytics_environment do
+        "ci" -> [%{field: :is_ci, op: :==, value: true}]
+        "local" -> [%{field: :is_ci, op: :==, value: false}]
+        _ -> []
+      end
+
+    {builds, _meta} = Gradle.list_builds(project.id, %{page_size: @recent_builds_page_size, filters: filters})
```
fortmarek, line 267: Only the `is_ci` filter is threaded into the recent-builds query here. The page-level date range (`analytics-date-range`) is applied to the analytics widgets above but not to this list, so the "recent builds" section shows results outside the selected window.

If the intent of hoisting filters to the page is "everything on this page respects these filters," this should also take `start_datetime`/`end_datetime` from `DatePicker.date_picker_params(params, "analytics")`. If the intent is "recent = always the latest 5 regardless of filters," that's fine, but I'd add a comment clarifying that — right now it reads like an oversight.

server/lib/tuist/builds/analytics.ex
```diff
@@ -35,6 +35,7 @@ defmodule Tuist.Builds.Analytics do
       )
 
     query
+    |> apply_filters(opts)
```
fortmarek, line 38: Good catch adding this (and the matching `maybe_filter_ci(opts)` in `gradle/analytics.ex:560`) — the configuration-insights analytics query was silently ignoring the filter opts before. Can you add a test in `test/tuist/builds/analytics_test.exs` that calls `configuration_insights/2` with e.g. `is_ci: true` and asserts the result set narrows? Small guardrail against regressions now that the filter is shared between multiple charts.

fortmarek (commented):
Nice simplification overall — removing the duplicate date picker and unifying filters at the page level reads well. A few concerns below, the dropdown CSS one is the most pressing because it breaks unrelated docs components.

Also: the PR description still lists Tests/Test Runs/Shards/Bundles/etc., but the final commit reverted those. Mind updating the description to match what's actually in the diff?

noora/css/dropdown.css
fortmarek: I believe this is also the reason for the bug where the individual labels (`Xcode`/`Gradle`) are not visible in the project dropdown:
image

asmitbm: I didn't know dropdowns were structured like that, will makes changes in the next commit

fortmarek (approved):
Great work, thanks 👏

PR #10228, opened 2026-04-09
fix(cli): use combined module map to avoid argument list too long errors

fortmarek:
## Summary

Fixes "Argument list too long" (E2BIG) errors during `tuist cache` and `tuist test` for targets with many transitive dependencies.

There are two sources of argument list bloat, addressed by two commits:

### 1. Combined module map (commit b279d26)

- The `ModuleMapMapper` previously added individual `-fmodule-map-file=<path>` flags for each transitive dependency's module map. For large dependency graphs, this produced 300+ build setting entries in `OTHER_C_FLAGS` and `OTHER_SWIFT_FLAGS`
- Now generates a single combined module map file per target using clang's `extern module` declarations, referenced by one `-fmodule-map-file` flag. Reduces flag count from O(N) to O(1)

### 2. Framework search paths response file (commit d6df2bf)

- When cached targets are replaced with xcframeworks, each xcframework lives in its own hash-based directory (`~/.cache/tuist/Binaries/<hash>/`). The `LinkGenerator` added each as a `FRAMEWORK_SEARCH_PATHS` entry, producing 100+ unique `-F` flags
- Swift compilation handles this fine (uses `.resp` response files), but C/ObjC compilation (e.g. `_vers.c`, `.m` files) receives all flags directly on the command line, exceeding `ARG_MAX`
- When a target has 20+ precompiled framework search paths, they are now written to a `Derived/FrameworkSearchPaths/<TargetName>.resp` file and referenced via `@file` in `OTHER_CFLAGS`, `OTHER_SWIFT_FLAGS`, and `OTHER_LDFLAGS`. Clang and ld natively support `@file` expansion
- Below the threshold, behavior is unchanged

## Test plan

- [x] `ModuleMapMapperTests` pass (3 existing tests updated + side effect assertions)
- [x] `LinkGeneratorTests` pass (33 tests including new `test_setupFrameworkSearchPath_consolidatesIntoResponseFile_whenManyPrecompiledPaths`)
- [x] Full workspace build succeeds (`xcodebuild build -scheme Tuist-Workspace`)

### E2E verification with 800-module fixture

Created a fixture project with 800 framework targets + one "GodModule" that depends on all 800 (with an ObjC source file to trigger C compilation):

1. `tuist cache` to cache all 801 targets as xcframeworks
2. Invalidate GodModule's cache
3. `tuist cache --generate-only` to regenerate with 800 cached xcframework dependencies

**Before fix** (tuist 4.176.4):
- `FRAMEWORK_SEARCH_PATHS` for GodModule: 800 entries pointing to `~/.cache/tuist/Binaries/<hash>`
- All expanded as `-F` flags on the C compiler command line

**After fix**:
- `FRAMEWORK_SEARCH_PATHS`: empty (cache paths removed)
- `OTHER_CFLAGS`: `@$(SRCROOT)/Derived/FrameworkSearchPaths/GodModule.resp`
- Response file: 800 `-F<path>` lines, read by clang via `@file` (bypasses ARG_MAX)
- `xcodebuild build -scheme GodModule` succeeds (including ObjC compilation)

Could not reproduce the actual E2BIG error locally because ARG_MAX is 1MB and local env is ~12KB (vs 200KB+ on CI), but the fix demonstrably moves all framework search paths out of the command line.

### DIFF
--- a/cli/Sources/TuistConstants/Constants.swift
+++ b/cli/Sources/TuistConstants/Constants.swift
@@ -47,6 +47,7 @@ public enum Constants {
         public static let entitlements = "Entitlements"
         public static let privacyManifest = "PrivacyManifests"
         public static let moduleMaps = "ModuleMaps"
+        public static let frameworkSearchPaths = "FrameworkSearchPaths"
         public static let sources = "Sources"
         public static let testPlans = "TestPlans"
         public static let signingKeychain = "signing.keychain"
--- a/cli/Sources/TuistGenerator/Generator/LinkGenerator.swift
+++ b/cli/Sources/TuistGenerator/Generator/LinkGenerator.swift
@@ -1,6 +1,7 @@
 import Foundation
 import Path
 import PathKit
+import TuistConstants
 import TuistCore
 import TuistLogging
 import TuistSupport
@@ -40,7 +41,7 @@ protocol LinkGenerating {
         path: AbsolutePath,
         sourceRootPath: AbsolutePath,
         graphTraverser: GraphTraversing
-    ) throws
+    ) throws -> [SideEffectDescriptor]
 }
 
 /// When generating build settings like "framework search path", some of the path might be relative to paths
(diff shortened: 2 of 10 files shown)


pepicrft (approved):

n-zaitsev:
Hello @fortmarek @pepicrft , is there any chance this would be merged any time soon ?

PR #10232, opened 2026-04-09
feat(server): add automations engine for flaky tests

fortmarek:
## Summary

Introduces a unified Automations engine for test-case hygiene and lands its first sub-kind: **Alert automations** (aka "alert rules"). Each alert evaluates a monitor condition on a cadence, runs one or more trigger actions when it fires, and optionally runs recovery actions when the condition no longer holds. The shape is intentionally set up so we can add non-alert kinds (scheduled, webhook, function) later without breaking contracts.

### Architecture: Automations → Alerts → Monitors · Events · Actions

The folder structure and naming reflect the umbrella:

- `Tuist.Automations.Alerts.Alert` — the configurable record (PG table `automation_alerts`). Owns the monitor, trigger/recovery config, trigger/recovery actions, and an enabled flag.
- `Tuist.Automations.Alerts.Event` — append-only ClickHouse log (`automation_alert_events`) of transitions per entity. `list_active_alert_events/1` uses `argMax` to surface currently-firing entities; `create_alert_event/1` appends a new record.
- `Tuist.Automations.Monitors.FlakyTestsMonitor` — today's single monitor, with `flakiness_rate` and `flaky_run_count` variants.
- `Tuist.Automations.Actions.*` — `ChangeStateAction`, `AddLabelAction`, `RemoveLabelAction`, `SendSlackAction`. `ActionExecutor` dispatches with a generic `%{type: :test_case, id: id}` entity map so the engine can later work on builds or bundles.

### Test-case `state`

- New `state` field on `test_cases` (`enabled` | `muted`), replacing `is_quarantined` as source of truth.
- API responses still include a derived `is_quarantined` boolean flagged `deprecated: true` in OpenAPI so SDK generators emit a warning.
- The `is_quarantined` column itself lives in a follow-up migration so this PR stays purely additive.
- MCP tools expose `state` directly (no back-compat obligation).
- Mute/unmute dropdown on the test-case page replaces the quarantine/unquarantine buttons.

### REST API

- `GET/POST/PUT/DELETE /api/projects/:account_handle/:project_handle/automations/alerts[/{alert_id}]` — operation IDs `listAutomationAlerts`, `createAutomationAlert`, etc.
- Response + request-body schemas pulled into modules (`TuistWeb.API.Schemas.AutomationAlert`, `AutomationAlertAction`); controller helpers (normalize, error formatting) pulled into `TuistWeb.API.RequestParams`.
- Gated by a dedicated `:automation_alert` authorization object (admin-only writes, user/admin/ops/account-token reads). The LiveView uses the same object's `:read` so UI and API gates don't drift.

### Alert builder UI

- Modal with Condition / Action / Recovery sections, multiple actions per rule, add/delete, edit-in-place.
- Slack action with channel picker; message template supports `{{test_case.name}}`, `{{test_case.module_name}}`, `{{test_case.suite_name}}`, `{{test_case.url}}`, `{{automation.name}}`; user-controlled values are mrkdwn-escaped to prevent injection.
- Recovery toggle with configurable window; recovery actions run *before* the "recovered" event is appended so a failed Slack ping / unmute / label removal doesn't leave the UI looking resolved while the side effects never happened.

### Legacy migration

- Backfill migration materialises `automation_alerts` rows from existing project settings (`auto_mark_flaky_tests`, `auto_quarantine_flaky_tests`, `flaky_test_alerts_*`). One-shot insert, no de-dup guard — volume is low.
- Removed workers: `FlakyThresholdCheckWorker`, `FlakyTestAlertWorker`, `ClearCooledDownFlakyTestsScheduler/Worker`.
- Removed the manual-mark-as-flaky Slack alert (`send_manual_flaky_test_alert`); users configure Slack via the alert's `send_slack` action instead.
- Legacy "Flaky test detection" and "Test quarantine" settings cards removed.

### Oban scheduler

The scheduler includes `:completed` in Oban's unique-state set alongside `:available`/`:scheduled`/`:executing`. Without it, a fast evaluation job would drop to `:completed` between scheduler ticks and the next tick would re-enqueue, collapsing the effective cadence to the scheduler's own interval. With `period: cadence_seconds(cadence)` + `:completed`, re-scheduling waits at least one cadence.

### Worktree bootstrap

Three mise tasks under `claude:` (`bootstrap`, `regenerate-launch-json`, `seed-db`) auto-configure a fresh worktree. The committed `.claude/settings.json` wires them into a `SessionStart` hook.

### Noora changes

- Dropdown component: icon slot rendered via portal for dynamic updates; `size="medium"` variant.
- `volume-3.svg` icon added.

## Test plan

- [ ] PostgreSQL + ClickHouse migrations apply cleanly on a fresh worktree.
- [ ] Mute/unmute dropdown writes `state` and the API still returns the derived `is_quarantined` value.
- [ ] CRUD via `/api/projects/.../automations/alerts`; admin-only writes, broader reads.
- [ ] Multi-action alerts (change_state + add_label + send_slack), recovery toggle, cooldown.
- [ ] Legacy `?quarantined=true` API filter keeps returning muted tests.
- [ ] Enqueue `AlertEvaluationWorker` from iex and verify events land in `automation_alert_events` and actions fire in order.
- [ ] Action integration tests in `test/tuist/automations/actions/*` go through `Tests.update_test_case` down to a ClickHouse write so contract drift surfaces here.
- [ ] ~70 tests covering schema validation, context CRUD, action handlers, executor, evaluation worker, API controller, LiveView, and MCP tools.


.claude/settings.json
fortmarek, line 1: these are claude specific, but figured, they'd be useful for anyone else using the claude worktrees

server/lib/tuist/automations/alert_rule.ex
fortmarek, line 1: I went a bit back-and-forth on this. Initially, this scheme was called `automation`, but in the end decided that this specific shape is really an alert rule, primarily because of the possible recovery, which is really an alert-specific concept.

server/lib/tuist/automations.ex
fortmarek, line 1: This currently contains just alert rules, but I'd probably still consider keeping alert rules under the automations module. Once we add automations beyond alert rules, we'd want to keep alert rules still under the automations module.

server/lib/tuist_web/controllers/api/automations/alerts_controller.ex
fortmarek, line 1: also somewhat unsure about whether to call this alert rules controller or automations controller. But the shape we currently return is very alert rule like

pepicrft:
> 1. Top-level modeling — AlertRule or Automation?

I'm inclined to use "Automations" as an umbrella platform capability, and then call this particular type of automation, "Alert automation". Down the line we can then introduce another kind of automation, for example functions, that's a scriptable and more generic version for teams that need more.

> 2. API endpoint shape — /alert_rules or /automations?

I think we can do `/api/projects/:account/:project/automations/alerts`

> 3. User-facing terminology — "Automations" or "Alert rules" in the UI?
I'm inclined to keep using automations. Today, we know automations and alert rules mean the same thing, but down the line, we'll have sub-categories. At the API level, I'd already established that hierarchy, we'll have to introduce breaking changes there.

> Authorization reuses the :project object instead of a dedicated :automation one — write ops become admin-only, reads stay open to user/admin/ops/account tokens.

What's the scope granularity that we are using for other project-scoped objects? If we have scope per sub-object, I'm incline to embrace the same model here.

mise/tasks/claude/bootstrap.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+#MISE description="Bootstrap this worktree for Claude Code (launch.json + server install)"
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+WORKTREE_ROOT="$(git rev-parse --show-toplevel)"
```
pepicrft: We can use `MISE_PROJECT_ROOT`.

mise/tasks/claude/regenerate-launch-json.sh
```diff
@@ -0,0 +1,55 @@
...
+mkdir -p "${LAUNCH_DIR}"
+cat > "${LAUNCH_FILE}" <<EOF
+{
+  "version": "0.0.1",
+  "configurations": [
+    {
+      "name": "Tuist Server",
+      "runtimeExecutable": "bash",
+      "runtimeArgs": ["-c", "cd server && exec mise run dev"],
+      "port": ${server_port}
```
pepicrft, line 39: Having to write this file because `port` can't be dynamic is a bit annoying. I'd consider opening an issue or contacting them somehow to support something like `port: "$ENV_VARIABLE"`

fortmarek:
> What's the scope granularity that we are using for other project-scoped objects? If we have scope per sub-object, I'm incline to embrace the same model here.

Can change that, yeah, this feels very tied to settings, but we can make a scope for the new entity.

server/test/tuist/automations/actions/change_state_action_test.exs
```diff
@@ -0,0 +1,26 @@
...
+  describe "execute/2" do
+    test "calls Tests.update_test_case with the requested state" do
+      test_case_id = Ecto.UUID.generate()
+      entity = %{type: :test_case, id: test_case_id}
+
+      expect(Tests, :update_test_case, fn ^test_case_id, %{state: "muted"} ->
+        {:ok, %{state: "muted"}}
+      end)
+
+      assert :ok = ChangeStateAction.execute(entity, %{"state" => "muted"})
```
pepicrft: I'd adjust these action tests to be more of an integration test to be more realistic. In other words, integrate all the way down to the DB operation, and just stub side effects. Otherwise we can easily change something in the contract of `Tests.update_test_case` causing things to break even though tests are passing.

server/priv/repo/migrations/20260411120000_backfill_automations_from_project_settings.exs
```diff
@@ -0,0 +1,93 @@
+defmodule Tuist.Repo.Migrations.BackfillAutomationsFromProjectSettings do
+  use Ecto.Migration
+
+  import Ecto.Query
+
+  def up do
+    projects_with_flaky_detection =
```
pepicrft: Considering the volume of this will be low, I think it's fine to run this in one shot, no?

server/test/support/tuist_test_support/fixtures/automations_fixtures.ex
```diff
@@ -0,0 +1,34 @@
+defmodule TuistTestSupport.Fixtures.AutomationsFixtures do
+  @moduledoc false
+
+  alias Tuist.Automations
+  alias TuistTestSupport.Fixtures.ProjectsFixtures
+
+  def automation_fixture(opts \\ []) do
```
pepicrft:
```suggestion
  def automation_alert_rule_fixture(opts \\ []) do
```

server/lib/tuist_web/plugs/api/authorization/authorization_plug.ex
```diff
@@ -14,12 +14,13 @@ defmodule TuistWeb.API.Authorization.AuthorizationPlug do
...
   def init(:test), do: :test
   def init(:build), do: :build
+  def init(:project), do: :project
 
   def init(opts) when is_list(opts) do
     opts
   end
 
-  @project_categories [:run, :bundle, :cache, :preview, :test, :build]
+  @project_categories [:run, :bundle, :cache, :preview, :test, :build, :project]
```
pepicrft: I'd treat automations as another object that belongs to project:


```suggestion
  @project_categories [:run, :bundle, :cache, :preview, :test, :build, :automation]
```

server/lib/tuist_web/controllers/api/alert_rules_controller.ex
```diff
@@ -0,0 +1,325 @@
...
+
+  defp format_errors(changeset) do
+    changeset
+    |> Ecto.Changeset.traverse_errors(fn {msg, opts} ->
+      Regex.replace(~r"%{(\w+)}", msg, fn _, key ->
+        opts |> Keyword.get(String.to_existing_atom(key), key) |> to_string()
+      end)
+    end)
+    |> Enum.map_join(", ", fn {field, errors} -> "#{field}: #{Enum.join(errors, ", ")}" end)
+  end
```
pepicrft: The scoping of this logic feels wrong. I don't think they should be at the `alert_rules_controller.ex` level.

server/lib/tuist_web/api/schemas/test_case.ex
```diff
@@ -69,7 +70,13 @@ defmodule TuistWeb.API.Schemas.TestCase do
...
       is_quarantined: %Schema{
         type: :boolean,
-        description: "Whether the test case is quarantined"
+        deprecated: true,
+        description: "Whether the test case is quarantined. Deprecated: use `state` instead."
+      },
+      state: %Schema{
+        type: :string,
+        enum: ["enabled", "muted"],
+        description: "The state of the test case."
```
pepicrft: Do we need to explain what the different states represent?

server/lib/tuist_web/api/schemas/automation_alert_action.ex
```diff
@@ -0,0 +1,26 @@
...
+      state: %Schema{type: :string, enum: ["enabled", "muted"], description: "Required for change_state actions."},
+      label: %Schema{type: :string, description: "Label name. Required for add_label and remove_label actions."},
+      channel: %Schema{type: :string, description: "Slack channel ID. Required for send_slack actions."},
+      channel_name: %Schema{type: :string, description: "Slack channel name for display."},
+      message: %Schema{
+        type: :string,
+        description: "Message template. Required for send_slack actions. Supports {{variable}} interpolation."
+      }
+    },
+    required: [:type]
```
pepicrft, line 34: Is this the only required field?

server/lib/tuist_web/controllers/api/alert_rules_controller.ex
```diff
@@ -0,0 +1,325 @@
...
+  alias TuistWeb.API.Schemas.AlertRuleAction
+  alias TuistWeb.API.Schemas.Error
+
+  plug(TuistWeb.Plugs.CastAndValidate,
+    json_render_error_v2: true,
+    render_error: TuistWeb.RenderAPIErrorPlug
+  )
+
+  plug(TuistWeb.Plugs.LoaderPlug)
+  plug(TuistWeb.API.Authorization.AuthorizationPlug, :project)
```
pepicrft: The creation of alert rules here is gated by the `project_create`, while the liveview is gated by `project_update`, which means a non-admin can create alert rules via the API even though they can't via the UI.

server/lib/tuist/automations/workers/automation_scheduler.ex
```diff
@@ -0,0 +1,41 @@
...
+  use Oban.Worker, max_attempts: 1, queue: :default
+
+  import Ecto.Query
+
+  alias Tuist.Automations.AlertRule
+  alias Tuist.Automations.Workers.AlertEvaluationWorker
+  alias Tuist.Repo
+
+  @impl Oban.Worker
+  def perform(_job) do
```
pepicrft, line 12: In the time window that we are using here (1m), we are only using `available`, `scheduled`, and `executing` jobs to check for uniqueness. This means, even if the cadence is 5m, we might end up running an alert rule again.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -0,0 +1,123 @@
...
+        alert_rule_id: alert_rule.id,
+        test_case_id: alert.test_case_id,
+        status: "recovered",
+        triggered_at: now,
+        recovered_at: now
+      })
+
+      entity = %{type: :test_case, id: alert.test_case_id}
+
+      case ActionExecutor.execute_actions(alert_rule.recovery_actions, alert_rule, entity) do
```
pepicrft: If this fails, we've already appended the `recovered` status to the alert rule, leaving the state machine inconsistency. That means in the UI it'll look as resolved, even though the unmute, label cleanup, or Slack notification did not happen.

server/lib/tuist/automations/actions/send_slack_action.ex
```diff
@@ -0,0 +1,72 @@
...
+    account_name = project.account.name
+    project_name = project.name
+    test_case_url = "#{base_url}/#{account_name}/#{project_name}/tests/test-cases/#{test_case.id}"
+
+    template
+    |> String.replace("{{test_case.name}}", test_case.name || "")
+    |> String.replace("{{test_case.module_name}}", test_case.module_name || "")
+    |> String.replace("{{test_case.suite_name}}", test_case.suite_name || "")
+    |> String.replace("{{test_case.url}}", test_case_url)
+    |> String.replace("{{automation.name}}", automation.name || "")
```
pepicrft: The interpolated values are included in the markdown without escaping. I'd consider doing escaping here.

pepicrft (approved):

mise/tasks/claude/regenerate-launch-json.sh
fortmarek: don't disagree, but need something that works _now_ 😅 how do you run this in codex?

pepicrft: Yeah, we can keep it like this. I was thinking out loud.

mise/tasks/claude/bootstrap.sh
fortmarek: Switched to `MISE_PROJECT_ROOT` in 64992551d1.

server/test/tuist/automations/actions/change_state_action_test.exs
fortmarek: Reworked the three action tests (`ChangeStateAction`, `AddLabelAction`, `RemoveLabelAction`) as integration tests in 64992551d1 — they now write a real `TestCase` through `IngestRepo.insert_all`, invoke the action, and assert via `Tests.get_test_case_by_id/1`. No more mimic stubs on `Tests.update_test_case`, so drift in the context contract surfaces here.

server/priv/repo/migrations/20260411120000_backfill_automations_from_project_settings.exs
fortmarek: Agreed — dropped the `left_join` + `is_nil(a.id)` guard in 64992551d1. It was defending against re-running the migration, which we don't do; for the single-shot deploy path, the filter was dead weight.

server/test/support/tuist_test_support/fixtures/automations_fixtures.ex
fortmarek: Went with `automation_alert_fixture` in 64992551d1 to match the new `Tuist.Automations.Alerts.Alert` naming.

server/lib/tuist_web/plugs/api/authorization/authorization_plug.ex
fortmarek: Wired `:automation_alert` into `@project_categories` in 64992551d1. It dispatches to the new `object :automation_alert` in `Tuist.Authorization` (admin-only create/update/delete, user+admin+ops+account-token read).

server/lib/tuist_web/controllers/api/alert_rules_controller.ex
fortmarek: Extracted into `TuistWeb.API.RequestParams` (normalize/1, format_errors/1) and moved the `serialize_alert` serializer to `TuistWeb.API.Schemas.AutomationAlert.from_model/1` in 64992551d1.

server/lib/tuist_web/api/schemas/test_case.ex
fortmarek: Filled out the `state` enum description in 64992551d1 to spell out that `enabled` means the test runs and counts toward pass/fail, while `muted` (formerly "quarantined") still runs (so we keep collecting flakiness signal) but failures don't fail the build.

server/lib/tuist_web/api/schemas/automation_alert_action.ex
fortmarek: Only `type` is listed as required at the OpenAPI level because JSON Schema doesn't express conditional requirements cleanly without `oneOf` + discriminator (and that blows up the spec for four action types). The server-side changeset on `Tuist.Automations.Alerts.Alert` enforces the per-type contract (`change_state` needs `state`, `add_label`/`remove_label` need `label`, `send_slack` needs `channel` + `message`) and returns a structured 422. Added a docstring on the schema explaining this in 64992551d1 so the contract is discoverable.

server/lib/tuist_web/controllers/api/alert_rules_controller.ex
fortmarek: Good catch. Aligned both in 64992551d1 — the API now uses the new `:automation_alert` authz object (admin-only write, broader read), and the LiveView uses `:automation_alert_read` on mount so the UI gate matches the API's read gate. Write gates are already consistent via the LiveView's event handlers going through the admin-gated write actions.

server/lib/tuist/automations/workers/automation_scheduler.ex
fortmarek: Fixed in 64992551d1. Added `:completed` to the unique-state set so a fast-running evaluation job (one that drops to `:completed` between scheduler ticks) no longer lets the next tick slip a duplicate job through. With `period: cadence_seconds(cadence)` + `:completed`, re-scheduling waits at least one cadence regardless of how quickly the previous run finished.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
fortmarek: Swapped the order in 64992551d1 — `ActionExecutor.execute_actions` now runs first, and we only append the `"recovered"` event when the actions succeed. On failure we log and keep the alert visually triggered so the next cadence re-attempts.

server/lib/tuist/automations/actions/send_slack_action.ex
fortmarek: Added mrkdwn escaping (`&` → `&amp;`, `<` → `&lt;`, `>` → `&gt;`) to the interpolated `{{test_case.name}}`, `{{test_case.module_name}}`, `{{test_case.suite_name}}`, and `{{automation.name}}` values in 64992551d1. The header block stays `plain_text` so its literal rendering is safe.

PR #10241, opened 2026-04-11
chore(cli): drop SwiftNIO for file-system operations

pepicrft:
## What changed

This PR removes the in-repo `FileSystem/` package and makes Tuist depend on the `tuist/FileSystem` branch that carries the new backend work instead:

- `https://github.com/tuist/FileSystem.git`
- branch: `feat/vendor-swift-file-system-backend`

The root package and the embedded Swift packages now all point to that branch, and the affected lockfiles were updated to the resolved revision.

## Why

`FileSystem` should remain its own package. The Tuist-side change should only consume it, not vendor it again into `tuist/tuist`.

The motivation for the backend work is to move toward a filesystem implementation that handles file descriptors more gracefully under pressure, especially around descriptor lifecycle and exhaustion, while keeping the rollout runtime-switchable.

This keeps the new backend work centralized in `tuist/FileSystem` while still allowing Tuist to opt into it at runtime.

## Runtime behavior

Default behavior remains `SwiftNIO`.

To opt into the vendored `swift-file-system` backend at runtime:

```bash
TUIST_FILESYSTEM_BACKEND=swift-file-system
```

Accepted aliases also include `swift_file_system` and `swiftfilesystem`.

## Related work

- tuist/FileSystem#315

## Validation

Executed successfully:

- `swift package resolve --replace-scm-with-registry`
- `swift package resolve --package-path processor/native/xcactivitylog_nif --replace-scm-with-registry`
- `swift package resolve --package-path xcode_processor/native/xcresult_nif --replace-scm-with-registry`
- `TUIST_FILESYSTEM_BACKEND=swift-file-system tuist generate --no-open`

Additional validation on the `tuist/FileSystem` branch:

- `swift test --replace-scm-with-registry`
- `TUIST_FILESYSTEM_BACKEND=swift-file-system swift test --replace-scm-with-registry`

## Notes

`swift package resolve --package-path cli/Sources/XcodeGraph --replace-scm-with-registry` still fails in this environment because `swiftlang.swift-docc-plugin` requires a registry scope that is not configured locally. That failure is unrelated to the `FileSystem` branch switch.

### DIFF
--- a/cli/Sources/TuistEnvironment/Environment.swift
+++ b/cli/Sources/TuistEnvironment/Environment.swift
@@ -1,3 +1,4 @@
+import FileSystem
 import Foundation
 import Mockable
 import Path
@@ -218,7 +219,7 @@ public struct Environment: Environmenting {
     }
 
     public func currentWorkingDirectory() async throws -> AbsolutePath {
-        return try AbsolutePath(validating: FileManager.default.currentDirectoryPath)
+        return try await FileSystem().currentWorkingDirectory()
     }
 
     private func variable(_ variableName: String) -> String? {
--- a/cli/Sources/TuistEnvironment/Environment.swift
+++ b/cli/Sources/TuistEnvironment/Environment.swift
@@ -1,8 +1,5 @@
-import _NIOFileSystem
-import FileSystem
 import Foundation
 import Mockable
-import NIOCore
 import Path
 
 #if canImport(Glibc)
@@ -221,7 +218,7 @@ public struct Environment: Environmenting {
     }
 
     public func currentWorkingDirectory() async throws -> AbsolutePath {
-        return try await AbsolutePath(validating: _NIOFileSystem.FileSystem.shared.currentWorkingDirectory.string)
+        return try AbsolutePath(validating: FileManager.default.currentDirectoryPath)
     }
 
     private func variable(_ variableName: String) -> String? {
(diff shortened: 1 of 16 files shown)


xcode_processor/native/xcresult_nif/Package.swift
```diff
@@ -17,7 +17,11 @@ let package = Package(
     ],
     dependencies: [
         .package(id: "tuist.Path", from: "0.3.8"),
-        .package(id: "tuist.FileSystem", from: "0.15.0"),
+        .package(
+            name: "FileSystem",
+            url: "https://github.com/tuist/FileSystem.git",
+            branch: "feat/vendor-swift-file-system-backend"
```
pepicrft: I'll revert this once we merge [this](https://github.com/tuist/FileSystem/pull/315)

cli/Sources/TuistEnvironment/Environment.swift
```diff
@@ -221,7 +218,7 @@ public struct Environment: Environmenting {
     }
 
     public func currentWorkingDirectory() async throws -> AbsolutePath {
-        return try await AbsolutePath(validating: _NIOFileSystem.FileSystem.shared.currentWorkingDirectory.string)
+        return try AbsolutePath(validating: FileManager.default.currentDirectoryPath)
```
fortmarek: I wonder if this should be defined in our `FileSystem` lib?

fortmarek (approved):
Let's  not forget to revert all the `Package.swift` and `Package.resolved` changes once the new file system ships.

cli/Sources/TuistEnvironment/Environment.swift
pepicrft: It's already there, so this function can just be a passthrough. I'll update it.

PR #10243, opened 2026-04-11
feat(cli): process test insights remotely

fortmarek:
## Summary
- `tuist inspect test`, `tuist test`, and `tuist xcodebuild test` now default `--mode`/`--inspect-mode` to `remote` for tuist-hosted instances and to `local` for self-hosted ones, so the default matches where it's cheapest to process the xcresult.
- The default is resolved after the config is loaded, so passing `--mode`/`--inspect-mode` explicitly always wins.
- `InspectTestCommand`'s inner `ProcessingMode` enum is replaced by the shared `TestProcessingMode` to avoid duplication.

## Test plan
- [x] `xcsiftbuild build -workspace Tuist.xcworkspace -scheme Tuist-Workspace`
- [x] `TuistKitTests/InspectTestCommandServiceTests` (incl. new `run_defaults_to_remote_for_tuist_dev_url` and `run_explicit_local_mode_overrides_url_default`)
- [x] `TuistKitTests/XcodeBuildTestCommandServiceTests`
- [x] `TuistKitTests/TestServiceTests`

### DIFF
--- a/cli/Sources/TuistKit/Services/TestProcessingMode.swift
+++ b/cli/Sources/TuistKit/Services/TestProcessingMode.swift
@@ -7,17 +7,16 @@ public enum TestProcessingMode: String, Sendable, CaseIterable, ExpressibleByArg
 
     /// Returns the default processing mode for the given Tuist server URL.
     ///
-    /// Defaults to `.remote` for the official Tuist hosts (`tuist.dev`, `staging.tuist.dev`,
-    /// `canary.tuist.dev`) and for `localhost`. All other URLs default to `.local` so
-    /// self-hosted instances continue to parse xcresults locally unless explicitly opted in.
+    /// Defaults to `.remote` for tuist-hosted instances and `.local` for self-hosted ones,
+    /// so self-hosted servers keep parsing xcresults locally unless they explicitly opt in.
     public static func `default`(for url: URL) -> TestProcessingMode {
-        let knownRemoteHosts: Set<String> = [
+        let tuistHostedHosts: Set<String> = [
             "tuist.dev",
             "staging.tuist.dev",
             "canary.tuist.dev",
             "localhost",
         ]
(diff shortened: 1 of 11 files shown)


cli/Sources/TuistKit/Services/TestProcessingMode.swift
```diff
@@ -1,6 +1,24 @@
...
+    ///
+    /// Defaults to `.remote` for tuist-hosted instances and `.local` for self-hosted ones,
+    /// so self-hosted servers keep parsing xcresults locally unless they explicitly opt in.
+    public static func `default`(for url: URL) -> TestProcessingMode {
+        let tuistHostedHosts: Set<String> = [
+            "tuist.dev",
+            "staging.tuist.dev",
+            "canary.tuist.dev",
+            "localhost",
+        ]
```
fortmarek: not super happy about this, but I think it's the best heuristic we have for selecting the right mode for self-hosted vs tuist-hosted teams

cli/Sources/TuistKit/Services/TestProcessingMode.swift
```diff
@@ -1,6 +1,24 @@
...
+
+    /// Returns the default processing mode for the given Tuist server URL.
+    ///
+    /// Defaults to `.remote` for tuist-hosted instances and `.local` for self-hosted ones,
+    /// so self-hosted servers keep parsing xcresults locally unless they explicitly opt in.
+    public static func `default`(for url: URL) -> TestProcessingMode {
+        let tuistHostedHosts: Set<String> = [
+            "tuist.dev",
+            "staging.tuist.dev",
+            "canary.tuist.dev",
```
pepicrft: What if we just check that the URL base domain is `tuist.dev`?

pepicrft (approved):

PR #10245, opened 2026-04-12
fix(server,cache): report discarded Oban jobs to Sentry

pepicrft:
## Summary

Adds `TuistCommon.ObanSentry` — a shared telemetry handler that reports Oban job failures to Sentry only when all retry attempts are exhausted (job state is `:discard`). Wired up in both cache and server.

## Context

The cache `S3TransferWorker` was crashing ~1,200 times/hour due to `cas` records from a canary node, but these errors were invisible because there was no Oban-to-Sentry integration. Rather than adding `cas` backwards compatibility (the canary should be redeployed), this PR adds observability so similar issues are caught early.

## Test plan

- [x] All 489 cache tests pass
- [x] `TuistCommon.ObanSentry` unit tests pass (reports on discard, skips retryable failures)
- [ ] Deploy and verify discarded Oban jobs appear in Sentry
- [ ] Redeploy canary node to stop `cas` record insertion


cache/test/cache/oban_sentry_integration_test.exs
```diff
@@ -0,0 +1,51 @@
...
+
+  defmodule FailingWorker do
+    use Oban.Worker, queue: :test, max_attempts: 1
+
+    @impl Oban.Worker
+    def perform(_job), do: raise("job failed permanently")
+  end
+
+  describe "discarded Oban job" do
+    test "reports to Sentry when a job exhausts all attempts" do
```
pepicrft: I'm doing a full integration test here with Oban & telemetry to ensure we test the contract of the format that Oban uses for the telemetry data and the data that we expect.

tuist_common/test/tuist_common/oban_telemetry_test.exs
```diff
@@ -0,0 +1,71 @@
+defmodule TuistCommon.ObanSentryTest do
+  use ExUnit.Case, async: false
```
pepicrft, line 2: I went with an integration test here to ensure the contract between our handler and the structure of Oban's emitted telemetry event is met.

tuist_common/lib/tuist_common/oban_telemetry.ex
```diff
@@ -0,0 +1,32 @@
...
+  Reports Oban job failures to Sentry only when all retry attempts have been exhausted.
+  """
+
+  def attach do
+    :telemetry.attach(
+      "oban-sentry-discard",
+      [:oban, :job, :exception],
+      &__MODULE__.handle_event/4,
+      :no_config
+    )
```
fortmarek, line 12: would probably use more agnostic naming instead of coupling this to Sentry (across this file)

We might also want to add telemetry not just for worker exceptions, but in general for workers. Right now, the worker analytics are available at tuist.dev/ops/oban, but would be nice to have this data in grafana, too.

fortmarek (approved):

PR #10247, opened 2026-04-12
fix(cli): keep xcstrings in Resources phase for static framework stale detection

pepicrft:
## Summary

Fixes xcstrings stale-string detection for static frameworks (both single-platform and multiplatform). The root cause was threefold:

1. **`PACKAGE_RESOURCE_BUNDLE_NAME`** on the main target causes swift-build's `XCStringsCompiler.shouldCompileCatalog()` to return false, delegating xcstrings compilation to the companion bundle — which has no Swift sources, so all strings appear stale
2. **xcstrings in Sources build phase** doesn't trigger `XCStringsCompiler` in regular xcodeproj builds (only works through PIF/SPM)
3. **xcstrings in the companion bundle** caused conflicting extraction since the companion has no Swift sources

### Changes:
- Keep `.xcstrings` in the main target's **Resources** build phase (not Sources)
- **Exclude** `.xcstrings` from the companion bundle's Resources
- **Remove** `PACKAGE_RESOURCE_BUNDLE_NAME` and `PACKAGE_RESOURCE_TARGET_KIND` from the main target so `XCStringsCompiler` processes the catalog there

## Test plan

- [x] Build compiles successfully
- [x] Manually verified: dynamic framework with multiplatform destinations — `unused_string` correctly marked STALE, used strings not marked
- [x] Manually verified: static framework with multiplatform destinations and patched pbxproj — `unused_string` correctly marked STALE, used strings not marked
- [x] Verified `xcodebuild archive` succeeds for static framework with both xcassets and xcstrings (no regression from removing PACKAGE_RESOURCE_BUNDLE_NAME)

### DIFF
--- a/cli/Sources/TuistGenerator/Generator/ProjectFileElements.swift
+++ b/cli/Sources/TuistGenerator/Generator/ProjectFileElements.swift
@@ -393,7 +393,7 @@ class ProjectFileElements {
     ) throws {
         guard products[targetName] == nil else { return }
         let fileType = try RelativePath(validating: productName).extension.flatMap {
-            Self.filetype(extension: $0)
+            Xcode.filetype(extension: $0)
         }
         let fileReference = PBXFileReference(
             sourceTree: .buildProductsDir,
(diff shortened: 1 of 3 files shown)


fortmarek (approved):

KaiOelfke:
I confirmed with generated_ios_app_with_static_framework_with_xcstrings that all cases from my markdown table and one of my own projects are working as expected with this PR. Thank you!

Ari4ka (approved):

PR #10249, opened 2026-04-12
fix(server): increase processor worker max attempts to survive deployments

fortmarek:
## Summary
- Increases `max_attempts` from 3 to 5 on `ProcessBuildWorker` and `ProcessXcresultWorker`
- Oban's default exponential backoff (`attempt^4 + 2` seconds) with 3 attempts only covers ~100s, which isn't enough to survive a processor service deployment on Render
- With 5 attempts the retry window extends to ~16 minutes, giving plenty of headroom for deployments

Fixes TUIST-B0

## Test plan
- [x] Verified the change is limited to `max_attempts` on both workers
- [ ] Monitor Sentry after next processor deployment to confirm no more `econnrefused` failures exhaust retries


fortmarek:
fwiw, we should move to better handling of redeployments of the processor at some point, but this should help for the time being.

PR #10252, opened 2026-04-13
fix(server): handle inactive Slack report installations

cschmatzler:
Resolves https://tuist.sentry.io/issues/TUIST-B5

This fixes the `MatchError` reported in Sentry when `Tuist.Slack.Workers.ReportWorker` tries to send a scheduled Slack report for an installation that Slack has already marked as inactive.

Previously, the worker expected `SlackClient.post_message/3` to return `:ok`, so a response like `{:error, "account_inactive"}` crashed the job on the pattern match and kept retrying the same invalid installation.

This PR updates the worker to:

- handle `{:error, "account_inactive"}` explicitly instead of crashing
- delete the inactive Slack installation and clear the associated project Slack channel configuration
- return `{:discard, :account_inactive}` so Oban discards the cleanup job instead of retrying it
- keep the report window anchored to the last *completed* report job, so discarded cleanup jobs do not affect the next report period
- add regression coverage for inactive installations, discarded jobs, transient failures, and the report period behavior

It also makes the account fixture derive a valid handle from the provided email so these tests can create stable account data without collisions.

### How to test locally

- cd server
- mix test test/tuist/slack/workers/report_worker_test.exs

### DIFF
--- a/server/lib/tuist/slack/workers/report_worker.ex
+++ b/server/lib/tuist/slack/workers/report_worker.ex
@@ -13,19 +13,22 @@ defmodule Tuist.Slack.Workers.ReportWorker do
   alias Tuist.Projects
   alias Tuist.Projects.Project
   alias Tuist.Repo
+  alias Tuist.Slack
   alias Tuist.Slack.Client, as: SlackClient
   alias Tuist.Slack.Reports
 
+  require Logger
+
   @impl Oban.Worker
   def perform(%Oban.Job{args: %{"project_id" => project_id}}) do
-    project = Projects.get_project_by_id(project_id)
+    case Projects.get_project_by_id(project_id) do
+      nil ->
+        :ok
 
-    if project do
-      project = Repo.preload(project, account: :slack_installation)
-      :ok = send_report(project)
+      project ->
+        project = Repo.preload(project, account: :slack_installation)
+        send_report(project)
     end
-
-    :ok
   end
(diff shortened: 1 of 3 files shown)


server/lib/tuist/slack/workers/report_worker.ex
```diff
@@ -13,19 +13,22 @@ defmodule Tuist.Slack.Workers.ReportWorker do
...
+  alias Tuist.Slack
   alias Tuist.Slack.Client, as: SlackClient
   alias Tuist.Slack.Reports
 
+  require Logger
+
   @impl Oban.Worker
   def perform(%Oban.Job{args: %{"project_id" => project_id}}) do
-    project = Projects.get_project_by_id(project_id)
+    case Projects.get_project_by_id(project_id) do
```
fortmarek, line 24: I understand this can occasionally happen when someone deletes a project, but I'd expect to be very rare? Is there another reason why this worker is being called with a project id that does not exist?

fortmarek (approved):

server/lib/tuist/slack/workers/report_worker.ex
cschmatzler: It's exactly the same behavior as `if project do` before, just written differently because I felt like the `nil -> :ok` shortcut flow is significantly easier to read with `case`.

PR #10261, opened 2026-04-13
feat(cli,gradle): auto-detect HTTP proxy from HTTPS_PROXY/HTTP_PROXY

pepicrft:
## Summary

When `HTTPS_PROXY` or `HTTP_PROXY` is set in the environment, the Tuist CLI and the Tuist Gradle plugin now automatically route the HTTP connections they manage through that proxy. There is nothing for users to configure: the same convention `curl`, `git`, and most developer tools follow.

The change applies to every HTTP client Tuist owns: cache, previews, analytics, registry access, and the calls the Gradle plugin makes back to Tuist services.

## How it works

- **CLI (Swift):** `URLSession+Tuist.swift` reads `HTTPS_PROXY`, `https_proxy`, `HTTP_PROXY`, `http_proxy` (in that order) from `Environment.current.variables` at first access of `URLSession.tuistShared`. On macOS the resolved URL is applied via `connectionProxyDictionary` (`kCFNetworkProxies*` keys, with sensible defaults of 443/80/8080 by scheme).
- **Gradle (Kotlin):** `TuistHttpClients` reads the same variables via `System.getenv(...)`, builds a `java.net.Proxy`, and applies it to the shared OkHttp client and to direct `HttpURLConnection.openConnection` calls.

## What was removed

This PR replaces an earlier explicit-configuration design. The following surfaces were removed entirely:

- `Tuist.swift` `proxy:` parameter and `Tuist.Proxy` enum
- `tuist.toml` `[proxy]` table parsing (CLI + Gradle)
- `settings.gradle.kts` `proxy = ...` DSL plus `Proxy.kt`, `ProxyResolver.kt`, `TuistProxyParameters.kt`
- `URLSession.configureTuistProxy` runtime API and the `applyProxy` plumbing in `ConfigLoader`

Net change: roughly 1300 fewer lines than the explicit-configuration design across CLI, Gradle plugin, and docs.

## Docs

- New dedicated `/guides/integrations/http-proxy` page (sidebar entry under Integrations) describing the env-var lookup order, what is and isn't proxied, and the inline-credentials format.
- `install-tuist.md` and `install-gradle-plugin.md` each have a one-line `## HTTP proxy` section linking to that guide.
- New changelog entry at `server/priv/marketing/changelog/2026.04.16-http-proxy-support.md`.

## Other fixes folded in

- Bumped `dompurify` 3.3.3 to 3.4.0 to clear Trivy advisory GHSA-39q2-94rc-95cp (this PR's `Security` red).

## Test plan

- [x] `xcodebuild build -workspace Tuist.xcworkspace -scheme tuist`: BUILD SUCCEEDED.
- [x] `mise x swiftformat -- swiftformat cli/ app/ --lint`: clean.
- [x] `tuist inspect dependencies --only implicit`: no implicit dependencies.
- [x] `./gradlew test` in `gradle/`: passes (`TuistHttpClientsTest` rewritten to cover the no-proxy default; the explicit `Proxy.Url` tests were removed along with the type).
- [x] `mix compile --warnings-as-errors` in `server/`: clean (sidebar still parses).
- [x] CI green on the proxy-relevant jobs (`Unit Tests`, `Linux Build`, `Linux Unit Tests`, `Lint`, `Build`, `Test` (gradle), `Build Acceptance Tests`, `Gradle Cache Acceptance`, etc.).

## Pre-existing CI reds (not introduced by this PR)

- `Test` (server): 7 failures in `Tuist.Gradle.AnalyticsTest` caused by hardcoded `~N[2026-01-15]` fixtures hitting the 90-day ClickHouse TTL today (2026-04-16). Same on `main`.
- `Docker build`: Trivy detected new ImageMagick CVEs published 2026-04-15 in the base image.

These need their own narrow follow-ups (use `Tuist.Time.utc_now()` in fixtures; bump base image).


server/priv/docs/en/guides/install-gradle-plugin.md
```diff
@@ -66,12 +66,40 @@ The following options are available in the `tuist` extension block in `settings.
 | `executablePath` | `String?` | `null` (uses `tuist` from PATH) | Path to the Tuist CLI executable. |
 | `url` | `String?` | `null` | The base URL of the Tuist server. If not set, it defaults to `"https://tuist.dev"` or the value defined in `tuist.toml`. |
 | `uploadInBackground` | `Boolean?` | `null` | Whether to upload build and test insights in the background. When `null` (default), uploads run in the background for local builds and in the foreground on CI. |
+| `proxy` | `Proxy` | `Proxy.None` | The HTTP proxy the plugin routes its traffic through. See <.localized_link href="#http-proxy">HTTP proxy</.localized_link>. |
```
pepicrft: I don't think the link to #http-proxy here should be `<.localized_link>`

cli/Sources/TuistHTTP/URLSession+Tuist.swift
```diff
@@ -17,21 +40,102 @@ private func tuistURLSessionConfiguration() -> URLSessionConfiguration {
...
+        case let .environmentVariable(name):
+            guard let value = resolveProxyEnvironmentValue(name: name) else {
+                return nil
+            }
+            return URL(string: value)
+        }
+    }
+
+    private func resolveProxyEnvironmentValue(name: String?) -> String? {
+        let env = ProcessInfo.processInfo.environment
```
pepicrft: This should use `Environment.current` to get environment variables so that we can mock it easily from tests.

cli/Sources/TuistHTTP/URLSession+Tuist.swift
```diff
@@ -17,21 +40,102 @@ private func tuistURLSessionConfiguration() -> URLSessionConfiguration {
...
     public static var tuistShared: URLSession {
-        _tuistURLSession
+        _sessionLock.lock()
+        defer { _sessionLock.unlock() }
+        return _tuistURLSession
+    }
+
+    /// Configures the HTTP proxy used by `URLSession.tuistShared`.
+    ///
+    /// Call this once, early in the lifecycle (e.g., right after loading `Tuist.swift`),
```
pepicrft: We should also think about how to make it configurable from the `tuist.toml`, which is also a valid configuration file, and document the schema changes for that.

cli/Sources/TuistHTTP/URLSession+Tuist.swift
```diff
@@ -7,7 +7,30 @@ import Foundation
...
+
+    /// Read the proxy URL from an environment variable.
+    ///
+    /// When `name` is `nil`, Tuist reads `HTTPS_PROXY` and falls back to `HTTP_PROXY`.
+    /// Both uppercase and lowercase variants are checked.
+    case environmentVariable(String?)
+
+    /// Use the given proxy URL directly.
+    case url(URL)
+}
```
pepicrft: Why do we need an additional symbol here? I was thinking we can resolve the value when the configuration is loaded, and then initialize the session configuration with it. Otherwise, we have many instances of a similar type.

cli/Sources/TuistConfigLoader/SwiftConfigLoader.swift
```diff
@@ -59,9 +61,16 @@
                 at: configPath
             )
             cachedConfigs[path] = config
+            applyProxy(from: config)
             return config
         }
 
+        /// Applies the proxy configured in `Tuist.swift` to the shared URLSession so that
+        /// every subsequent network request Tuist makes goes through it.
+        private func applyProxy(from config: TuistConfig.Tuist) {
```
pepicrft: What about the loading of `tuist.toml`?

server/priv/docs/en/guides/install-gradle-plugin.md
pepicrft: Fixed — it is now a plain Markdown `[HTTP proxy](#http-proxy)` link, since the anchor is on the same page and not cross-document. (08975481fa)

cli/Sources/TuistHTTP/URLSession+Tuist.swift
pepicrft: Done, but indirectly — I deleted the env-var lookup from TuistHTTP altogether (see the "additional symbol" thread below). Resolution now lives in `TuistConfig.Tuist.Proxy.resolvedURL(environment:)`, which takes the env dict as a parameter. `ConfigLoader` calls it with `Environment.current.variables`, so tests can mock by passing any `[String: String]`. (08975481fa)

cli/Sources/TuistHTTP/URLSession+Tuist.swift
pepicrft: Done on both sides. `tuist.toml` now accepts a `[proxy]` table with exactly one of `url` / `environment_variable`, enforced at decode time:

```toml
project = "org/project"

[proxy]
url = "http://proxy.corp:8080"
```

The Swift side validates in `TuistTomlProxy.init(from:)`. The Gradle side reads the same table via `TomlParser` and merges it with the DSL-provided proxy in `ProxyResolver` (DSL wins, TOML is the fallback). Both docs pages now document the schema. (08975481fa)

cli/Sources/TuistHTTP/URLSession+Tuist.swift
pepicrft: Good call — `TuistHTTPProxy` is gone. `URLSession.configureTuistProxy(_ proxyURL: URL?)` now takes a plain URL. Resolution (env var → URL) lives once in `TuistConfig.Tuist.Proxy.resolvedURL(environment:)` and `ConfigLoader` calls it before handing the URL to the HTTP layer. No parallel symbol tree to keep in sync. (08975481fa)

cli/Sources/TuistConfigLoader/SwiftConfigLoader.swift
pepicrft: Moved the proxy application out of `SwiftConfigLoader` and into `ConfigLoader.loadConfig`, which is the single entry point for both `Tuist.swift` and `tuist.toml`. Whichever loader produces the config, the same `applyProxy` call runs afterwards. (08975481fa)

cli/Tests/TuistConfigLoaderTests/TuistTomlConfigLoaderTests.swift
```diff
@@ -93,6 +93,88 @@ struct TuistTomlConfigLoaderTests {
...
+
+        let rootDirectoryLocator = MockRootDirectoryLocating()
+        given(rootDirectoryLocator).locate(from: .any).willReturn(temporaryDirectory)
+
+        let subject = TuistTomlConfigLoader(
+            fileSystem: fileSystem,
+            rootDirectoryLocator: rootDirectoryLocator
+        )
+
+        await #expect(throws: (any Error).self) {
```
pepicrft: Validate the type of error that's being thrown.

cli/Sources/ProjectDescription/Tuist.swift
```diff
@@ -41,6 +41,87 @@ public struct Tuist: Codable, Equatable, Sendable {
...
+    /// are supported:
+    ///
+    /// - ``none`` (default): Tuist makes requests directly.
+    /// - ``environmentVariable(_:)``: Tuist reads the proxy URL from an environment variable.
+    ///   When the parameter is `nil` (the default), Tuist reads `HTTPS_PROXY` at runtime —
+    ///   matching the convention used by `curl`, `git`, and most developer tools. Pass a
+    ///   different name (e.g. `"HTTP_PROXY"` or a custom variable) to read somewhere else.
+    /// - ``url(_:)``: Tuist uses the proxy URL you pass directly. Include credentials inline
+    ///   if the proxy requires authentication: `http://user:password@proxy.corp:8080`.
+    public enum Proxy: Codable, Equatable, Sendable, ExpressibleByStringLiteral {
```
fortmarek: I'm honestly not a fan of making the proxy configurable in the `Tuist.swift` with possibly custom environment variables. Should we consider supporting just the conventional `HTTPS_PROXY`/`HTTP_PROXY`? I can't think of a good use case for this manifest addition.

And the same for Gradle, I'd consider reading this env variables instead of allowing to codify this in the manifest.

cli/Sources/TuistHTTP/URLSession+Tuist.swift
```diff
@@ -7,31 +7,74 @@ import Foundation
     import TuistHAR
 #endif
 
+private let _tuistURLSession: URLSession = makeTuistURLSession()
+
+private func environmentProxyURL() -> URL? {
+    let environment = ProcessInfo.processInfo.environment
```
pepicrft: I'm replacing this with Environment.current.environmentVariables

cli/Sources/ProjectDescription/Tuist.swift
pepicrft: I changed the approach to be opt-in when the variables are present.

gradle/src/main/kotlin/dev/tuist/gradle/TuistHttpClients.kt
```diff
@@ -0,0 +1,109 @@
...
+import okhttp3.OkHttpClient
+import retrofit2.Retrofit
+import retrofit2.converter.gson.GsonConverterFactory
+import java.net.HttpURLConnection
+import java.net.InetSocketAddress
+import java.net.URI
+import java.util.concurrent.TimeUnit
+
+/**
+ * Single source of truth for the HTTP machinery the Tuist Gradle plugin uses.
```
fortmarek, line 12: HTTP machinery is a really weird way to put it 😅

server/priv/docs/en/guides/integrations/http-proxy.md
fortmarek, line 1: not sure if I would put this into `integrations`, but I also understand we don't have a good section for this. Maybe later on, we could have a `Networking` section?

fortmarek (approved):

server/priv/docs/en/guides/integrations/http-proxy.md
pepicrft: I suggest we keep it here, and once we iterate on the docs, we can find a home for it.

PR #10267, opened 2026-04-13
feat(server): show account dropdown in docs nav

pepicrft:
## Summary
- Surfaces the dashboard's account dropdown in the docs top nav when the user is authenticated, reusing the existing `TuistWeb.AccountDropdown` function component
- Makes the dropdown's primary action configurable via a new `primary_action` attr (`account_settings` default, `dashboard` for docs) and adds `log_out_return_to` so logout keeps the user on the current docs page
- Moves `account_dropdown.css` to `server/assets/shared/css/components/` and wires it into both the app and docs bundles; adds the `ThemeSwitcher` hook to `docs.js` so the theme picker works outside the dashboard
- Threads `return_to` through the login redirect URL in addition to the session so the round trip from docs login works even if the session state hiccups

image

> [!NOTE]
> In a follow-up PR I'll add the project selector, a piece of information that the documentation context will make use of :)


server/assets/docs/docs.js
```diff
@@ -3,6 +3,7 @@ import { Socket } from "phoenix";
 import { LiveSocket } from "phoenix_live_view";
 import topbar from "../app/js/vendor/topbar.js";
 import Noora from "noora";
+import ThemeSwitcher from "../app/js/ThemeSwitcher.js";
```
fortmarek: we brought in the theme switcher, but in the docs, changing to dark mode doesn't do anything, does it?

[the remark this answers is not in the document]
pepicrft: Good catch — pushed 60f28c2548 calling observeThemeChanges() before the LiveSocket is created, matching app.js.

fortmarek (approved):

server/assets/docs/docs.js
pepicrft: I'll hide it until we support dark mode in the docs.

PR #10268, opened 2026-04-14
feat(cli): preserve test targets for local SPM packages

sabade-omkar:
## Summary

Local Swift packageshave their test targets incorrectly filtered out during project generation. This happens because `SwiftPackageManagerGraphLoader` passes all SPM packages as 
`PackageType.external` regardless of whether they are local path dependencies or remote, so `PackageInfoMapper` and downstream graph mappers treat them identically, dropping all test targets.

## Fix

Propagate package origin through the pipeline so local package test targets are preserved while remote ones are correctly filtered:

 1. `PackageType.ExternalOrigin` — new enum (`.local` / `.remote`) added to `PackageType.external`
 2. `SwiftPackageManagerGraphLoader` — classifies dependency kind (`local, fileSystem, localSourceControl → .local; remoteSourceControl, registry → .remote`) and passes origin to the mapper
 3. `PackageInfoMapper` — skips test targets only for remote packages; tags local package test targets with `tuist:local-swift-package-test` metadata for downstream mappers
 4. `PruneOrphanExternalTargetsGraphMapper` — checks the metadata tag instead of `product == .unitTests`, so remote test targets are correctly pruned while local ones are preserved
 5. `FocusTargetsGraphMappers` — preserves both local project unit tests and tagged local package tests
 
## How to test locally

 1. Add a local Swift package with a test target as a path dependency
 2. Run `tuist generate` and open the workspace
 3. Verify the local package's test target appears in the Xcode project and can be built/run
 4. Verify remote package test targets do not appear


fortmarek:
## Review changes

I pushed several commits on top of this PR to address review feedback. Here's a summary of all changes and the reasoning behind them:

### 1. Remove broader behavioral change in FocusTargetsGraphMappers

The original PR preserved all local unit tests (`project.type == .local`) when focusing on specific targets, not just local SPM package test targets. This changed existing focus behavior -- `tuist generate App` would no longer prune unrelated local test targets. Reverted to original behavior: only tagged local SPM test targets get special treatment.

### 2. Remove unrelated changes

Reverted the `processor/native/xcactivitylog_nif/Package.resolved` and `Package.swift` changes (FileSystem version bump) since they're unrelated to this fix.

### 3. Restore relaxed test assertion

The `map_whenNameContainsUnderscoreAndPeriod` test was relaxed from a full structural comparison to just checking target names. Restored the original full assertion.

### 4. Move local SPM test target pruning to FocusTargetsGraphMappers

During e2e testing we discovered that the `PruneOrphanExternalTargetsGraphMapper` tag exemption was never reached -- the `FocusTargetsGraphMappers` (which runs first in the `automation` mapper chain with `excludingExternalTargets: true`) was marking local SPM test targets as prunable, and `TreeShakePrunedTargetsGraphMapper` was removing them before the orphan pruner could save them.

Restructured the responsibility:

- **`PruneOrphanExternalTargetsGraphMapper`**: now completely skips tagged local SPM test targets (early return). It no longer decides their fate.
- **`FocusTargetsGraphMappers`**: the single decision point. Preserves local SPM test targets only when (a) there are no explicit filters (plain `tuist generate`) and (b) at least one of their direct dependencies is in the filtered graph. When focusing on specific targets (`tuist generate App`), they are pruned like any other unfocused target.
- **`ExternalProjectsPlatformNarrowerGraphMapper`**: local SPM test targets are orphans (nothing depends on them), so the top-down platform traversal never reaches them. Added logic to inherit narrowed destinations from their direct dependencies.

### E2E verification

Built the `tuist` executable from this branch (`swift build --configuration debug --replace-scm-with-registry --product tuist`) and created a test project with the following structure:

**`Tuist/Package.swift`** -- declares a local path dependency:
```swift
let package = Package(
    name: "App",
    dependencies: [
        .package(path: "../LocalPackage"),
    ]
)
```

**`LocalPackage/Package.swift`** -- local package with a test target:
```swift
let package = Package(
    name: "LocalPackage",
    platforms: [.iOS(.v16)],
    products: [
        .library(name: "LocalLib", targets: ["LocalLib"]),
    ],
    targets: [
        .target(name: "LocalLib"),
        .testTarget(name: "LocalLibTests", dependencies: ["LocalLib"]),
    ]
)
```

**`Project.swift`** -- iOS app depending on the local package, plus a standalone framework with no dependency on it:
```swift
let project = Project(
    name: "App",
    targets: [
        .target(
            name: "App",
            destinations: .iOS,
            product: .app,
            bundleId: "dev.tuist.App",
            infoPlist: .default,
            sources: ["App/Sources/**"],
            dependencies: [.external(name: "LocalLib")]
        ),
        .target(
            name: "Standalone",
            destinations: .iOS,
            product: .framework,
            bundleId: "dev.tuist.Standalone",
            infoPlist: .default,
            sources: ["Standalone/Sources/**"]
        ),
    ]
)
```

**Test results:**

| Command | `LocalLibTests` in generated project? | Expected? |
|---|---|---|
| `tuist generate --no-open` | Yes (24 references in pbxproj) | Yes -- unfocused generate includes all local package tests |
| `tuist generate App --no-open` | No (0 references) | Yes -- focusing on App should not include unrelated test targets |
| `tuist generate Standalone --no-open` | No (LocalPackage project not generated at all) | Yes -- Standalone has no dependency on LocalPackage |

pepicrft (approved):

PR #10274, opened 2026-04-14
fix(server): preserve build run list state

cschmatzler:
## Summary
- preserve the current query string when navigating from Xcode and Gradle build run lists into a build run detail page
- keep the same query parameters on the detail-page back actions so returning to the list restores the active filters and pagination state
- add a shared query helper and targeted LiveView coverage for both Xcode and Gradle flows

## Validation
- `cd server && mix test test/tuist_web/live/build_runs_live_test.exs test/tuist_web/live/build_run_live_test.exs test/tuist_web/live/gradle_build_runs_live_test.exs test/tuist_web/live/gradle_build_live_test.exs`
- `cd server && mix format --check-formatted`
- `cd server && mix credo`

### DIFF
--- a/server/lib/tuist_web/live/build_run_live.ex
+++ b/server/lib/tuist_web/live/build_run_live.ex
@@ -141,6 +141,10 @@ defmodule TuistWeb.BuildRunLive do
     {:noreply, socket}
   end
 
+  def build_runs_path(%{selected_project: project, uri: uri}) do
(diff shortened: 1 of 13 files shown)


cschmatzler:
* **fix(server): preserve build run list state #10274 👈**
* `main`

---
This stack of pull requests is managed by [jj-ryu](https://github.com/dmmulroy/jj-ryu).

fortmarek (approved):
Thanks 🙏

PR #10278, opened 2026-04-14
fix(server,cache): harden NixOS platform (docker liveRestore, drop root SSH, SSH-config-driven deploys)

fortmarek:
## Summary

Three hardening changes to the processor and cache NixOS platforms, prompted by an incident where `colmena switch` on prod killed the processor container and Kamal didn't auto-restart it.

- **Enable `virtualisation.docker.liveRestore = true`** so dockerd restarts (triggered by `colmena switch`) no longer kill running containers. Without this, every NixOS config apply caused ~30–60s of app downtime and left containers in a `hasBeenManuallyStopped=true` state that blocks Kamal auto-restart.
- **Drop the root-only `cschmatzler@tahani` key** from `authorizedKeys`. It was a nixos-anywhere bootstrap artifact that never got rotated. All legitimate operators (mfort, cschmatzler, pedro) are already in `wheel` with passwordless sudo, so Colmena can deploy via any of them.
- **Set `deployment.targetUser = null`** so Colmena defers to each operator's `~/.ssh/config` for the SSH user. Keeps the flake pure (no `--impure`), sudo handles privilege escalation. Operators add a one-liner to their SSH config (see updated handbook).

Already applied live to processor/processor-canary/processor-staging; cache hosts still need `colmena apply` to pick up `liveRestore`. One-time container restart per host is expected when enabling liveRestore; after that, future switches are transparent.

## Test plan

- [x] `colmena apply --on processor-staging` — Live Restore Enabled: true, app healthy
- [x] `colmena apply --on processor-canary` — same
- [x] `colmena apply --on processor` — same
- [ ] Apply to cache hosts (`cache/platform/`) after merge
- [ ] Verify future `colmena switch` on these hosts no longer restarts containers

### DIFF
--- a/cache/platform/configuration.nix
+++ b/cache/platform/configuration.nix
@@ -163,6 +163,7 @@
   virtualisation.docker = {
     enable = true;
     logDriver = "json-file";
+    liveRestore = true;
     daemon.settings = {
       "default-ulimits" = {
         nofile = {
--- a/cache/platform/flake.nix
+++ b/cache/platform/flake.nix
@@ -49,6 +49,11 @@
     mkColmenaNode = diskConfig: hostname: {
       deployment = {
         targetHost = "${hostname}.tuist.dev";
+        # Defer SSH user to the operator's ~/.ssh/config so the flake stays
+        # pure (no --impure / env vars). Privilege escalation via passwordless
+        # sudo for wheel users.
+        targetUser = null;
+        privilegeEscalationCommand = ["sudo" "-H" "--"];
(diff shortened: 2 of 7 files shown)


cli/Sources/TuistKit/Services/TestService.swift
pepicrft, line 1040: These changes are unrelated, no?

pepicrft (approved):

pepicrft (approved):

PR #10282, opened 2026-04-15
fix(cli): Add warning for missing xctestplan files

yhkaplan:
Was investigating why generated xcschemes didn't turn out the way I wanted to and found that the xctestplan I generate on CI wasn't generated at the correct timing. This would have saved me time.

### How to test locally

1. Run unit tests
2. Create a Project.swift that references an xctestplan that's missing and run tuist generate

### DIFF
--- a/cli/Sources/TuistLoader/Models+ManifestMappers/TestAction+ManifestMapper.swift
+++ b/cli/Sources/TuistLoader/Models+ManifestMappers/TestAction+ManifestMapper.swift
@@ -3,6 +3,7 @@ import Foundation
 import Path
 import ProjectDescription
 import TuistCore
+import TuistLogging
 import TuistSupport
 import XcodeGraph
 
@@ -61,13 +62,17 @@ extension XcodeGraph.TestAction {
                     }
                 } else {
                     // Handle as literal path
-                    if try await fileSystem.exists(resolvedPath) && resolvedPath.extension == "xctestplan" {
-                        let testPlan = try await TestPlan.from(
-                            path: resolvedPath,
-                            isDefault: resolvedTestPlans.isEmpty,
-                            generatorPaths: generatorPaths
-                        )
-                        resolvedTestPlans.append(testPlan)
+                    if try await fileSystem.exists(resolvedPath) {
+                        if resolvedPath.extension == "xctestplan" {
+                            let testPlan = try await TestPlan.from(
+                                path: resolvedPath,
+                                isDefault: resolvedTestPlans.isEmpty,
+                                generatorPaths: generatorPaths
+                            )
+                            resolvedTestPlans.append(testPlan)
+                        }
+                    } else {
+                        Logger.current.warning("\(resolvedPath.pathString) does not exist")
                     }
                 }
             }
--- a/cli/Sources/TuistLoader/Models+ManifestMappers/TestAction+ManifestMapper.swift
+++ b/cli/Sources/TuistLoader/Models+ManifestMappers/TestAction+ManifestMapper.swift
@@ -3,7 +3,7 @@ import Foundation
 import Path
 import ProjectDescription
 import TuistCore
-import TuistLogging
+import TuistAlert
 import TuistSupport
 import XcodeGraph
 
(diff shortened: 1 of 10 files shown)


cli/Sources/TuistLoader/Models+ManifestMappers/TestAction+ManifestMapper.swift
```diff
@@ -61,13 +62,17 @@ extension XcodeGraph.TestAction {
...
+                        if resolvedPath.extension == "xctestplan" {
+                            let testPlan = try await TestPlan.from(
+                                path: resolvedPath,
+                                isDefault: resolvedTestPlans.isEmpty,
+                                generatorPaths: generatorPaths
+                            )
+                            resolvedTestPlans.append(testPlan)
+                        }
+                    } else {
+                        Logger.current.warning("\(resolvedPath.pathString) does not exist")
```
pepicrft: Let's use the APIs in `AlertController.current` instead. This utility acts as a collector, ensuring these things are printed at the end of the command execution. I'd also add a bit of more context to help the person find out where the reference is, for example the scheme whose test action is referencing this test plan.

pepicrft (approved):
This looks good to me. Thoughts @fortmarek ?

cli/Sources/TuistLoader/Models+ManifestMappers/TestAction+ManifestMapper.swift
```diff
@@ -61,13 +66,20 @@ extension XcodeGraph.TestAction {
...
+                        }
+                    } else {
+                        let schemeContext = schemeName.map { " referenced by the scheme '\($0)'" } ?? ""
+                        AlertController.current.warning(
+                            .alert("Test plan \(resolvedPath.basename) does not exist at \(resolvedPath.pathString)\(schemeContext)")
                         )
-                        resolvedTestPlans.append(testPlan)
                     }
                 }
             }
```
joeThomas311292, line 87: testPlans = resolvedTestPlans two lines below assigns [] when all plans were skipped. That empty array is fundamentally different from nil. Because testAction.testPlans becomes Optional([]), the ?.map in SchemeDescriptorsGenerator returns Optional([]) (non-nil but empty), and XcodeProj's serializer faithfully emits <TestPlans></TestPlans> with no children — causing xcodebuild to silently run zero tests.
Should this be:
testPlans = resolvedTestPlans.isEmpty ? nil : resolvedTestPlans
So XcodeProj omits <TestPlans> entirely rather than emitting it empty?

cli/Tests/TuistLoaderTests/Models+ManifestMappers/TestAction+ManifestMapperTests.swift
```diff
@@ -185,6 +187,35 @@ struct TestActionManifestMapperTests {
...
+
+        // When
+        let testAction = try await XcodeGraph.TestAction.from(
+            manifest: manifest,
+            generatorPaths: generatorPaths,
+            schemeName: schemeName
+        )
+
+        // Then
+        #expect(testAction.testPlans?.isEmpty == true)
```
joeThomas311292: testAction.testPlans == nil

cli/Sources/TuistLoader/Models+ManifestMappers/TestAction+ManifestMapper.swift
joeThomas311292: Shouldn't this be a defensive guard in[ SchemeDescriptorsGenerator.swift line 294 ](https://github.com/yhkaplan/tuist/blob/c876de099f1c73b9f2a9750e559e820e394e679d/cli/Sources/TuistGenerator/Generator/SchemeDescriptorsGenerator.swift#L294). Even with the nil-guard, the generator is fragile — ?.map propagates an empty array as Optional([]) rather than collapsing to nil. 

Shouldn't this:
// current
```
let testPlans: [XCScheme.TestPlanReference]? = testAction.testPlans?.map {
    XCScheme.TestPlanReference(reference: "container:\($0.path.relative(to: rootPath))", default: $0.isDefault)
}
```
 be changed to
// proposed
```
let testPlans: [XCScheme.TestPlanReference]? = testAction.testPlans.flatMap { plans in
    guard !plans.isEmpty else { return nil }
    return plans.map {
        XCScheme.TestPlanReference(reference: "container:\($0.path.relative(to: rootPath))", default: $0.isDefault)
    }
}
```
flatMap on an Optional returns nil when the closure returns nil, so an empty array cleanly becomes nil and XcodeProj never creates the element.
Without both of these, the warning fires but the broken xcscheme is still written — Xcode sees <TestPlans></TestPlans> and silently runs no tests. The warning is a big improvement for discoverability, but the output remains incorrect.

joeThomas311292 (approved):

fortmarek (approved):

fortmarek:
Thanks @yhkaplan!

did a couple of minor adjustments based on @pepicrft's and @joeThomas311292's feedback 😌

PR #10287, opened 2026-04-15
feat(server): remember per-tab query state for sidebar navigation

fortmarek:
## Summary
Remember the last-used query string on dashboard pages (build runs, test runs, module cache runs, etc.) per browser tab so that sidebar navigation and detail-page back buttons restore the user's filters, sort, pagination, and selected tab.

Addresses the "Losing Filters in Detail Pages" and "Losing filter state when navigating via the sidebar" feedback Roland raised in Slack.

Supersedes #10274 — the URL-propagation approach introduced there is replaced by the mechanism described below, which handles sidebar navigation as well as the detail ↔ list flow.

## Design
A single JS module (`assets/app/js/QueryMemory.js`, ~35 lines) does all the work:

- **Remember**: on `phx:navigate` (server `push_patch`, link navigations, popstate) and on `phx:replace-url` (the app-custom silent URL update used on a few list pages), plus once on initial page load, if the current page is under a `[data-query-memory]` marker, write `sessionStorage["tuist:query-memory:" + pathname] = location.search.slice(1)`.
- **Restore**: a document-level capture-phase `click` listener runs before LiveView's own click handler. If the target is a bare `<a href="/...">` (no existing query) and we have a remembered query string for that pathname, it rewrites the `href` in place. LiveView then reads the rewritten `href` and navigates with the restored query.

**Per-tab isolation** is inherited from `sessionStorage` (already per tab natively — no `tab_id` plumbing required).

**Scope** is a single `data-query-memory` attribute on the `<main>` element in `layouts/project.html.heex`, covering every route under the `:project` live_session. New dashboard pages get the behavior for free; new links need zero annotation.

## Opt-in / opt-out granularity
The marker is intentionally only at the project layout root. We considered finer-grained control — a per-page `data-query-memory` opt-in on each list template, or a per-link `data-no-query-memory` opt-out — but neither solved a problem we actually have today. Every dashboard page's query string is something a user would want restored, and nothing outside the project layout needs the behavior. We can revisit if a concrete case shows up (e.g. a dashboard route whose query params shouldn't be sticky, or a link that needs to force a clean URL).

## Why not session/cookie/localStorage or a server-side cache?
- **Plug session / cookie**: shared across tabs, grows the cookie.
- **localStorage**: filters silently resurrect across sessions, which is surprising.
- **Server-side Cachex / Redis keyed by `{user_id, tab_id}`** (the earlier iteration of this PR): since `tab_id` itself lives in `sessionStorage`, the server entry's effective lifetime is capped at the tab's lifetime anyway. All the server-side machinery (on_mount hook, routes map, push_event, connect param plumbing, per-link JS hook) produced no user-visible advantage over `sessionStorage` alone. The refactor on this branch deletes ~300 lines for that reason.
- **`sessionStorage` + document-level interceptor**: same lifetime, zero server state, zero per-link annotation, a single opt-in point at the layout.

## Test plan
- [x] `mix test test/tuist_web/live/{build_runs,build_run,gradle_build_runs,gradle_build,layout}_live_test.exs` — 35/35 pass.
- [ ] Smoke test in `mise run dev`: set filters on `/builds/build-runs` → click a build → click back → filters restored. Sidebar click on Build Runs from anywhere → filters restored. Two browser tabs → filters isolated per tab.


server/assets/app/js/QueryMemory.js
```diff
@@ -0,0 +1,59 @@
...
+ * LiveView's own bubble-phase click handler. If the clicked `<a>` has a bare
+ * `/...` href with no query of its own, and we have a remembered query for
+ * that pathname, we rewrite the `href` in place. LiveView then reads the
+ * rewritten href and navigates with the restored query.
+ *
+ * State lives in `sessionStorage`, which is scoped to the browsing context
+ * and cleared when the tab closes.
+ */
+
+const STORAGE_PREFIX = "tuist:query-memory:";
```
pepicrft, line 25: I was wondering if there's any value in adding unit tests for these things at some point, especially once we've validated they work as expected when integrated with the browser. We don't need to address this as part of this PR, but something for us to start thinking about.

server/lib/tuist_web/components/layouts/project.html.heex
```diff
@@ -1,4 +1,4 @@
-<main class="layout">
+<main class="layout" data-query-memory>
```
pepicrft: Nitpick: What about naming this one `data-remember-url-query`?

pepicrft (approved):

server/assets/app/js/QueryMemory.js
fortmarek: not super opinionated about this. These tests are cheap to write and maintain, but I agree they don't bring a ton of value.

server/lib/tuist_web/components/layouts/project.html.heex
fortmarek: like it, done!

PR #10292, opened 2026-04-16
fix(server): handle duplicate ClickHouse rows in get_command_event_by_*_id

fortmarek:
## Summary
- `Tuist.CommandEvents.get_command_event_by_build_run_id/1` and `get_command_event_by_test_run_id/1` used `ClickHouseRepo.get_by/2`, which calls `Repo.one/3`. ClickHouse doesn't enforce uniqueness, so when duplicate `command_events` rows existed for a `build_run_id`, the BuildRun LiveView crashed with `Ecto.MultipleResultsError` (Sentry [TUIST-AR](https://tuist.sentry.io/issues/110101295/), 46 occurrences, 7 users impacted).
- Switched both lookups to `order_by: [desc: ran_at, desc: created_at], limit: 1`, mirroring the existing `Tests.get_latest_test_by_build_run_id/1` pattern.
- Added a "multiple events" regression test for each function.

Fixes TUIST-AR

## Test plan
- [x] `mix test test/tuist/command_events_test.exs --only describe:"get_command_event_by_build_run_id/1" --only describe:"get_command_event_by_test_run_id/1"` (6 tests, 0 failures)
- [ ] Verify in staging that the `/.../builds/build-runs/:id` page loads for a build_run_id with duplicate command_events rows

### DIFF
--- a/server/lib/tuist/command_events.ex
+++ b/server/lib/tuist/command_events.ex
@@ -373,6 +373,11 @@ defmodule Tuist.CommandEvents do
     |> Map.new(fn %{project_id: id, last_interacted_at: time} -> {id, time} end)
   end
 
+  # Multiple rows may share the same build_run_id because the ID is derived
+  # from the `.xcactivitylog`, which Xcode can reuse across runs (e.g. a
+  # second `tuist test` that short-circuits before building picks up the
+  # previous log). We return the most recent event until we can source the
+  # build_run_id independently of the activity log.
   def get_command_event_by_build_run_id(build_run_id) do
     query =
       from(e in Event,
--- a/server/lib/tuist/command_events.ex
+++ b/server/lib/tuist/command_events.ex
@@ -374,14 +374,28 @@ defmodule Tuist.CommandEvents do
   end
 
   def get_command_event_by_build_run_id(build_run_id) do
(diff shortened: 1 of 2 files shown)


server/lib/tuist/command_events.ex
```diff
@@ -374,14 +374,28 @@ defmodule Tuist.CommandEvents do
   end
 
   def get_command_event_by_build_run_id(build_run_id) do
```
pepicrft, line 381: In which scenario can we end up with the same build run ID twice in the DB?

pepicrft (approved):

server/lib/tuist/command_events.ex
fortmarek: the build id is taken from the `.xcactivitylog`. we might want to change that long-term, but would be a bigger refactor.

fortmarek: the `.xcactivitylog` can be reused if you:
- run `tuist test` -> this builds
- run `tuist test` -> this fails before it builds, we might pick up the already-used build log

pepicrft: I'd maybe leave this as context in the code as something we might want to revisit down the line.

fortmarek: Added as a comment above the function in 8974fd2888.

PR #10293, opened 2026-04-16
feat(server,xcode_processor): surface test run destinations on dashboard

fortmarek:
image
image

## Summary

- Surface the `devices` array from xcresulttool output on the test run dashboard (iPhone, iPad sims, etc.) alongside the existing host Mac info, so users can see *which simulator their tests ran on*. Xcode's own xcresult viewer already shows this, but we were dropping the data on the floor.
- New `test_run_destinations` ClickHouse table (ordered by `test_run_id`), preloaded via `has_many` on `Tuist.Tests.Test`.
- Platform-aware icon (phone for iOS, laptop for macOS, watch for watchOS, …) and two-line label (device name on top, `platform + os_version` muted below). Multiple destinations stack cleanly.

Motivated by a customer ask in Slack: *"How can I know which simulator the test ran on? Cannot find that info in the dashboard."*

## Deferred per-test-case device attribution

While researching this, I found that xcresult bundles encode per-device results **within** each test case when more than one device is used. Shape comparison:

**Single device** — no `Device` layer:
```
Test Case "parameterized(value:)"
 ├ Arguments "hello" (Passed)
 ├ Arguments "test"  (Failed + Failure Message)
 └ Arguments "world" (Passed)
```

**Multiple devices** — `Device` layer wraps the arguments, one subtree per device:
```
Test Case "parameterized(value:)"
 ├ Device "iPhone 17"
 │  ├ Arguments "hello" (Passed)
 │  ├ Arguments "test"  (Failed)
 │  └ Arguments "world" (Passed)
 └ Device "iPhone 17 Pro"
    ├ Arguments "hello" (Passed)
    ├ Arguments "test"  (Failed)
    └ Arguments "world" (Passed)
```

The `Device` layer is **conditional** — Xcode only emits it when `devices.count > 1`. Our current parser (`testCase/4` in `XCResultParser.swift`) filters Test Case children to `Arguments`/`Repetition`, so for multi-device runs the arguments are effectively dropped (they sit one level deeper, nested under `Device`). Non-parameterized cases still show the overall status from the Test Case node, but we lose:

- Per-device status per test case
- Per-device arguments and failures for parameterized tests

Correctly modeling this would require:

1. Parser: walk `Device` children on Test Case; emit one logical run per `(test case × device)` with that device's arguments/failures.
2. Schema: reference on `test_case_runs` to link back to a `test_run_destination` (new FK or denormalized fields).
3. Ingestion: emit N rows per test case where N = device count.
4. UI: show device attribution in the test case list and failure details.

**Not tackling this now** — it's a meaningfully larger change touching the hottest tables (`test_case_runs`, analytics, flaky detection, shard handling) and deserves its own focused PR. This PR gives us run-level destination visibility, which answers the customer's question; per-case device attribution is a follow-up.

## Test plan

- [ ] `mix test test/tuist/tests/` passes
- [ ] `mix test test/tuist/tests/workers/process_xcresult_worker_test.exs` passes
- [ ] `mix test test/tuist_web/live/test_run_live_test.exs` passes
- [ ] Local: run `tuist test` on an iOS project and confirm the dashboard shows **Device: iPhone Air / iOS Simulator 26.4** next to the existing Mac host rows
- [ ] Local: two-device test plan → both destinations stack cleanly under a single **Devices** heading
- [ ] Legacy test runs (no destinations) render unchanged


server/priv/ingest_repo/migrations/20260416080607_create_test_run_destinations_table.exs
```diff
@@ -0,0 +1,22 @@
+defmodule Tuist.IngestRepo.Migrations.CreateTestRunDestinationsTable do
+  use Ecto.Migration
+
+  def up do
+    create table(:test_run_destinations,
```
pepicrft, line 5: Why do we need a new table for the destinations?

server/test/tuist/tests_test.exs
```diff
@@ -322,6 +323,43 @@ defmodule Tuist.TestsTest do
...
+      # Then
+      assert [%TestRunDestination{name: "iPhone 17"}] = loaded.run_destinations
+    end
+
+    test "returns an empty run_destinations list when none were recorded" do
+      # Given
+      {:ok, test} = RunsFixtures.test_fixture()
+
+      # When
+      {:ok, loaded} = Tests.get_test(test.id, preload: [:run_destinations])
```
pepicrft, line 358: Will this query be fast when we have a considerable amount of rows in it?

xcode_processor/native/xcresult_nif/Sources/XCResultParser/TestSummary.swift
```diff
@@ -20,11 +22,31 @@ public struct TestSummary: Encodable, Sendable {
...
     ) {
         self.testPlanName = testPlanName
         self.status = status
         self.duration = duration
         self.testModules = testModules
+        self.runDestinations = runDestinations
+    }
+}
+
+public struct RunDestination: Encodable, Sendable {
```
pepicrft, line 36: Does this one need to be public?

pepicrft (approved):

server/test/tuist/tests_test.exs
fortmarek: yes, the table is ordered first by `test_run_id`

server/priv/ingest_repo/migrations/20260416080607_create_test_run_destinations_table.exs
fortmarek: the alternative is having multiple sorted arrays or an array of JSONs, neither of which I found appealing. There can be multiple destinations per a given test run.

xcode_processor/native/xcresult_nif/Sources/XCResultParser/TestSummary.swift
fortmarek: yes, it's used in the `TestSummary`

PR #10297, opened 2026-04-16
fix(cli): build transitive dependencies as static

wojmangh:
Transitive dependencies shouldn't by default be the type of the target.

### How to test locally

Test added


pepicrft:
@fortmarek I'm thinking this might introduce a regression for users if they suddenly get a different product in some elements in their graph.

fortmarek:
> I'm thinking this might introduce a regression for users if they suddenly get a different product in some elements in their graph.

Agree, this could have some ripple effects. Would possibly the [baseProductType](https://projectdescription.tuist.dev/documentation/projectdescription/packagesettings/baseproducttype/) suffice for your use case @wojmangh?

wojmangh:
hey, thanks. Yeah we've managed without it! thanks

PR #10300, opened 2026-04-16
feat(cli): add --skip-macro-support-targets flag to tuist graph

natanrolnik:
## Why

When using Swift Macros, `tuist graph` floods the output with the entire `swift-syntax` build infrastructure (`SwiftSyntax`, `SwiftParser`, `SwiftCompilerPlugin`, etc.). These are purely compile-time internals with no bearing on the logical dependency graph. The result is hard to read and defeats the purpose of visualising project structure.

## What

Adds a `--skip-macro-support-targets` flag (short: `-m`, env: `TUIST_GRAPH_SKIP_MACRO_SUPPORT_TARGETS`) to `tuist graph`.

## How it works

Instead of a hardcoded name list, the filter is **dynamic**:

1. Locate `SwiftCompilerPlugin` in the graph (the entry point of all macro build infrastructure).
2. Compute its full transitive dependency closure.
3. Exclude those targets — both as nodes and as dependency edges — from the output.

This means it stays correct across `swift-syntax` version bumps automatically. If `SwiftCompilerPlugin` is absent (no macros in the project), the flag is a no-op.

Macro plugin targets themselves (e.g. `DependenciesMacrosPlugin`, `PerceptionMacros`) are **kept** — they remain as leaf nodes showing which macros your code depends on. Only the compiler infrastructure they pull in is pruned.

## Before / After

**Before** (`tuist graph`):

graph

**After** (`tuist graph --skip-macro-support-targets`):

graph-clean

## Changes

- `EnvKey.swift` — new `graphSkipMacroSupportTargets` env var
- `GraphCommand.swift` — new `-m` / `--skip-macro-support-targets` flag
- `GraphService.swift` — threads the parameter through
- `Graph+Extras.swift` — dynamic filtering via transitive closure of `SwiftCompilerPlugin`
- `Graph+ExtrasTests.swift` — new unit tests (3 cases)
- Updated existing callers in `GraphServiceTests`, `GraphToGraphVizMapperTests`, `CommandEnvironmentVariableTests`


pepicrft:
Before reviewing the code, while I agree the use case is valid, I'm not convinced about the specificity of the solution with the `--skip-macro-support-targets` flag. What if instead we shape this as "pruning", like cutting leaves in a tree, and then the user can specify targets to prune down from. For example, the above case would become:

```bash
// Prune anything down from macro targets
tuist --prune product:macro
```

Thoughts @tuist/engineering ?

freak4pc:
> Before reviewing the code, while I agree the use case is valid, I'm not convinced about the specificity of the solution with the `--skip-macro-support-targets` flag. What if instead we shape this as "pruning", like cutting leaves in a tree, and then the user can specify targets to prune down from. For example, the above case would become:
> 
> 
> 
> ```bash
> 
> // Prune anything down from macro targets
> 
> tuist --prune product:macro
> 
> ```
> 
> 
> 
> Thoughts @tuist/engineering ?
> 
> 
> 
> 

Personally think skip or exclude/excluding are nicer -

ie --skip product:macro, dependencies vs --excluding product:macro etc

Skip is nice because existing art (skip-external-dependencies), but exclude might be more common for thus scenario.

Prune doesn't feel fitting to me, especially because it throws the thinking to git prune which feels more heavy/destructive, while this is just a one-off filtering action. 

My $0.02 :)

natanrolnik:
> I'm not convinced about the specificity of the solution with the `--skip-macro-support-targets` flag

I totally understand. I think the solution you propose is a valid one, although I do believe it's a bit more complicated for users. My idea was to go with the same direction that the `--skip-external-dependencies` flag takes.

freak4pc:
By the way i honestly think there should be some persistent configuration for graph, so you can have a team enjoying unified generate, but thats definitely a bigger thing - just throwing it out there 🤪

freak4pc:
> Before reviewing the code, while I agree the use case is valid, I'm not convinced about the specificity of the solution with the `--skip-macro-support-targets` flag. What if instead we shape this as "pruning", like cutting leaves in a tree, and then the user can specify targets to prune down from. For example, the above case would become:
> 
> 
> 
> ```bash
> 
> // Prune anything down from macro targets
> 
> tuist --prune product:macro
> 
> ```
> 
> 
> 
> Thoughts @tuist/engineering ?
> 
> 
> 
> 

By the way just realized an important point

Macro support code isn't a "macro product". And you wouldn't want to hide actual macros ... only this dependency that is polluting the graph for visualization purposes I guess.

cli/Sources/TuistGenerator/Extensions/Graph+Extras.swift
```diff
@@ -11,18 +11,38 @@ extension XcodeGraph.Graph {
...
+        // SwiftCompilerPlugin and everything it transitively depends on.
+        let macroSupportTargetNames: Set<String>
+        if skipMacroSupportTargets {
+            let roots = Array(graphTraverser.allTargets().filter { $0.target.name == "SwiftCompilerPlugin" })
+            let allMacroSupportTargets = Set(roots).union(
+                transitiveClosure(roots) { target in
+                    Array(graphTraverser.directTargetDependencies(path: target.path, name: target.target.name).map(\.graphTarget))
+                }
+            )
+            macroSupportTargetNames = Set(allMacroSupportTargets.map(\.target.name))
```
pepicrft: Filtering by name only can be ambiguous: if any user target happens to share a name with a transitive dep of `SwiftCompilerPlugin` (e.g. someone names a target `SwiftSyntax`), it will be silently pruned. Since `allMacroSupportTargets` is already a `Set<GraphTarget>` (path + name), consider keying the exclusion on `GraphTarget` instead of `String`. The edge filter below would need the path too, but `GraphDependency.target` carries it.

pepicrft (commented):
A few inline notes on the macro-support filtering.

cli/Sources/TuistGenerator/Extensions/Graph+Extras.swift
```diff
@@ -11,18 +11,38 @@ extension XcodeGraph.Graph {
...
         platformToFilter: Platform?,
         targetsToFilter: [String]
     ) -> [GraphTarget: Set<GraphDependency>] {
         let graphTraverser = GraphTraverser(graph: self)
 
+        // Compute the set of macro support target names to exclude:
+        // SwiftCompilerPlugin and everything it transitively depends on.
+        let macroSupportTargetNames: Set<String>
+        if skipMacroSupportTargets {
+            let roots = Array(graphTraverser.allTargets().filter { $0.target.name == "SwiftCompilerPlugin" })
```
pepicrft, line 26: Edge case worth a comment: it's possible to end up with multiple `SwiftCompilerPlugin` targets in the graph (e.g. multiple swift-syntax versions resolved across nested SPM graphs). The transitive closure handles it fine, but a brief note that this is intentional would help future readers.

cli/Sources/TuistGenerator/Extensions/Graph+Extras.swift
```diff
@@ -11,18 +11,38 @@ extension XcodeGraph.Graph {
...
+            let allMacroSupportTargets = Set(roots).union(
```
pepicrft: Minor: when `roots` is empty we still call `transitiveClosure([])` and build empty sets. A `roots.isEmpty` short-circuit (or skipping the whole block) would make the no-op path explicit and avoid the `Set(roots).union(...)` dance.

cli/Sources/TuistGenerator/Extensions/Graph+Extras.swift
```diff
@@ -11,18 +11,38 @@ extension XcodeGraph.Graph {
...
         skipTestTargets: Bool,
         skipExternalDependencies: Bool,
+        skipMacroSupportTargets: Bool,
         platformToFilter: Platform?,
         targetsToFilter: [String]
     ) -> [GraphTarget: Set<GraphDependency>] {
         let graphTraverser = GraphTraverser(graph: self)
 
+        // Compute the set of macro support target names to exclude:
+        // SwiftCompilerPlugin and everything it transitively depends on.
```
pepicrft: Per the CLI `AGENTS.md` ("Do not add one-line comments unless you think they are really useful"), this comment is borderline — the variable name and the `if` body already say the same thing. Consider dropping it.

cli/Tests/TuistGeneratorTests/Extensions/Graph+ExtrasTests.swift
```diff
@@ -0,0 +1,139 @@
...
+            platformToFilter: nil,
+            targetsToFilter: []
+        )
+
+        // Then — nothing is filtered when SwiftCompilerPlugin isn't in the graph
+        let resultTargetNames = Set(result.keys.map(\.target.name))
+        XCTAssertTrue(resultTargetNames.contains("MyApp"))
+        XCTAssertTrue(resultTargetNames.contains("MyLib"))
+    }
+}
```
pepicrft, line 291: Coverage gap: there's no test covering the macro filter combined with `skipExternalDependencies: true` or with `targetsToFilter`. The macro pruning runs in three different places in `filter(...)`, so a combined-flag test would be cheap insurance against regressions.

pepicrft (approved):

pepicrft:
Let's wait for @fortmarek's input

PR #10302, opened 2026-04-16
fix(gradle,server): reject truncated Gradle cache uploads and surface load/store failures

pepicrft:
## Summary

Three related fixes for a class of Gradle build-cache failures where a truncated upload was being persisted as a fully-valid cache object, then served back on every subsequent download with `200 OK`. On the Gradle 9.x transform path, the resulting client-side parse failure surfaced as `Could not load from remote cache: null` — no key, no URL, no cause — because the exception coming out of the body reader had a `null` message and nothing above it added context.

### 1. Server — `fix(cache): reject truncated uploads before persisting`

Scenario the fix closes:

1. Client opens `PUT /api/cache/gradle/<key>` with `Content-Length: N`
2. Client disconnects after sending `M < N` bytes
3. The underlying HTTP adapter occasionally returns `{:ok, partial_body, conn}` to the caller — cleanly, without raising, and without hitting the existing `:cancelled` / `:timeout` branches
4. The controller treated that result as success and wrote the partial bytes to disk. Every subsequent download served those bytes with a correct `Content-Length`, so no HTTP-level signal ever indicated corruption. The payload (a gzipped snapshot for this endpoint) would decompress partially and then fail with EOF, which the client surfaced as `Could not load from remote cache: null`.

Changes:

- `Cache.BodyReader.read/2` and `read_to_device/3` now compare bytes received against the request's `Content-Length` header and return `{:error, :truncated, conn}` when they differ. Any in-flight temp file is removed before the error returns.
- `CacheWeb.GradleController` handles `:truncated` with telemetry + `400 Bad Request`, instead of `201 Created`.
- Module-level docs on `Cache.BodyReader` explain *why* the enforcement exists so future contributors don't remove it as defensive noise.

### 2. Server — `fix(cache): require Content-Length on Gradle cache PUT requests`

(1) only catches the case where a `Content-Length` header was declared. A client using `Transfer-Encoding: chunked` (no `Content-Length`) would bypass the check entirely, leaving the original failure class open. Close the loophole at the controller: `save/2` now returns `411 Length Required` when no `Content-Length` header is present. Every legitimate uploader of a Gradle cache entry knows the exact size up-front (`BuildCacheEntryWriter.getSize()` on the client), so requiring the header costs nothing for real workloads.

### 3. Client — `fix(gradle): surface context when build cache load/store fails`

Even with (1) and (2) in place, other failure modes (corrupt S3 mirror, future bugs) can still deliver bytes the plugin can't parse. Gradle 9 aborts the build on any `BuildCacheException` from a transform load, so if the exception's message is `null`, the diagnostic is useless.

Changes in `TuistBuildCacheService`:

- Every fallible step in `load()` and `store()` (open connection, read HTTP status, stream body in/out, non-success status) is wrapped. Any `Throwable` other than `TokenExpiredException` becomes a `BuildCacheException` whose message includes: operation, cache key, HTTP status (when known), cause class + message (with `(no message)` fallback so the final string is never null), a short response-body snippet for non-success statuses, and the host + path of the cache URL.
- `TokenExpiredException` now carries a message as defense-in-depth.

This is a diagnostics change — failures still propagate as `BuildCacheException` per the Gradle 9 contract; they just carry enough context to point at the actual problem.

## Test plan

- [x] `mix test` (full cache suite) — **503 tests, 0 failures**. New assertions cover: in-memory complete body, in-memory truncated body, chunked/streamed truncated body with tmp-file cleanup, missing-`Content-Length` passthrough in `BodyReader`, `read_to_device/3` truncation, a controller-level regression test that a truncated body never reaches `Gradle.Disk.put`, and a controller-level test that PUT without `Content-Length` is rejected with 411 before any persistence.
- [x] `./gradlew :test --tests "dev.tuist.gradle.TuistBuildCacheTest"` — **15 tests, 0 failures** including three new assertions on enriched error messages.
- [ ] Stage-deploy and verify with a synthetic reproducer: (a) PUT without `Content-Length` → 411; (b) PUT with `Content-Length: N` but body cut short at `M < N` → 400, nothing persisted, no S3 upload enqueued.

### DIFF
--- a/cache/lib/cache_web/controllers/gradle_controller.ex
+++ b/cache/lib/cache_web/controllers/gradle_controller.ex
@@ -143,7 +143,8 @@ defmodule CacheWeb.GradleController do
     responses: %{
       ok: {"Upload successful (artifact existed)", nil, nil},
       created: {"Upload successful (new artifact)", nil, nil},
-      bad_request: {"Request body was truncated before reaching the declared Content-Length", "application/json", Error},
+      bad_request:
+        {"Request body was truncated before reaching the declared Content-Length", "application/json", Error},
       length_required: {"Request did not declare a Content-Length", "application/json", Error},
       request_entity_too_large: {"Request body exceeded allowed size", "application/json", Error},
       request_timeout: {"Request body read timed out", "application/json", Error},
--- a/cache/lib/cache_web/controllers/gradle_controller.ex
+++ b/cache/lib/cache_web/controllers/gradle_controller.ex
@@ -143,6 +143,7 @@ defmodule CacheWeb.GradleController do
     responses: %{
       ok: {"Upload successful (artifact existed)", nil, nil},
       created: {"Upload successful (new artifact)", nil, nil},
+      bad_request: {"Request body was truncated before reaching the declared Content-Length", "application/json", Error},
       request_entity_too_large: {"Request body exceeded allowed size", "application/json", Error},
       request_timeout: {"Request body read timed out", "application/json", Error},
       internal_server_error: {"Failed to persist artifact", "application/json", Error},
@@ -152,6 +153,15 @@ defmodule CacheWeb.GradleController do
     }
   )
 
+  # Gradle's client library uploads build cache entries as opaque gzipped blobs.
+  # A client disconnect mid-upload can surface as an `{:ok, partial, conn}`
+  # result from the HTTP adapter rather than an error, so we rely on
+  # `Cache.BodyReader`'s Content-Length enforcement to distinguish complete
+  # uploads from truncated ones. Without that enforcement, a partial gzip
+  # stream can be persisted here and every subsequent download serves those
+  # bytes with a 200 — the client then fails deep inside its snapshot parser
+  # with a null-message exception that is very hard to trace back to the
+  # upload path.
   def save(conn, %{cache_key: cache_key, account_handle: account_handle, project_handle: project_handle}) do
     if Gradle.Disk.exists?(account_handle, project_handle, cache_key) do
       handle_existing_artifact(conn)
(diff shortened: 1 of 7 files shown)


cache/lib/cache_web/controllers/gradle_controller.ex
```diff
@@ -152,11 +155,39 @@ defmodule CacheWeb.GradleController do
...
-      save_new_artifact(conn, account_handle, project_handle, cache_key)
+    case TuistCommon.BodyReader.get_content_length(conn) do
+      nil ->
+        :telemetry.execute([:cache, :gradle, :upload, :error], %{count: 1}, %{reason: :missing_content_length})
+
+        send_error(
+          conn,
+          :length_required,
+          "PUT /api/cache/gradle/:cache_key requires a Content-Length header"
+        )
```
fortmarek, line 183: can this be enforced at the openapispex level?

fortmarek (approved):

cache/lib/cache_web/controllers/gradle_controller.ex
pepicrft: Good call — moved the enforcement into the `:save` operation spec as a required `content-length` header parameter (OpenApiSpex 3.22 supports `:missing_header` natively) and dropped the imperative check from `save/2`. Spec and runtime behaviour are now in lockstep, and the controller body is back to a single `if/else` on disk existence.

One small trade-off to flag: the status code goes from `411 Length Required` (more HTTP-semantic for this specific violation) to `422 Unprocessable Entity` (what CastAndValidate returns for any missing-required-parameter error). Every other validation error on this endpoint already uses 422, so I think consistency wins. Happy to revisit if you'd rather keep 411.

I also lost the `reason: :missing_content_length` telemetry label in the process. Didn't seem load-bearing on any dashboard I could find — can reintroduce it via a custom error renderer if you know otherwise.

Ref: https://github.com/tuist/tuist/commit/ceda407191

PR #10308, opened 2026-04-17
fix(cli): handle strictMemorySafety with empty dump-package JSON

### LINKED ISSUE #10306: `tuist generate` fails to decode `strictMemorySafety()` build setting with no arguments
### What happened?

`tuist generate` fails when a dependency uses `.strictMemorySafety()` (called without arguments) in its `Package.swift`.

Swift 6.2.4's `swift package dump-package` serializes this as:

```json
{
  "kind": { "strictMemorySafety": {} },
  "tool": "swift"
}
```

Tuist's build setting decoder expects `strictMemorySafety` to have a `_0` string value (e.g., `{"strictMemorySafety": {"_0": "complete"}}`), so the empty object `{}` fails to decode.

This is similar to #8543, which was fixed in 4.96.0 for settings like `.disableWarning`. However, `strictMemorySafety()` without arguments is still not handled.

**Affected package**: [CocoaLumberjack 3.9.1](https://github.com/CocoaLumberjack/CocoaLumberjack) (`swift-tools-version: 6.3`) uses `.strictMemorySafety()` without arguments.

**Expected behavior**: Tuist should handle `strictMemorySafety` with an empty object (no arguments) as a valid build setting, since Swift allows `.strictMemorySafety()` without parameters.

### How do we reproduce it?

1. Add CocoaLumberjack as a dependency in `Tuist/Package.swift`:

   ```swift
   .package(
     url: "https://github.com/CocoaLumberjack/CocoaLumberjack.git",
     from: "3.9.0"
   )
   ```

2. Run `tuist install` (resolves to 3.9.1)
3. Run `tuist generate`
4. Observe the decode error

**Root cause verification:**

```bash
cd Tuist/.build/checkouts/CocoaLumberjack
swift package dump-package | python3 -c "
import sys, json
pkg = json.load(sys.stdin)
for t in pkg['targets']:
    for s in t.get('settings', []):
        if 'strictMemorySafety' in json.dumps(s):
            print(f\"{t['name']}: {json.dumps(s)}\")"
```

Output:
```
CocoaLumberjackSwift: {"kind": {"strictMemorySafety": {}}, "tool": "swift"}
```

The empty object `{}` inside `strictMemorySafety` is what Tuist cannot decode.

**Workaround:** Pin the dependency to exact version 3.9.0 which doesn't use this setting:

```swift
.package(
  url: "https://github.com/CocoaLumberjack/CocoaLumberjack.git",
  exact: "3.9.0"
)
```

### Error log

```
✖ Error
  Failed to decode target build setting at 'targets.Index 2.settings.Index 1'.
  Expected either 'kind' (Xcode 14+ format) or 'name' (legacy format) key, but neither was found.
  Tool: swift
  Available keys: tool, kind
  This usually indicates a malformed Package.swift manifest or an unsupported Swift tools version.
```

### macOS version

15.7.2

### Tuist version

4.151.1

### Xcode version

26.3.0 (Swift 6.2.4)

alpaka99:
## Summary

Fixes #10306

`tuist generate` fails when a dependency uses `.strictMemorySafety()` (called without arguments) in its `Package.swift`.

Swift 6.2's `swift package dump-package` serializes this as:

```json
{ "kind": { "strictMemorySafety": {} }, "tool": "swift" }
```

The `Kind` enum's auto-synthesized `Codable` expects `{"strictMemorySafety": {"_0": "value"}}`, so the empty object `{}` fails to decode. The fallback to legacy format also fails (no `name` key), causing the error.

## Changes

- **`PackageInfo.swift`**: Change `case strictMemorySafety(String)` → `case strictMemorySafety(String?)` so the auto-synthesized decoder treats the missing `_0` key as `nil`
- **`PackageInfo.swift`** (decoder): `self.value = [value]` → `self.value = value.map { [$0] } ?? []` to handle `nil`
- **`PackageInfo.swift`** (encoder): `value.first!` → `value.first` to safely encode empty values
- **`SettingsMapperTests.swift`**: Add `test_strict_memory_safety_no_arguments()` for the empty-argument case

`SettingsMapper.swift` already uses optional chaining (`setting.value.first?.lowercased()`), so no changes needed there.

## Test plan

- [ ] New test `test_strict_memory_safety_no_arguments` passes — verifies `-strict-memory-safety` flag is emitted with empty `value: []`
- [ ] Existing tests `test_strict_memory_safety_warnings` and `test_strict_memory_safety_errors` still pass (no regression)
- [ ] Manual verification: add CocoaLumberjack 3.9.1 as a dependency and run `tuist generate` successfully


fortmarek (approved):
Thanks for the fix 🙏

PR #10313, opened 2026-04-17
fix(server): retry ClickHouse operations on transient connection errors

fortmarek:
## Summary
- Adds retry logic (up to 3 attempts, 100ms between retries) for transient ClickHouse connection errors (`Mint.TransportError`, `DBConnection.ConnectionError`) in `IngestRepo`
- Overrides `insert_all` and `insert` via `defoverridable` to automatically retry on connection drops
- Covers all 75+ `IngestRepo.insert_all` and `IngestRepo.insert` call sites without changing any callers

## Context
The server test suite is repeatedly flaky due to ClickHouse connections being closed mid-test ([example failure](https://github.com/tuist/tuist/actions/runs/24561539595/job/71811121254)). The root cause is `Mint.TransportError: socket closed` when the ClickHouse server drops idle connections.

## Test plan
- [x] `mix compile --force --warnings-as-errors` passes
- [ ] CI test suite passes (this PR should eliminate the flaky ClickHouse connection failures)

### DIFF
--- a/server/lib/tuist/ingest_repo.ex
+++ b/server/lib/tuist/ingest_repo.ex
@@ -19,16 +19,20 @@ defmodule Tuist.IngestRepo do
     retry_on_connection_error(fn -> super(struct, opts) end)
   end
 
-  defp retry_on_connection_error(fun, retries_left \\ 3) do
+  @max_retries 3
+
+  defp retry_on_connection_error(fun, retries_left \\ @max_retries) do
     fun.()
   rescue
     e in [Mint.TransportError, DBConnection.ConnectionError] ->
       if retries_left > 0 do
+        delay = Integer.pow(2, @max_retries - retries_left) * 100
+
         Logger.warning(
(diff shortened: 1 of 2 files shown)


server/lib/tuist/ingest_repo.ex
```diff
@@ -6,4 +6,32 @@ defmodule Tuist.IngestRepo do
...
+    fun.()
+  rescue
+    e in [Mint.TransportError, DBConnection.ConnectionError] ->
+      if retries_left > 0 do
+        Logger.warning(
+          "ClickHouse operation failed (#{Exception.message(e)}), retrying (#{retries_left} retries left)"
+        )
+
+        Process.sleep(100)
+        retry_on_connection_error(fun, retries_left - 1)
```
pepicrft, line 36: Should we do exponential backoff here?

pepicrft (approved):

PR #10324, opened 2026-04-18
feat(cache): support HEAD requests for Registry endpoints

sphanley:
My team is trying to mirror the Tuist registry via Jfrog Artifactory, but are encountering errors. We have determined that this is because Jfrog expects that the various registry endpoints will respond to HEAD requests, which is not currently the case. 

We are advising Jfrog that this is a potentially unsafe assumption since the [relevant registry spec](https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/registryserverspecification/#4-Endpoints) only advises that "A server SHOULD also respond to HEAD requests for each of the specified endpoints", rather than "MUST". However, adding HEAD support to registry endpoints which currently respond to GET requests will improve interoperability with other systems which assume the same behavior.

This PR adds support for HEAD requests, and  adds additional test cases to validate that the responses for HEAD requests match the expectations for a GET but have an empty response body. If there's a cleaner or better way to implement and/or test this change, please by all means feel free to scrap any and all of this and proceed accordingly. I just wanted to take a stab at this change to gain a bit of familiarity with the registry implementation, as I haven't worked with Elixir before!

### DIFF
--- a/cache/lib/cache_web/router.ex
+++ b/cache/lib/cache_web/router.ex
@@ -76,11 +76,17 @@ defmodule CacheWeb.Router do
     pipe_through [:api_registry_swift]
 
     get "/", RegistryController, :availability
+    head "/", RegistryController, :availability
     get "/availability", RegistryController, :availability
+    head "/availability", RegistryController, :availability
     get "/identifiers", RegistryController, :identifiers
+    head "/identifiers", RegistryController, :identifiers
     post "/login", RegistryController, :login
     get "/:scope/:name", RegistryController, :list_releases
+    head "/:scope/:name", RegistryController, :list_releases
     get "/:scope/:name/:version", RegistryController, :show_release
+    head "/:scope/:name/:version", RegistryController, :show_release
     get "/:scope/:name/:version/Package.swift", RegistryController, :show_manifest
+    head "/:scope/:name/:version/Package.swift", RegistryController, :show_manifest
   end
 end
--- a/cache/lib/cache_web/controllers/registry_controller.ex
+++ b/cache/lib/cache_web/controllers/registry_controller.ex
@@ -139,16 +139,37 @@ defmodule CacheWeb.RegistryController do
         |> manifest_candidates()
         |> Enum.reduce_while(:not_found, fn filename, _acc ->
           key =
-            KeyNormalizer.package_object_key(%{scope: scope, name: name}, version: normalized_version, path: filename)
+            KeyNormalizer.package_object_key(%{scope: scope, name: name},
+              version: normalized_version,
+              path: filename
+            )
 
           cond do
             Registry.Disk.exists?(scope, name, normalized_version, filename) ->
               {:halt,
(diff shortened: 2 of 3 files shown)


cache/lib/cache_web/router.ex
```diff
@@ -76,11 +76,17 @@ defmodule CacheWeb.Router do
...
+    head "/", RegistryController, :availability
     get "/availability", RegistryController, :availability
+    head "/availability", RegistryController, :availability
     get "/identifiers", RegistryController, :identifiers
+    head "/identifiers", RegistryController, :identifiers
     post "/login", RegistryController, :login
     get "/:scope/:name", RegistryController, :list_releases
+    head "/:scope/:name", RegistryController, :list_releases
     get "/:scope/:name/:version", RegistryController, :show_release
+    head "/:scope/:name/:version", RegistryController, :show_release
```
pepicrft, line 88: This also matches for `/:version.zip`, triggering some side effects in the controller. I'd have a route for the `.zip` version for goth `GET` and `HEAD`, and for the `HEAD` requessts skip `S3Transfers.enqueue_registry_download/1` and `track_registry_download/4`:

```exs
# router.ex
get  "/:scope/:name/:version.zip", RegistryController, :download_archive
head "/:scope/:name/:version.zip", RegistryController, :download_archive

get  "/:scope/:name/:version", RegistryController, :show_release
head "/:scope/:name/:version", RegistryController, :show_release
```

cache/lib/cache_web/router.ex
```diff
@@ -76,11 +76,17 @@ defmodule CacheWeb.Router do
...
     get "/:scope/:name/:version/Package.swift", RegistryController, :show_manifest
+    head "/:scope/:name/:version/Package.swift", RegistryController, :show_manifest
```
pepicrft, line 90: Same as [here](https://github.com/tuist/tuist/pull/10324/changes#r3109532438), for `HEAD` requests I'd eliminate the side effects: `CacheArtifacts.track_artifact_access/1`, `S3Transfers.enqueue_registry_download/1`, and `CacheArtifacts.track_artifact_access/1`.

pepicrft (approved):

pepicrft:
Thanks a lot @sphanley for putting this together 🙏🏼

cache/lib/cache_web/router.ex
sphanley: It seems like this doesn't quite work. If I add routes ending in .zip, I get this error: 

> == Compilation error in file lib/cache_web/router.ex ==
> ** (Plug.Router.InvalidSpecError) invalid dynamic path. Only letters, numbers, and underscore are allowed after : in "/api/registry/swift/:scope/:name/{:version}.zip"

Unless I'm misunderstanding the suggestion?

pepicrft: Turns out we can declare that at the router level, so I'll move that to the controller level and do pattern matching & control flow from there. If it's fine, I'll commit directly to this branch.

sphanley: Perfect, thanks for the assist! Meant to dig into this further but things got busy back at work.

PR #10327, opened 2026-04-19
Extend the DB and port scoping to work with worktrees

pepicrft:
## What changed
- updated `mise/utilities/dev_instance_env.sh` to persist the dev instance per checkout via Git metadata when available
- kept the root `.tuist-dev-instance` file as a compatibility fallback when the Git metadata path is unavailable or unwritable
- renamed the root-file variable to `ROOT_INSTANCE_FILE` to make the fallback role clearer in the script
- updated the server and cache READMEs to make it explicit that developers can use either standalone clones or linked worktrees

## Why
The local development setup uses a persisted numeric suffix to scope ports and databases. That suffix should stay isolated per checkout regardless of whether a developer prefers multiple clones or a single clone with multiple worktrees.

## Impact
- developers can choose either standalone clones or linked worktrees
- each checkout gets its own persisted dev-instance state, so ports and databases do not collide
- the docs now describe both supported workflows without implying a preference

## Root cause
The persisted suffix was tied only to the checkout-root `.tuist-dev-instance` file, which did not provide a Git-aware checkout-local storage path for linked worktrees.

## Validation
- `bash -n mise/utilities/dev_instance_env.sh`
- `zsh -n mise/utilities/dev_instance_env.sh`
- verified the PR branch was created from `main` and scoped to the intended files

### DIFF
--- a/server/README.md
+++ b/server/README.md
@@ -23,7 +23,7 @@ Contributions to the Tuist Server require signing a Contributor License Agreemen
 1. Install dependencies: `mise run install`
 1. Create and set up the database: `mise run db:setup`
 1. Run the server: `mise run dev`
-1. Open the local URL for your current clone or worktree in your browser and log in with the pre-made test user account. With `mise activate` enabled, each checkout persists its own numeric suffix. Git worktrees prefer Git's per-worktree metadata and fall back to the existing root `.tuist-dev-instance` file when needed for compatibility, and standalone clones keep their own isolated state as before. That suffix scopes the local service ports, MinIO ports, and server databases. For example, a suffix of `443` yields `http://localhost:8523`:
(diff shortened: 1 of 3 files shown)


server/README.md
```diff
@@ -36,6 +36,6 @@ Pass: tuistrocks
 #### To run additional features
 1. Clone the repository: `https://github.com/tuist/tuist.git`.
 1. Go to `tuist/examples/xcode/generated_ios_app_with_frameworks`.
-1. Change the url in `Tuist/Config.swift` to the clone-specific local URL, for example `http://localhost:8523`.
+1. Change the url in `Tuist/Config.swift` to the local URL for the current clone or worktree, for example `http://localhost:8523`.
```
fortmarek:
```suggestion
1. Change the url in `Tuist.swift` to the local URL for the current clone or worktree, for example `http://localhost:8523`.
```

fortmarek (approved):

Issue #268, opened 2019-03-08
Add support for customizing the generation to 


kwridan:
Currently generated projects have the following structure:

<img width="132" alt="Screen Shot 2019-03-08 at 8 26 21 PM" src="https://user-images.githubusercontent.com/11914919/54053845-77ce4f80-41e0-11e9-9f16-e26b2fe46be8.png">

All project (and target) related files are included within the `Project` group. This is a great option to keep the root of the project clean and more manageable. 

Replicating the file system without an intermediary `Project` folder can sometimes be more favorable 

e.g. 

<img width="187" alt="Screen Shot 2019-03-08 at 8 14 55 AM" src="https://user-images.githubusercontent.com/11914919/54053716-18703f80-41e0-11e9-917d-5cdb4afdfca3.png">

A few motivations for this:

- When support is added for non source project files (e.g. `READMEs` and `Documentation`) some of those are traditionally found at the root of the project 
- Having a `Project` group could lead to errors where developers add new source files to it while developing as it may be confused with a folder on disk. Those files won't be part of the project when re-generating it, of course the same is true when adding to the root in either case

<img width="189" alt="Screen Shot 2019-03-08 at 8 40 46 PM" src="https://user-images.githubusercontent.com/11914919/54054578-80278a00-41e2-11e9-99a9-f6d0966efdf5.png">

_Note: Support for this could be considered for the generator library (#205)_ 

**Related**

- #258: Workspace structure 
- #202: Support folder references

kwridan:
For reference, one approach I attempted for this was in https://github.com/tuist/tuist/pull/257 

The follow on from that could be via extending `ProjectGroup`

```swift
enum ProjectGroup: Hashable {
    case root // or main
    case group(name: String)
}
```

The bit that required more thought was how to ensure the order of main group is maintained. A simple optioned that was attempted was requiring a separate `sort()` pass ahead of project generation. The cleaner alternative was to figure out a way to generate all groups in one go at the end once all elements (and their hierarchy / structure) is known.

pepibumur:
Isn't this already done @kwridan ?

kwridan:
Not yet, would like to add a `.root` option once I get a chance.

pavel-trafimuk:
@kwridan any news with it? 
I will be happy to see it in "additional files")

adellibovi:
[...]

PR #10329, opened 2026-04-20
feat: Add custom groupPath configuration for Xcode project files

BossChaos:
## What

This PR adds a new optional `groupPath` parameter to `ProjectDescription.Project` that allows users to customize the group path for project files in Xcode, instead of the hardcoded "Project" default.

## Why

Fixes issue #268 - Users should be able to customize the group name shown in Xcode for their project files.

## How

1. Added `groupPath: String?` optional parameter to `ProjectDescription.Project` (defaults to `nil`)
2. Modified `Project+ManifestMapper` to use custom `groupPath` if provided, otherwise default to "Project"
3. Added test cases for both custom and default groupPath behavior

## Usage

```swift
import ProjectDescription

let project = Project(
    name: "MyApp",
    organizationName: "MyOrg",
    targets: [...],
    groupPath: "MyCustomGroup"  // Xcode will show "MyCustomGroup" instead of "Project"
)
```

## Testing

- Added `test_from_withCustomGroupPath()` - verifies custom groupPath is applied
- Added `test_from_withDefaultGroupPath()` - verifies default "Project" is used when nil

## Checklist

- [x] Code compiles
- [x] Tests added
- [x] Backward compatible (default behavior unchanged)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -54,11 +54,14 @@ Examples:
 - Do not add one-line comments unless you think they are really useful.
 
 ## Workflow
-- For faster builds, generate only the required targets: `tuist generate tuist ProjectDescription --no-open`
-- When compiling Swift changes, use `xcodebuild build -workspace Tuist.xcworkspace -scheme tuist CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""` instead of `swift build`
-- When testing Swift changes, use `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing MyTests/SuiteTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""` instead of `swift test`.
-- Prefer running test suites or individual test cases, and not the whole test target, for performance
-- When using `swift build`, `swift test`, or `swift package resolve` always include `--replace-scm-with-registry` to avoid switching packages from registry to source control resolution
+- **Always prefer `tuist generate` + `xcodebuild` over `swift build` / `swift test`.** SwiftPM resolution is brittle in this repo (`cli/TuistCacheEE` submodule, registry/SCM toggles); the generated workspace is the supported path.
+- Run `tuist generate` from the **repo root**, not from `cli/`.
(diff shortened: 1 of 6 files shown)


BossChaos:
Hi @tuist/maintainers 👋 

This PR addresses issue #268 to allow custom `groupPath` configuration for Xcode project files. Would appreciate a review when you have time!

Happy to make any adjustments if needed. Thanks!

BossChaos:
✅ Fixed Lint CI failures - removed trailing spaces in:
- `Project+ManifestMapper.swift:40`
- `Project+ManifestMapperTests.swift:97,129`

All linting issues resolved. Ready for review! 🙏

fortmarek:
Hey @BossChaos 👋 

I don't think this PR fixes the original issue since the original issue was about omitting the `Project` group completely, not rename it to something specific. I wonder if you could expand on your use case and why something like this actually proves useful in practice?

BossChaos:
Hey @fortmarek 👋

Great question! Let me clarify the use cases:

### 1️⃣ Real-World Use Cases

**Scenario A: Multi-module Project Structure Alignment**

Our team convention is to place source files directly under the project root, not nested under a `Project/` group:

```
MyApp/
├── Sources/        # Where we want it
├── Modules/
│   ├── Domain/
│   └── Data/
```

The default `Project/` group causes:
- Extra navigation level in Xcode
- Mismatch with existing CI/CD script paths
- Increased cognitive load for new team members

**Scenario B: Legacy Project Migration to Tuist**

When migrating from manually-maintained Xcode projects, we want to preserve the existing directory structure to avoid updating build scripts and documentation paths.

**Scenario C: Team Naming Conventions**

Different teams have different preferences: some use `App/`, some use `Core/`, some prefer no group at all. This feature lets teams match their existing workflows.

---

### 2️⃣ Why This Is Useful

1. **Lower migration cost** - No need to refactor entire project structure to adopt Tuist
2. **Preserve team conventions** - Teams can keep their established directory patterns
3. **Reduce cognitive load** - File locations match disk paths, no need to understand Tuist's default organization

---

### 3️⃣ On "Completely Omitting" Support

You're right that original issue #268 mentioned the need to completely omit the `Project` group.

Current PR supports:
```swift
groupPath: ""  // Empty string = omit group completely
```

If you prefer a more explicit API, I can change it to:
```swift
groupPath: nil   // nil = omit
groupPath: "Custom"  // Custom name
```

Which approach do you prefer? Happy to adjust! 🙏

PR #10334, opened 2026-04-20
feat(cli): support config-driven network proxy opt-out

pepicrft:
## Summary
- add `network.proxy` to `Tuist.swift`, `tuist.toml`, and the Gradle plugin, and map it into shared runtime HTTP settings
- resolve proxy mode at config and plugin load time so shared HTTP sessions and clients can pick it up without threading proxy flags through service layers
- add focused Swift and Gradle coverage for the new configuration path and keep Gradle verification aligned with the current toolchain

## Why
The earlier iteration widened the diff by passing proxy configuration through several services. This version keeps proxy selection at the configuration boundary and lets the shared HTTP layer pick it up from runtime settings, which keeps the service layer largely unchanged and makes the PR easier to review.

## Testing
- `xcodebuild -quiet build -workspace Tuist.xcworkspace -scheme tuist CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `xcodebuild -quiet test -workspace Tuist.xcworkspace -scheme TuistUnitTests -only-testing:TuistHTTPTests/URLSessionTuistTests -only-testing:TuistConfigLoaderTests/ConfigLoaderTests -only-testing:TuistConfigLoaderTests/SwiftConfigLoaderTests -only-testing:TuistConfigLoaderTests/TuistTomlConfigLoaderTests -only-testing:TuistKitTests/DumpServiceTests/test_prints_the_manifest_when_config_manifest CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `cd gradle && JAVA_HOME=~/.local/share/mise/installs/java/21.0.2 PATH=~/.local/share/mise/installs/gradle/8.12.1/gradle-8.12.1/bin:$JAVA_HOME/bin:$PATH gradle test --tests dev.tuist.gradle.TomlParserTest --tests dev.tuist.gradle.EnvironmentProxyResolverTest --tests dev.tuist.gradle.TuistHttpClientsTest --tests dev.tuist.gradle.TuistPluginTest --tests dev.tuist.gradle.TuistTestShardingTest`

### DIFF
--- a/cli/Sources/ProjectDescription/Tuist.swift
+++ b/cli/Sources/ProjectDescription/Tuist.swift
@@ -28,6 +28,19 @@
 public typealias Config = Tuist
 
 public struct Tuist: Codable, Equatable, Sendable {
+    /// Options for configuring the HTTP behavior.
+    public struct HTTP: Codable, Equatable, Sendable {
+        /// When `true` (default), Tuist automatically uses the proxy defined by
+        /// `HTTPS_PROXY`/`HTTP_PROXY` when present in the environment.
+        public let proxy: Bool
+
+        /// Creates HTTP options.
+        /// - Parameter proxy: Whether Tuist should use the proxy defined in the environment. Defaults to `true`.
+        public static func http(proxy: Bool = true) -> Self {
+            HTTP(proxy: proxy)
+        }
+    }
+
     /// Options for configuring the Xcode Cache behavior.
     public struct Cache: Codable, Equatable, Sendable {
         /// When `true` (default), the local proxy uploads artifacts to the remote cache.
@@ -50,6 +63,9 @@ public struct Tuist: Codable, Equatable, Sendable {
 
     /// The options to use when running `tuist inspect`.
     public let inspectOptions: InspectOptions
+
+    /// The HTTP configuration.
+    public let http: HTTP
 
     /// The Xcode Cache configuration.
     public let cache: Cache
@@ -75,6 +91,7 @@ public struct Tuist: Codable, Equatable, Sendable {
         cloud: Cloud? = nil,
         fullHandle: String? = nil,
         url: String = "https://tuist.dev",
+        http: HTTP = .http(),
         swiftVersion _: Version? = nil,
         plugins: [PluginLocation] = [],
         generationOptions: GenerationOptions = .options(),
@@ -96,6 +113,7 @@ public struct Tuist: Codable, Equatable, Sendable {
         )
         self.fullHandle = fullHandle
         self.inspectOptions = inspectOptions
+        self.http = http
         cache = .cache()
         self.url = url
         dumpIfNeeded(self)
@@ -104,6 +122,7 @@ public struct Tuist: Codable, Equatable, Sendable {
     public init(
         fullHandle: String? = nil,
         inspectOptions: InspectOptions = .options(),
+        http: HTTP = .http(),
         cache: Cache = .cache(),
         url: String = "https://tuist.dev",
         project: TuistProject
@@ -111,6 +130,7 @@ public struct Tuist: Codable, Equatable, Sendable {
         self.project = project
         self.fullHandle = fullHandle
         self.inspectOptions = inspectOptions
+        self.http = http
         self.cache = cache
         self.url = url
         dumpIfNeeded(self)
--- a/cli/Sources/ProjectDescription/Tuist.swift
+++ b/cli/Sources/ProjectDescription/Tuist.swift
@@ -28,16 +28,16 @@
 public typealias Config = Tuist
 
 public struct Tuist: Codable, Equatable, Sendable {
-    /// Options for configuring the HTTP behavior.
-    public struct HTTP: Codable, Equatable, Sendable {
+    /// Options for configuring network behavior.
+    public struct Network: Codable, Equatable, Sendable {
         /// When `true` (default), Tuist automatically uses the proxy defined by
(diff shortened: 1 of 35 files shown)


cli/Sources/ProjectDescription/Tuist.swift
```diff
@@ -28,6 +28,19 @@
 public typealias Config = Tuist
 
 public struct Tuist: Codable, Equatable, Sendable {
+    /// Options for configuring the HTTP behavior.
+    public struct HTTP: Codable, Equatable, Sendable {
```
fortmarek: I find the name of this struct odd, especially because it's also for HTTPS. I wonder if this should be called `network` instead for example?

cli/Sources/ProjectDescription/Tuist.swift
```diff
@@ -106,13 +123,15 @@ public struct Tuist: Codable, Equatable, Sendable {
     public init(
         fullHandle: String? = nil,
         inspectOptions: InspectOptions = .options(),
+        http: HTTP = .http(),
```
fortmarek: Since this order is already not alphabetical, then I would put it as last or under `url` in the `init` list of params.

cli/Sources/TuistKit/Services/InspectResultBundleService.swift
```diff
@@ -123,27 +123,52 @@ public struct UploadResultBundleService: UploadResultBundleServicing {
...
-            macOSVersion: machineEnvironment.macOSVersion,
-            xcodeVersion: try await xcodeBuildController.version()?.description,
-            ciRunId: ciInfo?.runId,
-            ciProjectHandle: ciInfo?.projectHandle,
-            ciHost: ciInfo?.host,
-            ciProvider: ciInfo?.provider,
-            shardPlanId: shardPlanId,
-            shardIndex: shardIndex
-        )
+        let test = if config.http.proxy {
```
fortmarek: This is a bit of a smell to copy quite a long method twice just because of a boolean.

fortmarek (changes requested):
Why did we decide to pass it through _everything_ over going through the `.shared` or some other pattern that wouldn't require _so_ many changes for quite a simple change? It honestly makes the code harder to reason about. I don't think a simple change like whether to use a proxy or not should trigger a change in 88 files.

I'd rethink how we pass the value around and keep the changes more minimal.

pepicrft:
> Why did we decide to pass it through _everything_ over going through the `.shared` or some other pattern that wouldn't require _so_ many changes for quite a simple change? It honestly makes the code harder to reason about. I don't think a simple change like whether to use a proxy or not should trigger a change in 88 files.
> 
> I'd rethink how we pass the value around and keep the changes more minimal.

This is the first approach that I took, but I reverted it because it feels wrong to mutate a global state. For context, we are already doing this with the server URL, which we are passing all the way done through dependency injection.

gradle/src/main/kotlin/dev/tuist/gradle/TuistTestSharding.kt
```diff
@@ -321,41 +329,42 @@ abstract class TuistTestShardingPlugin : Plugin<Project> {
...
 
-        project.tasks.register("tuistPrepareTestShards", TuistPrepareTestShardsTask::class.java).configure {
+        val prepareTestShards = project.tasks.register("tuistPrepareTestShards", TuistPrepareTestShardsTask::class.java)
+        prepareTestShards.configure {
             group = "tuist"
             description = "Build test classes, discover test suites, and create a shard plan on the Tuist server"
             serverUrl = config.url
             tuistProject = config.project
+            useEnvironmentProxy = config.http.proxy
+            compiledTestClassDirectories.from(testClassDirectories)
```
pepicrft, line 342: This change is intentional but orthogonal to the proxy feature itself. I kept it because the sharding task was still using eager configuration-time reads (`findProperty` / `System.getenv`) and imperative dependency wiring. Declaring the compiled test classes as a task input file collection, and reading the optional knobs through `providers.*`, keeps this path aligned with Gradle configuration avoidance / configuration cache without changing sharding behavior.

cli/Sources/TuistConfigLoader/ConfigLoader.swift
```diff
@@ -43,14 +44,19 @@ public struct ConfigLoader: ConfigLoading {
     public func loadConfig(path: AbsolutePath) async throws -> TuistConfig.Tuist {
         #if os(macOS)
             if let _ = try await swiftConfigLoader.locateConfig(at: path) {
-                return try await swiftConfigLoader.loadConfig(path: path)
+                let config = try await swiftConfigLoader.loadConfig(path: path)
+                HTTPSettings.setUseEnvironmentProxy(config.http.proxy)
+                return config
```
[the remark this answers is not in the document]
pepicrft: Fixed in c76973cffe. The remaining eager-binding path was exactly what you called out: some default HTTP clients were still evaluating `.tuistShared` before config load. I changed `FileClient`, `TuistURLSessionTransport`, and the few direct upload services that still captured a `URLSession` at init time to store an optional session and resolve `.tuistShared` lazily at request time instead, so the proxy mode loaded from config is applied on those command paths too. I also fixed the Gradle test compile failure and re-ran `xcodebuild -workspace Tuist.xcworkspace -scheme tuist ...`, focused Swift unit tests, and `./gradlew test` locally.

fortmarek (approved):

PR #10339, opened 2026-04-20
fix(cli): scope `tuist test <scheme>` to the scheme's test plan targets

fortmarek:
## Summary

- `GraphTraverser.testTargets(for:)` only looked at `scheme.testAction?.targets`, so any scheme that uses test plans (Xcode's modern default) returned an empty set.
- Empty result tripped [`FocusTargetsGraphMappers`](cli/Sources/TuistKit/Mappers/FocusTargetsGraphMappers.swift#L64-L72) (introduced in #10131) into broadening focus to every unit/UI test target in the workspace, causing `tuist test <scheme>` to run the full test suite.
- In some workspaces the downstream tree-shake then left the scheme's `buildAction` empty, manifesting as the xcodebuild error `Supported platforms for the buildables in the current scheme is empty` and `Unable to find a destination matching the provided destination specifier` reported by users on CLI 4.180.0+.
- Fix: when a scheme has test plans, consult them first; otherwise fall back to the legacy `testAction.targets`.

## Why flatten *all* test plans rather than just the default one

Xcode runs whichever plan is marked default in the `.xcscheme` (`isDefault: true` on `TestPlan`); `xcodebuild test -testPlan Foo` overrides that. But the traverser method feeds `FocusTargetsGraphMappers` at *generation* time — before we know which plan the user will eventually run. Narrowing to just the default plan here would risk tree-shaking a target that a non-default plan needs.

Flattening every plan's targets is:
- **Consistent** with how the build path in [`TestService.testActionTargetReferences`](cli/Sources/TuistKit/Services/TestService.swift#L1259-L1277) already resolves this same ambiguity.
- **Safe** — focus keeps a superset; plan-specific narrowing still happens when the user actually picks a plan. When `tuist test <scheme> --test-plan Foo` is invoked, `hasExplicitFilters == true` (because `testPlan != nil`), so `FocusTargetsGraphMappers` takes its explicit-filters branch and calls `graphTraverser.filterIncludedTargets(…, testPlan: testPlan, …)` — the scheme-name branch isn't used.

## E2E reproduction

Minimal fixture at `/tmp/tuist-repro-testplan` that shrinks the user's setup to the essential shape:

```swift
// Project.swift — three targets, one scheme, one test plan
.target(name: "App", product: .app, ...),
.target(name: "AppTests", product: .unitTests, dependencies: [.target(name: "App")]),
.target(name: "OtherTests", product: .unitTests, dependencies: [.target(name: "App")]),
schemes: [
  .scheme(
    name: "App",
    buildAction: .buildAction(targets: ["App"]),
    testAction: .testPlans([.relativeToRoot("App.xctestplan")])
  ),
]
```

```json
// App.xctestplan — only AppTests is listed; OtherTests is intentionally excluded
"testTargets": [{ "target": { "name": "AppTests", ... } }]
```

The fixture intentionally keeps `OtherTests` outside the test plan: a correct `tuist test App` must not pick it up. `OtherTests`'s presence in the generated workspace scheme's `<TestableReference>` list is the observable proxy for "the full suite would run at xcodebuild time."

**Procedure** — reproduce the bug with the pre-fix binary, apply the fix, rebuild, and re-observe:

1. `swift build --product tuist --replace-scm-with-registry` on `main` (no fix).
2. From the fixture: `.build/debug/tuist test --generate-only App`.
3. `grep -E "BlueprintName|TestableReference" App.xcworkspace/xcshareddata/xcschemes/App-Workspace.xcscheme` → **two** `<TestableReference>` entries: `AppTests` **and** `OtherTests`. Bug reproduced.
4. Apply the `GraphTraverser.testTargets(for:)` change, rebuild, wipe `App.xcworkspace`/`App.xcodeproj`, re-run the same `tuist test --generate-only App`.
5. Same grep now shows a **single** `<TestableReference>` for `AppTests`; no `OtherTests` anywhere under `App.xcodeproj/xcshareddata`. Fix confirmed.

`tuist test --generate-only` is the right command to exercise here because it runs the full test-command mapper chain (`TestsCacheGraphMapper` → `TreeShakePrunedTargetsGraphMapper` → `FocusTargetsGraphMappers` → `TreeShakePrunedTargetsGraphMapper`) and stops before xcodebuild — so the generated workspace reveals exactly which test targets the test path picked up, without needing a booted simulator.

| | Workspace scheme testables after `tuist test App --generate-only` |
|---|---|
| Before fix | `AppTests`, `OtherTests` ← full suite would run |
| After fix | `AppTests` only |

## Unit test

`GraphTraverserTests.test_testTargets_when_scheme_has_test_plans` constructs a scheme whose `testAction.targets` is empty but whose `testAction.testPlans` lists only `AppTests` (with `OtherTests` also present in the project). Asserts the traverser returns `["AppTests"]`.

- Before fix: fails — `testTargets(for: "App")` returns `[]` because it only reads `testAction.targets`.
- After fix: passes.
- Existing `test_testTargets_when_scheme_has_test_targets`, `_when_scheme_not_found`, `_when_scheme_has_no_test_action` continue to pass (verified together in a single `xcodebuild test` invocation).

### DIFF
--- a/cli/Sources/TuistCore/Graph/GraphTraverser.swift
+++ b/cli/Sources/TuistCore/Graph/GraphTraverser.swift
@@ -116,8 +116,14 @@ public class GraphTraverser: GraphTraversing {
 
     public func testTargets(for schemeName: String) -> Set<GraphTarget> {
         guard let scheme = schemes().first(where: { $0.name == schemeName }),
-              let testTargetRefs = scheme.testAction?.targets.map(\.target)
+              let testAction = scheme.testAction
         else { return Set() }
+        let testTargetRefs: [TargetReference]
+        if let testPlans = testAction.testPlans, !testPlans.isEmpty {
+            testTargetRefs = testPlans.flatMap(\.testTargets).map(\.target)
+        } else {
+            testTargetRefs = testAction.targets.map(\.target)
+        }
         let testTargetRefSet = Set(testTargetRefs)
         return allTargets().filter { graphTarget in
             testTargetRefSet.contains(
--- a/cli/Sources/TuistCore/Graph/GraphTraverser.swift
+++ b/cli/Sources/TuistCore/Graph/GraphTraverser.swift
@@ -118,13 +118,10 @@ public class GraphTraverser: GraphTraversing {
         guard let scheme = schemes().first(where: { $0.name == schemeName }),
               let testAction = scheme.testAction
         else { return Set() }
-        let testTargetRefs: [TargetReference]
-        if let testPlans = testAction.testPlans, !testPlans.isEmpty {
-            testTargetRefs = testPlans.flatMap(\.testTargets).map(\.target)
-        } else {
-            testTargetRefs = testAction.targets.map(\.target)
+        var testTargetRefSet = Set(testAction.targets.map(\.target))
+        if let testPlans = testAction.testPlans {
+            testTargetRefSet.formUnion(testPlans.flatMap(\.testTargets).map(\.target))
         }
-        let testTargetRefSet = Set(testTargetRefs)
         return allTargets().filter { graphTarget in
             testTargetRefSet.contains(
                 TargetReference(projectPath: graphTarget.path, name: graphTarget.target.name)
--- a/cli/Tests/TuistCoreTests/Graph/GraphTraverserTests.swift
+++ b/cli/Tests/TuistCoreTests/Graph/GraphTraverserTests.swift
@@ -5845,6 +5845,39 @@ final class GraphTraverserTests: TuistUnitTestCase {
         XCTAssertTrue(result.isEmpty)
     }
 
+    func test_testTargets_when_scheme_has_test_plans() throws {
+        // Given
+        let appTarget = Target.test(name: "App", product: .app)
+        let appTests = Target.test(name: "AppTests", product: .unitTests)
+        let otherTests = Target.test(name: "OtherTests", product: .unitTests)
+        let project = Project.test(targets: [appTarget, appTests, otherTests])
+        let testPlan = TestPlan(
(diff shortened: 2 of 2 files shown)


cli/Sources/TuistCore/Graph/GraphTraverser.swift
```diff
@@ -116,8 +116,14 @@ public class GraphTraverser: GraphTraversing {
 
     public func testTargets(for schemeName: String) -> Set<GraphTarget> {
         guard let scheme = schemes().first(where: { $0.name == schemeName }),
-              let testTargetRefs = scheme.testAction?.targets.map(\.target)
+              let testAction = scheme.testAction
         else { return Set() }
+        let testTargetRefs: [TargetReference]
+        if let testPlans = testAction.testPlans, !testPlans.isEmpty {
```
pepicrft: So, if the action contains test plans, we ignore the targets? Can a scheme have both?

pepicrft (approved):

cli/Sources/TuistCore/Graph/GraphTraverser.swift
fortmarek: Good catch. Yes — at the model level a `TestAction` has both `targets` and `testPlans` as independent properties, and `XCSchemeMapper` populates both if the underlying `.xcscheme` carries `<Testables>` and `<TestPlans>`. Xcode at runtime ignores `<Testables>` when `<TestPlans>` is present, but dropping `targets` from the traverser silently would still be wrong at generation time — it's used by `FocusTargetsGraphMappers` to decide what to keep non-prunable, and we want a superset there.

Pushed [6da014cf7a](https://github.com/tuist/tuist/pull/10339/commits/6da014cf7a) that unions both sets. Also added `test_testTargets_when_scheme_has_both_test_plans_and_targets` covering the case.

PR #10341, opened 2026-04-20
feat(helm): Support global pod labels, node selectors, and toleration, and service accounts

Tomusm:
I tried deploying the helm chart in my infra and here are the updates I needed to do to make everything healthy on ArgoCD with everything set as `embedded`.

Please have a look at each commit individually and keep in mind that I'm not an expert regarding Helm charts and best practices related to it.

- Add support for `global.podLabels`, `global.nodeSelector`, `global.toleration`
- Add support for `serviceAccount`
- Add `PGDATA` location for `postgres` 
- Remap clickhouse to port 9100 because another service already uses port 9000 and our internal service mesh was not happy about it -> Not sure about this, feel free to propose a better alternative.
- Pin the version of the app itself (1.152.4) and the cache (0.29.0) -> Those are from the time I tried the deployment, not sure if they are up to date anymore
- Set podSecurityContext for the cache with the group number of the user of the pod -> I'm not too sure about this tbh, but it fixes the fact that nothing could get written on the pod. Please double check.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -23,6 +23,8 @@ permissions:
   contents: read
 
 env:
+  MISE_VERSION: "2026.4.18"
+  MISE_GITHUB_TOKEN: ${{ github.token }}
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
   MISE_GITHUB_ATTESTATIONS: 0
   MISE_LOCKED: 1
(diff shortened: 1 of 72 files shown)


pepicrft (approved):
Thanks a lot @Tomusm. I made some configuration overridable such that's portable and not specific to your organization.

pepicrft:
Regarding the CLA signature, the server/ changes were done by me, so @Tomusm doesn't need to sign anything.

PR #10342, opened 2026-04-21
fix(server): retry ClickHouse flush on socket drops

fortmarek:
## Summary

The ingestion `Buffer.do_flush/1` writes RowBinary batches to ClickHouse via `IngestRepo.query!/3`. The repo-level retry added in [6df900d138](https://github.com/tuist/tuist/commit/6df900d138) only wraps `insert_all`/`insert`, so this path had no recovery when ClickHouse dropped an idle pool connection mid-request.

Under CI load, the `Ch` pool occasionally hands out a connection the server is about to close. The next `query!` raises `Mint.TransportError: socket closed`, the `:insert_and_flush` `GenServer.call` exits, and the test that triggered the flush fails — e.g. the two `Tuist.BuildsTest` failures on [run 24676402525](https://github.com/tuist/tuist/actions/runs/24676402525):

```
(exit) exited in: GenServer.call(Tuist.Builds.Build.Buffer, {:insert_and_flush, <binary>}, :infinity)
   ** (EXIT) an exception was raised:
       ** (Mint.TransportError) socket closed
           (ecto_sql 3.13.2) lib/ecto/adapters/sql.ex:1098: Ecto.Adapters.SQL.raise_sql_call_error/1
           (tuist 0.1.0) lib/tuist/ingestion/buffer.ex:83: Tuist.Ingestion.Buffer.handle_call/3
```

Fix: wrap the flush in `IngestRepo.with_retry/1` so it shares the same 3-attempt exponential-backoff policy as `insert_all`/`insert` and test-teardown `TRUNCATE`s.

- **`idle_interval`:** `DBConnection`'s default `:idle_interval` is 1s, which already pings more aggressively than ClickHouse's 300s `keep_alive_timeout`. The race we see is a connection closed *between* checkout and first byte — no idle ping can avoid it.
- **Bumping `keep_alive_timeout`:** both local and CI launch ClickHouse via `server/mise/tasks/clickhouse/start.sh` with `--keep_alive_timeout=300`, so that knob is already maxed for the flakes we observe.
- **Buffer-local retry:** we deliberately centralized retry into `IngestRepo` ([1d57731155](https://github.com/tuist/tuist/commit/1d57731155)) to avoid duplication. This change reuses that policy instead of reintroducing a parallel one.

## Test plan

- [x] `mix test test/tuist/ingestion/buffer_test.exs` — 12/12 pass
- [ ] CI green on the server pipeline
- [ ] Confirmed by watching the `Tuist.BuildsTest` class of failure disappear from future flake reports

No reliable local repro — the failure only surfaces under CI load. The fix is derived from the stack-trace pattern and the shape of the existing retry policy.


fortmarek:
Will fix the failing Android workflow in a separate PR.

fortmarek:
@pepicrft merging this as this actively breaks all workflows.

happy to make changes to this if you'd do things differently.

PR #10357, opened 2026-04-21
chore(server): Make our ClickHouse development setup work with worktrees

pepicrft:
Our current setup is incompatible with worktrees. We are currently treating ClickHouse as a global process through an assigned port, while configuring it from each worktree to persist its data in a directory scoped to the worktree. This PR changes that by making it truly global, with its storage being globally in the system, while scoping each worktree to a DB whose name is bound to the worktree.

## Summary
- Move dev and test ClickHouse port and database selection into `runtime.exs`
- Keep the local ClickHouse daemon shared across checkouts while retaining per-worktree database names
- Simplify the compile-time config files so they only hold static settings

## Testing
- Verified `Config.Reader.read!` loads `runtime.exs` successfully
- Verified the updated shell scripts still parse with `bash -n`
- Performed a local ClickHouse start/stop smoke check against the shared daemon path

### DIFF
--- a/server/mise/tasks/clickhouse/start.sh
+++ b/server/mise/tasks/clickhouse/start.sh
@@ -3,9 +3,29 @@
 
 set -euo pipefail
 
+CLICKHOUSE_RUNTIME_DIR="${TUIST_SERVER_CLICKHOUSE_RUNTIME_DIR:-${XDG_STATE_HOME:-${HOME}/.local/state}/tuist/clickhouse}"
+CLICKHOUSE_HTTP_PORT="${TUIST_SERVER_CLICKHOUSE_HTTP_PORT:-8123}"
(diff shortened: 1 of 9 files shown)


server/mise/tasks/clickhouse/start.sh
```diff
@@ -3,9 +3,29 @@
 
 set -euo pipefail
 
+CLICKHOUSE_RUNTIME_DIR="${TUIST_SERVER_CLICKHOUSE_RUNTIME_DIR:-${XDG_STATE_HOME:-${HOME}/.local/state}/tuist/clickhouse}"
```
pepicrft, line 6: I'm moving this to work globally with the state living in `~/.local/state/tuist/clickhouse`. That way all the worktrees use the same instance (identified by the port) but work with different databases, each representing a worktree.

fortmarek (approved):

PR #10359, opened 2026-04-21
feat(server): add /ops/accounts dashboard with Enterprise upgrade flow

fortmarek:
image

image

## Summary

Adds an internal ops dashboard at \`/ops/accounts\` that lists every account with its current plan and lets ops drive the three billing actions we previously had to run out of the Stripe dashboard:

- **Manage on Stripe** — opens the Stripe customer portal (creating a bare Stripe customer first if the account doesn't have one yet).
- **Upgrade to Enterprise** — opens a modal to fill in billing name/email/address (country is a dropdown) and monthly/yearly cadence. On submit, we update the Stripe Customer with the address and create or switch the subscription to Enterprise with \`collection_method: send_invoice, days_until_due: 30\`. No card is captured and no redirect to Stripe Checkout — the customer receives an invoice by email, which matches how we sell Enterprise today.
- **Cancel plan** — flags the active subscription as \`cancel_at_period_end\` (after a confirm dialog), so the customer retains access until the end of the billing period; our DB row waits for the webhook to flip.

Pro stays self-serve (it needs a card + our metered pricing, Checkout already handles that well), so there's no "Upgrade to Pro" from ops.

### Under the hood

- \`Tuist.Accounts.list_accounts/1\` is a thin \`Flop.validate_and_run!/3\` wrapper. Cross-column "substring search" is expressed as a \`:search\` custom Flop filter on the \`Account\` schema — Flop's built-in filters AND per-column, so a custom field keeps the call-site clean. Today it matches on \`name\`; broadening later is a one-line change.
- \`Tuist.Billing.upgrade_to_enterprise/2\` handles the customer update plus the create-or-switch subscription flow.
- \`Tuist.Billing.cancel_subscription_at_period_end/1\` wraps the Stripe call and lets the existing webhook-driven \`on_subscription_change\` keep our local \`Subscription\` in sync.
- CSS scoped to \`#ops-accounts\` mirrors \`#ops-cache\` so the page heading and filters render with the shared ops styling.
- New sidebar entry in the ops layout.

### Why invoice-billed Enterprise instead of Checkout?

Enterprise deals are hand-sold and invoice-billed in practice. Collecting the address in-product + creating a \`send_invoice\` subscription avoids bouncing ops through the Stripe Checkout flow (which is card-first) and avoids pulling Stripe Elements into the ops surface. For anything more bespoke (payment methods, tax IDs), **Manage on Stripe** opens the customer portal which already covers it.

## How to test locally

- [ ] Log in as a user listed in \`Tuist.Environment.ops_user_handles/0\` and visit http://localhost:8080/ops/accounts.
- [ ] Search narrows rows by handle; pagination appears once you have more than 30 accounts (seeded dev has 3, so seed more to exercise it or see the test).
- [ ] Open the per-row menu → **Manage on Stripe**; you should be redirected to the Stripe customer portal (the portal must be configured in the Stripe dashboard for the test key).
- [ ] Open the per-row menu → **Upgrade to Enterprise**; fill in the modal (address required, country dropdown includes \`US\`, \`GB\`, \`DE\`, etc.) and submit. The Stripe Customer should have the new address, and a \`send_invoice\` Enterprise subscription should exist; the row's Plan badge flips to "Enterprise" after the webhook syncs.
- [ ] For a Pro/Enterprise account, open the menu → **Cancel plan**, confirm; the Stripe subscription should have \`cancel_at_period_end: true\`.
- [ ] \`mise run dev\` + \`mix test test/tuist_web/live/ops_accounts_live_test.exs\` covers list/search/pagination/manage/upgrade/cancel (6 tests).


server/test/tuist/accounts_test.exs
```diff
@@ -3975,4 +3975,26 @@ defmodule Tuist.AccountsTest do
...
+        Accounts.list_accounts(%{
+          page: 1,
+          page_size: 10,
+          filters: [%{field: :search, op: :==, value: "acme"}]
+        })
+
+      # Then
+      names = accounts |> Enum.map(& &1.name) |> Enum.sort()
+      assert names == ["acme", "acmetools"]
+      assert meta.total_count == 2
```
pepicrft: This is unnecesssary, no?

server/lib/tuist_web/components/app_layout_components.ex
```diff
@@ -375,6 +375,12 @@ defmodule TuistWeb.AppLayoutComponents do
         navigate={~p"/ops"}
         selected={@current_path == "/ops"}
       />
+      <.sidebar_item
```
pepicrft, line 378: What about hiding this for on-premis instances?

pepicrft (approved):

PR #10366, opened 2026-04-22
feat(server): per-account Prometheus /metrics endpoint + Grafana app plugin

fortmarek:
Implements [RFC: Per-account /metrics endpoint and Grafana integration](https://community.tuist.dev/t/per-account-metrics-endpoint-and-grafana-integration/974).

## Summary

- **Server**: `GET /api/accounts/:handle/metrics` returns Prometheus/OpenMetrics text with per-account build, test, cache, and CLI counters, authenticated by a new `account:metrics:read` account-token scope and rate-limited to ~1 req/10s.
- **CLI**: OpenAPI client regenerated so `tuist account tokens create --scopes account:metrics:read ...` works.
- **Grafana plugin** (`grafana/`, id `tuist-tuist-app`): configuration page with Alloy snippet generator plus two curated dashboards (`Tuist Xcode`, `Tuist Gradle`) mirroring the Tuist project dashboard's layout.
- **Docs**: new [Grafana integration guide](/en/guides/integrations/grafana) plus a [Metrics reference](/en/references/metrics) page whose body is generated at page-compile time from `Tuist.Metrics.Schema`, so it can't drift from the scrape endpoint.
- **Release pipeline**: on every `main` push with a `(grafana)`-scoped commit, `.github/workflows/release.yml` bumps `package.json`, regenerates the CHANGELOG, signs (`--signature-type community`), and ships a GitHub release with the zip attached.

## Server architecture

- ETS-backed per-account aggregator in `Tuist.Metrics.Aggregator` — counters use atomic `:ets.update_counter/3`; histograms serialised through the owning GenServer.
- Cluster fan-out in `Tuist.Metrics.snapshot/1`: scrapes RPC every peer and merge snapshots so rolling deploys don't oscillate.
- Telemetry bridge in `Tuist.Metrics.Telemetry` — subscribes to `[:tuist, :run, :command]`, `[:tuist, :cache, :event]`, and new `[:tuist, :metrics, …]` events emitted from build/test/gradle write sites.
- Declarative schema in `Tuist.Metrics.Schema` with bounded-cardinality labels and shared histogram buckets. The same module drives the [Metrics reference](/en/references/metrics) doc via `Tuist.Metrics.Schema.Markdown.render/0`.

Metric families exposed (per namespace):

| Namespace | Counters | Histograms |
|---|---|---|
| Xcode | build_runs_total, test_runs_total, test_cases_total, cache_events_total | build_run_duration_seconds, test_run_duration_seconds |
| Gradle | build_runs_total, test_runs_total | build_run_duration_seconds, test_run_duration_seconds |
| CLI | invocations_total | invocation_duration_seconds |

## Grafana plugin

- Thin scaffold built with `@grafana/create-plugin`, using pnpm + mise-pinned Node.
- Config page stores the token as `secureJsonData`; exposes an Alloy scrape-snippet generator parameterised by server URL + handle + interval.
- Dashboards are **single-project** (exact `project="$project"` filter), organised into rows matching the Tuist dashboard pages:
  - **Tuist Xcode**: Builds / Tests / Xcode cache
  - **Tuist Gradle**: Builds / Tests
- Duration timeseries show p50, p90, p99, and avg — matching the Tuist Builds page's duration selector.

## Architectural choice: app plugin vs datasource plugin

Grafana integrations come in two shapes:

- **App plugin (scrape-based)** — expose a Prometheus-style `/metrics` endpoint; plugin ships configuration UI + scrape-snippet generator + dashboards. Users run an Alloy/Agent collector that remote-writes to their Prometheus. PromQL, rate/trend queries, and alerting all work out of the box. Precedent: every Grafana Cloud integration for OSS services (Linux Host, NGINX, MySQL, Redis, Kafka, Kubernetes, MongoDB…).
- **Datasource plugin (query-time)** — plugin queries a vendor-hosted query API at panel-render time. Credentials live entirely in Grafana's `secureJsonData`; no collector, no env var. Precedent: every commercial SaaS integration (Sentry, Datadog, New Relic, Honeycomb, GitHub, Jira).

### Why this PR ships the app-plugin shape

1. Our `/metrics` endpoint returns an in-memory ETS aggregator snapshot, not a time-series. `rate()` / `increase()` / multi-day trends require sampling over time and persisting the samples — that's what a Prometheus + scraper pair provides. Building the equivalent on the Tuist side means a PromQL-compatible (or custom) query API over ClickHouse, which is a distinct product commitment.
2. Matches the "your Prometheus, your retention, your PromQL" property self-hosted customers expect — no vendor lock-in on the metrics layer.
3. The endpoint is a day of work; a query API is a quarter.
4. Gets a Tuist entry into the Grafana plugin catalogue quickly so the integration has a public landing page and install path before we invest in the datasource track.

### The datasource plugin is the likely next track, not a "someday"

After a deeper spike into the tradeoffs, the datasource plugin is probably the shape that actually matches Tuist's user base better, and we should scope it as a dedicated follow-up project rather than drift into it opportunistically.

**User-base reality check**. The scrape-based app plugin assumes users run Alloy / Grafana Agent / vanilla Prometheus somewhere. Segmentation of Tuist's actual user base:

| Segment | Share (rough) | Already runs Prometheus/Alloy | Realistic to expect setup |
|---|---|---|---|
| Solo devs / indie studios | ~40% | No | No |
| Small mobile teams (2–10, no SRE) | ~35% | No | Unlikely |
| Mid-size mobile (10–50, some platform) | ~20% | Sometimes | Maybe |
| Large orgs with dedicated SRE | ~5% | Yes | Yes |

Tuist is a mobile-tooling product. The typical user opens Xcode and wants to know why builds are slow — not a platform engineer running a Kubernetes-monitoring Alloy fleet. The Grafana Cloud caveat compounds this: **Fleet Management is a paid Advanced/Enterprise feature**; Free and Pro users don't have it. The indie dev trying Tuist on Grafana Cloud Free would be stuck at the "deploy a collector" step.

So the app plugin in this PR primarily serves the ~5% of users who already operate Prometheus — the audience that needs Tuist's help *least*. The datasource plugin serves the other ~95%.

**Pros / cons of a datasource plugin** (vs this PR's app plugin):

| | App plugin (this PR) | Datasource plugin (future) |
|---|---|---|
| Setup friction | 4 steps incl. run a collector | 2 steps: install + paste token |
| Works on Grafana Cloud Free | Needs user-run Alloy | Zero config |
| Credentials location | Grafana + collector env var | Grafana `secureJsonData` only |
| Historical data | User's Prometheus retention | Tuist's ClickHouse retention |
| Freshness | Scrape interval (10–15s) | Live at panel render |
| Cardinality (commit SHA, branch, device, module) | Explodes Prometheus | Free — scoped ClickHouse queries |
| Alerting path | Prometheus + Alertmanager (mature) | Grafana Alerting only |
| PromQL federation with other metrics | Yes | No |
| If Tuist is down | Prometheus still serves history | Every dashboard breaks |
| ClickHouse read load | Zero (just scrape aggregates) | Per-panel-render |
| Our ongoing cost | Endpoint + schema | Query API + caching + versioning |
| Data portability if user leaves | Takes their Prometheus with them | History stays in our infra |

**Implementation outline if/when we take it on**. Three pieces, roughly 5–6 weeks of one engineer for an MVP:

1. **Server-side query API** (`POST /api/accounts/:handle/metrics/query`, ~2–3 weeks). Canned-queries flavor to start — 8–10 named queries with typed parameters covering the Xcode/Gradle dashboard needs (`build_duration_percentile`, `build_count_by_status`, `test_success_rate`, `test_case_flakiness_top_n`, `cache_hit_rate`, `cli_invocations_by_command`, etc.). Response in Grafana DataFrame shape. Reuse the `account:metrics:read` scope; rate-limit at ~1 req/s per account (dashboard refreshes are burstier than scrapes); add `Last-Modified`/`If-Modified-Since` for Grafana's datasource cache. Skip full PromQL support — deceptively hard and not worth the scope for a v1.

2. **Datasource plugin** (~2 weeks). Scaffold with `pnpm dlx @grafana/create-plugin@latest --type datasource`, plugin id `tuist-tuist-datasource`. Frontend (React/TS): `ConfigEditor` (same three fields as the current app plugin), `QueryEditor` (dropdown of query types, parameter inputs, project/scheme auto-populated via `metricFindQuery`), `DataSource.ts` implementing `query`, `testDatasource`, `metricFindQuery`. Backend (Go, ~300 LOC, recommended not strictly required): `QueryData` handler proxying to our API with the `secureJsonData` token injected as `Authorization: Bearer`. The backend matters for Grafana Alerting to evaluate queries server-side and to keep the token out of the browser.

3. **Dashboards bound to the datasource** (~1 week). Two JSONs (Xcode, Gradle) that target the datasource query shape. Can't reuse the PromQL dashboards — different query language. Both sets coexist in the catalogue; users pick the plugin that matches how they run Grafana.

Plus ~0.5 week for signing, release pipeline, and catalogue submission. First-time review for a separate plugin ID takes 1–3 business days on Grafana's side.

**Bonus: the query API is reusable.** Once it exists behind the datasource plugin with a stable contract, the same endpoint unblocks a future `tuist metrics query …` CLI subcommand, Slack integrations, custom dashboards in other tools, and any analytics UI we want to ship inside Tuist itself. The framing is really "a public Tuist analytics API, first surfaced through Grafana" — which is a nicer product story than "a Grafana-specific endpoint". Worth keeping that in mind when we sit down to design the query shape.

## How this was tested end-to-end

No `mise run dev` script (intentionally removed to keep the repo surface small). The full pipeline below is what I ran on `staging.tuist.dev` to confirm server → Prometheus → Grafana → dashboards.

1. **Deploy the server to staging** via `Server Staging Deployment` workflow_dispatch with this branch's SHA. Verify `https://staging.tuist.dev/ready` returns `200` afterwards.

2. **Create a metrics token** using your existing staging OAuth session — no new 1Password entry needed:

   ```bash
   ACCESS=$(jq -r .accessToken ~/.config/tuist/credentials/staging.tuist.dev.json)
   curl -sS -X POST https://staging.tuist.dev/api/accounts/<handle>/tokens \
     -H "Authorization: Bearer $ACCESS" \
     -H "Content-Type: application/json" \
     -d '{"scopes":["account:metrics:read"],"name":"grafana-e2e"}' | jq -r .token
   ```

3. **Smoke-scrape the endpoint** to see the 11 metric families show up (empty bodies until traffic is fired):

   ```bash
   curl -H "Authorization: Bearer $TOKEN" https://staging.tuist.dev/api/accounts/<handle>/metrics | head
   ```

4. **Generate staging traffic** so the counters move. Easiest is a minimal Tuist project:

   ```bash
   mkdir /tmp/tuist-smoke && cd /tmp/tuist-smoke
   cat > Tuist.swift <<'SWIFT'
   import ProjectDescription
   let tuist = Tuist(fullHandle: "<handle>/<project>", url: "https://staging.tuist.dev")
   SWIFT
   cat > Project.swift <<'SWIFT'
   import ProjectDescription
   let project = Project(name: "Smoke", targets: [
     .target(name: "Smoke", destinations: .iOS, product: .app,
             bundleId: "dev.tuist.Smoke", sources: [])
   ])
   SWIFT
   for _ in 1 2 3; do tuist generate --no-open; done
   ```

   Within a few seconds, `tuist_cli_invocations_total{command="generate",project="<handle>/<project>"}` appears in the scrape.

5. **(Optional) synthesise Xcode builds** by POSTing directly to `/api/projects/<handle>/<project>/builds` with varied scheme/status/duration payloads. See the RFC discussion thread for a sample loop.

6. **Build + sign the plugin locally**:

   ```bash
   cd grafana
   export GRAFANA_ACCESS_POLICY_TOKEN=$(op read 'op://tuist/TUIST_GRAFANA_PLUGIN_TOKEN/password')
   mise run bundle --signature-type community
   ```

   Produces `grafana/tuist-tuist-app-<version>.zip`.

7. **Load the plugin in a local Grafana** against staging:

   ```bash
   docker run --rm -p 3000:3000 \
     -e GF_PLUGINS_ALLOW_LOADING_UNSIGNED_PLUGINS=tuist-tuist-app \
     -e GF_AUTH_ANONYMOUS_ENABLED=true -e GF_AUTH_ANONYMOUS_ORG_ROLE=Admin \
     -v "$PWD/dist:/var/lib/grafana/plugins/tuist-tuist-app" \
     grafana/grafana:latest
   ```

   Open <http://localhost:3000/plugins/tuist-tuist-app>, **Enable**, open **Configuration**. Fill in the server URL (`https://staging.tuist.dev`), handle, and token from step 2. Save.

8. **Run a Prometheus that scrapes staging** (pointing your own instance or a throwaway container):

   ```yaml
   scrape_configs:
     - job_name: tuist-staging
       scheme: https
       metrics_path: /api/accounts/<handle>/metrics
       authorization:
         type: Bearer
         credentials: <token>
       static_configs:
         - targets: ["staging.tuist.dev"]
   ```

   Confirm `up{job="tuist-staging"} == 1` in Prometheus.

9. **Import the dashboards** from the plugin's **Dashboards** tab. Open `Tuist Xcode` — the stat cards populate from the traffic generated in step 4/5 within one scrape interval.

## Release pipeline

For every `main` push that includes a `(grafana)`-scoped commit:

1. `release-grafana` reads `op://tuist/TUIST_GRAFANA_PLUGIN_TOKEN/password`, bumps `grafana/package.json`, regenerates `grafana/CHANGELOG.md`, runs `mise run bundle --signature-type community`.
2. `commit-and-release` publishes a GitHub release (`grafana@<version>`) with the signed zip attached.
3. First-time only: submit the zip to the public Grafana plugin catalogue at <https://grafana.com/auth/sign-in?redirectPath=/plugins/submit>. After approval, subsequent releases are ingested automatically.

## RFC comments addressed

- ✅ `[:tuist, :metrics, …]` event namespace separates user-facing metrics from internal telemetry.
- ✅ Metrics docs generated from the schema (`{{TUIST_METRICS_TABLE}}` marker expanded at page-compile time from `Tuist.Metrics.Schema`).


grafana/src/components/AppConfig.tsx
fortmarek, line 1: Image

server/lib/tuist/metrics/schema.ex
fortmarek, line 1: this is one of the more important pieces cc @pepicrft

grafana/.config/types/bundler-rules.d.ts
```diff
@@ -0,0 +1,37 @@
+// Image declarations
+declare module '*.gif' {
```
pepicrft, line 2: Were these files scaffolded by the Grafana's init workflow?

grafana/.config/types/webpack-plugins.d.ts
```diff
@@ -0,0 +1,83 @@
+declare module 'replace-in-file-webpack-plugin' {
```
pepicrft, line 1: Quite odd they are still using Webpack, but I guess it does its job.

grafana/mise/tasks/test.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+#MISE description="Test the Tuist Grafana app plugin"
+set -euo pipefail
+pnpm run test:ci
```
pepicrft, line 8: How do `ci` tests differ from the local ones?

grafana/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
pepicrft, line 1: I don't think we need this since it'll be inherited from the one at the root.

server/lib/tuist/metrics/telemetry.ex
```diff
@@ -0,0 +1,243 @@
...
+          command: event_string(event, :name),
+          is_ci: to_string(!!event_get(event, :is_ci)),
+          status: command_status(event)
+        }
+    end
+  end
+
+  defp cli_labels(_), do: nil
+
+  defp cache_labels(%{project_id: project_id}) do
```
pepicrft, line 171: This new cache bridge assumes cache events always carry `project_id`, but the existing `Tuist.Telemetry.event_name_cache()` emits from `Tuist.CommandEvents.create_command_event/1` only send `count` and `event_type`. That means `cache_labels/1` returns `nil` for the main Xcode cache path, so `tuist_xcode_cache_events_total` never increments and the new cache panels stay empty.

server/lib/tuist/tests.ex
```diff
@@ -312,6 +312,8 @@ defmodule Tuist.Tests do
             "#{project.account.name}/#{project.name}",
             :test_created
           )
+
+          emit_test_run_metrics(test, project, test_case_runs)
         end)
```
pepicrft, line 322: This emits test-run metrics from the initial insert path, but multi-shard runs start life as `in_progress` and get finalized later in the update branch. Because that later branch never calls `emit_test_run_metrics/3`, sharded runs remain counted once under `in_progress` with the first shard duration instead of their eventual success or failure.

server/lib/tuist/metrics/telemetry.ex
```diff
@@ -0,0 +1,243 @@
...
+  end
+
+  # ---- Utility -----------------------------------------------------------
+
+  defp ms_to_seconds(ms) when is_integer(ms), do: ms / 1_000
+  defp ms_to_seconds(ms) when is_float(ms), do: ms / 1_000
+  defp ms_to_seconds(_), do: 0
+
+  defp command_status(event) do
+    if Map.get(event, :status) == :success or event_get(event, :error_message) in [nil, ""] do
```
pepicrft: `Event.changeset/1` normalizes command status to `0` or `1` before telemetry sees it, but this check only treats the atom `:success` as success and otherwise falls back to whether `error_message` is present. Since the analytics API does not require `error_message` on failures, a failure with status `1` and no message will be exported as success in `tuist_cli_invocations_total` and the duration histogram.

server/lib/tuist/metrics/aggregator.ex
```diff
@@ -0,0 +1,224 @@
+defmodule Tuist.Metrics.Aggregator do
```
pepicrft, line 1: With the growth of metrics and tenants, I expect the memory used by this one to grow, so I'd consider adding some telemetry there so we can know the state of this in every instance, and setup alerts down the road.

pepicrft, line 1: Let's add some tests to this one to prevent regressions in the aggregation logic.

server/lib/tuist/metrics/telemetry.ex
```diff
@@ -0,0 +1,243 @@
...
+    :telemetry.attach_many(@handler_id, @events, &__MODULE__.handle_event/4, nil)
+  end
+
+  def detach do
+    :telemetry.detach(@handler_id)
+  end
+
+  @doc false
+  def handle_event(event, measurements, metadata, _config) do
+    handle(event, measurements, metadata)
```
pepicrft, line 42: `:telemetry` events are processed synchronously, so anything happening in the handlers here will block the caller. I'm not sure what the performance of `Metrics.xxxx` APIs is, but I might consider doing the processing asynchronously, potentially with Broadway so we have control over the processing resources.

server/test/tuist/metrics/telemetry_test.exs
```diff
@@ -0,0 +1,223 @@
...
+  alias TuistTestSupport.Fixtures.ProjectsFixtures
+
+  setup do
+    MetricsTelemetry.attach()
+    on_exit(&MetricsTelemetry.detach/0)
+
+    case GenServer.whereis(Aggregator) do
+      nil -> start_supervised!(Aggregator)
+      _ -> Aggregator.reset()
+    end
```
pepicrft: Why do we need this? I'd expect the application supervision tree to initialize it.

server/test/tuist/metrics/telemetry_test.exs
```diff
@@ -0,0 +1,223 @@
...
+  use TuistTestSupport.Cases.DataCase, async: false
+
+  alias Tuist.Metrics
+  alias Tuist.Metrics.Aggregator
+  alias Tuist.Metrics.Telemetry, as: MetricsTelemetry
+  alias TuistTestSupport.Fixtures.AccountsFixtures
+  alias TuistTestSupport.Fixtures.ProjectsFixtures
+
+  setup do
+    MetricsTelemetry.attach()
```
pepicrft, line 11: I'd ask the agent if it can turn this one into `async: true`. The telemetry handlers are global, so that's something we can't change, but events can have additional metadata, which we can pass here, and that can be used to _point_ to the aggregator, whose lifecycle can be bound to the test.

server/test/tuist/metrics_test.exs
```diff
@@ -0,0 +1,151 @@
...
+  alias Tuist.Metrics
+  alias Tuist.Metrics.Aggregator
+
+  setup do
+    # The Aggregator is normally supervised. If a previous test left counters
+    # in the ETS table, clean them.
+    case GenServer.whereis(Aggregator) do
+      nil -> start_supervised!(Aggregator)
+      _ -> Aggregator.reset()
+    end
```
pepicrft: I'd assume `Aggregator` is running, since we have it in the application supervised tree, and just call reset here. This logic reads very defensive about something else changing elsewhere in the app.

server/lib/tuist.ex
```diff
@@ -101,6 +101,12 @@ defmodule Tuist do
       Xcode.XcodeProject,
       Xcode.XcodeTarget,
       Loops,
+      Metrics,
```
pepicrft, line 104: Every time I see this, I wonder if we should invest some time in turning it into something valuable towards ensuring the codebase scales well architecture-wise. Our current usage is just adding pieces to the list to mitigate a warning.

server/priv/docs/en/guides/integrations/grafana.md
```diff
@@ -0,0 +1,58 @@
...
+account. The **Tuist** Grafana app plugin bundles two dashboards (Xcode
+and Gradle) and a one-click scrape-config generator for Grafana Alloy or
+Agent.
+
+## Setup {#setup}
+
+### 1. Install the plugin
+
+On Grafana Cloud: *Administration → Plugins → search "Tuist" → Install*.
+On self-hosted Grafana: `grafana-cli plugins install tuist-tuist-app`.
```
pepicrft, line 20: `tuist-tuist-app` reads redundant and also inconsistent with the fact it's a plugin. Shouldn't it be `tuist-plugin`? What's an app?

server/priv/docs/en/guides/integrations/grafana.md
```diff
@@ -0,0 +1,58 @@
...
+
+### 2. Create an account token
+
+```bash
+tuist account tokens create <your-handle> \
```
pepicrft, line 25: Out of curiosity, do they support OIDC? It'd be a much better integration experience for the user.

server/priv/docs/en/guides/integrations/grafana.md
```diff
@@ -0,0 +1,58 @@
...
+does not show it again.
+
+### 3. Configure the plugin
+
+In *Administration → Plugins → Tuist → Configuration*, fill in:
+
+- **Server URL** — `https://tuist.dev`, or your self-hosted URL.
+- **Account handle** — the account you want to scrape.
+- **Metrics token** — the `tuist_...` value from step 2.
+- **Scrape interval** — defaults to `15s` (minimum `10s`).
```
pepicrft, line 40: I believe we are currently holding unbounded memory metrics for all accounts, which can be an issue from a memory perspective. I'd suggest to clean them periodically, and then that piece of information is relevant here, because it determines the ceiling of this interval.

pepicrft: I'd also maybe consider making this an opt-in. It doesn't make sense to use memory resources for accounts that are not interested in this.

server/test/tuist_web/controllers/api/metrics_controller_test.exs
```diff
@@ -0,0 +1,152 @@
...
+  alias Tuist.Accounts
+  alias Tuist.Environment
+  alias Tuist.Metrics
+  alias Tuist.Metrics.Aggregator
+  alias TuistTestSupport.Fixtures.AccountsFixtures
+
+  setup do
+    stub(Environment, :tuist_hosted?, fn -> true end)
+
+    case GenServer.whereis(Aggregator) do
```
pepicrft: Same here, can't we assume this is already up and running as part of the app supervision tree?

pepicrft (approved):

grafana/.config/types/bundler-rules.d.ts
fortmarek: yup

grafana/.config/types/webpack-plugins.d.ts
fortmarek: agree, but that's what the scaffold went with 🤷

grafana/LICENSE
fortmarek: I thought so, too, but claude mentioned it's better if the license is additionally in the root of the plugin once we ask for the plugin review by grafana.

grafana/mise/tasks/test.sh
fortmarek: otherwise, tests are watched ... this was also created by the grafana scaffold. not in live with this pattern, but would maybe keep the plugin as standard as possible.

grafana/.config/types/bundler-rules.d.ts
fortmarek: Yes — the entire `.config/` tree is scaffolded by `pnpm dlx @grafana/create-plugin@latest`. We re-run `update --force` periodically to pull upstream improvements; deliberate deviations are tracked in [grafana/AGENTS.md](https://github.com/tuist/tuist/blob/main/grafana/AGENTS.md#upgrading-grafanacreate-plugin).

grafana/.config/types/webpack-plugins.d.ts
fortmarek: Agreed — `@grafana/create-plugin` is still webpack-default as of v12. Their Rspack option is behind a flag and has known bugs with their test utilities. Will swap upstream eventually.

grafana/mise/tasks/test.sh
fortmarek: Addressed in [`e044d03dbb`](https://github.com/tuist/tuist/pull/10366/commits/e044d03dbb) — added an inline comment in `test.sh`. tl;dr: `test:ci` = `jest --passWithNoTests --maxWorkers 4` (one-shot), `test` = `jest --watch --onlyChanged` (TDD loop). The "ci" suffix is inherited from `@grafana/create-plugin`'s scaffold naming; both CI and local pre-push hooks want the one-shot form.

grafana/LICENSE
fortmarek: Kept because Grafana's plugin catalogue validator expects a LICENSE file **inside the bundled zip**, not just at the repo root. Webpack copies `grafana/LICENSE` into `dist/` during bundle. Could point webpack at the root `LICENSE.md` instead (needs `toType: 'file'` to avoid creating a directory); happy to de-duplicate as a follow-up if you'd prefer a single source of truth.

server/lib/tuist/metrics/telemetry.ex
fortmarek: Fixed in [`809c9e441a`](https://github.com/tuist/tuist/pull/10366/commits/809c9e441a). Threaded `project_id` onto the `[:tuist, :cache, :event]` emissions in `Tuist.CommandEvents.create_command_event/1` so `cache_labels/1` can resolve the account. The existing listener (prom_ex_plugin) doesn't care about the extra metadata field.

server/lib/tuist/tests.ex
fortmarek: Fixed in [`809c9e441a`](https://github.com/tuist/tuist/pull/10366/commits/809c9e441a). `emit_test_run_metrics/3` is now skipped while `status == "in_progress"` and called from the shard-update branch once the final shard flips the run to success/failure. Metrics emit exactly once per run with the correct terminal status and accumulated duration.

server/lib/tuist/metrics/telemetry.ex
fortmarek: Fixed in [`809c9e441a`](https://github.com/tuist/tuist/pull/10366/commits/809c9e441a). `command_status/1` now matches integer `0`/`1` (what `Event.changeset/1` actually normalizes to) and the `:success`/`:failure` atoms; dropped the `error_message`-presence fallback since the API doesn't require a message on failures. Regression test added in `telemetry_test.exs`.

server/lib/tuist/metrics/aggregator.ex
fortmarek: Added [`Aggregator.table_info/0`](https://github.com/tuist/tuist/pull/10366/commits/809c9e441a) returning `%{size, memory_words}` as a hook point. Wiring it to a periodic `[:tuist, :metrics, :aggregator, :stats]` telemetry emission is scoped as a follow-up so the scheduling interval + alert thresholds get their own review. The memory-ceiling work and opt-in-per-account thread linked separately make more sense in the same follow-up PR.

fortmarek: Added [`Tuist.Metrics.AggregatorTest`](https://github.com/tuist/tuist/pull/10366/commits/809c9e441a#diff-fb6de16b4bca6551f60cb41b83c1f0a2b1eb826d8b4eee36e9f9b2b9f9b) in [`809c9e441a`](https://github.com/tuist/tuist/pull/10366/commits/809c9e441a) covering: counter atomicity under concurrent writes (200 parallel increments), histogram bucket distribution + late observations, account isolation (writes under account A don't leak into account B's snapshot), reset, and a smoke test for the new `table_info/0` helper.

server/lib/tuist/metrics/telemetry.ex
fortmarek: Agreed — this is the right direction. The bridge handlers do a Postgres lookup (`Tuist.Projects.get_project_by_id/1`) inline with the emitter, which is exactly the kind of work that should sit behind back-pressure. Broadway gives us that plus isolation of the metrics path from the hot path. Keeping it scoped as a separate PR since it touches every emitter + the telemetry test shape (would let us flip to `async: true` at the same time — #3130024056). Flagged it as a follow-up in my consolidated reply.

server/test/tuist/metrics/telemetry_test.exs
fortmarek: Fixed in [`809c9e441a`](https://github.com/tuist/tuist/pull/10366/commits/809c9e441a). Dropped the defensive `case GenServer.whereis(Aggregator) do nil -> start_supervised!/1 ...` in all three test files — just `Aggregator.reset()` in `setup`, trusting the supervised instance. You were right that it read as if something elsewhere was tearing it down.

server/test/tuist/metrics/telemetry_test.exs
fortmarek: Agreed — it's the cleanest path. The emitter would need to accept a per-test aggregator ref through telemetry metadata (or stash it in `Process.put/get` as you suggested). Folding it into the Broadway refactor in #3129979975 since both refactors touch the same surface and the test shape depends on the new async pipeline.

server/test/tuist/metrics_test.exs
fortmarek: Fixed in [`809c9e441a`](https://github.com/tuist/tuist/pull/10366/commits/809c9e441a) — `Aggregator.reset()` in setup only.

server/lib/tuist.ex
fortmarek: Agreed — this module is acting as a manual registry for Boundary. This PR keeps following the existing convention so the scope stays on metrics, but happy to back a separate issue to replace it with real architecture guardrails (per-directory boundaries, layered dependencies). Wanted to flag it for the same reason every time I touch it.

server/priv/docs/en/guides/integrations/grafana.md
fortmarek: Grafana enforces `<org-slug>-<plugin-name>-<plugin-type>` IDs across the catalogue — we can't drop the org prefix even though ours matches the product name. The double `tuist` happens because our Grafana org and the product share the name; same pattern as `redis-redis-datasource`, `percona-percona-datasource`, `mongodb-mongodb-datasource` in the public catalogue.

The `-app` suffix is Grafana's *plugin type*, not a product label: `app` = bundle of UI pages + dashboards + config (what we have), vs `datasource` = answers queries at panel-render time, vs `panel` = custom visualization. So the "it's a plugin" angle is actually covered — it's describing *which kind* of plugin. Keeping as is given it's also already community-signed under this ID.

server/priv/docs/en/guides/integrations/grafana.md
fortmarek: Grafana Cloud has OIDC for *user sign-in to Grafana itself*, but the plugin catalogue's config-page flow for third-party plugins doesn't have an OIDC path — plugins reach their own backends with whatever credentials the user pastes into the config page.

Accepting Tuist OAuth tokens (not just account tokens) for a "sign in with Tuist to configure the plugin" flow would be a much nicer UX, but it's a genuinely new server-side auth flow (dynamic client registration or a pre-registered Grafana client) and a UI rewrite on the plugin config page. I've scoped it alongside the datasource-plugin track in the [PR description](https://github.com/tuist/tuist/pull/10366#issue-body-top).

server/priv/docs/en/guides/integrations/grafana.md
fortmarek: Agreed — the ETS table currently grows with unique `(account_id, metric, labels)` tuples and only clears on app restart. Two follow-up paths: (a) periodic prune of entries not observed in the last N hours, (b) a per-account cap that drops the oldest entries. (a) is easier to ship; (b) is more principled. Will fold this into the same follow-up as the periodic memory telemetry (#3129925069) so the observability and the cleanup land together with consistent semantics.

fortmarek: Strong agree — for Tuist Cloud specifically, ~99% of accounts will never scrape metrics, so pre-aggregating their telemetry in ETS is pure waste. Obvious design: a boolean on the `Account` schema (or a feature-flag toggle) that the aggregator checks on the write path. Landing it as its own change since it touches account settings UI + a migration, not just the metrics module.

server/test/tuist_web/controllers/api/metrics_controller_test.exs
fortmarek: Fixed in [`809c9e441a`](https://github.com/tuist/tuist/pull/10366/commits/809c9e441a) — `Aggregator.reset()` in setup only.

server/lib/tuist.ex
fortmarek: > Every time I see this, I wonder if we should invest some time in turning it into something valuable towards ensuring the codebase scales well architecture-wise. Our current usage is just adding pieces to the list to mitigate a warning.

(me talking): agree, although, I think it does serve some sort of purpose by limiting what's accessible in cross-module references. But probably worth iterating on this further.

server/lib/tuist/metrics/aggregator.ex
fortmarek: Update: pulled this into the PR in [`4b8faa63b4`](https://github.com/tuist/tuist/pull/10366/commits/4b8faa63b4) instead of deferring. The Aggregator now schedules `:emit_stats` every 60s (configurable via `:stats_interval_ms`) and publishes `[:tuist, :metrics, :aggregator, :stats]` carrying `%{size, memory_bytes, memory_words}`. Attach a handler in prod to forward to Prometheus / StatsD / wherever the ops alerts live. Regression test covers the emission path end-to-end.

server/priv/docs/en/guides/integrations/grafana.md
fortmarek: Good push — ran the numbers and partially walked it back. Here's the calculus on a realistically-active account (3 projects, 2 schemes each, ~15 CLI commands, 2 concurrent Xcode versions, is_ci × status combos):

| Metric | label combos | ETS rows per combo | rows |
|---|---|---|---|
| tuist_xcode_build_runs_total (counter) | ~54 | 1 | 54 |
| tuist_xcode_build_run_duration_seconds (histogram) | ~54 × 1.5 (versions) | 13 (11 buckets + count + sum) | ~1,050 |
| tuist_xcode_test_runs / duration | same | | ~1,100 |
| tuist_xcode_test_cases_total | ~24 | 1 | 24 |
| tuist_xcode_cache_events_total | ~9 | 1 | 9 |
| Gradle × 4 (counter + histogram × 2) | ~same shape | | ~1,100 |
| tuist_cli_invocations / duration | ~180 combos × (1 + 12) | | ~2,340 |

At ~200B per row (key + value + ETS overhead), that's **~5,500 rows ≈ 1 MB per active account**. 1K active accounts = ~1 GB ETS — uncomfortable but not critical. 10K active = ~10 GB — genuine problem.

**Biggest lever**: histograms multiply label cardinality by ~13 (every bucket is its own row), so dropping a version label from a histogram saves ~13 rows per combo, vs 1 row for a counter. Also: [no bundled dashboard queries version labels](https://github.com/tuist/tuist/blob/main/grafana/src/dashboards) (I grepped).

Pushed [`3cc9c9ff92`](https://github.com/tuist/tuist/pull/10366/commits/3cc9c9ff92) dropping `xcode_version`/`macos_version` from the Xcode duration histograms and `gradle_version`/`jvm_version` from the Gradle duration histograms. Versions still live on the matching counters for drill-down queries (`rate(tuist_xcode_build_runs_total{xcode_version="15.4"}[5m])` still works).

Concrete effect: for any customer with >1 concurrent toolchain version in flight (the common case during upgrades), per-account histogram memory drops by 40–50%. The `[:tuist, :metrics, :aggregator, :stats]` event from [`4b8faa63b4`](https://github.com/tuist/tuist/pull/10366/commits/4b8faa63b4) makes the real effect measurable in prod.

**My take on opt-in**: I'd hold off. With the histogram trim + the stats telemetry in this PR, we can watch per-node memory in production and revisit the opt-in decision when we actually have signal. At current scale (≤1K active), memory is not a problem. At 10K+ active, we'll want something — but the right something might be periodic eviction by last-seen, or a global ETS cap with alerting, not necessarily per-account opt-in. Better to decide against real numbers than guess now.

fortmarek: Pulled the eviction in — [`11a9f69f2d`](https://github.com/tuist/tuist/pull/10366/commits/11a9f69f2d).

Design: a separate `:tuist_metrics_last_scrape` ETS table tracks per-account timestamps.

- **Seed on first write.** `increment_counter` and `observe_histogram` do `:ets.insert_new` on the timestamp table. Brand-new accounts get a clock that starts ticking from their first observation. Subsequent writes don't perturb an already-present timestamp, so the write path stays essentially free for active accounts.
- **Refresh on scrape.** `Aggregator.snapshot/1` records a scrape each time it's called — which means both the local node's scrape AND every `:rpc.call` during the cluster fan-out refresh the timestamp on the node they land on. Any account actively reading `/metrics` stays alive on every node that holds its data.
- **Sweep every 5 minutes** (configurable). Finds accounts with a last-scrape older than 30 minutes (configurable). Drops their counters + histograms from the main table and removes the timestamp entry. 30-minute staleness comfortably tolerates a 1-minute scrape interval missing dozens of cycles — so a collector restart or network blip doesn't lose a customer's dashboards.

Emits `[:tuist, :metrics, :aggregator, :evicted]` with `%{accounts, rows}` per sweep so ops can tell normal churn apart from cardinality explosions. Knobs live on `start_link/1` via `:eviction_interval_ms` and `:scrape_staleness_ms` options if we want to tune them per env.

This makes the opt-in feature flag unnecessary for the memory argument: any account that isn't actively scraped goes away. Accounts that are scraping are the ones that explicitly want the data. The feature-flag conversation can be purely about product ("should this endpoint be available on free tier?") rather than an infra guardrail.

server/lib/tuist.ex
pepicrft: Yeah... I think it serves a purpose, but we are not taking any advantage of it at the moment.

fortmarek: Yep, exactly. It's serving the *syntactic* purpose Boundary asks for (every module declared once) without serving the *actual* purpose (enforcing architectural layers). I'll open a separate issue so we have a real tracker for doing the Boundary-gets-teeth work — feels like its own design conversation, not something to staple onto this PR.

fortmarek: Tracker: #10431.

PR #10368, opened 2026-04-22
feat(server): migrate from Render to Syself Kubernetes (Helm chart, Alloy, ESO, cascade deploys)

fortmarek:
## Summary

Moves the Tuist server from Render to a Syself-managed Kubernetes cluster on Hetzner, with all three environments (staging, canary, production) on the same shape.

## Current state

- **Staging** — live. `/ready` returns 200 via `https://staging.tuist.dev` (Cloudflare edge → Hetzner LB `167.235.217.233` → Syself). Cutover complete.
- **Canary** — deployed. Hetzner LB: `91.98.11.250`. `/ready` returns 200 when addressed with the `canary.tuist.dev` host header. DNS flip to the canary LB is still pending (will be done post-merge).
- **Production** — cluster up, ESO + 1P ClusterSecretStore + Alloy installed, CI kubeconfig loaded. **First deploy is blocked on Hetzner's Load Balancer quota** (we hit 5/5 — 3 auto-created apiserver LBs + staging Service LB + canary Service LB). Support ticket filed to raise the limit. Once raised, a re-dispatch should land production cleanly; no further code changes required.

### What's in the PR

**Cluster side** — reproducible Syself onboarding:
- [`infra/k8s/syself-onboarding.md`](infra/k8s/syself-onboarding.md) — end-to-end bring-up runbook (ClusterStack subscription → Hetzner secrets → workload cluster CR → ESO/ClusterSecretStore → Alloy → CI ServiceAccount).
- [`infra/k8s/syself/`](infra/k8s/syself/) — Cluster CRs for all three envs ([`staging`](infra/k8s/syself/workload-cluster-staging.yaml), [`canary`](infra/k8s/syself/workload-cluster-canary.yaml), [`production`](infra/k8s/syself/workload-cluster-production.yaml)), plus [`cluster-stack.yaml`](infra/k8s/syself/cluster-stack.yaml) (ClusterStack subscription), [`ci-service-account.yaml`](infra/k8s/syself/ci-service-account.yaml), [`ingress-nginx-values.yaml`](infra/k8s/syself/ingress-nginx-values.yaml).

**Helm chart** — managed-cloud shape on top of the existing self-host chart:
- [`values-managed-common.yaml`](infra/helm/tuist/values-managed-common.yaml) + per-env overlays (`values-managed-{staging,canary,production}.yaml`) — external DBs (Supabase + ClickHouse Cloud + Tigris), ESO-driven `MASTER_KEY`, Helm pre-install migration hook, probes, labels, autoscaling/PDB config.
- Chart templates extended with `Service.metadata.annotations` + `externalTrafficPolicy` so the Hetzner CCM can provision LoadBalancers with the right location + tier.
- ESO `ClusterSecretStore` + `ExternalSecret` glue for 1Password-sourced secrets.
- Everything is gated behind `server.managedSecrets` (default `false`) so self-hosters see no behavior change.

**Observability** — [`infra/helm/alloy/`](infra/helm/alloy/): in-cluster Grafana Alloy that scrapes the server's `/metrics`, receives OTLP traces via gRPC, tails pod logs through the Kubernetes API, and forwards everything to Grafana Cloud (Prometheus / Tempo / Loki). Replaces the external `tuist-fmpt` Alloy pserv. Self-hosters continue to use the chart's embedded observability stack (unchanged).

**CI** — two workflows with single responsibilities:
- [`server-deployment.yml`](.github/workflows/server-deployment.yml) — build + deploy to ONE chosen environment. Triggered by `workflow_dispatch` (for ad-hoc deploys) or called from the cascade via `workflow_call`. No cascade, no tests.
- [`server-production-deployment.yml`](.github/workflows/server-production-deployment.yml) — promotion cascade on `push to main`: canary → acceptance tests → production. Commits with `(hotfix)` in the subject fast-path to production; canary + tests still run as a safety net. `workflow_dispatch` runs the cascade against a pinned SHA.
- Replaces the three Render-era `server-{staging,canary,production}-deployment.yml` workflows (deleted). No `gcloud`, no Cloud Build config, no `values-gke.yaml`.
- **Slack notifications** reuse the existing [`notify-failure.yml`](.github/workflows/notify-failure.yml) pattern (listens for `workflow_run` completions, pulls the webhook from the `cache` 1Password vault via `CACHE_OP_SERVICE_ACCOUNT_TOKEN`). Added "Server Deployment" to its watch-list so direct dispatches are covered. A new [`notify-deploy-success.yml`](.github/workflows/notify-deploy-success.yml) mirrors it for successes, scoped to "Server Production Deployment" only — staging/canary successes stay silent by design. No new secrets required.

**Runtime cleanup** — `server/rel/env.sh.eex` and `server/config/runtime.exs` drop the Render env-var fallbacks (`RENDER_DISCOVERY_SERVICE`, `RENDER_SERVICE_NAME`, `RENDER_INTERNAL_IP`, `RENDER_GIT_COMMIT`); only `TUIST_CLUSTER_*` + `TUIST_GIT_COMMIT` feed the release now.

## Per-env cluster shapes

| Env | Control plane | Workers | Why | Cost/mo |
|---|---|---|---|---|
| staging | 3 × cpx22 | 2 × cpx22 (shared vCPU) | Low-traffic pre-prod; cheapest viable. One cpx22 saturates on system pods (cilium-operator, hubble, CSI controllers, metrics-server, ESO, Alloy) before the app schedules, so two is the minimum. | **~€45** |
| canary | 3 × cpx22 | 2 × cpx22 (shared vCPU) | Mirrors staging; canary gates production but doesn't serve meaningful real traffic. | **~€45** |
| production | 3 × cpx22 | 2 × ccx23 (dedicated vCPU, 4 vCPU / 16 GB / 160 GB NVMe) | Dedicated cores eliminate CPU noisy-neighbor jitter on the one env where p99 matters; 32 GB total RAM is plenty for the Phoenix BEAM + system pods; API-driven so HPA still works for traffic spikes. | **~€87** |

All-cloud **~€175/mo total** across three envs.

## Why not bare metal?

We considered bare metal (Hetzner Robot AX41-NVMe class, ~€50/mo per box) for production and deliberately decided against it for this migration. Bare metal's wins — **dedicated local NVMe at ~1M IOPS**, **massive RAM cheap per GB**, **zero CPU noisy-neighbor** — all assume the node runs workloads that can consume them. Our production workload doesn't:

1. **We run no stateful services on the cluster.** Postgres (Supabase), ClickHouse (ClickHouse Cloud), and object storage (Tigris) are all external managed offerings. Phoenix writes nothing to disk beyond logs — Alloy tails stdout, so local disk IOPS never matters.
2. **Phoenix latency is I/O-bound on those external databases**, not CPU-bound. A Tuist request spends most of its wall-clock time waiting on Supabase / ClickHouse round trips. Local CPU jitter is in the noise.
3. **Autoscaling matters more than peak-perf hardware for our traffic shape.** Traffic spikes at release time (our own releases and downstream teams'); HPA on cloud nodes absorbs those in seconds. Bare metal can't autoscale — you order boxes ahead of need and pay for peak forever.
4. **Bare metal provisioning is manual + slow.** Order a server on Hetzner Robot → wait 10 min to a few hours for delivery → SSH bootstrap → register `HetznerBareMetalHost` → scale the MachineDeployment. Scaling is measured in hours. We want minutes.

**When does bare metal become the right call?** When we bring Postgres and ClickHouse into the cluster. An AX41-NVMe with local NVMe at ~1M IOPS runs Postgres / ClickHouse orders of magnitude better than any Hetzner Cloud block-storage disk (capped ~3k IOPS on ccx, shared on cpx), at a fraction of the cost of managed providers. That's the migration where bare metal pays off.

## GKE was evaluated, then dropped

We stood up a parallel GKE POC and validated the same Helm chart + ESO flow against it. Worked fine, just more expensive (~2–3×) and a different vendor than our Hetzner-based cache fleet. The GKE-specific bring-up doc (`gke-onboarding.md`), `values-gke.yaml`, and `server/cloudbuild.yaml` are removed here.

## Cutover plan after merge

1. **Staging is already cut over.** Merging this PR only changes the deploy pipeline; staging has been serving traffic on Syself for hours via the existing workflow runs + the Cloudflare A record flip.

2. **Unblock production**: wait for Hetzner to raise our Load Balancer quota (support ticket filed). Once raised, re-dispatch `Server Deployment` with `environment=production` and the remaining Service LB provisions; production is then reachable at its Hetzner LB IP, with `/ready` returning 200.

3. **Stop deploying to Render.** The old `server-{staging,canary,production}-deployment.yml` workflows are deleted in this PR, so the next push to `main` after merge goes to the new Kubernetes cascade — **not** to Render. Render stays up passively until we actively delete the services (step 5).

4. **Canary DNS cutover.** Flip `canary.tuist.dev` CNAME → Hetzner canary LB IP (`91.98.11.250`) in Cloudflare. Low traffic so no gradual ramp needed.

5. **Gradual production cutover via Cloudflare Load Balancing.** `tuist.dev` stays on its current Render A record while we add the Syself LB IP as a second origin in a Cloudflare Load Balancer pool. Ramp weights ~5% → 25% → 50% → 100% Syself over a week, watching Grafana Cloud SLO dashboards + Sentry. Rollback = set weight back to 0% (30s edge propagation). Once 100% and stable for a few days, we delete the Render record, delete the Render services, and remove `origin.staging` / any other Render-era DNS.

6. **Delete the old Render-era GitHub Environments.** The `server-canary`, `server-staging`, and `server-production` Environments were used by the now-deleted Render workflows and nothing references them anymore. Delete them via GitHub UI → Settings → Environments → click each one → Delete. Not urgent; they're cost-free to leave around, but cleaner gone.

7. **End-to-end TLS as a follow-up.** All three envs currently sit behind Cloudflare's edge proxy (TLS terminates at Cloudflare, plain HTTP to the Hetzner LB). Good enough shipping, but a separate PR to install cert-manager + ingress-nginx + external-dns in each cluster closes the edge-to-origin hop.

## Where to watch deployments

- **GitHub Actions** (primary) — <https://github.com/tuist/tuist/actions/workflows/server-production-deployment.yml> for the push-to-main cascade, <https://github.com/tuist/tuist/actions/workflows/server-deployment.yml> for ad-hoc single-env dispatches. Each cascade run shows canary → acceptance tests → production.
- **Lens / k9s** against the workload kubeconfigs — `Deployments` / `ReplicaSets` / `Pods` in the env's namespace (`tuist-staging` / `tuist-canary` / `tuist`). `helm history` surfaces each release + rollback target.
- **Slack** — failure pings for any listed deploy workflow on main, plus a success ping when production lands green. Webhook comes from the `cache` 1P vault.
- **Grafana Cloud** — deploy events show up indirectly via pod restart spikes + build-info metrics. A dedicated Deployments dashboard is a planned follow-up.

## Test plan

- [x] Staging cluster up, `/ready` returns 200 via `https://staging.tuist.dev`
- [x] Canary cluster up, `/ready` returns 200 via the Hetzner LB IP (`91.98.11.250`)
- [x] Production cluster up; ESO + Alloy + CI kubeconfig wired
- [x] `server-deployment.yml` succeeds end-to-end on staging + canary (plan ✓ build+push ✓ deploy ✓)
- [ ] Production deploy succeeds — blocked on Hetzner LoadBalancer quota bump
- [ ] First merge to main runs the cascade cleanly
- [ ] DNS cutover for `canary.tuist.dev` and the Cloudflare LB ramp for `tuist.dev`


infra/helm/alloy/templates/externalsecret.yaml
fortmarek, line 1: down the line, all secrets should be pulled in from here and we should stop using the master key, so it's easier to rotate individual secrets

infra/helm/tuist/templates/server-deployment.yaml
```diff
@@ -64,16 +89,38 @@ spec:
...
             {{- if .Values.server.license.certificateBase64 }}
             - name: TUIST_LICENSE_CERTIFICATE_BASE64
               valueFrom:
                 secretKeyRef:
-                  name: {{ include "tuist.componentName" (dict "root" . "component" "app-secrets") }}
+                  name: {{ $secretName }}
                   key: server-license-certificate-base64
             {{- end }}
+            {{- if or .Values.server.masterKey .Values.server.managedSecrets }}
+            - name: MASTER_KEY
```
fortmarek, line 103: as mentioned above, we should aim to remove the usage of the master key in favor of having all secrets from 1p.

the environment-specific variables that are not necessarily secret can then be encoded directly in the helm chart.

infra/helm/tuist/values-managed-production.yaml
```diff
@@ -0,0 +1,78 @@
...
+      value: http://processor.tuist.dev:4002
+    - name: TUIST_OTEL_EXPORTER_OTLP_ENDPOINT
+      value: http://alloy-tuist-alloy.observability.svc.cluster.local:4317
+
+  autoscaling:
+    enabled: true
+    minReplicas: 2
+    maxReplicas: 5
+    targetCPUUtilizationPercentage: 70
+    targetMemoryUtilizationPercentage: 80
```
pepicrft, line 35: Production scales to 2+ replicas here, but `values-managed-common.yaml` still leaves `server.cluster.enabled=false`, so the chart never renders the headless Service or `TUIST_CLUSTER_*` env vars. Given the app uses PubSub-backed behavior like live updates and the in-memory rate limiter, is it intentional for each pod to run isolated once HPA scales out?

infra/helm/alloy/values.yaml
```diff
@@ -0,0 +1,98 @@
...
+# Kubernetes pod log tailing. Alloy uses the discovery.kubernetes component
+# to find pods, then loki.source.kubernetes to stream their stdout logs via
+# the kubelet API and forward them to Grafana Cloud Loki. Apps just write to
+# stdout — no in-app Loki client required.
+podDiscovery:
+  # Scope to specific namespaces so Alloy only tails the Tuist workloads
+  # (not kube-system, ESO, etc.). Empty list = all namespaces.
+  namespaces:
+    - tuist-staging
+    - tuist  # placeholder for production/canary
```
pepicrft: This chart reads as reusable per workload cluster, but the checked-in defaults still bake in staging-specific values: the Prometheus target points at `tuist-staging`, pod discovery includes `tuist-staging`, and the config template hard-codes `env=staging`. If the intended model is `generic chart + per-environment values`, should those overrides live in the repo too? As written, canary/production installs still seem to depend on out-of-repo values.

pepicrft (commented):
Left two follow-up comments around clustering and the Alloy chart contract.

pepicrft (approved):
Nice work on the migration. I left two follow-up comments around clustering and the Alloy chart setup, but overall this looks good to me.

fortmarek:
Addressed both in cf46eeaec4:

**1. Production cluster.enabled** — you're right, leaving it off with replicaCount≥2 would have silently broken PubSub fan-out and the in-memory rate limiter. Enabled on the production overlay (canary stays at replicaCount=1 so it's unaffected, but happy to flip it too if we want symmetry):

```yaml
# values-managed-production.yaml
server:
  cluster:
    enabled: true
    applicationName: tuist
```

The chart's existing `server-service-headless.yaml` template + `TUIST_CLUSTER_*` env wiring pick this up automatically.

**2. Alloy per-env values in the repo** — extracted. The shared `values.yaml` is now env-agnostic (`env`, `scrape.targets`, `podDiscovery.namespaces` all default to empty; the `env` label is a required value, template fails loudly if unset). Three overlays checked in:

- `infra/helm/alloy/values-staging.yaml`
- `infra/helm/alloy/values-canary.yaml`
- `infra/helm/alloy/values-production.yaml`

Each points Prometheus scrape at its own cluster's server Service and scopes pod discovery to the single workload namespace (`tuist-staging` / `tuist-canary` / `tuist`). The three running Alloy releases have been re-applied with their respective overlay.

PR #10371, opened 2026-04-22
fix(server): restore missing vitepress docs path redirects

pepicrft:
## Summary
- restore the missing VitePress-era docs path redirects in Phoenix
- keep the docs redirect plug focused on path normalization and moved-page redirects only
- remove the Phoenix-side host redirect handling and Render domain change for `docs.tuist.dev`

## Why
The missing behavior was the path-level redirect map that existed in VitePress for moved docs pages like `guides/features/insights`, `guides/develop/build/cache`, and related reorganizations. Host-level redirects for `docs.tuist.dev` are already handled elsewhere, so Phoenix only needs to preserve the page-path redirects that were lost in the migration.

## Impact
- old VitePress page paths now resolve to the current docs destinations again
- renamed pages like `guides/features/insights` redirect to `guides/features/build-insights`
- chained legacy paths like `guides/develop/build/cache` collapse to the live `guides/features/cache` page
- external ProjectDescription reference redirects continue to work

## Validation
- added resolver and plug coverage for renamed, chained, and external docs redirects
- smoke-checked the Phoenix redirect resolver for `insights`, `guides/develop/build/cache`, and ProjectDescription redirects
- smoke-checked the plug for `/docs/en/...` normalization plus chained redirects with query preservation
- `mix compile` passes
- narrow `mix test` runs are still blocked locally by unrelated environment issues in this worktree, including a missing test `users` table and one endpoint port collision during startup


server/lib/tuist_web/live/docs_overview_live.ex
```diff
@@ -201,7 +201,7 @@ defmodule TuistWeb.DocsOverviewLive do
               </div>
             </.link>
             <.link
-              patch={docs_path("/en/guides/features/builds/insights")}
+              patch={docs_path("/en/guides/features/build-insights")}
```
pepicrft, line 204: this link was wrong

fortmarek (approved):

PR #10373, opened 2026-04-22
feat(server): move filters to page level for tests related pages

asmitbm:
Move filters to page level

Modified pages:

- Tests
- Test Runs
- Test Cases
- Flaky Tests
- Quarantined Tests
- Shards


server/lib/tuist_web/live/test_runs_live.ex
```diff
@@ -366,11 +366,15 @@ defmodule TuistWeb.TestRunsLive do
...
-    ]
+    {start_datetime, end_datetime} = socket.assigns.analytics_period
+
+    flop_filters =
+      [
+        %{field: :project_id, op: :==, value: project.id},
+        %{field: :status, op: :!=, value: "in_progress"},
+        %{field: :ran_at, op: :>=, value: start_datetime},
+        %{field: :ran_at, op: :<=, value: end_datetime}
+      ] ++ build_flop_filters(filters, search) ++ page_level_environment_filters(socket.assigns)
```
pepicrft, line 382: These new page-level filters now change the dataset behind this paginated query, but `analytics_period_changed` still preserves any existing `before` or `after` cursor. If the user is on a later page and then changes the environment or date range, we reuse a cursor from the previous result set and can skip the first page of filtered results or show an empty table. Can we clear the cursors when those top-level filters change?

server/lib/tuist_web/live/shards_live.ex
```diff
@@ -338,8 +338,12 @@ defmodule TuistWeb.ShardsLive do
...
       Filter.Operations.decode_filters_from_query(params, socket.assigns.available_filters)
 
+    {start_datetime, end_datetime} = socket.assigns.analytics_period
+
     flop_filters = [
-      %{field: :project_id, op: :==, value: project.id}
+      %{field: :project_id, op: :==, value: project.id},
+      %{field: :ran_at, op: :>=, value: start_datetime},
+      %{field: :ran_at, op: :<=, value: end_datetime}
       | build_sharded_run_flop_filters(filters, search)
```
pepicrft, line 352: Adding the date range to this paginated query means `analytics_period_changed` now changes the result set without clearing any existing `before` or `after` cursor. After paginating, changing the range can reuse a cursor from the previous dataset and jump into the middle of the filtered rows or come back empty even when matches exist. Can we drop the cursors when the top-level date filter changes?

server/lib/tuist_web/live/test_runs_live.ex
```diff
@@ -366,11 +366,15 @@ defmodule TuistWeb.TestRunsLive do
...
-      | build_flop_filters(filters, search)
-    ]
+    {start_datetime, end_datetime} = socket.assigns.analytics_period
+
+    flop_filters =
+      [
+        %{field: :project_id, op: :==, value: project.id},
+        %{field: :status, op: :!=, value: "in_progress"},
+        %{field: :ran_at, op: :>=, value: start_datetime},
+        %{field: :ran_at, op: :<=, value: end_datetime}
```
pepicrft, line 381: I tried this locally and the table still does not appear to honor the selected page-level filters. With a custom date range of March 10, 2026 to March 13, 2026, the analytics updates but the list still shows runs from now like `in 2 minutes`, which should be outside the selected window. So it looks like the list filtering is still not working end to end.

pepicrft: @asmitbm I'd go page by page, checking if the filters are working.

pepicrft:
@asmitbm some behaviors that I noticed:

- In the test runs page, the period filter doesn't apply to the list of test runs. I'd expect that to be applied:
image

- The same happens in test cases
image

- The same happens in quarantined tests. Additionally the font of last run seems wrong:
image

- The shards page suffers from the same thing:
image

- Flaky tests is the only page that works fine.

- I noticed some filters duplication in the "Tests" page. For example, on this page, I can see _environment_ and _period_ in the first and second cards. Is that by design?

image

asmitbm:
I applied the same filter for test runs page, but the filter works for me

Screenshot 2026-04-22 at 16 14 59

asmitbm:
> The same happens in quarantined tests. Additionally the font of last run seems wrong:

Can you hit Cmd+Shift+R to fix the font?

For this page, the date filter shouldn’t apply to the table. Quarantined tests represent the current state, we want to see what’s quarantined right now, regardless of when they last ran. The date range is only relevant for the analytics chart, where it shows trends over time.

asmitbm:
> The shards page suffers from the same thing:

Shards page does work fine for me

Screenshot 2026-04-22 at 16 55 18

asmitbm:
> I noticed some filters duplication in the "Tests" page. For example, on this page, I can see environment and period in the first and second cards. Is that by design?

There are no filters in other cards. Maybe try Cmd+Shift+R? I can see that page level filters are not there and instead they are section card filters

Screenshot 2026-04-22 at 16 59 47

pepicrft:
My bad, I was running it from the wrong copy of the repo. A few things that I noticed. Something is off with the flaky tests:

image


Same with quarantined tests:

image


Everything else is working fine.

pepicrft (approved):

asmitbm:
> Odd, I still get the same issue with flaky tests: 

Haven't pushed the changes yet, testing it right now

PR #10375, opened 2026-04-22
feat(server): page level filter for xcode cache page

asmitbm:
Page level filters for xcode cache page


server/lib/tuist_web/live/xcode_cache_live.ex
```diff
@@ -215,10 +220,14 @@ defmodule TuistWeb.XcodeCacheLive do
...
 
   defp assign_recent_builds(%{assigns: %{selected_project: project}} = socket, _params) do
+    {start_datetime, end_datetime} = socket.assigns.analytics_period
+
     options = %{
       filters: [
         %{field: :project_id, op: :==, value: project.id},
-        %{field: :cacheable_tasks_count, op: :>, value: 0}
+        %{field: :cacheable_tasks_count, op: :>, value: 0},
+        %{field: :inserted_at, op: :>=, value: start_datetime},
```
pepicrft, line 224: Now that `@builds` is filtered by `start_datetime` and `end_datetime`, the top-level `Enum.empty?(@builds)` check will show the setup empty state for projects that already have cache data whenever the selected window has no matches. That also hides the period filter, which makes the state hard to recover from. We should keep a separate unfiltered `has_builds?` check, or always render the page shell and let only the Recent Builds card go empty.

server/lib/tuist_web/live/gradle_cache_live.ex
```diff
@@ -241,7 +246,21 @@ defmodule TuistWeb.GradleCacheLive do
   defp analytics_trend_label(_), do: dgettext("dashboard_gradle", "since last month")
 
   defp assign_recent_builds(%{assigns: %{selected_project: project, selected_account: account}} = socket, _params) do
-    {builds, _meta} = Gradle.list_builds(project.id, %{page_size: @recent_builds_page_size})
+    {start_datetime, end_datetime} = socket.assigns.analytics_period
+    analytics_environment = socket.assigns.analytics_environment
+
+    filters =
+      [
+        %{field: :inserted_at, op: :>=, value: start_datetime},
```
pepicrft, line 249: Now that `@builds` is filtered by the selected period and environment, the top-level `Enum.empty?(@builds)` check will show the setup empty state for projects that already have Gradle cache data whenever the chosen filters return no matches. That also hides the filters needed to recover. We should keep a separate unfiltered `has_builds?` check, or always render the page shell and let only the Recent Builds card go empty.

server/lib/tuist_web/live/xcode_cache_live.ex
```diff
@@ -215,10 +220,14 @@ defmodule TuistWeb.XcodeCacheLive do
...
+        %{field: :inserted_at, op: :<=, value: end_datetime}
```
pepicrft, line 225: The cache pages use `inserted_at <= end_datetime`, but the existing build pages use `DateTime.add(end_datetime, 1, :second)` for the same date picker values. Because custom end dates are truncated to `23:59:59`, the current filter can drop builds created at `23:59:59.xxx` on the selected end day, which makes the Recent Builds list look out of sync with the chosen period.

server/lib/tuist_web/live/gradle_cache_live.ex
```diff
@@ -241,7 +246,21 @@ defmodule TuistWeb.GradleCacheLive do
...
+        %{field: :inserted_at, op: :<=, value: end_datetime}
```
pepicrft, line 250: The cache pages use `inserted_at <= end_datetime`, but the existing build pages use `DateTime.add(end_datetime, 1, :second)` for the same date picker values. Because custom end dates are truncated to `23:59:59`, the current filter can drop builds created at `23:59:59.xxx` on the selected end day, which makes the Recent Builds list look out of sync with the chosen period.

pepicrft:
@asmitbm is recent builds supposed to inherit the date period filter? If so, it doesn't work for me:
image

asmitbm:
> @asmitbm is recent builds supposed to inherit the date period filter? If so, it doesn't work for me:

Can you check now? I pushed a commit while you were reviewing
Screenshot 2026-04-22 at 16 09 51

pepicrft:
> Can you check now? I pushed a commit while you were reviewing

Beautiful! It works now. Let's fix the formatting issue

pepicrft (approved):

PR #10376, opened 2026-04-22
feat(server): page level filters for bundles

asmitbm:
Page level filters for bundles page

## Summary

- Align bundle widgets, chart, and table so they all apply the same filter set (project + period + app + type + branch) — previously each component enforced a different subset, which could leave the chart showing data from other apps, the table showing other branches, or the widgets showing a stale "latest ever" bundle regardless of the selected date range.
- Make the branch selector app-aware: on load (or when switching apps), the dropdown defaults to `main` if the selected app has default-branch bundles, otherwise to `Any`, so users don't land on an empty dashboard when their app happens to live on a feature branch.
- Strict date/branch filtering end-to-end: if no bundles exist for the current filters, all three views report "no data" together instead of one view quietly falling back to older or unrelated bundles.

## What changed

- `Tuist.Bundles.last_project_bundle/2`: new `period: {from, to}` option so the widget's "latest" lookup is scoped to the selected date range.
- `Tuist.Bundles.project_bundles_by_date/2` (chart source): now filters by `start_datetime..end_datetime` and `:name`, removing the cross-range fallback that could pull in bundles from outside the selected period.
- `Tuist.Bundles.has_bundles_in_project_default_branch?/2`: accepts an optional `:name` filter, used to decide per-app whether `main` is a sensible default.
- `TuistWeb.BundlesLive`: computes `effective_git_branch` once and threads it through the widget, chart, and table. `bundle_size_branch` auto-switches to `Any` when the selected app has no default-branch bundles (covers both first load and stale URL params). Adds `maybe_add_branch_filter/2` to the table's Flop filters.


pepicrft:
I changed the period to one that doesn't include any bundle, and the widgets still showed data:
image

I also think we should adjust the "... yet" copy. Maybe something like "... for the selected period"?
image

pepicrft (approved):

asmitbm:
> I also think we should adjust the "... yet" copy. Maybe something like "... for the selected period"?

I think that's a common empty state shared across different states like when there's actually no bundle data at all. I will check if we can change the copy specifically when the filter is applied.

pepicrft:
> I will check if we can change the copy specifically when the filter is applied.

I like the idea. "Yet" has the connotation of "data hasn't arrived", but in the case of being filtered, that's not really the case.

asmitbm:
> I changed the period to one that doesn't include any bundle, and the widgets still showed data:

@pepicrft I applied the same filters, but got this:
Screenshot 2026-04-23 at 14 58 45

fortmarek (approved):

PR #10380, opened 2026-04-22
feat(server): use feature flags for Kura cache endpoint selection

pepicrft:
## Summary
- Replace the cache endpoint `technology` query parameter with the shared `x-tuist-feature-flags` header and resolve Kura when the `kura` flag is present
- Update the CLI and Gradle cache clients to use the existing feature-flag plumbing instead of `TUIST_KURA`
- Keep cache endpoint persistence technology-aware and add a compatibility migration to rename `cache_technology` to `technology` for databases that already applied the earlier migration state

## Testing
- `mise x -- ./gradlew test --tests dev.tuist.gradle.CacheEndpointResolverTest --tests dev.tuist.gradle.TuistHttpClientsTest --tests dev.tuist.gradle.TuistHttpClientTest`
- `mise run lint`
- `swift test --replace-scm-with-registry --filter CacheURLStoreTests`
- `mise x -- env MIX_ENV=test mix ecto.migrate`
- `mise x -- mix test test/tuist_web/headers_test.exs test/tuist_web/controllers/api/cache_controller_test.exs`


fortmarek (commented):
Instead of a client selecting the technology, should this instead be a server-side configuration (since the server already lists available endpoints and knows the account associated)?

I'm not sure how much of a fan I am of making this configurable from the client, too. At least it's not what I'd start with.

pepicrft:
> I'm not sure how much of a fan I am of making this configurable from the client, too. At least it's not what I'd start with.

It really depends on how we'd like accounts that are not `tuist` to opt into it. I'm inclined to suggest that they run it in a `branch` before, which means they need client-side control @fortmarek. If you disagree with that, I can do server-side driven, but I'm not sure I like that approach for the rollout.

server/lib/tuist_web/controllers/api/cache_controller.ex
```diff
@@ -13,40 +13,52 @@ defmodule TuistWeb.API.CacheController do
...
+         required: false,
+         description: "The name of the account to get custom cache endpoints for."
+       ]},
+      {:"x-tuist-feature-flags",
+       [
+         in: :header,
+         type: :string,
+         required: false,
+         description: "Comma-separated client feature flags. Include `kura` to resolve Kura cache endpoints."
+       ]}
```
fortmarek: not sure if this is something that needs to be part of the openapi interface? I'd probably lean skipping including feature flags in the openapi definition, but don't feel too strongly about it.

server/lib/tuist_web/controllers/api/cache_controller.ex
```diff
@@ -13,40 +13,52 @@ defmodule TuistWeb.API.CacheController do
...
+      cache_ttl: to_timeout(minute: 1)
+    ]
+    when action not in [:endpoints]
+  )
 
-  plug TuistWeb.API.Authorization.BillingPlug when action not in [:endpoints]
+  plug(TuistWeb.API.Authorization.BillingPlug when action not in [:endpoints])
 
-  plug :sign
+  plug(:sign)
```
fortmarek: a bit noisy, would be great if the formatter would keep the presence or omission of brackets always aligned.

server/data-export.md
```diff
@@ -18,7 +18,7 @@ Sensitive authentication data (passwords, tokens) are excluded from exports.
 - Organization memberships and roles
 - Account billing information and subscriptions
 - API tokens and project tokens (existence only, not values)
-- Custom cache endpoint configurations
+- Custom cache endpoint configurations, including technology-specific endpoint groups (for example, default and Kura)
```
fortmarek: don't see how this is relevant to data export and our customers

fortmarek (approved):

server/lib/tuist_web/controllers/api/cache_controller.ex
pepicrft: I'll investigate this one separately.

server/lib/tuist_web/controllers/api/cache_controller.ex
pepicrft: I went back and forth, but I think I'll drop it, since the consumers of the API (as a product) is not something they are interested in knowing.

PR #10382, opened 2026-04-22
feat(cli,server,gradle): add feature flag headers

pepicrft:
## Summary
- Add feature-flag header propagation for the CLI and Gradle plugin
- Send enabled flags as a comma-separated list in `x-tuist-feature-flags`
- Add server helpers to read the header from a `conn` as feature presence
- Cover the new transport format with unit tests across clients and server

## Testing
- `gradle` plugin unit tests passed
- Swift `xcodebuild build -workspace Tuist.xcworkspace -scheme tuist ...` passed
- Android unit tests were added, but local execution is still blocked here by the missing Android SDK


cli/Sources/TuistServer/ClientFeatureFlags.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+import TuistEnvironment
+
+public enum ClientFeatureFlags {
+    public static let headerName = "x-tuist-feature-flags"
+    private static let environmentPrefix = "TUIST_FEATURE_"
```
fortmarek:
```suggestion
    private static let environmentPrefix = "TUIST_FEATURE_FLAG_"
```

Would be maybe a bit more specific about the name to lessen the potential for user-defined env variables

android/app/src/main/java/dev/tuist/app/data/network/FeatureFlagsHeaders.kt
fortmarek, line 1: I'd skip adding the feature flag functionality for android probably

server/lib/tuist_web/headers.ex
```diff
@@ -26,4 +28,72 @@ defmodule TuistWeb.Headers do
   def put_cli_version(conn, version) do
     Plug.Conn.put_req_header(conn, "x-tuist-cli-version", version)
   end
+
+  def get_client_feature_flags(conn) do
```
fortmarek, line 32: how is this used? should we parse the feature flags into the conn context eagerly? This looks like it will be used on demand, which I'm not sure is what we want.

fortmarek (approved):

server/lib/tuist_web/headers.ex
pepicrft: I think on-demand parsing is fine for now. The header is small, parsing is cheap, and we don’t currently have multiple consumers in the same request. If we start reading it in more places, we can move it into conn context via a plug later.

PR #10386, opened 2026-04-22
chore(server): drop legacy flaky columns replaced by the automations engine

fortmarek:
## Summary

Follow-up to #10232. Drops the legacy columns and dead code paths left behind by the switch to the Automations engine, now that the backfill migration in #10232 has copied every project's flaky settings into automation alerts.

**PostgreSQL \`projects\` columns removed**
- \`auto_mark_flaky_tests\`, \`auto_mark_flaky_threshold\`
- \`auto_quarantine_flaky_tests\`
- \`flaky_cooldown_days\`
- \`flaky_test_alerts_enabled\`
- \`flaky_test_alerts_slack_channel_id\`, \`flaky_test_alerts_slack_channel_name\`

**ClickHouse \`test_cases\` column removed**
- \`is_quarantined\` (replaced by \`state\`; backfill already flipped \`is_quarantined = true\` → \`state = 'muted'\` in #10232's \`20260410120000_add_state_to_test_cases\`)

**Code paths removed**
- \`Tuist.Slack.send_flaky_test_alert/4\` + its helper blocks — superseded by \`SendSlackAction\` in the automation engine.
- \`Project\` schema fields, \`create_changeset\` / \`update_changeset\` casts, threshold/cooldown validators.
- The two legacy Slack channel fields from \`Tuist.Slack.clear_project_slack_fields_query/1\`.
- Unused \`maybe_add_suite/2\` helper.
- Seed data for the removed fields.
- \`slack_test.exs\` describe block for \`send_flaky_test_alert/4\`.

## Merge ordering

**Must merge after #10232.** That PR contains the backfill migration (\`20260411120000_backfill_automations_from_project_settings\`) which reads the columns this PR drops. If this lands first, the backfill breaks. The migration timestamps enforce the correct runtime order, but the source code dependency (the backfill's query is typed against the columns) still requires the PR order.

## Test plan

- [x] \`mix test test/tuist/slack_test.exs test/tuist/projects_test.exs test/tuist/automations_test.exs test/tuist_web/live/project_automations_live_test.exs test/tuist_web/controllers/api/projects_controller_test.exs\` — 164 passing locally
- [x] \`mix compile --warnings-as-errors\` clean
- [x] \`mix credo\` clean on touched files
- [x] Ran \`ecto.drop/create/load/migrate\` locally to confirm the drop migrations apply end-to-end
- [ ] CI green before merging, and merged only after #10232


server/lib/tuist_web/live/project_automations_live.ex
```diff
@@ -63,95 +120,356 @@ defmodule TuistWeb.ProjectAutomationsLive do
...
+        Map.put(action, "message", message)
+      end)
 
-          {:error, _changeset} ->
-            {:noreply, socket}
-        end
+    {:noreply, assign(socket, create_automation_form_recovery_actions: actions)}
+  end
+
+  def handle_event("save_automation", _params, %{assigns: assigns} = socket) do
```
pepicrft, line 305: This LiveView now mounts with `automation_alert_read`, but these mutation handlers never enforce the matching `automation_alert_create/update/delete` checks. That means any non-admin project user who can open the page can still trigger create/edit/delete events over the LiveView socket, bypassing the stricter REST controller policy. Could we gate these handlers the same way the controller does, or keep `mount/3` admin-only if the page is meant to be mutable?

server/lib/tuist/tests.ex
```diff
@@ -702,9 +689,9 @@ defmodule Tuist.Tests do
       end
 
     events =
-      case {Map.get(old_test_case, :is_quarantined, false), Map.get(new_attrs, :is_quarantined)} do
-        {false, true} -> [:quarantined | events]
-        {true, false} -> [:unquarantined | events]
+      case {Map.get(old_test_case, :state, "enabled"), Map.get(new_attrs, :state)} do
```
pepicrft, line 692: Renaming these emitted events to `muted` / `unmuted` looks like a breaking contract change for the rest of the stack. Later quarantine-attribution queries in this module, `Tuist.Tests.Analytics`, and the public events enum in `API.TestCasesController` still key off `quarantined` / `unquarantined`, so newly muted tests will stop showing up correctly in attribution, charts, and the events API unless those consumers migrate in the same PR.

server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
```diff
@@ -0,0 +1,101 @@
...
+    cutoff = DateTime.add(DateTime.utc_now(), -window, :second)
+
+    triggered_test_case_ids =
+      ClickHouseRepo.all(
+        from(tcr in TestCaseRun,
+          where: tcr.project_id == ^project_id,
+          where: tcr.inserted_at >= ^cutoff,
+          where: tcr.is_flaky == true,
+          group_by: tcr.test_case_id,
+          having: count(tcr.id) >= ^threshold,
```
pepicrft, line 60: This changes the semantics of the migrated threshold. The legacy worker counted `count(DISTINCT (scheme, git_commit_sha))` groups, but this monitor counts every flaky `test_case_run`. Since the backfill copies existing thresholds directly into the new automation, projects with retries or repeated flaky runs on the same commit/scheme will start triggering much earlier than before. Is that behavior change intentional, or should this preserve the old distinct-group counting so the migration stays behaviorally compatible?

PR #10388, opened 2026-04-22
feat(kura): add Kura to the monorepo

pepicrft:
Kura is a technology to build tenant-scoped cache clusters with bounded resources. I'm moving it to the monorepo for fast iteration. The interface from the Tuist client and to the Tuist server mimics the one of our current cache server, but the implementation is done in Rust and designed for building a decentralized mesh.

## Summary
- Import Kura into the repository as a top-level workspace under `kura/`
- Add monorepo intent docs and root README links for the new service
- Wire Kura into GitHub Actions failure notifications and a dedicated workflow

## Testing
- Unit tests passed in `kura/`
- End-to-end ShellSpec suite passed in `kura/`


kura/ops/helm/kura/values-scaleway.yaml
fortmarek, line 1: we will most likely end up using syself + hetzner, but we can keep this here for now

kura/spec/e2e/clients_spec.sh
```diff
@@ -0,0 +1,144 @@
...
+        break
+      fi
+      sleep 1
+    done
+
+    The variable second_build should include 'remote cache hit'
+    The variable second_artifact should include "${marker}"
+  End
+
+  It 'builds Buck2 targets against the REAPI surface in multiple regions'
```
fortmarek, line 75: I wonder if it would have been better to first focus the support on just Bazel, Xcode Cache, Module Cache, and Gradle since the other kinds of caches are not something we want to actively support in the near-ish future.

kura/spec/e2e/cluster_spec.sh
```diff
@@ -0,0 +1,164 @@
...
+  It 'keeps Xcode artifacts readable after a node restart'
+    status="$(status_only -X POST \
+      "${KURA_US_URL}/api/cache/cas/artifact-1?tenant_id=acme&namespace_id=ios" \
+      -H "content-type: application/octet-stream" \
+      --data-binary "xcode-binary")"
+    The variable status should eq 204
+
+    capture_into body \
+      wait_for_contains \
+      "${KURA_EU_URL}/api/cache/cas/artifact-1?tenant_id=acme&namespace_id=ios" \
```
fortmarek, line 69: we know we don't want to have these new caching nodes shared across tenants. And I wonder how much we want to distinguish between projects. I don't really see a point in having either given where we want to go. In other words, would consider dropping support for both query parameters.

kura/src/artifact/kind.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+impl ArtifactKind {
+    pub const fn all() -> [Self; 4] {
+        [Self::Keyvalue, Self::Xcode, Self::Gradle, Self::Module]
+    }
+
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::Keyvalue => "keyvalue",
```
fortmarek: nit: would probably have the casing as `KeyValue` and `key_value`

kura/src/artifact/kind.rs
```diff
@@ -0,0 +1,128 @@
...
+            Self::Xcode => "xcode",
+            Self::Gradle => "gradle",
+            Self::Module => "module",
```
fortmarek: will we also have `content_addressable_storage` artifact? In which case, Xcode Cache artifacts should be just regular KV and CAS artifacts with some extra metadata stored in the appropriate table, so we can appropriate the right data in our dashboard.

fortmarek: In similar vain, should we consider treating module cache and gradle artifacts as CAS artifacts (since that's what they are)?

kura/README.md
```diff
@@ -0,0 +1,393 @@
...
+## 🔌 Protocol Surfaces
+
+Kura exposes multiple cache protocols behind one service:
+
+The following HTTP surfaces are Tuist-specific client protocols. They exist so Tuist clients can talk to Kura without changing their cache behavior:
+
+- 🍎 `Xcode CAS`: `POST/GET /api/cache/cas/{id}?tenant_id=...&namespace_id=...`
+- 🗂️ `Keyvalue / action-cache style entries`: `PUT /api/cache/keyvalue?tenant_id=...&namespace_id=...`
+- 🐘 `Gradle`: `PUT/GET /api/cache/gradle/{cache_key}?tenant_id=...&namespace_id=...`
+- 📦 `Multipart module cache uploads`: `POST /api/cache/module/start?...`, `POST /api/cache/module/part?...`, `POST /api/cache/module/complete?...`, `HEAD/GET /api/cache/module/{id}?...`
```
fortmarek: I think the list of what we support should be higher up

fortmarek: Would rename the Multipart module cache uploads to just "Module Cache" (with a link to our docs)

kura/README.md
fortmarek, line 1: This is fine for now. Before we consider productizing it, we might want to cut down on the content here (and emojis 😅) to focus on the important and then have a dedicated docs page.

fortmarek (approved):
A couple of notes from a Claude review:
- Add a formatting check `cargo fmt --check`
- Consider Swatinem/rust-cache instead of raw actions/cache@v4. It handles target/ cleanup and hashing better; the current scheme can grow indefinitely because kura/target is always added to the key/contents without pruning.
- No supply-chain scan. For a new networked service with JWT and mTLS, consider adding cargo audit (or cargo-deny) as a required job. The diff adds 3,400+ lines of Cargo.lock, so this is a meaningful moment to turn scanning on.
- Stale Elixir boilerplate in ignore files. kura/.gitignore:1 and kura/.dockerignore:1 include _build, deps/, mix.exs, mix.lock, priv, lib, log.nonode@nohost, data.nonode@nohost, erl_crash.dump, *.ez, cache-*.tar, undefined — all irrelevant for a Rust service. Likely a copy-paste from the Elixir server/ project. The dockerignore is also missing .dockerignore / Dockerfile / README.md (minor) and has target (no slash) while .gitignore uses /target/.


As discussed before lunch, I'd also consider mostly aligning on the gRPC interface, definitely for Xcode Cache, but that's out of scope for this PR.

Otherwise, lgtm 📦

kura/spec/e2e/cluster_spec.sh
pepicrft: This will go away once we move to the new tenant-scoped infrastructure. Fromt he perspective of data storage, we don't persist any tenant information.

kura/src/artifact/kind.rs
pepicrft: Agreed. I think the right direction is to unify at the storage model first and keep the current protocol-specific APIs as adapters for now. The branch already normalizes persisted metadata around client and artifact_class, but storage/layout still partitions by protocol kind, so there’s still follow-up work to do. I’d prefer to move the internal storage model toward generic action-cache/blob artifacts first, then converge the API surface later.

pepicrft:
@fortmarek I addressed the actionable items from your top-level review in `2dd2b5e54e`:

- added a `Format` job with `cargo fmt --check`
- switched the Rust jobs from raw `actions/cache` to `Swatinem/rust-cache`
- added an `Audit` job with `cargo audit`
- cleaned up the stale Elixir boilerplate in `kura/.gitignore` and `kura/.dockerignore`

A couple of related follow-ups came with that:

- `kura/rust-toolchain.toml` now declares `clippy` and `rustfmt`, and the workflow unsets `RUSTUP_TOOLCHAIN` so CI actually honors the checked-in toolchain config
- `cargo audit` surfaced a real `rustls-webpki` advisory, so I updated the lockfile to `0.103.13`
- I also added a narrow audit ignore for the `jsonwebtoken` RSA advisory because Kura only uses HMAC JWT verification, not RSA algorithms

The gRPC-alignment note is still intentionally out of scope for this PR, as you suggested. Locally I re-ran `cargo fmt --check`, `cargo clippy --locked --all-targets -- -D warnings`, `cargo test --locked`, and `cargo audit`, and they all pass in the same shape the workflow now uses.

PR #10393, opened 2026-04-23
feat(server): add total flaky tests and total tests counts to dashboard

fortmarek:
Screenshot 2026-04-23 at 17 04 37
Screenshot 2026-04-23 at 17 04 44

Surfaces project-scoped totals on the **Flaky Tests** and **Test Cases** dashboard pages so users can see, at a glance, how many tests are currently considered flaky and how many test cases exist in total — without needing to switch pages.

Based on this feature request: add a number of total flaky tests on `/tests/flaky-tests` (similar to the quarantined-tests page), plus a total test cases count that puts flaky numbers into perspective. Same total test cases count on `/tests/test-cases`.

**What changed:**

- New `Tuist.Tests.project_test_cases_count/1` and `Tuist.Tests.project_flaky_test_cases_count/1` helpers that run scoped `FINAL` counts against the `test_cases` ReplacingMergeTree table.
- Flaky Tests page now renders two always-visible legends (`Flaky Tests`, `Total Tests`) above the flaky-runs chart.
- Test Cases page renders a `Total Tests` legend at the top of the Test Cases card.
- Both counts are loaded via `assign_async` so they don't block the first render.

### How to test locally

1. Start the server: \`mise run dev\`.
2. Open a project dashboard with test data, e.g. \`/<account>/<project>/tests/flaky-tests\` — the top of the card should show \"Flaky Tests\" and \"Total Tests\" legends.
3. Open \`/<account>/<project>/tests/test-cases\` — the top of the Test Cases card should show a \"Total Tests\" legend.
4. Run the test suite:
   - \`mix test test/tuist/tests_test.exs\`
   - \`mix test test/tuist_web/live/flaky_tests_live_test.exs\`
   - \`mix test test/tuist_web/live/test_cases_live_test.exs\`


server/lib/tuist/tests/analytics.ex
```diff
@@ -164,6 +165,124 @@ defmodule Tuist.Tests.Analytics do
...
+      values: values,
+      dates: dates
+    }
+  end
+
+  defp active_test_cases_count(project_id, endpoint, window_seconds) do
+    window_start = DateTime.add(endpoint, -window_seconds, :second)
+
+    ClickHouseRepo.one(
+      from(tcr in TestCaseRun,
```
pepicrft: `test_cases_count_analytics/2` only uses the date range and never applies `opts[:is_ci]` to the `TestCaseRun` query. `TestCasesLive.assign_analytics/2` does pass the page's CI/local filter in `opts`, so the new Test cases widget and chart stay project-wide while the other widgets switch scope, which produces incorrect numbers whenever users change the Environment dropdown.

pepicrft (commented):
Inline review note.

server/lib/tuist_web/live/flaky_tests_live.ex
```diff
@@ -241,11 +255,26 @@ defmodule TuistWeb.FlakyTestsLive do
...
+    |> assign(:analytics_trend_label, analytics_trend_label(preset))
+    |> assign(:analytics_selected_widget, analytics_selected_widget)
     |> assign_async(:flaky_runs_analytics, fn ->
       {:ok, %{flaky_runs_analytics: Analytics.test_run_analytics(project.id, Keyword.put(opts, :is_flaky, true))}}
     end)
+    |> assign_async(:flaky_tests_analytics, fn ->
+      {:ok,
+       %{
+         flaky_tests_analytics:
+           Analytics.flaky_tests_analytics(project.id, start_datetime: start_datetime, end_datetime: end_datetime)
```
pepicrft: This builds an environment-scoped `opts` list, but the new `flaky_tests_analytics/2` call drops it and only passes the dates. That means the Flaky Runs widget and table honor the CI/local filter while the new Flaky Tests widget and chart remain project-wide, so the page mixes two different scopes as soon as the user changes Environment.

pepicrft (commented):
Inline review note.

server/lib/tuist_web/live/test_cases_live.html.heex
```diff
@@ -228,7 +255,9 @@
              (@analytics_selected_widget == "test_case_run_count" and
                 @test_case_runs_analytics.result.count != 0) or
              (@analytics_selected_widget == "failed_test_case_run_count" and
-                @failed_test_case_runs_analytics.result.count != 0))
+                @failed_test_case_runs_analytics.result.count != 0) or
+             (@analytics_selected_widget == "test_cases_count" and
+                @test_cases_count_analytics.result.count != 0))
```
pepicrft: `test_cases_count_analytics.count` is the value at the last bucket, not a total over the whole selected range. If the project had active test cases earlier in the period but none in the trailing 14-day window at `end_datetime`, `values` will still contain non-zero history while `count` is 0. This condition renders the empty state instead of the real chart, so the new widget loses the exact rise-and-fall history it was added to show.

pepicrft (commented):
Inline review note.

pepicrft (approved):
-

pepicrft:
Lol, apologies for the messed-up comments. It shouldn't have dumped those.

PR #10399, opened 2026-04-23
fix(server): redirect SSO callback to pending invitation instead of 401

fortmarek:
### Summary

When a Tuist user is invited to an organization whose SSO provider differs from the one they originally signed up with, the SSO callback would 401 with `:existing_user_not_member` and leave them with no path forward. The invitation accept page also required authentication, but the only login button presented to them in practice was the org's SSO — so they ended up looping through the same 401.

This PR breaks the loop in one place: when the SSO callback finds an existing user, the user is *not* a member of the authenticating org, *and* there is a pending invitation for that email + org, the callback now redirects to `/auth/invitations/<token>` instead of raising 401.

The invitation page itself stays behind `:require_authenticated_user`, so the redirect bounces the visitor to `/users/log_in` (with `:user_return_to` preserved). They can authenticate via any of their existing identities — Google, GitHub, password, etc. — instead of being forced through the org's SSO. After login they're returned to the invitation page where Accept is clickable.

### What changes

- **`auth_controller.ex`** — in `link_existing_user_and_log_in`, before raising `:existing_user_not_member`, check for a pending invitation matching `(user.email, sso_organization)` via `pending_invitation_for/2`. If one exists, redirect to the invitation accept page. If not, the existing 401 still fires.
- **`accounts.ex`** — adds `Accounts.get_invitation_by_token/1` (token-only, no `%User{}` argument). Used by the LiveView to render an explanatory "Wrong account" message when the session user's email doesn't match the invitee email, instead of the generic "Invitation not found" path that hid that case.
- **`accept_invitation_live.ex`** — small render refactor: a single tagged `@invitee_state` (`{:ok, invitation}` / `{:mismatched, email}` / `:not_found`) drives a `cond` over the four render branches (accepted, declined, ok, mismatched, not-found). Authentication is unchanged — the page still uses `:ensure_authenticated` and accept/decline still run through `phx-click`.

### Security

Authentication for accepting an invitation is unchanged from main — you still need to be logged in as the matching user. The SSO callback's existing safeguard against custom-IdP account takeover (`:existing_user_not_member`) is preserved for the case where there is *no* pending invitation; the redirect path only fires when an admin has already issued an invitation, which is the recorded-on-server signal of intent.

### Tests

- `test/tuist_web/controllers/auth_controller_test.exs` — new test asserts the SSO callback redirects to `/auth/invitations/<token>` when a pending invitation exists, and that membership/oauth_identity/invitation are not yet mutated until the user explicitly accepts.
- `test/tuist_web/live/accept_invitation_live_test.exs` — coverage for the unauthenticated redirect, accept, decline, wrong-account, and not-found paths.

### How to test locally

1. Create user A via password signup with email `a@example.com`.
2. As an org admin, configure SSO (Okta or generic OAuth2) on the org.
3. Invite `a@example.com` to that org.
4. Log out. Run `tuist auth login` as user A and pick the org's SSO.
5. Expect a redirect to `/auth/invitations/<token>` instead of a 401, then the standard login page (where you can pick Google/password instead of SSO), then the accept page, then dashboard membership.

Original draft attempted to also let the invitation token act as a one-click magic-link login (so SSO could redirect straight to a page that auto-accepted on click). That widened the trust on the invitation token more than this fix needs to and was reverted in 3c946b6.


server/lib/tuist_web/controllers/auth_controller.ex
```diff
@@ -198,29 +198,40 @@ defmodule TuistWeb.AuthController do
...
-        |> Authentication.log_in_user(existing_user, auth_params)
-      else
-        Authentication.log_in_user(conn, existing_user, auth_params)
-      end
-    else
-      log(
-        :warning,
-        "Refused to link existing user #{existing_user.id} via custom SSO provider #{auth.provider}: user is not a member of the authenticating organization."
-      )
+        |> redirect(to: ~p"/auth/invitations/#{invitation.token}")
```
pepicrft, line 236: Redirecting to the invitation page here sends an unauthenticated user back through `:require_authenticated_user`, which overwrites any existing `:user_return_to`. In the device-code login flow that replaces `/auth/device_codes/:device_code?...` with the invitation URL, so after the user signs in with their existing identity they land on the invitation page and the original `tuist auth login` request is abandoned. We should preserve the prior return target before this redirect or keep the invitation page out of the authenticated pipeline.

server/lib/tuist_web/live/accept_invitation_live.ex
```diff
@@ -174,19 +219,20 @@ defmodule TuistWeb.AcceptInvitationLive do
...
+      invitee: user,
       organization: socket.assigns.organization
     })
 
     {:noreply, assign(socket, accepted: true)}
   end
 
-  def handle_event("decline_invitation", _, socket) do
+  def handle_event("decline_invitation", _params, socket) do
     Accounts.delete_invitation(%{invitation: socket.assigns.invitation})
```
pepicrft, line 254: These handlers currently trust the rendered UI state instead of re-checking `@invitee_state` or reloading the invitation server-side. A crafted `accept_invitation` or `decline_invitation` event from the `Wrong account` or `Invitation not found` states will call `Accounts.*` with `nil` assigns and crash the LiveView with a function-clause error. It would be safer to reject non-`{:ok, invitation}` states before mutating.

pepicrft (approved):
I left two inline follow-up comments for the return-target handling and the LiveView mutation guards so they are tracked in the PR.

fortmarek:
@pepicrft good catch — addressed in 8a82388.

Before issuing the redirect to the invitation page, the SSO callback now stashes the prior return target (`oauth_return_to` if set, else the existing `user_return_to`) under a separate session key `:post_invitation_return_to`. The require-auth bounce can clobber `:user_return_to` freely after that, since the original URL is no longer there.

The accept handler in `AcceptInvitationLive` reads `post_invitation_return_to` from the session at mount, and on Accept click it redirects to that path instead of rendering the success page when set. So the device-code flow resumes cleanly: SSO → invitation page → log in via existing identity → return to invitation page → click Accept → land back on `/auth/device_codes/<code>?type=cli`, which completes the original `tuist auth login`.

Tests:
- `auth_controller_test.exs` — asserts `:post_invitation_return_to` is set to the original `user_return_to` when SSO redirects to the invitation page.
- `accept_invitation_live_test.exs` — asserts the LiveView redirects to that path when Accept is clicked.

fortmarek:
@pepicrft fixed in 2ff519d.

Both `handle_event/3` clauses now head-match on `%{assigns: %{invitee_state: {:ok, _}}} = socket` and fall through to a no-op clause for any other state. The buttons that fire these events are still only rendered in the `{:ok, invitation}` branch — this is the server-side belt to the UI's suspenders.

Two new tests pin the behavior:
- `ignores accept_invitation pushed from a not-found state` — pushes `accept_invitation` and `decline_invitation` via `render_hook/3` against a LiveView mounted with an invalid token, asserts the page still says "Invitation not found" (no crash).
- `ignores accept_invitation pushed from the wrong-account state` — same for a session user whose email doesn't match the invitee, asserts no membership change and the invitation survives.

PR #10403, opened 2026-04-23
fix(cli): propagate default-enabled Swift package traits

pepicrft:
## Summary
- fixes Swift package generation when a dependency enables SwiftPM's default traits
- expands selected default traits to their underlying trait names before filtering trait-gated package dependencies
- keeps generated dependencies and `SWIFT_ACTIVE_COMPILATION_CONDITIONS` aligned for cases like Hummingbird's `ConfigurationSupport`

## Validation
- Added regression tests in `PackageInfoMapperTests` for default trait expansion and explicit `traits: []`.
- `swift build --replace-scm-with-registry --target TuistLoader`
- Red/green external reproduction with a temporary Swift test package that exercises `SwiftPackageManagerGraphLoader.load`:
  - against `origin/main`, the Hummingbird-style reproduction failed because `Configuration` was missing from `HummingbirdCore`
  - against this branch, the same reproduction passed
- `xcodebuild test` is still not usable in the generated workspace because the `Tuist-Workspace` schemes have empty `Testables`.
- `swift test` for the repo still doesn't run these loader tests directly because `TuistLoaderTests` is not part of the current checked-in SwiftPM package graph.

### DIFF
--- a/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
+++ b/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
@@ -226,7 +226,7 @@ public struct SwiftPackageManagerGraphLoader: SwiftPackageManagerGraphLoading {
         )
 
         let packageInfoDictionaryById = Dictionary(uniqueKeysWithValues: packageInfos.map { ($0.id, $0.info) })
-        let enabledTraitsPerPackage = enabledTraits(
+        let enabledTraitsPerPackage = Self.enabledTraits(
             rootPackageInfo: rootPackage,
             packageInfos: packageInfoDictionaryById
         )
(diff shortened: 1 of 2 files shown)


cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
```diff
@@ -363,17 +365,62 @@ func enabledTraits(
...
+            enabledTraits,
+            packageTraits: packageInfos[dependency.identity]?.traits ?? []
+        )
+
+        guard !resolvedTraits.isEmpty else { continue }
+        result[dependency.identity, default: []].formUnion(resolvedTraits)
+    }
+}
+
+private func resolvedEnabledTraits(
```
fortmarek: I think a top level method is not really needed, any reason why it's not part of the loader struct itself?

fortmarek (approved):

cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
pepicrft: Not really. I'll place it in the `struct`.

PR #10409, opened 2026-04-23
fix(server): stabilize clickhouse transaction-backed tests

pepicrft:
## Summary
- point the server test setup at the Tuist `ch` and `ecto_ch` forks with transaction support enabled locally
- replace ClickHouse test cleanup-by-truncate with sandbox-backed transactions, inline test-mode ingestion writes, and UTC-safe analytics/query behavior
- mark the ClickHouse-heavy analytics suites as `async: true` and document the `system.tables` metadata lookup as the one test-only transaction bypass

## Why
The previous ClickHouse test path relied on truncation and background buffering, which made test isolation different from Postgres, hid transitive ClickHouse dependencies from developers, and left some analytics behavior dependent on local timezone/session details.

## Impact
Developers can use the standard test cases without needing to think about whether ClickHouse is involved, while the local transaction-enabled ClickHouse setup now supports the same mental model much more closely.

## Validation
- `mise x -- bash -lc 'export TUIST_SERVER_CLICKHOUSE_HTTP_PORT=28123 TUIST_SERVER_TEST_PORT=4616 MIX_ENV=test MIX_BUILD_PATH=_build/codex-clickhouse-pr; cd server && mix test'`


mise/utilities/dev_instance_env.sh
```diff
@@ -141,6 +141,8 @@ export TUIST_SERVER_CLICKHOUSE_NATIVE_PORT="9000"
 export TUIST_SERVER_CLICKHOUSE_INTERSERVER_HTTP_PORT="9009"
 export TUIST_SERVER_CLICKHOUSE_MYSQL_PORT="9004"
 export TUIST_SERVER_CLICKHOUSE_POSTGRESQL_PORT="9005"
+export TUIST_SERVER_CLICKHOUSE_KEEPER_PORT="9181"
```
pepicrft, line 144: Transactions depend on a coordination layer named keeper, so we need to configure it here.

server/lib/tuist/command_events.ex
```diff
@@ -432,13 +458,13 @@ defmodule Tuist.CommandEvents do
       add_filters(
         from(e in Event,
           as: :event,
-          group_by: fragment("formatDateTime(?, ?)", e.ran_at, ^date_format),
+          group_by: fragment("formatDateTime(?, ?, 'UTC')", e.ran_at, ^date_format),
```
pepicrft, line 459: This is needed needed for deterministic analytics/query semantics across machines and test runs

fortmarek (approved):

PR #10417, opened 2026-04-23
fix(cli,server): add muted/unmuted test case event types

justine-acorns:
### Summary

The server emits `muted` / `unmuted` test case events (a rename of `quarantined` / `unquarantined`), but the OpenAPI spec's `event_type` enum was never updated to include them. The CLI's generated decoder rejects the response as a result, breaking \`tuist test case events\` end-to-end with:

\`\`\`
✖ Error Client encountered an error invoking the operation 'listTestCaseEvents'
\`\`\`

### Changes

- **server** (\`test_cases_controller.ex\`): add \`muted\` and \`unmuted\` to the \`event_type\` enum. Legacy \`quarantined\` / \`unquarantined\` values are kept so older events still decode.
- **cli** (generated \`server.yml\` + \`Types.swift\`): mirror the two new enum cases. These changes would normally come from running \`mise run generate-api-cli-code\`, but are applied by hand here to keep the diff minimal and avoid pulling in unrelated generated drift.
- **cli** (\`TestCaseEventsCommandService.swift\`): extend the display switch with the new cases.
- **tests**: update \`run_with_text_output\` to also assert \`Muted\` renders, since that's the case this PR unblocks.

### Notes

- While investigating, I noticed \`.github/workflows/server.yml\` contains a full OpenAPI-drift check job (regenerate spec, fail CI if git status is dirty) that is entirely commented out. Re-enabling it would have caught this before release. Happy to open a separate issue/PR for that if useful.
- There's a broader discussion to be had about whether forward-evolving enum fields (event kinds, CI providers, platforms, etc.) should use an extensible-enum pattern so the CLI doesn't hard-fail on unknown server values. Out of scope here; this PR only restores the currently-documented behavior.

### How to test locally

\`\`\`bash
swift test --package-path . --replace-scm-with-registry --filter TestCaseEventsCommandServiceTests
\`\`\`

Running \`tuist test case events <id>\` against a project whose history contains \`muted\` events now succeeds and renders \`Muted (automatic)\`, where previously the decode failed the whole request.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -926,6 +926,7 @@ var targets: [Target] = [
             "TuistServer",
             "TuistEnvironmentTesting",
             "TuistOIDC",
+            "TuistTesting",
             "TuistUserInputReader",
             pathDependency,
             fileSystemDependency,
@@ -986,6 +987,7 @@ var targets: [Target] = [
             "TuistConfigLoader",
             "TuistEnvironmentTesting",
             "TuistNooraTesting",
+            "TuistTesting",
             mockableDependency,
         ],
         path: "cli/Tests/TuistRegistryCommandTests"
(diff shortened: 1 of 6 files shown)


fortmarek (approved):
Thanks for the fix!

PR #10423, opened 2026-04-24
fix(cli): restore static framework .xcstrings localization

### LINKED ISSUE #10420: Strings from String catalogs for static frameworks can't be found via Bundle.module
### What happened?

With the recent PRs @pepicrft fixed the stale marking / compiler extraction of string catalogs for static frameworks for normal sources / resource manifests and buildable folders.

But unfortunately this also seems to have created a regression where the actual localized string can't be looked up in the running app resulting in keys to be shown instead.

Image

Image

### How do we reproduce it?

[generated_ios_app_with_static_framework_with_xcstrings.zip](https://github.com/user-attachments/files/27035876/generated_ios_app_with_static_framework_with_xcstrings.zip)

I modified the fixture sample and added English as well as German localizations for the GreetingView, which is shown when the app target is run. I also added the CFBundleAllowMixedLocalizations: true Info.plist key value just in case.

With tuist 4.185.1 this results in keys being shown.
With tuist 4.179.3 this results in localized strings being shown.

### Error log

Run app and see if strings are localized.

### macOS version

26.4.1 (25E253)

### Tuist version

4.185.1

### Xcode version

Version 26.5 beta 2 (17F5022i)

pepicrft:
## Summary
- restore static-framework `.xcstrings` handling so localized catalogs are compiled into the companion resource bundle again
- keep `.xcstrings` in the framework target's sources so Xcode still performs extraction while `Bundle.module` continues to resolve through the synthesized bundle
- share `.xcstrings` across the source and resource targets when splitting buildable folders
- cover the regression with mapper tests and the static framework `.xcstrings` acceptance test

## Testing
- `mise run cli:lint`
- built the `tuist` scheme and regenerated `examples/xcode/generated_ios_app_with_static_framework_with_xcstrings`, then verified localized lookups resolve from the companion bundle again
- reproduced the same regression on the sample attached to `#10420` and verified the patched CLI restores companion-bundle lookups there too

Closes #10420

### DIFF
--- a/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
+++ b/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
@@ -517,8 +517,8 @@ extension BuildableFolder {
         sharedEntries: [BuildableFolderFile],
         resourceOnlyEntries: [BuildableFolderFile]
     ) {
-        let originalOnlyEntries = resolvedFiles.filter { $0.path.shouldStayOnlyOnOriginalTargetWhenSplittingResources }
-        let sharedEntries = resolvedFiles.filter { $0.path.shouldBeSharedAcrossTargetsWhenSplittingResources }
+        let originalOnlyEntries = resolvedFiles.filter(\.path.shouldStayOnlyOnOriginalTargetWhenSplittingResources)
+        let sharedEntries = resolvedFiles.filter(\.path.shouldBeSharedAcrossTargetsWhenSplittingResources)
         let resourceOnlyEntries = resolvedFiles.filter {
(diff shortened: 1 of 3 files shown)


fortmarek (approved):

adarhef:
@pepicrft Hey, can I ask why we still have the resource bundle in the first place for localizable strings? Or at all actually?
I thought it was removed at some point since static frameworks are capable of holding resources, and then I started noticing the companion bundles again recently.

I'm also noticing on my end that string extraction is wonky after installing the latest CLI. My localizable strings all get erased from the `xcstrings` file when running a build now. They're in the sources list for the static framework target and are copied in the companion bundle, but still I'm not getting correct extraction.

When I manually add the `xcstrings` file to the copy bundle resources phase of the static framework, string extraction works as expected.

PR #10425, opened 2026-04-24
perf(server): speed up test_case_runs slow queries with lookup MVs

fortmarek:
## Summary

Two alerting query patterns on the `test_case_runs` ClickHouse table scan millions of rows per execution because their filter columns are not part of the table's sort key `(project_id, test_case_id, ran_at, id)`:

- **Pattern 1** — cross-run flakiness lookup by `(project_id, git_commit_sha)` in [`Tuist.Tests.get_existing_ci_runs_for_commit`](https://github.com/tuist/tuist/blob/main/server/lib/tuist/tests.ex#L1043). Reads 4-6M rows per execution, p90 ~4.2s.
- **Pattern 2** — `FlakyTestsMonitor.evaluate_by_run_count` filters by `(project_id, inserted_at, is_flaky)` and reads up to 39M rows per execution (up to 8.7s).

Projections cannot be used because the source table is a `ReplacingMergeTree` (ClickHouse issue [#46968](https://github.com/ClickHouse/ClickHouse/issues/46968)) — the 20260319 reorder migration explicitly dropped them for this reason. This PR follows the existing lightweight TO-table MV pattern already used by `flaky_test_case_runs_mv` and `test_case_runs_by_test_run_mv`.

### Changes

- New MV `test_case_runs_by_commit` (7 columns: `project_id, git_commit_sha, is_ci, status, id, test_case_id, inserted_at`) ordered by `(project_id, git_commit_sha, is_ci, status, id)`, with partitioned backfill.
- `get_existing_ci_runs_for_commit` now queries the slim MV for `{id, test_case_id, status}`. `mark_test_case_runs_as_flaky` re-fetches full rows from the main table by `id` (leveraging the `idx_id` bloom filter) only for the small cross-run-flaky subset — usually 0 rows, in the worst case a handful.
- `FlakyTestsMonitor.evaluate_by_run_count` and `flaky_candidate_ids` now hit the existing `flaky_test_case_runs` MV, which stores only flaky rows and is ordered `(project_id, ran_at, test_case_id)`. The `evaluate/1` denominator query stays on the main table since it needs non-flaky rows too, but now benefits from prefix alignment on `test_case_id IN (candidates)`.

### Why not the original recommendations?

- **Projection on `(project_id, git_commit_sha, is_ci, status)`** — unusable with `ReplacingMergeTree`.
- **Bloom-filter GRANULARITY 1 → 4** — that makes the index *coarser*, not more selective; would degrade pruning, not improve it.
- **Tightened `inserted_at` cast** — the deeper issue is that `inserted_at` isn't a sort-key prefix, so cast hygiene alone doesn't move the needle.

### Local benchmarks

2M seeded rows, `project_id=2382` weighted to 70% of data:

| Query | Before (main table) | After (MV) | Reduction |
|---|---|---|---|
| Pattern 1: commit lookup | 13ms, 808,976 rows | 2.3ms, 8,192 rows | **~100×** fewer rows |
| Pattern 2: run-count flaky | 26ms, 1,410,176 rows | 8ms, 99,772 rows | **~14×** fewer rows |
| `flaky_candidate_ids` | 16ms, 1,410,176 rows | 8ms, 99,772 rows | **~14×** fewer rows |

## How to test locally

1. Run `mise run dev` to start the server (runs the new ingest migration).
2. Seed some `test_case_runs` data and verify `test_case_runs_by_commit` is populated by the MV trigger.
3. Run the affected suites:
   - `mix test test/tuist/tests_test.exs`
   - `mix test test/tuist/automations/`
   - `mix test test/tuist/tests/`
   - `mix test test/tuist_web/live/flaky_tests_live_test.exs`

## Test plan

- [x] `mix test` passes for affected modules (464 tests, 0 failures locally)
- [x] `mix format --check-formatted` passes
- [x] `mix credo --strict` passes on new/changed files
- [x] Local ClickHouse benchmark confirms row-count reduction on both patterns
- [ ] Staging deploy: verify migration runs and MV backfill completes for production volumes
- [ ] Post-deploy: observe alert query durations drop in Grafana


server/lib/tuist/tests.ex
```diff
@@ -1898,10 +1900,13 @@ defmodule Tuist.Tests do
   defp mark_test_case_runs_as_flaky([]), do: :ok
 
   defp mark_test_case_runs_as_flaky(runs) when is_list(runs) do
+    ids = runs |> Enum.map(& &1.id) |> Enum.uniq()
+
+    full_runs =
+      ClickHouseRepo.all(from(tcr in TestCaseRun, where: tcr.id in ^ids))
```
pepicrft: Because `test_case_runs` is a `ReplacingMergeTree`, reading it back by `id` without `FINAL` can return multiple versions of the same row once a run has already been reinserted as flaky. The old `Enum.uniq_by(& &1.id)` protected us from that, but this path now re-inserts every returned version. Would it be safer to add `hints: ["FINAL"]` here or re-dedupe `full_runs` by `id` before building `updated_runs`?

pepicrft (approved):
Looks good overall. I left one small inline follow-up about duplicate handling in the refetch path, but I’m happy with the direction and perf improvement.

PR #10426, opened 2026-04-24
feat(cli): generate .xctestplan files from ProjectDescription

fortmarek:
## Summary

Adds the ability to declare test plans in Swift and have Tuist generate the `.xctestplan` files, so the list of test targets stays in sync with the manifest as features are added or removed. Generated and hand-maintained plans can be mixed in a single scheme.

Motivated by a [community question](https://tuist.dev/community) about selective-testing schemes that compose dynamic lists of per-feature test targets.

### New API

A single `.testPlan(...)` factory picks the right shape from its parameters:

```swift
import ProjectDescription

Scheme(
    name: "SelectiveTests",
    testAction: .testPlans([
        .testPlan(
            name: "UnitTests",
            testTargets: allFeatureModules.map { .testableTarget(target: .target("\($0)Tests")) },
            isDefault: true
        ),
        .testPlan(name: "SnapshotTests", testTargets: snapshotTestables),
        .testPlan(path: "TestPlans/Legacy.xctestplan"),
    ])
)
```

- **`.testPlan(path:)`** — reference an existing hand-maintained `.xctestplan`. Glob patterns supported.
- **`.testPlan(name:testTargets:path:isDefault:)`** — Tuist generates the file from Swift. Defaults to `Derived/TestPlans/<name>.xctestplan` (gitignored alongside other derived artefacts); pass an optional `path:` to pin the file to a checked-in location when external tooling like `xcodebuild -testPlan` needs a predictable path.
- **Convenience factories** mirror `Path`: `.relativeToManifest(_:)`, `.relativeToRoot(_:)`, `.relativeToCurrentFile(_:)`.
- **String literal** — `TestPlan` conforms to `ExpressibleByStringLiteral`, so `"Foo.xctestplan"` in the list is shorthand for `.testPlan(path: "Foo.xctestplan")`.

### Backwards compatibility

All existing manifest patterns continue to work unchanged through the single new overload:

```swift
.testPlans(["Default.xctestplan", "Other.xctestplan"])           // via ExpressibleByStringLiteral
.testPlans([.relativeToManifest("App.xctestplan")])              // via TestPlan.relativeToManifest
.testPlans([.relativeToManifest("*.xctestplan")])                // glob, resolver unchanged
```

Verified against every `.testPlans(...)` call site in the repo (fixtures under `examples/xcode/`).

## Design notes

- `TestPlan` is a struct wrapping a `Source` enum (`.path(Path)` / `.generated(name:testTargets:path:)`). Users never touch the enum cases directly — construction goes through the `testPlan(...)` factory, the `relativeTo...` convenience factories, or bare string literals. The mapper switches on `entry.source` internally.
- Default generated location lives under `Derived/`, which Tuist already gitignores, so no git churn and no "who owns this file?" ambiguity.
- The `.xctestplan` is written **after** the owning `.xcodeproj`. Until `XcodeProj` serializes a project it uses temporary PBX object references, so the blueprint identifiers embedded in each test-plan target only become stable post-write. `XcodeProjWriter.writeTestPlans(_:)` runs right after `xcodeProj.write(...)`.
- `XcodeGraph.TestPlan` grew an `isGenerated` flag so the linter can skip the existence check for plans Tuist itself writes, and so `TestPlanGenerator` only emits descriptors for plans that opt in.
- Reading existing plans back from an on-disk project (via `XcodeGraphMapper`) works unchanged — `isGenerated` defaults to `false`.
- Test plan JSON output uses Xcode's key order and 2-space indent; the configuration UUID is deterministic per plan to minimize churn when the file is pinned to a checked-in location.
- A single shared `TestPlan.resolve(entries:...)` handles both path and generated entries (glob expansion, default-plan inference, missing-file warnings), so legacy string-literal / `.relativeToManifest` call sites go through the exact same code path as the new factory.

## Test plan

- [x] New `TestPlanGeneratorTests` verifies the generated JSON (keys, identifier, enabled flag).
- [x] New `SchemeDescriptorsGeneratorTests` case asserts only generated plans produce descriptors.
- [x] New `TestActionManifestMapperTests` cases cover: default-plan inference, explicit `isDefault`, the derived path default, the explicit-path override, and mixing path + generated entries in a single list.
- [x] Pre-existing mapper tests (string literals, `.relativeToManifest`, glob expansion, missing-file warning) still pass unchanged.
- [x] All `TuistLoaderTests` (227) and relevant `TuistGeneratorTests` (62) still pass.


cli/Sources/ProjectDescription/TestAction.swift
```diff
@@ -3,8 +3,18 @@
...
+    ///
+    /// Construct each entry with `TestPlan.testPlan(path:)` for a hand-maintained `.xctestplan`,
+    /// or `TestPlan.testPlan(name:testTargets:)` for a plan whose file Tuist generates from
+    /// Swift. Both kinds can be mixed freely. `TestPlan.relativeToManifest(_:)`,
+    /// `.relativeToRoot(_:)`, and `.relativeToCurrentFile(_:)` are convenience factories for
+    /// the path form.
+    ///
+    /// Because `TestPlan` conforms to `ExpressibleByStringLiteral`, bare string literals are
+    /// accepted too — `testPlans(["Foo.xctestplan"])` is equivalent to
+    /// `testPlans([.testPlan(path: "Foo.xctestplan")])`.
```
fortmarek: this is kept for backwards compatibility primarily.

cli/Sources/XcodeGraph/Sources/XcodeGraph/Models/XCTestPlan.swift
```diff
@@ -0,0 +1,74 @@
...
+
+    public let configurations: [Configuration]?
+    public let defaultOptions: [String: String]?
+    public let testTargets: [TestTarget]
+    public let version: Int?
+
+    public init(
+        testTargets: [TestTarget],
+        configurations: [Configuration]? = nil,
+        defaultOptions: [String: String]? = nil,
```
pepicrft: The shared `XCTestPlan` model now decodes `defaultOptions` as a string dictionary, but real Xcode test plans in this repo already store booleans and nested objects there. I reproduced this with the checked-in `All.xctestplan` fixture, which now fails with a type mismatch at `defaultOptions.codeCoverage`. Because the loader and `XcodeGraphMapper` both use this model, existing path-based test plans regress immediately.

cli/Sources/TuistLoader/Models+ManifestMappers/TestPlan+ManifestMapper.swift
```diff
@@ -1,35 +1,154 @@
...
+                    schemeName: schemeName,
+                    fileSystem: fileSystem,
+                    into: &resolved
+                )
+            case let .generated(name, testTargets, path):
+                let resolvedPath: AbsolutePath = if let explicitPath = path {
+                    try generatorPaths.resolve(path: explicitPath)
+                } else {
+                    derivedDirectory.appending(component: "\(name).xctestplan")
+                }
```
pepicrft, line 42: When no explicit path is provided, the generated plan path is derived only from the plan name. Two schemes in the same manifest that both generate a `UnitTests` plan will therefore write to the same `Derived/TestPlans/UnitTests.xctestplan`, and the later side effect silently overwrites the earlier one. That makes one scheme run the other scheme's target set unless duplicate paths are rejected or namespaced.

cli/Sources/TuistGenerator/Extensions/TestPlanDescriptor+Extras.swift
```diff
@@ -0,0 +1,46 @@
...
+            else {
+                continue
+            }
+            let containerRelativePath = projectPath.relative(to: rootPath).pathString
+            testTargets.append(
+                TestPlanDescriptor.TestTarget(
+                    pbxTarget: pbxTarget,
+                    containerPath: "container:\(containerRelativePath)",
+                    isEnabled: !testableTarget.isSkipped
+                )
```
pepicrft, line 39: This conversion keeps only whether a target is enabled and discards `TestableTarget.parallelization`. `TestPlanDescriptor.encode()` then never sets the `parallelizable` field, and elsewhere omitted `parallelizable` is interpreted as the Swift Testing only mode. A manifest-generated plan therefore cannot preserve enabled or disabled parallelization from the user's `TestableTarget` list.

pepicrft (commented):
Leaving the follow-up findings inline for visibility.

pepicrft (approved):
Aside from minor comments, it looks good.

PR #10427, opened 2026-04-24
ci(server): fix OOM in Gradle Cache Acceptance by sharing one BEAM across setup

fortmarek:
## Summary

The Gradle Cache Acceptance job was intermittently failing on main and PRs (e.g. [run 24880633451](https://github.com/tuist/tuist/actions/runs/24880633451/job/72847850714?pr=10368)) with `mix ecto.create` killed at exit 137 (OOM).

The workflow ran `mix compile` in one step and then `mix ecto.create` in a separate step. In the fresh BEAM, Elixir's manifest tracking re-checked 153 app files, and the parallel compiler's peak memory (one in-flight module copy per worker) overshot the runner's memory ceiling before the DB setup could finish.

## Changes

- Collapse `compile` + ecto tasks + seeds into a single `mix do` so the compile peak happens once inside one BEAM.
- Drop the now-redundant standalone `Compile server` step.
- Run `mix phx.server` with `ELIXIR_ERL_OPTIONS=\"+S 1\"` so any stale-manifest recompile on server start is serial (not 4-wide parallel) and fits in memory.
- Use `nohup` + `disown` to keep the backgrounded BEAM alive across the step boundary, otherwise the e2e step finds nothing listening.
- Wait for the TCP listener with `nc -z` over 180 attempts and dump `/tmp/main-server.log` on failure so future regressions are debuggable.

The same set of changes has 3 consecutive green Gradle Cache Acceptance runs on branch `claude/objective-spence-4ccf3a` (which is behind the unrelated PR #10366).

## Test plan

- [ ] Gradle Cache Acceptance passes on this PR
- [ ] Merged into main and subsequent runs stay green

### DIFF
--- a/.github/workflows/gradle-cache-acceptance.yml
+++ b/.github/workflows/gradle-cache-acceptance.yml
@@ -95,20 +95,9 @@ jobs:
       - name: Install server deps
         working-directory: server
         run: mix deps.get
-      - name: Compile server
-        working-directory: server
-        run: mix compile
       - name: Start ClickHouse
         working-directory: server
         run: mise run --no-deps clickhouse:start
-      - name: Setup main server
-        working-directory: server
-        run: |
-          mix ecto.create
-          mix run priv/repo/timezone.exs
-          mix ecto.load
-          mix ecto.migrate
-          mix run priv/repo/seeds.exs
       - name: Restore cache deps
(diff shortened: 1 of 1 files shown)


pepicrft (commented):
Much better

PR #10428, opened 2026-04-24
feat(server): run processor in-cluster via Oban queue + least-privilege role

fortmarek:
## Summary

Collapses the xcactivitylog processor into the server codebase and turns it into an Oban queue consumer. Instead of a separate Docker image + HTTP webhook path, the same `ghcr.io/tuist/tuist` image runs as either a web server or a processor pod — `TUIST_PROCESSOR_MODE=true` flips the BEAM to `queues: [process_build: N]` only and skips the Phoenix endpoint. SKIP LOCKED on `oban_jobs` gives natural least-busy dispatch with zero coordination logic on the server side.

Production runs 2 replicas for redundancy across rolling deploys + node loss; canary and staging run 1 replica each (staging is squeezed onto a 2-cpx22 cluster where a second pod doesn't fit alongside the server + DaemonSets — the redundancy goal is met where it matters).

## Why one image (instead of two)

The standalone `processor/` Elixir app went away. There's exactly one Docker image, `ghcr.io/tuist/tuist`, and the same release boots in either role. That's a deliberate choice over keeping a separate `ghcr.io/tuist/processor` image:

- **Worker / schema lockstep**. The processor runs `Tuist.Builds.Workers.ProcessBuildWorker` directly out of the server's codebase, against the server's Ecto schemas, against the same encrypted `priv/secrets/<env>.yml.enc`. With two images we'd have to extract the worker (and any schema it touches) into a shared library and version-match it across deploys; a mismatch would crash jobs at runtime. With one image, "what enqueued the job" and "what executes the job" are bit-for-bit identical by construction.
- **One CI build, two deploy targets**. The Helm chart points both `Deployment`s at the same image tag pushed by the existing `server-deployment.yml` workflow. No second build job, no second cache scope, no risk of drift between server and processor SHAs.
- **Configuration is small**. `TUIST_PROCESSOR_MODE=1` + `TUIST_WEB=0` + a narrowed `oban_queues` list is all that distinguishes processor mode from server mode at runtime. The web endpoint is still supervised but doesn't bind a port; the processor simply doesn't start the Phoenix listener. ~30 MB of Swift runtime libs travel along on every server pod that doesn't need them — measured against the alternative (a whole second image build pipeline + the worker-extraction refactor), that footprint is immaterial.

The trade-off is that processor pods boot the full Tuist supervision tree (Repos, Vault, Cachex, ingestion buffers, Oban, OTel, etc.) for what's nominally a queue consumer. ~600–800 MiB RSS at idle. Trimming the supervision tree is a viable follow-up if/when that overhead becomes the bottleneck.

## What's in the PR

**Code consolidation (`server/`, `processor/` deleted)**
- Standalone `processor/` Elixir app gone — `Processor.BuildProcessor` → `Tuist.Processor.BuildProcessor`, `Processor.XCActivityLogNIF` → `Tuist.Processor.XCActivityLogNIF`. The C bridge's `ERL_NIF_INIT` is updated to register against the new module name; mismatch caused a silent load failure caught in staging e2e.
- Swift NIF source moved to `server/native/xcactivitylog_nif/`. Server Dockerfile builds the NIF and bundles it in every image (~30 MB of Swift runtime libs — immaterial).
- `ProcessBuildWorker` simplified to a single local-processing path: download from S3 via `Tuist.Storage`, parse via `Tuist.Processor.BuildProcessor`, write via `Tuist.Builds`. No HTTP dispatch, no webhook secret, no HMAC signing.
- Drops dead `processor.url` and `processor.webhook_secret` keys from `priv/secrets/{stag,can,prod}.yml.enc` (folded in from #10498, which is now superseded).

**Oban queue gating (`server/config/runtime.exs`)**
- `Environment.processor_mode?/0` reads `TUIST_PROCESSOR_MODE`. When set, the Oban config narrows to `[process_build: queueConcurrency]` only; when unset, the server runs every queue as before.
- `Environment.delegate_process_build?/0` reads `TUIST_DELEGATE_PROCESS_BUILD`. When true, the server pod drops `:process_build` from its queue list entirely so jobs land exclusively on the processor fleet (the chart sets this whenever `processor.enabled: true`). Without it, the server's local Oban worker raced the dedicated processor and won enough jobs to defeat the point of the separate fleet — caught in staging e2e.
- `Environment.process_build_queue_concurrency/0` — per-pod concurrency, defaults to 5 on processor pods. Tunable via `TUIST_PROCESS_BUILD_QUEUE_CONCURRENCY`.
- Postgrex options gated on URL port: when port == 6543 (Supabase Supavisor transaction-mode pooler), drops the `tcp_keepalives_*` startup parameters (which Supavisor rejects with `protocol_violation`) and switches to `prepare: :unnamed` (transaction mode can't reuse named prepared statements). Direct Postgres on 5432 keeps the existing behavior unchanged.

**Helm chart (`infra/helm/tuist/`)**
- New `processor-deployment.yaml` — same image as the server Deployment, `TUIST_PROCESSOR_MODE=1`, `TUIST_WEB=0`, soft `topologySpreadConstraints` across nodes, `maxSurge=1` / `maxUnavailable=0`, preStop drain, emptyDir `/tmp`, no HTTP probes (Oban heartbeat into `oban_peers` is the liveness signal).
- `processor-pdb.yaml` — `minAvailable: 1` (off in staging where there's only ever one pod).
- `processor-external-secrets.yaml` — ESO syncs only the `tuist_processor` role **password** from 1P (`PROCESSOR_DATABASE_PASSWORD`) and templates the full `DATABASE_URL` from `processor.database.{host,port,name,username,params}` chart values. Password rotation is a single-item update.
- `server-deployment.yaml` adds `TUIST_DELEGATE_PROCESS_BUILD=1` whenever `processor.enabled` (handoff above).
- `values-managed-common.yaml` flips processor on; per-env overlays set the Supabase pooler host and `TUIST_DEPLOY_ENV` (matches the server's).
- Staging overlay drops `replicaCount` to 1 + `maxSurge=0/maxUnavailable=1` for both server and processor + tighter memory requests because the 2-worker cpx22 cluster is too tight for surge-1 deploys with two app pods. Production keeps the canonical 2-replica + PDB shape.

**Supabase role provisioning (`infra/supabase/`)**
- `tuist-processor-role.sql` — idempotent `CREATE ROLE tuist_processor` + least-privilege grants: `SELECT/INSERT/UPDATE/DELETE` on `oban_jobs` + `oban_peers`, read-only on `projects` + `accounts`. The build-run tables (`builds`, `build_*`, `cacheable_tasks`, `cas_outputs`, `build_machine_metrics`) live in **ClickHouse**, not Postgres, so the role correctly has no grants on them — those writes go through `Tuist.IngestRepo` / `Tuist.ClickHouseRepo` using credentials baked into `priv/secrets/<env>.yml.enc`. Explicit `REVOKE ALL` keeps future Postgres tables out by default. Password is bounced through a transaction-scoped GUC (`set_config(..., true)`) so it doesn't linger on the connecting session.
- `README.md` — one-shot bootstrap instructions (psql with `-v pw=$PW` against the Supabase direct URL), rotation snippet (`op item edit` + `ALTER ROLE`), and the "why not an Ecto migration" rationale.

**CI / deploy (`.github/workflows/`)**
- Standalone `processor.yml`, `processor-deploy.yml`, `processor-staging-deploy.yml` deleted. Build + deploy happens in the existing `server-deployment.yml` cascade now that there's only one image.

## Rolling-deploy behavior

**Production** (`replicaCount: 2`, `maxSurge: 1`, `maxUnavailable: 0`, PDB `minAvailable: 1`):

- **Steady state**: 2 pods, one per Hetzner ccx23 worker, thanks to `topologySpreadConstraints` with `whenUnsatisfiable: ScheduleAnyway` (soft anti-affinity — prefers separation, allows co-location during surge).
- **Rolling deploy**: new ReplicaSet scales up by 1 (3 pods total briefly, surge pod co-locates on an existing worker), wait for new pod Ready, scale old down by 1, repeat. At least 2 ready pods at all times → `:process_build` always has a consumer.
- **PDB**: backstop for involuntary disruption (node drain, eviction, OS upgrade) — k8s refuses voluntary disruption that would drop below 1 ready pod.
- **In-flight parse during pod termination**: BEAM gets SIGTERM, enters `terminationGracePeriodSeconds: 1260` (21 min) graceful shutdown, finishes any claimed parse, exits, kubelet reaps. If a parse runs longer than 21 min (very unusual), `Oban.Plugins.Lifeline` reclaims it on the next poll cycle of any healthy pod.

**Canary** (`replicaCount: 1`): single pod on a single VM. Same rolling-deploy mechanics as the server (already established in this env).

**Staging** (`replicaCount: 1`, `maxSurge: 0`, `maxUnavailable: 1`, PDB off): the 2-cpx22 cluster has ~1 GiB free per VM after system DaemonSets, so a surge pod doesn't fit alongside the existing processor + the server pod. Old pod terminates first → ~30–60 s gap with no processor consumer → new pod boots → drains queue. Jobs enqueued during the gap sit in `oban_jobs` until the new pod's first poll. Acceptable on staging given the traffic shape.

So: **two Hetzner VMs each running a processor pod in production**, one VM in canary, one VM in staging. The "two replicas for redundancy" goal in the PR title applies to production specifically.

## Pre-deploy checklist (per env)

Before the first deploy with `processor.enabled: true`, each env needs:

1. **Postgres role provisioned** — run `infra/supabase/tuist-processor-role.sql` against the env's Supabase project (direct URL, port 5432), passing the password via `-v pw=$PW`. See `infra/supabase/README.md`.
2. **1Password item** — `PROCESSOR_DATABASE_PASSWORD` in the env's `tuist-k8s-<env>` vault containing the password from step 1. (`op item create` on first setup.)
3. **Supabase pooler host set** — `processor.database.host` in `values-managed-<env>.yaml` (or `--set` at deploy time). Dashboard → Settings → Database → Connection pooling → Host.
4. **ESO SA has vault access** — the 1Password Service Account backing the cluster's `ClusterSecretStore "onepassword"` needs read access to the new item. Usually covered by vault-level grants; verify with `kubectl get externalsecret tuist-tuist-processor-external-secrets` showing `SecretSynced=True` after the first reconcile.
5. **Old polling consumers shut down** — any pre-k8s deploy still pointed at the env's Postgres (Render service, leftover preview env) will race the new processor on SKIP LOCKED. They run stale code that 5xx's every job. Caught in staging e2e — the Render staging service was still polling and discarding every job.

## Cutover plan

1. ✅ **Staging** — checklist done, deploy succeeded, processor consumes the queue end-to-end (verified via real build upload at `marek/flaky-demo`; job claimed by processor pod in <1s, completed in 1.3s).
2. **Canary next** — same checklist against the canary Supabase project, then `environment=canary`.
3. **Production last** — same again. Normal push-to-main cascade flows through once processor is healthy.

## Post-deploy cleanup

Once all three envs have been serving from the Oban-queue path for a release cycle and rollback to the Kamal pipeline is ruled out:

1. **Archive the old processor secrets in the `cache` vault** (if they're still there):
   ```bash
   for env in "" _CANARY _STAGING; do
     for base in PROCESSOR_WEBHOOK_SECRET PROCESSOR_SECRET_KEY_BASE PROCESSOR_S3_ACCESS_KEY_ID PROCESSOR_S3_SECRET_ACCESS_KEY; do
       op item delete --archive --vault cache "${base}${env}" 2>/dev/null || true
     done
   done
   # Leave SENTRY_DSN_PROCESSOR alone — other services may still reference it.
   ```
2. **Terminate the three Hetzner dedicated servers** that ran the old processor (`processor.tuist.dev`, `processor-canary.tuist.dev`, `processor-staging.tuist.dev`). Verify with `dig +short processor.tuist.dev` that nothing still points at them.
3. **Drop the `processor-*.tuist.dev` DNS records** in Cloudflare.

## Test plan

- [x] `mix format --check-formatted` — server
- [x] `mix credo` — touched files
- [x] `mix test` — worker tests 14/14 after rewrite
- [x] `mix compile --warnings-as-errors` — clean
- [x] `helm lint` — staging / canary / production managed overlays
- [x] `helm template` renders cleanly for managed (all three envs) and self-hosted (processor disabled by default, opt-in via `processor.enabled: true`)
- [x] `helm template` emits clear `required` error when `processor.database.host` is missing in managed mode
- [x] Postgres role provisioned on staging Supabase project
- [x] `PROCESSOR_DATABASE_PASSWORD` populated in `tuist-k8s-staging` vault
- [x] Staging deploy succeeded; processor pod claims + completes a real build job (verified end-to-end against `marek/flaky-demo`)
- [ ] Canary + production follow the same playbook
- [ ] PDB verified: `kubectl rollout status` during a redeploy keeps `minAvailable: 1` (production only — staging is intentionally single-pod)
- [ ] Post-deploy cleanup once the cutover soaks


infra/helm/tuist/templates/processor-deployment.yaml
```diff
@@ -0,0 +1,124 @@
...
+            - name: DATABASE_URL
+              valueFrom:
+                secretKeyRef:
+                  name: {{ $processorSecretName }}
+                  key: processor-database-url
+            {{- if or .Values.server.masterKey .Values.server.managedSecrets }}
+            # MASTER_KEY unlocks the rest of priv/secrets/<env>.yml.enc
+            # (ClickHouse URL, S3 creds, Sentry, Stripe, etc.). The release
+            # boots with the same secrets the web server would, minus the
+            # DATABASE_URL override above.
```
fortmarek, line 102: I'm slowly working on removing the dependency on the phoenix secrets, but step-by-step.

server/priv/secrets/can.yml.enc
fortmarek, line 1: removing the processor-related secrets.

infra/helm/tuist/values-managed-production.yaml
```diff
@@ -90,6 +88,28 @@ objectStorage:
       xcodeCache: tuist-xcode-cache
       registry: tuist-registry
 
+processor:
+  # Supabase production project (yltsvvtmlktxdgpcbylg) — dedicated pooler on
+  # port 6543 (transaction mode). Same hostname as the direct Postgres
+  # endpoint, just a different port. Pooler creds are the `tuist_processor`
+  # role provisioned via infra/supabase/tuist-processor-role.sql.
+  database:
+    host: db.yltsvvtmlktxdgpcbylg.supabase.co
```
pepicrft, line 97: Do we want to have these values here? I'd expect them to be in some kind of values file.

fortmarek: This is the values file? 🤔 

where else would you put this? The supabase reference is not really private.

fortmarek: This is the values file? 🤔 

where else would you put this? The supabase reference is not really private.

pepicrft:
One naming thought: instead of three booleans (`TUIST_PROCESSOR_MODE`, `TUIST_WEB`, `TUIST_DELEGATE_PROCESS_BUILD`), I'd consider a single `TUIST_MODE=web|processor` enum.

Pros: mutually exclusive by construction, self-documenting, and easy to grow (e.g. `scheduler`, `ingest` later) without piling on more flags. `TUIST_WEB=0` becomes redundant since processor mode implies no Phoenix listener.

I'd keep `TUIST_DELEGATE_PROCESS_BUILD` separate though, it's a different axis (cluster topology, not pod role). Maybe rename to something like `TUIST_DROP_QUEUES=process_build` if you want it to feel more general.

Not blocking, just a thought before this lands and the names get sticky.

pepicrft (approved):
Approving. A few non-blocking nits I'd consider as follow-ups (or fold in here if quick):

1. The `preStop: sleep 10` on the processor pod doesn't really do anything useful since there's no Service to drain. I'd either drop it or replace with something Oban-aware (e.g. pause queues so in-flight finishes but no new claims).

2. `oban_queues` in [server/config/runtime.exs](server/config/runtime.exs) duplicates the queue list across two branches. Adding a new queue means remembering to update both. Easy to derive from a base list + a conditional `process_build` entry.

3. `TUIST_DATABASE_POOLED` could be derived from the URL port (6543 vs 5432) instead of an extra env var. The PR description even mentioned this was the original plan. Less to misconfigure.

4. Worth a quick sanity check that the processor pod winning Oban leader election (Pruner, Lifeline) is fine. Cron is already gated, but the others will run wherever the leader lands.

Nice work overall, the description is genuinely useful and the SQL bootstrap + GUC trick is well thought out.

pepicrft (approved):

fortmarek:
On nit 3 — leaning toward keeping `TUIST_DATABASE_POOLED` explicit. Port 6543 is Supabase Supavisor's convention specifically; PgBouncer defaults to 6432, RDS Proxy uses 5432, and any other transaction-mode pooler is free to pick whatever. A self-hoster putting PgBouncer on its default 6432 would silently get `:named` prepares + `tcp_keepalives_*` startup parameters from a port heuristic, which is exactly the failure shape the conditional is meant to avoid.

Saying "this URL points at a pooler" is a behavioral fact the operator knows; saying "port 6543 ⇒ pooler" is a vendor assumption that ages badly. Happy to revisit if we ever standardize on one pooler across self-host + managed.

The other three nits are folded in (commit 6df568cd36): preStop dropped, oban_queues derived from a base list, and the Cron-on-processor-leader bug is fixed (it actually was a bug — previously if the processor won leader the empty crontab would fire and no scheduled jobs would be enqueued cluster-wide).

Plus the comment-thread suggestion: TUIST_PROCESSOR_MODE + TUIST_WEB collapsed into TUIST_MODE=web|processor.

PR #10429, opened 2026-04-24
feat(server,cli): add "Skip" quarantine mode for test cases

fortmarek:
Implements the "Skip" quarantine mode described in the [Mute & Skip quarantine modes RFC](https://community.tuist.dev/t/adding-mute-and-skip-quarantine-modes/961), building on the automations engine landed in #10232.

Quarantined test cases now have two distinct modes:

- **Mute** — keep running, mask failures. Good for "watch this one but don't fail CI". This is the existing behavior; the state is just renamed end-to-end (`is_quarantined` → `state: "muted"`).
- **Skip** — exclude from execution via `-skip-testing`. Good for broken/slow/persistently-flaky tests that shouldn't even run.

### Server

- New `"skipped"` value on `test_cases.state` (LowCardinality string column — no migration needed) plus `ChangeStateAction` / `Automations.Alerts.Alert` accept it.
- `Tuist.Tests.determine_test_case_events/2` emits `:skipped` / `:unskipped` transitions in addition to the existing muted/unmuted ones. Legacy `quarantined` / `unquarantined` event types are also matched where we pull "who quarantined this" info so pre-automations data still renders correctly.
- `list_quarantined_test_cases/2` and `get_quarantine_actors/1` include both `muted` and `skipped` tests; the returned `QuarantinedTestCase` now carries `state` so the UI can tag each row.
- OpenAPI: `TestCase.state` enum grows to `enabled` | `muted` | `skipped`, `listTestCases?state=` filter accepts the new value, `is_quarantined` (deprecated) is `true` for both `muted` and `skipped`.
- MCP `list_test_cases` state enum + description updated.
- LiveView:
  - Test case detail page: state dropdown gains a Skipped option, badges for muted/skipped render distinctly, history timeline handles the new event types.
  - Test cases list: Muted / Skipped badges.
  - Quarantined tests page: new **Mode** column with Muted/Skipped badges + Mode filter (+ `state_filter` plumbed into `Tests.list_quarantined_test_cases`).
  - Automations editor: Skipped option in both trigger and recovery change_state dropdowns.
- Seeds spread flaky test cases across enabled/muted/skipped (~30/40/30) so the new state is exercisable locally.

### CLI

- OpenAPI client regenerated off the updated spec; the `state` query param on `listTestCases` is plumbed through `ListTestCasesService`.
- New `TestCaseListService` exposes `listTestCases(config:state:)` returning `[TestIdentifier]` for a single state. `TestQuarantineService` shrinks to its `markQuarantinedTests` / `onlyQuarantinedTestsFailed` post-processing helpers.
- `TestService.run` and `XcodeBuildTestCommandService.run` (the `tuist xcodebuild test` passthrough) each fire two concurrent `listTestCases(state:)` calls (`.muted` / `.skipped`). Skipped identifiers are appended to xcodebuild's `-skip-testing` args; muted identifiers feed post-run masking. `--skip-quarantine` short-circuits the fetch entirely at the call site.
- Gradle plugin (`gradle/src/main/kotlin/dev/tuist/gradle/`): `TuistTestQuarantineService` returns a `QuarantinedTests` struct splitting `muted` / `skipped`. The plugin filters skipped tests out via `Test.filter.excludeTestsMatching("Class.method")` in `doFirst`, and feeds only muted tests into the existing post-run failure mask.
- `TestCaseListCommandService` / `TestCaseShowCommandService` print a `state` column/field, `TestCaseEventsCommandService` handles the new event-type enum cases.

### Tests

- Server: 26 new/updated tests across `tests_test.exs`, `test_cases_controller_test.exs`, `quarantined_tests_live_test.exs`, `test_case_live_test.exs`, `change_state_action_test.exs`, `alert_test.exs`.
- CLI: new `TestCaseListServiceTests` covering the state-filter pass-through; `TestServiceTests.test_run_passes_skipped_quarantined_tests_as_skip_testing` verifying skipped identifiers reach xcodebuild's skip args; `--skip-quarantine` test asserts the fetch is short-circuited; `ChangeStateAction` tests extended to cover skipped.
- Gradle: `TuistTestQuarantineTest` covers the muted/skipped split, the `null`-state back-compat fallback to muted, and the `QuarantinedTests.all` union.

### End-to-end smoke (server → CLI → xcodebuild)

Verified locally by wiring a real round-trip:

**Setup**

1. Built the CLI from this branch: `xcodebuild build -workspace Tuist.xcworkspace -scheme tuist ...` → `.../Tuist-*/Build/Products/Debug/tuist`.
2. Started the Phoenix server (`mise run dev` → `http://localhost:8977`) and seeded the dev DB so ~42 flaky test cases land in `state: "skipped"` and ~51 in `state: "muted"`.
3. Minted a project-scoped token for the seeded `tuist/tuist` project via IEx (`Tuist.Projects.create_project_token/1`).
4. Wrote a minimal Tuist fixture at `/tmp/SkipModeFixture` with `fullHandle: "tuist/tuist"`, `url: "http://localhost:8977"`, an `App` macOS app + `APITests` unit-test target, then `tuist generate`.

**Run**

```sh
CI=1 TUIST_TOKEN=$(cat /tmp/tuist-token) TUIST_URL=http://localhost:8977 \
  tuist test APITests --verbose
```

**Observations** (captured from the `--verbose` log)

1. Two state-filtered API requests against the live Phoenix, fired concurrently:
   - `GET /api/projects/tuist/tuist/tests/test-cases?state=muted&page_size=500&page=1` → 51 rows (`is_quarantined: true`, `state: "muted"`).
   - `GET /api/projects/tuist/tuist/tests/test-cases?state=skipped&page_size=500&page=1` → 42 rows (`is_quarantined: true`, `state: "skipped"`).
2. CLI log from `TestService` (combined log line at the call site after the two state-filtered fetches):
   ```
   Found 93 quarantined test(s): 51 muted, 42 skipped
   ```
3. xcodebuild invocation captured verbatim from `XcodeBuildController` debug log — note the `-skip-testing` arguments only for skipped identifiers:
   ```
   /usr/bin/xcrun xcodebuild test -scheme APITests -workspace /private/tmp/SkipModeFixture/SkipModeFixture.xcworkspace \
     -destination platform=macOS,arch=arm64 \
     -resultBundlePath /Users/.../result-bundle \
     -skip-testing APITests/CacheTests/testPerformance \
     -skip-testing APITests/ViewModelTests/testPerformance \
     -skip-testing SecurityTests/CacheTests/testUserLogout \
     -skip-testing SecurityTests/NetworkTests/testNetworkRequest \
     -skip-testing PerformanceTests/CacheTests/testErrorHandling \
     … 37 more …
   ```
4. Unique `-skip-testing` arg count: **42** — exactly matches the 42 server-side `state: "skipped"` rows. The 51 muted tests are absent from the skip list (they still run under mute semantics).

This closes the server → HTTP → CLI → xcodebuild loop end-to-end for skip mode.

### How to test locally

1. `mise run db:setup && mise run db:seed` — the updated seeds now populate ~30% of flaky test cases as `skipped` so there's data to play with.
2. Run `mise run dev`, open any project's **Tests → Quarantined Tests** page. Confirm the new Mode column shows both Muted and Skipped badges and that the Mode filter works.
3. On a test case detail page, use the state dropdown to flip between Enabled → Muted → Skipped. History timeline should show `skipped`/`unskipped` events with a distinct icon/color.
4. In **Settings → Automations**, create a change_state automation and confirm the Skipped option is available for both trigger and recovery.
5. CLI: `tuist test ...` against a project that has skipped tests — xcodebuild should be invoked with `-skip-testing <target>/<suite>/<method>` for each skipped test. Muted tests still run and their failures are masked as before.


cli/Sources/TuistKit/Services/TestCaseListService.swift
```diff
@@ -0,0 +1,52 @@
...
+    ) {
+        self.listTestCasesService = listTestCasesService
+    }
+
+    func listTestCases(
+        fullHandle: String,
+        serverURL: URL,
+        state: Operations.listTestCases.Input.Query.statePayload
+    ) async throws -> [TestIdentifier] {
+        let response = try await listTestCasesService.listTestCases(
```
pepicrft, line 35: This still goes through `Tests.list_test_cases/3`, which applies the default 14-day active window whenever there is no name filter or `active_period`. That is fine for `muted` tests because they still run and refresh `last_ran_at`, but `skipped` tests intentionally never run, so they age out of this fetch after two weeks and `tuist test` / `tuist xcodebuild test` start executing them again. We need either a quarantine-specific fetch path or to bypass the active-window filter when fetching skipped tests for execution control.

gradle/src/main/kotlin/dev/tuist/gradle/TuistTestQuarantine.kt
```diff
@@ -47,24 +83,24 @@ class TuistTestQuarantineService(
...
+            QuarantinedTests.EMPTY
         }
 
-        cachedExclusions = result
+        cached = result
         return result
     }
 
-    private fun fetchQuarantinedTests(): Map<String, List<TestIdentifier>> {
+    private fun fetchQuarantinedTests(): QuarantinedTests {
```
pepicrft, line 103: The Gradle path still calls `/tests/test-cases?quarantined=true`, and the API maps that to `state in ["muted", "skipped"]` before going through the same 14-day active-window logic in `Tests.list_test_cases/3`. Once a skipped test has been excluded long enough that `last_ran_at` is older than the cutoff, it disappears from this response and Gradle stops excluding it even though its server state is still `skipped`.

pepicrft (approved):
Left some comments.

cli/Sources/TuistKit/Services/TestCaseListService.swift
fortmarek: Fixed in `3f03693a4e` server-side: `Tests.list_test_cases/3` now skips the 14-day `last_ran_at` cutoff when the filter list contains `state == "muted"`, `state == "skipped"`, or the legacy `state IN ["muted","skipped"]` (i.e. `?quarantined=true`). `?state=enabled` and `?quarantined=false` keep the window. New regression test in `tests_test.exs` seeds tests with `ran_at = 30 days ago`, marks one skipped + one muted, asserts the default fetch returns nothing while each state filter still surfaces them.

gradle/src/main/kotlin/dev/tuist/gradle/TuistTestQuarantine.kt
fortmarek: Same fix in `3f03693a4e` covers the Gradle path. The `?quarantined=true` query (which the controller still maps to `state IN ["muted","skipped"]`) now bypasses the active-window filter, so skipped tests stay in the response indefinitely while their server state is `skipped`. No CLI/Gradle plugin change required.

PR #10440, opened 2026-04-24
feat(server): replace Alloy chart with Grafana k8s-monitoring wrapper

fortmarek:
## Summary

Swaps the hand-rolled `infra/helm/alloy/` chart for a thin wrapper around the upstream [`grafana/k8s-monitoring`](https://github.com/grafana/k8s-monitoring-helm) v4.0.3, and wires the install + cutover into CI so no manual `helm` commands are required. Goal: get the managed clusters' full Kubernetes picture (nodes, workloads, events, kube-state-metrics, cAdvisor) visible in Grafana Cloud's **Observability → Kubernetes** app, so we stop reaching for Lens for cluster inspection.

**Live dashboard:** [Grafana Cloud → Kubernetes → Overview](https://tuist.grafana.net/a/grafana-k8s-app/home?from=now-1h&to=now&refresh=1m&var-cluster=%24__all&var-namespace=%24__all) — `tuist-staging` is live there now; `tuist-canary` and `tuist-production` will appear automatically after this PR merges and the cascade rolls each cluster.

### What changes operationally

| Signal | Before (`helm/alloy`) | After (`helm/k8s-monitoring`) |
|---|---|---|
| Server app metrics | Static scrape of `tuist-tuist-server:9091` | Auto-discovered via existing `prometheus.io/scrape` annotation |
| Server traces | OTLP gRPC :4317 → Tempo | Same, via `alloy-receiver` |
| Server logs | stdout tailed via kubelet API → Loki | stdout tailed via DaemonSet `/var/log/pods` → Loki (narrower blast radius) |
| **kube-state-metrics** | — | Deployed + scraped |
| **node-exporter** | — | DaemonSet (node CPU/mem/disk/net) |
| **kubelet + cAdvisor** | — | Scraped (container resource usage) |
| **Kubernetes Events** | — | Streamed to Loki as structured logs |

### Contract preserved

- Same Grafana Cloud Prometheus / Loki / Tempo stack (URLs + numeric usernames unchanged).
- Same 1Password-via-ESO token sync: three items (`PROMETHEUS_TOKEN`, `LOKI_TOKEN`, `TEMPO_TOKEN`), same `ClusterSecretStore` named `onepassword`.
- Same OTLP gRPC receiver on :4317 for server traces. The Service name changes to `k8s-monitoring-alloy-receiver.observability.svc.cluster.local:4317` — the three managed server values files are updated to match.
- `env` label kept on metrics + logs (via `extraLabels`); added as `deployment.environment` resource attribute on traces (OTLP doesn't take `extraLabels`).

### CI automation

Two new jobs in `.github/workflows/server-deployment.yml`, both sharing the `server-deploy-<env>` concurrency lane with `deploy` so two cascades can't interleave and pair a new chart install with an old server rollout:

- `observability-install` — runs in parallel with `build`, gates `deploy`. Executes `helm dependency update` + `helm upgrade --install k8s-monitoring`. Idempotent; ~10 s no-op on subsequent runs.
- `observability-cleanup` — runs after a successful server deploy and does `helm uninstall alloy --ignore-not-found`. One-shot per cluster; no-op afterwards. Keeping it in CI rather than leaving the cleanup to a human closes the window where both charts run side-by-side duplicating metrics + logs. After all three clusters cut over, this job + `infra/helm/alloy/` get deleted together in a follow-up PR.

The cascade workflow's path filter also now includes `infra/helm/k8s-monitoring/**` so chart edits re-run canary → tests → prod.

### Worker bump

Staging + canary `workerMachineTypeHcloud` bumped `cpx22` (4 GB) → `cpx31` (8 GB). The extra Alloy DaemonSets + kube-state-metrics + node-exporter left no memory headroom on `cpx22` for a rolling server update to schedule a fresh pod alongside the draining one. Production stays on `ccx23` (16 GB). ~€20/mo increase across staging + canary.

### Reviewer notes

- Self-hosters are untouched — this chart is managed-cluster-only. The embedded `observability.enabled` path in `infra/helm/tuist/` still exists for self-host.
- Old `infra/helm/alloy/` stays in-tree (marked deprecated in its README) so a rollback is one git revert away.
- `Chart.lock` is committed; `charts/*.tgz` is gitignored (CI runs `helm dependency update` before install). 276 KB tarball not worth vendoring.
- Four Alloy collectors: `alloy-metrics` (clustered DaemonSet, scrapes), `alloy-logs` (DaemonSet, tails /var/log/pods), `alloy-receiver` (1-replica Deployment, OTLP gateway), `alloy-singleton` (1-replica Deployment, cluster events). Managed by the upstream alloy-operator — Services get created at runtime, not by `helm template`.

## How to test locally

```bash
helm dependency update infra/helm/k8s-monitoring
helm lint infra/helm/k8s-monitoring -f infra/helm/k8s-monitoring/values-staging.yaml
helm template k8s-monitoring infra/helm/k8s-monitoring -n observability \
  -f infra/helm/k8s-monitoring/values-staging.yaml
```

All three env overlays (`staging` / `canary` / `production`) lint clean.

## Rollout plan

**Pre-merge** (already done): applied the updated Cluster CRs to the Syself management cluster; CAPI is rolling staging + canary workers cpx22 → cpx31. Staging k8s-monitoring chart was also hand-installed during PR validation.

**Merge** handles canary + prod automatically via the cascade:

1. Push to main triggers the cascade workflow.
2. Canary leg: `observability-install` brings up `k8s-monitoring` on canary; `deploy` rolls the server with the new OTLP endpoint; `observability-cleanup` removes the old `alloy` release.
3. Acceptance tests run.
4. Production leg: same three steps against prod.

Staging is already past step 1 + 2 but still has the legacy `alloy` release running — the next deploy targeting staging will pick it up and clean it via the new `observability-cleanup` job, or you can run `helm uninstall alloy -n observability` against the staging kubeconfig now to close the dup-telemetry window immediately.

After all three clusters are cut over, a follow-up PR deletes `infra/helm/alloy/` and the `observability-cleanup` job together.


infra/helm/k8s-monitoring/values.yaml
```diff
@@ -0,0 +1,179 @@
...
+    # Clustering lets multiple alloy-metrics replicas hash-partition the
+    # scrape targets, so adding replicas linearly scales capacity. Required
+    # by the clusterMetrics / hostMetrics / annotationAutodiscovery features.
+    alloy-metrics:
+      presets: [clustered]
+    # filesystem-log-reader mounts /var/log so the DaemonSet can tail pod
+    # logs from disk (matches the podLogsViaLoki feature).
+    alloy-logs:
+      presets: [filesystem-log-reader]
+    alloy-singleton:
```
pepicrft, line 179: `clusterEvents` is assigned to `alloy-singleton`, but this collector only sets `alloy: {}`. When I rendered the chart locally, the resulting Alloy CR had no `spec.controller` override, so it falls back to the Alloy chart defaults instead of a 1-replica singleton. That means every node can watch and forward the same Kubernetes Events, duplicating them in Loki. This probably wants the upstream `singleton` preset, or an explicit `controller.type: deployment` with `replicas: 1`.

infra/helm/alloy/README.md
```diff
@@ -1,4 +1,19 @@
-# Grafana Alloy for the Tuist managed cluster
+# Grafana Alloy for the Tuist managed cluster (DEPRECATED)
+
+**Superseded by [`infra/helm/k8s-monitoring/`](../k8s-monitoring/README.md).** The new chart covers the same telemetry paths (Grafana Cloud Prometheus / Loki / Tempo, same ESO token sync) plus cluster / pod / node metrics and events, so the Grafana Cloud Kubernetes app works out of the box.
+
+## Cutover / cleanup
+
+Cleanup is deliberately manual — `helm uninstall` is destructive and the cutover per cluster is a one-time event, so it doesn't belong in the every-deploy CI workflow. Once the new receiver is live on a cluster and the server is pushing traces to it, drop the old release:
```
pepicrft: The cutover plan keeps the legacy `alloy` release running until a later manual uninstall, but that old chart still scrapes `/metrics` and tails pod logs. Once `k8s-monitoring` is installed it starts collecting the same server metrics and logs immediately, so the rollout duplicates telemetry until someone performs the cleanup step. That will skew dashboards/alerts and likely increase log volume during the migration window.

.github/workflows/server-deployment.yml
```diff
@@ -151,11 +154,50 @@ jobs:
...
+    # Installs / upgrades the grafana/k8s-monitoring wrapper chart ahead of
+    # the server rollout so the Alloy OTLP receiver is ready before the
+    # server starts pushing spans to it. Idempotent: on every run it's a
+    # no-op if Chart.lock + values haven't changed.
+    name: Observability chart (${{ inputs.environment }})
+    needs: plan
+    runs-on: namespace-profile-default
+    environment:
+      name: server-k8s-${{ needs.plan.outputs.environment_key }}
+    concurrency:
```
pepicrft, line 172: I believe splitting `observability-install` and `deploy` into different concurrency groups lets them interleave across SHAs. Two queued runs can then produce `observability(B) -> deploy(A)`, which pairs an older server config with a newer monitoring chart. Since this PR changes the OTLP endpoint and the chart in lockstep, I think both jobs should stay in the same concurrency lane.

pepicrft (changes requested):
I found a couple of rollout issues that look worth fixing before merge: the cluster-events collector is documented as singleton but the rendered Alloy CR doesn't actually force singleton behavior, the migration plan leaves both telemetry pipelines active at once, and the new observability/app deploy lanes can interleave across SHAs. I left inline comments on the specific lines.

pepicrft (approved):

PR #10446, opened 2026-04-24
feat(kura): add warm rollout and rollback primitives

pepicrft:
## Why
Kura needs warm in-place rollouts for PVC-owning `StatefulSet` pods.

- Public reads are local-only today, so routing traffic to a half-warm pod creates avoidable misses.
- We cannot assume every client retries cleanly during shutdown.
- We must preserve `one Kura process = one PVC = one writer`.

## What This PR Implements
This PR adds the runtime and operational primitives needed for warm rollouts and safe single-ordinal rollback.

- explicit traffic lifecycle: `joining -> serving -> draining`
- single-writer fencing in the data directory
- public readiness that stays false until discovery and bootstrap settle
- drain-first shutdown with a shared deadline for public HTTP, gRPC, and internal HTTP
- a rollout status surface that exposes the data needed to gate promotions
- a generic rollout gate that is not coupled to Kubernetes
- a Helm adapter for partitioned `StatefulSet` rollouts
- an adjacent-version compatibility and rollback harness on the same persistent volume

## Runtime
- Kept `/up` as liveness and made `/ready` the public readiness signal.
- Added `/status/rollout` as a machine-oriented rollout status endpoint with generation, state, ring size, bootstrap progress, inflight work, outbox depth, memory pressure, and FD timeout counters.
- Added `RuntimeState` and `DataDirLock` to track traffic state, inflight work, and enforce single-writer startup.
- Readiness now depends on writer-lock ownership, initial discovery, bootstrap completion for the current peer set, and a short settle window.
- Membership generation changes move a node back out of `serving` until it reconverges.
- Public HTTP rejects new requests while draining and closes HTTP/1.1 connections.
- gRPC rejects new RPCs while draining and uses bounded connection age to help long-lived clients move off the node.
- Shutdown now uses one shared drain deadline and waits for inflight HTTP and gRPC work to complete before forcing task shutdown.

## Rollout Gating
- Added a generic rollout gate in `ops/rollout/gate.sh`.
- The gate depends only on Kura's rollout status endpoint, not on Kubernetes probes or Prometheus text parsing.
- Promotion now requires every node to report the same membership generation, every node back in `serving`, expected ring size cluster-wide, zero bootstrap inflight peers, outbox depth near baseline, no critical memory pressure, and no new FD timeout activity.
- Added a Kubernetes adapter in `ops/helm/kura/rollout.sh` that stages revisions with `StatefulSet` partitions and delegates health gating to the generic rollout gate.
- The Kubernetes adapter now only handles revision rollout and transport to pods; it does not own rollout semantics.

## Helm
- Switched readiness probing to `/ready` and kept liveness on `/up`.
- Added a startup probe and drain-on-`preStop` behavior.
- Derived `terminationGracePeriodSeconds` from the app's own drain timeout plus small lifecycle buffers.
- Defaulted PVC access mode to `ReadWriteOncePod`, with `ReadWriteOnce` documented as the fallback when the CSI driver requires it.

## Compatibility And Rollback Validation
- Added `test/e2e/kura_compatibility_rollout.sh`.
- The harness validates `PREVIOUS_REF -> HEAD -> PREVIOUS_REF` on the same persistent Docker volumes.
- Today that validation is focused on the artifact CAS path, so it proves the adjacent-version rollback contract there without claiming a full protocol matrix.

## What This PR Does Not Do
- It does not add a rollout controller or operator.
- It does not add peer read-through for public cache misses.
- It does not change the compatibility contract to allow breaking wire or storage changes.
- It does not try to model every production environment; the shell helpers are operational building blocks, not a full deployment system.

## Testing
- `cargo test`
- `helm lint ops/helm/kura`
- `helm template kura ops/helm/kura --namespace kura`
- `helm template kura ops/helm/kura --namespace kura --set updateStrategy.rollingUpdatePartition=2`
- `bash -n ops/rollout/gate.sh`
- `bash -n ops/helm/kura/rollout.sh`
- `bash -n test/e2e/kura_compatibility_rollout.sh`

### DIFF
--- a/kura/AGENTS.md
+++ b/kura/AGENTS.md
@@ -3,6 +3,7 @@
 This node covers the `kura/` workspace, a Rust service for low-latency cache meshes that replicate artifacts and metadata across peer nodes.
 
 ## Key Boundaries
+- High-level architecture overview: `docs/architecture.md` — start here when onboarding or reasoning about how subsystems interact
 - Entry points: `src/main.rs`, `src/app.rs`
 - Public HTTP and gRPC surfaces: `src/http.rs`
 - Storage, metadata, and replication state: `src/store.rs`, `src/state.rs`
@@ -20,5 +21,6 @@ This node covers the `kura/` workspace, a Rust service for low-latency cache mes
 
 ## Maintenance Notes
 - Keep `README.md` aligned with any protocol, configuration, or deployment changes
+- Keep `docs/architecture.md` in sync when changing how subsystems fit together (storage planes, replication model, traffic lifecycle, rollouts, observability surface)
 - When changing cache protocol behavior, update the relevant shellspec coverage under `spec/e2e/`
 - Keep Helm and local observability assets in `ops/` in sync with runtime configuration changes
--- a/kura/AGENTS.md
+++ b/kura/AGENTS.md
@@ -10,6 +10,7 @@ This node covers the `kura/` workspace, a Rust service for low-latency cache mes
 - Observability and analytics: `src/metrics.rs`, `src/telemetry.rs`, `src/analytics.rs`
 - Peer TLS support: `src/peer_tls.rs`
 - Operational assets: `docker-compose.yml`, `ops/`, `test/e2e/`, `spec/e2e/`
+  - See `ops/AGENTS.md` for Helm, rollout helpers, and observability config boundaries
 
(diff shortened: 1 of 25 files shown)


fortmarek (approved):
Aligned with the architecture 🙌

PR #10449, opened 2026-04-24
chore(server): bump XCLogParser to 0.2.47

fortmarek:
## Summary

Bumps \`MobileNativeFoundation.XCLogParser\` from \`0.2.46\` → \`0.2.47\`.

\`0.2.47\` ships [MobileNativeFoundation/XCLogParser#249](https://github.com/MobileNativeFoundation/XCLogParser/pull/249), which fixes a long-standing bug: \`addCompilationTimesToTarget\` was producing **negative** \`compilationDuration\` values for targets in incremental Xcode builds. Xcode reuses substep entries from prior build sessions for files that don't need recompilation; those substeps keep their original (older) timestamps and aren't flagged \`wasFetchedFromCache\`, so the latest substep's \`compilationEndTimestamp\` could pre-date the target's \`startTimestamp\`.

Downstream effect on the server: those negative \`Int\` values were sent to ClickHouse's \`UInt64 build_targets.compilation_duration\` column, where 2's-complement underflow turned them into huge positive numbers. The Build Insights UI then rendered them as e.g. \`5,124,095,576,030h 7m 41s\` (reported by Vinted: see \`a62b41e1-d59b-4b70-8c67-6b1eb3dc569a\`). After the bump, affected targets will report \`0ms\` (correct: no real compile work in this session) instead of the corrupted value.

Also adds an assertion in [\`XCActivityLogParserTests\`](processor/native/xcactivitylog_nif/Tests/XCActivityLogParserTests/XCActivityLogParserTests.swift:51) that \`target.compilation_duration >= 0\` as a guard against future regressions.

## Test plan

- [x] \`swift package resolve --replace-scm-with-registry\` succeeds in both \`Package.swift\` files.
- [x] \`swift test --filter XCActivityLogParserTests\` passes (12/12) against \`0.2.47\`.

## Follow-up

Existing corrupted rows in ClickHouse (\`build_targets.compilation_duration\` near \`UInt64.max\`) are not touched by this PR. Backfill or display-side filtering can be done separately.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "1f6d96b59d4ac0c449d7302ca00a15240d8bc98a9c2e8b6100f7ea87bf0ece99",
+  "originHash" : "ff91f77210b8991b3fa1311d8c0904413d69d86abdceada53823faec67d87d14",
   "pins" : [
     {
       "identity" : "1024jp.gzipswift",
@@ -407,7 +407,7 @@
       "kind" : "registry",
       "location" : "",
       "state" : {
-        "version" : "0.2.46"
+        "version" : "0.2.47"
       }
     },
     {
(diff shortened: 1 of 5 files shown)


pepicrft (approved):

pepicrft:
@fortmarek, the failing acceptance tests are odd. The runs on the dashboard show as passing with "Unknown" scheme, even though they are reported as failing on the client.

fortmarek:
> The runs on the dashboard show as passing with "Unknown" scheme, even though they are reported as failing on the client.

This is due to failing remote processing of xcresults, will have a fix for that soon.

PR #10450, opened 2026-04-24
feat(cli,server): add tuist test case update command

justine-acorns:
## Summary

Adds a `tuist test case update` CLI command backed by a new `PATCH /projects/:account/:project/test-cases/:id` server endpoint. The command accepts either a test case UUID or a `Module/Suite/TestName` identifier and can change the test case's `state` (via `--state`) and/or its flaky flag (via `--flaky`/`--no-flaky`). At least one of the two must be provided.

Primary motivation: give operators a one-shot, scriptable way to mutate a test case; in particular to unquarantine tests that were flagged as flaky by auto-quarantine but turned out to be false positives, and to toggle the flaky flag from CI without going through the dashboard.

```bash
tuist test case update AppTests/LoginTests/testHappyPath --state enabled
tuist test case update 123e4567-e89b-12d3-a456-426614174000 --state muted
tuist test case update AppTests/testFoo --no-flaky --json
tuist test case update AppTests/testFoo --state enabled --flaky
```

## Changes

- **server**: new `PATCH /api/projects/:account_handle/:project_handle/tests/test-cases/:test_case_id` endpoint on `TestCasesController`. Accepts optional `state` and `is_flaky`; returns 400 on empty body or invalid state, 404 when the test case is missing or belongs to another project (including the TOCTOU case where it disappears between lookup and update), 403 when the caller lacks project write access. Response mirrors the existing test case payload shape (`id`, `name`, `module`, `suite`, `is_flaky`, `is_quarantined`, `state`, `url`) and records the corresponding `muted`/`unmuted`, `skipped`/`unskipped`, `marked_flaky`/`unmarked_flaky` events on transitions. Scoped to project `:user` role to mirror the dashboard's existing mute/flaky permissions.
- **cli**: new `TestCaseUpdateCommand` and `TestCaseUpdateCommandService`, plus `UpdateTestCaseService` in `TuistServer`. Input supports both test case UUID and `Module[/Suite]/TestName`; reuses the existing `TestCaseIdentifier` parsing pattern from `test case show`. Flags: `--state`, `--flaky/--no-flaky`, `--project`, `--path`, `--json`.
- **OpenAPI**: hand-edited client + types + spec to register the new `updateTestCase` operation and its `Ok`/`BadRequest`/`Forbidden`/`NotFound` responses. `state` is modeled as an open string so new server-side states (e.g. `skipped` from #10429) don't break clients pinned to an older spec.
- **tests**: server controller tests cover enabled -> muted, muted -> enabled, marking flaky, state + is_flaky together, 400 on empty body, 400 on invalid state, 404 on missing test case, 404 on cross-project access, 404 on the TOCTOU path, and 403 on unauthorized. CLI service tests cover missing full handle, missing update fields, UUID + JSON, UUID + text, name lookup, explicit `--project` flag, both `--flaky` and `--no-flaky`, and invalid identifier.

## Test plan

- [x] `xcodebuild test ... TuistTestCommandTests/TestCaseUpdateCommandServiceTests` passes
- [x] `xcodebuild test ... TuistTestCommandTests TuistKitTests/TestQuarantineServiceTests` passes (affected modules)
- [x] `mix format --check-formatted` clean on touched server files
- [x] CI: SwiftPM Build, Unit Tests, Linux Build, Linux Unit Tests, Server Test all green
- [ ] Smoke test against staging: update a test case by UUID and by name; toggle `state` and `is_flaky` independently and together

### DIFF
--- a/cli/Sources/TuistTestCommand/TestCaseUpdateCommand.swift
+++ b/cli/Sources/TuistTestCommand/TestCaseUpdateCommand.swift
@@ -1,0 +1,72 @@
+import ArgumentParser
+import Foundation
+import Path
+import TuistEnvKey
+import TuistNooraExtension
+import TuistServer
+
+public struct TestCaseUpdateCommand: AsyncParsableCommand, NooraReadyCommand {
+    public init() {}
+    public static var configuration: CommandConfiguration {
+        CommandConfiguration(
+            commandName: "update",
+            abstract: "Updates a test case."
+        )
+    }
+
+    @Argument(
+        help: "The test case identifier. Either a UUID or the format Module/Suite/TestCase (or Module/TestCase).",
+        envKey: .testCaseUpdateIdentifier
+    )
+    var testCaseIdentifier: String
+
+    @Option(
+        help: "The new state for the test case.",
+        envKey: .testCaseUpdateState
+    )
+    var state: ServerTestCaseState?
+
+    @Flag(
+        inversion: .prefixedNo,
+        exclusivity: .exclusive,
+        help: "Mark the test case as flaky (--flaky) or not flaky (--no-flaky).",
+        envKey: .testCaseUpdateFlaky
+    )
+    var flaky: Bool?
+
+    @Option(
+        name: [.customLong("project"), .customShort("P")],
+        help: "The full handle of the project. Must be in the format of account-handle/project-handle.",
+        envKey: .testCaseUpdateProject
+    )
+    var project: String?
+
+    @Option(
+        name: .shortAndLong,
+        help: "The path to the directory or a subdirectory of the project.",
+        completion: .directory,
+        envKey: .testCaseUpdatePath
+    )
+    var path: String?
+
+    @Flag(
+        help: "The output in JSON format.",
+        envKey: .testCaseUpdateJson
+    )
+    var json: Bool = false
+
+    public var jsonThroughNoora: Bool = true
+
+    public func run() async throws {
+        try await TestCaseUpdateCommandService().run(
+            project: project,
+            testCaseIdentifier: testCaseIdentifier,
+            state: state,
+            isFlaky: flaky,
+            path: path,
+            json: json
+        )
+    }
+}
+
+extension ServerTestCaseState: @retroactive ExpressibleByArgument {}
--- a/cli/Tests/TuistTestCommandTests/Services/TestCaseUpdateCommandServiceTests.swift
+++ b/cli/Tests/TuistTestCommandTests/Services/TestCaseUpdateCommandServiceTests.swift
@@ -1,0 +1,362 @@
+import Foundation
+import Mockable
+import Testing
+import TuistConfig
+import TuistConfigLoader
+import TuistEnvironment
+import TuistEnvironmentTesting
+import TuistNooraTesting
+import TuistServer
+
+@testable import TuistTestCommand
+
+struct TestCaseUpdateCommandServiceTests {
+    private let getTestCaseService = MockGetTestCaseServicing()
+    private let updateTestCaseService = MockUpdateTestCaseServicing()
+    private let serverEnvironmentService = MockServerEnvironmentServicing()
+    private let configLoader = MockConfigLoading()
+    private let subject: TestCaseUpdateCommandService
+
+    init() {
+        subject = TestCaseUpdateCommandService(
+            getTestCaseService: getTestCaseService,
+            updateTestCaseService: updateTestCaseService,
+            serverEnvironmentService: serverEnvironmentService,
+            configLoader: configLoader
+        )
+    }
+
+    @Test(.withMockedEnvironment()) func run_when_full_handle_is_not_passed_and_absent_in_config() async throws {
+        // Given
+        let tuist = Tuist.test(fullHandle: nil)
+        let directoryPath = try await Environment.current.pathRelativeToWorkingDirectory(nil)
+        given(configLoader).loadConfig(path: .value(directoryPath)).willReturn(tuist)
+        let serverURL = URL(string: "https://\(UUID().uuidString).tuist.dev")!
+        given(serverEnvironmentService).url(configServerURL: .value(tuist.url)).willReturn(serverURL)
+
+        // When/Then
+        await #expect(throws: TestCaseUpdateCommandServiceError.missingFullHandle, performing: {
+            try await subject.run(
+                project: nil,
+                testCaseIdentifier: "test-case-id",
+                state: .enabled,
+                isFlaky: nil,
+                path: nil,
+                json: false
+            )
+        })
+    }
+
+    @Test(.withMockedEnvironment()) func run_when_neither_state_nor_flaky_provided() async throws {
+        // When/Then
+        await #expect(throws: TestCaseUpdateCommandServiceError.missingUpdateFields, performing: {
+            try await subject.run(
+                project: nil,
+                testCaseIdentifier: "test-case-id",
+                state: nil,
+                isFlaky: nil,
+                path: nil,
+                json: false
+            )
+        })
+    }
+
+    @Test(
+        .withMockedEnvironment(arguments: ["--json"]),
+        .withMockedNoora
+    ) func run_with_json_output_and_uuid() async throws {
+        // Given
+        let fullHandle = "\(UUID().uuidString)/\(UUID().uuidString)"
+        let tuist = Tuist.test(fullHandle: fullHandle)
+        let directoryPath = try await Environment.current.pathRelativeToWorkingDirectory(nil)
+        given(configLoader).loadConfig(path: .value(directoryPath)).willReturn(tuist)
+        let serverURL = URL(string: "https://\(UUID().uuidString).tuist.dev")!
+        given(serverEnvironmentService).url(configServerURL: .value(tuist.url)).willReturn(serverURL)
+        let updated = ServerUpdatedTestCase.test(
+            module: .test(name: "AuthTests"),
+            name: "testLogin()",
+            state: .enabled
+        )
+        given(updateTestCaseService).updateTestCase(
+            fullHandle: .value(fullHandle),
+            testCaseId: .value("test-case-id"),
+            state: .value(.enabled),
+            isFlaky: .value(nil),
+            serverURL: .value(serverURL)
+        ).willReturn(updated)
+
+        // When
+        try await subject.run(
+            project: nil,
+            testCaseIdentifier: "test-case-id",
+            state: .enabled,
+            isFlaky: nil,
+            path: nil,
+            json: true
+        )
+
+        // Then
+        let payload = try #require(decodeJSON(ui()))
+        #expect(payload["name"] as? String == "testLogin()")
+        #expect((payload["module"] as? [String: Any])?["name"] as? String == "AuthTests")
+        #expect(payload["state"] as? String == "enabled")
+    }
+
+    @Test(
+        .withMockedEnvironment(arguments: ["--json"]),
+        .withMockedNoora
+    ) func run_with_json_output_and_name_identifier() async throws {
+        // Given
+        let fullHandle = "\(UUID().uuidString)/\(UUID().uuidString)"
+        let tuist = Tuist.test(fullHandle: fullHandle)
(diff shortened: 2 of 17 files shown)


server/lib/tuist_web/controllers/api/test_cases_controller.ex
```diff
@@ -288,6 +289,140 @@ defmodule TuistWeb.API.TestCasesController do
...
+        if test_case.project_id == selected_project.id do
+          actor_id = Authentication.authenticated_subject_account(conn).id
+
+          {:ok, updated_test_case} = Tests.update_test_case(test_case_id, attrs, actor_id: actor_id)
+
+          json(conn, %{
+            id: updated_test_case.id,
+            name: updated_test_case.name,
+            module: %{
+              id: updated_test_case.module_name,
```
pepicrft: `module.id` is populated with `module_name`, so both `id` and `name` end up as the same string. The GET response returns the actual module id here — this PUT response will diverge. Worth fetching (or preloading) the module and using its real id (and same check for `build_suite/1` below).

server/lib/tuist_web/controllers/api/test_cases_controller.ex
```diff
@@ -288,6 +289,140 @@ defmodule TuistWeb.API.TestCasesController do
...
+  def update(
+        %{assigns: %{selected_project: selected_project}, params: %{test_case_id: test_case_id}, body_params: body_params} =
+          conn,
+        _params
+      ) do
+    attrs = Map.take(body_params, [:state, :is_flaky])
+
+    if map_size(attrs) == 0 do
+      conn
+      |> put_status(:unprocessable_entity)
```
pepicrft: This 422 isn't documented in the OpenAPI `responses` for `updateTestCase` (only 200/403/404 are listed). Either add it to the schema or rely on the existing 400 path from `CastAndValidate` for malformed bodies — the `state: "not-a-real-state"` test already exercises that route.

cli/Sources/TuistTestCommand/TestCaseUpdateCommand.swift
```diff
@@ -0,0 +1,72 @@
...
+import TuistEnvKey
+import TuistNooraExtension
+import TuistServer
+
+public struct TestCaseUpdateCommand: AsyncParsableCommand, NooraReadyCommand {
+    public init() {}
+    public static var configuration: CommandConfiguration {
+        CommandConfiguration(
+            commandName: "update",
+            abstract: "Updates a test case's state or flaky flag."
```
pepicrft: I'd not mention the specific thing that we are updating in the description, so that we can add other properties in the future and this description will still be relevant.

cli/Tests/TuistTestCommandTests/Services/TestCaseUpdateCommandServiceTests.swift
```diff
@@ -0,0 +1,355 @@
...
+            project: nil,
+            testCaseIdentifier: "test-case-id",
+            state: .enabled,
+            isFlaky: nil,
+            path: nil,
+            json: true
+        )
+
+        // Then
+        #expect(ui().contains("testLogin()"))
```
pepicrft: Let's confirm that what we get is a JSON.

pepicrft: I'd do that with all the tests in this suite that use `--json`.

server/lib/tuist/authorization.ex
```diff
@@ -486,6 +486,17 @@ defmodule Tuist.Authorization do
       allow([:authenticated_as_account, scopes_permit: "project:tests:read"])
       allow([:authenticated_as_account, scopes_permit: "project:tests:write"])
     end
+
+    action :update do
+      desc("Allows users of a project to update a test.")
```
pepicrft:
```suggestion
      desc("Allows users of a project to update a tests data.")
```

server/lib/tuist/authorization.ex
```diff
@@ -486,6 +486,17 @@ defmodule Tuist.Authorization do
...
+      allow([:authenticated_as_user, user_role: :user])
+
+      desc("Allows the admin of a project to update a test.")
```
pepicrft:
```suggestion
      desc("Allows the admin of a project to update tests data.")
```

server/lib/tuist/authorization.ex
```diff
@@ -486,6 +486,17 @@ defmodule Tuist.Authorization do
...
+      allow([:authenticated_as_user, user_role: :admin])
+
+      desc("Allows an account token with project:tests:write scope to update tests.")
```
pepicrft:
```suggestion
      desc("Allows an account token with project:tests:write scope to update tests data.")
```

server/lib/tuist/authorization.ex
```diff
@@ -486,6 +486,17 @@ defmodule Tuist.Authorization do
       allow([:authenticated_as_account, scopes_permit: "project:tests:read"])
       allow([:authenticated_as_account, scopes_permit: "project:tests:write"])
     end
+
+    action :update do
+      desc("Allows users of a project to update a test.")
+      allow([:authenticated_as_user, user_role: :user])
```
pepicrft, line 492: I'm not sure a non-admin should be allowed to update tests data. Thoughts @fortmarek?

pepicrft (approved):

pepicrft:
@justine-acorns I'd also add an MCP tool that mimics this for consistency.

justine-acorns:
> @justine-acorns I'd also add an MCP tool that mimics this for consistency.

Okay! I didn’t see other mutating functionality in the MCP, so I wasn’t sure if adding that would be outside the scope of what your team wanted the MCP to include. Can definitely add that!

server/lib/tuist_web/controllers/api/test_cases_controller.ex
justine-acorns: Looking at `index` and `show`, I see `id: test_case.module_name` as well. Is there a unique module ID available?

server/lib/tuist_web/controllers/api/test_cases_controller.ex
```diff
@@ -288,6 +289,130 @@ defmodule TuistWeb.API.TestCasesController do
...
+      forbidden: {"You don't have permission to update this resource", "application/json", Error}
+    }
+  )
+
+  def update(
+        %{assigns: %{selected_project: selected_project}, params: %{test_case_id: test_case_id}, body_params: body_params} =
+          conn,
+        _params
+      ) do
+    attrs = Map.take(body_params, [:state, :is_flaky])
```
pepicrft, line 399: This currently accepts `{}` and returns `200` with no change because we pass `Map.take(body_params, [:state, :is_flaky])` straight into `Tests.update_test_case/3` without checking that at least one field was set. The MCP tool and CLI both reject empty updates, so it would be good to keep the controller contract aligned here and fail with `400` instead.

server/lib/tuist/mcp/components/tools/update_test_case.ex
```diff
@@ -0,0 +1,172 @@
...
+        },
+        "identifier" => %{
+          "type" => "string",
+          "description" =>
+            "Test case identifier in Module/Suite/TestCase or Module/TestCase format. " <>
+              "Required when not using test_case_id. Must be combined with account_handle and project_handle."
+        },
+        "state" => %{
+          "type" => "string",
+          "enum" => ["enabled", "muted"],
```
pepicrft: This matches the controller on this branch, but it is narrower than the controller-level state model we use on `main` after #10429. Once that lands in the base, `skipped` is also a valid test case state, so `is_flaky`-only updates on a skipped test will still return `state: "skipped"`. I think we should either widen this surface on rebase or document why the MCP tool intentionally excludes skip-mode quarantine.

pepicrft (commented):
A couple of follow-ups here before this lands.

server/lib/tuist/mcp/components/tools/update_test_case.ex
justine-acorns: Oh awesome! That is a great addition. Thank you for the context. I'll rebase and make sure this aligns with the latest available states throughout.

server/lib/tuist_web/controllers/api/test_cases_controller.ex
```diff
@@ -296,6 +299,155 @@ defmodule TuistWeb.API.TestCasesController do
...
+      true ->
+        update_test_case(conn, selected_project, test_case_id, attrs)
+    end
+  end
+
+  defp update_test_case(conn, selected_project, test_case_id, attrs) do
+    case Tests.get_test_case_by_id(test_case_id) do
+      {:ok, test_case} ->
+        if test_case.project_id == selected_project.id do
+          actor_id = Authentication.authenticated_subject_account(conn).id
```
pepicrft, line 421: This strict pattern match crashes with `MatchError` (→ HTTP 500) when `Tests.update_test_case/3` returns anything other than `{:ok, _}`.

`Tests.update_test_case/3` already has at least one error path: it calls `get_test_case_by_id/1` internally with a `with` clause and short-circuits to `{:error, :not_found}`. Although the controller pre-fetches the test case just above, this leaves a TOCTOU window (test case deleted between the controller's check and the `Tests.update_test_case` call) and, more importantly, makes any future internal error path crash the request.

I confirmed locally with this test (passes by raising `MatchError`):

```elixir
test "crashes with MatchError when Tests.update_test_case returns an error", %{
  conn: conn, user: user, project: project
} do
  test_case =
    RunsFixtures.test_case_fixture(
      project_id: project.id,
      state: "muted",
      last_ran_at: NaiveDateTime.utc_now()
    )

  stub(Tests, :get_test_case_by_id, fn _id -> {:ok, test_case} end)
  stub(Tests, :update_test_case, fn _id, _attrs, _opts -> {:error, :not_found} end)

  assert_raise MatchError, fn ->
    conn
    |> put_req_header("content-type", "application/json")
    |> put(
      "/api/projects/#{user.account.name}/#{project.name}/tests/test-cases/#{test_case.id}",
      %{state: "enabled"}
    )
  end
end
```

Suggested fix: turn this into a `case` and surface `{:error, :not_found}` as a 404 (and any other `{:error, _}` as a 422/500 with a real body).

server/lib/tuist_web/router.ex
```diff
@@ -497,6 +497,7 @@ defmodule TuistWeb.Router do
             end
 
             get "/:test_case_id", TestCasesController, :show
+            put "/:test_case_id", TestCasesController, :update
```
pepicrft: The PR description says `PATCH /projects/.../test-cases/:id` and "state-only", but the route is `put` and the body also accepts `is_flaky`. `PATCH` is the conventional verb for partial updates with optional individual fields (which is exactly what this endpoint does); `PUT` semantically implies full replacement. Worth either switching the verb to `patch` or updating the description so future readers don't get tripped up.

server/lib/tuist/authorization.ex
```diff
@@ -486,6 +486,17 @@ defmodule Tuist.Authorization do
...
       allow([:authenticated_as_account, scopes_permit: "project:tests:write"])
     end
+
+    action :update do
+      desc("Allows users of a project to update a tests data.")
+      allow([:authenticated_as_user, user_role: :user])
+
+      desc("Allows the admin of a project to update tests data.")
+      allow([:authenticated_as_user, user_role: :admin])
+
```
pepicrft, line 496: Heads up: this allows any project `:user` (i.e. every member of the project) to mute/unmute or mark/unmark tests as flaky. Most other `:update` actions in this file are admin-only (e.g. `organization`, `member`, `invitation`). The PR description frames this as an operator unquarantine command, which sounds more admin-y. Could you confirm this is the intended scope? If yes, no change needed - just worth being explicit since it's a noticeable departure from the surrounding pattern.

justine-acorns: Yes, that was my intention! I wanted this to mirror the permission scope from the Tuist web dashboard. Anyone can modify the state or flaky flag for a test, so I wanted the test update API to allow that as well. I can change it if that is a deviation from the usual pattern and you have concerns though.

pepicrft (approved):

server/lib/tuist/authorization.ex
fortmarek: I think updating tests should be available to users, don't see that as an admin task.

PR #10464, opened 2026-04-26
fix(server): wire in-cluster Valkey for canary/production rate limiter

fortmarek:
### Symptom

Acceptance tests against `canary.tuist.dev` failed with `unknownError(500)` on every test that hits `POST /api/auth`:

- `generated_project_with_caching_enabled()`
- `multiplatform_app_module_cache()`
- `multiplatform_app_selective_testing()`

Server logs (canary pod):

```
[error] ** (MatchError) no match of right hand side value:
    {:error, %Redix.ConnectionError{reason: :closed}}
    (hammer_backend_redis 7.1.0) lib/hammer/redis/token_bucket.ex:85
    (tuist 0.1.0) lib/tuist_web/controllers/api/auth_controller.ex:253
```

### Root cause

The auth rate limiter uses Hammer.Redis when `Tuist.Environment.redis_url()` is set, otherwise an in-memory bucket ([`auth.ex`](server/lib/tuist_web/rate_limit/auth.ex#L10-L17)). On Render, both canary and production pointed at managed Valkey instances (`redis-canary`, `redis-production`). Those services were torn down in the k8s migration, but `priv/secrets/{can,prod}.yml.enc` still referenced `redis://red-d35ffquuk2gs73bgekfg:6379`.

`redis_url` resolved non-nil → Hammer tried Redis → DNS lookup failed → connection `:closed` → `{:ok, _} = ...hit(...)` MatchError → 500.

### Fix

Add an embedded Valkey 8 Deployment + Service to the Tuist Helm chart, gated by `redis.enabled`. Inject `TUIST_REDIS_URL` into the server pod, then **clear `redis_url` from `priv/secrets/{can,prod,stag}.yml.enc`** so the env var is the single source of truth. Without `redis_url` the rate limiter cleanly falls back to the per-pod InMemory bucket — a safer failure mode than dialing a dead host.

- `redis.enabled: true` in canary + production overlays
- Staging stays as-is (in-memory bucket fits a single-pod, low-traffic env)
- No PVC: the token bucket is ephemeral by design; pod restart resetting limits is acceptable
- `--save "" --appendonly no` so Valkey doesn't try to fsync to a non-existent volume
- NetworkPolicy restricts ingress on the Valkey Service to server pods (toggle via `redis.embedded.networkPolicy.enabled`, default true)
- Image pinned to `valkey/valkey:8.1-alpine`, with conservative resource requests/limits
- `TUIST_REDIS_URL` is only emitted when `tuist.redisUrl` resolves non-empty, so `mode: external` with no `external.url` won't ship an empty value to the pod

### Scope note

This PR also bundles a separate RBAC commit (`fix(server): grant CI deployer SA cluster-admin (chart needs it)`) that's logically independent — the k8s-monitoring chart bundles ClusterRoles whose creation requires `escalate`/`bind` on cluster-roles, which is functionally cluster-admin anyway, so the previous "narrow" framing was abandoned. Already applied to all three clusters out-of-band; this commit just reconciles the manifest in-repo. Kept in this PR rather than split because the RBAC change unblocks the same Acceptance Tests path.

### Manual validation

```bash
helm template tuist ./infra/helm/tuist \
  -f ./infra/helm/tuist/values-managed-common.yaml \
  -f ./infra/helm/tuist/values-managed-canary.yaml \
  --set server.image.tag=test | grep -A1 TUIST_REDIS_URL
# - name: TUIST_REDIS_URL
#   value: "redis://tuist-tuist-redis:6379"
```

Staging template diff is a no-op — `redis.enabled` defaults to false.

### Related

Follow-up to [#10462](https://github.com/tuist/tuist/pull/10462) and [#10463](https://github.com/tuist/tuist/pull/10463). Original failing run: <https://github.com/tuist/tuist/actions/runs/24941287267>.


pepicrft (approved):
LGTM on the Valkey wiring, fix is at the right layer and the deployment shape (single replica, Recreate, no PVC, save/appendonly disabled) matches the ephemeral nature of the bucket. A few things worth following up on, none blocking:

1. `TUIST_REDIS_URL` is emitted whenever `redis.enabled`, but `tuist.redisUrl` returns empty when `mode: external` without `external.url` set. Worth either gating the env var on the helper being non-empty, or adding a `required` assertion, so we don't ship an empty value to the pod.

2. No NetworkPolicy or auth on the Valkey Service. Any pod in the cluster can talk to it. Data is ephemeral so impact is low, but a NetworkPolicy restricting ingress to the server pod selector would be cheap insurance, especially given the cluster-admin widening in this same PR.

3. `resources: {}` by default. A small default request/limit would avoid surprise OOM kills on a busy node.

4. The RBAC widening to `cluster-admin` is reasonable given what k8s-monitoring pulls in transitively, but it feels out of scope for a PR titled around the rate limiter and isn't mentioned in the description. Could be worth splitting or at least calling out explicitly so it's easier to find later.

5. `8-alpine` is a floating tag. Pinning to a minor or digest would help reproducibility, low priority.

PR #10467, opened 2026-04-26
feat(cli): add CopyFileElement.buildProduct for embedding target products in Copy Files phases

freak4pc:
Adds a new `.buildProduct(name:condition:codeSignOnCopy:)` case to `CopyFileElement` so that a dependent target's build product can be referenced in a Copy Files build phase. The product is resolved from `BUILT_PRODUCTS_DIR` at build time rather than from the filesystem at generation time.

This enables the standard macOS Login Item embedding pattern where a helper `.app` is copied into `Contents/Library/LoginItems` via a Copy Files phase referencing the helper target's product.

Resolves <https://github.com/tuist/tuist/issues/YYY>

### How to test locally

Unit tests and/or fixtures

### DIFF
--- a/cli/Sources/ProjectDescription/CopyFileElement.swift
+++ b/cli/Sources/ProjectDescription/CopyFileElement.swift
@@ -1,4 +1,5 @@
-/// A file element from a glob pattern, a folder reference, or a build product which is conditionally applied to specific platforms
+/// A file element from a glob pattern, a folder reference, or a build product which is conditionally applied to specific
+/// platforms
 /// with an optional "Code Sign On Copy" flag.
 public enum CopyFileElement: Codable, Equatable, Sendable, ExpressibleByStringInterpolation {
     /// A file path (or glob pattern) to include with an optional PlatformCondition to control which platforms it applies.
--- a/cli/Sources/ProjectDescription/CopyFileElement.swift
+++ b/cli/Sources/ProjectDescription/CopyFileElement.swift
@@ -1,5 +1,5 @@
-/// A file element from a glob pattern or a folder reference which is conditionally applied to specific platforms with an optional
-/// "Code Sign On Copy" flag.
+/// A file element from a glob pattern, a folder reference, or a build product which is conditionally applied to specific platforms
+/// with an optional "Code Sign On Copy" flag.
 public enum CopyFileElement: Codable, Equatable, Sendable, ExpressibleByStringInterpolation {
     /// A file path (or glob pattern) to include with an optional PlatformCondition to control which platforms it applies.
     /// "Code Sign on Copy" can be optionally enabled for the glob.
@@ -9,9 +9,14 @@ public enum CopyFileElement: Codable, Equatable, Sendable, ExpressibleByStringIn
     /// to. "Code Sign on Copy" can be optionally enabled for the folder reference.
     case folderReference(path: Path, condition: PlatformCondition? = nil, codeSignOnCopy: Bool = false)
 
+    /// A reference to a dependent target's build product (e.g. a helper `.app` for Login Items embedding).
+    /// The product is resolved from `BUILT_PRODUCTS_DIR` during the build, not from the filesystem at generation time.
+    case buildProduct(name: String, condition: PlatformCondition? = nil, codeSignOnCopy: Bool = false)
+
     private enum TypeName: String, Codable, Sendable {
         case glob
         case folderReference
+        case buildProduct
     }
 
     private var typeName: TypeName {
@@ -20,6 +25,8 @@ public enum CopyFileElement: Codable, Equatable, Sendable, ExpressibleByStringIn
             return .glob
         case .folderReference:
             return .folderReference
+        case .buildProduct:
+            return .buildProduct
         }
     }
 
--- a/cli/Sources/ProjectDescription/CopyFileElement.swift
+++ b/cli/Sources/ProjectDescription/CopyFileElement.swift
@@ -1,6 +1,5 @@
-/// A file element from a glob pattern, a folder reference, or a build product which is conditionally applied to specific
-/// platforms
-/// with an optional "Code Sign On Copy" flag.
+/// A file element from a glob pattern, a folder reference, or a build product which is conditionally
+/// applied to specific platforms with an optional "Code Sign On Copy" flag.
 public enum CopyFileElement: Codable, Equatable, Sendable, ExpressibleByStringInterpolation {
     /// A file path (or glob pattern) to include with an optional PlatformCondition to control which platforms it applies.
     /// "Code Sign on Copy" can be optionally enabled for the glob.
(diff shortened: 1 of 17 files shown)


fortmarek:
Pushed c2fd521 to address the review feedback. Summary of changes and reasoning:

### Correctness

- **`CopyFileElement.path` → `AbsolutePath?`** (XcodeGraph). Returning `.root` for `.buildProduct` was a footgun: every build product aliased to `/`, which silently broke any path-based comparison/dedup. Made it optional so the type stops lying. Two consumers needed updating: `CopyFilesContentHasher` (now switches on cases directly so `path` is bound non-optionally) and `CopyFilesAction.cleanPackages` (guards on the optional and passes buildProduct entries through).
- **Dedicated `BuildPhaseGenerationError.missingProductReference(targetName:)`**. The previous code threw `missingFileReference` with a synthetic `/BUILT_PRODUCTS_DIR/<name>` AbsolutePath, which read confusingly when surfaced. The new case has a message that tells the user to declare the helper target as a dependency — which is the actual fix.
- **Restored deterministic Copy Files ordering**. The original code sorted `action.files` by path; the first revision of this PR dropped that sort, which would reshuffle Copy Files phases in PBXProj output for *every* project (not just ones using `.buildProduct`). Added a `copyFileElementSortKey` helper that sorts file/folderReference by path (matching pre-PR behaviour) and buildProduct by name, with a case index prefix to keep them grouped.
- **Round-trip `.buildProduct` in `PBXCopyFilesBuildPhaseMapper`**. The Xcode-to-graph mapper previously emitted only `.file(...)`, so a Copy Files phase referencing `BUILT_PRODUCTS_DIR` would either be lost or distorted on import. It now detects `sourceTree == .buildProductsDir` and emits `.buildProduct(name:)`.

### Cosmetic / docs

- Reflowed the awkward two-line doc comment on `ProjectDescription.CopyFileElement`.
- Added a comment explaining why `RemoveHeadersOnCopy` is set alongside `CodeSignOnCopy` for buildProduct (matches what Xcode emits for "Embed App Extensions" / "Embed Frameworks").

### Tests

- Refactored the mixed files/buildProducts test to use the shared `createFileElements`/`createProductFileElements` helpers (consistent with other tests in the file).
- Added `generateCopyFilesBuildPhases_withBuildProduct_throwsWhenProductIsMissing` — verifies the new `.missingProductReference` error fires when the helper target isn't registered, which is the most likely user error.
- Updated the existing buildProduct test's expected ordering to match the restored sort.
- Adjusted `PBXCopyFilesBuildPhaseMapperTests` for the optional `path?`.

### Verification
- `xcsiftbuild build` (tuist scheme): clean.
- 323 `TuistGeneratorTests` pass.
- 228 `TuistHasherTests` + `TuistLoaderTests` pass.
- `XcodeGraphTests` (CopyFile suite) and `PBXCopyFilesBuildPhaseMapperTests` pass via `swift test`.

fortmarek:
Worth recording the design space we considered for "embed a target's product into a Copy Files phase," since this PR introduces a new API surface and we want to be intentional about it.

### What Tuist already auto-embeds

App extensions, Watch apps, App Clips, XPC services, macOS bundles, and System Extensions all have automatic Embed phases driven by the `isEmbeddable*` / `canEmbed*` predicates on `Target`. The user only declares `.target(name:)` and Tuist emits the right Copy Files phase at the right destination.

**Login Items don't fit any of these.** A Login Item helper has to be an `.app` (with `LSBackgroundOnly`) embedded under `Contents/Library/LoginItems/` so `SMLoginItemSetEnabled` / `SMAppService` can find it. It can't be modelled as `.appExtension` (different runtime, different destination), and a sibling `.app` dependency is ambiguous — it could mean Login Item, SharedSupport, a separate distributable, etc.

### Workarounds that don't cut it

- **Run Script phase** doing `cp -R "$BUILT_PRODUCTS_DIR/Helper.app" "$TARGET_BUILD_DIR/$WRAPPER_NAME/Contents/Library/LoginItems/"`. Loses codesign-on-copy integration, breaks archive bundling, codesigning of the embedded helper has to be wired separately.
- **Glob** like `.glob(pattern: "${BUILT_PRODUCTS_DIR}/Helper.app")`. Globs are evaluated at generation time against the filesystem; the product doesn't exist yet. Doesn't work.
- **Different product type**. Login Items can't be `.appExtension` — wrong runtime model.

So the gap is real and we need *some* new surface.

### Design alternatives

**A. Typed dependency** — `.target(name: "Helper", embedAt: .loginItems)`. Single declaration; the dependency edge and the embed are inseparable, so users can't forget one. Downside: covers only a fixed enumeration of destinations and doesn't extend to "embed product X into arbitrary Copy Files destination Y."

**B. Auto-embed predicate** — `isEmbeddableLoginItem()` + `canEmbedLoginItems()`, matching the existing pattern. Cleanest for the exact Login Items case, but it can't disambiguate intent for `.app` → `.app` macOS dependencies (Login Item vs SharedSupport vs sibling app), so it only works if you mark intent somewhere — at which point you've reinvented (A).

**C. This PR** — `CopyFileElement.buildProduct(name:)` inside an explicit `CopyFilesAction`. Most flexible: any product into any Copy Files destination, including future patterns we haven't enumerated (SharedSupport, vendored plug-ins, custom helper folders). Downside: the user declares the helper twice — once in `dependencies` for build ordering, once in `copyFiles` for the embed — and the new `missingProductReference` error is what catches the omission.

### Recommendation

We lean toward keeping the PR's shape (C). Reasons:

1. **Generality matches Xcode's underlying model.** Copy Files phases in Xcode can target arbitrary destinations and arbitrary product references; the PR's API mirrors that 1:1. (A) and (B) bake in an enumeration that we'd have to expand every time a new "embed at custom path" pattern shows up.
2. **It composes with everything already in `CopyFilesAction`.** Code-sign-on-copy, platform conditions, mixing with file/folderReference entries — all work the same way as for existing cases.
3. **Round-trip is now symmetric.** With the reverse-mapper change in this branch, an Xcode project using "drag from Products group into a Copy Files phase" round-trips through Tuist as `.buildProduct(name:)` — without (C), we'd silently drop or distort those entries on import.
4. **The double-declaration cost is low.** Users declaring `.target(name:)` in `dependencies` is already idiomatic for build ordering, and the dedicated error message (`"Make sure 'X' is declared as a dependency of this target"`) makes the failure mode obvious.

(A) is still appealing as *sugar* on top — a future PR could add `.target(name:, embedAt:)` for the common Login Items / SharedSupport cases, lowering it to a synthesised `CopyFilesAction` + `.buildProduct` underneath. The escape hatch in this PR stays for everything (A) doesn't enumerate.

pepicrft (approved):
Thanks for the design analysis. I was at first hesitant to the model, since it doesn't align with Tuist's tradition of mapping the graph to the right build settings and phases, but there are cases like this one that escape that model and for which we should provide escape hatches.

PR #10476, opened 2026-04-27
chore(cli): bump FileSystem to 0.17.0 and drop SwiftNIO

pepicrft:
## Summary
- Bump `tuist.FileSystem` from `0.16.2` to `0.17.0`.
- Remove the `TUIST_FILESYSTEM_BACKEND` env var and the `isSwiftFileSystemBackendEnabled` flag — FileSystem 0.17 drops the file system backend concept, so the conditional concurrency branches are no longer needed (call sites now always run concurrently with `Int.max`).
- Drop `apple.swift-nio` as a direct dependency of the CLI; it had no remaining direct usage. `grpc-swift-nio-transport` (used by gRPC) stays.

## Test plan
- [x] `xcodebuild build -scheme tuist` succeeds
- [x] `tuist test TuistUnitTests --test-targets TuistHasherTests/HeadersContentHasherTests --test-targets TuistHasherTests/GraphContentHasherTests` passes
- [ ] CI green

### DIFF
--- a/cli/Sources/TuistInspectCommand/Services/TargetImportsScanner.swift
+++ b/cli/Sources/TuistInspectCommand/Services/TargetImportsScanner.swift
@@ -2,7 +2,6 @@
     import FileSystem
     import Mockable
     import Path
-    import TuistEnvironment
     import XcodeGraph
 
     @Mockable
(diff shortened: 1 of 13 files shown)


cli/Sources/TuistInspectCommand/Services/TargetImportsScanner.swift
```diff
@@ -30,9 +29,8 @@
                 filesToScan.append(contentsOf: headers.public)
                 filesToScan.append(contentsOf: headers.project)
             }
-            let maxConcurrentTasks = Environment.current.isSwiftFileSystemBackendEnabled ? Int.max : 100
```
pepicrft, line 33: We've been running this without caps, and we had no issues. I don't have 100% confidence this change will be free of issues, but we can revert if needed, pointing to this PR in case users report issues.

fortmarek (approved):
👋

PR #10479, opened 2026-04-27
chore(ci): validate Helm charts at the PR stage

fortmarek:
## Summary
Adds `.github/workflows/helm.yml`, which validates every Helm chart in the repo on PRs that touch `infra/helm/**`, `kura/ops/helm/**`, or the workflow file itself. Three layers, each progressively stricter:

1. **`helm lint`** — Chart.yaml well-formedness, template parse, value reference sanity. One job per chart (`infra/helm/{tuist,k8s-monitoring,platform}`, `kura/ops/helm/kura`). Runs `helm dependency update` first when the chart has a `dependencies:` block so subchart references resolve.
2. **`helm template`** — full render with every committed values overlay. The matrix covers all 8 combinations the deploy workflows actually use:
   - `tuist` × {staging, canary, production} (each with `values-managed-common.yaml` as the base, matching `server-deployment.yml`)
   - `k8s-monitoring` × {staging, canary, production}
   - `platform/values-hetzner.yaml`
   - `kura/values-scaleway.yaml`
3. **`kubeconform`** — schema-validates the rendered manifests against the built-in Kubernetes API and the [datreeio CRDs catalog](https://github.com/datreeio/CRDs-catalog) (so `ExternalSecret` etc. validate too). `-strict` rejects unknown fields, so typo'd keys in resource specs fail the build. `-ignore-missing-schemas` skips operator-specific CRDs without registered schemas rather than vendoring everything.

## Why
PRs like [#10473](https://github.com/tuist/tuist/pull/10473) modify the umbrella chart but nothing today verifies the chart even templates cleanly until `server-deployment.yml` runs `helm upgrade --install` against a live cluster. By that point a typo costs a rolled-back deploy and a `helm history` dive — much cheaper to fail in CI before merge.

`helm template` catches the obvious class of breakage (bad pipelines, missing required values, dependency resolution failures), but `helm template` happily emits `apiVersion`/field combos that the API server would reject. `kubeconform` closes that gap without needing a real cluster.

## Behaviour
- Both jobs use `fail-fast: false` so all chart × overlay legs report independently.
- Tools install via `mise` (`helm`, `kubeconform`) using the same `MISE_VERSION` (`2026.4.18`) as the rest of the workflows.
- Triggers: `push` to `main`, `pull_request` (non-draft), `merge_group`.
- Concurrency: cancels in-progress runs on the same head ref.

## Verified locally
Ran the exact commands the workflow runs against every chart × overlay:

| Chart × overlay | kubeconform result |
| --- | --- |
| tuist (production) | 13/13 valid |
| tuist (canary) | renders cleanly |
| tuist (staging) | renders cleanly |
| k8s-monitoring (production) | 35/35 valid |
| k8s-monitoring (canary) | renders cleanly |
| k8s-monitoring (staging) | renders cleanly |
| platform (hetzner) | 91 valid, 20 skipped (operator CRDs without registered schemas) |
| kura (scaleway) | 5/5 valid |

## Test plan
- [ ] Workflow runs green on this PR
- [ ] Introduce a deliberate typo in one of the chart templates (e.g. wrong `apiVersion`) on a scratch branch — `kubeconform` should fail
- [ ] Introduce a missing required value (e.g. remove `cluster.name` from a k8s-monitoring overlay) — `helm template` should fail

### DIFF
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -88,24 +88,15 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # Tuist umbrella: every deploy passes values-managed-common.yaml
-          # plus the per-env overlay, so render the same way here. Only
-          # canary and production are checked: production is the
+          # Only production overlays are checked. Production is the
           # load-bearing overlay (HPA, PDB, NetworkPolicy, second
-          # deployment) and canary is the rollout gate. Staging diverges
-          # minimally from canary and would just duplicate signal.
-          - name: tuist (canary)
-            chart: infra/helm/tuist
-            release: tuist
-            values: values-managed-common.yaml values-managed-canary.yaml
+          # deployment) so it exercises the templates the most. Canary
+          # and staging diverge minimally from it and would just
+          # duplicate signal.
           - name: tuist (production)
             chart: infra/helm/tuist
             release: tuist
             values: values-managed-common.yaml values-managed-production.yaml
-          - name: k8s-monitoring (canary)
-            chart: infra/helm/k8s-monitoring
-            release: k8s-monitoring
-            values: values-canary.yaml
           - name: k8s-monitoring (production)
             chart: infra/helm/k8s-monitoring
             release: k8s-monitoring
--- a/.github/workflows/helm.yml
+++ b/.github/workflows/helm.yml
@@ -62,7 +62,6 @@ jobs:
       - uses: jdx/mise-action@v4.0.1
         with:
           version: ${{ env.MISE_VERSION }}
-          cache: "false"
       - name: Install helm
         run: mise use -g "helm@${HELM_VERSION}"
       - name: helm dependency update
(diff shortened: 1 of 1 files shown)


.github/workflows/helm.yml
```diff
@@ -0,0 +1,149 @@
...
+            helm dependency update "$CHART"
+          fi
+      - name: helm template
+        env:
+          CHART: ${{ matrix.chart }}
+          RELEASE: ${{ matrix.release }}
+          VALUES: ${{ matrix.values }}
+        run: |
+          args=()
+          for v in $VALUES; do
```
pepicrft: Word-splitting `$VALUES` works for current filenames, but a values file with a space in it would silently break. A YAML list per matrix entry would be sturdier — not a blocker though.

.github/workflows/helm.yml
```diff
@@ -0,0 +1,149 @@
...
+            args+=("-f" "$CHART/$v")
+          done
+          helm template "$RELEASE" "$CHART" "${args[@]}" > rendered.yaml
+          echo "Rendered $(grep -cE '^kind:' rendered.yaml) resources, $(wc -l <rendered.yaml) lines."
```
pepicrft: `grep -cE '^kind:'` will also match nested `kind:` fields inside CRD specs (e.g. `kind: ConfigMap` under a template). Fine for an informational log line, just don't rely on it as a true resource count.

.github/workflows/helm.yml
```diff
@@ -0,0 +1,149 @@
...
+          - name: platform (hetzner)
+            chart: infra/helm/platform
+            release: platform
+            values: values-hetzner.yaml
+    steps:
+      - uses: actions/checkout@v4
+      - uses: jdx/mise-action@v4.0.1
+        with:
+          version: ${{ env.MISE_VERSION }}
+          cache: "false"
```
pepicrft: `cache: "false"` reinstalls helm/kubeconform on every run. Probably only seconds, but trivially recoverable by enabling the mise cache.

.github/workflows/helm.yml
```diff
@@ -0,0 +1,149 @@
...
+      - name: helm lint
+        env:
+          CHART: ${{ matrix.chart }}
+        run: helm lint "$CHART"
+
+  render:
+    name: Render (${{ matrix.name }})
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    if: github.event.pull_request.draft == false || github.event_name != 'pull_request'
```
pepicrft, line 97: This `if:` guard is duplicated with the `lint` job (line 52). For two jobs duplication is fine, but if a third job lands it'd be worth lifting to a workflow-level guard or a reusable check.

pepicrft (approved):

PR #10485, opened 2026-04-27
feat(server): add ephemeral preview environments on dedicated kubernetes cluster

pepicrft:
## Summary

Per-PR / per-commit ephemeral preview deploys on a dedicated `tuist-preview` Syself/Hetzner cluster. The worker pool scales to zero at idle so resting cost is ~€5/mo (single small control plane); workers spin up on demand and the hourly sweep workflow scales them back down once nothing is live.

Each preview is one Helm release in its own namespace, running the full embedded stack (server + cache + postgres + clickhouse + minio) on a single labeled+tainted worker. URLs follow:

```
pr-1234.preview.tuist.dev      # PR builds (stable across pushes; redeploy upgrades)
sha-abc1234.preview.tuist.dev  # one-off commits / branches without a PR
```

TLS comes from a wildcard cert pre-issued at cluster bootstrap (cert-manager DNS-01 against Cloudflare); per-PR ingresses pick it up via ingress-nginx's `--default-ssl-certificate` flag — no per-namespace plumbing.

## How previews are triggered

**Label-driven (default):** add the `preview` label to a PR. Subsequent pushes auto-redeploy; removing the label or closing the PR tears down. The PR timeline gets a "Preview – PR #1234" GitHub Deployment card (active during the deploy, set inactive on teardown) and a single auto-edited bot comment with the URL.

**Manual (one-offs):** `gh workflow run preview-deploy.yml -f pr_number=1234 -f ttl_hours=24` or `-f commit_sha=abc1234 -f ttl_hours=4`. Useful for re-deploying a closed PR or for branches without a PR.

Fork PRs are filtered out at the resolve job — secrets don't propagate to fork-based runs anyway.

## What's new

**Chart**
- `server.externalSecrets.license.{item,field}` — license syncs from 1Password through the same ESO ClusterSecretStore pattern already used for `MASTER_KEY`, gated independently of `managedSecrets` so previews don't need the encrypted `priv/secrets` blob. Backwards-compatible with the existing inline-license path; mutually exclusive with `server.license.key` (helper `fail`s if both are set).
- New `tuist.licenseEnv` helper centralises the license env wiring across deployment + migration job.
- `values-preview.yaml` overlay: pin to `role=preview` node, embed all dependencies, scaled-down resource requests, ingress configured for the wildcard-cert path.

**Cluster**
- `infra/k8s/syself/workload-cluster-preview.yaml` — CAPI Cluster CR (1 × cpx22 control plane + 0 × cpx31 workers default).
- `infra/k8s/syself/preview-mgmt-rbac.yaml` — narrow `preview-scaler` SA on the management cluster (only `get/patch` on the preview Cluster, read-only on MachineDeployments) for the workflows to scale the worker pool.

**CI**
- `.github/workflows/preview-deploy.yml` — `pull_request` (label-driven) + `workflow_dispatch`. Split into `resolve` (gates on event filter, decides deploy vs. teardown), `start-deployment`, `build`, `scale-up` (concurrency: `preview-cluster-scale`, shared with sweep), per-release `deploy`, and `teardown`. Different PRs deploy in parallel while the management Cluster CR patches stay one-at-a-time.
- `.github/workflows/preview-sweep.yml` — hourly cron + manual dispatch. Backstop deletion for any namespace past its `tuist.dev/expires-at` label, scales the worker pool back to 0 when no previews remain.

**Local validation**
- `infra/k8s/kind-preview.yaml` + `infra/mise/tasks/helm/preview-{up,down}.sh` — multi-node kind config (control plane + general worker + labeled+tainted preview worker) and end-to-end task. Supports both ESO mode (`OP_SERVICE_ACCOUNT_TOKEN`) and inline fallback (`TUIST_LICENSE_KEY`) for offline validation.

**Docs**
- `infra/k8s/syself-onboarding.md` §9 — preview cluster bootstrap end-to-end (cluster apply → ESO → wildcard DNS+cert → both CI ServiceAccounts → first deploy).

## Test plan

Verified live on a multi-node kind cluster:

- [x] All 8 workloads schedule to the `role=preview` worker; default workloads stay on the general worker
- [x] ESO + 1Password license sync works against the real `tuist-k8s-preview` vault (license bytes round-trip with `| trim` correctly handling trailing newline)
- [x] `helm install --atomic --wait` succeeds in ~70s
- [x] Postgres + ClickHouse migrations both run forward in the migration job
- [x] Server pod boots and serves `GET /ready` → `HTTP 200`, `GET /` → `HTTP 302` (Phoenix redirect to configured `appUrl`)
- [x] Both modes of license sourcing (ESO + inline) render correctly via `helm template`; mutually-exclusive guard `fail`s when both are set
- [x] Preview overlay leaves Ingress unrendered without CI overrides; renders correctly with both default-ssl-certificate and reflector paths

Not exercised here (require real Syself + Cloudflare):

- [ ] CAPI scale-up timing on Hetzner cold-boot
- [ ] cert-manager DNS-01 against Cloudflare for the wildcard cert
- [ ] GitHub Actions secret + environment resolution
- [ ] `pull_request` label-trigger event flow + GitHub Deployment status transitions

## Follow-ups (separate PRs)

- **Durable node labels at bootstrap.** Custom Syself ClusterClass variant that sets `role=preview` label + taint at kubelet registration, so the workflow's relabel-before-deploy step becomes belt-and-braces rather than load-bearing. Genuinely upstream-dependent — needs Syself to expose label/taint config via ClusterClass variables.

## Sharp edges

- Worker labels/taints aren't durable across MachineDeployment scale events; the deploy workflow re-applies them before `helm install` as a safeguard. See follow-up above for the long-term fix.
- One-replica worker = single point of failure. Acceptable for previews; flagged in the onboarding doc.

### DIFF
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -1,0 +1,295 @@
+name: Preview Deploy
+
+# Deploy an ephemeral preview environment for a PR or a specific commit.
+#
+# Usage:
+#   - Run from the GitHub UI (Actions → Preview Deploy → Run workflow).
+#   - Provide EITHER pr_number OR commit_sha. ttl_hours sets when the sweep
+#     workflow auto-deletes the namespace.
+#
+# Convention:
+#   pr_number=1234   → release pr-1234,    namespace preview-pr-1234,
+#                      host pr-1234.preview.tuist.dev
+#   commit_sha=abc…  → release sha-abc1234, namespace preview-sha-abc1234,
+#                      host sha-abc1234.preview.tuist.dev (7-char short SHA)
+#
+# Required GitHub Environment secrets (environment: server-k8s-preview):
+#   KUBECONFIG       - base64 kubeconfig for the preview WORKLOAD cluster's
+#                      CI ServiceAccount (cluster-admin in preview-system).
+#   KUBECONFIG_MGMT  - base64 kubeconfig for the SYSELF MANAGEMENT cluster's
+#                      `preview-scaler` SA (narrow Role: get/patch on
+#                      Cluster/tuist-preview, read on MachineDeployments).
+#
+# Both are minted at cluster bootstrap; see infra/k8s/syself-onboarding.md §10.
+
+on:
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: PR number to deploy (omit if using commit_sha)
+        required: false
+      commit_sha:
+        description: Commit SHA to deploy (omit if using pr_number)
+        required: false
+      ttl_hours:
+        description: Auto-delete after N hours (max 168 = 7 days)
+        required: true
+        default: "24"
+
+permissions:
+  contents: read
+  packages: write
+  pull-requests: write
+  id-token: write
+
+env:
+  REGISTRY: ghcr.io
+  IMAGE_NAME: tuist/tuist
+  HELM_CHART_PATH: infra/helm/tuist
+  PREVIEW_BASE_DOMAIN: preview.tuist.dev
+
+jobs:
+  resolve:
+    name: Resolve inputs
+    runs-on: ubuntu-latest
+    outputs:
+      release: ${{ steps.r.outputs.release }}
+      namespace: ${{ steps.r.outputs.namespace }}
+      host: ${{ steps.r.outputs.host }}
+      ref: ${{ steps.r.outputs.ref }}
+      short_sha: ${{ steps.r.outputs.short_sha }}
+      expires_at: ${{ steps.r.outputs.expires_at }}
+    steps:
+      - id: r
+        env:
+          PR: ${{ inputs.pr_number }}
+          SHA: ${{ inputs.commit_sha }}
+          TTL_HOURS: ${{ inputs.ttl_hours }}
+        run: |
+          set -euo pipefail
+          if [ -n "$PR" ] && [ -n "$SHA" ]; then
+            echo "::error::Provide pr_number OR commit_sha, not both."
+            exit 1
+          fi
+          if [ -z "$PR" ] && [ -z "$SHA" ]; then
+            echo "::error::Provide either pr_number or commit_sha."
+            exit 1
+          fi
+
+          # Clamp TTL to [1, 168].
+          if ! [[ "$TTL_HOURS" =~ ^[0-9]+$ ]] || [ "$TTL_HOURS" -lt 1 ] || [ "$TTL_HOURS" -gt 168 ]; then
+            echo "::error::ttl_hours must be an integer between 1 and 168."
+            exit 1
+          fi
+          EXPIRES_AT=$(( $(date +%s) + TTL_HOURS * 3600 ))
+
+          if [ -n "$PR" ]; then
+            REF=$(gh pr view "$PR" --repo "${{ github.repository }}" --json headRefOid -q .headRefOid)
+            RELEASE="pr-${PR}"
+          else
+            REF="$SHA"
+            RELEASE="sha-${SHA:0:7}"
+          fi
+
+          SHORT_SHA="${REF:0:12}"
+          NAMESPACE="preview-${RELEASE}"
+          HOST="${RELEASE}.${PREVIEW_BASE_DOMAIN}"
+
+          echo "release=$RELEASE"     >> "$GITHUB_OUTPUT"
+          echo "namespace=$NAMESPACE" >> "$GITHUB_OUTPUT"
+          echo "host=$HOST"           >> "$GITHUB_OUTPUT"
+          echo "ref=$REF"             >> "$GITHUB_OUTPUT"
+          echo "short_sha=$SHORT_SHA" >> "$GITHUB_OUTPUT"
+          echo "expires_at=$EXPIRES_AT" >> "$GITHUB_OUTPUT"
+
+          {
+            echo "## Preview deploy"
+            echo "- Release: \`$RELEASE\`"
+            echo "- Namespace: \`$NAMESPACE\`"
+            echo "- Host: https://$HOST"
+            echo "- Image tag: \`sha-$SHORT_SHA\`"
+            echo "- Expires: $(date -u -d "@$EXPIRES_AT" '+%Y-%m-%d %H:%M UTC')"
+          } >> "$GITHUB_STEP_SUMMARY"
+        env:
+          GH_TOKEN: ${{ github.token }}
+
+  build:
+    name: Build + push
+    needs: resolve
+    runs-on: namespace-profile-default-with-volume
+    timeout-minutes: 40
+    outputs:
+      image_tag: ${{ steps.tag.outputs.image_tag }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.resolve.outputs.ref }}
+          submodules: false
+      - id: tag
+        run: echo "image_tag=sha-${{ needs.resolve.outputs.short_sha }}" >> "$GITHUB_OUTPUT"
+      - uses: namespacelabs/nscloud-setup-buildx-action@v0
+      - uses: docker/login-action@v3
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - uses: docker/build-push-action@v6
+        with:
+          context: .
+          file: ./server/Dockerfile
+          target: runner
+          push: true
+          tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ steps.tag.outputs.image_tag }}
+          build-args: |
+            TUIST_HOSTED=1
+            MIX_ENV=prod
+            APT_CACHE_BUST=${{ needs.resolve.outputs.ref }}
+          # Share the same cache scope as the managed-env builds — same image,
+          # same Dockerfile, layers are interchangeable.
+          cache-from: type=gha,scope=server-cluster
+          cache-to: type=gha,mode=max,scope=server-cluster
+
+  deploy:
+    name: Deploy ${{ needs.resolve.outputs.release }}
+    needs: [resolve, build]
+    runs-on: namespace-profile-default
+    environment:
+      name: server-k8s-preview
+      url: https://${{ needs.resolve.outputs.host }}
+    # Serialize per-release so two runs against the same PR don't race.
+    concurrency:
+      group: preview-deploy-${{ needs.resolve.outputs.release }}
+      cancel-in-progress: false
+    timeout-minutes: 30
+    env:
+      KUBECONFIG: ${{ github.workspace }}/.kube/config
+      KUBECONFIG_MGMT: ${{ github.workspace }}/.kube/mgmt
+      RELEASE: ${{ needs.resolve.outputs.release }}
+      NAMESPACE: ${{ needs.resolve.outputs.namespace }}
+      HOST: ${{ needs.resolve.outputs.host }}
+      IMAGE_TAG: ${{ needs.build.outputs.image_tag }}
+      EXPIRES_AT: ${{ needs.resolve.outputs.expires_at }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ needs.resolve.outputs.ref }}
+      - uses: jdx/mise-action@v3.2.0
+        with:
+          install_args: "helm kubectl"
+          cache: "false"
+      - name: Load kubeconfigs
+        run: |
+          mkdir -p "$(dirname "$KUBECONFIG")"
+          echo "${{ secrets.KUBECONFIG }}" | base64 -d > "$KUBECONFIG"
+          chmod 600 "$KUBECONFIG"
+          echo "${{ secrets.KUBECONFIG_MGMT }}" | base64 -d > "$KUBECONFIG_MGMT"
+          chmod 600 "$KUBECONFIG_MGMT"
+
+      - name: Scale preview worker pool up
+        # Patches the management Cluster CR to ensure md-0 has at least 1
+        # replica. Idempotent: a no-op when a worker is already up.
+        run: |
+          set -euo pipefail
+          ORG_NS=$(KUBECONFIG="$KUBECONFIG_MGMT" kubectl config view --minify -o jsonpath='{.contexts[0].context.namespace}')
+          CURRENT=$(KUBECONFIG="$KUBECONFIG_MGMT" kubectl -n "$ORG_NS" get cluster tuist-preview \
+            -o jsonpath='{.spec.topology.workers.machineDeployments[?(@.name=="md-0")].replicas}')
+          echo "md-0 replicas currently: $CURRENT"
+          if [ "${CURRENT:-0}" -lt 1 ]; then
+            echo "==> Scaling md-0 to 1..."
+            KUBECONFIG="$KUBECONFIG_MGMT" kubectl -n "$ORG_NS" patch cluster tuist-preview \
+              --type=json \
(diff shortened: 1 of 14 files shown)


.github/workflows/preview-deploy.yml
```diff
@@ -0,0 +1,295 @@
...
+#   KUBECONFIG       - base64 kubeconfig for the preview WORKLOAD cluster's
+#                      CI ServiceAccount (cluster-admin in preview-system).
+#   KUBECONFIG_MGMT  - base64 kubeconfig for the SYSELF MANAGEMENT cluster's
+#                      `preview-scaler` SA (narrow Role: get/patch on
+#                      Cluster/tuist-preview, read on MachineDeployments).
+#
+# Both are minted at cluster bootstrap; see infra/k8s/syself-onboarding.md §10.
+
+on:
+  workflow_dispatch:
```
fortmarek, line 38: Worth being explicit in the PR description that this is **manual-only** today (`workflow_dispatch`); the phrase "per-PR / per-commit ephemeral preview deploys" reads as if pushing to a PR auto-deploys.

**Follow-up to consider** (probably its own PR, not this one): label-driven automation similar to the Render integration. Shape would be:

- `pull_request: [labeled, synchronize, unlabeled, closed]` triggers, gated on the `preview` label being present.
- Add label or push to a labeled PR -> deploy / redeploy.
- Remove label or close PR -> uninstall the release.
- Keep the 24h TTL as a safety net so an orphaned namespace still gets swept.

Two things to think through before wiring it up:

1. **Forks**: `pull_request` from a fork can't read secrets, and `pull_request_target` runs the workflow definition from `main` with secrets but checks out fork code (the classic supply-chain footgun). Easiest gate: `github.event.pull_request.head.repo.full_name == github.repository` so only internal PRs auto-deploy and forks stay manual.
2. **Cost on chatty PRs**: every push redeploys, which spins the worker pool back up if it just scaled down. Probably fine given the workload is single-node, but worth deciding whether to debounce (e.g. only redeploy on `synchronize` if the PR is `ready_for_review`).

.github/workflows/preview-deploy.yml
```diff
@@ -0,0 +1,295 @@
...
+            echo "::error::ttl_hours must be an integer between 1 and 168."
+            exit 1
+          fi
+          EXPIRES_AT=$(( $(date +%s) + TTL_HOURS * 3600 ))
+
+          if [ -n "$PR" ]; then
+            REF=$(gh pr view "$PR" --repo "${{ github.repository }}" --json headRefOid -q .headRefOid)
+            RELEASE="pr-${PR}"
+          else
+            REF="$SHA"
```
fortmarek: Minor: `commit_sha` is not validated to be hex like `ttl_hours` is (`^[0-9]+$`). Not exploitable since downstream usage is always quoted, but a `^[0-9a-f]{7,40}$` check would be consistent with how `pr_number` is implicitly validated by `gh pr view` failing on non-numerics.

.github/workflows/preview-deploy.yml
```diff
@@ -0,0 +1,295 @@
...
+          # Share the same cache scope as the managed-env builds — same image,
+          # same Dockerfile, layers are interchangeable.
+          cache-from: type=gha,scope=server-cluster
+          cache-to: type=gha,mode=max,scope=server-cluster
+
+  deploy:
+    name: Deploy ${{ needs.resolve.outputs.release }}
+    needs: [resolve, build]
+    runs-on: namespace-profile-default
+    environment:
```
fortmarek, line 266: **Per-PR GitHub Deployments to consider.** Today every preview deploys under the same single Actions Environment, so the PR timeline doesn't get a distinct "Preview – PR #1234" card the way Render's GitHub integration produced.

The clean shape is to keep `server-k8s-preview` as the single Actions Environment (so `KUBECONFIG` / `KUBECONFIG_MGMT` and any reviewer gates stay in one place) and additionally create a per-PR Deployment via the [Deployments REST API](https://docs.github.com/rest/deployments/deployments):

- `POST /repos/{owner}/{repo}/deployments` with `environment: pr-1234`, `transient_environment: true`, `auto_merge: false`.
- Walk it through `pending -> in_progress -> success` with `environment_url` set to the preview host.
- The sweep workflow can mark the Deployment as `inactive` when it tears the namespace down so the timeline reflects real state.

Roughly a 30-line addition to the workflow with no cluster-side changes. Could land here or as a follow-up.

.github/workflows/preview-deploy.yml
```diff
@@ -0,0 +1,295 @@
...
+          cache: "false"
+      - name: Load kubeconfigs
+        run: |
+          mkdir -p "$(dirname "$KUBECONFIG")"
+          echo "${{ secrets.KUBECONFIG }}" | base64 -d > "$KUBECONFIG"
+          chmod 600 "$KUBECONFIG"
+          echo "${{ secrets.KUBECONFIG_MGMT }}" | base64 -d > "$KUBECONFIG_MGMT"
+          chmod 600 "$KUBECONFIG_MGMT"
+
+      - name: Scale preview worker pool up
```
fortmarek: **Race on the management Cluster CR.** Both this step and the sweep workflow `kubectl patch` `cluster/tuist-preview` without a `resourceVersion` precondition. Two concurrent deploys (different PRs) racing the sweep will collide on optimistic concurrency. The patches are idempotent so worst case is one workflow re-runs, but a workflow-level concurrency group:

```yaml
concurrency:
  group: preview-cluster-scale
  cancel-in-progress: false
```

on both this workflow and the sweep would tighten it without the per-release group having to absorb the contention.

fortmarek (commented):
## Overview

Well-scoped change. Adds per-PR / per-commit preview deploys on a dedicated `tuist-preview` Syself cluster scaled to zero at idle, each preview running the full embedded stack on a single labeled+tainted worker behind a wildcard cert, with a TTL label and an hourly sweep handling cleanup. The chart picks up a new `server.externalSecrets.license.{item,field}` block so previews can sync the license from 1Password without the encrypted `priv/secrets` blob.

The in-tree `kind` validation harness ([infra/mise/tasks/helm/preview-up.sh](https://github.com/tuist/tuist/blob/3fd0ef476259/infra/mise/tasks/helm/preview-up.sh)) is a strong touch and exercises both the ESO and inline license paths end-to-end. License helper centralisation across deployment + migration job is a clean refactor and is backwards-compatible with existing inline-license users. `--default-ssl-certificate` over reflector is the right call for an ephemeral cluster (one Secret, no per-namespace plumbing). `preview-mgmt-rbac.yaml` is appropriately narrow (`get/patch` on a single named Cluster). Onboarding doc §9 is complete and walks the bootstrap end-to-end.

## Headlines

- **One likely bug**: the `kubectl run` smoke pod can't schedule on the tainted preview workers (inline comment).
- **Two design points worth considering before this lands as the long-term shape**: per-PR GitHub Deployments so the PR timeline gets distinct preview cards (the way Render's integration produced), and label-based automatic triggering as a follow-up.
- A few smaller items inline (silent precedence in the license helper, race on the management Cluster CR under concurrent deploys, PR-comment spam on redeploys, minor `commit_sha` validation gap, Redis-disabled assumption to verify).

Nothing here blocks merging the bootstrap workflow IMO; the smoke-test scheduling fix is the only one that will fail on the first real run.

.github/workflows/preview-deploy.yml
```diff
@@ -0,0 +1,295 @@
...
+        run: |
+          kubectl label namespace "$NAMESPACE" \
+            "tuist.dev/expires-at=$EXPIRES_AT" \
+            "tuist.dev/preview=true" \
+            --overwrite
+
+      - name: Smoke test
+        run: |
+          kubectl -n "$NAMESPACE" rollout status deploy/${RELEASE}-tuist-server --timeout=10m
+          kubectl -n "$NAMESPACE" run smoke-$RANDOM --rm -i --restart=Never \
```
fortmarek: **Likely bug: this smoke pod cannot be scheduled.**

Two steps earlier the workflow taints every non-control-plane node with `role=preview:NoSchedule`. The smoke pod is created via `kubectl run` and inherits no tolerations, so it can't land on any worker (and CAPI control-planes are also tainted by default). The cluster only has md-0 workers in the Syself topology; there is no untainted general worker like in the kind harness, so the first real run will hang in `Pending` until `kubectl run`'s implicit timeout.

Fixes:

- `kubectl run ... --overrides='{"spec":{"tolerations":[{"key":"role","operator":"Equal","value":"preview","effect":"NoSchedule"}],"nodeSelector":{"role":"preview"}}}'`, or
- drop the side-pod entirely and `kubectl exec` into the existing server pod / use `kubectl port-forward` from the runner.

.github/workflows/preview-deploy.yml
```diff
@@ -0,0 +1,295 @@
...
+            --image=curlimages/curl --command -- \
+            curl -fsS "http://${RELEASE}-tuist-server/ready"
+
+      - name: Comment on PR
+        if: ${{ inputs.pr_number != '' }}
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr comment "${{ inputs.pr_number }}" --repo "${{ github.repository }}" --body "$(cat <<EOF
```
fortmarek: Per-PR redeploys spam the PR with new comments. Stable URL across pushes is the design intent, but every push gets its own "Preview deployed" comment.

Options: find the previous bot comment (e.g. by a marker string in the body) and edit it via `gh api`, or only post on the first deploy per PR. If the per-PR Deployments suggestion lands, the timeline card largely supersedes the need for the comment anyway.

infra/helm/tuist/templates/_helpers.tpl
```diff
@@ -198,6 +198,33 @@ http://{{ include "tuist.componentName" (dict "root" . "component" "otel-collect
...
+*/}}
+{{- define "tuist.licenseEnv" -}}
+{{- $appSecret := include "tuist.componentName" (dict "root" . "component" "app-secrets") -}}
+{{- $esoSecret := include "tuist.componentName" (dict "root" . "component" "server-external-secrets") -}}
+{{- $useEso := ne (.Values.server.externalSecrets.license.item | default "") "" -}}
+{{- if or $useEso .Values.server.license.key }}
+- name: TUIST_LICENSE_KEY
+  valueFrom:
+    secretKeyRef:
+      name: {{ ternary $esoSecret $appSecret $useEso | quote }}
```
fortmarek, line 219: **Silent precedence when both license sources are set.** If both `externalSecrets.license.item` and `server.license.key` are configured, ESO silently wins via the `ternary` and the inline value never ships. Documented in the helper's doc-comment but worth either:

- raising a `fail` in the helper when both are set, or
- adding a `values.yaml` note that the two are mutually exclusive.

Otherwise an operator who adds a quick inline override in an ESO-enabled environment will be debugging a no-op.

infra/helm/tuist/values-preview.yaml
```diff
@@ -0,0 +1,121 @@
...
+    persistence:
+      size: 5Gi
+    resources:
+      requests:
+        cpu: 100m
+        memory: 256Mi
+      limits:
+        memory: 1Gi
+
+redis:
```
fortmarek, line 111: Confirmed previews don't need any of the Redis-backed paths (rate limiting, channels, etc.)? The PR test plan verified `/ready` and `/`, but those don't necessarily exercise Redis-dependent flows. If anything Redis-backed gets hit during a preview session it'll surface as a runtime error rather than a startup failure.

fortmarek (approved):

.github/workflows/preview-deploy.yml
pepicrft: Good call on the wording — updated the PR body to make manual-only explicit. The label-driven shape (PR labeled `preview` → auto-deploy on push, unlabel/close → teardown) is the right v2; logging it as a separate follow-up rather than expanding scope here.

.github/workflows/preview-deploy.yml
pepicrft: Fixed in 6b3b923 — added `^[0-9a-f]{7,40}$` for `commit_sha` and an explicit `^[0-9]+$` for `pr_number` (was relying on `gh pr view` to fail later).

.github/workflows/preview-deploy.yml
pepicrft: Strong agree. Out of scope for this PR but the right shape: keep `server-k8s-preview` as the single Actions Environment, layer per-PR Deployments via the REST API on top. Logging it alongside the label-driven trigger as the v2 follow-up.

.github/workflows/preview-deploy.yml
pepicrft: Fixed in 6b3b923. Split the scale step into its own `scale-up` job with `concurrency.group: preview-cluster-scale`; the sweep workflow uses the same group name. This serializes the management-CR patches across both workflows but lets the long-running `helm install` in the per-release `deploy` job run in parallel for different PRs.

.github/workflows/preview-deploy.yml
pepicrft: Real bug, good catch. Fixed in 6b3b923. Dropped the in-cluster pod entirely. `kubectl rollout status` already covers internal readiness (the rollout only completes when the pod's own `/ready` probe passes); replaced the curl-from-a-pod with a `curl https://${HOST}/ready` retry loop from the runner, which is a strictly better signal — exercises ingress + wildcard cert + DNS at the same time.

.github/workflows/preview-deploy.yml
pepicrft: Fixed in 6b3b923. Embedded a hidden `<!-- preview-deploy:${RELEASE} -->` marker in the comment body and the workflow now `gh api`s the issue's comments, finds one starting with the marker, and `PATCH`es it (or creates a fresh one if none exists). Stable URL → single comment that updates in place.

infra/helm/tuist/templates/_helpers.tpl
pepicrft: Fixed in 6b3b923 — added a `fail` in the `tuist.licenseEnv` helper when both `externalSecrets.license.item` and `license.key` are set. `helm install` aborts with the message "...are mutually exclusive — pick one license source." instead of silently dropping the inline value.

infra/helm/tuist/values-preview.yaml
pepicrft: Confirmed via the chart values: `redis.enabled` defaults to `false` and the comment in [values.yaml](https://github.com/tuist/tuist/blob/main/infra/helm/tuist/values.yaml#L255-L258) explicitly notes the rate limiter falls back to in-memory bucket in that mode. Anything else Redis-backed (channels, etc.) would surface as a runtime error rather than startup failure — accepting that for previews since the alternative (running an embedded Redis per preview) is overkill for a smoke environment. If a Redis-required flow becomes part of regular preview testing we'd flip `redis.enabled: true` in `values-preview.yaml` (the chart's embedded mode is already wired).

.github/workflows/preview-deploy.yml
pepicrft: Update — folded this into the PR after all (commit 48fa814). The workflow now triggers on `pull_request: [labeled, synchronize, unlabeled, closed]` gated on the `preview` label, with fork-PR filtering. Adding the label deploys, pushing to a labeled PR redeploys, removing the label or closing the PR tears down. workflow_dispatch is preserved for one-off commits / re-deploying a closed PR.

.github/workflows/preview-deploy.yml
pepicrft: Update — folded this in too (commit 48fa814). Each PR-driven preview now creates a transient GitHub Deployment with `environment=preview-pr-<N>`, transitions in_progress → success (with environment_url) → inactive on teardown. The PR timeline gets the "Preview – PR #1234" card the way the Render integration produced. workflow_dispatch runs against a `commit_sha` skip the deployment object (no PR to attach to).

PR #10489, opened 2026-04-27
feat(server): manage multi-region Kura servers

pepicrft:
## Summary

This PR adds account-scoped, multi-region Kura server management to Tuist.

A Kura server is modeled as one account plus one region. Account settings and `/ops` now let Tuist create a regional Kura server, inspect status, see the public endpoint and node readiness, review deployment history, and destroy the server. Active Kura servers mirror their public URL into `account_cache_endpoints`, so operators and account owners manage Kura through the lifecycle instead of editing derived endpoints directly.

Production is controller-backed. The Tuist server records desired state in PostgreSQL, writes `KuraInstance` custom resources to the `kura` namespace, and lets the Go Kura controller reconcile the Kubernetes workload. EU Central uses the server pod's in-cluster ServiceAccount in the existing production cluster. US East and US West use production-synced kubeconfigs so their `KuraInstance` resources land in regional Hetzner workload clusters.

## What Changed

- Adds `Tuist.Kura` as the server context for Kura regions, servers, deployments, cache endpoint mirroring, rollout scheduling, destroy scheduling, node listing, and reconciliation.
- Adds `kura_servers` and `kura_deployments` PostgreSQL tables, plus data export documentation for both tables.
- Adds account settings and `/ops` Kura views for deployment, status, node readiness, endpoint, version, deployment history, and destroy controls.
- Adds `Tuist.Kubernetes.Client` with in-cluster and kubeconfig-backed modes. Local controller testing also uses kubeconfig, so the old `kubectl` transport is gone.
- Adds the `kura.tuist.dev/v1alpha1` `KuraInstance` CRD and `infra/kura-controller`, a Go controller that reconciles ConfigMap, Services, StatefulSet, PodDisruptionBudget, PVC templates, labels, and status.
- Replaces the legacy Helm shell-out provisioner with `Tuist.Kura.Provisioner.KubernetesController`.
- Builds and pushes the Tuist server image, Kura controller image, and Kura runtime image with the same deploy SHA. Helm passes that tag as `TUIST_KURA_RUNTIME_IMAGE_TAG`.
- Removes the GitHub Releases polling path. Runtime rollouts are now driven by deploy-time image tags and the Kura reconciler.
- Removes per-action rollout and destroy Oban jobs. User actions only mutate Postgres, and the minute reconciler is the single lifecycle loop.
- Removes the default Cilium bandwidth cap so we do not launch with an unvalidated noisy-neighbor throttle.

## Region Mapping

The production overlay exposes:

```text
TUIST_KURA_AVAILABLE_REGIONS=eu-central,us-east,us-west
```

Managed Kura maps those product regions to concrete Hetzner-backed clusters:

| Product region | Cluster ID | CAPI cluster | Kubernetes client | Hetzner location |
| `eu-central` | `eu-central-1` | `tuist` | in-cluster ServiceAccount | `fsn1` |
| `us-east` | `us-east-1` | `tuist-kura-us-east` | `TUIST_KURA_KUBECONFIG_US_EAST_1` | `ash` |
| `us-west` | `us-west-1` | `tuist-kura-us-west` | `TUIST_KURA_KUBECONFIG_US_WEST_1` | `hil` |

The public DNS convention is:

```text
{account_handle}-{cluster_id}.kura.tuist.dev
```

Examples: `tuist-eu-central-1.kura.tuist.dev`, `tuist-us-east-1.kura.tuist.dev`, and `tuist-us-west-1.kura.tuist.dev`.

## Operational Requirements

Production needs these runtime pieces before all three managed Kura regions work end to end:

- The `KuraInstance` CRD applied to each workload cluster. The deploy workflow applies `infra/helm/tuist/crds/` to the main production cluster and to each US regional cluster.
- The Kura controller deployed in the main production cluster and in the US regional clusters. Regional clusters use `infra/helm/tuist/values-managed-kura-region.yaml`, which renders only controller resources.
- The production server namespace secret `tuist-kura-kubeconfigs`, created by CI from the 1Password documents `kubeconfig: kura-us-east-1` and `kubeconfig: kura-us-west-1`.
- The `tuist` production cluster with its `fsn1` Kura worker pool for `eu-central`.
- The `tuist-kura-us-east` cluster in `ash` and `tuist-kura-us-west` cluster in `hil`, each with a Kura worker group sized for the default three replicas.
- hcloud CSI, hcloud cloud-controller-manager, external-dns, and the platform chart bootstrapped on the regional workload clusters.
- Deploy workflow access to push `ghcr.io/tuist/kura:<sha-tag>` alongside the server and controller images.

Bootstrap commands for the regional clusters:

```bash
mise run k8s:bootstrap-workload tuist-kura-us-east production "kubeconfig: kura-us-east-1"
mise run k8s:bootstrap-workload tuist-kura-us-west production "kubeconfig: kura-us-west-1"
```

## Tenant And Capacity Model

Kura runs on shared Hetzner Cloud worker pools, not on per-account dedicated hardware. The isolation boundary is application-level:

- Each server has one tenant ID derived from the account handle.
- Each `(account, region)` maps to one `KuraInstance` and one isolated Kubernetes workload set.
- Kura pods are pinned to nodes labeled for the Kura pool.
- Each production Kura cluster has autoscaler bounds for the Kura worker pool.

This PR intentionally targets Hetzner Cloud VMs with `hcloud-volumes`. Bare metal or local NVMe storage can be evaluated later with a separate storage and capacity model.

## Reconciliation Behavior

Create flow:

1. Account settings creates a `kura_servers` row and an initial `kura_deployments` row for the selected account, region, and current runtime image tag.
2. `Tuist.Kura.Reconciler` runs every minute. If the backing `KuraInstance` is missing or has image drift, it applies the desired `KuraInstance` to the target region cluster and marks the deployment `:running`.
3. The reconciler keeps observing `KuraInstance.status` until Kubernetes reports the requested image as ready.
4. When observed state matches desired state, the reconciler marks the deployment `:succeeded`, marks the server `:active`, and updates the account cache endpoint list.
5. If applying the desired state fails, the deployment and server move to failed states.

Destroy flow:

1. Account settings marks the server `:destroying` and removes the cache endpoint up front.
2. The reconciler observes destroying servers every minute. If Kubernetes still has the `KuraInstance`, it requests deletion. If Kubernetes returns `:not_found`, it marks the server `:destroyed`.

Automatic update flow:

1. A production deploy builds `ghcr.io/tuist/kura:<sha-tag>` and sets `TUIST_KURA_RUNTIME_IMAGE_TAG` on the server deployment.
2. The reconciler schedules deployments for active Kura servers whose observed image tag differs from that runtime tag.
3. The same reconciler loop rolls customer Kura pods to the lockstep runtime image.

## Testing And Validation

- `mise x elixir@1.19.5 -- mix test test/tuist/kubernetes/client_test.exs test/tuist/kura_test.exs test/tuist/kura/reconciler_test.exs test/tuist/kura/regions_test.exs test/tuist/kura/provisioner/kubernetes_controller_test.exs test/tuist/kura/deployment_test.exs test/tuist/kura/server_test.exs test/tuist_web/live/account_settings_live_test.exs`
- `helm template` production render verified `TUIST_KURA_RUNTIME_IMAGE_TAG` is set on the server deployment and not on the processor.
- `helm template` regional render verified the controller-only release for Kura regional clusters.
- Workflow YAML parse for `.github/workflows/server-deployment.yml` and `.github/workflows/server-production-deployment.yml`.
- `git diff --check`.


pepicrft:
@fortmarek I just realized I'm missing the provisioning of nodes. The logic assumes the cluster will have nodes available on which we can schedule Kura instances.

fortmarek:
Coming from #10499 (xcresult / Mac mini fleet), a few things on the provisioning model + bandwidth story worth flagging. None of it blocks this PR's current scope.

## Logical vs physical tenant boundary

The Helm-release-per-tenant boundary here is a *logical* boundary, not a *physical* one:

- `topologySpreadConstraints` keep one release's three replicas on three distinct nodes, but nothing keeps different accounts' releases off the same workers. Once two tenants share a node, they share the host NIC.
- No `kubernetes.io/egress-bandwidth` / `ingress-bandwidth` annotations, no Cilium Bandwidth Manager wiring. For a cache that's bursty and latency-sensitive, that's the noisy-neighbor surface that shows up as "Kura got slow" once the cluster has more than a handful of active tenants.

For a Dedicated tier later, a separate Helm release alone won't buy hardware isolation if both releases land on the same workers. Hardware isolation needs node-level scheduling: a per-account `MachineDeployment` with taints + tolerations + a matching `nodeSelector`. That's the pattern #10499 puts in place for the macOS fleet (`tuist.dev/account=<id>:NoSchedule` + per-account `MachineDeployment`); the same shape fits a Linux Kura fleet via a sibling provider, covered in #10499's "Co-located Linux fleet for Kura cache nodes" section.

Cheap follow-ups:

- Add ingress/egress bandwidth annotations to the StatefulSet template, sized for the pool tier (cluster CNI is Cilium, which honors them).
- Document explicitly that tenant separation today is application-layer (`KURA_TENANT_ID`) + per-tenant Helm release + per-pod PVC, not hardware-level. Helps whoever next triages a noisy-neighbor incident.

## Mesh topology: backbone + edge

Kura is a peer-to-peer mesh, so backbone-vs-colocation isn't either/or:

- **Hetzner**: high-bandwidth multi-tenant backbone, holds the bulk of the cache.
- **Scaleway** (per #10499's colocation idea): smaller edge fleet in `fr-par-1` next to the Mac mini runners, hot-only, peer-replicates from backbone on miss.
- Runners hit edge first (sub-ms RTT), miss falls through to backbone (~15ms one-time, warmed at edge after). External CI hits backbone directly.

Three things worth verifying before committing this shape:

1. **Cross-cluster peering is a one-knob change**: extending `KURA_PEERS` / DNS discovery to span Hetzner + Scaleway StatefulSets shouldn't need protocol-level work. Worth a quick check that adding a Scaleway peer to a Hetzner-based release falls out of the existing peer config. (Routing efficiency itself isn't a concern at our scale: 3-4 nodes per customer per release, so even broadcast-on-miss is bounded at O(n=4).)
2. **Cross-provider egress**: only pencils out if the hot-set is much smaller than total cache. Typical for build artifacts but worth modeling.
3. **Failure semantics**: edge down = sub-ms degrades to ~15ms; backbone down = edge serves what it has, cold misses fail. Different incident shape than "one cluster down."

## Bare-metal Hetzner for the backbone (follow-up)

For Kura's workload (bandwidth + IOPS bound, CPU incidental), Hetzner bare metal is meaningfully better than Cloud VMs at similar price points: dedicated NIC (10/25 Gbps options without sharing), local NVMe (5-7 GB/s sustained vs ~200-400 MB/s on `hcloud-volumes`), no co-tenant interference. `cluster-api-provider-hetzner` from Syself supports the Robot API, so the substrate stays one CRD pattern.

Provisioning model has to flip though: ~10 min auction-host time means we can't spin up a 3-node fleet on opt-in per account. Right shape there is a small pool of bare-metal hosts running shared Kura StatefulSets with the bandwidth shaping above; per-tenant separation comes from `KURA_TENANT_ID` + Cilium limits, not per-account hardware. Persistence likely flips to `emptyDir` since cache is rebuildable and peer-replication warms a fresh pod from peers.

Suggest a focused load test on AX52 + 10 Gbps add-on vs CCX33 before committing direction. Separate follow-up, not blocking this PR.

## On the "Kubernetes controller + CRDs: rejected for now" framing

The "second control plane" argument is the strongest part of the current call, but #10499 introduces a CAPI provider with custom CRDs + controller-runtime as part of the runners substrate. Once that's in flight, the marginal cost of one more CRD (a `TuistKura` resource owned by a small Go controller in the same chart) is much lower than bootstrapping operator infra from zero.

The "swap behind the provisioner behaviour later" escape hatch tends not to close cleanly in practice. The LiveView surfaces `helm install` stdout as `kura_deployment_log_lines` rows; the Oban worker contracts on rollout-stage transitions; `account_cache_endpoints` is mirrored from Postgres state. Each is shaped by the helm-shell-out impl, and the swap gets re-priced as a multi-week migration rather than the contained change the abstraction promises once it has accreted callers.

Worth considering: ship this as-is, *and* commit to migrating to a CRD-backed flow once the CAPI infrastructure from #10499 lands, rather than treating the swap as conditional. Otherwise the LiveView + Oban code accretes around the helm-shell model for another quarter and the swap becomes prohibitive even when the operator infra is sitting right there.

pepicrft:
@fortmarek here's where the review landed:

- **Tenant boundary**: pivoted to a shared Hetzner Robot pool (`tuist.dev/role=kura`), application-layer `KURA_TENANT_ID`, and per-pod Cilium bandwidth caps (small 100M / medium 250M / large 500M from `Tuist.Kura.Specs`). Per-account `MachineDeployment` doesn't fit Robot's ~10 min provisioning, as you flagged.
- **P1s**: switched ClickHouse reads from `IngestRepo` to `ClickHouseRepo`. Kura tables documented in `data-export.md`.
- **Mesh topology / cross-cluster peering**: inline note on `kura.seedPeers` confirms the runtime accepts comma-separated peers, so Hetzner backbone + Scaleway edge is a future `extraSeedPeers` chart knob, no protocol-level work.
- **CRD migration**: not in this PR. The PR description now lists the accretion points (helm-stdout mirror, Oban rollout-stage transitions, `account_cache_endpoints` mirror) and the adoption path (emit `TuistKura` CRs from existing `kura_servers` rows, controller adopts release-named resources via `ownerReferences`, PVCs survive). I'd rather wait until CAPH (#10586) is already in-cluster so we're not bootstrapping operator infra from zero. Happy to push earlier if you'd argue for it.
- **Bandwidth defaults + bare-metal direction**: picked from first principles, unvalidated. Will tune with real load before prod rollout.

CI is greening up after `a6990ec0b9` (format / credo / Trivy pass). Mind taking another look?

pepicrft:
Quick pass on the four "findings outside this PR's diff" from Blick. All four turn out to be false positives because Blick couldn't see the full file context, but worth confirming:

1. **`router.ex:0` Unclear authorization for new /ops Kura routes** — the new Kura `live` route is inside the existing `:ops` `live_session` (`router.ex:697-708`), which already has `[:current_user, :read, :ops]` authorization. Same gate as every other /ops route.
2. **`authorization.ex` Verify Kura objects added to `@project_categories`** — n/a. Kura management is /ops-only LiveView, not a public API. There's no Kura create/destroy endpoint exposed to CLI/account tokens, so no project-scoped category to register.
3. **`authorization.ex` Verify create action has `scopes_permit` check for account tokens** — same as above. Account tokens never call a Kura management endpoint; the CLI only resolves the public Kura URL through the existing cache-endpoint flow, which already has its own auth on Kura via the Lua extension.
4. **`create_kura_tables.exs:0` Migration timestamp column types not visible in diff** — verified: all `timestamps()`, `started_at`, `finished_at` use `:timestamptz` per the project rule (`migration.ex:29, 64, 65, 67`). Ecto schemas use `:utc_datetime` per the `lib/` rule. Both consistent.

No code change needed for any of these.

fortmarek:
Re-review of head 629cd4a9f1.

## Addressed since last review ✅

- `RolloutWorker` account-scoping (now symmetric with `DestroyServerWorker`)
- Reconciler cross-tenant scan documented inline
- `chart_path/1` trust model documented
- Trivy ignorefile passed explicitly
- Syself references dropped from chart values

## Outstanding

### P0 — live log tail won't update

[`ops_account_kura_deployment_live.html.heex` L51](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/server/lib/tuist_web/live/ops_account_kura_deployment_live.html.heex#L51)

`phx-update="ignore"` tells LiveView to leave the container's children alone after first render. The polling logic at [`ops_account_kura_deployment_live.ex` L60-L79](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/server/lib/tuist_web/live/ops_account_kura_deployment_live.ex#L60-L79) keeps appending to `@log_lines`, but those updates won't reach the DOM — /ops will appear frozen on whatever lines were present when the terminal first rendered. This is the headline /ops feature.

Suggested fix: `phx-update="stream"` with `stream(:log_lines, ...)`, or `phx-update="append"` with `id={"log-line-#{line.sequence}"}` per row.

### P2 — handle → DNS/release name unvalidated

[`helm_kubernetes.ex` L365](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/server/lib/tuist/kura/provisioner/helm_kubernetes.ex#L365)

`String.downcase` alone doesn't enforce RFC1123 (lowercase alphanumeric + `-`, ≤53 chars for the helm release name). An account name with `_`, `.`, or one that pushes `kura-{handle}-{cluster_id}` past 53 chars will produce an opaque helm error at rollout time rather than a clear validation error at create time.

Consider adding a `validate_format/3` on `Server.create_changeset/2` that mirrors RFC1123 label rules + length.

### P2 — `parse_id/1` is loose

[`ops_account_kura_deployment_live.ex` L81-L86](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/server/lib/tuist_web/live/ops_account_kura_deployment_live.ex#L81-L86) and [`ops_account_live.ex` L72-L77](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/server/lib/tuist_web/live/ops_account_live.ex#L72-L77)

`{n, _}` matches `Integer.parse("0abc")` → `{0, "abc"}`, so `/ops/accounts/0abc/...` resolves to account ID `0` and 404s downstream rather than rejecting trailing garbage. Prefer matching `{n, ""}` strictly.

### P3 — sequence numbering split across two sources of truth

[`rollout_worker.ex` L108-L122](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/server/lib/tuist/kura/workers/rollout_worker.ex#L108-L122)

The process dict is the source of truth on the happy path, but `fail_running/1` and the synthetic-failure-after-streamed-output path call `next_sequence/1` outside `log_sink/1`. In those paths the fallback queries ClickHouse via `Kura.next_log_sequence/1`, which returns `1` on any error — under a transient ClickHouse outage you'd get sequence `1` written on top of existing rows. MergeTree won't dedupe; the tail would show out-of-order lines but no data loss.

Threading a small sequencer through `inputs.on_log_line` would make the source of truth singular.

### P3 — `ensure_cache_endpoint` invariant uncommented

[`kura.ex` L345-L352](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/server/lib/tuist/kura.ex#L345-L352)

`on_conflict: :nothing` keyed on `(account_id, technology, url)` is correct *because* URLs are deterministic per `(account, region)` today — a stale row can't outlive a destroy → re-create cycle. Worth a one-line comment locking in that assumption to protect future provisioner authors who might generate non-deterministic URLs.

---

The P0 (live log tail) is the only one I'd block on; the rest are polish that can land in a follow-up.

fortmarek:
One more thing worth pinning down before this lands — apparent contradiction between the Hetzner overlay's storage class and the bare-metal node selector.

[`values-managed-provider-hetzner.yaml`](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/kura/ops/helm/kura/values-managed-provider-hetzner.yaml) pins Kura pods to bare-metal Robot nodes:

```yaml
nodeSelector:
  tuist.dev/role: kura       # AX/EX hosts onboarded via CAPH
```

But the same overlay also sets:

```yaml
persistence:
  storageClass: hcloud-volumes
  size: 200Gi
```

`hcloud-volumes` is the [Hetzner Cloud CSI driver](https://github.com/hetznercloud/csi-driver). It attaches Cloud-API-managed block volumes to Hetzner Cloud VMs (CCX/CPX) at ~200–400 MB/s over the network. It cannot attach volumes to Robot bare-metal servers — Robot servers aren't visible through the Hetzner Cloud API, so the CSI driver has nothing to call.

[`values-managed.yaml` L33-L39](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/kura/ops/helm/kura/values-managed.yaml#L33-L39) reinforces the contradiction by anchoring the `ReadWriteOncePod` access mode to "the Hetzner CSI driver on the cluster-stack version we run (>= 1.34)", which is the Cloud driver.

So one of these has to be wrong:

1. **Robot was the intent** — then `storageClass` needs to point at local NVMe (e.g. `local-path-provisioner`, OpenEBS LocalPV, or Rook on local disks). The PR description's "5–7 GB/s sustained NVMe" rationale assumes this.
2. **Cloud VMs were the intent** — then the `tuist.dev/role=kura` pool is actually CCX/CPX instances, and the PR description's "bare metal AX/EX hosts" / "10/25 Gbps dedicated NIC" / "local NVMe" framing doesn't match what'll actually run.

The performance argument for bare metal mostly evaporates if persistence routes through `hcloud-volumes` — you keep the dedicated NIC but lose the IOPS win, which is the bigger half of the equation for a cache. Worth resolving the inconsistency before rollout, since downstream decisions (per-spec PVC sizes, Cilium bandwidth caps sized against NIC capacity, the "shared pool + Cilium caps" tenancy story) all assume one substrate or the other.

fortmarek:
Resolving my earlier storage-class comment: per offline confirmation the actual substrate decision is Hetzner **Cloud VMs**, not Robot bare metal. So the chart is correct as-is (`storageClass: hcloud-volumes` + `tuist.dev/role=kura` label propagated through a future `hcloud-worker` MachineDeployment) and what needs updating is the PR description + module docs:

- PR description's "Tenant model" section: drop "Hetzner Robot AX/EX hosts" / "shared bare-metal pool" framing; this is a Cloud worker pool.
- PR description's "Why bare-metal Hetzner instead of Cloud VMs?" section: remove or invert — the answer turned out to be Cloud VMs.
- [`Tuist.Kura.Provisioner.HelmKubernetes` moduledoc](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/server/lib/tuist/kura/provisioner/helm_kubernetes.ex#L29-L34) similarly references "shared bare-metal pool of Hetzner Robot nodes".
- [`values-managed-provider-hetzner.yaml` L8-L14](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/kura/ops/helm/kura/values-managed-provider-hetzner.yaml#L8-L14) comment talks about "Hetzner Robot AX/EX hosts onboarded via cluster-api-provider-hetzner".

One concrete follow-up beyond doc cleanup: the [Cilium bandwidth caps in `Tuist.Kura.Specs`](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/server/lib/tuist/kura/specs.ex#L23-L48) (small 100M / medium 250M / large 500M) are commented as "sized for a 1 Gbps Hetzner AX NIC with light oversubscription". For a `ccx`-class Cloud VM that NIC budget is different (and shared at the hypervisor), so the caps may need re-tuning once we have load data — or at minimum the comment justifying the chosen values needs updating to match the real substrate.

Also worth confirming explicitly: there's still no `md-kura` MachineDeployment in any of the per-env Cluster CRs (`infra/k8s/clusters/cluster-{staging,canary,production,preview}.yaml`), so before `/ops` Kura works in production, that pool needs to be added.

fortmarek:
A question on the operational direction for capacity, not blocking — wanted to surface it before the Kura MachineDeployment lands in the cluster CR since the choice locks in some adjacent wiring.

The PR's tenant model relies on a pre-warmed shared pool of Kura-pool nodes. There's no per-customer K8s-node provisioning; helm `--wait` is the only thing keeping a rollout honest, and a full pool causes the third StatefulSet replica to stay Pending and the deployment to fail at `HELM_TIMEOUT=10m`. So the question is what the steady-state capacity story looks like.

Two shapes I can see:

**(A) Static `replicas: N`, operator-driven**
Just `replicas: N` in `cluster-{env}.yaml`, no autoscaler annotations. Same model as `md-0` today. New customer beyond the preconfigured headroom = operator opens a PR bumping `replicas`, mgmt cluster reconciles. Simple, predictable, but every burst becomes manual toil and the failure mode for "we're full" is a customer-facing rollout timeout in /ops.

**(B) Cluster-autoscaler-bounded, like `md-processor`**
```yaml
metadata:
  annotations:
    cluster.x-k8s.io/cluster-autoscaler-node-group-min-size: "N"
    cluster.x-k8s.io/cluster-autoscaler-node-group-max-size: "M"
```
`min-size` carries the always-warm headroom you actually want; CA scales up to `max-size` when a new customer's Pending pods can't fit. Bursts are absorbed without a PR; cap stays in the CR.

There's one gotcha if we pick (B): cluster-autoscaler needs the pool's labels and taints declared up front via `node.cluster.x-k8s.io/node-template/{label,taint}` annotations on the MD, otherwise CA can't match a Pending Kura pod (which carries the `tuist.dev/role=kura` toleration + nodeSelector) to the right node group and won't scale up. So (B) needs a small extra annotation block on top of the autoscaler bounds — easy, just non-obvious.

My read is (B) is the right shape long-term — Kura pods are per-customer and arrive in 3-replica bursts, which is exactly the pattern CA handles well — but doing (A) for the first few customers while load characteristics settle is also reasonable.

What does the rollout plan look like? Is the intent to pick a `min-size` aligned with the spec catalog (e.g. enough always-warm nodes for K medium-spec customers) and tune from telemetry, or is the plan operator-toil for the first phase?

pepicrft:
@fortmarek yes, we should go with option B here: Hetzner Cloud VMs, not Robot bare metal.

I updated the PR body and the stale module/chart comments so the storage class + worker-pool story now matches what this PR actually deploys: `hcloud-volumes` on a dedicated Cloud worker pool labelled `tuist.dev/role=kura`.

I was thinking we should add the actual `md-kura` MachineDeployment in a follow-up PR. This PR can land the product control plane and chart wiring, and the follow-up can make the per-env cluster change with the exact Cloud worker type, autoscaler bounds, taints, and labels.

fortmarek:
Pulling on the single-region thread from above — this is a meaningful omission for a tier being positioned as a low-latency cache mesh, and I don't think it should ship to customers in this state.

Current reality:

- `Tuist.Kura.Regions.all/0` returns `[eu, local]`. The `eu` region resolves to `cluster_id: "eu-1"` which points at the same fsn1 (Hetzner Falkenstein) workload cluster the central Tuist server already runs on.
- A US / APAC / SA customer's RTT to Kura is identical to their RTT to `tuist.dev`. Kura adds operational surface — helm rollouts, /ops state machine, ClickHouse log pipeline, per-account StatefulSets, per-account public hostnames, TLS automation — without any geographic locality benefit for those customers.
- fsn1 going down takes the central server *and* every customer's Kura with it. The blast radius strictly grows.

For a feature whose pitch is bandwidth-bound CI workloads getting closer-to-CI cache reads, EU-only inverts the value prop for a large fraction of customers. They'd be paying for a tier that adds latency-without-locality and inherits the central server's failure surface on top of its own.

Two viable rollout shapes I can see:

1. **Multi-region at GA**: launch with at least US-east + EU (APAC ideally), so the tier delivers what it advertises everywhere it's sold.
2. **EU-only with explicit scope**: launch the tier with an upfront geography restriction on the marketing surface ("Kura is currently available for EU-hosted CI; US/APAC on the roadmap"), customers self-select, no over-promising.

The PR's shape is multi-region-ready — `Regions` is a list, per-region kubeconfigs are structurally separate, public URL templates are per-region — so this isn't a protocol or architecture gap. What's missing is operational expansion:

- A second Cluster CR in a non-fsn1 Hetzner location (Hetzner Cloud has `ash` US-east, `hil` US-west, `sin` Singapore)
- That cluster's bootstrap (Cilium, HCCM, hcloud-csi, ESO, ingress) and its kubeconfig synced via 1P/ESO under `kura.kubeconfigs.{region}_1`
- A second `Regions.eu_region/0`-shaped entry
- DNS + TLS automation extended to a new `*.kura.tuist.dev` apex per region
- A Kura MachineDeployment in each new cluster CR

Plus Pedro's earlier note that `kura.seedPeers` handles cross-region peering without protocol work, so inter-region replication is a chart-values knob, not engineering.

Worth pinning down before the marketing surface goes live: what's the multi-region timeline relative to external GA, and which of the two rollout shapes are we picking?

fortmarek:
Pulling on the single-region thread from above — this is a meaningful omission for a tier being positioned as a low-latency cache mesh, and I don't think it should ship to customers in this state.

Current reality:

- `Tuist.Kura.Regions.all/0` returns `[eu, local]`. The `eu` region resolves to `cluster_id: "eu-1"` which points at the same fsn1 (Hetzner Falkenstein) workload cluster the central Tuist server already runs on.
- A US / APAC / SA customer's RTT to Kura is identical to their RTT to `tuist.dev`. Kura adds operational surface — helm rollouts, /ops state machine, ClickHouse log pipeline, per-account StatefulSets, per-account public hostnames, TLS automation — without any geographic locality benefit for those customers.
- fsn1 going down takes the central server *and* every customer's Kura with it. The blast radius strictly grows.

For a feature whose pitch is bandwidth-bound CI workloads getting closer-to-CI cache reads, EU-only inverts the value prop for a large fraction of customers. They'd be paying for a tier that adds latency-without-locality and inherits the central server's failure surface on top of its own.

Two viable rollout shapes I can see:

1. **Multi-region at GA**: launch with at least US-east + EU (APAC ideally), so the tier delivers what it advertises everywhere it's sold.
2. **EU-only with explicit scope**: launch the tier with an upfront geography restriction on the marketing surface ("Kura is currently available for EU-hosted CI; US/APAC on the roadmap"), customers self-select, no over-promising.

The PR's shape is multi-region-ready — `Regions` is a list, per-region kubeconfigs are structurally separate, public URL templates are per-region — so this isn't a protocol or architecture gap. What's missing is operational expansion:

- A second Cluster CR in a non-fsn1 Hetzner location (Hetzner Cloud has `ash` US-east, `hil` US-west, `sin` Singapore)
- That cluster's bootstrap (Cilium, HCCM, hcloud-csi, ESO, ingress) and its kubeconfig synced via 1P/ESO under `kura.kubeconfigs.{region}_1`
- A second `Regions.eu_region/0`-shaped entry
- DNS + TLS automation extended to a new `*.kura.tuist.dev` apex per region
- A Kura MachineDeployment in each new cluster CR

Plus Pedro's earlier note that `kura.seedPeers` handles cross-region peering without protocol work, so inter-region replication is a chart-values knob, not engineering.

Worth pinning down before the marketing surface goes live: what's the multi-region timeline relative to external GA, and which of the two rollout shapes are we picking?

fortmarek:
Stepping back on the substrate question — the helm shell-out from the application runtime is the load-bearing design choice that drags in everything else (admin-cert kubeconfig, helm/kubectl/bash in the container, chart vendoring, the `kura_deployment_log_lines` pipeline, the `phx-update="ignore"` log tail, the process-dict sequencer, the cluster-wide `concurrency: 1` queue). It's also out of step with the pattern that just landed in [#10653](https://github.com/tuist/tuist/pull/10653) for the runners warm pool, which is doing a strictly harder problem in a much cleaner shape. I'd push to converge with that pattern before this merges rather than carry it into prod and migrate later.

What [#10653](https://github.com/tuist/tuist/pull/10653) demonstrates is already in the codebase as a reusable primitive:

- [`Tuist.Kubernetes.Client`](https://github.com/tuist/tuist/blob/60a9ce2b54acb3cef0b56a36225d48ca28583496/server/lib/tuist/kubernetes/client.ex) — ~130 lines, native HTTPS via `Req` over `Tuist.Finch`, in-cluster ServiceAccount auth from `/var/run/secrets/kubernetes.io/serviceaccount/{token,ca.crt}`. No subprocess, no Port, no tmp files, no kubeconfig handling.
- A namespace-scoped `Role` + `RoleBinding` ([`runners-namespace.yaml`](https://github.com/tuist/tuist/blob/60a9ce2b54acb3cef0b56a36225d48ca28583496/infra/helm/tuist/templates/runners-namespace.yaml)) that grants exactly the verbs the application needs, nothing more.
- A reconciler-driven model where the server creates K8s objects directly, observes them via `list_*`/`get_*`, and mirrors observed state into Postgres rows. No log-stream pipeline because there's no helm to capture stdout from.

I think Kura should adopt the same shape in this PR. Concretely:

**Should be in this PR:**

1. **Drop the admin-cert kubeconfig from 1P; use an in-cluster ServiceAccount with a narrow Role** scoped to the `kura` namespace. Today the kubeconfig is the one `clusterctl get kubeconfig` mints — cluster-admin equivalent — sourced via ESO and written to a tmp file. A Phoenix code-execution vulnerability becomes a cluster compromise. Mirror what `runners-namespace.yaml` does for the runners namespace: a Role granting `statefulsets, services, ingresses, configmaps, secrets, persistentvolumeclaims, poddisruptionbudgets` for the `kura` namespace plus whatever cluster-scoped reads are actually needed, bound to the server's ServiceAccount. This alone defangs the security blast radius.

2. **Generalize `Tuist.Kubernetes.Client`** with the verbs Kura needs on top of what runners already uses: `get/2`, `replace/3` (PUT), `patch/3` (JSON Patch), `delete/2`. Same in-cluster auth, same `Req`/`Finch` transport. Cheap addition, reusable for future controllers in the codebase.

3. **Replace helm with native manifest application.** Stripped of provider overlays and conditionals, the per-account release is six resources: StatefulSet, Service, Ingress, PDB, ConfigMap (Lua hook), and the StatefulSet's `volumeClaimTemplates` (PVCs are managed by the StatefulSet controller). Build a `Tuist.Kura.Manifests` module that returns those as plain Elixir maps for a given `(server, account, region, image_tag)` — same shape as `Tuist.Runners.PodSpec` but for six resources instead of one. The `RolloutWorker` becomes "render manifests, replace each via the client, drive rollout state". Drop `priv/kura_chart`, drop helm/kubectl/bash from `runner-base` in the Dockerfile. The Lua hook ships as a regular release asset (à la `priv/static/skills`).

4. **Replace the helm-stdout log pipeline with native status mirroring.** The `kura_deployment_log_lines` ClickHouse table, the 2s-polling LiveView log tail, the process-dict sequencer in `RolloutWorker.log_sink/1` — all of that exists *because* helm shells out and stdout has to surface somewhere. Once the rollout is native API calls, deployment status can be driven by a periodic reconciler that GETs the StatefulSet and compares `.status.{updatedReplicas, readyReplicas, currentRevision}` against the deployment's `image_tag`. `/ops` shows the deployment row + a Grafana deep link for pod logs (which already ship via `k8s-monitoring`). Two tangential wins from this: the P0 `phx-update="ignore"` bug from the review disappears (no log tail to break), and the P3 sequence-numbering split disappears (no log lines).

**Optional / defer:**

5. **Port `rollout.sh`'s partitioned warm-rollout state machine** to Elixir. The algorithm is straightforward — walk ordinals N-1→0, PATCH `spec.updateStrategy.rollingUpdate.partition`, wait for the pod's `controller-revision-hash` label to flip, poll `/status/rollout` on each pod for the health gate (`outbox` and `fd_timeouts` deltas vs baseline). This is the same logic [`kura/ops/rollout/gate.sh`](https://github.com/tuist/tuist/blob/629cd4a9f19ab3335d8db339f9bba6f18d650d1b/kura/ops/rollout/gate.sh) implements today, but as an explicit Elixir state machine. For first installs you don't need this — a vanilla rolling update is fine. It only matters for version upgrades, which are operator-driven and infrequent. Reasonable to ship the rest first and add this back as the upgrade path matures.

What this looks like net:

| Today | After |
|---|---|
| helm + kubectl + bash in container | nothing extra |
| Admin-cert kubeconfig from 1P → ESO → tmp file | In-cluster SA, namespace-scoped Role |
| `priv/kura_chart` vendored at build | Elixir manifest renderer |
| Erlang Port + tmp files + chmod | `Req` over `Tuist.Finch` |
| Stdout per-line → ClickHouse → LiveView poll | `.status` mirrored, Grafana for logs |
| `kura_deployment_log_lines` table + 2s poll | Gone |
| `:kura_rollout` queue serialized cluster-wide | Per-server reconciliation, parallel across servers |
| P0 `phx-update="ignore"` | N/A |
| P3 sequence-numbering split | N/A |

The total code delta is a deletion, not an addition — the chart vendoring, the log-stream pipeline, the LiveView log tail, the kubeconfig dance, the rollout-stdout sequencer, and the test coverage attached to all of those goes away. What's added is one Manifests module + a few verbs on the existing Kubernetes client + a status reconciler.

The application-domain code (`Tuist.Kura` context, the schemas, state machines, `Regions` catalog, `Specs` catalog, `account_cache_endpoints` mirroring, the Lua hook, the `Reconciler` for orphaned rows, `/ops` LiveView shell) all stays. The surgery is purely on the cluster-comms substrate.

I'd really push to do (1)-(4) in this PR rather than ship and migrate later — (1) is a security-review concern that's hard to walk back once production has the cluster-admin secret deployed, and (2)-(4) get the team converged on one Kubernetes-comms pattern (`Tuist.Kubernetes.Client`) instead of two. Doing it later means duplicate test coverage for the helm path that gets deleted, plus another PR's worth of operational migration. Doing it now makes this PR smaller, not bigger, because the deletions outweigh the additions.

pepicrft:
@fortmarek I pushed the final follow-up and wanted to make the mapping to your review explicit.

Addressed points:
- Helm shell-out and admin kubeconfig: replaced the runtime Helm path with an in-cluster KuraInstance controller. The server now uses Tuist.Kubernetes.Client and its production ServiceAccount. The Kura controller owns the Helm work inside the cluster.
- Narrow production permissions: added namespace-scoped RBAC for the server to manage KuraInstance resources and controller RBAC for Kura resources, instead of passing a cluster-admin kubeconfig through 1Password/ESO/tmp files.
- Log tail, ClickHouse log rows, and phx-update issue: removed that whole path. /ops now shows deployment state and links to Grafana for pod logs. There is no LiveView log tail or sequence-numbering pipeline anymore.
- Validation hardening: tightened ID parsing, validated the provisioner node ref as an RFC1123 Kubernetes label, made the Kura server row lock account-scoped, and changed the stranded deployment reconciler to grouped update_all calls instead of per-row DB writes.
- Substrate and storage: aligned the implementation with Hetzner Cloud worker nodes, hcloud-volumes, and a CAPI topology worker named kura. The old bare-metal framing is gone. I also avoided md-kura in the topology name.
- Capacity: added the production Kura worker pool with node.cluster.x-k8s.io/pool=kura, autoscaler bounds, and current CAPI autoscaler annotations.
- Region rollout: production now exposes eu-central, us-east, and us-west through TUIST_KURA_AVAILABLE_REGIONS. The account settings page can offer all managed regions that are not already occupied by the account.
- DNS and data path: changed Kura exposure from nginx Ingress to a per-KuraInstance Service type LoadBalancer. external-dns watches the Service hostname, so tuist-*.kura.tuist.dev resolves to that instance-specific Hetzner LB rather than the shared ingress controller. Runtime traffic no longer goes through the shared nginx ingress path; it goes DNS -> per-instance Hetzner LB -> Service/NodePort with ExternalTrafficPolicy=Local -> the local Kura pod on port 4000. This still uses a per-instance LB, not raw pod or node IP DNS. The controller also deletes any legacy Kura Ingress for the instance.
- Dashboard source of truth: added infra/grafana-dashboards/tuist-kura.json, named tuist-kura with the title Tuist Kura.

Local validation is passing on 8ed1c0ed99: focused Kura tests, Kura controller go test/go vet, Credo, format, dashboard JSON validation, and Helm render checks. PR checks restarted after the amend.

fortmarek:
Re-reviewing head `8ed1c0ed99`. This is a substantial architectural rewrite that lands almost everything from the prior reviews — the surgery on the cluster-comms substrate landed in full, plus a bunch more.

## What the rewrite addresses ✅

**The shelling-out smell is gone.** Helm/kubectl/bash dropped from the container at [`server/Dockerfile`](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/server/Dockerfile) (only the Lua hook ships now, at `priv/kura/hooks/tuist.lua`). The chart vendoring at `priv/kura_chart` is gone. The new flow is exactly the shape #10653 demonstrated:

- A new [`KuraInstance` CRD](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/infra/helm/tuist/crds/kura.tuist.dev_kurainstances.yaml) carries the desired state.
- A new Go controller in [`infra/kura-controller/controllers/kurainstance_controller.go`](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/infra/kura-controller/controllers/kurainstance_controller.go) owns the StatefulSet/Service/PDB/ConfigMap reconciliation.
- The server speaks K8s natively via [`Tuist.Kubernetes.Client`](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/server/lib/tuist/kubernetes/client.ex) using in-cluster ServiceAccount auth and server-side-apply.
- The new [`Tuist.Kura.Provisioner.KubernetesController`](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/server/lib/tuist/kura/provisioner/kubernetes_controller.ex) is the production path; `HelmKubernetes` is now dev-only.

**The log-streaming pipeline is gone.** `kura_deployment_log_lines` ClickHouse table deleted, `Tuist.Kura.DeploymentLogLine` schema deleted, the LiveView log tail rewritten ([`ops_account_kura_deployment_live.ex` L88-L104](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/server/lib/tuist_web/live/ops_account_kura_deployment_live.ex#L88-L104)) to a Grafana Explore deep link. The `phx-update="ignore"` P0 disappears with the table. The process-dict sequencer is gone with the streaming.

**Cluster-wide rollout serialization is gone.** The `:kura_rollout` queue is gone — `RolloutWorker` now lives on `:default`. Each rollout creates a CR; the controller handles serialization at the cluster level via the CR's reconciliation loop.

**Multi-region catalog wiring is in.** `Tuist.Kura.Regions` now has three managed regions (`us-east`, `us-west`, `eu-central`) plus the local-controller dev region. Production exposes them via `TUIST_KURA_AVAILABLE_REGIONS=eu-central,us-east,us-west`.

**Capacity / autoscaling addressed.** A dedicated `kura` MachineDeployment landed in `cluster-production.yaml` with cluster-autoscaler bounds (3–12 replicas, ccx13 dedicated-vCPU pool).

**Direct LB per instance.** The controller drops the Ingress entirely and writes a per-instance `Service` of type `LoadBalancer` with Hetzner-managed certs, `external-dns` annotations, `ExternalTrafficPolicy=Local` for source-IP preservation. Removes the ingress-controller hop on the cache hot path.

**Polish from the prior review:**
- `parse_id/1` is now strict `{n, ""}` in both LiveViews.
- `provisioner_node_ref` validated as RFC1123 label, ≤53 chars ([`server.ex` L44-L45](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/server/lib/tuist/kura/server.ex#L44-L45)).
- `Reconciler` is now batched (single Oban.Job lookup, single `update_all`), and idempotent against the case where the controller activated the server before the BEAM crashed.
- `Tuist.Kubernetes.Client` carries a dev-mode `kubectl` shim for local testing alongside the in-cluster path.

Bonus: a [Grafana dashboard](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/infra/grafana-dashboards/tuist-kura.json) and a `VersionMonitorWorker` hourly cron that auto-schedules deployments when a new Kura release lands and active servers are behind. Customer-facing controls live in `account_settings_live.ex` (gated by FunWithFlags `:kura`); `/ops` is read-only oversight + destroy.

## What's still outstanding

### P1 — All three regions resolve to the same cluster

[`Tuist.Kura.Regions`](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/server/lib/tuist/kura/regions.ex#L80-L119) exposes three managed regions, but their `provisioner_config` carries no `kubernetes_client` overrides — only `cluster_id`, `public_host_template`, and `storage_class`. Production has `TUIST_KURA_AVAILABLE_REGIONS=eu-central,us-east,us-west` enabled.

The KubernetesController falls back to the in-cluster `Tuist.Kubernetes.Client` when no `kubernetes_client_opts` are present, which means **all three regions create their `KuraInstance` CR in the same fsn1 cluster the Tuist server runs in.** [`infra/k8s/clusters/`](https://github.com/tuist/tuist/tree/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/infra/k8s/clusters) only has `cluster-{staging,canary,production,preview}.yaml` — no `cluster-us-east.yaml` or equivalent.

So a customer clicking "US East" in account settings would:

- Get a CR named `kura-{handle}-us-east-1`
- Get a public host `{handle}-us-east-1.kura.tuist.dev`
- …backed by pods running in Falkenstein, with all the latency that implies for US/APAC users.

Two paths forward, both probably necessary before the tier is externally available outside EU:

1. Stand up a workload cluster per region (a `cluster-us-east.yaml`, `cluster-us-west.yaml` referencing the existing `tuist-hcloud` ClusterClass with `region: ash`/`hil`), bootstrap each via `bootstrap-workload.sh`, and add per-region kubeconfigs to the secrets bundle. Wire each region's `kubernetes_client` opts to point at the right cluster.
2. In the meantime, restrict `TUIST_KURA_AVAILABLE_REGIONS` to `eu-central` and surface the limitation in the customer-facing UI ("US East / US West coming soon"), so customers don't self-select a region that silently lands in EU.

The code shape is multi-region-ready; what's missing is the operational expansion. Worth resolving the marketing-vs-reality gap before the FunWithFlags `:kura` flag is enabled for accounts outside EU.

### P2 — `wait_until_ready` holds an Oban worker for up to 10 min

[`KubernetesController.wait_until_ready/4`](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/server/lib/tuist/kura/provisioner/kubernetes_controller.ex#L279) `Process.sleep`-loops until the CR's `.status.phase == "Ready"` or the 10-minute deadline. The `RolloutWorker` is on `:default` (concurrency 10), and `:default` is genuinely shared across most workers in the codebase — most use `queue: :default` explicitly, the rest omit the option and Oban defaults them to `:default`. Only `:process_build` and `:process_xcresult` are carved out for the dedicated processor fleets.

So if 10 customers create or upgrade Kura servers in close succession (or `VersionMonitorWorker` fires when a new Kura release lands and N active servers are behind), all those rollouts hold `:default` slots while sleep-polling. While they hold the slots, no other `:default` job runs — `AlertWorker` (`*/10 * * * *`), `AutomationScheduler` (`* * * * *`), `Slack.Workers.ReportWorker`, the slack/billing/usage crons, etc. all back up.

Two ways to fix:

1. **Carve out a `:kura_rollout` queue** with bounded concurrency (e.g. 5). One-line addition to `base_queues`. Bounds the starvation to that queue and leaves `:default` for everything else.
2. **Stop blocking the worker.** `RolloutWorker` creates the CR and returns immediately; a periodic cron walks `kura_deployments` in `:running` state, polls each CR's `.status.phase`, and transitions the row. The controller already has a 30s requeue loop ([`kurainstance_controller.go` L107](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/infra/kura-controller/controllers/kurainstance_controller.go#L107)) writing observed status into the CR — the server can just observe that asynchronously instead of holding a slot waiting for it. Closer to the eventual-consistency reconciliation pattern #10653 uses, and removes the worker-pool pressure entirely.

(1) is a one-line fix that mostly addresses the immediate concern. (2) is the cleaner shape long-term.

### P2 — bandwidth caps deserve a re-examination

[`kurainstance_controller.go` L339-L344](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/infra/kura-controller/controllers/kurainstance_controller.go#L339-L344) hardcodes `kubernetes.io/{ingress,egress}-bandwidth: 250M` for every instance. The original justification was twofold: per-tier billing differentiation (small/medium/large) and noisy-neighbor mitigation on a shared substrate. The first reason is gone — `Tuist.Kura.Specs` was deliberately dropped in `3a3f72f92b`. With one cap for everyone, that purpose evaporates.

The second reason (noisy-neighbor protection) is real but worth weighing against the cost: 250M egress ≈ 31 MB/s. A hot Xcode CI cache fetch is often several hundred MB; at 31 MB/s that's 10+ seconds added per build on the cache hot path. For a tier whose pitch is "closer-to-CI artifact reads", capping at a third of a Gbps is customer-visible and cuts directly against the value prop. Cilium caps are also hard ceilings — they don't yield to host headroom even when the NIC is idle.

Two options worth considering:

- **Drop the cap for GA** and add a Grafana alert on per-pod `container_network_transmit_bytes_total` p95 crossing some threshold. Catch noisy neighbors when they actually happen and react targeted (per-customer cap, rate limit at the Kura request layer, or a kindly worded email). The customer-perceived cost of a cap is paid every CI run; the cost of an actual noisy-neighbor incident is paid once.
- **Or raise the cap** to a sanity backstop level (e.g. 750M / 750M) — high enough that normal cache traffic never hits it but a single runaway pod can't take out the whole NIC.

The 250M number reads like it survived from the medium-tier value when the spec catalog still existed and didn't get re-examined when tiers were dropped.

### P3 — `Reconciler` moduledoc still references `rollout.sh`

[`reconciler.ex` L24](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/server/lib/tuist/kura/reconciler.ex#L24): "rollout.sh is idempotent, but the cluster-side state may be mid-helm-upgrade and an operator should look before re-triggering." The system no longer uses helm or `rollout.sh` in the production path, so the rationale needs updating to reference the controller-driven reconciliation.

### P3 — `HelmKubernetes` provisioner kept as dev-only legacy

[`HelmKubernetes`](https://github.com/tuist/tuist/blob/8ed1c0ed99d55d3ca41e2e601d55a7137a87f263/server/lib/tuist/kura/provisioner/helm_kubernetes.ex) is still in the codebase, only reachable via the `local` region (not `local-controller`, which uses `KubernetesController`). With production fully migrated, the helm provisioner is dead-code-on-prod and dev-redundant — the `local-controller` region exercises the production path via kind, which is more useful coverage. Worth deleting in a follow-up to keep one provisioner / one path.

### P3 — handle-format error message could be clearer

`provisioner_node_ref` is validated against RFC1123, but the ref is computed in code as `kura-{handle.downcase}-{cluster_id}`. An account handle with `_` would fail the regex with the message "must be a valid Kubernetes RFC1123 label", which is correct but doesn't clearly point a non-Kubernetes operator at the actual offending input (the account handle). Consider hoisting the validation to `account.name` itself with a customer-friendly message.

## Net

This went from "should this even ship in this shape" to "ship after queue carve-out + bandwidth-cap re-examination + multi-region cluster expansion". The architectural concerns are resolved; what's left is a mix of ops-side rollout sequencing and small polish.

fortmarek:
One architectural inconsistency worth pulling on before this lands: every other componentized image in the codebase follows the lockstep model — built with the deploy SHA in `server-deployment.yml`'s build job, deployed via `helm --set component.image.tag=$IMAGE_TAG`, version-locked to the server. The runner image, the xcresult-processor image, and the kura-controller image you just added all work this way. Only the **Kura runtime image** (`ghcr.io/tuist/kura:X.Y.Z`, the per-customer pod) is on a separate, asynchronous GitHub-Releases-polling track.

Today's flow:

```
PR merges → CI tags ghcr.io/tuist/kura@X.Y.Z (separate motion)
        → ≤1h later, VersionMonitorWorker polls GitHub Releases via Req
        → diffs against active kura_servers, schedules deployments
        → RolloutWorker fires, controller rolls out
```

Lockstep equivalent:

```
PR merges → CI builds ghcr.io/tuist/kura:sha-XXX alongside server + controller
        → helm passes --set kura.runtime.image.tag=$IMAGE_TAG
        → reconciler ensures KuraInstance.spec.image == env tag, schedules rollouts on drift
        → controller reconciles
```

What lockstep buys, beyond consistency with the rest of the codebase:

- **One deploy pipeline, one rollback boundary.** Today `kura@0.5.3` lands in production within an hour of being tagged, with no canary soak. Lockstep means a bad Kura runtime can't land in prod without first going through canary like every other change.
- **Coordinated server-runtime changes.** A new Lua hook field that requires both server-side rendering changes and Kura runtime support ships in one SHA. Async means shipping Kura first, waiting for fleet upgrade, then shipping the server — extra coordination per change.
- **Removes ~70 lines of code + a dependency.** `Kura.latest_versions/1` (with `KeyValueStore` cache + retry options + GitHub auth headers + failure-doesn't-cache logic), `Kura.extract_kura_release/1`, `VersionMonitorWorker`, `Kura.schedule_latest_version_deployments/0`, `Kura.servers_needing_version_query/1`, plus the `Tuist.Environment.github_token_update_package_releases` dependency. All replaced by reading `Tuist.Environment.kura_runtime_image_tag()` from an env var helm sets at deploy time.
- **Drops the second release cadence.** `kura@X.Y.Z` tagging stops being a separate motion; merge-to-main is the one trigger.

What async buys (and whether those cases are real today):

- **Kura's semver visibility as a customer-facing fact.** Customers see "running 0.5.2" instead of "sha-8ed1c0ed99d5". Solvable independently — display `KuraInstance.status.observedImage` regardless of how the tag is sourced.
- **Hot-fix Kura without a server deploy.** Tag `kura@0.5.4` for a CVE, fleet upgrades within the hour. Real long-term, but not a workflow today — and a server deploy isn't actually expensive given the existing canary → production cascade.
- **Future customer pinning.** Enterprise customer says "we certified 0.5.2, hold us there". Real future product surface, but not present today and the lockstep model doesn't preclude adding it later (a per-customer override on `KuraInstance.spec.image`).

For the team's current scale and maturity, the costs of async (extra surface, inconsistent with everything else, decoupled rollback boundary, second release motion) outweigh the benefits. Kura isn't a customer-pinnable product surface yet, hot-fix-without-server-deploy isn't a real workflow today, and Kura the cache mesh is in this monorepo / same release pipeline as everything else.

Concrete migration:

1. Add a Kura runtime image build to the existing `build` job in `server-deployment.yml`, same SHA as the server and controller.
2. Helm passes `--set kura.runtime.image.tag=$IMAGE_TAG`.
3. The chart sets `TUIST_KURA_RUNTIME_IMAGE_TAG` on the server pod env.
4. `Tuist.Environment.kura_runtime_image_tag/0` reads it; replaces `latest_versions/1` as the source of truth.
5. Fold the version-drift loop into `Reconciler` (it already walks `kura_servers`), or keep `VersionMonitorWorker` but have it read the env var instead of polling GitHub. Either way drops the GitHub dependency.
6. Drop the `kura@X.Y.Z` tag motion from the Kura release pipeline.

Roughly the same code shape, materially less to operate, and consistent with the runners + xcresult processor + kura-controller patterns the team already understands.

fortmarek:
Tying together a few separate threads on this review (the `wait_until_ready` queue starvation, the Reconciler-as-crash-recovery, and the lockstep-image-tag suggestion) — they're all symptoms of the same architectural choice, and converge if Kura adopts the eventually-consistent reconciliation pattern that [#10653](https://github.com/tuist/tuist/pull/10653) just established for runners.

What the runners reconciler does, every 60s ([`runners/reconciler.ex`](https://github.com/tuist/tuist/blob/60a9ce2b54acb3cef0b56a36225d48ca28583496/server/lib/tuist/runners/reconciler.ex)):

1. **Observe K8s state** via `Tuist.Kubernetes.Client.list_pods(ns, label_selector)`.
2. **Count `alive` vs `target`** from `PoolConfig`.
3. **Compute `gap = max(0, target - alive)`.**
4. **Create `gap` Pods** to close it. `runner_assignments` rows are auxiliary metadata, not authoritative state.

Holds the Oban slot for milliseconds. No `Process.sleep` polling, no `wait_until_ready`, no per-attempt worker. There's no "stuck running" recovery job because there's no synchronous wait-for-readiness step that can leave state hanging — BEAM crashes mid-tick are uninteresting, the next 60s tick reconverges. The `Reconciler` is *the loop*, not a recovery patch.

Kura today does something materially different:

- `RolloutWorker` per server, blocks ≤10 min in `wait_until_ready` polling for `.status.phase = "Ready"`.
- `kura_deployments` rows are mutated synchronously by the worker.
- BEAM crash → row stuck `:running` → separate `Reconciler` job exists *only* to clean up.
- `VersionMonitorWorker` polls GitHub Releases hourly to detect drift.
- Three workers (`RolloutWorker`, `DestroyServerWorker`, `VersionMonitorWorker`) plus the `Reconciler` recovery job.

Same shape as runners would collapse all of that into one cron-driven reconciler:

```elixir
def reconcile do
  desired_image = Tuist.Environment.kura_runtime_image_tag()
  servers = Repo.all(from s in Server, where: s.status not in [:destroyed])

  Enum.each(servers, fn server ->
    case server.status do
      :destroying ->
        Client.delete_kura_instance(...)  # idempotent
        if no_cr -> mark_destroyed(server)

      _ ->
        case Client.get_kura_instance(...) do
          {:ok, cr} ->
            apply_if_drift(cr, server, desired_image)  # image drift, spec drift
            sync_status(server, cr["status"])           # observed → Postgres

          {:error, :not_found} ->
            Client.apply(manifest(server, desired_image))  # BEAM-crashed-before-create case

          {:error, reason} ->
            Logger.warning("kura: reconcile read failed", server: server.id, reason: inspect(reason))
        end
    end
  end)
end
```

What this collapses, by concern:

| Concern from this review | Resolution |
|---|---|
| `wait_until_ready` blocks `:default` queue ≤10min per rollout | Reconciler holds slot for ms, observes status async via the next tick |
| `Reconciler` exists only as crash-recovery patch | Reconciler is the primary loop; crashes are uninteresting because next tick reconverges |
| `VersionMonitorWorker` + GitHub Releases polling | Folded in: drift check against `Tuist.Environment.kura_runtime_image_tag()` (env var) |
| Lockstep image versioning | Falls out naturally — env-supplied tag is the source of truth, no separate cadence |
| `RolloutWorker` / `DestroyServerWorker` / `VersionMonitorWorker` machinery | Collapse to one Oban cron worker (`ReconcileKuraWorker`) calling `Reconciler.reconcile/0` |

The `kura_deployments` table either becomes audit-only (rows derive status from observation) or gets dropped entirely. /ops shows current `KuraInstance.status` + the Grafana link that's already there; "deployment history" can be reconstructed from a small audit log if needed (timestamps on `kura_servers.current_image_tag` transitions).

The argument for doing this in *this* PR rather than a followup:

- It's not additional work on top of what's there — it's a structural simplification that *deletes* code (the per-server worker, the `wait_until_ready` polling, the GitHub-Releases poller, the Reconciler-as-recovery, the per-attempt deployment row mutation logic). Net diff is negative.
- Four unrelated review concerns collapse into one resolution. Carrying them as separate followups means four times the review surface.
- The pattern is the canonical shape in the codebase now. Shipping Kura with a different shape means new contributors need to learn two patterns instead of one, and any future "K8s-orchestrated component" PR has to choose which precedent to follow.
- Doing it later means migrating live production state from the synchronous-worker shape to the reconciler shape, which is more work than doing it before any customer has a `kura_servers` row.

One nuance worth being explicit about: Kura's reconciler does more per-tick work than runners' because it carries version-drift detection and status mirroring per server. But the *outer shape* (cron, observe, close gap, return in ms) is identical, and the runners pattern absorbs that complexity cleanly — Kura's `reconcile/0` body is bigger than runners', the worker scaffolding is the same.

fortmarek:
Tying off another thread on this review (the question of what `HelmKubernetes` is for now that the controller path is the production shape): the dev story can collapse to one transport, deleting ~700 lines and a kubectl/helm dependency. Suggesting this as a small followup since it's not blocking but the cleanup is unambiguous.

`Tuist.Kubernetes.Client` has three modes today:

| Mode | Transport | Used by |
|---|---|---|
| `:in_cluster` | Projected ServiceAccount at `/var/run/secrets/.../token` | Server pod talking to its own cluster (eu-central) |
| `:kubeconfig` | Parses a kubeconfig YAML, native Req over Finch | Server pod talking to a remote cluster (us-east, us-west) |
| `:kubectl` | Shells out to `kubectl`, parses JSON output | Local dev (`local-controller` region) |

`:kubectl` exists for one reason: when the dev loop was built, host-process → kind-cluster didn't have a non-shellout option. `:kubeconfig` (added in v4 for cross-region prod) works from any process — host, pod, pod-talking-to-a-different-cluster. No cluster-membership requirement. So `local-controller` could just use:

```elixir
kubernetes_client: [mode: :kubeconfig, kubeconfig_path: "~/.kube/config"]
```

…against a kind cluster created by `mise run dev:kura-cluster` (one-time setup), same code path production uses for us-east/us-west.

What that simplification deletes:

| Code | Lines | Why it exists |
|---|---|---|
| `HelmKubernetes` provisioner | ~485 | Dev-only legacy of the helm-shellout era |
| `helm_kubernetes_test.exs` | ~250 | Tests for the above |
| `:kubectl` mode in `Tuist.Kubernetes.Client` | ~150 | Pre-`:kubeconfig` dev shim |
| `kind` cluster auto-create + `kind_kubeconfig` in the client | ~50 | Lazy-on-request convenience |

What it keeps:

- `local-controller` region as the only dev region (now using `:kubeconfig`, not `:kubectl`)
- `kind create cluster` lifted to a mise task instead of being lazy-invoked from a request handler
- `Provisioner.HelmKubernetes.image_tag_from_image/1` moved to a shared `Tuist.Kura.Image` util (~10 lines, the one helper `KubernetesController` actually uses)
- The dev experience: `mise run dev` brings up kind + Phoenix, "Create Kura server" in /ops works exactly as today

Tradeoff: auto-create-on-demand convenience goes away. Today, click "Create Kura server" without a kind cluster and the server boots one for you (~90s). Under the simplification, devs run `mise run dev:kura-cluster` first (or the main `mise run dev` task does it eagerly). Mildly worse for first-time setup, eliminates a class of "request handler just started a 90s shellout" surprise.

Three followups in increasing scope:

1. **Drop `HelmKubernetes` + `local_region/0`.** ~700 lines deleted, no dev capability lost (already shadowed by `local-controller`). Move `image_tag_from_image/1` to `Tuist.Kura.Image`. **Unambiguous cleanup.**

2. **Migrate `local-controller` from `:kubectl` to `:kubeconfig`.** Add `mise run dev:kura-cluster` (or fold into `mise run dev`). Update `local_controller_region/0`'s `kubernetes_client` opts to read the kind kubeconfig path.

3. **Drop `:kubectl` mode entirely** from `Tuist.Kubernetes.Client`. ~150 lines deleted. One transport for non-in-cluster cases (`:kubeconfig`) plus the in-cluster shape — two modes total, not three.

Net: one transport for "talk to a remote cluster" regardless of whether it's the dev-laptop's kind or a Hetzner workload cluster, one provisioner (`KubernetesController`), one dev region (`local-controller`). The codebase ends up with the smallest set of patterns that covers every case.

Happy to do this in a followup PR if it sounds reasonable.

fortmarek (approved):
Left three extra comments that might be worth addressing before merging this. Otherwise, this is looking great, can't wait to see this rolled out 💯

PR #10490, opened 2026-04-27
ci: add blick agentic PR review for elixir scopes

pepicrft:
## Summary

- Wires up [tuist/blick](https://github.com/tuist/blick) to run agentic reviews on PRs touching the Elixir codebases (`server`, `cache`, `processor`, `xcode_processor`, `tuist_common`, `noora`).
- Uses `opencode` + Kimi via Fireworks (`fireworks-ai/accounts/fireworks/models/kimi-k2p5`), matching the model used in `tuist/blick` itself.
- Ships a project-specific skill at `.blick/skills/tuist-elixir-review/SKILL.md` covering the things only this repo knows — authorization invariants, tenancy, `IngestRepo` write-only, Mimic placement, migration `:timestamptz`, `data-export.md` updates, and currency-symbol i18n. Generic Elixir style is explicitly out of scope (mix format + credo cover that).
- Workflow filters by path (Elixir scopes only), skips draft PRs, and only publishes on review success.

## Required secret

`FIREWORKS_API_KEY` must be set as a repository secret. Already configured ✓

## Validation

End-to-end run on the previous revision succeeded — blick + opencode + Kimi all installed and ran across all six scopes via the live Fireworks API. Commit `5b9db8c` is the first revision with the project-specific skill content; the next workflow run will exercise it.

## Test plan

- [x] Add `FIREWORKS_API_KEY` to repo secrets
- [x] Confirm `mise use -g github:tuist/blick` and the opencode install both succeed on the runner
- [x] Confirm the workflow runs and exits cleanly with no findings on a non-Elixir-touching diff
- [ ] Land this PR, then verify the next Elixir PR gets a real review
- [ ] Open a small PR touching only `cli/` and confirm the workflow is skipped (paths filter)


blick.toml
```diff
@@ -0,0 +1,12 @@
+[agent]
+kind = "opencode"
+model = "fireworks-ai/accounts/fireworks/models/kimi-k2p5"
```
pepicrft, line 3: Let's use it for a few days @fortmarek and see what the token consumption looks like.

noora/blick.toml
```diff
@@ -0,0 +1,4 @@
+[[reviews]]
+name = "elixir"
+skills = ["elixir-code-review"]
+prompt_file = "../.blick/prompts/elixir.md"
```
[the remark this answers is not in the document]
pepicrft: @fortmarek this is an example of a code review comment.

.blick/skills/elixir-code-review/SKILL.md
```diff
@@ -0,0 +1,81 @@
+---
+name: elixir-code-review
```
pepicrft: This is vendored from [here](https://github.com/openclaw/skills/tree/main/skills/anderskev/elixir-code-review).

.blick/skills/elixir-code-review/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+- [ ] No blocking calls in GenServer callbacks
+- [ ] Proper use of call vs cast (sync vs async)
+
+### Documentation
+- [ ] Public-API modules (controllers, contexts, public boundaries) have @moduledoc describing purpose
+- [ ] Public-API functions have @doc
+- [ ] Doctests for pure functions where appropriate
+- [ ] No @doc false on genuinely public functions
+
+This repository intentionally does **not** use typespecs. Do **not** flag
```
fortmarek: The override says "do not flag missing `@spec`/`@type`/`@typep`", but `references/documentation.md` still teaches `@spec` patterns extensively and ends with the review question "Do all public functions have @doc and @spec?". The model loading that reference will likely surface typespec-related findings anyway. Either strip the `@spec` sections from the reference doc or add a stronger banner there.

.blick/skills/elixir-code-review/SKILL.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Code Style
+- [ ] Module names are CamelCase, function names are snake_case
+- [ ] Pipe chains start with raw data, not function calls
+- [ ] Private functions grouped after public functions
+- [ ] No unnecessary parentheses in function calls without arguments
+
+### Pattern Matching
+- [ ] Functions use pattern matching over conditionals where appropriate
+- [ ] With clauses have else handling for error cases
```
fortmarek: "With clauses have else handling for error cases" will over-flag. A `with` without `else` returning the un-tagged error tuple is often correct and idiomatic Elixir. Consider tempering this rule in the override.

.blick/skills/elixir-code-review/references/otp-basics.md
```diff
@@ -0,0 +1,131 @@
...
+def fetch_data(pid) do
+  GenServer.call(pid, :fetch_data, 10_000)  # 10 second timeout
+end
+
+# Handle timeout in caller
+case GenServer.call(pid, :fetch, 5_000) do
+  {:ok, data} -> data
+  {:error, reason} -> handle_error(reason)
+rescue
+  exit -> {:error, :timeout}
```
fortmarek: The `rescue exit ->` clause here doesn't actually catch `:exit` signals — that's `catch :exit, _`. If the model ingests this as a recommended pattern, it'll propagate the bug. Worth fixing upstream or in our copy.

.blick/skills/elixir-code-review/_meta.json
```diff
@@ -0,0 +1,17 @@
+{
```
fortmarek: Bigger question: do we want to vendor someone else's opinionated Elixir skill at all? It's mostly generic style/OTP that `mix format` + `credo` already cover, and we're already overriding it for typespecs. I think we'd get more value from a short project-specific `.blick/prompts/elixir.md` covering things only this repo knows (authorization invariants, tenant scoping, `IngestRepo` write-only, Mimic mock placement, `data-export.md` updates, etc.) — see the top-level review comment.

.github/workflows/blick.yml
```diff
@@ -0,0 +1,56 @@
+name: blick
+
+on:
+  pull_request:
```
fortmarek, line 4: No `paths:` filter — every PR (Swift CLI, docs, handbook) installs mise + opencode + blick and runs the agent before no-opping. Suggest:

```yaml
on:
  pull_request:
    paths:
      - "server/**"
      - "cache/**"
      - "processor/**"
      - "xcode_processor/**"
      - "tuist_common/**"
      - "noora/**"
      - ".blick/**"
      - "blick.toml"
      - ".github/workflows/blick.yml"
```

.github/workflows/blick.yml
```diff
@@ -0,0 +1,56 @@
...
+  cancel-in-progress: true
+
+env:
+  MISE_VERSION: "2026.4.18"
+  MISE_GITHUB_ATTESTATIONS: 0
+
+jobs:
+  review:
+    name: Review
+    if: github.event.pull_request.head.repo.full_name == github.repository
```
fortmarek: Add `github.event.pull_request.draft == false` so we don't pay LLM cost on every push to a draft. This PR being draft is the exception for end-to-end validation — flip it on after.

.github/workflows/blick.yml
```diff
@@ -0,0 +1,56 @@
...
+          mise use -g github:tuist/blick
+          mise use -g npm:opencode-ai
+
+      - name: Run blick review
+        env:
+          FIREWORKS_API_KEY: ${{ secrets.FIREWORKS_API_KEY }}
+        run: blick review --base "origin/${{ github.event.pull_request.base.ref }}" --json --stream
+
+      - name: Publish review
+        if: always()
```
fortmarek: If the review step fails (Fireworks 5xx, OOM, agent crash) `blick publish` still runs because of `if: always()`. Worth confirming it gracefully no-ops rather than posting an empty/confusing review.

cache/blick.toml
```diff
@@ -0,0 +1,3 @@
+[[reviews]]
```
fortmarek, line 1: All six per-scope `blick.toml` files are byte-identical. Not blocking, but if blick supports it, a single root entry like `[[reviews]] paths = ["server", "cache", ...]` would be tidier. If not, fine.

fortmarek (commented):
Overall a small, well-scoped CI addition — but I'd hold off merging until two things are addressed, and I'd encourage rethinking the vendored skill before this becomes the long-term shape.

### Description ↔ diff mismatch

The description says *"A shared prompt at `.blick/prompts/elixir.md` overrides the skill's `@spec` guidance"*, but that file isn't in the diff — the override is inlined into `SKILL.md` instead. Either the description needs updating, or the prompt file is missing. Worth confirming whether blick actually picks up `SKILL.md` overrides the way we expect.

### Bigger direction: drop the vendored skill, write a project-specific prompt

The current setup pays maintenance cost (vendored skill drift, conflicting overrides) for content the model already knows. The agent's high-leverage work is the things only this repo knows — and we have plenty of those:

- **Authorization** (`lib/tuist/authorization.ex` + `AuthorizationPlug`): new `object :foo` requires a matching `@project_categories` entry; `:authenticated_as_project` without `:projects_match` is a critical bug; `:public_project` / `:ops_access` should be read-only; new actions should cover user/project/account-token subjects or document the exclusion.
- **Tenancy**: bare `Repo.get(Bundle | Run | Cache, id)` without project/account scope is a tenant leak.
- **Write-only repo**: `IngestRepo.all/get/one` is wrong (per `CLAUDE.md`).
- **Tests**: `Mimic.copy(...)` belongs in `test_helper.exs`, not in test files.
- **Migrations**: timestamps must be `:timestamptz`; schema changes require a `data-export.md` update.
- **i18n**: currency symbols must not be wrapped in `dgettext`.

Roughly 60–80 lines of project-specific rules will catch real bugs and outperform the 400-line generic skill. Most of what's in the vendored skill (naming, pipe chains, function ordering, nesting depth, `String.to_atom` on input, module docs) is already either in `mix format` or available as Credo checks (`PipeChainStart`, `StrictModuleLayout`, `Nesting`, `UnsafeToAtom`, `ModuleDoc`) — cheaper, faster, deterministic, and runs locally.

Happy to draft the prompt as a follow-up.

### Recommendation

Approve once the description ↔ diff mismatch is resolved; treat the vendored-skill question and the workflow filter improvements as follow-ups so this can land and start producing reviews.

.blick/skills/elixir-code-review/SKILL.md
pepicrft: Good catch — banner added to the top of `references/documentation.md` and the `@spec`-heavy `@spec Patterns` section + the "@doc and @spec" review question stripped. The example header is now `### @doc` (no @spec). Should keep the model from drifting toward typespec findings.

.blick/skills/elixir-code-review/SKILL.md
pepicrft: Tempered — rule now reads: `with` has an `else` clause **only when** the error needs transforming or distinguishing; bare `with` letting un-tagged tuples flow through is explicitly not a finding.

.blick/skills/elixir-code-review/references/otp-basics.md
pepicrft: Fixed in our copy — example now uses `try / catch :exit, {:timeout, _}` since GenServer.call timeouts are EXITs, not exceptions.

.blick/skills/elixir-code-review/_meta.json
pepicrft: Agreed in principle — capturing as a follow-up rather than expanding scope here. Added it to the PR description's Follow-ups section so it doesn't get lost. Happy to hand off to you for the project-specific prompt draft when you have a moment.

.github/workflows/blick.yml
pepicrft: Added the `paths:` filter exactly as suggested, plus `types: [opened, synchronize, reopened, ready_for_review]` so flipping out of draft also triggers it.

.github/workflows/blick.yml
pepicrft: Added `github.event.pull_request.draft == false` to the job `if`. I'll mark this PR ready-for-review once `FIREWORKS_API_KEY` is in to validate end-to-end.

.github/workflows/blick.yml
pepicrft: Switched `if: always()` to `if: success() && hashFiles('.blick/runs/latest/**') != ''`. Now only publishes when the review step succeeded and produced a run artifact, so a Fireworks 5xx / agent crash won't post an empty review.

cache/blick.toml
pepicrft: Checked the blick README — `[[reviews]]` doesn't take a `paths` glob; scope is determined by which `blick.toml` is closest to a changed file. Consolidating would require a feature blick doesn't expose, so keeping the six tomls as-is. If blick grows that, happy to migrate.

fortmarek (approved):

PR #10495, opened 2026-04-27
feat(server): alert on ingress-nginx 4xx/5xx for tuist-server

fortmarek:
## Summary

- Wires Prometheus alert rules for the `tuist-tuist-server` Ingress, evaluated by Grafana Cloud Mimir, so the next proxy-layer regression in the shape of [#10492](https://github.com/tuist/tuist/pull/10492) (1 MB body-size limit silently rejecting bundle metadata for `tuist inspect bundle`) is caught before a customer reports it in Slack. Sentry is blind to these because nginx returns the error before the request ever reaches Phoenix.
- Ships two alerts on `nginx_ingress_controller_requests`, scoped to the tuist-server Ingress and aggregated `by (env, cluster)` so each environment fires independently:
  - `TuistServerIngress5xx` — `severity: critical`, `> 0.05 req/s` for 5m
  - `TuistServerIngressBlocking4xx` — `severity: warning` on 413/429/499 only (excludes noisy 401/404), `> 0.1 req/s` for 10m
- Adds an `alerts/` directory + post-install/post-upgrade `mimirtool rules sync` Job to the k8s-monitoring chart. Auth reuses the existing ESO Grafana Cloud Secret (`prometheus-username` / `prometheus-password`) — no extra credentials. Sync runs from every environment install and is idempotent.

## Why a Helm-managed sync Job

There is no in-cluster Prometheus or Alertmanager — metrics flow to Grafana Cloud Mimir, which evaluates rules server-side. So `PrometheusRule` CRs (the operator-native pattern) wouldn't be picked up by anything: there's no Prometheus Operator running, and the [grafana/k8s-monitoring chart's `prometheusOperatorObjects`](https://github.com/grafana/k8s-monitoring-helm/tree/main/charts/k8s-monitoring/charts/feature-prometheus-operator-objects) feature only handles `ServiceMonitor` / `PodMonitor` / `Probe`, not `PrometheusRule`. The sync-job pattern keeps rule definitions declarative and version-controlled in this repo, scoped to the same Helm lifecycle as the rest of the observability stack.

## Test plan

- [x] `helm lint` clean across all three overlays (staging / canary / production)
- [x] `helm template` renders both the ConfigMap and the post-install Job for each overlay; rendered command, env vars, and volume mounts inspected
- [x] `promtool check rules` passes on the rule file (after stripping the Mimir `namespace:` key)
- [x] Embedded YAML in the ConfigMap re-parses cleanly as a Mimir rule namespace with two alerts
- [ ] Deploy to canary, confirm the `*-alert-rules-sync` Job lands `Completed`, and verify the rule group appears under namespace `tuist-server-ingress` in Grafana Cloud → Alerts & IRM → Mimir-managed rules
- [ ] Manually fire the 4xx alert by sending a >50 MB body through the canary ingress — expect a 413 from nginx, then the alert in Grafana Cloud after the `for: 10m` window
- [ ] After ~24h on canary, confirm no false-positives at the chosen thresholds (`> 0.05` for 5xx, `> 0.1` for 413/429/499); tune in a follow-up if baseline says otherwise

### DIFF
--- a/infra/helm/k8s-monitoring/README.md
+++ b/infra/helm/k8s-monitoring/README.md
@@ -11,6 +11,7 @@ Wraps [`grafana/k8s-monitoring`](https://github.com/grafana/k8s-monitoring-helm)
 | node-exporter | Deployed as DaemonSet (node CPU / mem / disk / net) |
 | kubelet + cAdvisor | Scraped (container resource usage) |
 | Kubernetes Events | Streamed to Loki as structured logs |
+| Alert rules | YAML under [`alerts/`](./alerts) synced to Grafana Cloud Mimir on every `helm upgrade` (see [Alerting](#alerting)) |
 
 With these in place the Grafana Cloud **Observability → Kubernetes** app populates automatically (Cluster / Namespace / Workload / Pod / Node views) without importing dashboards by hand.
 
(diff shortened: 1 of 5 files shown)


fortmarek:
Closing — going with a Grafana Cloud UI-defined alert instead. For one alert, the click-through is more pragmatic than maintaining a sync mechanism.

PR #10497, opened 2026-04-27
feat(server): add manual-mark trigger for test case automations

fortmarek:
> Adds event-driven automation triggers so customers can react the moment a test case changes — manually marked / unmarked as flaky, or transitioned to a different quarantine state — without waiting for the scheduled flakiness monitors to converge.

### Summary

- **New monitor type `test_updated`** on `automation_alerts`. `trigger_config.events` is a non-empty subscription array (Stripe-style): pick any combination of `marked_flaky`, `unmarked_flaky`, `state_changed_to_enabled`, `state_changed_to_muted`, `state_changed_to_skipped`. No thresholds, no window, no recovery — discrete one-shots.
- **Event-driven dispatch**: `Tuist.Automations.dispatch_test_case_event/2`, called from `Tuist.Tests.update_test_case/3` when the update carries an `actor_id` (so system writes can't recurse). The dispatcher maps the raw test-case event (`:muted`, `:unmuted`, …) to a subscription key, fans out to every enabled `test_updated` alert whose `events` array contains that key, and runs its `trigger_actions`.
- **Automation attribution on the test case timeline**: new `automation_id` column on `test_case_events` (ClickHouse migration). `ActionExecutor.apply_merged_attrs/3` stamps it; `list_test_case_events/2` preloads the `:automation`; the timeline shows `By automation <name>` instead of the generic `Automatically by Tuist`.
- **Live updates**: `Tests.update_test_case/3` broadcasts on `Tests.test_case_topic(id)` after every state change. `TestCaseLive` subscribes on mount and re-renders the state pill + history on the broadcast — so when an automation mutes a test the open detail page reflects it without a refresh. The broadcast fires *before* `dispatch_event_driven_automations/3` so the nested automation broadcast lands last and isn't clobbered by the outer pre-automation snapshot. The buffer is flushed in-line on state-change writes to avoid the 5s ingest lag.
- **LiveView dashboard**: one consolidated "Test updated" entry in the `When` dropdown with a Noora inline-checkbox list for the event subscription; hides threshold / window / recovery section entirely for the event-driven type; switching the metric forces recovery off and strips the now-redundant `add_label flaky` / `remove_label flaky` defaults; Add (recovery) action dropdowns moved to the section headers to match the existing notifications-modal pattern; action-row dropdowns and trash buttons sized to large to match.

### Follow-up worth considering

- The schema is still named `Tuist.Automations.Alerts.Alert`. The `Alert` name is a fossil from the pre-automations era; threshold-based monitor types (`flakiness_rate`, `flaky_run_count`) genuinely are alerts in the classic sense, but the new `test_updated` type isn't. We could either rename the umbrella to `Automation` (one schema, `monitor_type` is the discriminator) or split into two schemas (`Alert` for threshold types, `EventTrigger` for event-driven). Out of scope for this PR; tracking for a follow-up.

### How to test locally

1. `mise run dev`, sign in to the dashboard, open a project's `Settings → Automations`.
2. Click `Add automation`, name it, switch `When` to **Test updated**, tick **Marked as flaky**.
3. Confirm threshold / window / comparison / recovery toggle have all disappeared.
4. Add a `Change state → Muted` action; save.
5. Open any test case detail page and click **Mark as flaky** — the state pill flips to **Muted** live (no refresh), and the timeline gains both **Marked as flaky** (Manually by @you) and **Muted** (By automation \<name\>) rows.

### Test plan

- [x] \`mix format --check-formatted\`
- [x] \`mix credo\` on the touched files
- [x] \`mix test\` for the touched suites: \`alerts\`, \`alert_evaluation_worker\`, \`automations\`, \`action_executor\`, \`tests\` (event-creation describe), and \`project_automations_live\`
- [x] \`mix gettext.extract\` — only \`.pot\` files committed; \`.po\` files left to Weblate's \`tuistit\`
- [x] \`server/data-export.md\` updated with the new \`test_case_events.automation_id\` column


server/lib/tuist_web/live/project_automations_live.html.heex
```diff
@@ -252,6 +272,28 @@
                     />
                   </div>
                 </div>
+                <div :if={@create_automation_form_metric == "test_updated"} data-part="form-row">
```
pepicrft, line 281: When this selector appears for `test_updated`, the full post-event render still includes the `Window` input because the two window input blocks above only check `@create_automation_form_window_type`. I validated this locally with a temporary full-render assertion. The existing test only checks the `render_hook/3` diff, so it misses the leftover field. I'd gate those two window input blocks with `not event_driven_monitor_type?(@create_automation_form_metric)` too.

server/lib/tuist_web/live/project_automations_live.ex
```diff
@@ -176,16 +187,45 @@ defmodule TuistWeb.ProjectAutomationsLive do
   end
 
   def handle_event("update_create_automation_form_metric", %{"data" => metric}, socket) do
+    event_driven? = event_driven_monitor_type?(metric)
+
+    trigger_actions = strip_redundant_actions(socket.assigns.create_automation_form_trigger_actions, metric)
```
pepicrft, line 193: This strips the default `add_label` action when switching to `test_updated`, which leaves a new form with `trigger_actions == []`. The Save button remains enabled because it only checks name and rolling-window validity, then the changeset rejects the empty action list and the modal silently stays open. I validated the full render locally. Either keep a valid default action or disable/show validation until an action is added.

server/lib/tuist/automations/action_executor.ex
```diff
@@ -47,10 +47,10 @@ defmodule Tuist.Automations.ActionExecutor do
   defp test_case_attr_change(%{"type" => "change_state", "state" => state}), do: {:state, state}
   defp test_case_attr_change(_), do: :pass
 
-  defp apply_merged_attrs(_entity, attrs) when map_size(attrs) == 0, do: :ok
+  defp apply_merged_attrs(_entity, attrs, _automation) when map_size(attrs) == 0, do: :ok
 
-  defp apply_merged_attrs(%{type: :test_case, id: id} = entity, attrs) do
-    case Tests.update_test_case(id, attrs) do
+  defp apply_merged_attrs(%{type: :test_case, id: id} = entity, attrs, automation) do
+    case Tests.update_test_case(id, attrs, alert_id: automation.id) do
```
pepicrft: This now requires every automation argument with test-case mutations to have `:id`. Production callers pass an `Alert`, but the existing integration tests in `tests_test.exs` still pass `%{name: ..., project_id: ...}` and now raise `KeyError`. I validated with `mix test test/tuist/tests_test.exs:7236`. Either make attribution optional here or update those tests/callers to pass an id-bearing automation.

pepicrft (approved):
Left a few non-blocking notes from local validation. Approved.

fortmarek:
Rejecting this Blick finding — it's a false positive.

`test_updated` is a **value** for the existing `monitor_type` column on `Alert` rows, sibling to `flakiness_rate` and `flaky_run_count`. All three live in the same `automation_alerts` table and are governed by the same `object :automation_alert` policy block in `lib/tuist/authorization.ex` for create/read/update/delete.

The diff introduces:
- no new `object` block in `Tuist.Authorization`,
- no new controller or route,
- no new API endpoint.

`@project_categories` in `AuthorizationPlug` is a registry of **resource categories** that mirror `object :foo` blocks consumed by router pipeline plugs. Adding `:test_updated` there would create a phantom category with no matching `object`, no policy rules, and nothing calling `plug AuthorizationPlug, :test_updated` — pure dead code that would muddy a future real audit. Leaving as-is.

PR #10498, opened 2026-04-27
chore(server): drop dead processor.url + processor.webhook_secret from priv/secrets

fortmarek:
Stacked context: paired with [#10428](https://github.com/tuist/tuist/pull/10428), which removes the HTTP-webhook path between the Tuist server and the standalone xcactivitylog processor in favor of an Oban queue. After that PR, \`Tuist.Environment.processor_url/0\` and \`Tuist.Environment.processor_webhook_secret/0\` are gone, so the matching keys in the encrypted secrets files are unreferenced.

## Summary

Removes \`processor.url\` and \`processor.webhook_secret\` from the encrypted \`priv/secrets/{stag,can,prod}.yml.enc\` files. Smaller secrets surface, cleaner rotation, no behavior change — the lookups are deleted in #10428 so the values were dead before this commit.

Self-hosted installs that still keep these keys in their own \`priv/secrets/<env>.yml.enc\` are unaffected: the lookups don't exist anymore, so any extra entries are simply ignored at boot.

## Why a separate PR

The encrypted-file diff is unreadable on GitHub (it's just changed ciphertext). Splitting it out from the much larger #10428 keeps that PR's diff clean and lets a reviewer eyeball the secrets edit by checking out the branch and running \`mix secrets.edit --env <env>\` to see exactly what was removed.

## Test plan

- [x] \`mix secrets.edit --env stag\` shows the removed keys are gone, all other keys intact
- [x] Same for \`can\` and \`prod\`
- [x] Staging deploy succeeds and the server boots cleanly with the new secrets file (already running on staging from #10428's deploys; this PR just lands the underlying file change)


fortmarek:
Folded into #10428 ([commit 01ee09149b](https://github.com/tuist/tuist/pull/10428/commits/01ee09149b)). Closing to consolidate.

PR #10499, opened 2026-04-28
feat(server): manage xcresult processor as a k8s Deployment via tart-cri (Mac minis as real k8s nodes)

fortmarek:
## Summary

Stacks on top of [#10428](https://github.com/tuist/tuist/pull/10428). Migrates the xcresult processor from nix-darwin-on-bare-metal to a fully cluster-managed shape. Mac minis join the cluster as real Kubernetes nodes via a small kubelet-shaped agent (`tart-kubelet`) installed during host bootstrap, and are provisioned declaratively via Cluster API (`cluster-api-provider-scaleway-applesilicon`). Workloads schedule via standard `Deployment`s with `nodeSelector: kubernetes.io/os=darwin`.

The PR went through several architecture iterations across the conversation thread:

1. nix-darwin Mac minis hitting the server via webhook
2. embedded Orchard + Virtual Kubelet provider (sibling control plane)
3. tart-cri + upstream macOS kubelet (Mac minis as real k8s nodes via CRI)
4. Virtual Kubelet driven by an Elixir fleet manager in the Tuist server
5. **`tart-kubelet`: a small kubelet-shaped Go agent on each Mac mini, registering as a real Node** (this final shape)

(3) was abandoned because upstream kubelet doesn't ship for darwin and cAdvisor is hard-coupled to Linux. (4) was abandoned because routing every Pod allocation through a central Elixir process re-introduces a single point of failure and concentrates trust around the cluster's Secrets API surface; pushing the agent onto each Mac mini matches what real kubelet does and keeps the trust boundary tight.

## Architecture

```
 K8s API server
       │
       │ tart-kubelet on Mac mini (darwin/arm64, launchd-supervised, SA token auth)
       ▼
 ┌──────────────────────────────────────────────┐
 │ tart-kubelet                                 │
 │   ├── Pod controller (controller-runtime)    │
 │   │   ├── envFrom / valueFrom resolver      │
 │   │   │   (reads Secrets / ConfigMaps from   │
 │   │   │    the API server with its own       │
 │   │   │    credentials, exactly like         │
 │   │   │    real kubelet)                     │
 │   │   ├── pull / clone / run via local Tart  │
 │   │   └── status update on the Pod object    │
 │   ├── Node controller (register + heartbeat) │
 │   └── GC: orphan VMs + stale OCI cache       │
 │       (periodic + on Pull no-space error)    │
 └──────────┬───────────────────────────────────┘
            │ exec(local tart CLI)
            ▼
       Tart VMs on the host
```

One Pod = one Tart VM. The first container in a Pod becomes the VM image; multi-container, init, and ephemeral Pods are filtered out by the reconciler so DaemonSets that tolerate every taint (csi-node etc.) don't spam events.

Provisioning happens off to the side via Cluster API:

```
 K8s API server                                     (control plane)
       │
       │ ScalewayAppleSiliconMachine CR
       ▼
 ┌──────────────────────────────────────────────────┐
 │ capi-scaleway-applesilicon manager               │
 │   1. Provisioning  → scaleway.CreateServer       │
 │   2. Bootstrapping → SSH, install Tart, drop     │
 │                       kubeconfig + tart-kubelet, │
 │                       load launchd plist         │
 │   3. Drift-detect  → re-upload tart-kubelet      │
 │                       binary + reload launchd    │
 │                       on operator-image bumps    │
 └──────────┬───────────────────────────────────────┘
            │ HTTPS (Scaleway SDK)         │ SSH (operator-managed key)
            ▼                              ▼
       Scaleway API                  Mac mini host
                                       (tart-kubelet)
                                              │
                                              │ self-registers as Node
                                              ▼
                                     K8s API server (Node)
```

The tart-kubelet binary is baked into the operator's container image and uploaded to each Mac mini over SSH at provision time and on every operator-image rebuild. No external release artifact, no separate URL pin: bump the operator image tag, the operator's drift-detection stage rolls the new agent across the fleet (each host re-uploads the binary, reloads its launchd job, re-registers; running Tart VMs survive the agent restart because of `Setsid` on `tart run`).

## Path to runners-as-a-service

This shape is deliberately the substrate for the runners product ([#10264](https://github.com/tuist/tuist/pull/10264)). Everything that lands here is reused; the runners product is a layer on top, not a fork.

Reused as-is by runners:

- **`infra/cluster-api-provider-scaleway-applesilicon/` and the `ScalewayAppleSiliconMachine` CRD.** Same declarative fleet for both internal workloads (xcresult-processor) and customer workloads (runners). `kubectl scale machinedeployment` to grow, `kubectl delete machine` to retire.
- **`infra/tart-kubelet/`.** The agent that runs Pods on Mac minis is workload-agnostic. Customer runner Pods schedule onto the same Nodes via `nodeSelector: kubernetes.io/os=darwin`.
- **The Tart image pipeline.** Runners get their own image (`ghcr.io/tuist/tuist-runner-macos:<sha>`) built by the same Packer flow, just with a different launchd job and base toolchain. The kubelet doesn't care.

What runners adds on top (none of which lives in this PR):

- **A `TuistRunner` CRD + controller** (in the Tuist server, alongside billing/accounts) that translates customer requests ("give me a runner with these specs") into Pods with the right env, image, and node affinity. Same shape as GitHub's Actions Runner Controller.
- **Warm-pool management.** The controller pre-creates N idle Pods per subscription tier so VM clone+boot (~30–60 s) is hidden from the customer's wait time.
- **Cache-locality scheduling.** The controller prefers the Mac mini that ran a project's last build via weighted nodeAffinity hints; tart-kubelet's image-cache + Tart's clone-from-cached-image keep that fast.
- **Per-account concurrency quota.** Soft cap on parallel runs per account on the pool tier so one customer can't starve everyone else. Enforced at the controller before Pod admission.
- **BYO image-pull credentials.** For customers running their own Tart images, the controller stages a per-Pod cred file and tart-kubelet picks it up before pulling. The curated-base flow (single registry, single creds) doesn't need this.

### Customer-to-machine assignment

Two tiers, both mapped to standard k8s scheduling primitives. The substrate doesn't care which tier a Pod belongs to; the controller picks labels + nodeSelector at admission and the scheduler does the rest.

**Pool (default).** Customer runner Pods carry `nodeSelector: kubernetes.io/os=darwin, tuist.dev/runtime=tart` and no node-affinity beyond cache-locality hints. kube-scheduler picks any free Mac mini in the shared `tuist-tuist-macos-fleet` MachineDeployment. The Tart VM is the customer-isolation boundary: separate kernel via Virtualization.framework, NAT'd network with no cross-VM reachability, CoW-cloned disk per VM. Billing per build-minute.

**Dedicated.** A per-account MachineDeployment, templated as `tuist-tuist-macos-fleet-<account_id>`. The operator labels its Nodes `tuist.dev/account=<id>` and taints them `tuist.dev/account=<id>:NoSchedule`; the customer's runner Pods carry the matching nodeSelector + toleration so no one else's Pods can land. Same chart, same operator (the `macos-fleet.yaml` template already iterates over a list of fleet entries; adding a dedicated customer is one more entry). Billing monthly, fixed. Useful for data-residency requirements (zone pin via `ScalewayAppleSiliconMachine.spec.zone`) or sustained workloads where per-build-minute pricing doesn't pencil out.

End-to-end flow either way:

1. Customer triggers a run (CLI command, CI hook, MCP tool).
2. Server's `Tuist.Runners` controller materializes a Pod with labels (`tuist.dev/account=<id>`, `tuist.dev/run-id=<uuid>`) and a tier-driven nodeSelector.
3. tart-kubelet reconciles the Pod into a Tart VM exactly as it does for the xcresult-processor today.
4. Pod deletion cleans up the VM and its CoW disk.

Three architectural reasons this split is right:

1. **Per-node failure isolation.** Each Mac mini's agent is independent. One agent crashing doesn't block other customers' jobs. With a centralized Elixir fleet manager (iteration 4 in the list above), every customer's job goes through one process; that's a SPoF *and* a contention bottleneck under load.
2. **Push-based status, not polling.** tart-kubelet updates Pod status as soon as the VM transitions; the runners controller subscribes via standard k8s watch. Customer dashboards see state changes within milliseconds of the Mac mini reporting them, not after the next 30 s polling tick.
3. **Standard k8s primitives all the way down.** Customer-facing APIs translate "give me a runner" into Pod manifests; everything below that uses standard kubelet/controller patterns. Operations are debugged with `kubectl describe pod`, `kubectl logs`, `kubectl get events` — the existing on-call playbook works for runners on day one.

### Co-located Linux fleet for Kura cache nodes

A core part of the runners value-proposition is that cache lookups during a customer's `tuist build` should feel free — sub-millisecond hits on a binary cache, not the 50-200 ms round-trip a cross-region cache lookup imposes today. The way to deliver that is to put `kura/` (the Rust distributed cache mesh) Pods physically next to the Mac mini fleet, in the same Scaleway zone (`fr-par-1`), on Scaleway's private network. Mac mini → Kura then becomes a single-DC RTT, ~0.5-2 ms.

We can't run Kura on the Apple Silicon fleet itself (Linux/Rust on macOS doesn't fit, and Mac mini cycles belong to customer builds). The right shape is a sibling fleet of Scaleway Linux instances, managed by a `cluster-api-provider-scaleway-instance` operator following the exact same pattern as `cluster-api-provider-scaleway-applesilicon`:

- A new `ScalewayInstanceMachine` CRD + reconciler. Talks to Scaleway's Instance API instead of the Apple Silicon API; the lifecycle stages (Provisioning → Bootstrapping → Drift-detection) are identical.
- Bootstrap is much simpler than the Mac mini path — no Tart, no tart-kubelet, no auto-login. Just `kubeadm join` against the cluster's API server. The host runs upstream Linux kubelet and Kura runs as ordinary k8s Pods.
- Same chart pattern: a `kuraFleet` block in `values-managed-<env>.yaml`, a `kura-fleet.yaml` template that emits a MachineDeployment, the same Scaleway IAM creds in 1Password (one Scaleway project covers both Apple Silicon + Instance APIs).

What this PR sets up that the cache fleet inherits:

- The CAPI provider operator skeleton (controller-runtime manager, per-fleet SSH key with Scaleway IAM registration, per-machine bootstrap Secret, TOFU host-key, drift-loop with retry cap + `FailureReason` transition, finalizer-clear on Scaleway 404). All of it is provider-agnostic; only the Scaleway API surface differs. Estimated reuse: ~70-80% of the Go code.
- The chart's `macos-fleet.yaml` template iterates over fleet entries and renders one MachineDeployment per entry. `kura-fleet.yaml` is a near-mirror with different labels and a different MachineTemplate kind.
- The same operational ergonomics: `kubectl scale machinedeployment` to grow, `kubectl delete machine` to retire, `kubectl get nodes -l tuist.dev/role=cache` to see them.

Network locality is the load-bearing piece. Scaleway's private network feature peers the Apple Silicon and Instance fleets without going through public IPs, and the `tuist.dev/runtime=tart` label on Mac minis vs `tuist.dev/role=cache` on Linux nodes lets the runners controller emit Pods that pin Kura nodes to the same zone as the Mac mini hosting their build via topologySpreadConstraints. The same `Tuist.Runners` controller that picks a Mac mini for a customer's build can pick a Kura cache Pod in the same DC with one extra hint.

## Path to BYOC (bring-your-own-cloud)

A directionally-related ask is "the customer brings their AWS credentials and we provision Mac/Linux runners in their account from the Tuist dashboard." Same shape as Render BYOC, Datadog Agent, GitHub self-hosted runners. This PR is intentionally compatible with that direction; some of the BYOC-blocking items are addressed here, the rest are flagged with rough effort estimates.

**Credential model is sts:AssumeRole, not access keys.** Customer pastes a CloudFormation stack URL, applies it once in their account, gets a role ARN with the trust policy back. Tuist holds nothing long-lived. CAPI's `AWSClusterRoleIdentity` shape maps cleanly: per-customer `Cluster` + `MachineDeployment` resources reference an identity pointing at the customer's ARN. CAPA does the EC2 call, our `tart-kubelet` runs on the EC2 Mac, registers into Tuist's cluster, runs Pods. Same agent code; only the provisioner swaps from CAPI-Scaleway to upstream CAPI-AWS.

**What's already done in this PR for the BYOC seam:**

- **`infra/macos-host-bootstrap/` is a standalone Go module.** Provider-neutral by design. `Config.UserPassword` (renamed from `SudoPassword`) doesn't care whether the password came from Scaleway's `CreateServer` response or from `dscl . -passwd` on an EC2 Mac. The package's doc enumerates which Config fields are provider-specific (IP, SSHUser, UserPassword, KnownHostFingerprint) vs. universal (auto-login, hostname, Tart, kubeconfig, launchd). A future `cluster-api-provider-aws-applesilicon` module imports it as-is.
- **Per-machine kubelet identity.** No more shared fleet token — `credentials.Manager.EnsureNodeIdentity` mints one ServiceAccount + service-account-token Secret + ClusterRoleBinding per Machine. A compromised host can only authenticate as its own Node identity, not impersonate every other host. This is the floor for multi-tenant; the next step (NodeAuthorizer-style Node-graph-restricted RBAC) is a follow-up but not blocking.
- **TOFU SSH host-key pin** rather than `ssh.InsecureIgnoreHostKey`. Persisted in the per-machine Secret on first reconcile, verified on every subsequent dial. The same code works against an EC2 Mac.

**What still has to land before BYOC ships:**

- **Worker-to-control-plane network.** Customer Nodes egress to Tuist's k8s API on 443 — straightforward. But `kubectl logs` and `kubectl exec` rely on the API server reaching back into each kubelet, which doesn't traverse a customer VPC. Either Konnectivity (upstream solution), or have tart-kubelet push logs/events to a Tuist-side collector. Latter is simpler and what we'd want for multi-tenant anyway.
- **Operator-to-Mac-mini network primitive.** Today the operator dials each Mac mini's public IP over SSH using a per-fleet ed25519 key. That works inside our Scaleway project but doesn't extend across a customer-VPC boundary — the EC2 Mac would either need a public IP (operationally fragile, security-team friction) or a reverse-tunnel solution. The natural fix is something like Tailscale (each Mac mini joins a Tuist-administered tailnet on first boot, the operator dials over WireGuard, port 22 stays closed to the internet, Tailscale SSH replaces our SSH-key auth) or Cloudflare Tunnel (`cloudflared` on each host, Cloudflare Access in front of SSH; same outbound-only shape, reuses an SaaS we already depend on, slightly weaker on the SSH-replacement angle). Out of scope for this PR — the per-machine SSH credentials + TOFU host-key the operator already does are tight enough for the single-tenant case — but worth picking up alongside the BYOC design.
- **CAPI provider for AWS EC2 Mac.** A sibling module to `cluster-api-provider-scaleway-applesilicon` that wraps upstream CAPA and handles EC2-Mac quirks (24h dedicated-host minimum allocation, region/AZ pinning).
- **Onboarding UX.** Connect-AWS flow, CloudFormation template, region picker, instance-type picker, quota surfacing.
- **NodeAuthorizer-style RBAC tightening.** Per-machine SAs prevent fleet-wide impersonation; the natural next step is restricting each SA to only the Node + Pods it owns. Standard k8s solution: enable NodeAuthorizer + NodeRestriction admission, authenticate kubelets as `system:node:<name>` via TLS bootstrap (CSRs + cluster-CA-signed certs). Bigger change than per-machine SAs but the right end state.

**Realistic effort outside this PR:**

| Phase | Scope | Effort |
| --- | --- | --- |
| MVP (Linux runners, single region, dedicated pool, logs via S3) | CAPA wiring + AssumeRole + onboarding + NodeAuthorizer migration | 6-8 weeks |
| Production BYOC (Mac, multi-region, elastic where possible, kubectl-logs equivalent) | + Konnectivity-or-equivalent + EC2 Mac quirks + tenant-isolation hardening | 2-3 months |
| Per-tenant cluster (data-residency / compliance) | tiny k8s control plane per customer in their VPC | additional month |

EC2 Mac BYOC is a fine v2 because the 24h commitment + per-day pricing genuinely change the product surface — pool elasticity is gone, only the dedicated tier makes sense. Linux/CI BYOC is the more natural MVP and reuses everything in this PR.

### `infra/tart-kubelet/` — new Go module (the agent on each Mac mini)

- [`cmd/tart-kubelet/`](infra/tart-kubelet/cmd/tart-kubelet/) — controller-runtime manager. Authenticates to the API server via `--kubeconfig=/etc/tart-kubelet/kubeconfig` (the same flag controller-runtime auto-registers; we deliberately don't define our own to avoid a panic). Cache filtered to Pods scheduled to this Node + the Node object itself.
- [`internal/podagent/`](infra/tart-kubelet/internal/podagent/) — Pod reconciler. Pull/Clone/Run idempotent on Pod identity; existing local clones are reused on agent restart instead of re-pulled.
- [`internal/podagent/garbage.go`](infra/tart-kubelet/internal/podagent/garbage.go) — disk GC. Periodic ticker + on-demand RunOnce. Drops local VM clones whose Pod no longer schedules to this Node and OCI cache entries no Pod's container image references. Wired into the reconciler's Pull error path so a "not enough space" failure triggers a sweep + retry.
- [`internal/envresolver/`](infra/tart-kubelet/internal/envresolver/) — resolves Pod env exactly the way kubelet does (envFrom Secret / ConfigMap dumps, valueFrom secretKeyRef / configMapKeyRef / common fieldRef paths). Optional refs silently skip on miss; required refs return an error so the Pod requeues.
- [`internal/nodeagent/`](infra/tart-kubelet/internal/nodeagent/) — Node registration + heartbeat. Sets `kubernetes.io/os=darwin`, `kubernetes.io/arch=arm64`, `tuist.dev/runtime=tart`, taints `tuist.dev/macos:NoSchedule`, advertises capacity from CLI flags.
- [`internal/tart/`](infra/tart-kubelet/internal/tart/) — local Tart wrapper (Pull / Clone / Set / Run / Stop / Delete / Get / List / IP). `tart run` is started directly via `os/exec` with `Setsid: true` so it lands in its own session and survives `launchctl bootout` during agent upgrades.

Tested with the standard Go testing tools: env resolution, Tart command construction, GC's `IsNoSpaceError` matcher, and the env-file staging path.

### `infra/cluster-api-provider-scaleway-applesilicon/` — the provisioner

Cluster API infrastructure provider. Watches `ScalewayAppleSiliconMachine` CRs and reconciles them against Scaleway's API, then SSHes in to bootstrap each Mac mini. The chart's `macosFleet.*` block declares the desired-state shape; operators only ever touch declarative resources.

- [`api/v1alpha1/`](infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/) — three CRDs: `ScalewayAppleSiliconMachine`, `ScalewayAppleSiliconMachineTemplate`, `ScalewayAppleSiliconCluster`. Status carries `tartKubeletBinarySHA` so the controller's drift loop converges (the field is in `crds/scalewayapplesiliconmachines.yaml` so the API server doesn't prune it on writes).
- [`controllers/`](infra/cluster-api-provider-scaleway-applesilicon/controllers/) — Reconciler with three lifecycle stages (Provisioning → Bootstrapping → Drift-detection on operator binary SHA), idempotent at every stage, finalizer-driven Scaleway release on delete.
- [`internal/scaleway/`](infra/cluster-api-provider-scaleway-applesilicon/internal/scaleway/) — Scaleway SDK wrapper (Apple Silicon API + IAM API).
- [`internal/bootstrap/`](infra/cluster-api-provider-scaleway-applesilicon/internal/bootstrap/) — SSH-driven bootstrap: passwordless sudo, macOS GUI auto-login (Tart's Virtualization.framework requires a live console session), Tart install via Homebrew, kubeconfig drop, tart-kubelet binary upload (read from `/opt/tart-kubelet/` baked into the operator's image), launchd plist install. The plist runs the agent as the SSH user (`UserName: m1`) so the same uid that holds the GUI console session also runs `tart`; Apple's VZ otherwise rejects with "Failed to get current host key".
- [`internal/credentials/`](infra/cluster-api-provider-scaleway-applesilicon/internal/credentials/) — per-fleet Ed25519 SSH key auto-generated on first reconcile and registered with Scaleway via IAM.
- [`internal/kubeconfig/`](infra/cluster-api-provider-scaleway-applesilicon/internal/kubeconfig/) — builds a self-contained kubeconfig per Mac mini from the configured external API server URL, the in-cluster CA, and the shared `tart-kubelet` ServiceAccount token Secret. UpdateTartKubelet refreshes the kubeconfig on every reconcile so token rotation and new hosts both pick up the right credentials.
- [`Dockerfile`](infra/cluster-api-provider-scaleway-applesilicon/Dockerfile) — multi-stage. Stage 1 builds the linux operator. Stage 2 cross-builds the darwin/arm64 tart-kubelet binary. Final image includes both. **One image tag pins both pieces; rolling the operator rolls the agent across the fleet.**

### `infra/xcresult-processor-image/`

Packer template against `ghcr.io/cirruslabs/macos-tahoe-xcode:26.4`. The image's `inject-env.sh` reads the env staged by tart-kubelet at `/Volumes/My Shared Files/env/tuist.env` (Tart `--dir env:<host>:ro` flow) into `/etc/tuist.env` before launchd starts the BEAM. Backward-compatible with the legacy `--user-data` JSON path so the same image works under both this architecture and the Orchard-based production fleet.

### Helm chart

- [templates/xcresult-processor-deployment.yaml](infra/helm/tuist/templates/xcresult-processor-deployment.yaml) — the macOS workload as a standard Deployment with `nodeSelector: kubernetes.io/os=darwin`. Image pin in `values-managed-<env>.yaml`.
- [templates/capi-scaleway-applesilicon.yaml](infra/helm/tuist/templates/capi-scaleway-applesilicon.yaml) — operator Deployment + ServiceAccount + ClusterRole. Reads Scaleway IAM creds + the agent rollout config from values.
- [templates/tart-kubelet-rbac.yaml](infra/helm/tuist/templates/tart-kubelet-rbac.yaml) — fleet-shared `ServiceAccount` + `ClusterRole` + `ClusterRoleBinding` granting Pod / Node / Secret / ConfigMap reads. The matching long-lived SA token Secret is annotated so the API server populates `data.token` + `data.ca.crt` automatically.
- [templates/macos-fleet.yaml](infra/helm/tuist/templates/macos-fleet.yaml) — declarative fleet: `Cluster` + `ScalewayAppleSiliconCluster` + `ScalewayAppleSiliconMachineTemplate` + `MachineDeployment`. Replicas come from `values-managed-<env>.yaml`.
- The existing `TUIST_DELEGATE_PROCESS_XCRESULT=1` handoff on the server Deployment from the consolidation commit stays.

### Decommissioning

- Standalone `xcode_processor/` (nix-darwin platform configuration, mise tasks, sops bundle) deleted in full.
- `.github/workflows/xcode-processor-deploy.yml` deleted; deploys flow through the standard `server-deployment` cascade now.
- `infra/vk-orchard/` and `infra/vk-applesilicon/` deleted (Virtual Kubelet provider iterations).
- `server/lib/tuist/orchard/` and the embedded Orchard control plane deleted.
- `server/lib/tuist/macos/` (the Elixir fleet manager from iteration 4) deleted.
- `infra/tart-cri/` (the upstream-kubelet-via-CRI iteration) deleted.

## (De)provisioning story

**Provisioning a Mac mini (declarative, end-to-end):**

```bash
# One-time per env: drop Scaleway IAM creds in 1P (access-key,
# secret-key, project-id) at op://tuist-k8s-<env>/SCALEWAY_API/.
# ESO syncs them into the cluster automatically.

helm upgrade tuist infra/helm/tuist \
  -f infra/helm/tuist/values-managed-<env>.yaml \
  --set macosFleet.enabled=true
```

That's it. The chart applies the `MachineDeployment` (replicas from `values-managed-<env>.yaml`); the operator picks up the new `ScalewayAppleSiliconMachine`, calls Scaleway's API to order the host, generates an Ed25519 SSH key on first reconcile, SSHes in to install Tart, drops a kubeconfig built from the chart-managed shared SA token, uploads the tart-kubelet binary it ships with, loads the launchd job. tart-kubelet authenticates and registers itself as a Node.

**Scaling the fleet:**

```bash
kubectl scale machinedeployment tuist-tuist-macos-fleet --replicas=4
```

That's the only path today and it's the source of truth. Two upgrades plotted for later, neither blocking on this PR:

- **/ops LiveView page.** The server pod gets a narrowly-scoped `ServiceAccount` with `patch` on `machinedeployments/scale` only (the `/scale` subresource is the smallest right that works). A LiveView reads current vs desired replicas from `.status.replicas`, lists each `ScalewayAppleSiliconMachine` CR with its phase (Provisioning / Bootstrapping / Ready), and exposes +/- buttons capped at a max defined in `values-managed-<env>.yaml`. CAPI takes 60-120 s to reconcile a new machine, so the page polls. Audit (who scaled, before, after) lands in oban. Scoped to the local cluster only, one /ops per env.
- **cluster-autoscaler.** CAPI integrates with cluster-autoscaler via the `cluster-api-autoscaler-node-group-min-size` / `-max-size` annotations on the MachineDeployment. Once configured, the autoscaler reacts to Pending Pods on its own and is the natural fit for the runners pool. Pick one source of truth: manual /ops UI vs autoscaler. If both touch `spec.replicas`, they will fight; the LiveView would either disable itself when the autoscaler is on or constrain its inputs to the autoscaler's `min/max` bounds.

**Replacing a wedged host:**

```bash
kubectl delete machine <machine-name>
# MachineSet immediately creates a replacement; operator releases the
# old Mac mini back to Scaleway in the delete-reconcile.
```

**Rolling a new tart-kubelet across the fleet:**

```bash
# Land the operator-image change. Bumping the chart's
# `macosFleet.image.tag` triggers the operator's drift-detection
# stage, which re-uploads the binary + reloads launchd on each Mac
# mini. Running Tart VMs survive (Setsid + state recovery).
helm upgrade ... --set macosFleet.image.tag=<new-sha>
```

**Decommissioning the entire fleet:**

```bash
helm upgrade tuist ... --set macosFleet.enabled=false
# CAPI core unwinds the MachineDeployment → Machines → Scaleway servers.
```

## What's intentionally not implemented yet

- **Cross-host pod-to-pod routing.** No CNI plugin yet — Pods on a Mac mini can egress freely but are not reachable from other nodes' Services. The xcresult-processor workload doesn't need it (egress only). The runners product will need it for log-streaming back to the cluster — gated on that work.
- **Pod resource accounting / Limits enforcement.** tart-kubelet advertises CPU + memory capacity via flags but doesn't actually slice CPU or memory across Pods on the same host. Concurrent Pods rely on Tart's own VM isolation and the per-host max-pods cap. Sufficient for one xcresult Pod per host today; runners' multi-tenant story will need at least a limit-bounded admission check.
- **`Exec` / `Attach` / `PortForward` / `Stats` / `Logs`.** The agent doesn't proxy these to the VM. `kubectl logs` against an xcresult-processor Pod returns nothing useful today; debugging happens via the host-side `/var/log/tart-vms/<vm>.log` and SSH into the VM. Adding these is gated on need.
- **Image GC by size.** The current GC drops orphan clones + OCI cache entries by Pod-membership, not by total disk usage. If a single host has a lot of distinct Pod images we may still hit `tart pull: not enough space`; the on-Pull retry path handles the immediate failure but a smarter LRU eviction is the proper fix.

## Pre-runners follow-ups

The second-pass review called out three items as "should land before customer runner Pods ride on this substrate." Two were small enough to land in this PR; the third is a deliberate-design item flagged for a follow-up.

**Landed in this PR:**

- **Pod-deletion finalizer on tart-kubelet.** Every Pod the reconciler accepts gets `tart-kubelet.tuist.dev/vm-cleanup` added on first reconcile after admission. On `DeletionTimestamp != nil` the reconciler runs deletePod (Stop + Delete on the VM) before removing the finalizer, so the API server can't complete deletion until the VM is actually gone. Closes the kubectl-force-delete / mid-flight-crash hole that previously left orphan VMs while kubectl reported the Pod gone.
- **Drift-loop retry cap in the CAPI provider.** New `Status.TartKubeletUpdateAttempts` counter + `--tartkubelet-max-update-attempts` flag (default 5). Once consecutive failures cross the cap, the CR transitions to a terminal Failed state with `FailureReason=TartKubeletUpdateExceededRetries`; the reconcile gate skips the drift loop while Failed so the operator stops SSH-hammering a wedged host every 60s. Recovery: `kubectl patch` to clear FailureReason + zero the counter. Cap=0 disables the terminal transition (escape hatch on the flag, not recommended for prod). 4 unit tests cover the increment / cap / before-cap / disabled paths.

**Deferred:**

- **Node-scoped per-machine RBAC.** The per-machine SA addresses identity *separation*; its bound ClusterRole is still fleet-wide read on Pods/Secrets/ConfigMaps. A compromised Mac mini can enumerate other Nodes' Pods (read-only). Proper Node-graph scoping means enabling NodeAuthorizer + NodeRestriction admission and authenticating each kubelet as `system:node:<name>` via TLS bootstrap (CSRs + cluster-CA-signed certs). Bigger lift than the first two; the BYOC plan above sequences this before the AWS path lands.

## Test plan

- [x] `go test ./...` in `infra/tart-kubelet/` and `infra/cluster-api-provider-scaleway-applesilicon/`.
- [x] `go vet ./...` clean for both modules.
- [x] `go build ./...` for both modules.
- [x] `mix compile --warnings-as-errors` clean after Elixir-side cleanup.
- [x] `mix format --check-formatted` clean.
- [x] `mix test` 2365 / 2365 green.
- [x] `helm lint` + `helm template` for default + `values-managed-staging.yaml`.
- [x] Operator image built + pushed.
- [x] xcresult-processor Tart image built + pushed.
- [x] **Mac mini in staging joins the cluster as a real Node, xcresult-processor Pod schedules + boots a Tart VM, the BEAM consumes the `:process_xcresult` Oban queue.**
- [x] **End-to-end xcresult upload from `tuist inspect test` reaches `status: succeeded` against staging, with parsed test cases + their pass/fail statuses landing in ClickHouse.**
- [ ] Canary + production cutover.

## Cutover plan

The standard deploy pipeline (`server-deployment.yml`) handles the rollout — no manual `kubectl apply` or `helm --set` flag flips. Each deploy:

1. Re-applies CRDs from `infra/helm/tuist/crds/` (idempotent).
2. Re-applies upstream CAPI core (idempotent).
3. Runs `helm upgrade --install` with the env's values overlay.

Cutover sequence:

1. Land this PR. Operator + tart-kubelet (baked into operator image) + the xcresult Tart image are pushed by CI on merge.
2. Drop Scaleway IAM creds in 1Password (canary + production deferred until staging proves out end-to-end).
3. Trigger the staging deploy. Mac minis bootstrap, tart-kubelet registers them as Nodes, xcresult-processor Pods schedule.
4. Verify: a real xcresult upload flows through to ClickHouse (`status: succeeded` on the test-run resource).
5. Canary, then production, same playbook.
6. Decommission the old nix-darwin Mac minis once the new k8s-native fleet has been healthy in production for at least a day.


.github/workflows/xcresult-processor-image.yml
```diff
@@ -53,38 +53,81 @@ jobs:
...
+      - name: Reclaim Tart disk
+        run: |
+          set -uo pipefail
+          tart list --format json || true
+          df -h /
+          tart stop tuist-xcresult-processor 2>/dev/null || true
+          tart delete tuist-xcresult-processor 2>/dev/null || true
+          tart prune --entries caches --older-than 0
+          tart prune --entries vms --older-than 1
+          df -h /
```
fortmarek, line 124: this is not exactly ideal, but the VM preparation needs to happen on bare metal (can't build a VM image in a VM...), so we need to adding some prep steps that would typically not be necessary in ephemeral runners.

infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
```diff
@@ -0,0 +1,433 @@
...
+
+
+func (r *ScalewayAppleSiliconMachineReconciler) reconcileDelete(
+	ctx context.Context,
+	machine *infrav1.ScalewayAppleSiliconMachine,
+) (ctrl.Result, error) {
+	logger := log.FromContext(ctx)
+	machine.Status.Phase = "Deleting"
+
+	if machine.Status.ServerID != "" {
```
pepicrft, line 407: This releases the Scaleway server but leaves the per-machine kubelet identity behind: the ServiceAccount, service-account-token Secret, ClusterRoleBinding, bootstrap Secret, and Node object are never removed. Since these tokens are long-lived and bound to a ClusterRole that can read Secrets and ConfigMaps cluster-wide, deleting a Machine leaves valid credentials behind. Can we clean up the node identity and Node object before removing the finalizer?

infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
```diff
@@ -0,0 +1,433 @@
...
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	patchHelper, err := patch.NewHelper(machine, r.Client)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+	defer func() {
```
pepicrft, line 131: This deferred patch error is swallowed because Reconcile does not use named return values. Assigning to the local err here cannot change the already evaluated return values, so a failed patch after CreateServer can return success without persisting Status.ServerID. The next reconcile can then order another Mac mini and leak hardware. Can we return the patch error explicitly or switch this function to named returns?

pepicrft (commented):
I left two blockers from the controller review.

infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
fortmarek: Done in commit `d4dc50e161`. `reconcileDelete` now runs four stages before removing the finalizer, each returning `RequeueAfter` on failure so the CR stays around until cleanup completes:

1. **Scaleway server** — unchanged.
2. **`credentials.Manager.DeleteNodeIdentity`** ([`internal/credentials/node_identity.go:180`](https://github.com/tuist/tuist/blob/claude/xcresult-processor-k8s/infra/cluster-api-provider-scaleway-applesilicon/internal/credentials/node_identity.go#L180)) — deletes the `ClusterRoleBinding` first so the token loses cluster-wide read on Secrets/ConfigMaps before the token Secret is removed, then deletes the token Secret and ServiceAccount.
3. **`credentials.Manager.DeleteMachineBootstrap`** ([`internal/credentials/machine.go:101`](https://github.com/tuist/tuist/blob/claude/xcresult-processor-k8s/infra/cluster-api-provider-scaleway-applesilicon/internal/credentials/machine.go#L101)) — deletes the per-machine bootstrap Secret (sudo password, SSH username, TOFU host fingerprint).
4. **Cluster Node object** — best-effort `client.Delete` against `&corev1.Node{Name: machine.Name}`. The kubelet can't deregister itself once the host is gone; without this the Node would linger as `NotReady` forever and confuse drain/scale tooling.

Each helper is idempotent (treats `NotFound` as success), so a partial-cleanup crash retries cleanly. The chart's ClusterRole and the kubebuilder markers grew the additional `delete` verb on `nodes` (the chart already had `delete` on `serviceaccounts` / `clusterrolebindings` / `secrets`).

infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
fortmarek: Switched to named returns in commit `d4dc50e161`. `Reconcile`'s signature is now:

```go
func (r *ScalewayAppleSiliconMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, err error) {
```

The deferred `patchHelper.Patch` can now promote a patch failure into the actual return value:

```go
defer func() {
    if patchErr := patchHelper.Patch(ctx, machine); patchErr != nil && err == nil {
        err = patchErr
    }
}()
```

Local error variables that previously shadowed the outer `err` were renamed to `getErr` / `helperErr` / `patchErr` to remove the ambiguity. So a patch failure after a successful `CreateServer` now propagates as a non-nil reconcile error, controller-runtime requeues, and the next reconcile sees the persisted `Status.ServerID` (no second Mac mini ordered).

pepicrft (approved):

fortmarek:
## State summary before merge

### What's validated end-to-end on staging

Full `tuist inspect test --mode remote` against `staging.tuist.dev` succeeded:
- CLI uploaded the result bundle ([test run d55fa63d](https://staging.tuist.dev/marek/tuist/tests/test-runs/d55fa63d-0d4b-4dce-bef3-b3b5ed273828))
- Server enqueued `:process_xcresult` Oban job
- xcresult-processor BEAM inside the Tart VM on the Mac mini consumed the job
- Both `AppFrameworkTests` (success) and `AppTests` (failure, expected from the fixture) modules parsed and persisted

### Fixes landed in this PR (chronological)

| Commit | What |
|---|---|
| `33020d8e0d` | `helm.sh/resource-policy: keep` on `ScalewayAppleSiliconMachine` — CR survives helm rollbacks instead of getting stuck Terminating |
| `b494b506ec` | Pre-install/pre-upgrade helm hook on the operator's ExternalSecret — blocks helm on ESO sync so the operator pod doesn't crashloop on missing secret |
| `625763a3c1` | Operator gets explicit `get` on `kube-public/cluster-info` — auto-discovery of API server URL works on Hetzner ClusterClass |
| `b5e97069f4` | reconcileNormal detects missing Node post-Bootstrap and re-runs Stage 2 (re-bootstraps existing Mac mini, no Scaleway re-provisioning) |
| `44ed0240b0` | bootstrap.go does `sudo killall -HUP loginwindow` after auto-login pref is written — brings up the Aqua session on headless Apple Silicon Macs that boot-time loginwindow refuses to populate without a display attached. Without this, VZ refuses with `Failed to get current host key` and every `tart run` fails. |
| `97abdeb57c` | tart-kubelet IP-wait 2m → 8m — absorbs macOS guest cold-boot before TartCreateFailed fires |
| `d2c7a1ffa4` | helm `--timeout 60m → 90m`, job `timeout-minutes: 75 → 105` — production cold-start headroom for ~50min Scaleway provisioning + Tart cold-boot |
| `f8f111e165`, `9d28795ae0` | Repoint `XCResultNIF` SwiftPM package + `XCResultParser` buildable folder to `server/native/xcresult_nif/` after the directory move |
| `32523be662` | `app.yml` cache key now includes `Package.swift` — Swift package edits invalidate the `.build` cache instead of restoring stale checkouts |
| `5b8bd9397d` | `xcresultProcessor.replicaCount=1` on canary, matching `macosFleet.replicas=1` (production stays at 2 with 2 Macs) |
| `096269ebaa` | tart-kubelet `--max-pods 2 → 3` — Apple's SLA caps 2 macOS *VMs* per host (Tart enforces this); kubelet `max-pods` is a Pod cap. The hcloud-csi-node DaemonSet takes 1 slot, leaving only 1 for xcresult-processor at the cap. Rolling-update surge needs 3 (csi + old + new) → max-pods=3. Without this, every xcresult-processor image bump deadlocks the rollout on every env. |
| `41f1a9cafb` | helm step-level `timeout-minutes: 95` — runner-level kill ceiling 5min over helm's `--timeout 90m`. Bounds the worst-case for hung helm cancellations. |

### 1Password vault items added (manual ops, done)

- `tuist-k8s-canary` and `tuist-k8s-production` vaults each have a fresh `SCALEWAY_API` item (`access-key`, `secret-key`, `project-id` fields), backed by per-env Scaleway IAM principals (`tuist-canary-capi`, `tuist-capi`) so a leaked staging key rotates without touching production. Verified the canary item is correctly synced by ESO on the canary cluster.

### Canary state at merge time

- The canary cluster is in a transitional state. Run [25435309787](https://github.com/tuist/tuist/actions/runs/25435309787) hung in `helm upgrade --wait` past its 90m timeout (helm ignored cancellation signals — exactly the case `41f1a9cafb` addresses going forward).
- The `ScalewayAppleSiliconMachine` CR survived (chart fix working). Mac mini at Scaleway server `80267854-b10e-4cfb-a9a0-b52b9b41bd33` (IP `62.210.195.181`) is allocated and Node-Ready in the cluster. Two `xcresult-processor` Pods are stuck (one Pending, one TartCreateFailed) reflecting the pre-`replicaCount=1` chart state.
- The post-merge canary deploy from main will pick up all the fixes and converge canary cleanly. No manual action required, just wait for the next canary deploy to land.
- If the next agent picks this up: kubeconfig at `~/.kube/tuist-canary.yaml`; the SSH key + bootstrap creds are in `tuist-tuist-macos-fleet-ssh` and `tuist-tuist-macos-fleet-0-bootstrap` Secrets in the `tuist-canary` namespace.

### Pre-existing CI failures (NOT in this PR's diff)

- **`Device Build`** keeps failing with `createBundle` JSON-decode errors against production tuist.dev. The CLI's OpenAPI types (`server.yml`, `Types.swift`) haven't changed in this branch since `cba310c549` (long before this PR). The failure shape is identical to what main saw after merging `0da02c90de fix(server): build createBundle response without round-tripping through ClickHouse` — looks like prod hasn't rolled that change yet, or there's a related CLI/server response shape mismatch. Worth investigating in a separate PR; orthogonal to the macOS fleet work.
- **`Test`** flaked once on `Cache.WebhookClientTest signed_post/3 allows a successful probe request after the cooldown window` — pre-existing flaky test in the cache module, unrelated to this PR. Re-run resolves.
- **`blick / server/elixir`** consistently flags the `process_xcresult_worker` Oban consumer for "tenant isolation gaps" via `Repo.get` on schemas the server itself enqueued. Internal-only Oban paths processing server-issued job IDs are explicitly out-of-scope per the project's established Blick-feedback policy. The findings are also tagged "outside this PR's diff" by Blick itself.

### Production deploy expectations after merge

The production cascade (`server-production-deployment.yml`) will run canary → acceptance tests → production. With 2 Mac minis, `xcresultProcessor.replicaCount=2` (chart default), and topology spread one-per-host, production cold-start should complete well within helm's 90m timeout (parallel Scaleway provisioning ~50min + Tart cold-boot ~5-8min on each Mac, plus the runner-level 95m step ceiling as backstop).

PR #10509, opened 2026-04-28
feat(server): backfill artifacts from PostgreSQL to ClickHouse (phase 2)

fortmarek:
Follow-up to [#10493](https://github.com/tuist/tuist/pull/10493). Phase 1 set up the destination table, the dual-write, and the per-bundle `artifacts_replicated_to_ch` flag. This PR fills in the long-running backfill that flips that flag for every existing bundle, so phase 3 can cut reads over to ClickHouse.

## Strategy

The migration paginates over bundles where `artifacts_replicated_to_ch = false`, fetches each batch's artifacts from PG, inserts them into CH, and flips the flag in a single PG `UPDATE`. Once the loop drains, every bundle's artifacts live in CH and the live dual-write keeps them in sync from there.

The flag itself doubles as the resume cursor — a killed run restarts and the `WHERE` filter naturally skips already-flipped bundles. This recovers both:
- pre-dual-write history (the bulk of the ~225M rows on production), and
- bundles whose dual-write hit a transient ClickHouse outage post-deploy (mirroring the contract documented in [Tuist.Bundles.ArtifactIngest](https://github.com/tuist/tuist/blob/main/server/lib/tuist/bundles/artifact_ingest.ex)).

## Why inline / raw schemaless queries

The migration is self-contained — no `from(b in Bundle, ...)`, no shared `Tuist.Bundles.ArtifactBackfill` module. Migrations are immutable history, but self-hosted upgrades may run them long after the live `Bundle` / `Artifact` schemas have drifted from the column shape this migration assumes. Referencing a live schema would silently rewrite the meaning of the migration on a multi-version jump. Raw `from(... in "bundles", ...)` and an inline `@ch_types` map pin the migration to the column shape that exists today, matching the convention used by [`BackfillTestRunsFromCommandEvents`](https://github.com/tuist/tuist/blob/main/server/priv/ingest_repo/migrations/20251118211224_backfill_test_runs_from_command_events.exs) and [`BackfillBuildRunsFromPostgres`](https://github.com/tuist/tuist/blob/main/server/priv/ingest_repo/migrations/20260211124936_backfill_build_runs_from_postgres.exs).

UUIDs come out of PG as 16-byte binaries (raw `:uuid` columns, no schema cast); a small `encode_for_ch` helper decodes them to 36-char strings via `Ecto.UUID.load!/1` and bumps timestamp precision before the CH `INSERT`.

## Production-safety knobs (matched against the other long-running CH backfills)

- **`Process.sleep(@throttle_ms = 500)` between batches** — same throttle [`BackfillTestRunsFromCommandEvents`](https://github.com/tuist/tuist/blob/main/server/priv/ingest_repo/migrations/20251118211224_backfill_test_runs_from_command_events.exs) and [`BackfillFirstRunEvents`](https://github.com/tuist/tuist/blob/main/server/priv/ingest_repo/migrations/20260123160000_backfill_first_run_events.exs) use. Gives live PG/CH traffic breathing room and keeps CH merges from falling behind `parts_to_throw_insert`.
- **`log: false`** on the per-batch artifact `SELECT`, the CH `INSERT`, and the PG flag `UPDATE`. Otherwise we'd dump full UUID arrays + every row payload at debug level for ~2,800 batches. The per-batch `Logger.info` summary still surfaces progress.
- **`timeout: :infinity`** on the CH insert and PG update. Default Ecto timeout is 15s; one batch can be tens of thousands of rows.

## Race window

Between the CH `INSERT` and the PG flag `UPDATE` for a given batch. A crash inside that window re-inserts that batch's rows on the next run. The destination is `MergeTree` (no row-level dedup), so duplicates persist until something else collapses them. Bounded by one batch (100 bundles) and an unlikely crash, so we accept it rather than pay the cost of a deduplicating engine or per-batch existence checks. Documented in the moduledoc.

## What's NOT in this PR

- **Read cutover** — phase 3.
- **Dropping PG `artifacts`** — phase 4.
- **Partial PG index on `(id) WHERE artifacts_replicated_to_ch = false`** — would help late-stage scans get faster as more rows flip, but adds a CREATE/DROP step. The migration is bounded at ~30-45 min already; left as a follow-up if production observes the late-stage scans dominating.

## Compliance

`server/data-export.md` updated to note that the CH `artifacts` table now holds the full history (dual-write + backfill), so it is viable for GDPR Article 20 / CCPA exports even before the phase 3 read cutover.

## How to test locally

1. `mise run db:reset && mise run dev` — boots cleanly; the new migration runs during startup and is a no-op on a fresh DB.
2. `mix test test/tuist/bundles_test.exs` — 53/53 passing.
3. End-to-end re-verified against the dev DB: reset 3 bundles' flags, ran the migration directly, observed log
   - `Backfilled batch: 3 bundles / 24 artifacts`
   - `Artifact backfill complete: 3 bundles, 24 artifacts replicated`
   - 500ms gap between batch summary and next cursor query
   CH `SELECT count() FROM artifacts` went from 184 → 208 (+24, exactly as expected). All flags flipped to `true`.

## Reviewer notes

- The migration is the long-running step. Production will run this as a pre-upgrade hook (~30-45 min for 225M rows). Runtime is dominated by PG read latency, not CH ingestion.
- `@bundle_batch_size = 100` — raise if production observes the per-batch round-trip dominates over CH insert throughput. Logs report per-batch counts, so it's easy to tune.


server/priv/ingest_repo/migrations/20260428120000_backfill_artifacts_from_postgres.exs
```diff
@@ -0,0 +1,185 @@
...
+        end
+
+        mark_replicated(bundle_ids)
+        {length(bundle_ids), length(artifacts)}
+    end
+  end
+
+  defp fetch_unreplicated_bundle_ids do
+    from(b in "bundles",
+      where: b.artifacts_replicated_to_ch == false,
```
pepicrft, line 126: This backfill can race with the live dual-write path in `Bundles.create_bundle/2`. New bundles commit their PostgreSQL rows before `artifacts_replicated_to_ch` flips to `true`, so while the migration is running they are still eligible for this scan. If we pick one up here before the live path reaches `mark_replicated!/1`, both paths will insert the same artifacts into the `MergeTree` table and the duplicates will persist. We should freeze the candidate set to preexisting bundles or add a stronger claim and update protocol before phase 3 reads from ClickHouse.

pepicrft (approved):
Approving to keep this moving. Please address the inline concurrency issue before phase 3 reads from ClickHouse.

fortmarek:
Good catch — fixed in c93e006a1c by inverting the optimistic path rather than adding a watermark.

Bundles are now inserted with \`artifacts_replicated_to_ch = true\` (the dual-write owns them from insert time). The CH insert path is a no-op on the flag in steady state; only the rescue flips the flag to \`false\` on failure. Bundles whose dual-write is in flight never enter the backfill scan, so the race goes away by construction rather than by a time buffer.

Trade-off documented in the commit message: a hard BEAM crash between PG commit and the rescue would leave \`flag = true\` with no CH rows. The rescue already catches the realistic failure modes; only a kernel-level kill would slip through. I think that's acceptable, but happy to revisit if you'd prefer a stronger claim/update protocol.

Phase 1's existing tests still pass — the "leaves flag false on CH unavailability" test asserts \`false\` after the rescue runs, just via a different code path. End-to-end re-verified against the dev DB.

PR #10525, opened 2026-04-29
fix(server): bound xcactivitylog parser, dedupe ProcessBuildWorker, cap dirty NIF wall time

fortmarek:
The `process_build` Oban queue was backing up: 22 jobs sitting in `executing` for 26+ minutes, 59k retryable. Investigation traced it to three independent issues stacking on each other.

### Why it was happening

1. **Some xcactivitylogs take much longer than they should to parse.** A monorepo iOS project running an all-tests scheme with `xcode_cache_upload_enabled: true` is the worst case shape. A contributing factor was `flattenBuildSteps` being recursive — on deeply-nested step trees it could blow the stack.
2. **Same `build_id` enqueued twice** (visible in the dashboard at attempt `1/5` for the same id). Race in `get_or_create_build` under concurrent CREATE: both requests find `not_found`, one wins the insert, the loser still falls through `if build.status == "processing"` and enqueues a duplicate job.
3. **The Swift NIF had no escape hatch.** `semaphore.wait()` with no timeout on an `ERL_NIF_DIRTY_JOB_CPU_BOUND` call meant a single pathological log permanently leaked a BEAM dirty CPU scheduler thread — Lifeline can rescue the *job* but cannot kill the OS thread.

### What changed

- **`ProcessBuildWorker`**: `unique: [keys: [:build_id], states: [available, scheduled, executing, retryable], period: :infinity]` + `timeout/1` returning 5 minutes. A build parse should never take longer than that.
- **`XCActivityLogNIF.swift`**: 4-minute bounded `semaphore.wait` with a heap-allocated `ParseResultBox`. The semaphore provides the happens-before edge for safe reads after wait succeeds; on timeout the parent abandons the box and ARC frees it once the Task finishes its (now ignored) write. Cancellation is propagated structurally via `defer { task.cancel() }`.
- **`XCActivityLogParser.swift`**: iterative `flattenBuildSteps` (no stack-overflow risk on deep trees, slightly less peak memory than the nested `flatMap`). Same call sites, same total work — this is purely a stack-safety + cache-locality change, not a complexity reduction.

NIF-level timeout (4 min) is intentionally lower than the worker timeout (5 min) so the NIF returns its own structured error before Oban kills the process.

### Trade-offs and known follow-ups

- The `get_or_create_build` race still exists. The Oban `unique` constraint absorbs the duplicate (the second `Oban.insert!` is a no-op on conflict), so it's harmless but cosmetic — worth a follow-up to dedupe at the controller layer alongside the near-duplicate `runs_controller.ex` path.
- `period: :infinity` means re-enqueueing a `build_id` while a previous job is still in `retryable` is silently dropped. Documented in the worker moduledoc; manual requeue requires cancelling the old job first.
- `XCLogParser` doesn't honor `Task.isCancelled`, so the `defer { task.cancel() }` is currently a no-op. The NIF still returns on time and the BEAM dirty scheduler thread is freed; the runaway Task continues to chew Swift cooperative-pool capacity until it finishes naturally. Worth a follow-up to thread `Task.checkCancellation()` through `parse`.
- Drain the 404-poison retryable jobs (storage_keys whose object no longer exists in S3 — see [TUIST-C5](https://tuist.sentry.io/issues/TUIST-C5), [TUIST-B1](https://tuist.sentry.io/issues/TUIST-B1)). One-shot `Oban.cancel_all_jobs` against prod, not a code change.

### How to test locally

- `cd server && mix test test/tuist/builds/workers/process_build_worker_test.exs` — 14/14 pass.
- `cd server/native/xcactivitylog_nif && swift test --replace-scm-with-registry` — 14/14 pass. `flattenBuildSteps` is exercised transitively through the existing fixture-driven `parse()` tests.

### DIFF
--- a/server/lib/tuist/builds/workers/process_build_worker.ex
+++ b/server/lib/tuist/builds/workers/process_build_worker.ex
@@ -25,10 +25,10 @@ defmodule Tuist.Builds.Workers.ProcessBuildWorker do
   require Logger
 
   # Cap one job's wall time so a single huge xcactivitylog cannot hold a worker
-  # slot indefinitely. Slightly higher than the NIF's internal timeout so the
-  # NIF returns its own error (with telemetry) before Oban kills the process.
+  # slot indefinitely. The NIF's internal timeout is set lower so it returns
+  # a structured error first; this is the outer guard for everything else.
   @impl Oban.Worker
-  def timeout(_job), do: to_timeout(minute: 20)
+  def timeout(_job), do: to_timeout(minute: 5)
 
   @impl Oban.Worker
   def perform(%Oban.Job{
--- a/server/lib/tuist/builds/workers/process_build_worker.ex
+++ b/server/lib/tuist/builds/workers/process_build_worker.ex
@@ -7,6 +7,14 @@ defmodule Tuist.Builds.Workers.ProcessBuildWorker do
   pods (`TUIST_MODE=processor`), so this worker's body executes there;
   the server pods enqueue jobs but never claim them. In self-hosted installs
   the server runs both roles in the same BEAM.
+
+  ## Uniqueness
+
+  Jobs are unique on `build_id` over `available | scheduled | executing |
+  retryable` with `period: :infinity` to absorb the duplicate enqueues from
+  the race in `get_or_create_build`. A re-enqueue while a previous job is
+  still in-flight (or sitting in retryable) is silently dropped — to manually
+  requeue a stuck job, cancel the existing one first.
   """
 
   use Oban.Worker,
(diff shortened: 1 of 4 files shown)


pepicrft (approved):
@fortmarek I was thinking it'd be awesome if we could attribute the jobs to accounts & projects in OTel such that we can, for a particular account, see the state of their processing jobs.

fortmarek:
> I was thinking it'd be awesome if we could attribute the jobs to accounts & projects in OTel such that we can, for a particular account, see the state of their processing jobs.

We can consider doing that, but I'm not 100 % how useful that is over going to the dashboard where you can already see that (and the processing queue should stay limited)

Issue #10325, opened 2026-04-19
String Catalogs (xcstring) in static frameworks marked stale, if added via buildable folder instead of resources


KaiOelfke:
### What happened?

Recently a few PR's got merged to fix stale detection for String Catalogs in static frameworks with resource bundle frameworks.

https://github.com/tuist/tuist/commit/e1c8912bf981358a1ccd5fb5e8cc017684f747f2
https://github.com/tuist/tuist/commit/46736d07a818aa6310c800e37935dfd3e457a9eb

There's also been some other attempts before that I don't list here.

With these fixes my previous testing matrix with generated_ios_app_with_static_framework_with_xcstrings and one of my own projects work as expected.

However, I have one more project, where things don't work. I debugged and looked at the difference. And the root cause seems to be using buildable folders. I'm not 100% sure about how buildable folders work in vanilla Xcode and Tuist generated projects in terms of identifying and mapping resources to the correct places and what's perhaps not supported. But I've seen some PRs and unit tests around resources integrated with buildable folders so I assume this is a valid use case.

I have a sample project attached. My guess is to fix this one just needs to compare the output of using my sample with sources / resources vs buildable folders to see what gets wrong.

### How do we reproduce it?

[generated_ios_app_with_static_framework_with_xcstrings.zip](https://github.com/user-attachments/files/26863299/generated_ios_app_with_static_framework_with_xcstrings.zip)

This is a project adapted from the example fixture with the same name in the tuist repo.

```swift
let project = Project(
    name: "StaticFramework",
    targets: [
        .target(
            name: "StaticFramework",
            destinations: .iOS,
            product: .staticFramework,
            bundleId: "dev.tuist.StaticFramework",
            deploymentTargets: .iOS("16.0"),
//            sources: "Sources/**",
//            resources: ["Resources/**"]
            buildableFolders: ["Sources", "Resources"]
        ),
    ]
)
```

The relevant change is using buildableFolders instead of sources & resources. One thing I noticed myself already is that the string catalog is not in the copy bundle phase for the source target, but it's in the copy bundle phase for the resource bundle framework. So that's opposite of the other configuration and some of the stuff the previous PRs fixed.

### Error log

String catalog json marks all as stale.

### macOS version

26.4 (25E246)

### Tuist version

4.182.0

### Xcode version

Version 26.5 beta 2 (17F5022i)

KaiOelfke:
@pepicrft tagging you here as you worked on these PRs.

[...]

PR #10532, opened 2026-04-29
fix(cli): keep static framework xcstrings on main target Resources phase

pepicrft:
## Summary

Aligns the group-based `target.resources` path with the buildable-folders path so `.xcstrings` end up on the main target's **Resources** phase (not Sources) for static frameworks. Closes the regression reported in the latest comments on #10325.

## Why this was broken

The current code (post #10423) places `.xcstrings` on the main target's **Sources** phase and on the companion resource bundle's Resources phase. With `PACKAGE_RESOURCE_BUNDLE_NAME` set on the main target (intentional, blocks the duplicate compile), swift-build's `XCStringsCompiler.shouldCompileCatalog` returns false. The Sources-phase placement is then only useful via the symbol-generation path, which is gated by `STRING_CATALOG_GENERATE_SYMBOLS`.

`STRING_CATALOG_GENERATE_SYMBOLS` defaults to `NO` and is auto-enabled by swift-build only for `project.isPackage` projects ([Settings.swift:4381-4385](https://github.com/swiftlang/swift-build/blob/main/Sources/SWBCore/Settings/Settings.swift#L4381-L4385)). Tuist-generated `.xcodeproj`s are not packages, so the Sources placement effectively does nothing for them.

The buildable-folders path was already routing `.xcstrings` to the main target's **Resources** phase via the partition logic (`originalTargetExplicitResources`), which is why Kai's update on #10325 said buildable folders worked but source/resource folders did not. adarhef's repro on #10325 (`target.resources` glob, no buildable folders) hits the broken path.

## What changes

In `ResourcesProjectMapper.mapTarget`, when the target is a static framework or a target without native resource support:

- `.xcassets` continues to be appended to the main target's Sources phase (typed asset symbols, unchanged).
- `.xcstrings` from `target.resources.resources` is now appended to `originalTargetExplicitResources` and assigned to `modifiedTarget.resources.resources`. This mirrors the shape produced by the buildable-folders partition.
- `.xcstrings` is still kept on the companion bundle's Resources phase, so `Bundle.module` lookups continue to resolve at runtime (the #10420 fix is preserved).
- `PACKAGE_RESOURCE_BUNDLE_NAME` and `PACKAGE_RESOURCE_TARGET_KIND` settings are unchanged, so the catalog is compiled exactly once (inside the companion bundle).

## Reproduction and verification

Used the [TuistReproduction.zip](https://github.com/user-attachments/files/27197369/TuistReproduction.zip) from adarhef's comment on #10325 (single static framework, `sources` + `resources` globs, `Localizable.xcstrings` under `Resources/`).

Diff of generated `.xcodeproj` between **before** and **after** (only the xcstrings entries change):

```
PBXResourcesBuildPhase (main target):
+ 32C5A55AB9D244854E919F62 /* Localizable.xcstrings in Resources */,

PBXSourcesBuildPhase (main target):
- DC945D9A387AADDBD3F91170 /* Localizable.xcstrings in Sources */,
```

`xcodebuild build` of the framework with the fix produces:

- `TuistReproduction_MyFramework.bundle/Contents/Resources/en.lproj/Localizable.strings` (catalog compiled once, in the bundle)
- No `Localizable.strings` inside `MyFramework.framework`
- One `CompileXCStrings` task in `target 'TuistReproduction_MyFramework'` (the bundle), zero in the main target

## Test plan

- [x] Unit test in `ResourcesProjectMapperTests` updated to assert the new shape (`mapWhenStaticTargetHasXcstringsKeepsThemOnMainResourcesAndBundle`)
- [x] Built `tuist` from this branch and regenerated adarhef's repro; `.pbxproj` matches the expected shape
- [x] Built the framework target and confirmed `Localizable.strings` is in the companion bundle and not in the framework
- [ ] Acceptance test `ios_app_with_static_framework_with_xcstrings` (would appreciate a CI run)
- [ ] Validate end-to-end in Xcode UI that string extraction / stale marking now runs on the `target.resources` path (manual on a populated catalog)

Refs: #10325, #10420, #10247, #10423, #10445

### DIFF
--- a/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
+++ b/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
@@ -94,22 +94,32 @@ public struct ResourcesProjectMapper: ProjectMapping { // swiftlint:disable:this
                 buildableFolders: resourceBuildableFolders
             )
             modifiedTarget.sources = target.sources.filter { $0.path.extension != "metal" }
-            // Asset catalogs and string catalogs are added to the main target's Sources build
-            // phase so Xcode generates typed symbols. This mirrors SwiftPM's PIF builder:
+            // Asset catalogs are added to the main target's Sources build phase so Xcode
+            // generates typed asset symbols. This mirrors SwiftPM's PIF builder:
             //   - https://github.com/swiftlang/swift-package-manager/blob/main/Sources/XCBuildSupport/PIFBuilder.swift#L944-L952
-            //   - https://github.com/swiftlang/swift-package-manager/blob/main/Sources/SwiftBuildSupport/PackagePIFProjectBuilder.swift#L345-L360
-            // Both are also compiled into the companion resource bundle via its Resources phase.
-            // The companion bundle declares PACKAGE_RESOURCE_TARGET_KIND = "resource" (above)
-            // so Xcode treats it as a pure resource container and skips string extraction.
-            // The main target carries PACKAGE_RESOURCE_TARGET_KIND = "regular" (below) so Xcode
-            // runs extraction here where the Swift source references live.
-            let codeGeneratingResourceExtensions: Set<String> = ["xcassets", "xcstrings"]
+            //   - https://github.com/swiftlang/swift-package-manager/blob/main/Sources/SwiftBuildSupport/PackagePIFProjectBuilder.swift#L347-L353
+            // They are also compiled into the companion resource bundle via its Resources phase.
+            let assetCatalogExtensions: Set<String> = ["xcassets"]
             for resource in target.resources.resources {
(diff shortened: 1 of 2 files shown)


adarhef:
Thank you for looking into it so quickly! Sounds like it would fix my issue.

I'm a bit confused though, I thought you said static frameworks must not have any resources directly in order to pass App Store validation? Or are they allowed to have `xcstrings` in particular?

pepicrft:
> I'm a bit confused though, I thought you said static frameworks must not have any resources directly in order to pass App Store validation? Or are they allowed to have xcstrings in particular?

Resources are included in a companion bundle, and then there are some settings necessary on the target side so Xcode sees this connection between the two. Otherwise it treats both as disjointed causing the _mark as stale_ feature to mark strings as stale.

fortmarek (approved):

PR #10544, opened 2026-04-29
feat(server): SCIM 2.0 provisioning + Authentication settings page

fortmarek:
## Summary

Adds inbound SCIM 2.0 provisioning so identity providers (Okta, Azure AD/Entra, JumpCloud, etc.) can manage organization membership in Tuist. Renames the per-org **SSO** settings page to **Authentication** and merges SCIM token management into it as a second card (Linear-style: SSO config on top, SCIM provisioning below).

### SCIM endpoints

Mounted under `/scim/v2/`:

- `POST/GET/PUT/PATCH/DELETE /Users` with `userName eq` and `externalId eq` filters
- `GET/PATCH /Groups` - synthetic `admins` and `users` groups mirroring the existing role hierarchy
- `GET /ServiceProviderConfig`, `/ResourceTypes`, `/Schemas` discovery
- Per-org bearer token backed by an organization-owned `AccountToken` with the `account:scim:write` scope, Bcrypt-hashed at rest, plaintext shown once
- `application/scim+json` mime type registered
- SCIM API requests are rate-limited by authenticated SCIM token
- `X-Forwarded-Proto` / `X-Forwarded-Host` are only reflected into `meta.location` URLs when the forwarded host matches the configured endpoint host

### Identity model

- New `users.active` boolean column. It is enforced in user-backed auth paths: bearer-token (`Tuist.Authentication.authenticated_subject/1`), password login (`Accounts.get_user_by_email_and_password/2`), session-cookie lookup (`Accounts.get_user_by_session_token/2`), JWT account tokens for personal accounts, personal account tokens, and organization account tokens created by an inactive user. Project tokens remain project/service credentials because they do not carry a user lifecycle owner.
- SCIM provisioning creates new users by email. If the email already belongs to a user in the same organization, SCIM updates their requested role. If the email belongs to an existing Tuist user outside the organization, Tuist returns `409 uniqueness` instead of claiming that user into the SCIM-managed organization.
- `DELETE /Users/:id` and Okta-style `PATCH active: false` remove the user’s explicit role in the organization and return an inactive SCIM representation to the IdP. They do not flip the global `users.active` flag, because a Tuist user can belong to multiple organizations.
- Group PATCH membership operations are scoped to current organization members; a SCIM token for one organization cannot add or remove arbitrary global users by ID.
- SCIM bearer tokens use the `tuist_scim_<id>_<secret>` format and are verified with a SCIM-specific hash input, so the same secret cannot be presented as a generic `tuist_<id>_<secret>` account token.

### UI

- Single **Authentication** tab replaces SSO. SSO config is the first section; SCIM provisioning the second.
- **Generate token** opens a modal (token name required), shows the bearer once with a clipboard copy button, and clears state on dismiss so the plaintext is unrecoverable after the modal closes.
- Existing SCIM-scoped account tokens listed in a Noora `<.table>` (Name, Created, Last used, trash action).
- Token revocation is scoped to the current organization before deletion.
- Layout tab renamed; `/sso` route removed (was a few weeks old).

### Docs and seeds

- Adds separate Authentication docs pages for SSO and SCIM, including exact Okta setup steps for the OIDC Web Application plus the SCIM 2.0 Test App (Header Auth), including the `Bearer ` token prefix.
- Documents that SCIM deprovisioning removes organization membership without globally disabling the user.
- Updates `server/data-export.md` for `users.active` and SCIM-scoped account tokens.
- Seeds the local `tuist` organization with Okta SSO defaults and, when `TUIST_OKTA_CLIENT_SECRET` / `TUIST_OKTA_SCIM_TOKEN` (or matching encrypted secrets) are present, configures Okta SSO plus a pre-generated SCIM-scoped account token named `Okta`.

### Test plan

- [x] `mix test test/tuist/scim_test.exs test/tuist/scim/filter_test.exs test/tuist_web/controllers/scim/users_controller_test.exs test/tuist_web/controllers/scim/groups_controller_test.exs test/tuist_web/live/authentication_settings_live_test.exs` - 63/63 pass
- [x] `mix compile --warnings-as-errors` clean
- [x] `mix esbuild app`
- [x] E2E smoke test with a real Okta trial tenant through a Cloudflare Tunnel:
  - Configured Tuist Okta SSO with an Okta OIDC Web Application.
  - Generated a Tuist SCIM token and configured Okta's **SCIM 2.0 Test App (Header Auth)** with the Tuist `/scim/v2` base URL.
  - Verified Okta's **Test API Credentials** succeeds. The Okta **API Token** field must include the `Bearer ` prefix because the Header Auth app forwards it as the `Authorization` header.
  - Enabled Okta provisioning to app for create users, update user attributes, and deactivate users.
  - Assigned an Okta test user and confirmed Tuist created the account and added it to the `tuist` organization members list.
  - Deprovisioned the Okta test user and confirmed the Tuist organization membership was removed.
  - Verified the generated-token modal clipboard button copies the plaintext token inside the dialog.

### Known follow-ups (deferred)

- Reserved-handle fallback: an IdP provisioning `admin@yourco.com` returns 400 today (handle is in `:blocked_handles`). Easy to add auto-suffix on `validate_exclusion` errors.
- Plan-tier gate (Pro/Enterprise only) - currently ungated.
- OpenAPI spec entry for SCIM endpoints (SCIM has its own discovery so this isn't strictly needed).


server/assets/app/js/Clipboard.js
```diff
@@ -1,11 +1,21 @@
...
-      if (value) {
-        navigator.clipboard.writeText(value);
+      if (!value) {
+        return;
       }
+
+      event.preventDefault();
+
+      copyTextToClipboard(value, {
+        container: this.el.closest("[role='dialog']"),
```
pepicrft, line 15: This seems very copuled to where this hook is being used no? Should we make it more generic?

pepicrft (commented):
I ran a full static + dynamic review of the SCIM implementation — reproduced all findings locally with ExUnit tests (36 tests, 0 failures). Two issues are critical (cross-tenant takeover chains); the rest are high/medium. Inline comments below mark the exact locations. Happy to pair on the fixes.

server/lib/tuist/scim.ex
```diff
@@ -0,0 +1,543 @@
...
+    end)
+  end
+
+  defp extract_member_ids(%{"value" => v}) when is_binary(v), do: [v]
+  defp extract_member_ids(_), do: []
+
+  defp add_member(organization, user_id, role) do
+    case Accounts.get_user_by_id(user_id) do
+      %User{} = user ->
+        :ok = Accounts.add_user_to_organization(user, organization, role: role)
```
pepicrft: **[Critical] Cross-tenant membership injection via `add_member/3`**

`Accounts.get_user_by_id/1` is a global lookup — it returns any user in the system regardless of which organization they belong to. An attacker who controls a SCIM token for org A can call `PATCH /scim/v2/Groups/{any_group_id}` with an `add` op containing the ID of a user from org B, and this line silently adds that user to org A without their consent.

Once the victim is in the attacker's org, they can be targeted by the deactivation path (see `set_active` comment below).

**Fix:** Before calling `add_user_to_organization`, assert the user is already a member of `organization`, or — if the intent is to provision on add — restrict `get_user_by_id` to members of the current org:

```elixir
defp add_member(organization, user_id, role) do
  case get_user(organization, user_id) do  # scoped to org
    {:ok, %User{} = user} ->
      :ok = Accounts.add_user_to_organization(user, organization, role: role)
      Accounts.update_user_role_in_organization(user, organization, role)
    _ ->
      {:error, :not_found}
  end
end
```

server/lib/tuist/scim.ex
```diff
@@ -0,0 +1,543 @@
...
+    with {:ok, user} <- get_user(organization, user_id) do
+      :ok = remove_user_role_from_organization(user, organization)
+      set_active(user, false)
+    end
+  end
+
+  defp set_active(%User{active: same} = user, same), do: {:ok, user}
+
+  defp set_active(%User{} = user, active) do
+    user |> User.active_changeset(active) |> Repo.update()
```
pepicrft: **[Critical] `set_active` flips a global flag without org-scoping**

`users.active` is a system-wide field. Setting it to `false` here locks the user out of *every* organization and every product surface (CLI auth, web login, JWT refresh all gate on `user.active`).

Combined with the `add_member` injection above, the full attack chain is:
1. Attacker POSTs `/scim/v2/Users` with victim's email → existing-user branch silently adds victim to attacker's org (no consent check in `provision_user`).
2. Attacker PATCHes `/scim/v2/Users/{victim_id}` with `active: false` → this line globally deactivates the victim.

**Fix:** Deactivation via SCIM should only mean "remove from this organization". Use `deactivate_user/2` (which already removes the org role) and leave `users.active` alone, or introduce a per-membership `active` state rather than a global flag.

server/lib/tuist/scim.ex
```diff
@@ -0,0 +1,543 @@
...
+  end
+
+  @doc """
+  Provisions a user from a SCIM `POST /Users` request.
+
+  If a user with this email already exists, they are added to the organization
+  with the requested role (idempotent if already a member). Otherwise a new
+  user account is created and added.
+  """
+  def provision_user(%Organization{} = organization, attrs) do
```
pepicrft, line 219: **[High] `provision_user/2` runs 3 separate writes outside a transaction**

`create_user`, `add_user_to_organization`, and `set_active` are three independent database operations. If the process crashes or the DB rejects any step after the first, the user is left in a partially-created state (e.g. user account exists but has no org membership, or has an org membership but is still marked inactive).

**Fix:** Wrap in `Ecto.Multi` / `Repo.transaction/1`:

```elixir
def provision_user(%Organization{} = organization, attrs) do
  email = ...
  Repo.transaction(fn ->
    user = ...
    :ok = Accounts.add_user_to_organization(user, organization, role: role)
    {:ok, user} = set_active(user, active)
    Repo.preload(user, :account)
  end)
end
```

server/lib/tuist/scim.ex
```diff
@@ -0,0 +1,543 @@
...
+    end
+  end
+
+  defp maybe_update_email(multi, user, %{user_name: new_email}) when is_binary(new_email) do
+    downcased = String.downcase(new_email)
+
+    if downcased == user.email do
+      multi
+    else
+      Multi.update(multi, :email, Ecto.Changeset.change(user, email: downcased))
```
pepicrft: **[High] `maybe_update_email` uses a raw changeset — no `unique_constraint` on email**

`Ecto.Changeset.change/2` skips all validations. If an IdP tries to rename a user to an email that already belongs to another Tuist user, Postgres raises a unique-violation exception that is **not** caught by the surrounding `Ecto.Multi` — it propagates as an uncaught `Ecto.ConstraintError` and results in a 500.

**Fix:** Use the proper changeset so the constraint is translated into a friendly `{:error, changeset}` tuple:

```elixir
Multi.update(multi, :email, User.email_changeset(user, %{email: downcased}))
```

If `User` doesn't expose a focused email changeset, add one that calls `validate_format` + `unique_constraint(:email)` + `put_change(:email, downcased)`.

server/lib/tuist/scim.ex
```diff
@@ -0,0 +1,543 @@
...
+          downcased = String.downcase(value)
+          from u in base, where: u.email == ^downcased
+
+        %{attribute: "externalId", op: :eq, value: value} ->
+          from u in base, where: u.email == ^String.downcase(value) or u.token == ^value
+
+        nil ->
+          base
+
+        _other ->
```
pepicrft, line 169: **[High] Unsupported filter attributes silently return the full member roster**

The `_other ->` clause falls through to `base` (the unfiltered members query), so `GET /scim/v2/Users?filter=displayName eq "anything"` returns all users in the organization rather than an empty list or a `400 invalidFilter`.

The SCIM spec (RFC 7644 §3.4.2.2) says the server MUST return a `400 invalidFilter` response if it does not support a requested filter attribute.

**Fix:**

```elixir
_other ->
  # unsupported filter — return an empty result set (or raise {:error, :unsupported_filter})
  from u in base, where: false
```

Or propagate `{:error, :unsupported_filter}` and have the controller reply with 400.

server/lib/tuist_web/controllers/scim/users_controller.ex
```diff
@@ -0,0 +1,157 @@
...
+      {:error, :missing_user_name} ->
+        send_scim_error(conn, 400, "userName is required", "invalidValue")
+
+      {:error, :invalid_email} ->
+        send_scim_error(conn, 400, "userName must be a valid email address", "invalidValue")
+
+      {:error, %Ecto.Changeset{} = changeset} ->
+        send_scim_error(conn, 400, format_changeset(changeset), "invalidValue")
+
+      {:error, _other} ->
```
pepicrft, line 69: **[High] `{:error, :email_taken}` falls into the catch-all 500**

When two concurrent `POST /scim/v2/Users` requests race for the same email, `provision_user` can return `{:error, :email_taken}`. That atom does not match any of the explicit `else` clauses above, so it hits `{:error, _other}` and the caller receives a 500.

SCIM clients (Okta, Azure AD) will retry on 500, potentially creating duplicate-provision storms.

**Fix:** Add an explicit clause:

```elixir
{:error, :email_taken} ->
  send_scim_error(conn, 409, "User already exists", "uniqueness")
```

server/lib/tuist/scim.ex
```diff
@@ -0,0 +1,543 @@
...
+  @doc """
+  Issues a new SCIM bearer token for an organization. Returns `{:ok, {token,
+  plaintext}}`. The plaintext is shown to the user once and never persisted.
+  """
+  def create_token(%Organization{} = organization, attrs \\ %{}) do
+    account = organization_account(organization)
+    raw = Base64.encode(:crypto.strong_rand_bytes(24))
+    encrypted = Bcrypt.hash_pwd_salt(raw <> Environment.secret_key_password())
+
+    case %AccountToken{}
```
pepicrft: **[High] `create_token/2` bypasses `AccountToken.create_changeset/1` validators**

This function uses `Ecto.Changeset.change/2` directly instead of `AccountToken.create_changeset/1`. As a result:
- Token names with spaces (e.g. `"Okta SCIM Token"`) are accepted here but would be rejected by the canonical changeset.
- Names longer than 32 chars (canonical limit) are accepted here (max 64 is set inline).
- Names are not downcased, so `"MyToken"` and `"mytoken"` can coexist.

This divergence will silently break any code that creates `AccountToken`s via the shared changeset path and assumes consistent name formatting.

**Fix:** Define a dedicated `AccountToken.scim_changeset/1` (or extend `create_changeset/1` with a `:scim` flag) so the token creation goes through a single codepath, and align the validation rules intentionally.

server/lib/tuist/scim.ex
```diff
@@ -0,0 +1,543 @@
...
+    else
+      _ -> {:error, :invalid_token}
+    end
+  end
+
+  def authenticate_token(_), do: {:error, :invalid_token}
+
+  defp account_token_plaintext(plaintext) do
+    case String.split(plaintext, "_", parts: 4) do
+      ["tuist", "scim", token_id, raw] -> "tuist_#{token_id}_#{raw}"
```
pepicrft: **[Medium] `account_token_plaintext/1` enables dual-format authentication**

This function strips the `scim_` segment before forwarding to the generic token lookup:
```
tuist_scim_<id>_<raw>  →  tuist_<id>_<raw>
```

That means the same secret authenticates against **two different token formats**. Any future code that distinguishes SCIM tokens from regular account tokens by prefix check will be bypassable by presenting the `tuist_<id>_<raw>` form.

**Fix:** Store a dedicated SCIM token type or prefix in the lookup itself so the SCIM bearer credential only matches `account:scim:write` scoped tokens and cannot be used where a plain account token is expected.

server/lib/tuist_web/plugs/scim_auth_plug.ex
```diff
@@ -0,0 +1,94 @@
...
+      ["bearer " <> token] -> String.trim(token)
+      _ -> nil
+    end
+  end
+
+  # Honors X-Forwarded-Proto and X-Forwarded-Host so meta.location URLs in
+  # SCIM responses are correct when running behind a tunnel/proxy (ngrok,
+  # Cloudflare, Render).
+  defp base_url(conn) do
+    forwarded_scheme = first_forwarded(conn, "x-forwarded-proto")
```
pepicrft: **[Medium] `X-Forwarded-Proto` / `X-Forwarded-Host` are reflected without a trust check**

`first_forwarded/2` accepts these headers unconditionally. Any client that can reach the SCIM endpoints (i.e. anyone with a valid SCIM token) can inject arbitrary values:

```
X-Forwarded-Host: attacker.example
X-Forwarded-Proto: https
```

This causes all `meta.location` URLs in the response to point to `https://attacker.example/scim/v2/...`, which can be used to redirect IdP follow-up requests (e.g. user detail fetches) to an attacker-controlled host.

**Fix:** Only trust these headers when the request originates from a known proxy tier (e.g. check `RemoteIp` allow-list, or use Phoenix's built-in `:force_ssl` / `plug :put_secure_browser_headers` mechanisms). At minimum, validate the forwarded host against the configured `TuistWeb.Endpoint` host.

server/lib/tuist_web/router.ex
```diff
@@ -183,6 +183,13 @@ defmodule TuistWeb.Router do
     plug ObservabilityContextPlug
   end
 
+  pipeline :scim_api do
```
pepicrft, line 186: **[Medium] `:scim_api` pipeline has no rate-limit plug**

Every other authenticated pipeline (`:api`, `:mcp_api`, etc.) includes a rate-limit plug. The SCIM pipeline omits one entirely, so an attacker with a valid token can hammer `/scim/v2/Users` at full speed — useful for email enumeration, user deactivation loops, or simply DoS-ing the provisioning flow.

**Fix:** Add a rate-limit plug to the pipeline, for example:

```elixir
pipeline :scim_api do
  plug :accepts, ["scim+json", "json"]
  plug TuistWeb.Plugs.SCIMAuthPlug
  plug TuistWeb.RateLimit.PersistentTokenBucket, ... # or InMemory
  plug SentryContextPlug
  plug ObservabilityContextPlug
end
```

server/lib/tuist/scim.ex
fortmarek: Fixed in 41dd0db590. Group membership changes now use the org-scoped `get_user/2` path instead of global `Accounts.get_user_by_id/1`, so add/replace/remove only operate on users already in the current organization. Added regression coverage in `Tuist.SCIMTest` and `TuistWeb.SCIM.GroupsControllerTest` for attempting to add a user from another organization.

server/assets/app/js/Clipboard.js
fortmarek: Addressed in the follow-up by keeping the hook thin and moving the reusable copy behavior to `assets/shared/js/clipboard.js`. The hook now only reads `data-clipboard-value`, prevents the click default, and passes the nearest dialog as the fallback container so the modal case works without baking modal-specific logic into the helper. `cdca77868a` also applies the required formatter output.

server/lib/tuist/scim.ex
fortmarek: Fixed in `41dd0db590`. SCIM deprovisioning and `active: false` no longer persist a global `users.active = false`; they remove the user from the current organization and return an inactive SCIM representation for that response. The cross-tenant attack chain is also closed because provisioning an existing user outside the organization now returns 409 instead of attaching that user. Added tests for both paths.

server/lib/tuist/scim.ex
fortmarek: Fixed in `41dd0db590`. Unsupported SCIM user filters now return `{:error, :unsupported_filter}` and the controller maps that to a 400 SCIM `invalidFilter` response instead of falling back to the full roster. Added controller coverage for unsupported attributes such as `displayName eq ...`.

server/lib/tuist/scim.ex
fortmarek: Fixed in `41dd0db590`. Added `User.email_changeset/2` with the same validation and unique constraint handling as the main user changeset, and `maybe_update_email` now uses it inside the transaction. There is also regression coverage for a conflicting SCIM email update returning a controlled error instead of a 500.

server/lib/tuist/scim.ex
fortmarek: Fixed in `41dd0db590`. `provision_user/2` now runs inside a `Repo.transaction`, including user creation or lookup, organization membership, role assignment, email updates, and the deprovisioning branch for `active: false`. Errors roll the transaction back instead of leaving partial membership state.

server/lib/tuist/scim.ex
fortmarek: Fixed in `41dd0db590`. SCIM token creation now goes through a dedicated `AccountToken.scim_changeset/1/2` instead of a raw changeset. It enforces the exact SCIM scope, organization ownership, uniqueness constraints, and SCIM display-name validation while still allowing the UI token names we want.

server/lib/tuist/scim.ex
fortmarek: Fixed in `41dd0db590`. The SCIM auth path now parses only the `tuist_scim_<id>_<secret>` format and verifies the `scim:`-prefixed secret directly. It no longer converts SCIM tokens into generic account-token plaintext, and a regression test confirms generic `tuist_<id>_<secret>` tokens are rejected by the SCIM plug.

server/lib/tuist_web/controllers/scim/users_controller.ex
fortmarek: Fixed in `41dd0db590`. The Users controller now handles `{:error, :email_taken}` explicitly and returns a 409 uniqueness SCIM error. `provision_user/2` maps both an existing out-of-organization email and unique-insert races into that error, and the controller test covers the 409 response.

server/lib/tuist_web/router.ex
fortmarek: Fixed in `41dd0db590`. Added `TuistWeb.RateLimit.SCIM` and `TuistWeb.Plugs.SCIMRateLimitPlug`, and the `:scim_api` pipeline now rate-limits after successful SCIM authentication. The limiter keys by SCIM token id, with a remote-IP fallback for unauthenticated requests.

server/lib/tuist_web/plugs/scim_auth_plug.ex
fortmarek: Fixed in `41dd0db590`. Forwarded host is now trusted only when it matches the configured Endpoint host, and forwarded scheme is limited to `http` or `https`. If those checks do not pass, SCIM resource locations fall back to the request host and scheme from the connection.

pepicrft (approved):

PR #10548, opened 2026-04-29
perf(server): speed up dev content compilation

fortmarek:
## Summary
- Lazy-load docs and marketing content at runtime in dev, with file watchers to refresh the caches after content edits.
- Restrict dev locales to English by default, with `TUIST_DEV_ALL_LOCALES=1` to opt back into all locales locally.
- Move dev CLDR data into the shared user cache and skip Boundary in dev compiles.

This intentionally does not duplicate #10547, which handles reverting the shared Mix `_build` directory across worktrees.

## Verification
- `mix format`
- `mix compile`
- `mix test test/tuist/docs_test.exs test/tuist/docs_sidebar_test.exs test/tuist/marketing/changelog_entry_parser_test.exs`


server/lib/tuist/cldr.ex
```diff
@@ -1,11 +1,24 @@
+all_locales = ["ar", "en", "es", "ja", "ko", "ru", "yue-Hant", "zh-Hans", "zh-Hant"]
```
pepicrft: This reads bad and I'd suggest iterating on it. Perhaps we can have a macro that decorates `use Cldr`. Also... Shouldn't we be using [`compile_env/3`](https://hexdocs.pm/elixir/1.12/Application.html#compile_env/3) since we are getting environment variables at compile time?

pepicrft:
The two file watchers (`Tuist.Docs.FileWatcher` and `Tuist.Marketing.FileWatcher`) are nearly identical — same GenServer scaffolding, same debounce timer, same `:file_event` handlers, same 100 ms delay. The only differences are the watched dirs, the extension filter, and which `RuntimeStore` gets `reload/0` called on it.

Could it collapse into a single `Tuist.ContentFileWatcher` parameterised on `{dirs, extensions, store}` and started twice from `dev_content_children/0`? Same for the two `RuntimeStore` modules, which share most of their structure too.

server/lib/tuist/locale.ex
```diff
@@ -12,6 +12,14 @@ defmodule Tuist.Locale do
     %{code: "zh_Hant", label: "Chinese (Traditional)", native: "繁體中文"}
   ]
 
+  dev_all_locales? = System.get_env("TUIST_DEV_ALL_LOCALES") in ~w(1 true TRUE yes YES)
+
+  @languages (if Mix.env() == :dev and not dev_all_locales? do
+                Enum.filter(all_languages, &(&1.code == "en"))
+              else
+                all_languages
+              end)
```
pepicrft: This filtering based on `TUIST_DEV_ALL_LOCALES` is repeated elsewhere, so I'd recommend to figure out a way to do compile-time reusing of this.

pepicrft:
The module-level env reads sitting outside the `defmodule` block (in `gettext.ex`, `cldr.ex`, and similar in `locale.ex`) are pretty rough to read — top-of-file bindings like `all_locales`, `dev_all_locales?`, `allowed_locales` floating before the module starts is unusual for this codebase and easy to miss.

Could the `TUIST_DEV_ALL_LOCALES` parsing move into a single helper (e.g. `Tuist.Environment.dev_all_locales?/0`), and the locale list selection happen inside each module via a module attribute? That would keep the dev gating in one place, avoid three copies of the same `~w(1 true TRUE yes YES)` parsing, and make each module read top-down like the rest of the codebase.

server/lib/tuist/marketing/runtime_loader.ex
```diff
@@ -0,0 +1,87 @@
+defmodule Tuist.Marketing.RuntimeLoader do
```
pepicrft: The `Runtime*` naming hides what these modules actually are. `RuntimeLoader` is a runtime port of NimblePublisher's compile-time build pipeline, and `RuntimeStore` is its cache + reload mechanism. The "Runtime" prefix only means "not at compile time," which is an implementation detail.

Could they live under a `NimblePublisher`-scoped namespace to make the lineage obvious? E.g.:

- `Tuist.Marketing.NimblePublisher.Builder` (this module)
- `Tuist.Marketing.NimblePublisher.Cache` (the GenServer)
- `Tuist.Docs.NimblePublisher.Cache`

That way it's immediately clear they exist to mirror NimblePublisher at runtime, and the name doesn't require readers to reverse-engineer the intent.

server/lib/tuist/marketing/blog.ex
```diff
@@ -3,19 +3,45 @@ defmodule Tuist.Marketing.Blog do
...
+      parser: PostParser,
+      highlighters: [],
+      html_converter: MDExConverter
+
+    @posts Enum.reverse(@posts)
+    @categories @posts |> Enum.map(& &1.category) |> Enum.uniq()
+
+    def get_posts, do: @posts
+    def get_categories, do: @categories
+  end
```
pepicrft, line 22: The same `if Mix.env() == :dev do ... else (use NimblePublisher) ... end` shape is repeated across five modules: `Blog`, `Changelog`, `Customers`, `Newsletter`, `Pages`. Each one declares its own `@*_opts` keyword list, defines a runtime accessor in the `:dev` branch, and re-declares the same `use NimblePublisher` call with the same options in the `else` branch.

This is the biggest source of duplication in the PR — five copies of the same conditional skeleton. Could it collapse into a single macro (e.g. `use Tuist.Marketing.Content, build: ..., from: ..., parser: ..., ...`) that expands into the dev-or-prod branches behind the scenes? Each call site would then be a one-liner and adding a sixth content type wouldn't mean copy-pasting the pattern again.

server/lib/tuist/application.ex
```diff
@@ -297,8 +297,9 @@ defmodule Tuist.Application do
         {Tuist.API.Pipeline, []},
         {Guardian.DB.Sweeper, [interval: 60 * 60 * 1000]},
         TuistWeb.Telemetry,
+        dev_content_children(),
```
pepicrft: The `List.flatten/1` here is a code smell — it's only needed because `dev_content_children/0` returns a list that has to be spliced into the children list. A cleaner alternative is to keep the children list flat and append the dev children conditionally:

```elixir
children =
  base_children() ++ dev_content_children() ++ [TuistWeb.Endpoint]
```

or insert via `List.insert_at/3` / pattern, without needing `List.flatten` to paper over a nested list. Right now a reader has to notice the `List.flatten` wrapper and then realise one of the entries in the literal list is itself a list — that's not obvious.

server/mix.exs
```diff
@@ -34,6 +34,9 @@ defmodule Tuist.MixProject do
   defp elixirc_paths(:dev), do: ["lib", "credo"]
   defp elixirc_paths(_), do: ["lib"]
 
+  defp compilers(:dev), do: Mix.compilers()
+  defp compilers(_env), do: [:boundary] ++ Mix.compilers()
```
pepicrft, line 38: Skipping the `:boundary` compiler in dev is a meaningful behavioural change — boundary violations now only surface in CI/prod compiles, not locally. Worth a short comment here explaining why (compile speed) so it doesn't read as accidental, and ideally a note in `server/AGENTS.md` so contributors know to run a non-dev compile if they want to validate boundaries before pushing.

server/lib/tuist_web/gettext.ex
```diff
@@ -1,3 +1,7 @@
+all_locales = ~w(ar en es ja ko pl pt ru tr yue_Hant zh_Hans zh_Hant)
+dev_all_locales? = System.get_env("TUIST_DEV_ALL_LOCALES") in ~w(1 true TRUE yes YES)
+allowed_locales = if Mix.env() == :dev and not dev_all_locales?, do: ["en"], else: all_locales
```
pepicrft: The `TUIST_DEV_ALL_LOCALES` parsing (`System.get_env(...) in ~w(1 true TRUE yes YES)`) is duplicated verbatim in three places: here, `lib/tuist/cldr.ex`, and `lib/tuist/locale.ex`. Three copies of the same env-flag truthiness check will drift.

Could this move into a single helper, e.g. `Tuist.Environment.dev_all_locales?/0`, that all three call sites use? Same applies to the `Mix.env() == :dev and not dev_all_locales?` gate that's also repeated.

server/lib/tuist/docs.ex
```diff
@@ -5,26 +5,39 @@ defmodule Tuist.Docs do
...
-  end
+    defp static_pages, do: RuntimeStore.pages()
+    defp static_page(slug), do: RuntimeStore.get_page(slug)
+    defp static_slugs, do: RuntimeStore.slugs()
+  else
+    alias Tuist.Docs.Loader
+
+    {pages, source_paths} = Loader.load_pages!()
+
+    for source_path <- source_paths do
```
pepicrft: This `if Mix.env() == :dev do ... else ... end` split inside `Tuist.Docs` is the same shape as the five branches added to the marketing modules — a runtime path that calls into a store, and a prod path that does compile-time embedding. If the marketing duplication gets factored into a macro/helper, it'd be worth pulling docs into the same abstraction so there's one canonical "dev = runtime, prod = compile-time" pattern instead of two implementations of it.

pepicrft (approved):
Solid dev-speed win and the prod path is preserved. Approving — left a batch of inline comments on duplication / naming / code-smell concerns that I think are worth addressing as a follow-up rather than blocking on here.

fortmarek:
thanks for the quick review, yeah, this should have been a draft, I haven't finished, yet 😅 my bad

server/lib/tuist/cldr.ex
fortmarek: Addressed in `67ce04be3e`: `Tuist.Cldr` now just uses `Tuist.Cldr.Configuration`, which wraps/decorates the `use Cldr` call. The `TUIST_DEV_ALL_LOCALES` read moved into compile-time app config and is exposed through `Tuist.Environment.dev_all_locales?/0` / `dev_single_locale?/0`, so the top-level env bindings are gone.

server/lib/tuist/locale.ex
fortmarek: Addressed in `67ce04be3e`: the locale switch is now centralized through compile-time config plus `Tuist.Environment.dev_all_locales?/0` and `Tuist.Environment.dev_single_locale?/0`. `Tuist.Locale`, Gettext, and Cldr all use those helpers instead of repeating the env parsing.

server/lib/tuist/marketing/blog.ex
fortmarek: Addressed in `67ce04be3e`: the repeated dev/prod NimblePublisher branch is now `use Tuist.Marketing.NimblePublisher.Content`. `Blog`, `Changelog`, `Customers`, `Newsletter`, and `Pages` each declare only their content-specific options and call `content_entries/0`.

server/lib/tuist/marketing/runtime_loader.ex
fortmarek: Addressed in `67ce04be3e`: the runtime NimblePublisher pieces now live under `Tuist.Marketing.NimblePublisher.*` and `Tuist.Docs.NimblePublisher.Cache`. The old `RuntimeLoader` / `RuntimeStore` names are gone, with `Builder` and `Cache` making the NimblePublisher lineage explicit.

server/lib/tuist/application.ex
fortmarek: Addressed in `67ce04be3e`: `get_children/0` now builds a flat base list and appends `dev_content_children()` plus `[TuistWeb.Endpoint]`, so `List.flatten/1` is gone.

server/mix.exs
fortmarek: Addressed in `67ce04be3e`: added a short comment next to `compilers(:dev)` explaining the compile-speed tradeoff, and added a note in `server/AGENTS.md` telling contributors to run a non-dev compile when they want local Boundary validation.

server/lib/tuist_web/gettext.ex
fortmarek: Addressed in `67ce04be3e`: the env flag parsing now lives in compile-time config and `Tuist.Environment.dev_all_locales?/0` / `dev_single_locale?/0`. Gettext, Cldr, and `Tuist.Locale` all use the helpers, so the repeated `System.get_env(...) in ~w(...)` checks are gone.

server/lib/tuist/docs.ex
fortmarek: Addressed in `67ce04be3e`: docs now use `use Tuist.Docs.Content`, which owns the dev-runtime vs non-dev compile-time split. The dev path delegates to `Tuist.Docs.NimblePublisher.Cache`, so the docs pattern is no longer hand-rolled inside `Tuist.Docs`.

fortmarek:
Replying to the two top-level review comments:

- Watcher/cache duplication: addressed in `67ce04be3e` by replacing the two watcher modules with `Tuist.ContentFileWatcher`, parameterized by `dirs`, `extensions`, and `cache`, and replacing the duplicate runtime stores with `Tuist.ContentCache`. The docs and marketing caches now wrap that shared cache as `Tuist.Docs.NimblePublisher.Cache` and `Tuist.Marketing.NimblePublisher.Cache`.
- Top-level env reads / repeated `TUIST_DEV_ALL_LOCALES` parsing: addressed in `67ce04be3e` by moving the env parsing into compile-time app config and exposing it through `Tuist.Environment.dev_all_locales?/0` / `dev_single_locale?/0`. `Tuist.Cldr` and `TuistWeb.Gettext` now use small configuration macros, and `Tuist.Locale` selects via a module attribute inside the module.

I also replied directly on each inline review thread with the specific change that addressed it.

PR #10551, opened 2026-04-29
chore(ci): bump Xcode to 26.4.1

pepicrft:
## Summary
- Update `.xcode-version` from 26.2 to 26.4.1 so CI runs use the latest Xcode.
- `.xcode-version-releases` is intentionally left at 26.0 (release builds pin separately).

## Test plan
- [ ] CI green across the workflows that read `.xcode-version` (cli, app, server-production-deployment).


.github/workflows/xcode-processor.yml
```diff
@@ -31,7 +31,7 @@ env:
 jobs:
   swift-build:
     name: Swift Build (XCResultNIF)
-    runs-on: namespace-profile-default-macos
+    runs-on: namespace-profile-default-macos-xcode-24
```
fortmarek: why are we adding a new profile? can't we just update/use the existing one?

.xcode-version
fortmarek, line 1: any reason why not to use the latest 26.4.1 version?

fortmarek (approved):
aligned with the bump, two comments 👇

.github/workflows/xcode-processor.yml
pepicrft: I refrained from updating the existing one because that'd cause the current builds to fail (i.e. the xcode version is bound to the image).

.xcode-version
pepicrft: Not really. I'll bump it.

PR #10554, opened 2026-04-30
fix(cli): intersect linkable dep destinations for orphan local SPM tests

### LINKED ISSUE #10536: Local SPM package test target with .macro dependency triggers fatal platform mismatch lint after #10268
### What happened?

Since Tuist **4.182.0** (PR #10268 — "feat(cli): preserve test targets for local SPM packages"), a local SPM package consumed via `.package(path:)` that contains all of:

1. `platforms: [.iOS, .macOS]`
2. A regular `.target` library
3. A `.macro` target
4. A `.testTarget` whose dependencies include both the library and the macro target

…will fail `tuist generate` with a fatal `GraphLinter` error when the workspace's app is iOS-only:

```
Target MyLibTests which depends on MyLib does not support the required platforms: macos.
The dependency on MyLib must have a dependency condition constraining to at most: ios.
```

The same project structure passed cleanly on Tuist **4.181.0** because the test target was tree-shaken as an orphan external target. PR #10268 intentionally preserves local SPM test targets in the graph (so they can be run via Tuist), but the platform alignment that should accompany this preservation is incomplete, and the existing `lintLinkableDependencies` rule then catches the resulting mismatch.

### Root cause (analysis)

`PruneOrphanExternalTargetsGraphMapper.swift` (introduced in #10268) skips pruning for targets tagged `localSwiftPackageTest`:

```swift
if target.metadata.tags.contains(TargetTags.localSwiftPackageTest) {
    return (target.name, target)
}
```

`ExternalProjectsPlatformNarrowerGraphMapper.swift` then attempts to inherit the test target's destinations from its in-package dependencies — **but only when the test target's own destinations are nil**:

```swift
if targetFilteredDestinations == nil,
   target.metadata.tags.contains(TargetTags.localSwiftPackageTest) {
    let depDestinations = target.dependencies.compactMap { dep -> Set<Destination>? in
        guard case let .target(name, _, _) = dep,
              let depTarget = project.targets[name]
        else { return nil }
        ...
    }
    if let merged = depDestinations.first {
        targetFilteredDestinations = depDestinations.dropFirst().reduce(merged) { $0.union($1) }
    }
}
```

When the test target depends on a `.macro` target in the same package:

- The `.target` library is narrowed to `[iOS]` by the iOS app's traversal.
- The `.macro` target retains `[macOS]` as its destination (host-only by nature).
- The test target's destinations get inherited as the **union** of its in-package deps → `[iOS, macOS]`.
- The library dep is now `[iOS]` while the test target requires `[iOS, macOS]`.
- `GraphLinter.lintLinkableDependencies` computes `requiredPlatforms - dependencySupportedPlatforms = [macOS]` and emits a fatal error.

The user has no way to add a dependency condition to the `.testTarget` in `Package.swift` that would actually make sense (the only working escape is to drop the library dep from the test target, drop the macro dep, drop `.macOS` from `platforms`, or remove the test target). Each is a workaround for a graph that the user did not author.

Suggested upstream fix candidates:

- Always intersect a `localSwiftPackageTest` target's destinations with the union of its in-package linkable dependency destinations (drop the `targetFilteredDestinations == nil` guard).
- Or, when computing `requiredPlatforms` in `lintLinkableDependencies`, restrict it to platforms that all in-package linkable deps actually support (treating in-package macro/host deps as not contributing to deployment platforms).
- Or, do not lint linkable-dependency platform alignment for synthesized SPM test targets (fall back to SwiftPM's own resolution).

### Expected behavior

A local SPM package consumed via `.package(path:)` whose internal structure is valid SwiftPM (the `.testTarget` builds and runs fine via `swift test`) should not produce a fatal Tuist lint error solely because Tuist now preserves the test target in the workspace graph.

### How do we reproduce it?

Minimal reproduction (≈ 15 files, no extra Tuist plugins).

```
repro/
├── Tuist.swift
├── Workspace.swift
├── Tuist/
│   └── Package.swift            # .package(path: "../LocalLib")
├── App/
│   ├── Project.swift            # iOS-only app, .external(name: "MyLib")
│   └── Sources/AppMain.swift
└── LocalLib/
    ├── Package.swift            # platforms: [.iOS, .macOS], .macro + .target + .testTarget
    ├── Sources/
    │   ├── MyLib/MyLib.swift
    │   └── MyLibMacros/MyLibMacros.swift
    └── Tests/MyLibTests/MyLibTests.swift
```

`LocalLib/Package.swift`:

```swift
// swift-tools-version: 5.9
import CompilerPluginSupport
import PackageDescription

let package = Package(
    name: "MyLib",
    platforms: [.iOS(.v16), .macOS(.v13)],
    products: [.library(name: "MyLib", targets: ["MyLib"])],
    dependencies: [
        .package(url: "https://github.com/swiftlang/swift-syntax.git", "601.0.0"..<"602.0.0"),
    ],
    targets: [
        .macro(
            name: "MyLibMacros",
            dependencies: [
                .product(name: "SwiftSyntaxMacros", package: "swift-syntax"),
                .product(name: "SwiftCompilerPlugin", package: "swift-syntax"),
            ]
        ),
        .target(name: "MyLib", dependencies: ["MyLibMacros"]),
        .testTarget(name: "MyLibTests", dependencies: ["MyLib", "MyLibMacros"]),
    ]
)
```

`Tuist/Package.swift`:

```swift
// swift-tools-version: 5.9
import PackageDescription
#if TUIST
import ProjectDescription
let packageSettings = PackageSettings()
#endif

let package = Package(
    name: "ExternalDeps",
    dependencies: [.package(path: "../LocalLib")]
)
```

`App/Project.swift`:

```swift
import ProjectDescription

let project = Project(
    name: "App",
    targets: [
        .target(
            name: "App",
            destinations: .iOS,
            product: .app,
            bundleId: "io.repro.app",
            deploymentTargets: .iOS("16.0"),
            sources: ["Sources/**"],
            dependencies: [.external(name: "MyLib")]
        ),
    ]
)
```

`Workspace.swift`:

```swift
import ProjectDescription
let workspace = Workspace(name: "ReproWorkspace", projects: ["App"])
```

`Tuist.swift`:

```swift
import ProjectDescription

let tuist = Tuist(
    project: .tuist(
        compatibleXcodeVersions: .all,
        generationOptions: .options()
    )
)
```

### Steps

```sh
# 4.181.0 — passes (test target is tree-shaken)
mise install tuist@4.181.0
~/.local/share/mise/installs/tuist/4.181.0/tuist install
~/.local/share/mise/installs/tuist/4.181.0/tuist generate --no-open

# 4.182.0 — fails with Fatal linting issues found
rm -rf Tuist/.build .build *.xcodeproj */*.xcodeproj *.xcworkspace
mise install tuist@4.182.0
~/.local/share/mise/installs/tuist/4.182.0/tuist install
~/.local/share/mise/installs/tuist/4.182.0/tuist generate --no-open

# 4.187.0 — same failure
~/.local/share/mise/installs/tuist/4.187.0/tuist install
~/.local/share/mise/installs/tuist/4.187.0/tuist generate --no-open
```

Notes:

- `tuist graph` does **not** run `lintLinkableDependencies` and reports success. The failure is only visible via `tuist generate` (or any path that runs the full graph linter).
- Reducing the test target's deps to just the macro dep, or removing the test target, makes the lint pass.
- Replacing `.package(path:)` with a remote `.package(url:)` also makes the lint pass (PR #10268 only preserves local SPM test targets).

### Error log

```sh
$ tuist generate --no-open
Using cache binaries for the following targets:
Target MyLibTests which depends on MyLib does not support the required platforms: macos. The dependency on MyLib must have a dependency condition constraining to at most: ios.
✖ Error
  Fatal linting issues found

  Sorry this didn't work. Here's what to try next:
   ▸ If the error is not actionable, let's discuss it in the <Troubleshooting & how to: https://community.tuist.dev/c/troubleshooting-how-to/6>
   ▸ If you are very certain it's a bug, <file an issue: https://github.com/tuist/tuist>
```

Bisect summary (clean reproduction project, identical sources):

| Tuist | Result |
| --- | --- |
| 4.181.0 | `tuist generate` succeeds. `MyLibTests` is pruned as orphan external target. |
| 4.182.0 | `tuist generate` fails with the fatal lint above. `MyLibTests` is preserved with `[iOS, macOS]` destinations while `MyLib` is narrowed to `[iOS]`. |
| 4.187.0 | Same failure as 4.182.0. |

The 4.181.0 → 4.182.0 boundary matches commit `3e50c1cb48` ("feat(cli): preserve test targets for local SPM packages", #10268).

### macOS version

26.2

### Tuist version

4.187.0 (also reproduces on 4.182.0–4.186.x; passes on 4.181.0)

### Xcode version

26.2

mqzkim:
Resolves https://github.com/tuist/tuist/issues/10536

## Summary

After #10268, local SPM test targets are preserved in the graph as orphans. `ExternalProjectsPlatformNarrowerGraphMapper`'s fallback derived their destinations as the *union* of their direct deps' destinations, but linking is an AND constraint — a test target can only run on platforms supported by every linked dep. The union also pulled `[.mac]` in from `.macro` deps (which `GraphTraverser` pins to host-only), producing fatal `lintLinkableDependencies` errors against iOS-only library deps:

> Target MyLibTests which depends on MyLib does not support the required platforms: macos. The dependency on MyLib must have a dependency condition constraining to at most: ios.

This switches the fallback to **intersect** destinations of the test's *linkable* deps (`Target.isLinkable()` — frameworks and libraries). Non-linkable deps (macros, bundles, plugins) don't participate in linking and don't constrain runtime platforms, so they're excluded naturally — no hard-coded product check. Approach validated in the [issue thread](https://github.com/tuist/tuist/issues/10536#issuecomment-4345584906) before opening this PR.

Unit coverage in `ExternalProjectsPlatformNarrowerGraphMapperTests` covers the main bug shape (test depending on library + macro), the macro-only edge case (declared destinations preserved), and incompatible linkable deps (empty intersection → `tuist:prunable`).

### How to test locally

Minimal repro attached: [issue-10536-repro.zip](https://github.com/user-attachments/files/27236535/issue-10536-repro.zip). On `main`, `tuist generate` fails with the lint above. With this patch, `tuist generate` succeeds and `xcodebuild build -scheme App -destination "platform=iOS Simulator,name=iPhone 17 Pro"` completes.

### DIFF
--- a/cli/Sources/TuistDependencies/Mappers/ExternalProjectsPlatformNarrowerGraphMapper.swift
+++ b/cli/Sources/TuistDependencies/Mappers/ExternalProjectsPlatformNarrowerGraphMapper.swift
@@ -69,12 +69,18 @@ public struct ExternalProjectsPlatformNarrowerGraphMapper: GraphMapping { // swi
            target.metadata.tags.contains(TargetTags.localSwiftPackageTest)
         {
             let linkableDestinations = target.dependencies.compactMap { dep -> Set<Destination>? in
-                guard case let .target(name, _, _) = dep,
+                guard case let .target(name, _, dependencyCondition) = dep,
                       let depTarget = project.targets[name],
                       depTarget.isLinkable()
                 else { return nil }
                 let depGraphTarget = GraphTarget(path: project.path, target: depTarget, project: project)
-                return externalTargetSupportedDestinations[depGraphTarget]
+                guard let depDestinations = externalTargetSupportedDestinations[depGraphTarget] else { return nil }
+
+                return orphanTestDependencyDestinations(
+                    depDestinations,
+                    target: target,
+                    dependencyCondition: dependencyCondition
+                )
             }
             if let first = linkableDestinations.first {
                 targetFilteredDestinations = linkableDestinations.dropFirst().reduce(first) { $0.union($1) }
@@ -97,5 +103,28 @@ public struct ExternalProjectsPlatformNarrowerGraphMapper: GraphMapping { // swi
             )
         }
         return target
+    }
+
+    private func orphanTestDependencyDestinations(
+        _ destinations: Set<Destination>,
+        target: Target,
+        dependencyCondition: PlatformCondition?
+    ) -> Set<Destination>? {
+        let inheritedDestinations = destinations.intersection(target.destinations)
+
+        guard let dependencyCondition,
+              let targetCondition = PlatformCondition.when(target.dependencyPlatformFilters)
+        else {
+            return inheritedDestinations
+        }
+
+        switch targetCondition.intersection(dependencyCondition) {
+        case .incompatible:
+            return nil
+        case let .condition(condition):
+            guard let condition else { return inheritedDestinations }
+            let allowedPlatformFilters = condition.platformFilters
+            return inheritedDestinations.filter { allowedPlatformFilters.contains($0.platformFilter) }
+        }
     }
 }
--- a/cli/Sources/TuistDependencies/Mappers/ExternalProjectsPlatformNarrowerGraphMapper.swift
+++ b/cli/Sources/TuistDependencies/Mappers/ExternalProjectsPlatformNarrowerGraphMapper.swift
@@ -62,20 +62,22 @@ public struct ExternalProjectsPlatformNarrowerGraphMapper: GraphMapping { // swi
 
         var targetFilteredDestinations = externalTargetSupportedDestinations[graphTarget]
 
-        // Local SPM test targets are orphans (nothing depends on them), so the top-down
-        // traversal never reaches them. Inherit destinations from their direct dependencies instead.
+        // Orphan local SPM test targets aren't reached by the top-down traversal. Intersect
+        // destinations of the test's linkable deps — linking is an AND constraint, and
+        // non-linkable deps (macros, bundles) don't constrain runtime platforms.
         if targetFilteredDestinations == nil,
            target.metadata.tags.contains(TargetTags.localSwiftPackageTest)
         {
-            let depDestinations = target.dependencies.compactMap { dep -> Set<Destination>? in
+            let linkableDestinations = target.dependencies.compactMap { dep -> Set<Destination>? in
                 guard case let .target(name, _, _) = dep,
-                      let depTarget = project.targets[name]
+                      let depTarget = project.targets[name],
+                      depTarget.isLinkable()
                 else { return nil }
                 let depGraphTarget = GraphTarget(path: project.path, target: depTarget, project: project)
                 return externalTargetSupportedDestinations[depGraphTarget]
             }
-            if let merged = depDestinations.first {
-                targetFilteredDestinations = depDestinations.dropFirst().reduce(merged) { $0.union($1) }
+            if let first = linkableDestinations.first {
+                targetFilteredDestinations = linkableDestinations.dropFirst().reduce(first) { $0.intersection($1) }
             }
         }
 
--- a/cli/Sources/TuistDependencies/Mappers/ExternalProjectsPlatformNarrowerGraphMapper.swift
+++ b/cli/Sources/TuistDependencies/Mappers/ExternalProjectsPlatformNarrowerGraphMapper.swift
@@ -62,9 +62,9 @@ public struct ExternalProjectsPlatformNarrowerGraphMapper: GraphMapping { // swi
 
         var targetFilteredDestinations = externalTargetSupportedDestinations[graphTarget]
 
-        // Orphan local SPM test targets aren't reached by the top-down traversal. Intersect
-        // destinations of the test's linkable deps — linking is an AND constraint, and
-        // non-linkable deps (macros, bundles) don't constrain runtime platforms.
+        // Orphan local SPM test targets aren't reached by the top-down traversal. Union
+        // destinations of the test's linkable deps — non-linkable deps (macros, bundles)
+        // don't constrain runtime platforms.
         if targetFilteredDestinations == nil,
            target.metadata.tags.contains(TargetTags.localSwiftPackageTest)
         {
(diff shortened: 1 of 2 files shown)


cli/Sources/TuistDependencies/Mappers/ExternalProjectsPlatformNarrowerGraphMapper.swift
```diff
@@ -62,20 +62,22 @@ public struct ExternalProjectsPlatformNarrowerGraphMapper: GraphMapping { // swi
...
+                      let depTarget = project.targets[name],
+                      depTarget.isLinkable()
                 else { return nil }
                 let depGraphTarget = GraphTarget(path: project.path, target: depTarget, project: project)
                 return externalTargetSupportedDestinations[depGraphTarget]
             }
-            if let merged = depDestinations.first {
-                targetFilteredDestinations = depDestinations.dropFirst().reduce(merged) { $0.union($1) }
+            if let first = linkableDestinations.first {
+                targetFilteredDestinations = linkableDestinations.dropFirst().reduce(first) { $0.intersection($1) }
```
pepicrft: I think this intersects conditional test deps too aggressively. SwiftPM target dependencies can carry platform conditions, and local package test targets preserve those conditions. Here we intersect every linkable dependency's inherited destinations without first applying each dependency's own condition, so a test target that links `LibA` only on iOS and `LibB` only on macOS gets collapsed to an empty destination set and marked prunable even though it should still be runnable on both platforms. The new tests only cover unconditional dependencies, so this regression would slip through.

mqzkim: @pepicrft You're right — I misread the conditional case. Treating it as an unconditional intersection collapses platforms that should remain valid (iOS-only LibA + macOS-only LibB → test should still run on both, not get pruned).
I'll update the mapper to honor each dep's PlatformCondition per-destination (a dep gated to iOS doesn't constrain the macOS column, and vice versa), revise the existing test that asserted the wrong expectation, and add coverage for the conditional case. Pushing a follow-up commit shortly.

mqzkim: Pushed a follow-up addressing both concerns:

- `75fca082c` switches the fold from `intersection` to `union`. Each linkable dep's traversed destinations already reflect the platform conditions applied upstream by `GraphTraverser`, so unioning resolves a test linking `LibA` on iOS and `LibB` on macOS to `{.iPad, .iPhone, .mac}` instead of pruning. Updated the existing `_has_disjoint_linkable_deps` test to assert the new expectation.
- `d6a500330` adds `_depends_on_platform_conditional_libraries`, which wires `PlatformCondition.when([.ios])` / `.when([.macos])` onto the test target's edges to lock in the regression case you flagged.

Let me know if I missed anything.

PR #10562, opened 2026-04-30
fix(cli): make shared cache and state writes safe across concurrent processes

### LINKED ISSUE #9588: Race Condition Failures in `tuist generate` Across Worktrees
### What happened?

One of my repos uses Tuist and has a lot of agents running concurrently. I've noticed flakiness in the `tuist generate` command, with 2 types of error messages:

1. 
```
JIT session error: Symbols not found: [
  _$s25ProjectDescriptionHelpers11CFWorkspaceV10tuistVal
  ue0aB09WorkspaceVvg,
  _$s25ProjectDescriptionHelpers11CFWorkspaceV4monoACvau
]
```
2.
```
File already exists: "Can't rename                    
  '/Users/connorneville/.local/state/tuist/.tmp-hPkYAJ'
  to '/Users/connorneville/.local/state/tuist/recent-pat
  hs.json' as it already exists." ('renamex_np' system  
  call failed with '(17) File exists'.)
```

Both of these are transient and succeed once there is no race, but they happen frequently enough with enough concurrent activity.

### How do we reproduce it?

On any Tuist-supported project, use this script which attempts `tuist generate` across 2 worktrees. 
```
#!/bin/bash
set -uo pipefail

REPO_ROOT="$(cd "$(dirname "$0")/.." && pwd)"
WT1="$REPO_ROOT/.claude/worktrees/jit-race-1"
WT2="$REPO_ROOT/.claude/worktrees/jit-race-2"

if [ ! -d "$WT1" ] || [ ! -d "$WT2" ]; then
  echo "Error: worktrees not found. Create them first:"
  echo "  git worktree add $WT1 -b jit-race-test-1 HEAD"
  echo "  git worktree add $WT2 -b jit-race-test-2 HEAD"
  exit 1
fi

echo "Clearing Tuist ProjectDescriptionHelpers cache..."
rm -rf ~/.cache/tuist/ProjectDescriptionHelpers/

echo "Starting concurrent generate-workspace in two worktrees..."
echo "  WT1: $WT1"
echo "  WT2: $WT2"
echo ""

(cd "$WT1" && .tuist generate --no-open) > /tmp/gen-ws-1.log 2>&1 &
PID1=$!

(cd "$WT2" && tuist generate --no-open) > /tmp/gen-ws-2.log 2>&1 &
PID2=$!

echo "PID 1: $PID1"
echo "PID 2: $PID2"
echo "Waiting for both to finish..."

wait $PID1
EXIT1=$?

wait $PID2
EXIT2=$?

echo ""
echo "=== Results ==="
echo "Process 1 (WT1) exit code: $EXIT1"
echo "Process 2 (WT2) exit code: $EXIT2"

if [ $EXIT1 -ne 0 ]; then
  echo ""
  echo "=== Process 1 output (FAILED) ==="
  cat /tmp/gen-ws-1.log
fi

if [ $EXIT2 -ne 0 ]; then
  echo ""
  echo "=== Process 2 output (FAILED) ==="
  cat /tmp/gen-ws-2.log
fi

if [ $EXIT1 -eq 0 ] && [ $EXIT2 -eq 0 ]; then
  echo "Both succeeded — race not triggered this run."
fi
```

In 10 runs, I received 2 of the `JIT session error`s and 8 of the `recent-paths.json` errors.

### Error log

See above.

### macOS version

26.0

### Tuist version

4.48.4

### Xcode version

26.2

pepicrft:
## Summary

Fixes [#9588](https://github.com/tuist/tuist/issues/9588) — running `tuist` concurrently across worktrees (or any setup that shares `XDG_CACHE_HOME` / `XDG_STATE_HOME` between processes) was hitting two races:

1. **`recent-paths.json`** — failed with `renamex_np ... 'File exists'`, and concurrent read-modify-write cycles silently dropped updates.
2. **`ProjectDescriptionHelpers` cache** — the module cache directory was created before `swiftc` finished writing the dylib, so a sibling process could short-circuit on the existence check at [ProjectDescriptionHelpersBuilder.swift:168](cli/Sources/TuistLoader/ProjectDescriptionHelpers/ProjectDescriptionHelpersBuilder.swift) and load a half-written module — surfacing as `JIT session error: Symbols not found` in the reporter's setup.

The previous workaround was to point each worktree at its own cache via `XDG_CACHE_HOME`, which defeats the cache and (per the issue) caused remote-cache redownloads and a billing spike.

## What changed

- New `FileLock` helper in `TuistSupport` wrapping `flock(2)`. The blocking acquire runs on a background queue so it does not stall the cooperative thread pool. The C `flock` symbol is resolved via `dlsym` to disambiguate from the `struct flock` exported by Darwin.
- `RecentPathsStore.remember` / `read` now serialize under an exclusive `FileLock` on `recent-paths.json.lock`, and writes go through Foundation's atomic `Data.write(to:options:[.atomic])` (plain rename, no `RENAME_EXCL`).
- `ProjectDescriptionHelpersBuilder` compiles into a sibling staging directory `<hash>.tmp.<pid>.<uuid>` and atomically publishes it to `<hash>` via `rename(2)`. If another process won the race, rename returns `ENOTEMPTY`/`EEXIST` — the loser removes its staging directory and falls back to the published cache. Sibling processes therefore only ever see a fully populated module directory.
- Reproduction tests for both races, plus the existing helpers integration tests still pass.

## Test plan

- [x] `RecentPathsStoreTests/remember_concurrent_writes_from_separate_stores` — failed 100% before, passes deterministically after.
- [x] `ProjectDescriptionHelpersBuilderIntegrationTests/test_build_is_atomic_across_concurrent_builders_sharing_cache` — failed ~75% before, passes 5/5 after.
- [x] Existing `ProjectDescriptionHelpersBuilderIntegrationTests` (`test_build_when_the_helpers_is_a_dylib`, `test_build_when_the_helpers_is_a_plugin`) still pass.
- [x] Existing `RecentPathsStoreTests/remember` still passes.


cli/Sources/TuistLoader/ProjectDescriptionHelpers/ProjectDescriptionHelpersBuilder.swift
```diff
@@ -169,22 +177,66 @@ public final class ProjectDescriptionHelpersBuilder: ProjectDescriptionHelpersBu
...
+                    if renameErrno != ENOTEMPTY, renameErrno != EEXIST {
+                        throw NSError(
+                            domain: NSPOSIXErrorDomain,
+                            code: Int(renameErrno),
+                            userInfo: [
+                                NSLocalizedDescriptionKey:
+                                    "Failed to publish ProjectDescriptionHelpers cache to \(moduleCacheDirectory.pathString)",
+                            ]
+                        )
+                    }
```
fortmarek: any reason why we're no returning our own error instead of `NSError`?

cli/Sources/TuistLoader/ProjectDescriptionHelpers/ProjectDescriptionHelpersBuilder.swift
```diff
@@ -169,22 +177,66 @@ public final class ProjectDescriptionHelpersBuilder: ProjectDescriptionHelpersBu
...
                 let timer = clock.startTimer()
-                try System.shared.runAndPrint(command, verbose: false, environment: Environment.current.manifestLoadingVariables)
+                do {
+                    try System.shared.runAndPrint(
+                        command,
+                        verbose: false,
+                        environment: Environment.current.manifestLoadingVariables
+                    )
+                } catch {
+                    try? await fileSystem.remove(stagingDirectory)
```
fortmarek: can this be in `defer` instead?

cli/Sources/TuistLoader/ProjectDescriptionHelpers/ProjectDescriptionHelpersBuilder.swift
```diff
@@ -169,22 +177,66 @@ public final class ProjectDescriptionHelpersBuilder: ProjectDescriptionHelpersBu
...
+                // between two directories on the same volume is atomic, and
+                // fails with `ENOTEMPTY` (or `EEXIST` on some platforms) when
+                // the destination directory exists and is non-empty — i.e. the
+                // "another process won the race" case. We treat that as
+                // success and remove our staging directory.
+                let renameResult = stagingDirectory.pathString.withCString { stagingCString in
+                    moduleCacheDirectory.pathString.withCString { destCString in
+                        rename(stagingCString, destCString)
+                    }
+                }
```
fortmarek: any reason why we can't we use `FileSystem` operation instead?

cli/Sources/TuistSupport/Utils/RecentPathsStore.swift
fortmarek, line 1: are we still using this for anything since we removed the MCP integration for generated projects?

cli/Sources/TuistSupport/Utils/POSIXDirectory.swift
```diff
@@ -0,0 +1,27 @@
...
+#endif
+
+import Foundation
+import Path
+
+public enum POSIXDirectory {
+    /// Idempotently ensures a directory exists at `path`. Equivalent to
+    /// `mkdir -p` for the leaf component: a single `mkdir(2)` call that
+    /// treats `EEXIST` as success — safe under concurrent processes
+    /// racing to create the same directory.
```
fortmarek: should we consider building this into the `FileSystem`?

cli/Sources/TuistSupport/Utils/FileLock.swift
fortmarek, line 1: we already have file locking for the credentials. Can we reuse the code between the two implementations?

fortmarek (approved):

cli/Sources/TuistSupport/Utils/RecentPathsStore.swift
pepicrft: Not at all. I'll remove it.

cli/Sources/TuistLoader/ProjectDescriptionHelpers/ProjectDescriptionHelpersBuilder.swift
pepicrft: Replaced the inline NSError with a typed FatalError enum (ProjectDescriptionHelpersBuilderError) in the latest push.

cli/Sources/TuistLoader/ProjectDescriptionHelpers/ProjectDescriptionHelpersBuilder.swift
pepicrft: Done. Cleanup is now in a defer keyed off a stagingDirectoryWasPublished flag, so the swiftc-failure and rename-failure paths share the same one-line cleanup.

cli/Sources/TuistLoader/ProjectDescriptionHelpers/ProjectDescriptionHelpersBuilder.swift
pepicrft: Switched to fileSystem.move. If the move throws but the destination now exists, we treat it as 'another process won the race' and fall back to the published directory.

cli/Sources/TuistSupport/Utils/RecentPathsStore.swift
pepicrft: No — RecentPathsStore was only there to feed the MCP recent-projects integration, which was already removed. I deleted the store, the RecentPathRememberableCommand protocol, the testing trait, the mock, and the conformances on every command.

cli/Sources/TuistSupport/Utils/POSIXDirectory.swift
pepicrft: Removed POSIXDirectory.swift entirely. fileSystem.makeDirectory already creates intermediates by default, so we don't need the helper anymore. If we hit this pattern again, upstreaming an idempotent makeDirectory variant into FileSystem sounds right.

cli/Sources/TuistSupport/Utils/FileLock.swift
pepicrft: Removed FileLock.swift. With RecentPathsStore gone, we don't need a lock — the helpers builder uses atomic publish via fileSystem.move instead. Any future locking should reuse TSCBasic.FileLock to match CachedValueStore and TuistMachineMetrics.

PR #10564, opened 2026-04-30
perf(server): cut ClickHouse CPU from active test cases analytics chart

fortmarek:
## Summary

The top ClickHouse CPU consumer was `SELECT uniqExact(test_case_id) FROM test_case_runs` issued by the Test Cases analytics chart — production telemetry showed 202 executions/hour, ~535 s of cumulative CPU, scanning ~30 B rows / 1.07 TB. The chart evaluates a 14-day rolling window at every bucket endpoint; with `test_case_runs` ordered by `(project_id, test_case_id, ran_at, id)`, the `ran_at` predicate forces a full per-project scan.

This PR pre-aggregates the answer:

- **New MV `test_case_runs_active_daily_stats`** (AggregatingMergeTree) stores `uniqExactState(test_case_id)` keyed by `(project_id, date, is_ci)`. Each chart bucket now merges ≤28 daily states (`14 days × 2 is_ci`) via `uniqExactMerge` instead of scanning raw runs.
- **`active_test_cases_count/4`** is rewritten to read the MV. The window becomes `window_days` calendar days inclusive of the endpoint's date, which matches the original behavior at every end-of-day bucket endpoint and the existing tests.
- **`test_cases` ILIKE searches** (Test Cases / Flaky / Quarantined listing pages) get `ngrambf_v1(3, 65536, 3, 0)` data-skipping indexes on `name`, `module_name`, and `suite_name`. Production traces showed 100–200 M rows scanned to return < 20 results; the bloom filter prunes granules where no 3-gram of the search term can exist. Same shape as the precedent on `xcode_targets`.

## Migration notes

- `20260430120000_create_test_case_runs_active_daily_stats_mv.exs` follows the partition-by-partition backfill pattern from `recreate_flaky_test_case_runs_mv` to avoid the ZooKeeper "table is shutting down" race on ClickHouse Cloud during a `POPULATE`. `@disable_ddl_transaction true` and `@disable_migration_lock true`. Backfill happens before the MV trigger is attached, mirroring the existing pattern.
- `20260430120001_add_ngram_indexes_to_test_cases.exs` uses `ADD INDEX IF NOT EXISTS` for idempotency and triggers `MATERIALIZE INDEX` to backfill index data for existing parts (asynchronous mutations — the migration returns once they're queued).
- Semantic note: the MV-backed window is calendar-day-inclusive of the endpoint's date. For end-of-day chart endpoints (the dominant case) this matches the prior second-precise window exactly; for hourly endpoints it collapses to the same value within a calendar day, which is consistent with how the chart is consumed.

## Test plan

- [x] Local ClickHouse: backfill matches raw query (`uniqExact == uniqExactMerge`, `is_ci`-split sums consistent with union via deduplication).
- [x] Verified MV trigger fires on new inserts.
- [x] `mix test test/tuist/tests/analytics_test.exs` — 49 tests, 0 failures.
- [x] `mix test test/tuist/tests_test.exs` — 198 tests, 0 failures.
- [ ] Confirm CPU drop on staging dashboard after the migration backfill completes.

### DIFF
--- a/server/lib/tuist/tests/analytics.ex
+++ b/server/lib/tuist/tests/analytics.ex
@@ -358,9 +358,13 @@ defmodule Tuist.Tests.Analytics do
     window_days = Tests.active_window_days()
 
     previous_endpoint = DateTime.add(start_datetime, -1, :second)
-    previous_count = active_test_cases_count(project_id, previous_endpoint, window_days, is_ci)
+    previous_count = active_test_cases_count(project_id, previous_endpoint, window_days, is_ci, date_period)
 
-    values = Enum.map(date_endpoints, &active_test_cases_count(project_id, &1, window_days, is_ci))
+    values =
+      Enum.map(
+        date_endpoints,
+        &active_test_cases_count(project_id, &1, window_days, is_ci, date_period)
+      )
 
     current_count = List.last(values) || 0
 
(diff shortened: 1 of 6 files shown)


server/lib/tuist/tests/analytics.ex
```diff
@@ -371,21 +372,21 @@ defmodule Tuist.Tests.Analytics do
...
-      where: tcr.ran_at <= ^endpoint,
-      where: tcr.inserted_at >= ^window_start,
-      select: fragment("uniqExact(?)", tcr.test_case_id)
+  # Reads `test_case_runs_active_daily_stats`, an AggregatingMergeTree MV that
+  # pre-computes `uniqExactState(test_case_id)` per (project_id, date, is_ci).
+  # Each call merges at most ~28 daily states (14 days × 2 is_ci) via
+  # `uniqExactMerge` instead of scanning `test_case_runs` for every chart
+  # bucket. The window is `window_days` calendar days inclusive of
+  # `endpoint`'s date.
+  defp active_test_cases_count(project_id, endpoint, window_days, is_ci) do
```
pepicrft: This changes the semantics for the `last-24-hours` preset. `active_test_cases_count/4` now truncates `endpoint` to a `Date`, so every hourly bucket within the same UTC day returns the same merged state and can count runs that happened later that day. We need either an hourly-safe aggregate or to keep the raw timestamp query for `:hour` ranges.

server/priv/ingest_repo/migrations/20260430120000_create_test_case_runs_active_daily_stats_mv.exs
```diff
@@ -0,0 +1,112 @@
...
+    CREATE TABLE IF NOT EXISTS test_case_runs_active_daily_stats (
+      project_id Int64,
+      date Date,
+      is_ci Bool,
+      test_case_ids_state AggregateFunction(uniqExact, UUID)
+    ) ENGINE = AggregatingMergeTree
+    ORDER BY (project_id, date, is_ci)
+    """)
+
+    backfill_by_partition()
```
pepicrft: This backfills before creating the MV trigger. On a live system, any `test_case_runs` rows inserted while the partition loop is running, or after a partition has already been processed, will never reach `test_case_runs_active_daily_stats` because there is no catch-up pass after `CREATE MATERIALIZED VIEW`. That leaves a permanent gap in the chart data.

pepicrft (approved):
Direction looks good overall. I left two inline comments that should be addressed before merging.

PR #10566, opened 2026-04-30
fix(cli): fix macro copy phase output collision on macOS consumer targets

freak4pc:
On a macOS native build of a target that consumes a Swift macro, Xcode emits warnings such as:

```
unexpected mutating task ('PhaseScriptExecution Copy\ Swift\ Macro\ executable\ into\ \$BUILT_PRODUCT_DIR …/Database.build/Script-….sh') with no relation to prior mutator ('Ld …/Build/Products/Debug/DatabaseMacros normal')

unexpected mutating task ('CodeSign …/Build/Products/Debug/DatabaseMacros') with no relation to prior mutator ('PhaseScriptExecution Copy\ Swift\ Macro\ executable\ into\ \$BUILT_PRODUCT_DIR …')
```

**Root cause:** the `Copy Swift Macro executable into $BUILT_PRODUCT_DIR` script declared `outputPaths` that collapse to `$BUILD_DIR/Debug/<exe>` on macOS native builds (empty `$EFFECTIVE_PLATFORM_NAME`) — the same path the macro target's `Ld` step writes. Xcode sees two unrelated mutators (script in consumer target, link in macro target) on one path with no declared dep edge and flags it. The script's `if src != dest` guard (added in #9995) keeps the on-disk artifact correct, but the declared output is misleading.

This PR splits phase generation by consumer destination:

- **macOS-exclusive consumer** → skip the phase entirely. The script body is a no-op (`src == dest` once the platform suffix collapses) and the macro target's `Ld` step has already placed the file at the expected path.
- **Mixed-destination consumer that supports macOS** (e.g. `destinations: [.iPhone, .mac]`) → still generate the phase, since it's needed to copy the host-built macro into `Debug-<platform>/<exe>` for non-macOS slices. Set `alwaysOutOfDate = true` instead of declaring `outputPaths` so the macOS slice of the same target doesn't trip the collision. Same pattern used elsewhere for always-run scripts whose outputs aren't consumed by the build graph (e.g. Crashlytics dSYM upload).
- **Non-macOS consumer** → unchanged. `$EFFECTIVE_PLATFORM_NAME` is non-empty so the collision doesn't occur.

### How to test locally

1. Generate any project where a target consumes a Swift macro on macOS (e.g. a target that uses `@Observable` via a macro package, or a SwiftSyntax-based macro target).
2. Build the consuming target for macOS in Xcode.
3. Before this change: build log shows `unexpected mutating task … with no relation to prior mutator …` warnings around the `Copy Swift Macro executable …` script and the macro target's `Ld` / `CodeSign` steps.
4. After this change: those warnings are gone, and the macro is still copied into `$BUILT_PRODUCT_DIR` for non-macOS slices.

`cli/Tests/TuistGeneratorTests/Generator/BuildPhaseGeneratorTests.swift` covers all three branches: existing iOS-only test asserts `outputPaths` are declared and `alwaysOutOfDate == false`; a new test asserts no phase is generated for macOS-exclusive consumers; another asserts `alwaysOutOfDate == true` with empty `outputPaths` for `[.iPhone, .mac]` consumers.

### DIFF
--- a/cli/Sources/TuistGenerator/Generator/BuildPhaseGenerator.swift
+++ b/cli/Sources/TuistGenerator/Generator/BuildPhaseGenerator.swift
@@ -511,11 +511,6 @@ struct BuildPhaseGenerator: BuildPhaseGenerating {
     ) throws {
         if directSwiftMacroExecutables.isEmpty { return }
 
-        // macOS-exclusive consumers: script body is a no-op (src == dest once
-        // `$EFFECTIVE_PLATFORM_NAME` is empty) and the declared output would
-        // collide with the macro target's `Ld` step on the same path.
-        if target.isExclusiveTo(.macOS) { return }
-
         let copySwiftMacrosBuildPhase = PBXShellScriptBuildPhase(name: "Copy Swift Macro executable into $BUILT_PRODUCT_DIR")
 
         let executableNames = directSwiftMacroExecutables.compactMap {
@@ -552,10 +547,11 @@ struct BuildPhaseGenerator: BuildPhaseGenerating {
         }
 
         if target.supports(.macOS) {
-            // Reached only for mixed-destination consumers (macOS-exclusive returned
-            // above). The phase is still needed to populate `Debug-<platform>/<exe>`
-            // for non-macOS slices, but declared outputs would collide on this
-            // target's own macOS slice — so omit them.
+            // Any consumer that supports macOS still needs the phase because the
+            // compiler/editor load plugin executables from `Debug...`, even when
+            // the active configuration is not Debug. Declared outputs would collide
+            // on native macOS builds where `$EFFECTIVE_PLATFORM_NAME` is empty, so
+            // omit them and always run the phase instead.
             copySwiftMacrosBuildPhase.alwaysOutOfDate = true
         } else {
             copySwiftMacrosBuildPhase.outputPaths = executableNames.flatMap { executable in
--- a/cli/Sources/TuistGenerator/Generator/BuildPhaseGenerator.swift
+++ b/cli/Sources/TuistGenerator/Generator/BuildPhaseGenerator.swift
@@ -91,6 +91,7 @@ struct BuildPhaseGenerator: BuildPhaseGenerating {
         let directSwiftMacroExecutables = graphTraverser.directSwiftMacroExecutables(path: path, name: target.name).sorted()
         try generateCopySwiftMacroExecutableScriptBuildPhase(
             directSwiftMacroExecutables: directSwiftMacroExecutables,
+            target: target,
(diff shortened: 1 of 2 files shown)


pepicrft:
I think we can narrow this a bit further.

For macOS-only targets, we should be able to skip generating the `Copy Swift Macro executable into $BUILT_PRODUCT_DIR` phase entirely. The macro executable already lands in `Build/Products/Debug/<exe>` on native macOS builds, so there is nothing to relocate and no reason to advertise outputs or force an always-run script.

For mixed-destination targets that also support macOS, `alwaysOutOfDate = true` still makes sense because the phase is needed for non-macOS destinations, but `outputPaths` collide on native macOS builds when `$EFFECTIVE_PLATFORM_NAME` is empty.

That should keep the warning fix while avoiding unconditional script phases for macOS-only macro consumers.

freak4pc:
Good point! 
I added the exclusivity check - but we still need the second condition for mixed targets (i.e. a target with both iOS & mac together). LMK what you think? Thanks! @pepicrft

pepicrft (approved):

fortmarek:
Alternative to `alwaysOutOfDate = true` for the mixed-destination branch: declare a stamp file under `\$(DERIVED_FILE_DIR)` as the script's output, and `touch` it at the end of the script.

```swift
let stamp = "\$(DERIVED_FILE_DIR)/copy-swift-macro.stamp"
copySwiftMacrosBuildPhase.outputPaths = [stamp]
copySwiftMacrosBuildPhase.shellScript = """
\(copyLines.joined(separator: "\n"))
mkdir -p "\$DERIVED_FILE_DIR"
touch "\$DERIVED_FILE_DIR/copy-swift-macro.stamp"
"""
```

Why it's worth considering over `alwaysOutOfDate`:

- `\$(DERIVED_FILE_DIR)` is per-target / per-configuration / per-platform, so the stamp can't collide with the macro target's `Ld` output on the macOS slice, and can't collide across consumer targets.
- Xcode's up-to-date check still works: stamp mtime vs. input mtime. Incremental rebuilds on iOS slices skip the phase when the macro binary hasn't changed, which `alwaysOutOfDate = true` doesn't allow.
- Standard touch-stamp pattern used elsewhere when a script's logical effect isn't the file you want Xcode tracking.

Trade-offs:

- Adds a tiny artifact in the derived files dir per consumer.
- If anything downstream had an `inputPaths` referencing the actual copied binary path, that edge would be lost. The current `alwaysOutOfDate` approach has the same gap, but worth a grep for `Debug\$EFFECTIVE_PLATFORM_NAME` references in input lists either way.

The macOS-exclusive `return` branch is unambiguously correct and should stay regardless. This suggestion only applies to the mixed-destination branch.

PR #10580, opened 2026-05-01
feat(server): cut bundle artifacts over to ClickHouse (phase 3+4)

fortmarek:
Follow-up to [#10493](https://github.com/tuist/tuist/pull/10493) (phase 1: dual-write + flag) and [#10509](https://github.com/tuist/tuist/pull/10509) (phase 2: backfill). Cuts reads and writes over to ClickHouse and retires the PG `artifacts` table and the bookkeeping flag now that CH holds the full history.

## What changes

- **Single source of truth.** `Tuist.Bundles.Artifact` is now the CH-backed schema; the temporary `ArtifactIngest` sibling is folded back into it. `Tuist.Bundles.Bundle` loses its cross-database `has_many :artifacts` and gains a virtual `:artifacts` field so `bundle.artifacts` keeps working for the JSON response.
- **Writes.** `Bundles.create_bundle/2` writes the bundle to PG and the artifacts to CH inside a single `Repo.transaction`. If `IngestRepo.insert_all` raises (transient CH outage), the PG insert rolls back, the caller sees the underlying error, and we never end up with a bundle row whose artifacts are missing. The previous rescue/flag-flip recovery path is gone — there is no longer a backfill cursor to recover from.
- **Reads.** `bundle_artifacts/1` and `get_bundle_artifact_tree/1` query `ClickHouseRepo`. CH stores `artifact_type` as `LowCardinality(String)`; `decode_artifact_type/1` normalizes it back to the atom shape that filters, comparisons, and the existing JSON serializers expect — so no callers (LiveView filters, MCP tool, controller) had to change.
- **Schema cleanup.** A new migration drops the `bundles.artifacts_replicated_to_ch` column and the legacy PG `artifacts` table. It lives in the **IngestRepo** migrations directory with a timestamp after the phase-2 backfill on purpose: `mix ecto.migrate` iterates `:ecto_repos` (`[Tuist.Repo, Tuist.IngestRepo]`) sequentially per-repo, so a Repo-side "drop column" migration at any timestamp would run before the IngestRepo backfill replays on fresh DBs (CI included) and break the backfill's `WHERE artifacts_replicated_to_ch = false` query. Same precedent as the backfill itself, which is an IngestRepo migration that drives PG via `Tuist.Repo.query!/2`. The CH `artifacts` table is untouched — it's the new source of truth.
- **Boundary export** for `Tuist.Bundles.Artifact` so `TuistWeb.API.Schemas.BundleArtifact` can call the new `Artifact.artifact_types/0` (replaces `Ecto.Enum.values/2`).

## Compliance

`server/data-export.md` no longer mentions the `bundles.artifacts_replicated_to_ch` bookkeeping flag. The `artifacts` entry remains under the ClickHouse section, which is now the only home for the data.

## How to test locally

1. `mise run db:reset && mise run dev` — boots cleanly; the IngestRepo runs the phase-2 backfill, then this PR's drop migration.
2. `mix test test/tuist/bundles_test.exs test/tuist/mcp/components/tools/bundle_tools_test.exs test/tuist/bundles/` — 79/79 pass.
3. Quick end-to-end: upload a bundle via the CLI / API, confirm `SELECT count() FROM artifacts` (CH) increments and the dashboard's bundle inspection page renders the artifact tree correctly.

## Reviewer notes

- The CH insert + PG transaction pairing means a CH outage will surface as a 5xx on bundle creation and rolls back the bundle insert. This is a deliberate change from the phase 1 behavior (which absorbed CH outages by leaving the bundle in PG with the replication flag flipped to `false`); now that CH is the only home for artifacts, there is no longer anywhere to recover the rows from after the fact, so failing loudly is the correct posture.
- `decode_artifact_type/1` is the single normalization point for atom ↔ string. If a future read path bypasses `bundle_artifacts/1` / `get_bundle_artifact_tree/1`, it will need to call this helper too.
- The drop migration is the last cross-DB phase, so the phase-2 backfill module (`priv/ingest_repo/migrations/20260428120000_backfill_artifacts_from_postgres.exs`) is no longer needed for new environments — but it stays in the tree as immutable history; the timestamp ordering inside the IngestRepo dir handles fresh-DB execution order on its own.


server/lib/tuist/bundles.ex
```diff
@@ -6,112 +6,53 @@ defmodule Tuist.Bundles do
...
-  require Logger
-
   @doc """
   Creates a bundle with associated artifacts.
+
+  The bundle row is written to PostgreSQL and the artifacts are written
+  to ClickHouse. Both writes happen inside a single PG transaction so an
+  exception from the ClickHouse insert (e.g. transient outage) rolls back
+  the bundle insert; the caller sees the underlying error and can retry.
   """
```
fortmarek, line 23: I will be moving the bundles table in a follow-up

pepicrft (changes requested):
I reran this on the PR head and hit two verified regressions.

1. `POST /api/projects/:account_handle/:project_handle/bundles` still crashes on the existing controller test at `test/tuist_web/controllers/api/bundles_controller_test.exs:22`. The failure is `KeyError: key :children not found` from `server/lib/tuist_web/controllers/api/bundles_controller.ex:397`. The regression comes from `Bundles.create_bundle/2` returning flattened ClickHouse rows in `bundle.artifacts` at `server/lib/tuist/bundles.ex:43-44`, while `bundle_to_map/1` still expects the nested artifact tree shape.

2. `Bundles.get_bundle(bundle.id, preload: [:artifacts])` still goes through the PostgreSQL association after the migration drops the `artifacts` table. I reproduced this with a one-off local regression case and got `ERROR 42P01 (undefined_table) relation "artifacts" does not exist` from the preload query. The remaining PG preload path is the combination of `has_many :artifacts, Artifact` in `server/lib/tuist/bundles/bundle.ex:63` and `preload: ^preload` in `server/lib/tuist/bundles.ex:67`.

Because both failures are reproducible, I can't approve this as-is.

fortmarek:
Thanks for the second pass.

**Issue 1** (`KeyError :children`): fixed at 0c4b96d. `create_bundle/2` no longer overrides `bundle.artifacts` with the flattened CH rows — it leaves the field as `%Ecto.Association.NotLoaded{}`, which `bundle_to_map/1`'s existing branch renders as `[]`. Same shape the original code returned. The line numbers in your reproduction predate that commit; current HEAD is 29efe24.

**Issue 2** (`preload: [:artifacts]` crashing): valid, fixed at 29efe24. `get_bundle/2` now strips `:artifacts` (and `{:artifacts, _}`) from the caller-supplied preload list before handing it to Ecto, since `bundle.artifacts` is always loaded via `bundle_artifacts/1` from CH regardless. Added a regression test in `test/tuist/bundles_test.exs` covering `Bundles.get_bundle(id, preload: [:artifacts, :uploaded_by_account])` so this can't backslide.

pepicrft (approved):

PR #10581, opened 2026-05-01
fix(server): partition cross-run flakiness detection by scheme

fortmarek:
## Summary

Two CI runs on the same commit but different schemes are different execution variants and should not flag each other as flaky. Previously, a deterministic per-scheme failure (e.g. snapshot tests run against the wrong simulator) was misclassified as flakiness whenever a different scheme produced a passing run on the same commit, surfacing as duplicate-looking entries in the Flaky Runs UI.

## What changed

- **Migration via atomic swap**: builds the new-schema storage table under a temporary name (`test_case_runs_by_commit_new`), backfills it from `test_case_runs` partitions, then atomically swaps it into the canonical `test_case_runs_by_commit` name with `EXCHANGE TABLES`. The previous (no-scheme) legacy data ends up parked under `test_case_runs_by_commit_v2` and is dropped in a follow-up migration. Keeping the canonical name means application code doesn't switch table identifiers across the deploy.
- New ORDER BY puts `scheme` in the prefix: `(project_id, git_commit_sha, scheme, is_ci, status, id)`.
- `Tuist.Tests.TestCaseRunByCommit` schema gains a `scheme` field (still bound to the canonical name).
- `Tuist.Tests.get_existing_ci_runs_for_commit/4` filters by scheme so historical runs from a different scheme can never participate in the cross-run comparison.
- Two new tests in the `cross-run flaky detection` describe block: one asserts that same-commit/different-scheme runs do **not** mark each other flaky; the other guards same-commit/same-scheme cross-run flakiness as a regression.

## Why scheme

Aligns with how Trunk and Buildkite Test Engine model execution variants: identity has to include the variant or the system generates false flakes. Datadog goes further (auto-fingerprints suite + name + parameters + configurations); scheme is the natural Xcode-world variant key and the smallest schema change that closes the most prevalent false-positive case. Destination/device fingerprinting is a reasonable follow-up.

## Rollout plan

1. **This PR** — runs the migration: build new table, backfill, swap into canonical name. Brief window (three DDLs) between dropping the legacy MV and creating the new MV; inserts that arrive in that window are not reflected in the by-commit lookup but are still in `test_case_runs` proper. Same trade-off the original MV migration accepted.
2. **Follow-up PR** (after the rollout is stable) — drop the parked legacy table `test_case_runs_by_commit_v2`. No application code change needed.

### How to test locally

1. `mix ecto.migrate` (the new ingest_repo migration creates, backfills, and swaps the table).
2. `mix test test/tuist/tests_test.exs --only describe:"cross-run flaky detection"` should pass all 7 cases including the two new ones.
3. Manually: ingest two CI runs for the same `git_commit_sha` with different `scheme` values (one passing, one failing) and verify neither run gets `is_flaky = true`.


pepicrft:
I found one migration issue.

In `server/priv/ingest_repo/migrations/20260501120001_add_scheme_to_test_case_runs_by_commit_mv.exs`, `backfill_by_partition()` runs while the existing materialized view is still attached to `test_case_runs_by_commit`. Any `test_case_runs` inserted after a partition has been copied are written only to the legacy storage table. Once the migration drops the view and exchanges the tables, those rows are parked in `test_case_runs_by_commit_v2` and never make it into the canonical table.

I validated this locally against ClickHouse 26.1 with the same sequence: create source plus old MV target, insert row 1, create and backfill the new target, insert row 2 before `DROP VIEW`, then drop the MV and `EXCHANGE TABLES`. The canonical table after the swap had `[1]`, while the parked legacy table had `[2,1]`.

Can we close this by preventing source writes during the backfill, or by doing a catch-up copy after dropping the old MV and before the exchange? If we do a catch-up insert from `test_case_runs`, we should also make sure duplicates cannot affect the lookup before ClickHouse merges them.

pepicrft (approved):

fortmarek:
Good catch, you're right — the naive sequence loses any row whose partition was backfilled before the legacy MV was dropped, since those land in the legacy storage that ends up parked in `_v2` after the swap.

Fixed in `1796d96f11` by adding a catch-up insert after the swap + new-MV creation:

1. Capture a `cutoff` timestamp before backfill starts.
2. Backfill partitions, drop legacy MV, EXCHANGE TABLES, create new MV (as before).
3. **New:** `INSERT INTO test_case_runs_by_commit SELECT ... FROM test_case_runs WHERE inserted_at >= cutoff` — pulls everything that was potentially missed during the migration window directly into the post-swap canonical table.
4. Rename `_new` to `_v2` for follow-up cleanup.

The destination is a `ReplacingMergeTree(inserted_at)` keyed on `(project_id, git_commit_sha, scheme, is_ci, status, id)`, so overlap with rows already copied by the partition backfill or written by the new MV is collapsed by background merges. To address your point about pre-merge duplicates affecting the lookup, I also added an `Enum.uniq_by(& &1.id)` in `Tuist.Tests.get_existing_ci_runs_for_commit/4` as a defensive dedupe for the brief window before merges complete.

Validated locally that the existing `cross-run flaky detection` describe block still passes (7/7).

PR #10582, opened 2026-05-02
fix(cli): skip fully cached missing test plans

pepicrft:
## Summary
- persist original test-plan-to-target membership in the selective testing graph
- treat `xctestproducts does not contain test plan` as a skip when the requested plan was fully pruned by selective testing
- add focused coverage for both the skip case and the unknown-plan failure case

## Root cause
Selective testing can prune every target in a test plan when all of them are already cached as passing. When that happens, Xcode does not emit an `.xctestrun` for the plan in `.xctestproducts`, so `test-without-building -testPlan ...` fails even though the correct outcome is effectively a skip.

## Validation
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing:TuistKitTests/TestServiceTests/test_run_testWithoutBuilding_skipsMissingTestPlanWhenItWasFullyCached -only-testing:TuistKitTests/TestServiceTests/test_run_testWithoutBuilding_doesNotSkipMissingUnknownTestPlan CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`


fortmarek (commented):
A couple of design suggestions before this lands. Both are about reducing the new persisted state.

## 1. Reuse `SelectiveTestingGraph` instead of adding a new sidecar

`SelectiveTestingGraph` is already written to (and read from) the same `.xctestproducts` location at the same point in the flow. Adding a second JSON sidecar doubles the I/O without much gain, since the two structures are produced together and consumed together. An optional field on `SelectiveTestingGraph` (default-empty for back-compat) keeps the bundle layout simpler and removes the new type, the new file, and the optional-decoding branch in `runTestWithoutBuildingFromBundle`.

## 2. The skip can probably be derived from the bundle itself

The bug is rooted in Xcode not emitting an `.xctestrun` for a fully pruned test plan. That same absence is also the cleanest signal we can read back: before invoking `xcodebuild test-without-building`, check whether the requested plan's `.xctestrun` exists in the bundle. If it doesn't, treat as the fully-pruned skip case. This avoids mirroring the user-filter logic (`-only-testing` / `-skip-testing` / passthrough `-skip-testing`) at the Tuist layer. xcodebuild already does that correctly when an `.xctestrun` exists; the only case it gets wrong is the missing-file one we're targeting.

## 3. Typo / unknown-plan regression worth covering

I think both the current implementation and the simpler ".xctestrun absence" detection share a hole: if the user passes `-testPlan SomeRandomPlanThatNeverExisted`, both would silently report success rather than surfacing Xcode's clear error.

- In the current PR: `runnableTargetNames` is computed from the build-time plan. If that plan was fully pruned, `runnableTargetNames = []`, so any later `-testPlan <typo>` also intersects to empty and skips.
- In the bundle-only approach: a missing `.xctestrun` for a typo plan is indistinguishable from a missing `.xctestrun` for a fully-pruned plan.

To fix this without bringing back the full pre/post-pruning sets, the minimum viable signal is a list of the test plans (or the scheme name when no plan was used) that `build-for-testing` actually attempted. The rule then becomes:

- Requested plan **not** in the attempted list: let xcodebuild run and fail with its native error.
- Requested plan **in** the list but its `.xctestrun` is missing: skip with success (fully pruned).
- Requested plan **in** the list and `.xctestrun` present: invoke xcodebuild as today.

That's one `[String]` field on `SelectiveTestingGraph` (e.g. `attemptedTestPlans`) instead of two `[String]` fields on a separate type, and it closes the typo regression that's not covered today. Would also be worth adding a test for the typo case alongside the existing fully-cached coverage.

pepicrft:
Thanks @fortmarek, all three addressed in a25966e:

1. Dropped the sidecar; `SelectiveTestingGraph` now carries `attemptedTestPlans: [String]` (decoded as empty for older bundles).
2. Skip now checks for the requested plan's `.xctestrun` in the bundle: `<plan>.xctestrun` for plans, `<scheme>_*.xctestrun` for plan-less schemes.
3. If the requested plan isn't in `attemptedTestPlans`, we fall through to xcodebuild and surface its native error. Added `test_run_testWithoutBuilding_doesNotSkipUnknownTestPlan` to cover the typo case.

fortmarek (commented):
Looks great, all three points addressed cleanly. A few small observations on the new shape:

## What's good

- `SelectiveTestingGraph.attemptedTestPlans` with an explicit `init(from:)` that defaults to `[]` keeps older bundles decodable. Synthesized `encode(to:)` still kicks in via the explicit `CodingKeys`, so writes pick up the new field automatically.
- Skip predicate is now a clean three-way: not in `attemptedTestPlans` (typo, fall through), in the list and `.xctestrun` present (run xcodebuild), in the list and `.xctestrun` missing (fully pruned, skip with success).
- `requestedTestPlanOrScheme` falls back to `schemeName`, and the same value gets passed to `uploadSkippedTestSummary` so the server gets the correct `testPlanName` for skipped runs (slight improvement over the existing `schemes.first?.name` call site).
- New `test_run_testWithoutBuilding_doesNotSkipUnknownTestPlan` locks in the regression we discussed, and the build-path test now asserts `attemptedTestPlans == [\"TestScheme\"]`.

## Minor things worth a look

1. **`bundleHasXCTestRun` prefix match has a theoretical collision case.** Today it returns true if `basename == name || basename.hasPrefix(\"\\(name)_\")`. If someone has both a scheme called `App` (no plans) and a test plan called `App_Smoke`, then a request for `-testPlan App_Smoke` would also match `App_iPhone16Pro_iOS18.xctestrun` via the prefix rule and incorrectly conclude the plan is present. Probably never happens in practice, but tightening the scheme-shape match to require something like a destination suffix (or just running the prefix branch only when `testPlanConfiguration?.testPlan == nil`) would close it cheaply.

2. **Backward-compat coverage.** The decoder default-empty path isn't directly exercised by a test. A small \"old bundle without `attemptedTestPlans`\" case (write JSON with only `testTargetHashes`, then run `test-without-building`, assert it falls through to xcodebuild) would lock in the back-compat behavior.

3. **Multi-plan scheme coverage.** `attemptedTestPlans` has a non-trivial branch for `scheme.testAction?.testPlans` being non-empty, but no test currently exercises the scheme-with-multiple-plans build path. Easy to add to the existing build assertion.

4. **Tiny nit:** the explicit `enum CodingKeys` is redundant with what Swift would synthesize for the two stored properties. You can drop it and keep just the custom `init(from:)` referencing `CodingKeys.self` synthesized from the property names. Not worth a churn, just noting.

Non-blocking on all four. LGTM modulo (1) which I'd pick up here or in a follow-up.

fortmarek (approved):

pepicrft:
Picked up all four in 28a19ee:

1. `bundleHasXCTestRun` now skips the prefix branch when a `-testPlan` is requested, so the `App_Smoke` collision case can't trip it.
2. New `test_run_testWithoutBuilding_doesNotSkipWhenLegacyGraphHasNoAttemptedTestPlans` writes a JSON without `attemptedTestPlans` and asserts xcodebuild is still called.
3. New `test_run_build_writesAllTestPlanNames_whenSchemeHasMultiplePlans` exercises the multi-plan build path and asserts both names land in the graph.
4. Dropped the redundant `CodingKeys` enum.

PR #10585, opened 2026-05-03
ci(workflows): optimize handbook, release, cli, and server pipelines

pepicrft:
> Trims CI inefficiencies surfaced by an audit of `.github/workflows/`.

#### `handbook.yml`
- Bump `actions/cache@v3` → `@v4` for the PNPM store restore step.

#### `release.yml`
- Drop `fetch-depth: 0` from release jobs that don't read git history: `release-cli-linux` (matrix x2), `release-ios`, `release-android`, `release-kura-binaries` (matrix x4), `release-kura-docker`. Jobs that run `git cliff` or compare against historical tags (`check-releases`, `release-cli`, `release-app`, `release-server`, `release-cache`, `release-kura-prepare`, `release-gradle`, `release-skills`, `release-noora`, `release-helm`, `commit-and-release`) still clone the full history.

#### `cli.yml`
- Add `restore-keys` fallback to every `.build` cache block (7 macOS jobs + 2 Linux jobs). Any `Package.swift`/`Package.resolved` bump used to invalidate the cache completely; now Swift incremental-compiles from the most recent stale `.build`.
- Flip the `Install Tuist dependencies` guards from `cache-matched-key == ''` to `cache-hit != 'true'`. With `restore-keys` present, a fallback restore would otherwise count as a hit and skip `tuist install` on stale deps. `cache-hit` is only true on exact primary-key match.
- Convert `cli-spm-build`'s cache from `actions/cache/restore@v4` to `actions/cache@v4` (full restore + post-job save). Today only `cli-cache` (main-only) saves the cache, so PRs always start from main's last save. With this change, the SwiftPM build saves on every PR push, so subsequent pushes to the same PR get a warm cache across all 4 macOS jobs.

#### `server.yml`
- Add the draft-PR skip to the `format` job. Every other server job (`test`, `credo`, `gettext`, `esbuild`, `security`) already has it; `format` was paying ~1.5 min on every draft push for nothing.
- Extract `nscloud-cache-action` + `mise-action` + mix cache restore + `mix deps.get` into `.github/actions/setup-server-mix/`. Used by `gettext`, `test`, `credo`, `format`, `esbuild`, `security`, `seed` — replaces ~30 lines per job. Future tweaks (e.g. mix cache key bump) become atomic.

### Notes on what was deliberately *not* changed

- `cache: "false"` on `jdx/mise-action` is intentional — it delegates caching to `nscloud-cache-action` (~1s NVMe restore vs ~38s GHA cache), per PRs #9530 and #9535.
- `nscloud-setup-buildx-action` provides the buildx layer cache transparently; no `cache-from`/`cache-to` needed.
- The originally-considered "serialized warm-build job" for CLI was dropped in favor of the `cli-spm-build` cache-save approach. With `restore-keys` increasing cache hit rate, a serialized warmer would have added wall-time tax on the common path; the cache-save approach is a strict win.

### How to test locally

No local test path. Validation is the next CI run on this PR (workflows still pass, no behavioral regressions) and the next release cascade (no broken `git cliff` invocations on the trimmed `fetch-depth` jobs).


fortmarek (commented):
The bulk of this looks good — `cache-hit` vs `cache-matched-key` flip is correct, the `fetch-depth: 0` removals all target jobs that don't read git history, and the `setup-server-mix` composite is a clean extraction.

One concern I'd want resolved before merging: the `restore-keys` fallback on the macOS `.build` cache. Two recent incidents are directly relevant.

### Prior incident 1 — Linux linker failures from stale `.build`

[e8fa267](https://github.com/tuist/tuist/commit/e8fa267c95) "fix(cli): only cache dependency checkouts in Linux CI jobs":

> Caching the entire .build directory caused flaky linker failures because stale compiled object files from previous runs could be restored, leading to undefined symbol errors at link time.

Severe enough that the fix was to stop caching `.build` outputs entirely; [040d69d](https://github.com/tuist/tuist/commit/040d69d925) renamed the prefix to invalidate lingering entries. Linux is back to caching full `.build` now, so either SwiftPM matured or this is tolerated — worth knowing which.

### Prior incident 2 — macOS workspace state pointing at moved paths

[0c6f20c](https://github.com/tuist/tuist/commit/0c6f20c116) / #10510 "fix(cli): include Package.swift in CI cache key":

> macOS jobs kept restoring stale `.build` workspace state pointing at the old `processor/native/xcactivitylog_nif` path, and `tuist install` was skipped on cache hit, so `tuist cache` / `tuist test` failed with \"Package.swift not found\".

### Why it matters here

The new `restore-keys: ${{ runner.os }}-cli-${{ hashFiles('.xcode-version') }}-` falls back to any prior `.build` regardless of `Package.swift`/`Package.resolved` content. The incident-2 scenario reproduces:

1. Someone moves a local SPM package
2. Primary key misses (good)
3. `restore-keys` falls back to the previous `.build` with stale workspace state (bad)
4. `cache-hit != 'true'` triggers `tuist install` → runs `swift package resolve`

Step 4 is the mitigation — *provided* `swift package resolve` rewrites `workspace-state.json` to point at the new path. The historical fix's wording (\"tuist install was *skipped* on cache hit\") implies running it would have repaired things, so the model should hold. But that's a *should*, not a *known*, and incident 1 shows partial-artifact reuse can produce undefined-symbol linker failures that `swift package resolve` won't fix — those need a clean `.build`.

### Suggestions

1. **Validate the incident-2 scenario directly** — push a commit on this PR that moves a local SPM package path and confirm `tuist cache` / `tuist test` still pass. A vanilla CI run on this PR won't exercise that boundary.
2. **Scope `restore-keys` more tightly**, e.g. include `Package.swift` (but not `Package.resolved`) in the fallback prefix:
   ```yaml
   key: ${{ runner.os }}-cli-${{ hashFiles('.xcode-version') }}-${{ hashFiles('Package.resolved', 'Package.swift') }}
   restore-keys: |
     ${{ runner.os }}-cli-${{ hashFiles('.xcode-version') }}-${{ hashFiles('Package.swift') }}-
   ```
   Keeps the win on `Package.resolved` churn (the common case) without falling back across structural `Package.swift` changes — which is the exact boundary that broke before.
3. **Leave a comment in the workflow** referencing #10510 / e8fa267 so the next person to touch these keys knows the scarred tissue.

Other minor notes:

- `setup-server-mix` defaults `mix-cache-key: mix`, shared across `gettext`/`credo`/`format`/`esbuild`/`security`/`seed`. Compatible today, but a future job needing isolated mix state would need to remember to override — worth a one-line note in the composite's inputs.
- `cli-spm-build` will now race other macOS jobs on the same primary cache key. \"Unable to reserve cache\" warnings already happen elsewhere, so this is just one more racer, not a new failure mode.

Happy to approve once we've either narrowed the restore-keys or confirmed via experiment that the wider fallback handles the moved-local-package case.

fortmarek (approved):

pepicrft:
All addressed in e67c72de5d. Restore-keys now hash `Package.swift` on both macOS and Linux jobs, added a comment block above `jobs:` pointing at #10510 / e8fa267, and beefed up the `mix-cache-key` description in `setup-server-mix` with the shared-default caveat and the test override. Held off on the moved-local-package experiment, the narrower fallback should keep us clear of that scenario, but tell me if you'd rather see a deliberate test before this lands.

PR #10595, opened 2026-05-04
feat(server): migrate bundles table from PostgreSQL to ClickHouse (phase 1)

fortmarek:
Mirrors the phase-1 setup [#10493](https://github.com/tuist/tuist/pull/10493) introduced for `artifacts`, applied to the `bundles` table. Establishes the destination CH table, the per-bundle backfill cursor, and starts dual-writing new bundles to ClickHouse. PostgreSQL stays the source of truth for bundle reads in this phase.

## What's in phase 1

- **CH `bundles` table** as `MergeTree` ordered by `(project_id, inserted_at, id)` and partitioned by `toYYYYMM(inserted_at)` — same shape as `build_runs` and the recently-created `artifacts` CH table. `install_size` widens to `Int64` (the original motivation for getting off PG `int4`). Enum-valued `type` and `supported_platforms` are stored as `LowCardinality(String)` / `Array(LowCardinality(String))` for human-readable ad-hoc queries.
- **`bundles.replicated_to_ch`** PG column. Default `false` so existing rows are picked up by the eventual phase-2 backfill; new rows get `true` from the schema's struct default. Doubles as the resume cursor.
- **`Tuist.Bundles.BundleIngest`** — CH-shaped sibling of `Tuist.Bundles.Bundle` with `Tuist.Ingestion.Bufferable`. Will fold back into `Bundle` in phase 4 once the PG schema goes away. Atom-valued enums encoded to strings, timestamps normalized to `NaiveDateTime` µs:6.
- **Dual-write** in `Bundles.create_bundle/2`. After the existing `Repo.transaction` (PG bundle insert + artifacts CH insert) commits, the bundle is shadow-written via `BundleIngest.Buffer.insert/1`. A CH outage on the shadow write is **logged**, the row's `replicated_to_ch` flag is flipped to `false`, and the caller still observes a successful `{:ok, bundle}` — PG is the source of truth for reads in this phase, so failing loud here would be wrong.

## Phase plan

1. **Phase 1 (this PR)**: CH table, dual-write, replication flag. Deploy is cheap — only fast DDL migrations.
2. **Phase 2 (follow-up PR)**: backfill migration covering `replicated_to_ch = false`. Splitting it out lets us observe dual-write working in production before committing to the long migration, scope the backfill to PG rows that pre-date dual-write, and revert either piece independently.
3. **Phase 3**: cut bundle reads over to CH. All bundle reads currently funnel through `lib/tuist/bundles.ex`, so the cutover is contained — callers don't change.
4. **Phase 4**: drop the PG `bundles` table and the `replicated_to_ch` column; fold `BundleIngest` back into `Bundle`. Update `server/data-export.md`.

## What's NOT in this PR

- **The backfill** — phase 2.
- **Read cutover** — phase 3.
- **Dropping PG `bundles`** — phase 4.

## Compliance

`server/data-export.md` updated to note the new CH `bundles` table under the analytics section, with a callout that PostgreSQL remains authoritative for export until the cutover completes.

## How to test locally

1. `mise run db:reset && mise run dev` — boots cleanly; the two new migrations run during startup.
2. `mix test test/tuist/bundles_test.exs` — 55/55 passing locally, including the two new tests:
   - `shadow-writes the bundle to ClickHouse` — verifies all fields land in CH (atom→string for `type` and `supported_platforms`, `Int64` size, nullable account/git fields) and that the PG row is marked `replicated_to_ch = true`.
   - `marks the bundle as not replicated when the ClickHouse dual-write fails` — stubs `IngestRepo.insert_all` to raise only for `BundleIngest`; the bundle still commits in PG with `replicated_to_ch = false` so the eventual backfill catches it.
3. Create a bundle via the CLI as usual; the row appears in both PG `bundles` and CH `bundles` tables.

## Reviewer notes

- Phase 1 deploys are low-risk: only fast DDL plus the new dual-write code path. Bake time for dual-write before opening phase 2: a few days at minimum, ideally a week, so we can spot any encoding issues against real-world bundles.
- The dual-write contract differs from the post-phase-4 `artifacts` write (which is synchronous inside the bundle transaction): bundle dual-write is async via `Buffer` and never propagates failures, because PG is still the source of truth for bundle reads. Once phase 3 cuts reads over and phase 4 removes the PG table, that contract will tighten the same way the artifacts one did.
- `BundleIngest.from_bundle/1` is the single boundary that converts the PG-shaped struct (atom enums, `DateTime`, possibly nil git fields) to the CH row shape. If a future write path bypasses `create_bundle/2`, it should go through this helper.

### DIFF
--- a/server/lib/tuist/bundles.ex
+++ b/server/lib/tuist/bundles.ex
@@ -7,11 +7,14 @@ defmodule Tuist.Bundles do
 
   alias Tuist.Bundles.Artifact
   alias Tuist.Bundles.Bundle
+  alias Tuist.Bundles.BundleIngest
   alias Tuist.Bundles.BundleThreshold
   alias Tuist.ClickHouseRepo
   alias Tuist.IngestRepo
   alias Tuist.Projects.Project
   alias Tuist.Repo
+
+  require Logger
 
   @doc """
   Creates a bundle with associated artifacts.
@@ -20,6 +23,13 @@ defmodule Tuist.Bundles do
   to ClickHouse. Both writes happen inside a single PG transaction so an
   exception from the ClickHouse insert (e.g. transient outage) rolls back
   the bundle insert; the caller sees the underlying error and can retry.
+
+  Once the transaction commits, the bundle row is shadow-written to the
+  ClickHouse `bundles` table. PostgreSQL remains the source of truth for
+  bundle reads in this phase, so a ClickHouse outage on the shadow write
+  is logged and the row's `replicated_to_ch` flag is flipped to `false`
+  so the eventual backfill picks it up; the caller still observes a
+  successful create.
   """
   def create_bundle(attrs \\ %{}, opts \\ []) do
     {artifacts, bundle_attrs} = Map.pop(attrs, :artifacts, [])
@@ -38,6 +48,7 @@ defmodule Tuist.Bundles do
 
     case result do
       {:ok, %{bundle: bundle}} ->
+        replicate_bundle_to_clickhouse(bundle)
         {:ok, Repo.preload(bundle, preload)}
 
       {:error, :bundle, changeset, _} ->
@@ -50,6 +61,24 @@ defmodule Tuist.Bundles do
   defp insert_artifacts_to_clickhouse(flattened) do
     IngestRepo.insert_all(Artifact, flattened)
     :ok
+  end
+
+  defp replicate_bundle_to_clickhouse(%Bundle{} = bundle) do
+    bundle
+    |> BundleIngest.from_bundle()
+    |> BundleIngest.Buffer.insert()
+
+    :ok
+  rescue
+    error ->
+      Logger.error("Failed to dual-write bundle #{bundle.id} to ClickHouse: #{Exception.message(error)}")
+
+      Repo.update_all(
+        from(b in Bundle, where: b.id == ^bundle.id),
+        set: [replicated_to_ch: false]
+      )
+
+      :ok
   end
 
   @doc """
--- a/server/lib/tuist/bundles.ex
+++ b/server/lib/tuist/bundles.ex
@@ -63,11 +63,13 @@ defmodule Tuist.Bundles do
     :ok
   end
 
+  # Synchronous on purpose: an async buffer write would return before the
+  # row reaches ClickHouse, so a later flush failure could not flip the
+  # `replicated_to_ch` flag and the backfill would skip the bundle even
+  # though it never landed in CH.
   defp replicate_bundle_to_clickhouse(%Bundle{} = bundle) do
-    bundle
-    |> BundleIngest.from_bundle()
-    |> BundleIngest.Buffer.insert()
-
+    row = BundleIngest.from_bundle(bundle)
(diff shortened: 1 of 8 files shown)


fortmarek:
Re. the Blick P3 finding on the `excellent_migrations:safety-assured-for-next-line` directive — that suppression syntax is the canonical one used 58 times across `server/priv/repo/migrations/` (e.g. [20260410120000_create_automation_alerts.exs](https://github.com/tuist/tuist/blob/main/server/priv/repo/migrations/20260410120000_create_automation_alerts.exs)). It's the official directive recognized by the `excellent_migrations` Hex library, not a typo. False positive — no change required.

server/lib/tuist/bundles.ex
```diff
@@ -52,6 +63,24 @@ defmodule Tuist.Bundles do
     :ok
   end
 
+  defp replicate_bundle_to_clickhouse(%Bundle{} = bundle) do
+    bundle
+    |> BundleIngest.from_bundle()
+    |> BundleIngest.Buffer.insert()
```
pepicrft: Could we avoid the buffered `BundleIngest.Buffer.insert/1` for this replication marker path, or make this call synchronous? In production `Bufferable` falls through to `Buffer.insert/2`, and `Buffer.insert/2` is a `GenServer.cast` unless `sync_writes` is enabled. The test config sets both `sync_writes: true` and `write_through_repo: true`, so the new failure test only covers the synchronous test-only path. In production a ClickHouse outage during the later buffer flush raises in the buffer process after this function has returned, leaving `replicated_to_ch = true`; the backfill then skips the bundle even though it never landed in CH. For this migration flag to be trustworthy, the CH write needs to fail in the caller before we leave the row marked replicated, or the buffer needs an error callback that flips the flag for the specific rows it lost.

pepicrft (changes requested):
I found one blocking issue around the replication marker. The bundle dual-write goes through the async ingestion buffer in production, so a later ClickHouse flush failure can leave `replicated_to_ch = true` for a row that never reached CH. The current tests run with synchronous/write-through buffer config, so they do not cover that production path.

fortmarek (commented):
Good catch — fixed in ec485edf1d. The dual-write now bypasses the buffer and calls `IngestRepo.insert_all(BundleIngest, [row])` directly, so a CH outage raises synchronously into the rescue clause and flips `replicated_to_ch = false` for the specific row that was lost. Dropped the now-unused `Bufferable` macro from `BundleIngest` and its supervision-tree entry too. Tests still pass 55/55.

server/lib/tuist/bundles.ex
fortmarek: Good catch — fixed in ec485edf1d. The dual-write now bypasses the buffer and calls `IngestRepo.insert_all(BundleIngest, [row])` directly, so a CH outage raises synchronously into the rescue clause and flips `replicated_to_ch = false` for the specific row that was lost. Dropped the now-unused `Bufferable` macro from `BundleIngest` and its supervision-tree entry too. Tests still pass 55/55.

pepicrft (approved):

PR #10597, opened 2026-05-04
feat(server): backfill bundles from PostgreSQL to ClickHouse (phase 2)

fortmarek:
Follow-up to [#10595](https://github.com/tuist/tuist/pull/10595). Phase 1 set up the destination CH `bundles` table, the dual-write, and the per-bundle `replicated_to_ch` flag. This PR fills in the long-running backfill that flips that flag for every existing bundle, so phase 3 can cut reads over to ClickHouse.

## Strategy

The migration paginates over `bundles` where `replicated_to_ch = false`, fetches each batch, inserts the rows into CH, and flips the flag in a single PG `UPDATE`. Once the loop drains, every bundle lives in CH and the live dual-write keeps things in sync from there.

The flag itself doubles as the resume cursor: a killed run restarts and the `WHERE` filter naturally skips already-flipped bundles. Two populations get caught:

1. Pre-dual-write history (column-defaulted to `false` by the phase-1 schema migration).
2. Bundles whose live dual-write hit a transient ClickHouse outage post-deploy and were rescued back to `false` by `Tuist.Bundles.replicate_bundle_to_clickhouse/1`.

Bundles whose dual-write is *in flight* never enter the scan because they are inserted with `replicated_to_ch = true` from the start (struct default in `Tuist.Bundles.Bundle`). The migration cannot race with a live dual-write.

## Why inline / raw schemaless queries

The migration is self-contained: no `from(b in Bundle, ...)`, no shared backfill module. Migrations are immutable history, but self-hosted upgrades may run them long after the live `Bundle` / `BundleIngest` schemas have drifted from the column shape this migration assumes. Referencing a live schema would silently rewrite the meaning of the migration on a multi-version jump. Raw `from(... in "bundles", ...)` and inline `@ch_types` / `@platform_by_int` / `@type_by_int` maps pin the migration to the column shape that exists today, matching the convention used by [`BackfillTestRunsFromCommandEvents`](https://github.com/tuist/tuist/blob/main/server/priv/ingest_repo/migrations/20251118211224_backfill_test_runs_from_command_events.exs), [`BackfillBuildRunsFromPostgres`](https://github.com/tuist/tuist/blob/main/server/priv/ingest_repo/migrations/20260211124936_backfill_build_runs_from_postgres.exs), and the artifacts phase-2 backfill.

UUIDs come out of PG as 16-byte binaries (raw `:uuid` columns, no schema cast); a small `encode_for_ch` helper decodes them to 36-char strings via `Ecto.UUID.load!/1` and bumps timestamp precision before the CH `INSERT`. Atom-valued enums (`type`, `supported_platforms`) are encoded against the frozen-at-write-time int to string maps so a future enum reordering doesn't silently rewrite the meaning of historic backfilled rows.

## Production-safety knobs (matched against the other long-running CH backfills)

- **`Process.sleep(@throttle_ms = 500)` between batches**: same throttle the test-runs and first-run-events backfills use. Gives live PG/CH traffic breathing room and keeps CH merges from falling behind `parts_to_throw_insert`.
- **`log: false`** on the per-batch bundle `SELECT`, the CH `INSERT`, and the PG flag `UPDATE`. Otherwise we'd dump full UUID arrays + every row payload at debug level for every batch. The per-batch `Logger.info` summary still surfaces progress.
- **`timeout: :infinity`** on the CH insert and PG update. Default Ecto timeout is 15s; one batch is up to 500 rows.
- **`@disable_ddl_transaction true` / `@disable_migration_lock true`** so the migration doesn't hold a long PG transaction or block other migrations.

## Race window

Between the CH `INSERT` and the PG flag `UPDATE` for a given batch. A crash inside that window re-inserts that batch's rows on the next run. The destination is `MergeTree` (no row-level dedup), so duplicates persist until something else collapses them. Bounded by one batch (500 bundles) and an unlikely crash, so we accept it rather than pay the cost of a deduplicating engine or per-batch existence checks. Documented in the moduledoc.

## What's NOT in this PR

- **Read cutover**: phase 3.
- **Dropping PG `bundles`**: phase 4.

## Compliance

`server/data-export.md` updated to note that the CH `bundles` table now holds the full history (dual-write + backfill), so it is viable for GDPR Article 20 / CCPA exports even before the phase 3 read cutover.

## How to test locally

1. `mise run db:reset && mise run dev`: boots cleanly; the new migration runs during startup and is a no-op on a fresh DB.
2. `mix test test/tuist/bundles_test.exs`: 55/55 passing.

## Reviewer notes

- The migration is the long-running step. Production will run this as a pre-upgrade hook. Runtime is dominated by PG read latency, not CH ingestion.
- `@batch_size = 500`: raise if production observes the per-batch round-trip dominates over CH insert throughput. Logs report per-batch counts, so it's easy to tune.


pepicrft:
I verified the migration behavior locally with isolated Postgres and ClickHouse databases. The implementation does what it says for the current false rows and CI is green.

One deployment caveat to handle before the phase 3 read cutover: if someone upgrades directly from before #10595 to this release while old pods can still serve traffic during the pre-upgrade migration, an old pod can insert a bundle after this backfill has drained. That row relies on the DB default `replicated_to_ch = false`, is not dual-written to ClickHouse, and the migration is already recorded as complete, so it will stay missing from CH.

My local reproduction ended in this state:

```text
pg_total=2
pg_unreplicated_after_completed_migration=1
migration_recorded=1
ch_total=1
```

I would make the rollout requirement explicit, or enforce that phase 2 cannot be the same upgrade that first introduces the bundle dual-write for self-hosted installs.

pepicrft (approved):
I verified the migration behavior locally with isolated Postgres and ClickHouse databases. The implementation does what it says for the current false rows and CI is green, so I am approving.

One deployment caveat to handle before the phase 3 read cutover: if someone upgrades directly from before #10595 to this release while old pods can still serve traffic during the pre-upgrade migration, an old pod can insert a bundle after this backfill has drained. That row relies on the DB default `replicated_to_ch = false`, is not dual-written to ClickHouse, and the migration is already recorded as complete, so it will stay missing from CH.

My local reproduction ended in this state:

```text
pg_total=2
pg_unreplicated_after_completed_migration=1
migration_recorded=1
ch_total=1
```

I would make the rollout requirement explicit, or enforce that phase 2 cannot be the same upgrade that first introduces the bundle dual-write for self-hosted installs.

PR #10599, opened 2026-05-04
feat(server): comparison-based automations with baseline establishment

fortmarek:
Screenshot 2026-05-05 at 16 43 32

## Summary

Two changes that go together:

1. **Comparison is a first-class field on the alert.** Pair the metric (`flakiness_rate` / `flaky_run_count`) with `>=`, `>`, `<`, or `<=` and a threshold. Existing alerts default to `>=` — no migration of trigger configs.
2. **First evaluation establishes the baseline silently.** The previous implicit "below comparisons scope to flagged tests, above comparisons scan everything" coupling is gone. A new alert records every currently-matching test case as a `triggered` AlertEvent *without firing trigger actions*; subsequent evaluations only fire on transitions across the threshold. The candidate set is uniform across comparisons (every test with at least one run in the window).

This makes the engine generic again — the worker doesn't know anything about flaky-specific scoping rules; the monitor just reports current matches; the state machine handles the rest. It also lets users configure cleanup automations (`Flaky runs < 1 over 30d` + `Unmark as flaky`) without spamming every passing test on creation.

### Schema
- `automation_alerts.baseline_established_at` (nullable timestamp). Migration backfills existing rows to `NOW()` so they don't re-baseline on first eval after deploy.
- `trigger_config.comparison` (string, default `gte`). Validated against `gte` / `gt` / `lt` / `lte`.

### UI
- `Comparison` dropdown alongside `When`, with `Greater or equal` / `Greater than` / `Less than` / `Less or equal`.
- `Count` / `Percent` instead of `Threshold (count)` / `Threshold (%)` so the label fits the compact column.
- `Mark as flaky` and `Unmark as flaky` are both available in the trigger and recovery action dropdowns; mutually exclusive within a single action list.
- Form-row widths pinned (dropdowns 11rem, compact inputs 7rem).
- Switching `Comparison` doesn't reset threshold or trigger actions — the user's input is preserved. Switching the metric does reset the threshold (count vs % aren't comparable units).

### Concurrency note (deferred)

Two alerts touching the same `test_case_id` concurrently can race inside `Tests.update_test_case/2` — both read the same row, both write back. ReplacingMergeTree's `inserted_at` resolves the row state; the event log can pick up duplicate `marked_flaky` / `unmarked_flaky` entries (`argMax` still resolves correctly) and notify-style actions can fire twice. For the detect+cleanup case both alerts want the same end state, so the race is benign in practice. A per-`test_case_id` lock would close this; tracked separately, not in this PR.

## Test plan

- [x] `mix test test/tuist/automations test/tuist_web/live/project_automations_live_test.exs` — 78 tests, 0 failures (baseline-establishment tests for the worker, comparison-based monitor tests for `lt` / `gt` / default `gte`, alert validation tests, LiveView dispatch tests)
- [x] `mix test test/tuist/tests_test.exs` — 200 tests, 0 failures
- [x] `mix format`, `mix credo --strict` on touched files
- [x] Manual smoke test of the Create automation modal across all four comparisons


pepicrft (approved):
LGTM. Two things worth a thought, neither blocking:

**Orphan path bypasses the recovery window.** The active-event path enforces `triggered_at < cutoff` (e.g. 14 days without trigger), but orphans get recovered on the very next worker tick if the monitor doesn't currently flag them. That means a manual UI/API mark can be undone within minutes. Reading the PR description this seems intentional (the whole point is "stayed flagged forever"), and there's no `triggered_at` to compare for orphans anyway, so it's structural rather than a missed check. A short comment in `recover_orphaned_flaky_test_cases` noting the asymmetry would save the next reader a head scratch.

**`already_recovered_ids` parameter is redundant.** The recovered ids are filtered from `active_events`, so they're already covered by `MapSet.new(active_events, & &1.test_case_id)`. The `MapSet.union` is a no-op and the "no double-recover" test passes via the active-events containment. Fine to keep as defensive, but you could drop the parameter.

Also: the new query runs per recovery-enabled alert per evaluation. Bounded by project parts thanks to the leading sort key, but worth keeping an eye on if alerts proliferate per project.

Nice extraction of `recover_test_case/2` and good test coverage on the four orphan paths.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -25,6 +25,37 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
...
+    end
+
+    :ok
+  end
+
+  # First evaluation after the alert was created: every test case currently
+  # matching the condition is part of the established state. Record them as
+  # `triggered` AlertEvents so subsequent evaluations only fire on
+  # transitions, but skip the trigger actions — there's no transition to
+  # announce yet, and firing for the entire matching set would spam users.
```
fortmarek, line 42: in other words, if you add an automation that sends a slack message when a test case flaky runs in the last 30d go below 3, you don't want that to actually trigger in the first run for all tests that comply with this condition (as you would get a _ton_ of slack messages, which is probably not what you want...)

server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
```diff
@@ -1,91 +1,181 @@
...
-        where: tcr.project_id == ^project_id,
-        where: tcr.ran_at >= ^cutoff,
-        select: tcr.test_case_id,
-        distinct: true
-      )
+  # The MV is keyed on `(project_id, date, test_case_id)`, so we round the
+  # cutoff to the start of the day. A 30-day window evaluated mid-day picks
+  # up a few hours of additional data on day -30 — acceptable for
+  # threshold-based alerts and well within the noise of test-run timing.
+  defp window_cutoff_date(window_seconds) do
```
pepicrft, line 77: I think this loses the configured semantics for sub-day windows. A `1m` or `1h` window now becomes `date >= today`, so it includes every run since midnight UTC. If cleanup automations are intended to support the same `m` and `h` windows as the parser accepts, this needs an exact timestamp filter somewhere. If we only want daily buckets, I would constrain validation/UI to day-level windows so users do not think minute/hour windows are honored.

server/lib/tuist_web/live/project_automations_live.ex
```diff
@@ -467,11 +503,26 @@ defmodule TuistWeb.ProjectAutomationsLive do
...
+      symbol: symbol,
+      threshold: threshold,
+      window: window
+    )
   end
 
   def automation_summary(_), do: ""
 
+  defp scope_phrase(comparison) when comparison in @below_comparisons,
+    do: dgettext("dashboard_projects", "a flaky-marked test's ")
```
pepicrft: This copy still says below comparisons are scoped to flaky-marked tests, but the monitor now evaluates every test case with at least one run in the window. I would remove this scope phrase or reword it so the summary matches the new generic baseline model.

pepicrft (approved):
Left a few follow-up comments. The overall direction looks good to me.

server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
fortmarek: Addressed in 6ee10b7339 — went with option (b): added a `valid_window?/1` Alert validator that rejects anything other than `Nd`, and dropped the `m` / `h` clauses from `parse_window/1` so the contract is explicit. `recovery_config.window` keeps minute/hour support since it's a wait-time, not a measurement window.

server/lib/tuist_web/live/project_automations_live.ex
fortmarek: Fixed in 6ee10b7339. Dropped the `scope_phrase` helper and the "a flaky-marked test's" prefix from both `automation_summary` clauses. The summary now reads e.g. "When flakiness rate < 5% over 30d" regardless of comparison direction — matches the generic baseline model.

server/priv/repo/migrations/20260505160000_add_baseline_established_at_to_automation_alerts.exs
```diff
@@ -0,0 +1,21 @@
+defmodule Tuist.Repo.Migrations.AddBaselineEstablishedAtToAutomationAlerts do
+  use Ecto.Migration
+
+  def up do
+    alter table(:automation_alerts) do
+      add :baseline_established_at, :timestamptz
```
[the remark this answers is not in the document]
fortmarek: Documented in fbb7f81242. Added `baseline established timestamp` to the Alert rules line under "Alerts & Monitoring" in `server/data-export.md`.

server/priv/ingest_repo/migrations/20260505170000_create_test_case_run_daily_stats_per_case_mv.exs
```diff
@@ -0,0 +1,116 @@
+defmodule Tuist.IngestRepo.Migrations.CreateTestCaseRunDailyStatsPerCaseMv do
```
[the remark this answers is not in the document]
fortmarek: Documented in fbb7f81242. Added an entry for `test_case_run_daily_stats_per_case` under "Analytics Data (ClickHouse)" in `server/data-export.md`, alongside the existing per-day MV entries.

PR #10600, opened 2026-05-04
fix(server): surface failed_processing status on test run pages

fortmarek:
### Summary

When a test run's xcresult parsing failed, the run was stored with the `failed_processing` status, but the UI rendered nothing for that state — the **Test Runs** list showed a blank Status column and the **Test Run detail** page showed an empty Status metadata field. Only the existing red header icon hinted that something was wrong, and it was indistinguishable from a regular test failure.

This PR brings the test run pages in line with the build run pages:

- **Test Runs list** — adds a `Failed processing` warning badge in the Status column.
- **Status filter** — adds `failed_processing` as a selectable filter option.
- **Test Run detail** — adds the `Failed processing` badge to the Status metadata row, and switches the header icon from the red `badge-failure` / `alert_circle` to the `badge-warning` / `alert_hexagon` pair already used by the build run page so processing failures are visually distinct from real test failures.
- Re-extracts `dashboard_tests.pot` to register the new translatable string.

### How to test locally

1. Insert a `test_runs` row with `status = "failed_processing"` (e.g. via `iex` or by simulating a worker failure in `process_xcresult_worker`).
2. Open the **Test Runs** list — the row now shows a yellow `Failed processing` badge in the Status column.
3. Open the **Filter** dropdown and confirm the **Status** filter offers `Failed processing`.
4. Open the run detail page — the header shows the warning hexagon icon and the Status metadata field shows the `Failed processing` badge.

### DIFF
--- a/server/lib/tuist_web/live/test_run_live.html.heex
+++ b/server/lib/tuist_web/live/test_run_live.html.heex
@@ -34,9 +34,9 @@
             <.circle_dashed />
           </div>
         </div>
-        <div :if={@run.status == "failed_processing"} data-part="badge-failure">
+        <div :if={@run.status == "failed_processing"} data-part="badge-warning">
           <div data-part="icon">
-            <.alert_circle />
+            <.alert_hexagon />
           </div>
         </div>
         <div :if={@run.status == "skipped"} data-part="badge-warning">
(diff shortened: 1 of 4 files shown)


pepicrft (approved):
Looks good. One small non-blocking note: `server/priv/gettext/dashboard_tests.pot` still references `lib/tuist_web/live/test_runs_live.ex:45` for `Failed processing`, but the filter option was removed in the second commit. It would be worth regenerating the template so the source locations stay clean.

PR #10601, opened 2026-05-04
fix(server): align Quarantined and Flaky Tests counts with their lists

fortmarek:
### Summary

Two related drift bugs surfaced as off-by-one mismatches between an analytics card and the list below it.

#### Quarantined Tests

The analytics card (event-replay) and the list (`TestCase.state`) drifted on at least one project — some tests had a `muted` event with no corresponding `state = "muted"` row.

Root cause traced to the pre-[`a3e34233da`](https://github.com/tuist/tuist/commit/a3e34233da) `ActionExecutor.execute_actions/3`, which dispatched each automation attribute action through its own `update_test_case/3` call. Each call did a read-modify-write against `test_cases`; the second call's read missed the first's just-written row, so the second write reverted `state` while leaving the first call's `muted` event in place. Background ReplacingMergeTree merge then collapsed both rows in favour of the later `enabled` one. The event survived because `test_case_events` is append-only.

`a3e34233da` (already merged) coalesced the actions into one `update_test_case/3` call, which closes the race at its source. This PR locks in that invariant so a future refactor can't reintroduce it silently.

#### Flaky Tests

The Flaky Tests card counted purely off `test_case_events` (currently flagged), while `list_flaky_test_cases/3` `inner_join`-ed `flaky_test_case_runs` over the analytics window. A test that's currently flagged but has no flaky runs in the window — a low-frequency test auto-flagged a while ago that hasn't run since, or a test whose flaky runs sit in a different `:is_ci` segment — was silently dropped from the list while the card counted it.

The fix: switch the stats join to `left_join` and coalesce `flaky_runs_count` to `0`. The list now includes every currently-flagged test; the date / environment opts scope only the enrichment columns (`flaky_runs_count`, `last_flaky_at`), not membership.

### What this PR adds

**Quarantined-tests invariant guard** — three integration tests in `server/test/tuist/tests_test.exs` under `describe "automation flow ↔ TestCase.state ⁄ test_case_events consistency invariant"`:
1. `[change_state: muted, add_label: flaky]` through `ActionExecutor` — asserts `state`/`is_flaky` and the latest quarantine event agree.
2. Recovery direction (`[remove_label: flaky, change_state: enabled]`) — same assertion.
3. Hard contract: `expect(Tests, :update_test_case, 1, …)` against `update_test_case/2` — fails immediately if `ActionExecutor` ever issues more than one call per multi-action automation.

**Flaky-tests query fix** — `build_flaky_test_cases_query/3` and `build_flaky_test_cases_count_query/3` switched to `left_join` on `flaky_stats_subquery`, with `coalesce(stats.flaky_runs_count, 0)` in the select so stale-flagged tests render with zeroed stats instead of being filtered out.

Three existing tests (`filters by time range`, `filters by CI environment`, `combines time range and environment filters`) updated to reflect the new semantic — they now assert that currently-flagged tests appear regardless of date/env opts and that those opts only scope the stats columns.

Adds one new invariant test: `list_flaky_test_cases/3` `total_count` equals `Analytics.flaky_tests_analytics/2`'s `count` for the same opts, including a currently-flagged test with no flaky runs in window.

**Plus** `Tests.active_quarantine_states/0` helper used by both the list query and the new consistency invariant test, and a side update to the `tuist-elixir-review` Blick skill that adds N+1-query detection to the review checklist.

### Out of scope (follow-ups)

- One-off backfill for affected rows (handled separately).
- Optional drift detector — periodic Oban / mix task that runs the list-vs-events count delta per project.

### How to test locally

```
mix test test/tuist/tests_test.exs --only describe:"automation flow"
mix test test/tuist/tests_test.exs --only describe:"list_flaky_test_cases"
```


server/test/tuist/tests_test.exs
```diff
@@ -6827,6 +6828,116 @@ defmodule Tuist.TestsTest do
...
+      # contract from regressing without depending on a ClickHouse
+      # part-visibility race that's hard to reproduce in tests.
+      project = ProjectsFixtures.project_fixture()
+
+      test_case =
+        RunsFixtures.test_case_fixture(project_id: project.id, state: "enabled", is_flaky: false)
+
+      IngestRepo.insert_all(TestCase, [test_case |> Map.from_struct() |> Map.delete(:__meta__)])
+
+      expect(Tests, :update_test_case, 1, fn id, attrs, opts ->
```
pepicrft: Can we make this expectation match the call shape from `ActionExecutor`? It calls `Tests.update_test_case(id, attrs)`, so this expectation is attached to `update_test_case/3` and will not observe the production call. That leaves the coalescing guard failing at verification time. Using `fn id, attrs -> Mimic.call_original(Tests, :update_test_case, [id, attrs]) end` should cover the path we want.

pepicrft (approved):
Approved. Please fix the inline test expectation before merging.

server/test/tuist/tests_test.exs
fortmarek: Good catch — fixed in 083b6ef16a. `ActionExecutor` calls `Tests.update_test_case(id, attrs)` so the expect now matches `/2`. With this, a regression that re-splits dispatch fails on the second call as intended.

PR #10614, opened 2026-05-05
fix(cli): preserve whitespace in plist template for Array-root scalars

winston-riley-zocdoc:
## Summary

- Fix a whitespace bug in the plist resource synthesizer where Array-root plists with scalar elements (e.g. `[String]`) produce `publicstatic` instead of `public static` in generated Swift code
- The bug is in `PlistsTemplate.swift` line 24: Stencil's `.smart` trim eats the space between `{{accessModifier}}` and `static` when the `nonisolated(unsafe)` branch is skipped. Changing `{% if` to `{%+ if` preserves the preceding space.
- Add a test covering Array-root scalar plists, which was previously untested

Storied commit, first commit is a failing test for verification. 

I suspect the regression might have been introduced here: https://github.com/tuist/tuist/pull/10195 (Anecdotally, the date this was merged roughly lines up with when I noticed the bad syntax being generated).

## Test plan

- [x] New test `render_plistWithArrayOfScalars_producesValidSpacing` fails before the fix and passes after
- [x] All existing `SynthesizedResourceInterfacesGeneratorTests` continue to pass

### DIFF
--- a/cli/Sources/TuistGenerator/Templates/PlistsTemplate.swift
+++ b/cli/Sources/TuistGenerator/Templates/PlistsTemplate.swift
@@ -21,7 +21,7 @@ extension SynthesizedResourceInterfaceTemplates {
       {% set rootType %}{% call typeBlock document.metadata %}{% endset %}
       {%- set containsAny %}{%- call typesContainAny document.metadata -%}{% endset %}
       {% if document.metadata.type == "Array" %}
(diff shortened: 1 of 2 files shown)


pepicrft (approved):
Looks good to me. The template change is narrowly scoped, and the added regression test covers the missing Array-root scalar case that was previously untested.

fortmarek (approved):
Thanks! looks good

PR #10615, opened 2026-05-05
feat(server): cut bundles over to ClickHouse (phase 3+4)

fortmarek:
> Finishes the bundles PG → CH migration started in #10595 (phase 1) and continued in #10597 (phase 2). Mirrors the artifacts cutover in #10580.

## Summary

- `Bundle` is now a ClickHouse-backed Ecto schema; `BundleIngest` is folded back in and deleted.
- All bundle reads cut over to `ClickHouseRepo` / `ClickHouseFlop`; writes go through `IngestRepo.insert_all`.
- New IngestRepo migration drops the PG `bundles` table and the `replicated_to_ch` column.
- `data-export.md` updated to drop the dual-write/backfill note for bundles.

## Implementation notes

- Atom-shaped enum fields (`type`, `supported_platforms`) are decoded back to atoms at every read boundary so existing callers don't change.
- `inserted_at` writes are deliberately truncated to seconds at the encode boundary because the PG `:utc_datetime` column was second-precision; without truncation, second-truncated period filters (e.g. the dashboard date picker) miss freshly-created bundles.
- `delete_bundle!/1` runs `ALTER TABLE … DELETE … SETTINGS mutations_sync = 1` against both `bundles` and `artifacts` so the dashboard redirect cannot observe the deleted row.
- `Bundle.changeset/2` and the dedicated `bundle_test.exs` were dropped; the controller's OpenAPI schema and the inline raises in `Bundles.create_bundle/2` cover input validation.

### How to test locally

- [ ] `mix test test/tuist/bundles_test.exs test/tuist/bundles/ test/tuist_web/live/bundle_live_test.exs test/tuist_web/live/project_bundle_settings_live_test.exs test/tuist_web/live/bundles_live_test.exs test/tuist_web/controllers/api/bundle_artifacts_controller_test.exs test/tuist_web/controllers/api/bundles_controller_test.exs test/tuist/mcp/components/tools/bundle_tools_test.exs test/tuist/vcs_test.exs test/tuist/alerts_test.exs` — 201 tests, 0 failures locally.
- [ ] Boot the dashboard, upload a bundle, verify it appears in the bundles list, the size widgets render, and deletion redirects without leaving the row visible.


server/lib/tuist/bundles.ex
```diff
@@ -7,107 +7,140 @@ defmodule Tuist.Bundles do
...
+  # `DateTime64(6)` (read back as `NaiveDateTime`); the rest of the
+  # codebase works with atoms and `DateTime`, so normalize at every
+  # read boundary that returns a `%Bundle{}` struct.
+  defp decode_bundle(nil), do: nil
+
+  defp decode_bundle(%Bundle{} = bundle) do
+    %{
+      bundle
+      | type: decode_type(bundle.type),
+        supported_platforms: Enum.map(bundle.supported_platforms || [], &String.to_existing_atom/1),
```
pepicrft: Validated locally by inserting a bundle row directly into ClickHouse with a dynamically generated unknown platform string and then calling `Bundles.get_bundle/1`; this raises `ArgumentError` before the caller can handle the row. Since ClickHouse is now the read source for bundles, a malformed backfill, manual repair, or future enum mismatch would turn dashboard and API reads into 500s. Can we decode through the known `Bundle.supported_platforms()` values and return a controlled result, or preserve unknown strings, instead of calling `String.to_existing_atom/1` directly?

pepicrft (changes requested):
Requested changes. I validated the enum decode concern locally by inserting a malformed ClickHouse bundle row and calling `Bundles.get_bundle/1`; the focused PR suite otherwise passes after resetting the local test databases.

The remaining issue is that a bad `supported_platforms` value in ClickHouse raises `ArgumentError` at the read boundary and would turn bundle reads into 500s now that ClickHouse is the source of truth.

pepicrft (approved):

PR #10617, opened 2026-05-05
feat(status): public status page worker backed by Grafana IRM

pepicrft:
> Adds a public status page for Tuist services, served from a tiny Cloudflare Worker and powered by the Grafana Cloud Incident API.

The new `status/` project is a stateless Cloudflare Worker (Hono + TypeScript) that renders the public status page. Both the list of components and the incident timeline are sourced live from Grafana Cloud:

- `IncidentsService.QueryIncidents` for active and recently-resolved incidents (last 14 days), filtered client-side by close time.
- `FieldsService.GetFields` for the public component list — each select option of the configured label field (default `affected_service`) becomes one row on the page; the option's value is the component id matched against incident labels, the description becomes the subtitle.

The page is rendered with the actual Noora component CSS (`noora-card`, `noora-banner`, `noora-status-badge`, `noora-badge`, `noora-line-divider`) inlined verbatim from `noora/css/`, so it stays visually identical to the rest of the Tuist surface without depending on Phoenix or Tailwind. It also exposes RSS (`/feed.rss`) and Atom (`/feed.atom`) feeds with `<link rel="alternate">` autodiscovery, a JSON snapshot at `/api/status.json`, and two upstream-debug endpoints (`/api/debug/incidents.json`, `/api/debug/fields.json`).

In dev, `USE_FAKE_DATA=true` (the default) renders the page with fixtures so design work doesn't need a token. To run against real Grafana data, set `GRAFANA_INCIDENT_API_URL`, `GRAFANA_INCIDENT_API_TOKEN`, and (optionally) override `GRAFANA_COMPONENT_LABEL_KEY` in `status/.dev.vars`. See `status/AGENTS.md` for full configuration and operational details.

### How to test locally

```
cd status
mise install
aube install   # one-time
aube run dev
open http://localhost:8787
```

That serves the page from fake fixtures. To exercise the real Grafana integration:

1. Create a Grafana Cloud service account in your stack with Viewer role and generate a `glsa_…` token.
2. In Grafana → Alerts & IRM → Label management, create labels with key `affected_service` and one row per public component (value = display name, description = subtitle).
3. In `status/.dev.vars`:
   ```
   USE_FAKE_DATA="false"
   GRAFANA_INCIDENT_API_URL="https://<stack-slug>.grafana.net/api/plugins/grafana-irm-app/resources"
   GRAFANA_INCIDENT_API_TOKEN="glsa_…"
   ```
4. Restart `aube run dev`. Visit:
   - `/` — the rendered status page.
   - `/api/status.json` — normalized snapshot.
   - `/api/debug/fields.json` — raw `FieldsService.GetFields` response if components come back empty.
   - `/api/debug/incidents.json` — raw `IncidentsService.QueryIncidents` payloads.

Type-check and dry-run build before merging:
```
aube run typecheck
aube run build
```

image


status/wrangler.jsonc
```diff
@@ -0,0 +1,16 @@
...
+  "$schema": "node_modules/wrangler/config-schema.json",
+  "name": "tuist-status",
+  "main": "src/index.ts",
+  "compatibility_date": "2025-04-01",
+  "compatibility_flags": ["nodejs_compat"],
+  "observability": {
+    "enabled": true,
+  },
+  "vars": {
+    "USE_FAKE_DATA": "true",
```
fortmarek: [P1] Production will serve fake data

The deploy workflow runs `wrangler deploy` with this same config, and the dry-run build showed `env.USE_FAKE_DATA` as true. Because `loadSnapshot` returns fixtures whenever this value is true, the production worker will ignore the Grafana token and publish canned incidents until this is set false for production.

status/src/grafana-irm.ts
```diff
@@ -0,0 +1,341 @@
...
+function statusFrom(input: string | undefined): IncidentStatus {
+  if (!input) return "investigating";
+  return input.toLowerCase() === "resolved" ? "resolved" : "investigating";
+}
+
+function labelKey(l: RawLabel): string | undefined {
+  return l.label?.key ?? l.key;
+}
+
+function labelValue(l: RawLabel): string | undefined {
```
fortmarek, line 104: [P1] Incident labels will not match components

Grafana documents incident labels as `{ key, label }`, but `labelValue` only reads `{ label: { value } }` or `{ value }`. Real incidents labeled with `affected_service` can therefore produce empty `affectedComponents`, leaving component and overall status operational during an outage.

status/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+});
+
+// Returns the raw upstream response from Grafana Incident, with no field
+// renaming or coercion. Useful for verifying the integration end-to-end and
+// diagnosing field-name drift. Only enabled when not in fake-data mode.
+app.get("/api/debug/incidents.json", async (c) => {
+  if (c.env.USE_FAKE_DATA === "true") {
+    return c.json({ error: "USE_FAKE_DATA is true; nothing to debug" }, 400);
+  }
+  const raw = await fetchRawIncidents(c.env);
```
fortmarek, line 44: [P2] Debug routes expose raw Grafana data

These debug endpoints are public whenever fake-data mode is off. They return raw incidents and all label fields from Grafana, which can include internal metadata beyond the sanitized public status page; gate them behind a non-production/debug secret or remove them from production.

status/src/grafana-irm.ts
```diff
@@ -0,0 +1,341 @@
...
+  const out: RawIncident[] = [];
+  let cursor: string | undefined;
+  for (let page = 0; page < 5; page++) {
+    const body: Record<string, unknown> = { query: { queryString, limit, orderDirection: "DESC" } };
+    if (cursor) (body.query as Record<string, unknown>).cursor = cursor;
+    const res = await rpc<QueryIncidentsResponse>(opts, "IncidentsService.QueryIncidents", body);
+    const batch = res.incidents ?? [];
+    out.push(...batch);
+    if (!res.cursor?.hasMore) break;
+    cursor = res.cursor.nextValue;
```
fortmarek, line 236: [P2] Cursor is sent in the wrong request shape

Grafana expects `cursor` as a top-level sibling of `query`, but this puts it inside `query`. If a response has `hasMore`, the worker can repeatedly fetch the first page, duplicate incidents, and miss later pages.

fortmarek (changes requested):
Initial review findings. The two P1s are release blockers for the status page: production currently serves fake data, and real Grafana labels may not map incidents to components.

status/wrangler.jsonc
pepicrft: Done in 0c1bac6 (separate from this push). Default is now `false`; local dev opts back in via `.dev.vars`.

status/src/grafana-irm.ts
pepicrft: Good catch, thanks. Fixed in 222fab2821. Tolerating `{ key, label }` (the actual shape) plus the older nested forms, with a regression test using a real label payload.

status/src/index.ts
pepicrft: Gated both routes behind `ENABLE_DEBUG_ROUTES="true"` in 222fab2821. Production returns 404; set it in `.dev.vars` to use them locally.

status/src/grafana-irm.ts
pepicrft: Right, fixed in 222fab2821. Cursor now sits next to `query`. Added a pagination test that asserts the body shape on both pages.

pepicrft:
@fortmarek thanks for the review. All four addressed:

- P1 fake data: `USE_FAKE_DATA` flipped to `false` in 0c1bac6.
- P1 label shape: tolerating `{ key, label }` plus the older nested forms in 222fab28, with a regression test against a real-shape payload.
- P2 debug routes: now gated behind `ENABLE_DEBUG_ROUTES="true"`; production returns 404, locals can opt in via `.dev.vars`.
- P2 cursor: moved out of `query` to a top-level sibling, with a pagination test that asserts both page bodies.

fortmarek:
A note on `status/src/fake-data.ts`: I think the fixture exists for a good reason. It lets people run and design the status page locally without a Grafana IRM token, while still exercising degraded services, active incidents, resolved incidents, maintenance, and feed entries.

I would not scope it under tests only if the Worker intentionally supports `USE_FAKE_DATA=true` at runtime for local/dev previews. It is runtime dev/demo data, not just test data.

That said, I think it would be safer if the shape made that intent explicit: for example `src/dev/fake-status.ts` or `src/fixtures/status.ts`, with production defaulting to real data and local-only config enabling fake data via `.dev.vars` or docs. The main issue is not that fake data lives in source; it is that the deployable config currently turns it on.

fortmarek (approved):

pepicrft:
@fortmarek good shout — moved to `status/src/dev/fake-status.ts` in 21cafce. Path now reads as runtime dev/demo data rather than test fixtures.

PR #10629, opened 2026-05-05
chore(server): Serve OpenAI Apps challenge

pepicrft:
## Summary

- Add a hosted-only `/.well-known/openai-apps-challenge` endpoint for OpenAI Apps domain verification.
- Return the public OpenAI challenge token directly from code so no runtime secret rollout is needed.
- Cover both Tuist-hosted and on-premise behavior in controller tests.

## Validation

- `git diff --check`
- `mise exec -- mix deps.get && mise exec -- mix format lib/tuist_web/controllers/well_known_controller.ex lib/tuist_web/router.ex test/tuist_web/controllers/well_known_controller_test.exs`
- `mise exec -- mix test test/tuist_web/controllers/well_known_controller_test.exs` could not complete locally because `mix deps.get` updates the out-of-date lockfile to Phoenix 1.8.6, which then fails compiling existing `billing_plug.ex` with duplicate `use Phoenix.VerifiedRoutes`.

### DIFF
--- a/server/lib/tuist_web/controllers/well_known_controller.ex
+++ b/server/lib/tuist_web/controllers/well_known_controller.ex
@@ -7,6 +7,7 @@ defmodule TuistWeb.WellKnownController do
   alias TuistWeb.AgentSkillsDiscovery
   alias TuistWeb.RequestOrigin
 
+  @openai_apps_challenge_token "YoBqoSMoA-RuEX8RMuCKrLnPCXDYUsYtKg-yjFBHmDQ"
   @mcp_path "/mcp"
   @oauth_token_path "/oauth2/token"
   @oauth_authorize_path "/oauth2/authorize"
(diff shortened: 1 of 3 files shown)


server/lib/tuist_web/controllers/well_known_controller.ex
```diff
@@ -7,6 +7,7 @@ defmodule TuistWeb.WellKnownController do
   alias TuistWeb.AgentSkillsDiscovery
   alias TuistWeb.RequestOrigin
 
+  @openai_apps_challenge_token "YoBqoSMoA-RuEX8RMuCKrLnPCXDYUsYtKg-yjFBHmDQ"
```
pepicrft, line 10: This piece of information is public so I leaned on hardcoding it here.

fortmarek (approved):

PR #10630, opened 2026-05-05
fix(server): remove former team member from about page

pepicrft:
## What changed

Some updates in our about page to reflect recent team changes.

## Validation

- Ran `git diff --check`
- Tried `mix format lib/tuist_web/marketing/controllers/marketing_html/about.html.heex`, but this worktree is missing server dependencies and Mix reported `Unknown dependency :ecto_sql given to :import_deps`


server/lib/tuist_web/marketing/controllers/marketing_html/about.html.heex
```diff
@@ -40,10 +40,7 @@
         </p>
       </article>
     </section>
-    <figure data-part="photos">
```
pepicrft, line 43: Let's bring this back later in the year.

PR #10632, opened 2026-05-05
fix(server): restore PG bundles table during phase-4 rollout

fortmarek:
## Summary

- The phase-4 migration (`20260505120000`) dropped the PostgreSQL `bundles` table as part of cutting reads to ClickHouse
- Pods still running pre-phase-4 code call `Repo.one()` (PostgreSQL) inside `Bundles.last_project_bundle/2`, causing `Postgrex.Error: relation "bundles" does not exist` (Sentry TUIST-EQ, 98 occurrences, escalating)
- This adds a new IngestRepo migration (`20260505160000`) that recreates an empty `bundles` table — old pods get `nil` results which they already handle gracefully, new pods use `ClickHouseRepo` and never touch the PG table

## Test plan

- [ ] Verify the migration runs cleanly: `mix ecto.migrate` completes without errors
- [ ] Confirm Sentry issue TUIST-EQ stops escalating after deployment

## Follow-up

Once the phase-4 rollout is complete and all pre-phase-4 pods are gone, the restored table and this migration can be removed with a cleanup PR.

### DIFF
--- a/server/priv/ingest_repo/migrations/20260505160000_restore_pg_bundles_for_rollout.exs
+++ b/server/priv/ingest_repo/migrations/20260505160000_restore_pg_bundles_for_rollout.exs
@@ -1,0 +1,69 @@
+defmodule Tuist.IngestRepo.Migrations.RestorePgBundlesForRollout do
+  @moduledoc """
+  Temporarily restores the PostgreSQL `bundles` table so that server pods still
+  running pre-phase-4 code can complete their Repo queries without crashing.
+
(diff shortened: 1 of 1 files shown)


fortmarek:
Closing — root cause is zombie pods on the legacy Syself cluster, not the new self-hosted CAPI cluster. Resurrecting the PG `bundles` table on the shared Supabase would let pre-phase-4 pods on Syself write stale rows the new cluster never reads. Will instead wind down the Syself workload cluster (except atlas nodes).

PR #10637, opened 2026-05-06
fix(cli): duplicate selective testing log messages and improve skip reason clarity

irena327:
## Purpose

  Fixes two issues with `tuist test` selective testing log output:

  1. **Duplicate log messages**: `shouldRunTest` was called twice in the test execution flow — once in
   the outer orchestration function and again inside `testSchemes()`. Both calls emitted the same
  "will be skipped" and "Testing the following targets" notices, resulting in duplicate output.

  2. **Misleading skip reason**: All skipped targets were reported as "have not changed since the last
   successful run" regardless of whether they were actually cached or simply not included in the
  `--test-targets` filter. A target not in `--test-targets` would appear in the skip message even
  though it was never requested.

  ### Changes

  - Removed the redundant `shouldRunTest` call from `testSchemes()`, replaced with a simple `guard
  !testSchemes.isEmpty` check
  - Moved `uploadSkippedTestSummary` handling to the single remaining call site
  - Added `requestedTestTargets` parameter to scope the skip message to only targets the user actually
   requested
  - When `--test-targets` is passed, the "skipped" list + "testing" list now adds up to exactly what
  was requested

  ### Before
  The following targets have not changed since the last successful run and will be skipped: <every
  target in the scheme not being tested, including ones never requested>
  Testing the following targets: A, B, C
  The following targets have not changed since the last successful run and will be skipped:
  Testing the following targets: A, B, C

  ### After
  The following targets have not changed since the last successful run and will be skipped: D, E
  Testing the following targets: A, B, C

  Where `--test-targets A B C D E` was passed and D, E had cache hits.

  1. Run `tuist test --test-targets TargetA TargetB TargetC` on a project with selective testing
  enabled
  2. Verify log messages appear only once (no duplicates)
  3. Verify the "will be skipped" list only contains targets from your `--test-targets` input that
  were cached, not unrelated targets from the scheme
  4. Verify that the skipped list + testing list = your `--test-targets` input
  5. Run `tuist test` without `--test-targets` and verify all cached scheme targets are reported as
  skipped

3 new tests:                                                                                                          
                                                            
  1. `test_run_tests_when_part_is_cached_only_logs_requested_targets_as_skipped` — Verifies that when --test-targets      
  TargetA TargetB is passed but TargetB and TargetC are both cached, only TargetB (which was requested) appears in the  
  "will be skipped" message. TargetC should not appear.                                                                 
  2. `test_run_tests_when_part_is_cached_does_not_log_duplicate_messages` — Verifies that "Testing the following targets" 
  appears exactly once in the log output (the original bug was it appeared twice).
  3. `test_run_tests_when_no_test_targets_filter_logs_all_cached_as_skipped` — Verifies that when no --test-targets is
  passed, all cached targets in the scheme (TargetB, TargetC) appear in the skip message.

### DIFF
--- a/cli/Sources/TuistKit/Services/TestService.swift
+++ b/cli/Sources/TuistKit/Services/TestService.swift
@@ -488,9 +488,19 @@ public struct TestService { // swiftlint:disable:this type_body_length
             mapperEnvironment: mapperEnvironment,
             graph: graph,
             action: action,
+            requestedTestTargets: testTargets,
             passthroughXcodeBuildArguments: passthroughXcodeBuildArguments
-        ), action == .build {
-            try await outputEmptyShardMatrixIfNeeded(isSharding: isSharding, action: action)
+        ) {
+            if action == .build {
+                try await outputEmptyShardMatrixIfNeeded(isSharding: isSharding, action: action)
+            } else {
+                let timer = clock.startTimer()
+                try await uploadSkippedTestSummary(
+                    schemeName: schemes.first?.name,
+                    config: config,
+                    timer: timer
+                )
+            }
             return
         }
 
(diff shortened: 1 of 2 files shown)


fortmarek (approved):
Looks great, thanks!

fortmarek:
Thanks again for the PR! Confirmed the failures are flakes, so going ahead and merging.

PR #10651, opened 2026-05-06
fix(server): restore main CI checks

fortmarek:
## Summary
- use LiveViewTest.render_async/2 in BuildsLiveTest so async assigns settle through the first-party test API
- add missing MCP tool metadata for update_test_case so production Docker compile passes with warnings-as-errors
- invalidate release .build caches when Package.swift changes and use macOS-compatible sed for the xcresult image pin bump

## Tests
- MIX_ENV=test TUIST_SERVER_TEST_POSTGRES_DB=tuist_test_codex_7250_green TUIST_SERVER_TEST_CLICKHOUSE_DB=tuist_test_codex_7250_green mix ecto.reset
- MIX_ENV=test TUIST_SERVER_TEST_POSTGRES_DB=tuist_test_codex_7250_green TUIST_SERVER_TEST_CLICKHOUSE_DB=tuist_test_codex_7250_green mix test test/tuist_web/live/builds_live_test.exs --warnings-as-errors
- MIX_ENV=test mix compile --warnings-as-errors
- git diff --check
- ruby -e "require 'yaml'; YAML.load_file('.github/workflows/release.yml')"
- tested the xcresult image-pin sed against a temp copy of infra/helm/tuist/values-managed-common.yaml

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -785,7 +785,7 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           path: .build
-          key: ${{ runner.os }}-cli-release-${{ hashFiles('.xcode-version-releases') }}-${{ hashFiles('Package.resolved') }}
+          key: ${{ runner.os }}-cli-release-${{ hashFiles('.xcode-version-releases') }}-${{ hashFiles('Package.swift') }}-${{ hashFiles('Package.resolved') }}
       - uses: jdx/mise-action@v4.0.1
         with:
           version: ${{ env.MISE_VERSION }}
@@ -884,7 +884,7 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           path: .build
-          key: linux-${{ matrix.arch }}-cli-release-${{ hashFiles('Package.resolved') }}
+          key: linux-${{ matrix.arch }}-cli-release-${{ hashFiles('Package.swift') }}-${{ hashFiles('Package.resolved') }}
 
       - name: Update version in Constants.swift
         run: |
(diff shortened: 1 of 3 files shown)


pepicrft (approved):

pepicrft:
Can we instead render the view doing the work synchronously? This can get flaky very quickly.

fortmarek:
Addressed this by removing the longer `render_async/2` timeout and letting BuildsLive tests opt into synchronous async-assign resolution instead. The synchronous path is gated by test config, so normal runtime keeps using `assign_async/3`.

Re-ran:
- `MIX_ENV=test TUIST_SERVER_TEST_POSTGRES_DB=tuist_test_codex_7250_verify TUIST_SERVER_TEST_CLICKHOUSE_DB=tuist_test_codex_7250_verify mix test test/tuist_web/live/builds_live_test.exs:73 --warnings-as-errors`
- `MIX_ENV=test TUIST_SERVER_TEST_POSTGRES_DB=tuist_test_codex_7250_verify TUIST_SERVER_TEST_CLICKHOUSE_DB=tuist_test_codex_7250_verify mix test test/tuist_web/live/builds_live_test.exs --warnings-as-errors`

fortmarek:
On the timeout: I switched these tests back to LiveViewTest.render_async/2 so we stay on LiveView's first-party async-assign test path instead of carrying test-only app/session logic. render_async/2 still needs a timeout, and these builds widgets cross analytics-backed async work that can be slower on CI. I don't see a deterministic first-party drain for that whole chain yet, so the explicit 1s timeout is the pragmatic guard for now. If we later expose a deterministic drain or remove the nested async dependency, we can drop the timeout.

PR #10653, opened 2026-05-06
feat: tuist runners — shared warm pool with dispatch-time binding

fortmarek:
## Summary

Customer-facing GitHub Actions runners on Tuist's Mac mini fleet, layered on top of the substrate that landed in [#10499](https://github.com/tuist/tuist/pull/10499). One cluster, N image pools sharing the same Mac mini host fleet, one ClickHouse-backed lifecycle table (`runner_jobs`) carrying every workflow_job from queue through completion, per-customer concurrency caps. Customer config collapsed into a single `accounts.runner_max_concurrent` column.

The PR walked through several architectural arcs to land at the current shape:

1. **End-to-end substrate.** Pre-bound + Burst Pod creation driven from the BEAM (Pod-watch driven), `runner_assignments` table for dispatch state, hardcoded `PoolConfig`.
2. **Variant C refactor.** Replaced the BEAM reconciler / watcher with a standalone Go `runners-controller` (controller-runtime, CRDs). Dropped the `runner_assignments` table; dispatch endpoint auth switched from per-Pod random tokens to **projected ServiceAccount tokens validated via TokenReview**.
3. **SharedWarm / no-cold-path collapse.** With one VM per host (v1 design choice — Apple's SLA permits up to two), SharedWarm size = host count, and at that setting every Pod is either polling or running a customer job. The cold-path Burst Pod creation is unreachable; customer-side `RunnerPool` CRs are unneeded; the controller manages Pods directly via owner refs.
4. **Customer config collapsed.** "Subscriptions" with `owner` / `labels` / `runner_group_id` / `allowed_repos` was bloat. For v1 a single `accounts.runner_max_concurrent` integer is the whole knob (`0` = runners disabled). Owner is the account's `name` (= GitHub org login by convention). Repo allowlisting falls back to GitHub's default runner group; per-repo scoping is a follow-up when multi-tenant onboarding demands it.
5. **Multi-image runner pools, not multi-image schema migration.** `RunnerPool.spec.dispatchLabel` is the customer-facing routing key (the label customers put in `runs-on`); helm `runnersFleet.pools[]` is a list, one entry per image. v1 ships a single `default` pool. Adding Xcode-15 compatibility or a future per-OS variant is a new `pools[]` entry, not a values-schema change or controller refactor — the lifecycle table, controller, Pod-side SA label, and dispatch endpoint all already key on the pool name.
6. **Pre-order, don't auto-order Mac minis.** Scaleway Apple Silicon stock is unreliable; auto-`CreateServer` on a `ScalewayAppleSiliconMachine` regularly fails with `resource server is out of stock`, and Apple's 24h licensing floor makes speculative ordering expensive. The CAPI provider now supports `spec.adoptPoolPrefix`: when set, the controller skips `CreateServer` and instead claims a pre-ordered host whose Scaleway-side name starts with the prefix. The rename to the Machine's name IS the claim — the prefix is gone, so the next reconcile won't double-claim. No auto-order fallback; if the pool is empty the Machine sits in `Provisioning` with a `NoAvailableHost` event until the operator pre-orders more capacity. Staging / canary / production all wire this to `tuist-pool-`.
7. **Two-store split: Postgres for the claim lock, ClickHouse for the lifecycle view.** ClickHouse `runner_jobs` (ReplacingMergeTree, keyed on `workflow_job_id`) is the customer-facing surface — what's queued, claimed, running, recently completed. Postgres `runner_claims` is a five-column thin table — one row per currently-claimed workflow_job — that gives us the two OLTP primitives ClickHouse can't provide across replicas: atomic claim (`INSERT … ON CONFLICT (workflow_job_id) DO NOTHING RETURNING …`) and per-account cap counting (`SELECT account_id, count(*) … GROUP BY account_id` against an index). The two stores stay in sync via a one-way protocol: every state transition that flips the PG "is this claimed?" bit pairs with a CH INSERT for visibility. CH is never queried for OLTP correctness; PG is never read for customer surfaces.

## Architecture

```
                    GitHub workflow_job: queued
                              │
                              ▼
            ┌─────────────────────────────────┐
            │ Tuist server (webhook handler)  │
            │  • match owner → account        │
            │  • check runner_max_concurrent  │
            │    > 0 (else :ignored)          │
            │  • LIST RunnerPool CRs, find    │
            │    one whose spec.dispatchLabel │
            │    ∈ workflow_job.labels        │
            │  • INSERT runner_jobs           │
            │    (CH, status='queued')        │
            └────────┬────────────────────────┘
                     │
                     ▼
            ┌──────────────────────────────────┐
            │ ClickHouse runner_jobs (RMT)     │
            │ keyed on workflow_job_id;        │
            │ status ∈ {queued, claimed,       │
            │   running, completed};           │
            │ customer-facing view + history.  │
            └──────────────┬───────────────────┘
                           │
   ┌───────────────────────┘
   │  warm Pod polls /api/internal/runners/dispatch with SA Bearer:
   │   1. TokenReview validates SA → ns/name
   │   2. resolve SA → fleet (from tuist.dev/runner-pool label)
   │   3. Claims.counts_per_account (PG GROUP BY) → ineligible
   │      accounts (those at runner_max_concurrent)
   │   4. Jobs.pick_queued (CH SELECT FINAL, oldest queued not in
   │      ineligible) → candidate
   │   5. Claims.attempt (PG INSERT … ON CONFLICT DO NOTHING)
   │      → {:ok, claim} or {:error, :lost_race}.
   │      Atomic by PK; loser bails (no mint, no labels).
   │   6. Jobs.record_claimed (CH state for customer visibility)
   │   7. PATCH polling Pod labels: + tuist.dev/runner-pool-owner=<account.name>
   │   8. fetch pool's dispatchLabel, mint JIT under account.name
   │   9. Jobs.record_running (CH state) and return 200 + JIT
   └─────────────────────────────────┐
                                     ▼
            Pod execs ./run.sh --jitconfig,
            registers under customer for one
            job, exits, VM halts.
                     │
                     ▼
            Pod terminates → RunnerPoolReconciler (Go) reaps
            the terminal Pod + sibling SA and boots a
            replacement (Pod + SA owned by the pool via
            owner ref; no RunnerAssignment intermediate)

      GitHub workflow_job: completed
                     │
                     ▼
            ┌──────────────────────────────────┐
            │ Webhook handler:                 │
            │  • Claims.delete on workflow_job │
            │    (frees the PG cap slot)       │
            │  • Jobs.complete (CH state with  │
            │    conclusion + completed_at)    │
            └──────────────────────────────────┘
```

## How requirements are satisfied

- **One host fleet, N image pools.** Helm renders one `RunnerPool` CR per entry in `runnersFleet.pools`. All pools share the same CAPI MachineDeployment (one Mac mini fleet) via `fleetSelector`; Pods bin-pack across hosts. v1 ships a single `default` pool.
- **Customer-facing image routing.** Each pool's `spec.dispatchLabel` is what customers put in `runs-on`. Webhook handler matches against every pool's label; dispatch endpoint includes the matched label in the JIT mint so GitHub binds the workflow_job to the right runner.
- **Per-customer max concurrency.** `accounts.runner_max_concurrent` (0 = disabled, N>0 = cap, across all pools the customer reaches). Enforced at *claim* time via an indexed Postgres `GROUP BY account_id` against `runner_claims`, so capped customers' queued work waits its turn instead of being dropped on GitHub's side.
- **Enqueue when busy, execute when a VM frees.** ClickHouse `runner_jobs` accumulates rows in `status='queued'`. When a Pod terminates, the controller reaps + boots a replacement on the freed host (~30-90 s clone+boot); the replacement's first poll claims the oldest eligible row.
- **Fair across customers.** `ORDER BY (enqueued_at ASC, workflow_job_id ASC)` is deterministic: two pollers always select the same candidate, then collapse on Postgres's `INSERT … ON CONFLICT DO NOTHING`. A capped customer's backlog is excluded by the ineligible-accounts skip-set; other customers stay unblocked.
- **Customer-visible state.** The same CH `runner_jobs` table powers the live UI surfaces (queued, running) and the historical "recent runs" surface. `workflow_job.completed` webhooks transition the row to `status='completed'` with the GitHub conclusion and delete the corresponding PG claim row to free the cap slot.
- **Crash-safe claims.** `StaleClaimsWorker` (cron, every minute, 5-min threshold) deletes PG claims older than the threshold and re-INSERTs a `queued` state row into CH so the next poll can re-claim. Mint failure inside `serve_claim` calls `Claims.release/2` directly to put the row back without waiting for the worker — the `(workflow_job_id, claimed_at)` tuple guards against a stale serve stomping a re-claim.

## Trust chain (SA-token auth)

No `runner_assignments` table, no dispatch tokens persisted, no per-Pod state in the BEAM. Per-customer config is a single int on `accounts` (Postgres); claim coordination is a thin five-column `runner_claims` table (Postgres); workflow_job lifecycle rows live in `runner_jobs` (ClickHouse); Pod state lives in K8s.

1. Controller mints SA per Pod. Each Pod owns a unique `ServiceAccount` carrying `tuist.dev/runner-pool=<fleet-name>` as a label.
2. tart-kubelet projects the SA token into the VM via TokenRequest; staged at `/etc/tuist-sa-token`.
3. VM POSTs to `/api/internal/runners/dispatch` with `Authorization: Bearer <sa-token>`.
4. Server validates via TokenReview → SA namespace + name.
5. Server resolves SA → fleet (read the pool label off the SA), GETs the matching `RunnerPool` CR for its `spec.dispatchLabel`.
6. Server claims a queue row for that fleet, looks up the account from the row, stamps the Pod with owner labels, mints a JIT under the account's name with the pool's `dispatchLabel`.

### `infra/runners-controller/` (Go)

- `RunnerPool` CRD: `spec.image`, `spec.fleetSelector`, `spec.dispatchLabel`, `spec.replicas`, `spec.podCPUMilli`, `spec.podMemoryMB`. That's the entire schema.
- `RunnerPoolReconciler` watches `RunnerPool` and its owned Pods. Maintains `spec.replicas` Pods + per-Pod SAs as direct owner-ref children; reaps terminal Pods + sibling SAs before booting replacements (siblings under the pool, so neither cascades the other).
- No `RunnerAssignment` CRD, no `PodGCReconciler`, no claim-annotation machinery.

### Helm chart

- `runner-pool.yaml` ranges over `runnersFleet.pools[]`, renders one `RunnerPool` CR per entry. v1 ships a single `default` pool; adding a second image is an append, not a schema change. Empty pools list with `runnersFleet.enabled: true` is a render-time error.
- `runnersFleet.hostCount` is the CAPI MachineDeployment replica count; sum of `pools[*].replicas` must fit under it at one VM per host.
- `runners-controller-crds.yaml` ships only the `RunnerPool` CRD (now with `dispatchLabel` required).
- `runners-namespace.yaml`: server SA gets `runnerpools: get,list,watch`, `serviceaccounts: get,list`, `pods: get,list,patch`. Cluster-scoped: `tokenreviews: create`, `nodes: get,list`. NetworkPolicy gates the Pod network only — Tart VM egress bypasses CNI (vmnet) and lives at the cluster perimeter; comment in template makes the gap explicit.
- `capi.enabled` gate so the chart can land on clusters where CAPI core isn't yet installed.

### Server (Elixir)

- `accounts.runner_max_concurrent` (integer, default 0) on Postgres. The only per-customer knob.
- `Tuist.Runners.Claims` + `Tuist.Runners.Claim` — Postgres `runner_claims` (five columns: `workflow_job_id` PK, `account_id`, `fleet_name`, `pod_name`, `claimed_at`). The OLTP claim-lock + cap-counter:
  - `attempt/4` — atomic `INSERT … ON CONFLICT (workflow_job_id) DO NOTHING RETURNING …`. Returns `{:ok, %Claim{}}` or `{:error, :lost_race}`. No verify-readback; the unique PK does the work.
  - `release/2` — `DELETE … WHERE workflow_job_id = ? AND claimed_at = ?`. The captured `claimed_at` guards against a stale serve stomping a re-claim.
  - `counts_per_account/1` — indexed `GROUP BY account_id`, returns `%{account_id => inflight}`. Powers the cap check.
  - `release_stale/1` — DELETEs claims older than the threshold, returns the released rows for re-surfacing in CH.
- `Tuist.Runners.Jobs` + `Tuist.Runners.Job` — ClickHouse `runner_jobs` (ReplacingMergeTree) lifecycle table:
  - `enqueue/1` — INSERTs the initial `queued` row (idempotent via RMT).
  - `pick_queued/2` — deterministic candidate selection (`ORDER BY enqueued_at, workflow_job_id`) excluding ineligible accounts.
  - `record_claimed/3`, `record_running/2`, `record_queued/1` — state transitions paired with the corresponding PG write so the customer-visible state stays in sync.
  - `complete/2` — terminal state from the `workflow_job.completed` webhook.
  - `status_counts/1` — aggregation for customer dashboards.
- `Tuist.Runners.Workers.StaleClaimsWorker` — cron worker. `Claims.release_stale` deletes PG rows older than 5 min; for each released claim, `Jobs.record_queued` re-INSERTs a `queued` state row in CH so the next poll can re-claim.
- `Tuist.Runners.Dispatch.handle_webhook/2` — branches on `action`. `queued` extracts the GitHub workflow_job metadata and calls `Jobs.enqueue/1`. `completed` matches by `workflow_job_id` and calls `Jobs.complete/2`. Pool matching matches `workflow_job.labels` against every `RunnerPool.spec.dispatchLabel` so the row is enqueued under the correct fleet.
- `Tuist.Runners.Dispatch.dispatch_label_for_pool/1` — inverse lookup at JIT-mint time so the runner registers with the label customers used in `runs-on`.
- `Tuist.Runners.dispatch_for_sa/2` — TokenReview, SA → fleet, ineligible accounts from `Claims.counts_per_account/1` + `accounts.runner_max_concurrent`, `Jobs.pick_queued/2` → candidate, `Claims.attempt/4` → atomic claim, `Jobs.record_claimed/3` for visibility, stamp Pod labels (operational visibility only — cap accounting reads PG), mint JIT, `Jobs.record_running/2`. Bails cleanly on `:lost_race`.
- `TuistWeb.RunnersController.dispatch` — POST `/api/internal/runners/dispatch`. Bearer-token extract → TokenReview → `dispatch_for_sa` → JSON-with-JIT or 204 No Content.
- `Tuist.Kubernetes.Client` surface: `create_token_review`, `get_service_account`, `get_runner_pool`, `list_runner_pools`, `patch_pod`. K8s `list_pods` is no longer in the dispatch hot path.

### tart-kubelet

- `internal/satoken` package mints projected SA tokens via TokenRequest API.
- Reconciler stages the token at `<envDir>/sa_token` when `Pod.Spec.AutomountServiceAccountToken: true`.
- RBAC gains `serviceaccounts: get,list,watch` + `serviceaccounts/token: create`.
- `--node-labels` flag for arbitrary node labels.

### Runner image

- `dispatch-poll.sh` POSTs with the projected SA token as `Authorization: Bearer`. Handles 200 (run job), 204 (warm-standby sleep + retry), 401/403 (abort + halt VM). `EXIT` trap halts the VM on every code path. `curl` is invoked without `-f` so 4xx/5xx HTTP statuses reach the case branches instead of being shadowed by transport-failure fallback.
- `inject-env.sh` copies the SA token from `/Volumes/My Shared Files/env/sa_token` to `/etc/tuist-sa-token`.
- `--disableupdate` on `./run.sh`; Renovate auto-bumps the actions/runner version baked into the image.

### Release pipeline

- New `release-runners-controller` job in `.github/workflows/release.yml`, mirroring the existing `release-capi-scaleway` shape.

## Pre-ordering Mac mini capacity

Mac mini stock is unreliable on Scaleway and Apple's 24h licensing floor makes speculative auto-ordering expensive. The CAPI provider gates this behind a per-Machine field:

* `ScalewayAppleSiliconMachine.spec.adoptPoolPrefix` (empty → legacy `CreateServer` auto-order; set → claim a pre-ordered host).
* Helm: `runnersFleet.adoptPoolPrefix` flows into the runners-fleet MachineTemplate. `macosFleet` stays on the auto-order path for now (xcresult-processor capacity is small and stable, so the operational complexity of pre-staging isn't worth it there yet).

Operator workflow when `adoptPoolPrefix` is set:

1. Order N Mac minis in the Scaleway console matching `machine.type` / `machine.zone` / `machine.os`. Name each one with the configured prefix (e.g. `tuist-pool-001`, `tuist-pool-fr-par-1-a`) — only the prefix is matched; the suffix is free-form.
2. Scale the MachineDeployment (`helm upgrade` with a bumped `hostCount`, or `kubectl scale`). The controller picks the first matching pre-ordered host and renames it to the Machine's name — the rename atomically removes the prefix (= the host is no longer in the pool) and makes future reconciles converge via the existing name-based idempotency path.
3. If no compatible host is available the Machine stays in `Provisioning` with a `NoAvailableHost` event so the operator sees the queue.

### Chart defaults

* `runnersFleet.enabled: false` and `macosFleet.enabled: false` are the chart-level defaults. The tuist chart is self-hostable, and most self-host operators don't have a Scaleway project + Mac minis to provision — keeping these off by default lets `helm install tuist` land a working server without templating CAPI MachineDeployments / Apple Silicon CRs / a tuist-runners namespace they have no hardware to back. Tuist-managed envs (staging / canary / production) each flip these on in their per-env values file.
* `runnersFleet.adoptPoolPrefix` defaults to empty (legacy auto-order) so a greenfield bring-up with no pre-staged hardware still works. All three Tuist-managed env values files override to `tuist-pool-`, the convention we use for pre-ordered Mac minis in the Scaleway console.

## Adding a second image (future)

The plumbing for a second runner image is already in place. To add one:

1. Build a new runner image with whatever delta (Xcode version, base OS, toolchain bundle).
2. Append an entry to `runnersFleet.pools` in the relevant per-env values file:
   ```yaml
   pools:
     - name: default
       replicas: 4
       runnerImage: "ghcr.io/tuist/tuist-runner@sha256:…"
       dispatchLabel: "tuist-macos"
     - name: xcode-15
       replicas: 1
       runnerImage: "ghcr.io/tuist/tuist-runner-xcode-15@sha256:…"
       dispatchLabel: "tuist-macos-xcode-15"
   ```
3. Customers route by adding the new dispatch label to their `runs-on`. No server, controller, or schema change.

Sum of `pools[*].replicas` must fit under `runnersFleet.hostCount` while we're at one VM per host. The chart doesn't enforce that arithmetically (intentional — overcommit shows up as unscheduled Pods rather than a deploy failure).

## Test plan

- [x] `go build ./...` + `go vet ./...` + `go test ./...` clean across `infra/runners-controller`, `infra/tart-kubelet`, and `infra/cluster-api-provider-scaleway-applesilicon`.
- [x] `mix compile --warnings-as-errors --no-deps-check` clean.
- [x] `mix test test/tuist/runners` — 19 tests against the Postgres + ClickHouse sandboxes. Claims: atomic claim, lost-race, release with stale-handle guard, counts_per_account, stale-release recovery. Jobs: enqueue idempotency, pick_queued ordering + ineligibility, record_claimed / record_running / record_queued state transitions, complete with conclusion, status_counts.
- [x] `mix credo` clean on touched files.
- [x] `mix excellent_migrations.check_safety` — runner migrations clean.
- [x] `helm lint` clean; `helm template` renders one `RunnerPool` CR per `pools[]` entry + the controller Deployment + the namespace RBAC.
- [x] `helm template` with `capi.enabled=false` produces zero `Cluster` / `MachineDeployment` / `ScalewayAppleSilicon*` objects.
- [x] `helm template` with `runnersFleet.enabled=true` but empty `pools` fails the render with a clear message.
- [x] **End-to-end staging smoke (Phase 1 substrate)** — [run #25555333979](https://github.com/tuist/tuist/actions/runs/25555333979): pre-bound `tuist-runner-*` exited cleanly after one job, replacement minted automatically.
- [ ] Repeat the staging smoke against the two-store shape end-to-end (webhook → PG claim → mint → CH state transitions → completion webhook).
- [ ] Negative network check from inside the VM (cluster API unreachable).
- [ ] Canary rollout: install CAPI core, flip `capi.enabled=true` + `macosFleet.enabled=true` + `runnersFleet.enabled=true`, set `accounts.runner_max_concurrent` on the tuist account, smoke.
- [ ] Production same playbook.


fortmarek:
## ⚠️ Pre-merge action required: GitHub App permission grant

After the org-scoped JIT-config switch (commit 157676bc4f), the runner reconciler + dispatch handler call `POST /orgs/{org}/actions/runners/generate-jitconfig`. That endpoint requires the **Self-hosted runners: Read and write** organization permission on the corresponding Tuist GitHub App.

Status across envs:

| App | Permission | Action needed |
|---|---|---|
| Tuist - Staging | ✅ already granted | none |
| Tuist - Canary | ❌ not granted | grant + org admin approves installation update |
| Tuist (Production) | ❌ not granted | grant + org admin approves installation update |

The grant is at https://github.com/settings/apps/<app-slug> → **Permissions & events** → **Organization permissions** → **Self-hosted runners** → set to **Read and write**, then approve the installation update from the org's settings page.

Without the grant, the staging deploy works in shared-pool mode (warm Pods running and polling the dispatch endpoint, which we've verified end-to-end), but no JIT configs can be minted, so no runner ever registers with GitHub and queued workflow_jobs sit in GH's queue.

The earlier `administration: write` requirement (a much broader permission — covers settings, secrets, collaborators, webhooks, deletion) is intentionally avoided in this PR by registering runners at the org level instead of the repo level. Same dispatch flow, same runner labels; just a more targeted permission scope that matches what ARC, Namespace, and BuildJet ship with by default.

infra/helm/tuist/values-managed-common.yaml
```diff
@@ -208,7 +208,7 @@ macosFleet:
     # release.yml has run for this scope, pin to a known-good
     # SHA-tagged build that the standalone
     # capi-provider-scaleway-applesilicon-image.yml workflow pushed.
-    tag: "0.1.0"
+    tag: "56858800"
```
fortmarek: Reviewer note: this SHA-tagged build (`56858800`) is the one carrying the CAPI MachineDeployment integration on this branch. After the PR merges and the next `feat(capi-scaleway)` / `fix(capi-scaleway)` lands on `main`, the release workflow will auto-bump this to a proper `0.x.y` semver tag and rewrite the line. Pinning to a SHA was the standalone `capi-provider-scaleway-applesilicon-image.yml` workflow stop-gap; the inline comment above already explains the rolling-forward path.

fortmarek:
**Procurement + scaling note: 1 VM/host today, two viable paths at customer scale**

Recording today's architecture/procurement conversation.

**Today (internal use): 10× M2-L on Scaleway `fr-par-1`, 1 VM/host.** Pre-buying for tuist/tuist's own CI workload. Internal-only, predictable build times worth more than packing density.

**At customer scale, two viable fleet shapes. Decision lands at external launch based on product positioning + early-customer telemetry.**

### Path A: Mixed fleet, 1 VM/host (dedicated)

- **Small tier**: M2-L (M2 Pro 10C / 16 GB / 512 GB at €0.21/h). Spec-comparable to Namespace's small (M4 Pro 6 vCPU / 14 GB at $0.06/min). 16 GB sits below Blacksmith's 24 GB but above Namespace's 14 GB.
- **Large tier**: M4-XL (M4 Pro 14C / 64 GB / 2 TB at €0.49/h). Matches and slightly beats Namespace large (M2 Pro 12 vCPU / 56 GB at $0.18/min) and Blacksmith 12vcpu (M4 12 vCPU / 48 GB at $0.16/min). No Scaleway SKU sits between M4-M and M4-XL, so M4-XL is forced for a market-spec large tier.
- **Both tiers run 1 VM per host (dedicated).** Deliberate differentiator vs Namespace/Blacksmith/WarpBuild, who all pack 2 VMs/host and accept ~7.5% expected build-time variance from L2/L3 cache and memory bandwidth contention. "Every job runs alone on its host" is a real marketing line.
- **Fleet sizing for 500k small + 250k large VM-min/mo at 50% util**: ~23× M2-L + ~12× M4-XL ≈ €7,818/mo on-demand (~€6,255/mo committed). 35 hosts.
- Two SKUs to procure and capacity-plan. Demand mix forecasting required at procurement time.

### Path B: Single fleet, M4-XL with up to 2 VMs/host (packed)

- **One host class**: M4-XL (M4 Pro 14C / 64 GB / 2 TB at €0.49/h), dynamically running 1 or 2 VMs.
- **Small tier**: M4-XL ÷ 2 = 7 vCPU / 32 GB per VM. RAM spec above Blacksmith (24 GB) and WarpBuild (22 GB), well above Namespace (14 GB).
- **Large tier**: M4-XL whole = 14 vCPU / 64 GB.
- **Fleet sizing for the same 500k + 250k demand**: ~23× M4-XL ≈ €8,227/mo on-demand. About 5% more on hardware than Path A, but 34% fewer hosts (23 vs 35) — less of every ops cost that scales with host count.
- **Cost tradeoff per small VM**: ~17% more than Path A (€179/VM-mo vs €153/VM-mo), gains 2× the RAM headroom (32 GB vs 16 GB) and 2× the disk.
- **Quality tradeoff**: ~7.5% expected build-time penalty averaged across all jobs at 50% utilization (~25% of jobs share a host, each suffering ~30% slowdown). Matches what every competitor accepts.
- **Single fungible fleet** dynamically serves whatever small/large mix shows up. No procurement-time forecasting risk.
- **Implementation cost** (not in this PR):
  - `tart-kubelet` advertising variable pod slots per Node.
  - `runnersFleet.replicas` reinterpreted as host count, not VM count.
  - Pod `resources.requests` halved when in 2-VM mode.
  - NetworkPolicy verifying VM-to-VM isolation on the same host.
  - Per-host contention dashboard for SLO monitoring.

### Decision factors

| Factor | Path A | Path B |
|---|---|---|
| Lower total hardware $ at scale | ✓ (~5%) | |
| Lower host count + ops surface | | ✓ (34% fewer hosts) |
| Marketing differentiator ("no co-tenancy") | ✓ | |
| Spec-competitive small tier (RAM) | | ✓ (32 GB vs 16 GB) |
| Demand-mix flexibility | | ✓ (single fungible fleet) |
| Implementation complexity | ✓ (already in this PR) | |
| Build-time predictability | ✓ (zero contention) | |

### Margin profile (both paths)

50% utilization, commit pricing:

- Small @ $0.08/min: ~93% gross margin per VM
- Large @ $0.16/min: ~96% gross margin per VM
- Break-even utilization is single-digit % at every market price point on both paths

Not blocking this PR. Decision factors that will sway it: whether early customers complain about 16 GB on M2-L (pushes toward Path B), whether they value "no co-tenancy" (pushes toward Path A), and the actual measured small/large demand split (a balanced split favors Path B's fungibility; a lopsided one favors Path A).

server/lib/tuist/runners.ex
```diff
@@ -0,0 +1,283 @@
...
+    * `{:error, :no_work_yet}` — queue empty, all accounts at
+      cap, or we lost a claim race; warm Pod keeps polling.
+    * `{:error, :not_found}` — SA gone (raced with GC).
+    * `{:error, :no_pool_label}` — SA missing the fleet label.
+    * `{:error, :unknown_account}` — claimed entry's account
+      went away.
+    * `{:error, :github_mint_failed}` — GitHub refused the JIT.
+    * `{:error, :not_in_cluster}` — server not running in-cluster.
+  """
+  def dispatch_for_sa(namespace, sa_name) when is_binary(namespace) and is_binary(sa_name) do
```
pepicrft, line 83: Blocking: this endpoint treats the runner ServiceAccount token as a reusable pool credential. The token is copied into the VM at `/etc/tuist-sa-token`, and the GitHub Actions job runs as root, so customer workflow code can read it and call this endpoint again while the first job is still running. Because this path only maps the SA to a pool, and `Claims.attempt/4` only conflicts on `workflow_job_id`, the second call can mint a JIT config for the next queued job in the same pool, including another account.

We need a single-use binding here: reject dispatch for an SA or pod that already has a live claim, or make `runner_claims` unique on `pod_name` and treat the first 200 as consuming that pod.

server/lib/tuist/runners/claims.ex
```diff
@@ -0,0 +1,168 @@
...
+
+  @doc """
+  Attempts to claim `workflow_job_id` for `pod_name`. Returns
+  `{:ok, %Claim{}}` on success (we won) or `{:error, :lost_race}`
+  if the workflow_job is already claimed by another pod.
+
+  The claim is atomic on the unique PK — no verify-readback
+  needed.
+  """
+  def attempt(workflow_job_id, account_id, fleet_name, pod_name)
```
pepicrft, line 99: Blocking: the account cap check is outside the claim insert and is not protected by a transaction or account-level lock. Two pollers can compute the same `ineligible` set before either insert lands, then claim different jobs for the same account. The primary key only serializes attempts for the same `workflow_job_id`, not the account budget.

Move the cap check into the same Postgres transaction as the claim, guarded by an account row lock or advisory transaction lock, so `accounts.runner_max_concurrent` is atomic across pools.

infra/helm/tuist/templates/runners-namespace.yaml
```diff
@@ -0,0 +1,237 @@
...
+#     requests to apiserver, etc.
+#   * Ingress: default-deny. The polling model doesn't need any
+#     inbound connection to the Pod.
+#   * Egress: HTTPS / HTTP to the public internet, with
+#     `exceptCIDRs` carving out cluster-internal ranges (Pod
+#     CIDR, Service CIDR, Node CIDR). Closes the lateral-pivot
+#     path from the Pod into ClickHouse / Postgres / the
+#     server's admin endpoints.
+#
+# **What this DOES NOT gate**: the Tart VM's own network traffic.
```
pepicrft, line 172: Blocking for production: this NetworkPolicy does not isolate the code that customers actually run. The comment correctly says Tart VM traffic bypasses CNI, but the PR relies on a node or perimeter firewall to drop RFC1918 destinations from VM IPs and I do not see that firewall being installed in the CAPI bootstrap or tart-kubelet path.

Until host-level rules exist and are tested, tenants are VM-isolated for filesystem and process state, but not network-isolated from cluster and private ranges.

infra/helm/tuist/values-managed-staging.yaml
```diff
@@ -154,7 +154,84 @@ xcresultProcessor:
...
+  # these on a fresh cluster.
+  networkPolicy:
+    enabled: true
+    exceptCIDRs:
+      - "10.0.0.0/8"
+      - "172.16.0.0/12"
+      - "192.168.0.0/16"
+  pools:
+    - name: default
+      replicas: 2
```
pepicrft: This staging shape is internally inconsistent. `hostCount` is 1, this pool asks for 2 replicas, and each runner Pod requests the full host. That means one warm runner is expected to sit unschedulable, which makes smoke tests and capacity telemetry noisy. Either set staging replicas to 1 or intentionally test two-per-host by splitting the pod requests and documenting that as the staging target.

infra/helm/tuist/templates/runners-controller-deployment.yaml
```diff
@@ -0,0 +1,194 @@
...
+    type: Recreate
+  selector:
+    matchLabels:
+      {{- include "tuist.componentLabels" (dict "root" . "component" "runners-controller") | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "tuist.componentLabels" (dict "root" . "component" "runners-controller") | nindent 8 }}
+      annotations:
+        prometheus.io/scrape: "true"
```
pepicrft, line 154: Scraping the controller's default metrics is not enough to operate this safely. Before enabling production runners, we need product and ops telemetry for dispatch latency, queued age by pool/account, claim failures by reason, JIT mint failures, stale-claim releases, cap saturation, RunnerPool desired vs observed, unschedulable runner Pods, and VM boot time. Those should have a dashboard and alerts, not just logs.

fortmarek: I'd say adding more telemetry is not something to be done in this PR. This PR is already huge as-is and it will be easier to add the right metrics once we roll this out internally and start to have the data we need.

server/lib/tuist/runners/workers/stale_claims_worker.ex
```diff
@@ -0,0 +1,98 @@
...
+  """
+
+  use Oban.Worker, queue: :default, max_attempts: 1
+
+  alias Tuist.Runners.Claims
+  alias Tuist.Runners.Jobs
+
+  require Logger
+
+  @stale_after_seconds 300
```
pepicrft, line 53: Blocking: this treats every claim older than 5 minutes as stale, but `runner_claims` is also the active cap slot for jobs that already reached `running` and stays there until `workflow_job.completed`. Any legitimate build longer than 5 minutes will be requeued and its cap slot freed, causing over-cap scheduling and incorrect lifecycle state.

The stale reaper needs to target only pre-JIT or pre-running claims, or the PG table needs a running state, heartbeat, or started_at with a different timeout.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
```diff
@@ -197,13 +198,37 @@ defmodule TuistWeb.Webhooks.GitHubController do
...
+        %{"installation" => %{"id" => id}} when is_integer(id) -> id
+        _ -> nil
+      end
+
+    if installation_id do
+      # Best-effort dispatch. We always 200 back to GH so it
+      # doesn't queue retries — `:no_idle_pod` and other
+      # operational failures are recoverable on the next event:
+      # GH's queue retains queued jobs and our watcher refills
+      # the warm pool the moment a Pod terminates.
```
pepicrft: Blocking: this is not recoverable on the next event for `workflow_job.queued`. If `match_pool/1` cannot list RunnerPools, no pool exists during a deploy, or enqueue fails before a queue row is persisted, GitHub will not send another `queued` event for the same waiting job. The job stays queued in GitHub, but no warm runner will mint because our ClickHouse queue never got the row.

Return non-2xx for transient enqueue failures so GitHub retries, or add a reconciler that polls GitHub for queued jobs by label.

server/lib/tuist/github/client.ex
```diff
@@ -335,6 +335,62 @@ defmodule Tuist.GitHub.Client do
...
+  that does only what the name implies.
+
+  Runners registered at the org level are usable by any repo in
+  the org subject to the runner-group's repo allowlist. The
+  default group ID is 1; pass a different `:runner_group_id` in
+  `attrs` to register the runner into a restricted group.
+
+  See: https://docs.github.com/en/rest/actions/self-hosted-runners#create-configuration-for-a-just-in-time-runner-for-an-organization
+  """
+  def generate_jit_config(installation_id, org, attrs) do
```
pepicrft: This path opts into GHES support elsewhere in `Tuist.GitHub.Client`, but JIT generation still takes only a numeric installation id and hardcodes `https://api.github.com`. A `workflow_job` webhook from a GHES installation can be accepted by `Dispatch.handle_webhook/2`, then fail to mint, or mint against github.com for an org with the same login.

Thread the resolved `GitHubAppInstallation` or client_url through dispatch and use the installation-scoped token/API URL here, or explicitly reject GHES runner jobs.

server/lib/tuist/runners/dispatch.ex
```diff
@@ -0,0 +1,257 @@
...
+  gives the pool name; this function maps name → label).
+  """
+  def match_pool(requested_labels) when is_list(requested_labels) do
+    needle_set = MapSet.new(requested_labels, &String.downcase/1)
+
+    case Client.list_runner_pools(namespace()) do
+      {:ok, []} ->
+        {:error, :no_pools}
+
+      {:ok, items} ->
```
pepicrft, line 184: `RunnerPoolSpec` says duplicate dispatch labels are enforced server-side, but this code accepts the first match from the Kubernetes list. List ordering is not a contract, so two pools with the same `dispatchLabel` would route nondeterministically and possibly boot the wrong image.

Detect more than one match and fail loudly, or enforce uniqueness at Helm render time.

server/lib/tuist/runners.ex
```diff
@@ -0,0 +1,283 @@
...
+            fleet: fleet_name
+          )
+
+          {:error, :no_work_yet}
+      end
+    end
+  end
+
+  defp serve_claim(namespace, sa_name, fleet_name, candidate, claim) do
+    case Accounts.get_account_by_id(candidate.account_id) do
```
pepicrft, line 135: This should re-check account eligibility before minting, not only when building the ineligible list. If ops sets `runner_max_concurrent` to 0 after jobs are queued, and the account currently has no active claims, `ineligible_accounts/0` will not include it because it only iterates accounts present in `runner_claims`. A warm pod can still claim and run those queued jobs. The same atomic claim path that enforces the cap should also reject disabled accounts.

infra/tart-kubelet/internal/podagent/reconciler.go
```diff
@@ -415,8 +480,27 @@ func (r *Reconciler) podStatus(ctx context.Context, pod *corev1.Pod) (*corev1.Po
...
+	running, err := r.Tart.IsRunning(ctx, entry.VMName)
+	if err != nil {
+		// pgrep failed unexpectedly — leave the Pod Running on
+		// the optimistic read so we don't flap on a transient
+		// host hiccup. The next reconcile retries.
+		status.Phase = corev1.PodRunning
+		return status, nil
+	}
+
+	if !running {
```
pepicrft, line 518: I think this clean-exit branch is probably unreachable for the normal single-shot runner path. `Reconcile` checks `RunHandle.Exited()` before calling `podStatus` and currently marks every exited `tart run` as PodFailed, including `exitErr == nil`. Since `dispatch-poll.sh` shuts the VM down after the GitHub runner exits, successful jobs can still surface as failed Pods and pollute the failure telemetry. Can we branch on `exitErr` in the earlier `RunHandle.Exited()` path, or move clean-exit handling into one place?

server/lib/tuist/github/client.ex
```diff
@@ -335,6 +335,62 @@ defmodule Tuist.GitHub.Client do
...
+    case App.get_installation_token(installation_id) do
+      {:ok, %{token: token}} ->
+        body = %{
+          "name" => Map.fetch!(attrs, :name),
+          "runner_group_id" => Map.get(attrs, :runner_group_id, 1),
```
pepicrft: Before enabling this broadly, I would avoid silently defaulting to the org default runner group. That group is commonly available to every repository in the org, and the dispatch path accepts any `workflow_job` for an enabled account owner as long as the label matches a pool. That means any repo with Actions enabled in the org can consume this Mac capacity and run on the same customer runner substrate. Can we make the runner group an explicit account-level setting with a restricted repo allowlist, or fail closed when it is not configured for production accounts?

infra/tart-kubelet/internal/satoken/satoken.go
```diff
@@ -0,0 +1,83 @@
...
+	if saName == "" {
+		saName = "default"
+	}
+	exp := m.ExpirationSeconds
+	if exp <= 0 {
+		exp = 3600
+	}
+
+	req := &authenticationv1.TokenRequest{
+		Spec: authenticationv1.TokenRequestSpec{
```
pepicrft, line 85: Can we make this token audience-bound to the Tuist dispatch endpoint and pass the same audience in TokenReview? As written, this mints the default Kubernetes API audience token and the server accepts it without an expected audience. If the token leaks from the guest, it is also a valid credential to the API server for this ServiceAccount, and the dispatch endpoint will accept any default-audience token for a labelled runner SA. A dedicated audience would make the credential single-purpose and reduce the blast radius.

server/lib/tuist/github/client.ex
fortmarek: > That means any repo with Actions enabled in the org can consume this Mac capacity and run on the same customer runner substrate. 

Runners are already opt-in via max concurrency which is set to 0 and thus currently only available to `tuist` repos.

fortmarek: If you want to restrict the actions just to `tuist/tuist` then I don't understand why?

server/lib/tuist/runners.ex
fortmarek: Addressed in d834dafa72: `Claims.attempt/4` now runs inside a Postgres transaction with an advisory lock on `account_id`, and one of the in-tx checks rejects the call if the polling pod already owns a live claim (`{:error, :pod_in_use}`). The replay attack you described — customer code in the VM reading `/etc/tuist-sa-token` and calling dispatch a second time — fails closed at that check. Test coverage in `test/tuist/runners/claims_test.exs` (`returns :pod_in_use when the same pod already has a live claim`).

server/lib/tuist/runners/claims.ex
fortmarek: Addressed in d834dafa72: cap check is now inside the same Postgres transaction as the insert, guarded by `pg_advisory_xact_lock($namespace, $account_id)` so two pollers serialise on the lock and can't both pass the count check in parallel. The check reads the live `runner_max_concurrent` from `accounts` and counts current claims for the account across all fleets before the INSERT. Test in `claims_test.exs` (`returns :over_cap when the account is already at runner_max_concurrent`) drives 3 attempts against a cap=2 account; the third gets `:over_cap`.

infra/helm/tuist/templates/runners-namespace.yaml
fortmarek: Addressed in a9f9201706: `installVMEgressFirewall` in `infra/macos-host-bootstrap` now writes a pfctl anchor at `/etc/pf.anchors/tuist.runners` that drops vmnet-source packets (192.168.64.0/22) destined for 10.0.0.0/8, 172.16.0.0/12, and 169.254.0.0/16 (incl. cloud metadata 169.254.169.254). The anchor is spliced into `/etc/pf.conf`, `pfctl -nf` validates before activation, and a `dev.tuist.pfctl-runners` LaunchDaemon re-arms the ruleset at every boot. The runners-namespace.yaml comment now points to where the firewall actually lives.

infra/helm/tuist/values-managed-staging.yaml
fortmarek: Addressed in 0417f642d5: dropped `runnersFleet.pools[0].replicas` from 2 to 1 to match `hostCount: 1`. Each runner Pod requests the full host, so the 2-replica shape was guaranteed to leave one Pod Pending — confirmed during the staging bring-up. With the change, staging is internally consistent and the 8e293b6a49 staging+canary hostCount=1 makes the whole tier 1+1.

server/lib/tuist/runners/workers/stale_claims_worker.ex
fortmarek: Addressed in d834dafa72: added a `lifecycle_state` column to `runner_claims` (`claimed` post-INSERT, `running` after JIT mint via `Claims.mark_running/2`), and `list_stale/1` now filters on `lifecycle_state = 'claimed'`. A long-running build holding a healthy cap slot is invisible to the worker — only pre-mint claims older than 5 min are reaped. A partial index `(claimed_at) WHERE lifecycle_state='claimed'` keeps the scan tight. Test (`skips claims that already transitioned to running`) confirms the worker leaves running rows alone.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
fortmarek: Addressed in d834dafa72: the webhook handler now returns 503 on `{:error, _}` from `Dispatch.handle_webhook/2` (transient failures — K8s LIST race, ClickHouse INSERT timeout, account lookup blowup) so GitHub retries with backoff. Permanent decisions (`{:ok, _}` / `:ignored`) still return 200; nothing useful comes from GitHub redelivering when we've made a final routing decision. Logger.warning carries the reason so ops can alert on a sustained 503 pattern.

server/lib/tuist/github/client.ex
fortmarek: Addressed in d834dafa72: `Runners.mint_jit/3` now resolves the full `GitHubAppInstallation` via `VCS.get_github_app_installation_for_account/1` (carrying `installation_id` AND `client_url`), and `GitHubClient.generate_jit_config/3` takes the installation map, derives the API URL via `installation_api_url/1`, and uses `pin_ghes_url/2` for the SSRF pin on GHES hosts. JIT generation now lands on the customer's GHES host for GHES installations and api.github.com for SaaS.

server/lib/tuist/runners/dispatch.ex
fortmarek: Addressed in d834dafa72: `Dispatch.match_pool/1` now uses `Enum.filter` and returns `{:error, :ambiguous_pool}` when more than one pool matches the requested label. The webhook handler maps that to 200 + `:ignored` (logger.error gives ops something to alert on) since GitHub can't redeliver a routing decision. Chart-side render-time uniqueness is the longer-term gate; this is the server-side fail-loud belt-and-suspenders.

server/lib/tuist/runners.ex
fortmarek: Addressed in d834dafa72: the eligibility re-check is now atomic — the same `Claims.attempt/4` transaction that performs the claim INSERT reads `runner_max_concurrent` and the live in-flight count under the per-account advisory lock. `ineligible_accounts/0` survives as a pre-filter optimisation (skip CH lookups for capped accounts) but is no longer the authoritative gate. Setting `runner_max_concurrent` to 0 after jobs are queued: the next attempt returns `{:error, :runners_disabled}` regardless of whether the account currently has any claims. Test in `claims_test.exs` (`returns :runners_disabled when runner_max_concurrent is 0`) covers it.

infra/tart-kubelet/internal/podagent/reconciler.go
fortmarek: Addressed in d5bb93fc8a: the `RunHandle.Exited()` branch in `Reconcile` now switches on `exitErr`. `exitErr == nil` is the dispatch-poll.sh clean-shutdown path after a successful job — published as PodSucceeded with reason `TartRunExited`. `exitErr != nil` is the genuine failure (guest crash, tart run died) — PodFailed as before. Single-shot successes no longer pollute the failure telemetry.

infra/tart-kubelet/internal/satoken/satoken.go
fortmarek: Addressed in d5bb93fc8a + d834dafa72: tart-kubelet's `ClientMinter` now defaults to `Audiences: [satoken.DispatchAudience]` (= `tuist-runners-dispatch`), and `Tuist.Kubernetes.Client.create_token_review/1` posts the same audience in the TokenReview spec and asserts it's in the response's validated `audiences` list before authenticating. A guest-leaked SA token no longer carries the default K8s API audience — it's single-purpose for this endpoint, and a TokenReview without our audience returns `{:error, :unauthenticated}`.

server/lib/tuist/runners.ex
```diff
@@ -0,0 +1,316 @@
...
+      {:ok, account} ->
+        pod_name = pod_name_from_sa(sa_name)
+
+        with {:ok, dispatch_label} <- Dispatch.dispatch_label_for_pool(fleet_name),
+             :ok <- stamp_owner_labels(namespace, pod_name, account),
+             {:ok, jit, runner_name} <- mint_jit(account, sa_name, dispatch_label),
+             :ok <- Claims.mark_running(candidate.workflow_job_id, runner_name),
```
pepicrft, line 142: This should keep the PG claim in `claimed` until the ClickHouse `running` transition has landed, or explicitly release/requeue if that transition raises. I reproduced the current sequence with a live ExUnit check: after `mark_running/2`, a failure before `Jobs.record_running/2` leaves the row in `running`, `Claims.list_stale/1` skips it, and the account cap stays consumed. That would strand the GitHub job because the JIT never reaches the VM.

infra/cluster-api-provider-scaleway-applesilicon/internal/scaleway/client.go
```diff
@@ -205,6 +205,91 @@ func (c *Client) findServerByName(ctx context.Context, name, zone string) (*appl
...
+			Zone:     scw.Zone(zone),
+			Page:     &page,
+			PageSize: &pageSize,
+		}, scw.WithContext(ctx))
+		if err != nil {
+			return nil, fmt.Errorf("list servers: %w", err)
+		}
+
+		for _, s := range resp.Servers {
+			if !strings.HasPrefix(s.Name, poolPrefix) {
```
pepicrft, line 275: This adoption path needs to check for an existing server named `claimName` before scanning the pool prefix. I reproduced the retry window with a fake Scaleway API and the real SDK client: if the first reconcile renames `tuist-pool-001` to the Machine name, then crashes or loses the status patch before `status.serverID` is persisted, the next reconcile only scans prefixed hosts and returns `ErrNoAvailableHost`. The host has been claimed, but the CR can no longer rediscover it.

.github/workflows/release.yml
```diff
@@ -2032,6 +2255,149 @@ jobs:
...
+          DIGEST="${{ steps.push.outputs.digest }}"
+          for f in \
+            infra/helm/tuist/values-managed-staging.yaml \
+            infra/helm/tuist/values-managed-canary.yaml \
+            infra/helm/tuist/values-managed-production.yaml; do
+            # Only rewrite if the file already has a non-empty digest pin
+            # — values-managed-{canary,production}.yaml ship with
+            # runnerImage: "" until canary/production are flipped on, and
+            # we don't want to populate them silently from the release flow.
+            if grep -qE '^\s*runnerImage:\s*"ghcr\.io/tuist/tuist-runner@sha256:' "$f"; then
```
pepicrft: This guard currently skips staging. I ran the release substitution against the current staging values and the grep does not match because staging is pinned as `ghcr.io/tuist/tuist-runner:08e66c9b`, not `@sha256:...`. A runner-image release can build and push a new image while leaving the only enabled environment on the old tag.

.github/workflows/runners-staging-smoke.yml
```diff
@@ -0,0 +1,79 @@
...
+  # specifically — the runner-pool smoke test isn't a default
+  # CI hook (would burn runner cycles on every PR), but during
+  # the staging bring-up we need a way to dispatch it without
+  # the workflow file being on main yet (gh workflow run
+  # requires that). Empty commits on the branch trigger the
+  # job.
+  push:
+    branches:
+      - feat/runners-shape-2
+    paths:
```
pepicrft, line 35: This does not match the comment above: with the `paths` filter, empty commits on `feat/runners-shape-2` will not trigger the smoke workflow. Only edits to this workflow file do. Either drop the paths filter for the bring-up branch or adjust the comment.

pepicrft (approved):
Approved after the focused runtime checks. I left a few inline comments for failure windows worth tightening before we rely on this path operationally.

server/lib/tuist/runners.ex
fortmarek: Addressed in 84498b8e8d. Pulled `Jobs.record_running/2` into a `record_running_safe/2` helper that catches the raise, logs the CH error, and returns `{:error, :record_running_failed}`. The `with`'s existing `else` then runs `release_safely/3`, the workflow_job goes back to queued, and the cap accounting matches reality.

infra/cluster-api-provider-scaleway-applesilicon/internal/scaleway/client.go
fortmarek: Addressed in 84498b8e8d. `AdoptByPrefix` now calls `findServerByName(claimName, zone)` up front — if a previous reconcile already renamed a pool host but lost the status patch, the next reconcile rediscovers the host by its claim name instead of scanning the (now-missing) prefix. Mirrors the same rediscovery pattern `CreateServer` uses.

.github/workflows/release.yml
fortmarek: Addressed in 84498b8e8d. Broadened the guard to `ghcr\.io/tuist/tuist-runner[@:]` (covers both `@sha256:...` and `:tag` pins) and rewrote the sed to match the same pattern. Staging on `:08e66c9b` will now get rewritten to the new digest on the next runner-image release.

.github/workflows/runners-staging-smoke.yml
fortmarek: Addressed in 84498b8e8d. Comment now describes the actual semantics: edits to this workflow file auto-fire the smoke; every other push needs an explicit `gh workflow run runners-staging-smoke.yml`. The bring-up branch's main use-case (re-validating against the live fleet when the smoke definition changes) still works.

PR #10657, opened 2026-05-06
fix(capi-scaleway): unblock first-time macOS-fleet helm rollouts

fortmarek:
## Summary

The first canary deploy that actually exercised `macosFleet.enabled=true` ([run 25441924886](https://github.com/tuist/tuist/actions/runs/25441924886/job/74637244699)) hung on `helm upgrade --install --wait` because two independent bugs compounded — the deploy worked on staging only because staging's xcresult-processor Pod was already happy from a previous bring-up that today's revisions didn't touch.

### Bug 1 — tart-kubelet destroys the VM after 8 min, then re-clones from scratch

`tart.Client.Run` blocked up to 8 min for the guest to obtain a DHCP lease, then called `cleanupAfterFailedRun` (stop + delete the VM) on timeout. Each reconcile retry then re-pulled, re-cloned, and re-cold-booted from zero — converging on the same wall. On canary's M4-S Mac mini the first-boot pipeline didn't get past 8 min and the deploy never made progress.

Fix: `Run` now returns as soon as `tart run` is alive (with a 5s sanity check that catches fast-fail cases). IP polling moves entirely to `podStatus`, so the Pod stays Pending across as many reconciles as the guest needs and transitions to Running the moment configd hands out a lease. Helm's `--wait --timeout 90m` is the right place for a top-level deadline; an inner one just forced waste. The Store entry is recorded before `Run` so `deletePod` and the GC keep tracking the VM through any partially-started state.

### Bug 2 — Hetzner CCM evicts Mac mini Nodes

Within ~30s of tart-kubelet missing a heartbeat (because of bug 1 wedging the host), the workload cluster's hcloud-cloud-controller-manager fired its `cloud-node-lifecycle` controller, called `cloudprovider.Instances.InstanceExists()` for the `scw-applesilicon://fr-par-1/...` providerID, got `false`, and deleted the Node. The `ScalewayAppleSiliconMachine` CR was still `Ready` but the kube `Node` was gone, so a fresh xcresult-processor Pod had nowhere to land.

Fix: pass `--controllers=*,-cloud-node-lifecycle` to HCCM. CAPI's `MachineHealthCheck` already covers dead Hetzner Linux nodes; Mac mini lifecycle is owned by the Scaleway Apple Silicon CAPI provider's finalizers on the CR. The `cloud-node-lifecycle` controller wasn't carrying weight we relied on.

### Closing the deploy gap

The HCCM values file was previously only applied when someone re-ran `mise run k8s:bootstrap-workload` by hand, so chart edits silently failed to propagate. The new `.github/workflows/hccm-deployment.yml` fires on push to main when `infra/k8s/mgmt/bootstrap/hccm-values.yaml` changes and runs `helm upgrade --install hccm` in parallel against canary, staging, and production. Shares the existing `server-deploy-<env>` concurrency lane so HCCM upgrades can't interleave with a server rollout.

The PR title is scoped `fix(capi-scaleway):` so `release.yml` cuts a new operator semver, rebuilds the operator image with the new tart-kubelet binary baked in, and rewrites `macosFleet.image.tag` in `values-managed-common.yaml` automatically. The follow-up release-bump commit is what flows the new tart-kubelet onto the Mac minis.

## Test plan

- [ ] After merge, `HCCM Deployment` workflow runs in parallel against canary/staging/production. Confirm the `hccm` Helm release in each cluster's `kube-system` rolls to a new revision and the `--controllers=*,-cloud-node-lifecycle` arg shows up in `kubectl -n kube-system describe deploy hcloud-cloud-controller-manager`.
- [ ] `release.yml` cuts a `capi-scaleway@<n>` release and opens an auto-bump commit/PR rewriting `macosFleet.image.tag` in `values-managed-common.yaml`. Merge that.
- [ ] Re-trigger the canary server deploy. Confirm `helm upgrade` completes (Mac mini Node stays Ready throughout, xcresult-processor Pod transitions Pending → Running once the guest's IP shows up) instead of hanging on first-time bring-up.
- [ ] `go test ./infra/tart-kubelet/...` (already green locally).

### DIFF
--- a/.github/workflows/hccm-deployment.yml
+++ b/.github/workflows/hccm-deployment.yml
@@ -1,0 +1,91 @@
+name: HCCM Deployment
+
+# Roll the hcloud-cloud-controller-manager chart values onto every
+# workload cluster on push to main. HCCM lives in kube-system and was
+# previously only installed during fresh-cluster onboarding via
+# `mise run k8s:bootstrap-workload`. That meant edits to
+# infra/k8s/mgmt/bootstrap/hccm-values.yaml had no effect on existing
+# clusters until someone re-ran bootstrap by hand — easy to forget,
+# and the kind of drift that bites at the worst possible time
+# (e.g. a CCM controller pruning Mac mini Nodes during a deploy).
+#
+# Idempotent. The matrix runs all environments in parallel: HCCM is
+# a workload-level component, blast radius is per-cluster, and the
+# change is a thin chart upgrade — no need for a canary→production
+# cascade like the application server has.
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - infra/k8s/mgmt/bootstrap/hccm-values.yaml
+      - .github/workflows/hccm-deployment.yml
+  workflow_dispatch: {}
+
+permissions:
+  contents: read
+
+jobs:
+  apply:
+    name: HCCM (${{ matrix.environment }})
+    strategy:
+      fail-fast: false
+      matrix:
+        # Per-cluster fixed region — sourced from
+        # infra/k8s/clusters/cluster-<env>.yaml's `region` topology
+        # variable. All workload clusters live in fsn1 today; spreading
+        # across regions is just a matter of overriding `region` per
+        # entry below — the chart's HCLOUD_LOAD_BALANCERS_LOCATION env
+        # follows.
+        include:
+          - environment: canary
+            region: fsn1
+          - environment: staging
+            region: fsn1
+          - environment: production
+            region: fsn1
+    runs-on: namespace-profile-default
+    environment:
+      name: server-k8s-${{ matrix.environment }}
+    # Hold the same lane as server-deployment.yml's deploy + observability
+    # jobs so HCCM upgrades and server rollouts can't race for the kube
+    # apiserver on the same cluster.
+    concurrency:
+      group: server-deploy-${{ matrix.environment }}
+      cancel-in-progress: false
+    timeout-minutes: 10
+    env:
+      KUBECONFIG: ${{ github.workspace }}/.kube/config
+    steps:
+      - uses: actions/checkout@v4
+      - uses: jdx/mise-action@v3.2.0
+        with:
+          install_args: "helm kubectl"
+          cache: "false"
+      - uses: 1password/install-cli-action@v2
+      - name: Load kubeconfig from 1Password
+        env:
+          OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
+        run: |
+          mkdir -p "$(dirname "$KUBECONFIG")"
+          op document get "kubeconfig: tuist-${{ matrix.environment }}" \
+            --vault "tuist-k8s-${{ matrix.environment }}" --output "$KUBECONFIG"
+          chmod 600 "$KUBECONFIG"
+      - name: Verify cluster reachability
+        # Same pre-flight as server-deployment.yml — fail fast if the
+        # 1P-stored kubeconfig points at a phantom apiserver instead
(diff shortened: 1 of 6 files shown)


pepicrft (commented):
I found two blockers that we should address before merging:

1. `.github/workflows/hccm-deployment.yml`: rendering the current Hetzner chart with this values file still points `HCLOUD_TOKEN` at `secretKeyRef.name: hcloud`, `key: token`, while our bootstrap creates `kube-system/hetzner` with key `hcloud` (`infra/mise/tasks/k8s/bootstrap-workload.sh`). On the first automated rollout the HCCM pod will fail env resolution and `helm --wait` will time out before this fix reaches clusters. Can we override `env.HCLOUD_TOKEN` to `hetzner/hcloud` in the values file, or make the workflow/bootstrap create the chart-default secret?

2. `infra/tart-kubelet/internal/tart/tart.go`: after the 5s sanity window we never consume the `cmd.Wait()` result again. If `tart run` exits at 6s or later before the guest gets an IP, `createPod` has already stored the VM entry and future reconciles skip creation, while `podStatus` keeps the Pod Pending on `Tart.IP` errors. That can still hang Helm indefinitely for non-immediate VM failures. Can we either keep tracking process exit from the reconciler, or have `podStatus` detect a stopped/exited VM and clear/fail/restart it?

fortmarek:
Thanks for the careful read — both addressed in dad6adb5f3.

**Blocker 2 (post-launch `tart run` exits)** — `Tart.Run` now returns a `*RunHandle` exposing `Done()` / `Exited()`. The reconciler stashes it in the Store entry and checks it on every reconcile *before* falling through to `createPod` / `podStatus`. A process that survives the 5s sanity window but exits later (guest panic 30s in, configd crash, BEAM bailout, etc.) is surfaced as a `PodFailed` with reason `TartRunExited`, the entry + VM are torn down via `deleteByKey`, and the owning ReplicaSet schedules a fresh replacement Pod. No more silent Pending → helm-`--wait`-timeout path. Locked in by two new tests in [tart_test.go](infra/tart-kubelet/internal/tart/tart_test.go): one for the immediate-exit case, one for the post-window-exit case (uses a `sleep 6` shim binary so the launcher goroutine has to actually catch the late exit).

Entries hydrated by `recoverState` after a kubelet restart still leave `Run` nil — we don't have a `cmd.Wait` to attach to those VMs because they're no longer a child of this kubelet — and the exit check explicitly skips when nil, falling back to the IP probe like before. Documented that on the `Entry` type.

**Blocker 1 (HCCM secret wiring)** — turns out the comment in `hccm-values.yaml` (and the matching one in `hcloud-csi-values.yaml`) was the source of the confusion. Verified live on canary + staging:

```
$ kubectl -n kube-system get deploy hcloud-cloud-controller-manager \
    -o jsonpath='{...env[?(@.name=="HCLOUD_TOKEN")].valueFrom.secretKeyRef}'
{"key":"token","name":"hcloud"}

$ kubectl -n kube-system get secret hcloud
NAME     TYPE     DATA   AGE
hcloud   Opaque   8      43h        # keys: token, hcloud, robot-*, apiserver-*, ...
```

The chart-default `hcloud/token` secret already exists on every workload cluster — caph (the Cluster API Hetzner provider) writes it during workload-cluster bring-up. Both HCCM and hcloud-csi pick up the chart default and have been doing so since cluster onboarding. The `hetzner` secret stamped by `bootstrap-workload.sh` is vestigial; nothing reads it today. So the first automated rollout via `.github/workflows/hccm-deployment.yml` will reuse the same secret reference HCCM is already wired to and won't break env resolution.

Fixed the misleading comments in both values files so a future reader doesn't have to chase the chain. Out of scope for this PR but worth a follow-up: drop the unused `hetzner` Secret + its bootstrap step.

pepicrft (approved):

PR #10658, opened 2026-05-06
perf(server): add project_id to flaky test_case_runs queries

fortmarek:
### Summary

Six queries against the `test_case_runs` ClickHouse table filtered by `test_case_id` (or other columns) without `project_id`. The table's `ORDER BY (project_id, test_case_id, ran_at, id)` means a missing `project_id` defeats prefix matching, so ClickHouse falls back from binary search to generic exclusion search and scans most granules.

### Pre-fix prod impact (last 7 days)

A Grafana `query_log` analysis surfaced what these queries actually cost in production:

| Function | Executions (7d) | p50 | p90 | p99 | Avg rows / bytes read per query |
|---|---|---|---|---|---|
| `get_test_case_flakiness_rate` | **4,939** | — | **185ms** | — | **5.29M rows / 88 MiB** |
| `get_flaky_runs_groups_count_for_test_case` | 347 | 42ms | 77ms | 174ms | 2.81M rows / 37.7 MiB |
| `list_flaky_runs_for_test_case` (groups) | 296 | 46ms | 138ms | 324ms | 2.82M rows / 37.8 MiB |
| `get_flaky_runs_groups_counts_for_test_cases` | not in window | — | — | — | — |
| `list_flaky_runs_for_test_case` (runs) | not in window | — | — | — | — |
| `get_flaky_run_group_for_test_case_run` | not in window | — | — | — | — |
| `get_cross_run_flaky_runs` | not in window | — | — | — | — |

`get_test_case_flakiness_rate` alone read **~26 billion rows / ~430 GiB** across the 7-day window. Each call reads more data than is justified for a single test case's flakiness rate; the cumulative impact dwarfs the other five queries combined.

The four "not in window" entries are tied to UI flows (paginated flaky-runs list on the test-case detail page, single test-case-run detail page) that get less traffic than the dashboard widget driving the flakiness rate. The fix still applies — they just won't show as much delta until someone opens those pages.

### Functions updated

`lib/tuist/tests.ex`:

- `get_flaky_runs_groups_count_for_test_case/1` → `/2`
- `get_flaky_runs_groups_counts_for_test_cases/1` → `/2`
- `list_flaky_runs_for_test_case/2` → `/3`
- `get_flaky_run_group_for_test_case_run/1` — pulls `project_id` from the existing struct argument
- `get_cross_run_flaky_runs/2` (private) — derives `project_ids` from the `current_flaky_runs` it already has

`lib/tuist/tests/analytics.ex`:

- `get_test_case_flakiness_rate/1` — pulls `project_id` from the `%TestCase{}` pattern match (no caller change)

`lib/tuist_web/live/test_case_live.ex` updated to pass `project.id` to `list_flaky_runs_for_test_case/3`. All other callers already had the full struct.

### How to test locally

`EXPLAIN indexes = 1` against local ClickHouse, same shape for every fixed query:

| | Search algorithm | Granules scanned |
|---|---|---|
| Before (no `project_id`) | generic exclusion search | 14/16 → 13 (after bloom) |
| After (with `project_id`) | **binary search** | **2/16 → 1 (after bloom)** |

Run the affected suites:

```sh
cd server
mix test test/tuist/tests_test.exs --only describe:"get_flaky_runs_groups_count_for_test_case/2" \
  --only describe:"get_flaky_runs_groups_counts_for_test_cases/2" \
  --only describe:"list_flaky_runs_for_test_case/3" \
  --only describe:"get_flaky_runs_for_test_run/1" \
  --only describe:"get_flaky_run_group_for_test_case_run/1"
mix test test/tuist/tests/analytics_test.exs --only describe:"get_test_case_flakiness_rate/1"
mix test test/tuist_web/live/test_case_live_test.exs
```

All 36 + 6 tests pass; `mix format --check-formatted` and `mix credo` clean.

### Post-deploy verification

Run against prod ClickHouse — should drop to zero (or near zero) rows once the deploy lands. Filter on the `tables` array so MV-backed queries (`test_case_runs_by_test_run`, `test_case_runs_active_daily_stats`, etc.) don't appear as false positives:

```sql
SELECT normalizeQuery(query), count(), round(quantile(0.9)(query_duration_ms)) AS p90_ms,
       formatReadableQuantity(avg(read_rows)) AS avg_read_rows
FROM system.query_log
PREWHERE event_date >= today() - 1
WHERE type = 'QueryFinish' AND query_kind = 'Select'
  AND has(tables, 'default.test_case_runs')
  AND query LIKE '%is_flaky%'
  AND query NOT LIKE '%project_id%'
GROUP BY 1 ORDER BY p90_ms DESC
```

A pre-fix run of this query (1-day window) returned 3 normalized shapes against the base `test_case_runs` table — matching #1 (`get_flaky_runs_groups_count_for_test_case`), #3a (`list_flaky_runs_for_test_case` groups), and `get_test_case_flakiness_rate`. Post-deploy, all three should fall out of the result set.

### Out of scope

The original report also flagged three full-table-scan probe queries (`SELECT * FROM <mv> ORDER BY <ts> DESC LIMIT 5`) on `test_case_branch_presence`, `test_case_runs_by_shard_id`, and `test_case_run_daily_stats_per_case`. Those are alert-rule probes (Grafana fingerprint `ffeb6l2ax5qtcf`) and don't appear anywhere in this repo — they need to be fixed in the Grafana alert rule definition by adding a bounded `WHERE` clause (e.g. `WHERE ran_at >= now() - INTERVAL 1 DAY`).

The reported `test_cases FINAL` JOIN is already resolved (MissingSeries) and the existing `FINAL` usage on `test_cases` is intentional — it's a `ReplacingMergeTree(inserted_at)` storing denormalized state, and `project_id` is already the leading PK column.

### DIFF
--- a/server/lib/tuist/tests.ex
+++ b/server/lib/tuist/tests.ex
@@ -2233,9 +2233,10 @@ defmodule Tuist.Tests do
   @doc """
   Returns the count of unique flaky run groups (scheme + commit_sha) for a test case.
   """
-  def get_flaky_runs_groups_count_for_test_case(test_case_id) do
+  def get_flaky_runs_groups_count_for_test_case(project_id, test_case_id) do
     query =
       from(tcr in TestCaseRun,
+        where: tcr.project_id == ^project_id,
         where: tcr.test_case_id == ^test_case_id,
         where: tcr.is_flaky == true,
         select: fragment("count(DISTINCT (scheme, git_commit_sha))")
@@ -2247,11 +2248,12 @@ defmodule Tuist.Tests do
   @doc """
   Returns a map of test_case_id => count of unique flaky run groups for multiple test cases.
   """
-  def get_flaky_runs_groups_counts_for_test_cases([]), do: %{}
+  def get_flaky_runs_groups_counts_for_test_cases(_project_id, []), do: %{}
 
-  def get_flaky_runs_groups_counts_for_test_cases(test_case_ids) do
+  def get_flaky_runs_groups_counts_for_test_cases(project_id, test_case_ids) do
     query =
       from(tcr in TestCaseRun,
+        where: tcr.project_id == ^project_id,
         where: tcr.test_case_id in ^test_case_ids,
         where: tcr.is_flaky == true,
         group_by: tcr.test_case_id,
@@ -2267,13 +2269,14 @@ defmodule Tuist.Tests do
   Fetches flaky runs for a specific test case, grouped by scheme and commit SHA.
   Returns paginated groups, each containing all runs with their failures.
   """
-  def list_flaky_runs_for_test_case(test_case_id, params \\ %{}) do
+  def list_flaky_runs_for_test_case(project_id, test_case_id, params \\ %{}) do
     page = Map.get(params, :page, 1)
     page_size = Map.get(params, :page_size, 20)
     offset = (page - 1) * page_size
 
     groups_query =
       from(tcr in TestCaseRun,
+        where: tcr.project_id == ^project_id,
         where: tcr.test_case_id == ^test_case_id,
         where: tcr.is_flaky == true,
         group_by: [tcr.scheme, tcr.git_commit_sha],
@@ -2292,6 +2295,7 @@ defmodule Tuist.Tests do
 
     flaky_runs_query =
       from(tcr in TestCaseRun,
+        where: tcr.project_id == ^project_id,
         where: tcr.test_case_id == ^test_case_id,
         where: tcr.is_flaky == true,
         order_by: [desc: tcr.ran_at]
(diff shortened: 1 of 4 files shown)


pepicrft (approved):
Looks good to me. The `project_id` prefix is now applied to the flaky `test_case_runs` queries that this PR targets, and the call-site changes look consistent.

Non-blocking follow-up: the same detail/API/MCP flows still call `test_case_reliability_by_id/2` and `test_case_analytics_by_id/1`, which query `test_case_runs` by `test_case_id` only. It'd be worth threading `project_id` through those helpers too so the rest of the test-case detail surface benefits from the same primary-key pruning.

PR #10660, opened 2026-05-06
fix(server): split quarantine dashboard chart into muted and skipped series

irena327:

Split the quarantine dashboard analytics chart into two separate series showing muted tests and
  skipped tests independently, instead of a single combined "Quarantined Tests" count.

  Previously the chart showed one line with the total number of quarantined tests. Now it displays two
   lines:
  - Muted (warning/yellow) — tests that are muted (failures ignored but still executed)
  - Skipped (destructive/red) — tests that are skipped entirely

  The data model already tracked these states separately via muted/unmuted and skipped/unskipped
  events — this change surfaces that distinction in the analytics chart.

> Changes

  - `server/lib/tuist/tests/analytics.ex `— quarantined_tests_analytics/2 now returns muted_count,
  skipped_count, muted_values, and skipped_values instead of a single count/values
  - `server/lib/tuist_web/live/quarantined_tests_live.html.heex` — Renders two chart series with
  separate legend items
  - `server/test/tuist/tests/analytics_test.exs` — Updated existing tests to match new return shape;
  added tests for skipped event tracking

### How to test locally

Added unit tests


fortmarek:
Hey @irena327! I pushed a few adjustments on top of your branch. Here's the rundown so the diff is easier to review:

**Three widgets instead of two legends.** Following the pattern from the Flaky Tests page, the analytics card now has three clickable widgets — **Quarantined tests** (total), **Muted**, **Skipped** — with their own trend badges (`since last month`, etc.). The chart below swaps to whichever widget is selected and the selection persists via `?analytics-selected-widget=…` in the URL.

**`quarantined_tests_analytics/2` returns the combined series too.** It now exposes `count`/`values` (muted + skipped) alongside `muted_count`/`muted_values`/`skipped_count`/`skipped_values` and per-mode trends, all derived from the same event replay so the widgets and chart can't disagree. I extracted a `quarantine_counts_at/2` helper so the per-bucket loop and the previous-period lookup share the same logic.

**Seed bug fix.** While verifying in the browser I noticed the Skipped widget was always 0 even though the table showed Skipped tests. Root cause: `priv/repo/seeds.exs` emitted `muted`/`unmuted` events for every quarantined test regardless of its `state`, so tests in `skipped` state had no matching events and the event-sourced analytics couldn't see them. The seed now branches on the test's actual state and emits the correct `muted`/`unmuted` or `skipped`/`unskipped` chain. Production code (`update_test_case`) was already correct, so live data is unaffected.

**CSS + chart polish.** Added `[data-part="widgets"]` flex layout and `[data-part="analytics-content"]` styling to `quarantined_tests.css` (mirroring `flaky_tests.css`) so the three widgets sit side-by-side on desktop and stack on mobile. Aligned the chart grid with the rest of the dashboards (`width: "93%"` + `right: "7%"`) so the `May 7` label no longer overflows the card.

**Tests.** Updated existing assertions to also cover `count`/`values`/`trend` and added a focused `trend reflects change between period start and end` test that exercises a non-zero combined trend (1 → 3 = +200%) plus per-mode trends.

**Changelog.** Added `2026.05.07-quarantine-mode-breakdown.md` describing the three-widget split for the public changelog.

Let me know if anything looks off or if you'd rather take a different direction on any of it!

fortmarek (approved):

fortmarek:
Thanks again @irena327!

PR #7222, opened 2025-01-06
Fix race condition when running multiple tuist dump commands in parallel

danibachar:
Resolves https://github.com/tuist/tuist/issues/7143

### Short description 📝

The migration to the new async file system introduces race conditions, when the tuist cli is used in an async fashion.
Following the description in [this issue](https://github.com/tuist/tuist/issues/7143), we decided to just disregard the `fileAlreadyExists` error that can be thrown by nio.

### How to test the changes locally 🧐

Testing it locally is a bit hard, one would need to build the tuist cli from source, and follow the steps in the issue on how to reproduce that.

Nevertheless, unit tests were added to cover the issue.

Notes - @fortmarek / @pepicrft - I have added a manual mock to the `FileSysteming` protocol. I have seen that you started to adopt the `Mockable` framework, and can definitely update the [FileSystem](https://github.com/tuist/FileSystem/tree/main) code with that, if preferred.

Moreover, I can also create some kind of integration/acceptance tests utilizing the short script I wrote to reproduce the original issue - although I think its an overkill - if you wish me to do that, can you please point me to where would be most appropriate to write that test?

### Contributor checklist ✅

- [x] The code has been linted using run `mise run lint-fix`
- [x] The change is tested via unit testing or acceptance testing, or both
- [x] The title of the PR is formulated in a way that is usable as a changelog entry
- [ ] In case the PR introduces changes that affect users, the documentation has been updated

### Reviewer checklist ✅

- [ ] The code architecture and patterns are consistent with the rest of the codebase
- [ ] Reviewer has checked that, if needed, the PR includes the label `changelog:added`, `changelog:fixed`, or `changelog:changed`, and the title is usable as a changelog entry

PR #10662, opened 2026-05-06
fix(cli): resolve race condition in CachedManifestLoader parallel writes

ze-diaz:
### Summary

Fix race condition in `CachedManifestLoader` when multiple tuist `dump`/`generate`/`build`/`test` commands run in parallel. The `write()` method used a check-then-act pattern (exists → mkdir / exists → remove → write) that races when two processes try to cache the same manifest simultaneously, causing EEXIST / renamex_np crashes.

Introduce `ConcurrentFileSystem.ignoringAlreadyExists` — a shared utility in `TuistSupport` that catches all "already exists" error variants from `File_System_Primitives` (mkdir, copy, move, symlink, atomic write) and `CocoaError.fileWriteFileExists`. Apply it in `CachedManifestLoader` via a new `FileSysteming.makeDirectoryIfNeeded(at:options:)` extension, and remove the **exists → remove** step before `writeText` since atomic writes handle replacement.

Re-introduces the fix from [#7222](https://github.com/tuist/tuist/pull/7222) (bd9d2735) which was inadvertently lost when `NIOFileSystem` was removed, adapted to work without catching NIO-specific error types.

### Test plan

- [x] Build the CLI: tuist generate tuist --no-open && xcodebuild build -workspace Tuist.xcworkspace -scheme tuist CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""
- [x] Run unit tests: tuist test TuistLoaderTests — new tests write_succeedsWhenMakeDirectoryRacesButDirectoryExists, write_propagatesErrorWhenDirectoryCreationFailsAndDirectoryDoesNotExist, and write_doesNotCallRemoveBeforeWriting should pass
- [x] Verify single-process caching still works: run tuist generate twice on the same project, confirm the second invocation hits the manifest cache (faster, no re-compilation logged)

### DIFF
--- a/cli/Sources/TuistLoader/Loaders/CachedManifestLoader.swift
+++ b/cli/Sources/TuistLoader/Loaders/CachedManifestLoader.swift
@@ -291,20 +291,11 @@ public class CachedManifestLoader: ManifestLoading {
         guard let cachedManifestContent = String(data: cachedManifestData, encoding: .utf8) else {
             throw ManifestLoaderError.manifestCachingFailed(manifest, cachedManifestPath)
         }
-        try await write(cachedManifestContent: cachedManifestContent, to: cachedManifestPath)
-    }
-
-    private func write(cachedManifestContent: String, to cachedManifestPath: AbsolutePath) async throws {
-        if try await !fileSystem.exists(cachedManifestPath.parentDirectory, isDirectory: true) {
-            try await fileSystem.makeDirectory(at: cachedManifestPath.parentDirectory, options: [.createTargetParentDirectories])
-        }
-        if try await fileSystem.exists(cachedManifestPath) {
-            try await fileSystem.remove(cachedManifestPath)
-        }
-        try await fileSystem.writeText(
-            cachedManifestContent,
-            at: cachedManifestPath
+        try await fileSystem.makeDirectory(
+            at: cachedManifestPath.parentDirectory,
+            options: [.createTargetParentDirectories]
         )
+        try await fileSystem.writeText(cachedManifestContent, at: cachedManifestPath)
     }
 }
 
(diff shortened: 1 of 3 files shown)


cli/Sources/TuistLoader/Loaders/CachedManifestLoader.swift
danibachar: Just wrap the calling function - this would not solve the issue if `fileSystem.writeText` throws

ze-diaz: Updated

danibachar: This was not addressed - like in this [PR](https://github.com/tuist/tuist/commit/bd9d2735f00efe9e975607cd40aad87d2565916b). Wrap `try await write(cachedManifestContent: cachedManifestContent, to: cachedManifestPath)` with `ignoringAlreadyExists`

This change still fails if `fileSystem.writeText` throws

danibachar: Leave the coed in the `private func write(cachedManifestContent: String, to cachedManifestPath: AbsolutePath) async throws` the same

cli/Sources/TuistLoader/Loaders/CachedManifestLoader.swift
```diff
@@ -295,12 +295,10 @@ public class CachedManifestLoader: ManifestLoading {
     }
 
     private func write(cachedManifestContent: String, to cachedManifestPath: AbsolutePath) async throws {
```
danibachar: Revert all changes to this function

ze-diaz: Reverted

cli/Sources/TuistLoader/Loaders/CachedManifestLoader.swift
ze-diaz: Mb, now it's wrapped

pepicrft (approved):
Thanks @ze-diaz 🙏🏼. I ended up pushing the fix down to FileSystem since it feels it should be addressed at a lower layer. As a result, the changes here became much simpler.

PR #10670, opened 2026-05-07
fix(capi-scaleway): keep Aqua session alive for VZ HostKey creation

fortmarek:
## Summary

Canary deploys after #10657 were hanging on the same wall the IP-wait decouple was supposed to break, but for a different reason. Apple's Virtualization framework refuses to mint a fresh HostKey for a macOS guest unless the calling user holds a live **Aqua** (loginwindow/GUI) launchd session at the moment of `tart run`. Without one, every VM exits in <1s with `VZErrorDomain Code=-9 / Failed to create new HostKey` and the Pod stalls in `TartCreateFailed` forever.

The bootstrap pass already configures auto-login + SIGHUPs loginwindow once to bring the session up. What it didn't do is keep the session alive past the boot window or recover when it goes away. Canary's host had no `gui/501` at all when we logged in to investigate — `launchctl print gui/501` returned `Could not print domain: 125`. A single `sudo killall -HUP loginwindow` brought it back in 1s and the deploy went green immediately.

This PR adds two layers so we don't depend on a human kicking loginwindow:

### 1. `disableIdleSleep` at bootstrap

New step in `bootstrap.Run` that turns off the macOS facilities that tear the Aqua session down:
- `pmset -a sleep 0 displaysleep 0 disksleep 0` — host/display/disk idle sleep off (`-a` because Mac mini servers are AC-only but the default profile selection isn't trustworthy).
- `com.apple.screensaver idleTime 0` — screensaver off.
- `com.apple.screensaver askForPassword 0` — no screen lock.
- `.GlobalPreferences com.apple.autologout.AutoLogOutDelay 0` — disable inactivity auto-logout (off-by-default on consumer Macs but Scaleway-baked images sometimes ship with it on).

Failures are fatal — there's no point shipping a host that will silently fall over an hour after bootstrap.

### 2. `tart.Client.EnsureGUISession` preflight

Runtime check that runs before every `tart run`:
- `sudo -n launchctl print gui/$(uid)` → look for `session = Aqua`. If found, return immediately.
- Otherwise `sudo -n killall -HUP loginwindow` and poll for up to 30s.

Self-healing for the cases the bootstrap hardening can't cover: WindowServer crash, manual logout via SSH, OS-update reboot. Idempotent (skip when Aqua's already up). Loginwindow restart doesn't disturb running VMs — they're Setsid-detached and live in their own pgid.

Plumbed as a function field on `Client` (`EnsureGUISession func(ctx context.Context) error`) so existing tests that construct `Client` directly skip it; `New()` wires `EnsureRealGUISession`. Two new tests:
- `TestRunInvokesEnsureGUISessionBeforeStartingTart` — the preflight runs before `tart run`.
- `TestRunSurfacesEnsureGUISessionFailure` — preflight failure short-circuits `Run` without launching tart.

The runtime check uses `sudo -n` and depends on the NOPASSWD sudoers entry the bootstrap already writes via `enablePasswordlessSudo`. Without it both the check and the kick would fail fast on a password prompt.

## Test plan

- [x] `go test ./...` in both `infra/tart-kubelet` and `infra/macos-host-bootstrap`.
- [ ] After merge, `release.yml` cuts a new `capi-scaleway@<n>` and the auto-bump rewrites `macosFleet.image.tag` (now that #10669 unblocks `commit-and-release`).
- [ ] Reboot the canary Mac mini after the new operator image rolls out and confirm `gui/501` comes up at boot without manual SIGHUP, and that the next `xcresult-processor` Pod schedules cleanly without hitting `TartCreateFailed`.

### DIFF
--- a/infra/macos-host-bootstrap/bootstrap.go
+++ b/infra/macos-host-bootstrap/bootstrap.go
@@ -134,6 +134,9 @@ func Run(ctx context.Context, cfg Config) (string, error) {
 	if err := enableAutoLogin(ctx, client, cfg.SSHUser, cfg.UserPassword); err != nil {
 		return hk.observed(), fmt.Errorf("auto-login: %w", err)
 	}
+	if err := disableIdleSleep(ctx, client); err != nil {
+		return hk.observed(), fmt.Errorf("disable idle sleep: %w", err)
+	}
 	if cfg.NodeName != "" {
 		if err := setHostname(ctx, client, cfg.NodeName); err != nil {
 			return hk.observed(), fmt.Errorf("set hostname: %w", err)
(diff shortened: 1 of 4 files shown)


fortmarek:
@pepicrft merging, but let me know if you'd take a different approach/have some comments

PR #10672, opened 2026-05-07
fix(cli): propagate selective testing and module cache analytics across --build-only / --without-building

fortmarek:
### Context

When `tuist test --build-only` and `tuist test --without-building -testProductsPath …` run as separate processes (typically across CI machines via a shared `.xctestproducts` bundle), the test phase had no access to the build phase's `RunMetadataStorage`. The resulting test run reported:

- empty Module Cache tab
- empty Selective Testing tab
- no link back to the build run

This setup is uncommon — most users hit `tuist test` in a single process, where `RunMetadataStorage` survives end-to-end — but it's the standard split for the "build once, test on many shards/machines" CI topology, and the analytics gap is visible on the test run page.

### What changed

Added a new `RunMetadata` Codable struct that captures the analytics fields the test phase needs: graph, binary cache items, selective-testing cache items, target content hash subhashes, and `buildRunId`. The build phase writes it as `run-metadata.json` alongside the existing `selective-testing-graph.json` inside the `.xctestproducts` bundle whenever `--build-only` produces one. The test phase restores it best-effort before `xcodebuild test-without-building` runs, in both `runTestWithoutBuildingFromBundle` and the shard execute path.

When `TrackableCommand` later assembles the command event for upload, `RunMetadataStorage` is now populated as if the test phase had generated the project itself, so the test run gets:

- Selective Testing tab with the original cache items
- Module Cache tab with the binary cache items
- a `buildRunId` link to the originating build run

### Notes for reviewers

- **Backwards compatible:** the read-side is best-effort; bundles produced before this change have no run-metadata file and the test phase falls back to today's behavior.
- **Sharding:** `archiveXCTestProducts` only excludes `.dSYM`, so the snapshot file rides along with the bundle to the server and back.
- **Path keys:** the graph's `[AbsolutePath: Project]` keys and the cache-item dictionaries are produced together on the build machine. They match each other inside the snapshot regardless of where the test machine extracts the bundle, so `CommandEventFactory.map`'s lookups work on the test side.
- **Failure handling:** snapshot read/write failures log a warning and never abort the test run.

### E2E verification

In addition to the new `TuistKitTests` cases (run-metadata round-trip + missing-file fallback), I verified the full upload path against a real Tuist server.

Setup: a minimal Tuist-generated macOS app fixture pointing at `tuist/tuist` on production.

```bash
# Build phase — writes run-metadata.json into the bundle
tuist test --build-only App -- \
  -destination 'platform=macOS' \
  -derivedDataPath /tmp/dd \
  -testProductsPath /tmp/snapshot-e2e.xctestproducts

# Test phase — restores run metadata, uploads with the build-run link
tuist test --without-building App -- \
  -destination 'platform=macOS' \
  -testProductsPath /tmp/snapshot-e2e.xctestproducts
```

Confirmed end to end on `tuist.dev/tuist/tuist`:

- Build run created: https://tuist.dev/tuist/tuist/builds/build-runs/847d1736-6e7e-4436-9710-208b254ec410
- Run-metadata file contained `"buildRunId": "847D1736-6E7E-4436-9710-208B254EC410"` (matches the build-run UUID).
- Test phase logged `Skipping project generation, using selective testing graph from .xctestproducts bundle...` — confirms the restore path ran.
- HTTP request bodies (from `~/.local/state/tuist/sessions/.../logs.txt`) carried the same `build_run_id` in both uploads:
  - `POST /api/projects/tuist/tuist/tests` → `build_run_id: 847D1736-…`
  - `POST /api/analytics?project_id=tuist%2Ftuist` → `build_run_id: 847D1736-…`
- Server response on the analytics upload redirected to `/builds/build-runs/847d1736-…` — this branch in `analytics_controller.ex` only fires when `build_run_id` is non-nil, so the link was persisted.
- Test run: https://tuist.dev/tuist/tuist/tests/test-runs/6699b64b-eacf-448f-a75b-50c64749cdb5 — opens with the "Build: {scheme}" link button driven by `:if={@run.build_run}` in `test_run_live.html.heex`.

### DIFF
--- a/cli/Sources/TuistCore/Analytics/RunMetadata.swift
+++ b/cli/Sources/TuistCore/Analytics/RunMetadata.swift
@@ -2,16 +2,16 @@ import Foundation
 import Path
 import XcodeGraph
 
-/// A snapshot of run metadata captured during the build phase
-/// (`tuist test --build-only`) and restored during the test execution phase
-/// (`tuist test --without-building`) when the two run as separate processes,
-/// typically across CI machines via a shared `.xctestproducts` bundle.
+/// Run metadata captured during the build phase (`tuist test --build-only`) and
+/// restored during the test execution phase (`tuist test --without-building`)
+/// when the two run as separate processes, typically across CI machines via a
+/// shared `.xctestproducts` bundle.
 ///
 /// Persisting this lets the test phase upload the same module-cache,
 /// selective-testing, and build-run analytics that the build phase produced,
 /// so the resulting test run page links back to the build run and surfaces
 /// the cache tabs even when the actual `xcodebuild` invocation runs no tests.
-public struct RunMetadataSnapshot: Codable {
+public struct RunMetadata: Codable {
     public let graph: Graph?
     public let binaryCacheItems: [AbsolutePath: [String: CacheItem]]
     public let selectiveTestingCacheItems: [AbsolutePath: [String: CacheItem]]
@@ -32,5 +32,5 @@ public struct RunMetadataSnapshot: Codable {
         self.buildRunId = buildRunId
     }
 
-    public static let fileName = "run-metadata-snapshot.json"
+    public static let fileName = "run-metadata.json"
 }
--- a/cli/Sources/TuistCore/Analytics/RunMetadataSnapshot.swift
+++ b/cli/Sources/TuistCore/Analytics/RunMetadataSnapshot.swift
@@ -1,0 +1,36 @@
+import Foundation
+import Path
+import XcodeGraph
+
+/// A snapshot of run metadata captured during the build phase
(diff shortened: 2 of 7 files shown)


pepicrft (approved):

ivan-gaydamakin:
@fortmarek We first run only the build, and only after that we run the tests.
Because on our self-hosted CI/CD we do not reset the virtual machine every time, the size of the `~/Library/Developer/XCTestDevices/` folder keeps growing (it even reached 2 TB with a 300 GB SSD :D).

Because of this, sometimes when tests start, they can freeze during launch.
We use a 5-minute timeout to detect that the test is stuck, then we kill the `xcodebuild` process and start the tests again without rebuilding. 

And “just copying the cache” does not work for us, because our cache size is around 10–20 GB :O.

PR #10674, opened 2026-05-07
feat(server): add rolling window option to flaky test automations

fortmarek:

Adds a "Rolling window" option to the flaky-test automation form alongside the existing "Last days" mode. Until now, an automation could only fire based on a calendar window like `30d`. With rolling windows, the condition is evaluated against the latest N runs per test case — useful for projects whose test volume varies across days, where a fixed time window can over- or under-react to recent activity.

The shape of `trigger_config` now carries an explicit `window_type`:

- `"last_days"` (default for backwards compatibility): existing `window: "30d"` behaviour, evaluated against `test_case_run_daily_stats_per_case`.
- `"rolling"`: new `rolling_window_size: 100` integer, evaluated against `test_case_runs` directly with a `row_number() OVER (PARTITION BY test_case_id ORDER BY ran_at DESC)` subquery.

Existing alerts without a `window_type` continue to work as before; validation defaults missing values to `"last_days"`.

The same window-type dropdown is wired up to the Recovery section so users can wait either "X days without a re-trigger" or "N new runs without a re-trigger" before recovery actions fire. The evaluation worker counts test_case_runs since the original trigger event when recovery is rolling.

### How to test locally

1. Boot the server (`mise run dev`) and open a project's automation settings (`/<account>/<project>/settings/automations`).
2. Open "Add automation". The condition row now has an extra "Over" dropdown with two options:
   - **Last days**: shows the existing `30d`-style text input.
   - **Rolling window**: shows a numeric "Last N runs" input.
3. Toggle Recovery on and confirm "Without trigger for" exposes the same dropdown (Last days / Rolling window).
4. Save an automation in each mode and confirm the trigger summary on the table reflects the chosen window (`over 30d` vs `over the last 100 runs`).
5. Edit a saved automation and confirm the form repopulates the right input based on the persisted `window_type`.
6. Trigger evaluation against a project with test case runs and confirm flakiness rate / flaky run count thresholds resolve based on the most recent N runs in rolling mode.

Test coverage is added for:

- Schema validation of `window_type`, `rolling_window_size`, and the legacy `window` field (`server/test/tuist/automations/alerts/alert_test.exs`).
- Monitor evaluation for both `flakiness_rate` and `flaky_run_count` in rolling mode (`server/test/tuist/automations/monitors/flaky_tests_monitor_test.exs`).
- Worker behaviour for rolling recovery (count-based cooldown) (`server/test/tuist/automations/workers/alert_evaluation_worker_test.exs`).
- LiveView form persistence and invalid-value handling for both trigger and recovery (`server/test/tuist_web/live/project_automations_live_test.exs`).

Run the targeted suites:

```bash
cd server
mix test test/tuist/automations/alerts/alert_test.exs \
         test/tuist/automations/monitors/flaky_tests_monitor_test.exs \
         test/tuist/automations/workers/alert_evaluation_worker_test.exs \
         test/tuist_web/live/project_automations_live_test.exs
```


server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
```diff
@@ -179,6 +207,71 @@ defmodule Tuist.Automations.Monitors.FlakyTestsMonitor do
...
+  # already absorbed the row, so a re-inserted run (e.g. is_flaky updated
+  # later) appears twice in the `groupArrayLast` array. That's bounded
+  # noise — ≤1% at the default window — well within the natural variance
+  # of a flakiness threshold.
+  #
+  # `monitor_type` and `comparison` are interpolated because each is
+  # constrained to a fixed allowlist, so there is no SQL-injection vector.
+  # Numeric inputs (`project_id`, `size`, `threshold`) flow through bound
+  # parameters.
+  defp rolling_triggered_test_case_ids(project_id, monitor_type, size, threshold, comparison) do
```
fortmarek, line 233: I'm not an extreme fan of this, but this felt like the best way to make the rolling window automation performant

server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
```diff
@@ -179,6 +207,69 @@ defmodule Tuist.Automations.Monitors.FlakyTestsMonitor do
...
+
+    case ClickHouseRepo.query(sql, %{project_id: project_id, size: size, threshold: threshold * 1.0}) do
+      {:ok, %{rows: rows}} ->
+        # ClickHouse returns UUID columns as 16-byte binaries here; the rest
+        # of the worker compares against string-encoded UUIDs from the Ecto
+        # path, so normalise.
+        Enum.map(rows, fn [binary] -> Ecto.UUID.load!(binary) end)
+
+      _ ->
+        []
```
pepicrft: This path should not turn a ClickHouse error into an empty triggered set. If this query returns `{:error, _}` because the MV is missing, the SQL changes, or ClickHouse has a transient issue, the worker treats every active event as no longer triggered and can run recovery actions. Let the error raise so Oban retries, matching the existing `ClickHouseRepo.all` paths.

server/priv/ingest_repo/migrations/20260508130000_create_test_case_runs_recent_per_case_mv.exs
```diff
@@ -0,0 +1,110 @@
...
+    backfill_by_partition()
+
+    # excellent_migrations:safety-assured-for-next-line raw_sql_executed
+    IngestRepo.query!("""
+    CREATE MATERIALIZED VIEW IF NOT EXISTS test_case_runs_recent_per_case_mv
+    TO test_case_runs_recent_per_case
+    AS SELECT
+      project_id,
+      assumeNotNull(test_case_id) AS test_case_id,
+      groupArrayLastState(#{@max_window_size})((ran_at, toUInt8(is_flaky))) AS recent_runs
```
pepicrft, line 62: This aggregate caps before we order by `ran_at`, so it does not strictly preserve the last N runs by run time. `groupArrayLastState` keeps the last values in ingestion and merge order, and the backfill queries do not specify an order. Sorting the merged array later only sorts the rows that survived the 1000-entry cap. Delayed uploads or ReplacingMergeTree re-inserts of older runs can push newer runs out of the state. We need an order-aware aggregate or a query path that orders by `ran_at` before applying the cap.

pepicrft (approved):
Approved with two comments.

PR #10675, opened 2026-05-07
feat(capi-scaleway): auto-roll xcresult-processor on fleet-shape change

fortmarek:
## Summary

Today's production fleet bring-up surfaced a sharp edge in the k8s scheduler model: it places Pods at admission time and **never rebalances them on its own**, so when a new Mac mini joins (or leaves) the fleet, the xcresult-processor Pods stay where the scheduler put them at the previous topology.

Concretely: production deployed with `replicas: 2` for the xcresult-processor, fleet-0 came Ready first, both replicas landed on it (legitimately — `whenUnsatisfiable: ScheduleAnyway` lets that happen). When fleet-1 came Ready ~50 min later, nothing moved. We had to `kubectl rollout restart deploy/...-xcresult-processor` by hand to let the topology spread relocate one Pod onto the new host.

This PR makes that automatic.

## How

New `FleetSpreadReconciler` in the CAPI operator. It watches `ScalewayAppleSiliconMachine`, computes a stable hash of the Ready members, and patches a target Deployment's `spec.template.metadata.annotations[tuist.dev/fleet-hash]` whenever the hash changes. K8s sees the new annotation, mints a new ReplicaSet, the rolling update fires, and the Deployment's existing `topologySpreadConstraints` redistribute Pods across the now-current host set.

Steady-state reconciles are no-ops (idempotent patch on matching hash).

### Hash semantics

- Only `Ready`, non-deleting machines count. A machine flapping through `Provisioning`/`Bootstrapping` doesn't roll the workload until the host actually comes up; a machine being torn down stops counting the moment it gets a `DeletionTimestamp` rather than when its finalizer finally clears.
- Hash is order-independent (sorted before SHA), so reordering between List calls doesn't churn ReplicaSets.

### Wiring

Operator binary:
- `--fleet-spread-deployment` / `--fleet-spread-namespace` flags. Empty disables the controller (OSS deployments without a workload to spread).
- `mgr.GetAPIReader()` for the Deployment Get — bypasses controller-runtime's informer cache so we don't need cluster-wide `list,watch` on Deployments. The `ScalewayAppleSiliconMachine` List goes through the normal cache.

Chart:
- Adds the two args under `{{- if .Values.xcresultProcessor.enabled }}`, pointing at the rendered xcresult-processor Deployment name.
- ClusterRole gets `get,patch` on `apps/deployments` *scoped via `resourceNames`* to that single Deployment — a compromised operator can't patch arbitrary workloads.

## Tests

`go test ./controllers/...` covers:
- Hash stability across reorderings, distinguishing membership changes, excluding non-Ready, excluding terminating machines (delete-timestamp set).
- Reconcile stamps the annotation on first Ready transition.
- Reconcile is idempotent — `ResourceVersion` doesn't budge when the hash already matches.
- Reconcile re-patches on fleet growth (the production scenario this PR is for).
- Reconcile tolerates a missing target Deployment (chart-bring-up race; helm orders resources by kind and CRs may sort before Deployments).

## Test plan

- [x] `go test ./...` in the operator module.
- [x] `helm template` confirms args + scoped RBAC render correctly.
- [ ] After the next operator release lands and rolls onto canary/production: scale `macosFleet.replicas` up by 1 (or remove a machine) and verify the xcresult-processor Deployment annotation flips and a new ReplicaSet rolls without manual intervention.
- [ ] Verify the operator pod doesn't grow CPU/memory under the new watch (List of all machines on every machine event in a small fleet should be ~free, but worth eyeballing in steady state).

## Worth flagging (potential follow-ups)

- **Rollout storms**: if a Mac mini flaps Ready/NotReady (network blip, partial bootstrap failure), the Deployment will roll repeatedly. With PDB `minAvailable: 1` worst case is a single Pod down at a time, but a debounce / "only react to the Ready set growing" mode could be added later.
- **Other workloads**: only one target Deployment per operator instance today. If we ever pin more workloads to the macOS fleet, generalize to a list.

### DIFF
--- a/infra/cluster-api-provider-scaleway-applesilicon/cmd/manager/main.go
+++ b/infra/cluster-api-provider-scaleway-applesilicon/cmd/manager/main.go
@@ -69,6 +69,9 @@ func main() {
 		tartKubeletMaxUpdateAttempts int
 
 		machineMaxConcurrentReconciles int
+
+		fleetSpreadDeployment string
+		fleetSpreadNamespace  string
 	)
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8080", "Prometheus metrics endpoint")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "Liveness/readiness probe endpoint")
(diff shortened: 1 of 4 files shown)


pepicrft (approved):
Reviewed the fleet-spread controller and Helm wiring. Verified locally:

- `mise exec go@1.25 -- go test ./...` from `infra/cluster-api-provider-scaleway-applesilicon`
- `mise exec -- helm lint infra/helm/tuist --set macosFleet.enabled=true --set xcresultProcessor.enabled=true`
- `mise exec -- helm template review infra/helm/tuist --namespace tuist --set macosFleet.enabled=true --set xcresultProcessor.enabled=true`
- rendered with `xcresultProcessor.enabled=false` and confirmed the fleet-spread RBAC and args are omitted

Looks good to me.

PR #10677, opened 2026-05-07
fix(server): restore Oban metrics in Grafana Cloud

fortmarek:
### Why

Since the Render → Syself migration the **Tuist – PromEx Oban** dashboard has been empty and the `ProcessXcresultWorker` / `ProcessBuildWorker` discarded-job alerts have been firing as `DatasourceNoData`. Three layered root causes:

1. **Annotation namespace mismatch.** The Grafana `k8s-monitoring` chart's `annotationAutodiscovery` feature defaults to the `k8s.grafana.com/*` annotation namespace, but our Pods carry the legacy `prometheus.io/*` annotations. So Alloy never discovered any Tuist Pod and not a single `tuist_oban_*` series reached Grafana Cloud.
2. **Duplicate scrape targets.** Pod-role discovery emits one target per declared container port (server pods have `http` + `metrics` + `epmd`); the rewrite to `:9091` collapsed the addresses but the targets still re-evaluated the relabel pipeline once per port.
3. **xcresult-processor pods unreachable.** They run inside Tart VMs on Mac mini nodes (`kubernetes.io/os: darwin`); Alloy is a Linux DaemonSet, the VM IPs sit behind the Mac mini's NAT bridge, and tart-kubelet had no host-side proxy to bridge them.

### What

**`infra/helm/k8s-monitoring/values.yaml`** — point `annotationAutodiscovery` at the `prometheus.io/*` namespace and add `metricsPortName: prometheus.io/port-name` so the chart's `keepequal` filter trims server / processor pods to a single target per pod. Verified via `helm template`.

**`infra/helm/tuist/templates/{server,processor}-deployment.yaml`** — add `prometheus.io/port-name: metrics` so the new filter has something to key on.

**`infra/tart-kubelet/`** — new in-process HTTP reverse proxy started per scrape-annotated VM Pod. Listens on `<node_ip>:<port>` and forwards to `<vm_ip>:<port>`, resolving the upstream IP via `tart ip` per request so a VM restart / new DHCP lease is transparent. The Pod's `Status.PodIP` is rewritten to `NodeIP`, so Alloy's existing pod-IP-based annotation autodiscovery just works without per-pod chart overrides. New `--node-ip` flag with auto-detect that walks UP non-loopback interfaces and skips `bridge*` / `vmnet*` so a kubelet restart with a running VM doesn't pick the Tart NAT bridge.

(HTTP reverse proxy rather than raw TCP — Prometheus reuses connections via HTTP keep-alive, which deadlocks both copy goroutines on idle keep-alive sockets in a naive bidirectional `io.Copy`. The reverse proxy handles each HTTP request as an independent operation.)

**`infra/helm/tuist/templates/xcresult-processor-deployment.yaml`** — opts in via `prometheus.io/scrape: "true"` + `prometheus.io/port: "9091"`, mirroring the other workload Pods. Container declares `hostPort: 9091` on the named `metrics` port so the kube-scheduler enforces hostPort uniqueness across Nodes (one xcresult-processor Pod per Mac mini, full stop). Rollout strategy moves to `maxSurge: 0 / maxUnavailable: 1` because surge-then-replace can't satisfy hostPort uniqueness on a fleet whose Nodes are already at capacity. `topologySpread` flips to `DoNotSchedule` for self-consistency. PDB.minAvailable=1 keeps a consumer up across the rollout. The forwarder bind + `PodIP=NodeIP` rewrite are gated on the forwarder actually starting, so a runtime collision (whatever the cause) leaves PodIP at the unreachable VM IP — Alloy harmlessly times out instead of mis-scraping the wrong VM.

**Source-IP allowlist on the forwarder.** `r.NodeIP` on Scaleway can be the Mac mini's public IP (no private interface that's still routable to Linux Alloy DaemonSet Pods elsewhere in the cluster), so the bind interface alone isn't a security boundary. Default allowlist is RFC1918 + IPv4 loopback + IPv4 link-local + IPv6 ULA + link-local + loopback — covers any realistic cluster Pod or Node CIDR while clamping out the WAN. Repeatable `--scrape-allowed-cidr` flag overrides for the rare cluster that routes Pod traffic through public IPs. Path filter restricts the proxy to `GET /metrics` only; `tart ip` resolution is cached for 30s so a burst of scrapes (or anyone hammering the proxy) doesn't translate 1:1 into subprocess invocations on the host.

**`server/lib/tuist/oban/prom_ex_plugin.ex`** — drops the custom `tuist_oban_jobs_recent_count` polling metric and its test. The canonical PromEx event counters now flow from xcresult-processor pods directly, and `sum by (worker, state) (increase(tuist_oban_job_processing_duration_milliseconds_count[30m]))` is the alert query.

**`infra/grafana-dashboards/xcresult-processor.json`** — single-pane Tuist xcresult Processor dashboard now that the pods are scrapeable: active consumers, pending jobs, throughput, processing/queue latency p50/p95/p99, queue depth by state, exception rate by error class, attempts-to-completion, and discards/cancels rate. All sourced from canonical PromEx event metrics, no custom series. Wrapped in the Grafana Git Sync v1 envelope so the **Tuist Dashboards** folder picks it up on next sync.

### What this isn't (and why we're not doing it now)

The kubelet-faithful answer to "make scraping work for VM Pods" is **a CNI plugin that integrates Tart VMs into the cluster Pod network** — same model Linux Pods have via Cilium. With that in place, `Pod.Status.PodIP` would be the real, routable address, Alloy's annotation autodiscovery would Just Work without any rewriting, and we'd delete the host forwarder + the PodIP-rewrite hack + the source-CIDR allowlist + the hostPort uniqueness constraint.

**It's not a small project.** A focused research pass turned up:

- **No CNI plugin for Tart or for macOS hosts exists in any public repo** (cirruslabs/tart, Anka, Veertu, Krustlet, kubevirt, kubernetes-sigs all checked). Whoever builds this builds from scratch.
- **macOS has no Linux-equivalent CNI building blocks.** No veth / netns / eBPF on XNU. `pf` is [deprecated by Apple](https://developer.apple.com/forums/thread/781996); no native TUN/TAP (`tuntaposx` is kext-based and Apple is removing kexts); Network Extensions explicitly can't intercept and modify packets. Sanctioned VM-network primitives are `vmnet.framework` and `VZFileHandleNetworkDeviceAttachment` (Tart's `--net-fd`) — both userspace.
- **Cilium runs in VXLAN tunnel mode on this cluster** ([cilium-values.yaml:21-22](infra/k8s/mgmt/bootstrap/cilium-values.yaml#L21-L22)), which is a Linux-kernel datapath — the Tart VM can't just join the VXLAN mesh. We'd flip Cilium to its WireGuard mode cluster-wide first (a one-line Helm value change + agent rollout, ~1 day), then have the VM peer into the mesh as a regular WG node.

The minimum-viable shape ("path A") is **WireGuard inside the VM, the Mac mini host stays out of the dataplane**: tart-kubelet allocates a Pod IP from the cluster CIDR, injects it + a WG config into the VM via cloud-init, the VM joins the cluster mesh as a peer, advertises its Pod IP back via PodStatus. The host-side "CNI plugin" is mostly bookkeeping (IPAM + cloud-init + the CNI-spec ADD/DEL shim the kubelet calls). **Realistic estimate: 3-4 weeks.** Breakdown:

- IPAM allocation in tart-kubelet (1-2 days)
- Cloud-init / Tart guest-init for Pod IP + WG config injection (3-5 days)
- CNI-spec-conformant binary the kubelet calls (3-5 days)
- Cilium VXLAN → WG flip + Linux Node rollout coordination (1-2 days)
- Node-side route announcement so Cilium peers learn the new Pod IPs (2-3 days)
- Removing the host forwarder + PodIP rewrite + source-CIDR allowlist (1 day)
- End-to-end test pass (Services, NetworkPolicy, kube-state-metrics, Alloy autodiscovery) (3-5 days)

**Why we're skipping it for now:** Apple's macOS SLA caps Tart at 2 concurrent VMs per host, and the xcresult-processor is sized at ~12 vCPU / 24 GiB — most of one Mac mini. The fleet is sized **1 Pod = 1 Mac mini** for HA (a single host losing power doesn't take xcresult capacity to zero), not for horizontal scale. As long as the workload pattern stays "one VM Pod per host," the hostPort uniqueness constraint costs us nothing operationally and the host forwarder is ~300 lines of contained Go. If a second VM-backed workload ever wants to share a Mac mini with xcresult-processor, that's the moment a real CNI starts paying for itself; until then, the simpler hack is the right trade.

(A heavier "path B" — host-side dataplane via Tart's `--net-fd` so the host can enforce per-VM NetworkPolicy — would push the estimate to 6-10 weeks. Not on the radar unless NetworkPolicy on VM Pods becomes a requirement.)

### Rollout

Auto-cascades on merge:

- `release.yml` detects changes under `infra/tart-kubelet/**` (and friends), bumps the `capi-scaleway` version, builds + pushes a new operator image with the new tart-kubelet binary baked in, patches `infra/helm/tuist/values-managed-common.yaml` to pin the new tag.
- The bump commit triggers `server-production-deployment.yml` → canary → acceptance tests → production. Helm rolls out the new operator image.
- The operator's drift loop notices `status.tartKubeletBinarySHA` differs from its embedded binary's SHA and re-uploads + reloads launchd on each Mac mini. Zero-downtime: the running Tart VM is detached via `setsid`, so launchd reload doesn't kill it, and the new tart-kubelet's startup recovery pass re-binds the existing VM.

### Follow-up in Grafana UI (not in this PR)

Repoint the `ProcessXcresultWorker` / `ProcessBuildWorker` discarded-job alerts off the legacy queries and onto:

```promql
sum by (worker) (
  increase(tuist_oban_job_processing_duration_milliseconds_count{
    state="discarded",
    worker="Tuist.Tests.Workers.ProcessXcresultWorker"
  }[30m])
) > 0
```

(and analogously for `ProcessBuildWorker`). `sum` aggregates across xcresult-processor replicas correctly because each pod's counter only covers the jobs it processed (Oban claims jobs with `SKIP LOCKED`).

### How to test locally

- `cd infra/tart-kubelet && go test ./...` — covers the HTTP reverse proxy (relay, per-request resolution, /metrics path filter, GET-only method filter, source-CIDR allow / deny, default CIDR coverage, resolver caching within TTL, 502 on resolve failure, idempotent Stop).
- `helm dependency update infra/helm/k8s-monitoring && helm template infra/helm/k8s-monitoring -f infra/helm/k8s-monitoring/values-production.yaml | grep -A4 'annotation_autodiscovery_pods'` — confirm the relabel `keep` rule sources from `__meta_kubernetes_pod_annotation_prometheus_io_scrape`.


infra/tart-kubelet/internal/podagent/proxy.go
```diff
@@ -0,0 +1,262 @@
...
+		return nil, errors.New("resolve function is required")
+	}
+	l, err := net.Listen("tcp", listenAddr)
+	if err != nil {
+		return nil, fmt.Errorf("listen on %s: %w", listenAddr, err)
+	}
+
+	cached := newCachedResolver(resolve, 30*time.Second)
+
+	rp := &httputil.ReverseProxy{
```
pepicrft, line 98: `govulncheck` reports this new reverse-proxy path as reaching `GO-2026-4918` through `golang.org/x/net` with the current module version. The fix is dependency-only: bump `golang.org/x/net` in the tart-kubelet module to at least `v0.53.0` before shipping this listener.

infra/tart-kubelet/internal/podagent/proxy.go
```diff
@@ -0,0 +1,262 @@
...
+				*req = *req.WithContext(ctx)
+				return
+			}
+			req.URL.Scheme = "http"
+			req.URL.Host = target
+			req.Host = target
+		},
+		ErrorHandler: func(w http.ResponseWriter, req *http.Request, err error) {
+			if rerr, ok := req.Context().Value(resolveErrKey{}).(error); ok && rerr != nil {
+				http.Error(w, "tart-kubelet: upstream resolve failed: "+rerr.Error(), http.StatusBadGateway)
```
pepicrft: Avoid returning the raw upstream or resolver error to the HTTP response here. This endpoint is reachable by any source in the allowed CIDRs, and these errors can include VM names, IPs, paths, or Tart command output. A generic 502 body with the detailed error logged server-side keeps the scrape surface smaller.

infra/tart-kubelet/internal/podagent/proxy.go
```diff
@@ -0,0 +1,262 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if !c.cachedAt.IsZero() && time.Since(c.cachedAt) < c.ttl {
+		return c.target, c.cachedErr
+	}
+
+	target, err := c.fn()
+	c.target = target
+	c.cachedErr = err
```
pepicrft: Negative results are cached for the full 30s here. If `tart ip` has one transient failure during a VM restart or DHCP churn, every scrape returns 502 until the TTL expires even if the VM gets a new IP immediately after. Cache successful targets, but retry errors on the next request or give them a much shorter TTL.

infra/helm/tuist/values.yaml
```diff
@@ -313,10 +313,15 @@ xcresultProcessor:
...
-      # Surge by 1 so a new VM comes up before the old one drains; keeps
-      # at least 1 VM consuming the queue across deploys.
-      maxSurge: 1
-      maxUnavailable: 0
+      # Terminate-then-replace, not surge-then-terminate: the Pod
+      # declares hostPort 9091 (so the host-side metrics forwarder
+      # in tart-kubelet has a single owner per Mac mini) and the
+      # scheduler refuses to bind the same hostPort to two Pods on
+      # the same Node. Surging would deadlock on a fleet whose
+      # nodes are all already at capacity. PDB.minAvailable=1 keeps
```
pepicrft: This comment relies on `PDB.minAvailable=1`, but the default values keep the PDB disabled. With `replicaCount=1`, this rollout strategy can take the only queue consumer down. Either make the PDB default match the comment or call out that single-replica deployments have rollout downtime.

pepicrft (commented):
Runtime checks I ran locally: `go test ./...`, `go test -race ./...`, and `go vet ./...` pass for the Go modules. `govulncheck` still flags the reachable `x/net` issue called out inline.

pepicrft (approved):

PR #10679, opened 2026-05-07
fix(server): make all non-web pods leader-ineligible in Oban

fortmarek:
## Summary

Two related fixes to Oban scheduling.

### 1. Make all non-web pods leader-ineligible

The xcresult-processor pod role joined the same Oban cluster as the web and build-processor pods but was never added to the `peer: false` guard introduced in #10530, leaving it leader-eligible with an empty crontab. When it wins the peer election, every scheduled job — `AutomationScheduler`, `AlertWorker`, `SyncStripeMetersWorker`, `ExpireStaleTestRunsWorker`, the daily/hourly Slack reports — silently stops firing.

Symptom: in production Oban Web, no executed-count for any cron worker over a multi-hour window.

`runtime.exs` had two parallel invariants expressed as denylists of non-leader pod roles:

- crontab populated unless `processor_mode?()` or `xcresult_processor_mode?()`
- `peer: false` set if `processor_mode?()`

When the xcresult-processor role landed (#10499), the first guard was extended but the second wasn't. Result: an xcresult-processor pod could (and did) win the leader election, run `Oban.Plugins.Cron` against an empty schedule, and prevent any cron jobs from being inserted.

Flipped both guards to an allowlist keyed off `Tuist.Environment.web?()`. Any future pod role (`:scheduler`, `:ingest`, etc.) is now leader-ineligible by default and has to opt in explicitly.

### 2. Run project-level crons on self-hosted deployments

The crontab was wholesale-gated on `tuist_hosted?()`, which meant self-hosted deployments never ran `AutomationScheduler`, `AlertWorker`, per-project Slack reports, or stale-test-run cleanup — even though those are project-level features any operator opts into.

Split the crontab into `shared_crons` (runs on any web deployment in prod/stag/can) and a hosted-only layer (Tuist's internal Slack ops reports, account-usage rollup feeding Stripe, Stripe metered billing). Self-hosted deployments now get the project-level schedule; Tuist-hosted gets both layers.

| Worker | Self-hosted | Tuist-hosted |
| --- | --- | --- |
| `Slack.Workers.ReportWorker` (per-project Slack) | ✅ | ✅ |
| `Alerts.Workers.AlertWorker` (per-project alert rules) | ✅ | ✅ |
| `Tests.Workers.ExpireStaleTestRunsWorker` (sharded-run cleanup) | ✅ | ✅ |
| `Automations.Workers.AutomationScheduler` | ✅ | ✅ |
| `Ops.DailySlackReportWorker` (internal Slack) | — | ✅ |
| `Ops.HourlySlackReportWorker` (internal Slack) | — | ✅ |
| `Accounts.Workers.UpdateAllAccountsUsageWorker` (Stripe rollup) | — | ✅ |
| `Billing.Workers.SyncStripeMetersWorker` (Stripe) | — | ✅ |

## How to test locally

Config-only change. Verify the runtime config evaluates as expected for each role:

```bash
TUIST_MODE=xcresult_processor TUIST_DEPLOY_ENV=prod TUIST_HOSTED=1 \
  iex -S mix run --no-start -e 'IO.inspect(Application.get_env(:tuist, Oban))'
```

Expect `peer: false` and empty crontab. Repeat with `TUIST_MODE=processor` (same expectation), unset `TUIST_MODE` + `TUIST_HOSTED=1` (full crontab, no `peer:` override), and unset `TUIST_MODE` + unset `TUIST_HOSTED` (shared crontab only — no Stripe/usage/Ops entries).

In production, after rollout: watch the `tuist@*` nodes in Oban Web — only web pods should appear, and `AutomationScheduler` / `AlertWorker` should resume firing on their cron cadence.

### DIFF
--- a/server/config/runtime.exs
+++ b/server/config/runtime.exs
@@ -470,23 +470,25 @@ config :tuist, Oban,
     {Oban.Plugins.Lifeline, rescue_after: to_timeout(minute: 30)},
     {Oban.Plugins.Cron,
      crontab:
-       cond do
-         not (Tuist.Environment.web?() and env in [:prod, :stag, :can]) ->
-           []
-
-         Tuist.Environment.tuist_hosted?() ->
-           [
+       if Tuist.Environment.web?() and env in [:prod, :stag, :can] do
+         hosted_only_crons =
+           if Tuist.Environment.tuist_hosted?() do
              # Tuist-hosted-only: report into Tuist's own Slack workspace,
              # roll up account usage that feeds Stripe metered billing,
              # and reconcile Stripe meters from the rolled-up usage.
-             {"0 10 * * 1-5", Tuist.Ops.DailySlackReportWorker},
-             {"0 * * * 1-5", Tuist.Ops.HourlySlackReportWorker},
-             {"@daily", Tuist.Accounts.Workers.UpdateAllAccountsUsageWorker},
-             {"@daily", Tuist.Billing.Workers.SyncStripeMetersWorker}
-           ] ++ shared_crons
+             [
+               {"0 10 * * 1-5", Tuist.Ops.DailySlackReportWorker},
+               {"0 * * * 1-5", Tuist.Ops.HourlySlackReportWorker},
+               {"@daily", Tuist.Accounts.Workers.UpdateAllAccountsUsageWorker},
+               {"@daily", Tuist.Billing.Workers.SyncStripeMetersWorker}
+             ]
+           else
+             []
+           end
 
-         true ->
-           shared_crons
+         hosted_only_crons ++ shared_crons
+       else
+         []
        end}
   ]
 
--- a/server/config/runtime.exs
+++ b/server/config/runtime.exs
@@ -435,18 +435,21 @@ oban_queues =
   end
 
 # Leader-only Oban work (Cron, Pruner, Lifeline, Oban.Met.Reporter) runs on
-# whichever node wins the peer election. Server pods are always running and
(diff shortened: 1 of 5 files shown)


pepicrft:
Looked at how to make the gate regression-proof. The two gates (crontab and `peer: false`) now both key off `Tuist.Environment.web?()`, which is great, but if someone reintroduces a denylist down the road, nothing fails until the next non-web mode ships and an incident reproduces.

One option: lift the cron list and peer rule into pure functions (something like `Tuist.Oban.RuntimeConfig.crontab/3` and `peer_eligible?/1`) and unit-test them against every value of `Tuist.Environment.mode/0`. CI would then catch any future drift before it lands in prod, instead of relying on the manual `iex` check from the PR description.

pepicrft:
Tangent surfaced while reviewing this. `Tuist.Environment.mode/0` silently falls back to `:web` for any unknown `TUIST_MODE`. I checked locally with a tiny script:

```
TUIST_MODE=(unset)            => :web
TUIST_MODE=processor          => :processor
TUIST_MODE=xcresult_processor => :xcresult_processor
TUIST_MODE=processsor (typo)  => :web
TUIST_MODE=ingest             => :web
```

So a typo or a partially-wired new mode in a deployment manifest would land that pod in leader eligibility with the full crontab, which is the exact failure shape this PR is fixing. Worth tightening `mode/0` to raise on unknown non-empty values so a misconfig fails the pod fast at boot? Could be a follow-up.

pepicrft (approved):

PR #10686, opened 2026-05-08
fix(server): retry transient ClickHouse transport errors marking flaky runs

fortmarek:
Fixes TUIST-DA

## Summary

- Extract the retry helper that already lived in `Tuist.IngestRepo` into a shared `Tuist.ClickHouseRetry` module.
- `Tuist.IngestRepo.with_retry/2` now delegates there (preserves the `Tuist.Ingestion.Buffer` call site).
- `Tuist.ClickHouseRepo` overrides `all/1,2` and `one/1,2` and funnels them through the same helper, so every read of the read-repo gets the same retry behavior writes already had.

## Why

Sentry [TUIST-DA](https://tuist.sentry.io/issues/116250005/) is `Mint.TransportError: socket closed` (or `: timeout`) raised from `ClickHouseRepo.all/1` inside `ProcessXcresultWorker`. Every event comes from the Scaleway Mac mini Tart VMs - the `:process_xcresult` queue runs there exclusively after [#10499](https://github.com/tuist/tuist/pull/10499). Hetzner web pods don't see this in Sentry because their pool is hot from constant traffic; the bursty xcresult-processor sits idle long enough for Scaleway egress NAT to drop pool sockets, and the next checkout grabs the dead one.

Writes were already protected: `Tuist.IngestRepo` overrides `insert_all` / `insert` / `all` to wrap them in a `with_retry` that catches `Mint.TransportError` + `DBConnection.ConnectionError` with exponential backoff (100/200/400ms, 4 attempts total). Reads through `Tuist.ClickHouseRepo` had no such cover, so a single transport blip burned the worker's full 5 Oban attempts (>16 min of backoff) before the job was finally marked failed.

The first cut of this PR tried two dead ends before landing on the shared-helper shape:

1. **TCP keepalive raw socket options** (reverted). Mint's TCP and SSL transports both call `Keyword.merge(opts, defaults)`, and Elixir 1.19's `Keyword.merge/2` validates both arguments as keyword lists - `{:raw, _, _, _}` 4-tuples raised `ArgumentError` before any socket could open.
2. **Per-call-site retry wrap** (reverted). Fixed exactly the one query Sentry was flagging and left every other `ClickHouseRepo.all` / `one` call exposed to the same flap.

## Test plan

- [x] `Tuist.ClickHouseRetryTest` covers success, `Mint.TransportError` retry-then-success, `DBConnection.ConnectionError` retry-then-success, exhaustion, and non-matching exception passthrough.
- [ ] Canary stays green after rollout.
- [ ] Watch [TUIST-DA](https://tuist.sentry.io/issues/116250005/) event rate over the next 24 h after canary then production. Expect rate to drop to near-zero. The `Fixes TUIST-DA` reference auto-resolves the issue on merge so we re-page if it ever fires again.


pepicrft (approved):
Approving. The extraction is clean, the helper is well-tested, and the fix lands on the call site Sentry was flagging. A couple of notes that don't block this PR.

**Coverage scope is narrower than it reads.** The PR description says "every read of the read-repo gets the same retry behavior writes already had," but `defoverridable all: 1, all: 2, one: 1, one: 2` only wraps those four. In Ecto, `one!/2`, `preload/2,3`, `aggregate/3,4`, `exists?/1,2`, `get/2,3`, `get_by/2,3`, and `stream/1,2` each call into `Ecto.Repo.Queryable.*` (or peers) directly. They do NOT go through `all` or `one`. There are real callers on `ClickHouseRepo` for most of these: `Tuist.Tests` and the build/test LiveViews use `preload`; `Tuist.Bundles` / `Tuist.Xcode` use `exists?`; `Tuist.CommandEvents` uses `aggregate`; tests use `one!`. A NAT-induced flap on one of those paths would still burn the worker's Oban attempts. The known failing path (`mark_test_case_runs_as_flaky` -> `ClickHouseRepo.all`) is covered so this PR fixes TUIST-DA, but the framing in the PR body / module doc is broader than the wiring. Worth either tightening the wording or following up to wrap the rest.

**Tiny duplication.** `IngestRepo.with_retry/2` hardcodes `\\ 3` to mirror `ClickHouseRetry`'s `@max_retries`. If the retry budget ever changes, these will drift silently. Pulling the default from the helper (or just calling `ClickHouseRetry.with_retry(fun)` for the 1-arity path) would remove the magic number.

Neither is blocking. Happy to see this go out and watch the canary.

pepicrft (approved):

PR #10691, opened 2026-05-08
fix(server): skip CSRF cross-origin check for FunWithFlags UI assets

fortmarek:
Fixes [TUIST-F1](https://tuist.sentry.io/issues/118505612/).

### Summary

`Plug.CSRFProtection` registers a `before_send` callback that raises `InvalidCrossOriginRequestError` on any non-XHR `GET` returning a JS content-type, regardless of `Origin` / `Sec-Fetch-Site` headers (see `Plug.CSRFProtection.cross_origin_js?/1`). `FunWithFlags.UI.Router` serves its bundled JS via `Plug.Static` after the `:browser_app` pipeline arms that callback, so every browser hit on `/ops/flags/assets/*.js` was 500-ing in production with that exception.

Set `plug_skip_csrf_protection` on the conn for paths starting with `/ops/flags/assets/`, registered as a single small plug in the existing `:ops` pipeline. Only static GETs are exempt — CSRF still applies to the FunWithFlags admin pages and the flag-toggle form submissions.

### How to test locally

1. Sign in as an ops user and navigate to `/ops/flags`.
2. Open a flag detail page (which loads `/ops/flags/assets/details.js`); the bundled JS should load without errors and the flag UI should render normally.
3. The added regression test covers the asset path: `mix test test/tuist_web/router_test.exs`.

### DIFF
--- a/server/lib/tuist_web/router.ex
+++ b/server/lib/tuist_web/router.ex
@@ -1016,7 +1016,13 @@ defmodule TuistWeb.Router do
   # FunWithFlags.UI serves its bundled JS via Plug.Static after the :browser_app
   # pipeline registers Plug.CSRFProtection's before_send callback, which raises
   # InvalidCrossOriginRequestError on any non-XHR GET that returns a JS response.
-  defp skip_csrf_for_fun_with_flags_assets(%Plug.Conn{path_info: ["ops", "flags", "assets" | _]} = conn, _opts) do
+  # Constrained to safe methods so any future state-changing route under this
+  # prefix still goes through CSRF protection.
+  defp skip_csrf_for_fun_with_flags_assets(
+         %Plug.Conn{method: method, path_info: ["ops", "flags", "assets" | _]} = conn,
+         _opts
+       )
+       when method in ["GET", "HEAD"] do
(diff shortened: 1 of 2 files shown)


pepicrft:
Security-wise, this looks acceptable to me because the bypass is constrained to authenticated ops static assets and does not remove CSRF from the FunWithFlags pages or form submissions.

The main residual risk is future expansion under `/ops/flags/assets/*`. If anything dynamic or state-changing ever lands under that prefix, it would inherit this exception. Tightening the guard to `GET`/`HEAD` and JS assets, or splitting these assets into a dedicated no-CSRF static pipeline, would make the exception easier to audit. I do not think that needs to block this fix.

pepicrft (approved):
Approved. This fixes the production asset loading issue while keeping ops authorization in place and leaving CSRF protection on the FunWithFlags admin UI paths.

fortmarek:
Tightened the guard to `GET`/`HEAD` in 87e16348a4 so any future state-changing route under `/ops/flags/assets/*` would still go through CSRF protection. Did not split into a dedicated no-CSRF static pipeline since the method-clause guard achieves the same audit clarity with a single function head.

PR #10692, opened 2026-05-08
fix(cli): allow folder resources to overlap sources

### LINKED ISSUE #10680: Regression in 4.182.0: package test target resource copy tries to add Swift test file to missing build phase
### What happened?

I bisected this issue with Codex 5.5 and reduced it to a minimal sample. The text below is by Codex.

Starting with Tuist 4.182.0, generating a project that depends only on a Swift package library product fails if that package also contains a test target with `resources: [.copy(".")]`.

The app target depends on the package library product only. It does not depend on the package test target.

Expected behavior: Tuist should generate the workspace successfully, as it did in 4.181.1.

Actual behavior: Tuist appears to process the package test target resources and tries to add the test source file to a build phase that has not been added to the generated project.

### How do we reproduce it?

1. Use the attached sample project.

The sample is intentionally minimal:

- One Tuist iOS app target
- One local Swift package dependency named `HybridLogicalClock`
- The app depends only on `.external(name: "HybridLogicalClock")`
- The Swift package has:
  - one tiny library target
  - one tiny test target
  - the test target uses `resources: [.copy(".")]`

2. From the sample project root, run:

```sh
tuist install
tuist generate --no-open
```

[HLCMinimalTuist.zip](https://github.com/user-attachments/files/27503370/HLCMinimalTuist.zip)

### Error log

4.181.1 succeeds
4.182.0 fails

**Error log**

```text
Using cache binaries for the following targets:
Generating workspace HLCMinimalTuist.xcworkspace
Generating project HLCMinimalTuist
Generating project HybridLogicalClock
✖ Error
  Trying to add a file at path /Users/kaioelfke/Desktop/Samples/HLCMinimalTuist/Packages/HybridLogicalClock/Tests/HybridLogicalClockTests/HybridLogicalClockTests.swift to a build phase that hasn't been added to the project.

  Sorry this didn’t work. Here’s what to try next:
   ▸ If the error is not actionable, let's discuss it in the <Troubleshooting & how to: https://community.tuist.dev/c/troubleshooting-how-to/6>
   ▸ If you are very certain it's a bug, <file an issue: https://github.com/tuist/tuist>
```

### macOS version

26.4.1

### Tuist version

4.182.0 and higher

### Xcode version

26.5 (17F42)

pepicrft:
> [!NOTE]
> TL;DR;
> - Before #10268: local Swift package test targets were pruned, so app-only package dependencies did not exercise package test resources.
> - Regression: #10268 correctly preserved those local package test targets, which exposed that `.copy(".")` can make a test target directory both a copied folder resource and a source group.
> - Fix: keep the test targets and let folder references coexist with logical groups at the same absolute path.

## Summary
- Track folder references separately from logical file groups when generating project file elements.
- Allow a copied resource folder to coexist with nested source file references at the same path.
- Add regression coverage for the Swift package test resource shape behind #10680.

Fixes #10680

## Testing
- `tuist generate tuist TuistGenerator TuistGeneratorTests ProjectDescription --no-open`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistGeneratorTests/ProjectFileElementsTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `mise run cli:lint --fix`

### DIFF
--- a/cli/Sources/TuistGenerator/Generator/ProjectFileElements.swift
+++ b/cli/Sources/TuistGenerator/Generator/ProjectFileElements.swift
@@ -71,6 +71,7 @@ class ProjectFileElements {
     // MARK: - Attributes
 
     var elements: [AbsolutePath: PBXFileElement] = [:]
+    var folderReferences: [AbsolutePath: PBXFileReference] = [:]
     var products: [String: PBXFileReference] = [:]
     var compiled: [AbsolutePath: PBXFileReference] = [:]
     var knownRegions: Set<String> = Set([])
@@ -416,7 +417,11 @@ class ProjectFileElements {
         sourceRootPath: AbsolutePath
     ) throws {
         // The file already exists
-        if elements[fileElement.path] != nil { return }
+        if elements[fileElement.path] != nil,
+           !(fileElement.isFolderReference && folderReferences[fileElement.path] == nil)
+        {
+            return
+        }
 
         let fileElementRelativeToSourceRoot = fileElement.path.relative(to: sourceRootPath)
         let closestRelativeRelativePath =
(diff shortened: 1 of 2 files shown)


cli/Sources/TuistGenerator/Generator/ProjectFileElements.swift
```diff
@@ -723,6 +750,24 @@ class ProjectFileElements {
         elements[fileAbsolutePath] = file
     }
 
+    func addFolderReferenceElementRelativeToGroup(
+        from _: AbsolutePath,
```
fortmarek, line 754: do we need this if we don't use it?

fortmarek (approved):

PR #10697, opened 2026-05-08
fix(infra): force-delete stuck Terminating pods before helm upgrade

fortmarek:
## Summary

Canary deploys have been looping between `failed` and `pending-rollback`. Every helm operation that uses `--wait` (both the recovery `helm rollback` and the main `helm upgrade --atomic --wait`) times out because an xcresult-processor pod is stuck Terminating for 23h on the single canary Mac mini.

With xcresult-processor's `nodeSelector` + `hostPort 9091` + `maxSurge=0` rollout, the deployment controller can't surge a new Pod onto a host whose slot is still held by the zombie, so the new ReplicaSet's Pod stays `Pending` with `FailedScheduling` ("Too many pods"). Helm's wait can never see `Ready`, the recovery rollback hits the same wall, and the release flips back and forth.

A Pod that has been Terminating past 5 minutes is effectively orphaned: the kubelet has lost track of the container. Force-deleting just removes it from etcd so the scheduler frees up the slot.

This adds a `Force-delete stuck Terminating pods` step right after `kubectl apply CRDs` (so the recovery rollback that runs after it can also complete on a healthy cluster).

Failing run for context: https://github.com/tuist/tuist/actions/runs/25552991784

## Test plan

- [ ] Re-run a canary deploy after merging; the new step should log either "Force-deleting tuist-tuist-xcresult-processor-..." or nothing, and the subsequent recovery / `helm upgrade` should reach `Ready`.
- [ ] Steady-state deploys (no zombies) should be no-ops at this step.

### DIFF
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -277,6 +277,39 @@ jobs:
         # schema changes ship the same way code changes do; no
         # operator-side manual step.
         run: kubectl apply -f "$HELM_CHART_PATH/crds/"
+      - name: Force-delete stuck Terminating pods
+        # A pod stuck Terminating for hours blocks rolling updates on
+        # capacity-constrained nodes, most acutely the macOS fleet:
+        # xcresult-processor pins to a single Mac mini via nodeSelector
+        # + hostPort 9091 and the rollout strategy is
+        # maxSurge=0/maxUnavailable=1. The new ReplicaSet's Pod stays
+        # Pending with FailedScheduling ("Too many pods") because the
+        # zombie still occupies the host's pod slot, and helm
+        # `--atomic --wait` (and the recovery rollback below) time out
+        # waiting for a Ready that can never happen. Anything stuck
+        # Terminating past 5 min is effectively orphaned: the kubelet
+        # has lost track of the container, and force-deleting just
+        # removes it from etcd so the scheduler frees up the slot.
+        run: |
+          set -euo pipefail
+
+          threshold=300  # seconds
+          now="$(date -u +%s)"
+
+          # `--ignore-not-found` covers a rare race where the pod gets
+          # cleaned up between the list and delete calls.
+          kubectl -n "$NAMESPACE" get pods \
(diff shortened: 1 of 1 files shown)


fortmarek:
Closing without merging. Findings from the investigation:

**Root cause of the failing run.** A `tuist-tuist-xcresult-processor` Pod was stuck `Terminating` for 23h on the single canary Mac mini. With `nodeSelector kubernetes.io/os=darwin` + `hostPort 9091` + `maxSurge=0/maxUnavailable=1`, the deployment controller couldn't surge a replacement onto the same host while the zombie still held the slot. So `helm --atomic --wait` never saw `Ready`, the recovery rollback hit the same wall, and the release oscillated between `failed` and `pending-rollback` across deploys.

**Why this is a tart-kubelet issue, not a workflow issue.** On the Hetzner Linux nodes the upstream node-controller path handles this in 5-10 min: NotReady → `node.kubernetes.io/unreachable:NoExecute` taint → force-eviction → CAPI deletes the Machine. The macOS path is different because tart-kubelet is our own kubelet shim and the Mac mini Node stayed `Ready` while one Pod's VM went away. That's already being fixed at the right layer:

- 5685880033 fix(infra,server): close terminal-Pod loop in tart-kubelet + watcher
- de4da11454 chore(infra): bump CAPI operator to terminal-pod-loop-fix build

The 23h-stuck Pod predates those commits rolling.

**Why not land the workflow safety net anyway.** A blanket force-delete in the deploy job would mask future regressions in the terminal-Pod loop (and on the Linux side, mask real eviction bugs). Better to let the kubelet-level fix prove itself and only revisit if a stuck-Terminating Pod recurs *after* the new tart-kubelet build has been running for a while.

**Unblock plan for the current canary.** Manual one-shot:

```
kubectl -n tuist-canary delete pod tuist-tuist-xcresult-processor-9c74997cd-pltwk \
  --grace-period=0 --force
```

The helm release is `failed` after rev 46, so the recovery step in the next run will skip and go straight to `helm upgrade --install`. With the slot freed, the rollout converges.

PR #10698, opened 2026-05-08
fix(server): log temp dir contents when xcresult extraction yields no bundle

fortmarek:
## Summary

- One tenant has been hitting `:xcresult_not_found` repeatedly since the new Tart VM image started serving prod xcresult processing on 2026-05-07. The error site only tells us that `find_xcresult/1` returned `nil`, which can mean two very different things:
  1. The Swift NIF returned `:ok` without actually writing files (silent extraction failure — read-only `/tmp`, FS quota, codesign/arch issue, etc.).
  2. The NIF wrote a layout the lookup doesn't recognize (e.g. a wrapping `result_bundle/` dir, so `Info.plist` lives one level deeper than the fallback checks).
- I pulled the actual failing archive, extracted it locally with the production NIF code (built from this branch on the same Darwin version the Mac mini reports), and the bundle parses cleanly end-to-end. So the bug isn't in the upload — it's in the runtime of the new image fleet, and the current logs aren't enough to tell which side.
- This patch logs `temp_dir` and the first 30 entries from a recursive glob right before returning `nil`. Next failing job in Sentry will have everything we need to pick between the two hypotheses.

## Why diagnostic-only

The archive itself is well-formed and the extraction code is verified correct on identical hardware/OS. Without log evidence from a real failing job, any "fix" would be speculation. The follow-up (recursive `Info.plist` search, NIF rebuild, /tmp probing on the VM, etc.) waits for the next failure to land in Sentry with the contents listing attached.

## Test plan

- [x] `mix test test/tuist/processor/xcresult_processor_test.exs` — 12 passing including new coverage on the diagnostic path
- [ ] After merge: watch [TUIST-EY](https://tuist.sentry.io/issues/TUIST-EY) for the next event; the breadcrumb should now carry the temp_dir listing

### DIFF
--- a/server/test/tuist/processor/xcresult_processor_test.exs
+++ b/server/test/tuist/processor/xcresult_processor_test.exs
@@ -2,6 +2,8 @@ defmodule Tuist.Processor.XCResultProcessorTest do
   use ExUnit.Case, async: true
   use Mimic
 
+  import ExUnit.CaptureLog
+
   alias ExAws.S3.Upload
   alias Tuist.Processor.XCResultNIF
   alias Tuist.Processor.XCResultProcessor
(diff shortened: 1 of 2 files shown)


server/test/tuist/processor/xcresult_processor_test.exs
```diff
@@ -321,36 +323,49 @@ defmodule Tuist.Processor.XCResultProcessorTest do
       assert {:ok, ^parsed_data} = XCResultProcessor.process_local(fixture_path)
     end
 
-    test "surfaces NIF decompression failures for AppleArchive payloads" do
+    test "logs temp_dir contents when extraction yields no recognizable bundle" do
       fixture_dir =
         Path.join(
           System.tmp_dir!(),
```
pepicrft: Let's use `@tag :tmp_dir`

pepicrft (approved):

server/test/tuist/processor/xcresult_processor_test.exs
fortmarek: Done in f6ae1aaad9.

PR #10701, opened 2026-05-08
feat(server,infra): add request_kind tagging + page load Apdex dashboard

fortmarek:
### Purpose

Two changes that together ship a robust per-request-kind observability story:

1. **`TuistWeb.Plugs.RequestKindPlug`** — writes a coarse `request_kind` (`page_load`, `page_image`, `api`, `mcp`, `scim`, `marketing`, `docs`, `marketing_feed`, `api_catalog`, `auth`) into `Logger.metadata` and the active OTel span. Wired into every router pipeline with a distinct external surface. The Loki handler now ships `request_kind` as structured metadata, and the console formatter prints it locally.

2. **Page Load Apdex dashboard** — Grafana Cloud dashboard ([UID `ma6s248`](https://github.com/tuist/tuist/blob/main/infra/grafana-dashboards/page-load-apdex.json)) tracking Apdex score, latency percentiles, request rate, and 5xx errors for the dashboard's HTML routes, scoped via `| request_kind=\"page_load\"`.

### Why the plug

The first cut of this dashboard discriminated web UI page loads from API calls with `|~ \"auth_account_handle\" != \"selected_account_handle\"` — relying on the implicit invariant that web UI requests only set `auth_account_handle` while API requests set both. That invariant is wrong: `LoaderPlug` resolves the account on `/:account_handle` dashboard URLs and calls `ObservabilityContextPlug.set_selection_context/1`, which sets `selected_account_handle` for those page loads too. The filter silently dropped traffic and the metric was structurally fragile.

`request_kind` makes the discriminator explicit, declared at the pipeline boundary in `router.ex`, and reusable for future API/mcp/scim slicing.

### Panels

8 panels across 4 rows:

- **Overview**: Apdex Score (T=1s, stat), Request Rate (stat), P99 Latency (stat), 5xx Error Rate (stat)
- **Apdex**: Apdex Score Over Time — two lines, T=1s (green) and T=200ms (orange)
- **Latency**: P50 / P95 / P99 percentiles
- **Traffic & Errors**: request-rate timeseries and 5xx-rate bars

Apdex is computed properly as `(satisfied + tolerating/2) / total` using three hidden Loki rate queries over a `regexp \"in (?P<ms>[0-9]+)ms\"` extraction, then a math expression — replacing a naive proxy ratio that always returned 100% because Phoenix consistently logs response times in `ms`.

### How to test locally

- `mix test test/tuist_web/plugs/request_kind_plug_test.exs` — unit tests for the plug.
- Tail the local console: requests should print `request_kind=page_load` (etc.) in the metadata block alongside the existing `auth_account_handle` / `request_id`.
- The dashboard itself is for the managed Grafana Cloud setup; once merged, Grafana's Git Sync (60 s interval, or webhook) pulls the file into the `Tuist Dashboards` folder. Verify by opening `/d/ma6s248/tuist-page-load-apdex` after the next sync.

### DIFF
--- a/infra/grafana-dashboards/page-load-apdex.json
+++ b/infra/grafana-dashboards/page-load-apdex.json
@@ -27,7 +27,7 @@
       }
     ],
     "cursorSync": "Off",
-    "description": "Apdex score and latency metrics for Tuist web page loads (request_kind=\"page_load\"). T=1s (satisfied <1s, tolerating 1-4s, frustrated >4s).",
+    "description": "Apdex score and latency metrics for Tuist web page loads. Discriminated by request_kind=\"page_load\" — extracted via | logfmt from the pod-log stdout stream that the per-node Alloy DaemonSet ships to Loki. T=1s (satisfied <1s, tolerating 1-4s, frustrated >4s).",
     "editable": true,
     "elements": {
       "panel-11": {
@@ -53,7 +53,7 @@
                       },
                       "spec": {
                         "app": "grafana-assistant-app",
-                        "expr": "sum(rate({service_name=\"tuist\", env=\"production\", container=\"server\"} | request_kind=\"page_load\" | regexp \"in (?P<ms>[0-9]+)ms\" | ms <= 1000 [$__interval]))",
+                        "expr": "sum(rate({env=\"production\", container=\"server\"} | logfmt | request_kind=\"page_load\" | regexp \"in (?P<ms>[0-9]+)ms\" | ms <= 1000 [$__interval]))",
                         "instant": false,
                         "legendFormat": "satisfied",
                         "queryType": "range",
@@ -76,7 +76,7 @@
                       },
                       "spec": {
                         "app": "grafana-assistant-app",
-                        "expr": "sum(rate({service_name=\"tuist\", env=\"production\", container=\"server\"} | request_kind=\"page_load\" | regexp \"in (?P<ms>[0-9]+)ms\" | ms > 1000 | ms <= 4000 [$__interval]))",
+                        "expr": "sum(rate({env=\"production\", container=\"server\"} | logfmt | request_kind=\"page_load\" | regexp \"in (?P<ms>[0-9]+)ms\" | ms > 1000 | ms <= 4000 [$__interval]))",
                         "instant": false,
                         "legendFormat": "tolerating",
                         "queryType": "range",
@@ -99,7 +99,7 @@
                       },
                       "spec": {
                         "app": "grafana-assistant-app",
-                        "expr": "sum(rate({service_name=\"tuist\", env=\"production\", container=\"server\"} | request_kind=\"page_load\" | regexp \"in (?P<ms>[0-9]+)ms\" [$__interval]))",
+                        "expr": "sum(rate({env=\"production\", container=\"server\"} | logfmt | request_kind=\"page_load\" | regexp \"in (?P<ms>[0-9]+)ms\" [$__interval]))",
                         "instant": false,
                         "legendFormat": "total",
                         "queryType": "range",
@@ -213,7 +213,7 @@
                       },
                       "spec": {
                         "app": "grafana-assistant-app",
-                        "expr": "sum(rate({service_name=\"tuist\", env=\"production\", container=\"server\"} | request_kind=\"page_load\" |~ \"Sent [0-9]+ in\" [$__interval]))",
+                        "expr": "sum(rate({env=\"production\", container=\"server\"} | logfmt | request_kind=\"page_load\" |~ \"Sent [0-9]+ in\" [$__interval]))",
                         "instant": false,
                         "legendFormat": "Page Loads/s",
                         "queryType": "range",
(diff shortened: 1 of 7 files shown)


fortmarek:
Apdex dashboard built directly in grafana before this one gets pulled in via git sync: https://tuist.grafana.net/goto/eflglg45sp3wgf?orgId=stacks-1028740

server/lib/tuist_web/router.ex
```diff
@@ -54,6 +55,7 @@ defmodule TuistWeb.Router do
   end
 
   pipeline :browser_app do
+    plug RequestKindPlug, "page_load"
```
pepicrft: I believe pipelines can take compile-time metadata that's then exposed as an assign. I'm inclined to do that instead over passing options to the plug. That way you can just define the plug once at the Endpoint level, and then set the kind if we've defined it at the router level.

pepicrft (approved):

server/lib/tuist_web/router.ex
fortmarek: Done in c636e3d3bb. Phoenix's `pipeline` macro doesn't expose compile-time metadata as an assign out of the box, so I went with the closest equivalent in spirit: `RequestKindPlug` is now installed once at the Endpoint and is option-less; pipelines declare their kind via `plug :put_request_kind, "..."`, which just assigns `:request_kind` to the conn. The endpoint plug uses `register_before_send/2` to copy the assign into `Logger.metadata` + the active OTel span at response time — registered after `Plug.Telemetry` so its before_send fires first (LIFO) and the kind is in scope when `Sent NNN in NNNms` is emitted.

Happy to swap to per-route `metadata: %{request_kind: ...}` instead if you'd rather have it on routes than pipelines, but that's ~50 sites vs 9.

infra/grafana-dashboards/page-load-apdex.json
```diff
@@ -0,0 +1,1374 @@
...
+                    "query": {
+                      "kind": "DataQuery",
+                      "group": "loki",
+                      "version": "v0",
+                      "datasource": {
+                        "name": "grafanacloud-logs"
+                      },
+                      "spec": {
+                        "app": "grafana-assistant-app",
+                        "expr": "sum(rate({service_name=\"tuist\", env=\"production\", container=\"server\"} | request_kind=\"page_load\" | regexp \"in (?P<ms>[0-9]+)ms\" | ms <= 1000 [$__interval]))",
```
pepicrft: I think this selector is going to make the dashboard empty. The managed k8s log path documents `env` plus Kubernetes labels like `container`, but not `service_name`; the in-process Loki handler uses `service_name="tuist-server"` and does not add `container`. Could we align this with the actual production log source before merging?

infra/grafana-dashboards/page-load-apdex.json
```diff
@@ -0,0 +1,1374 @@
...
+                    "query": {
+                      "kind": "DataQuery",
+                      "group": "loki",
+                      "version": "v0",
+                      "datasource": {
+                        "name": "grafanacloud-logs"
+                      },
+                      "spec": {
+                        "app": "grafana-assistant-app",
+                        "expr": "sum(rate({service_name=\"tuist\", env=\"production\", container=\"server\"} | request_kind=\"page_load\" |~ \"Sent [0-9]+ in\" [$__interval]))",
```
pepicrft: `request_kind` is added to Logger metadata and the console formatter, so on the managed pod-log path it will be text in the log line, not a Loki label. These LogQL filters need to parse it first, for example `| logfmt | request_kind="page_load"`, or we need to configure Alloy/Loki to promote it as structured metadata or a label.

pepicrft (commented):
Found two issues that look like they will keep the new dashboard from returning data.

infra/grafana-dashboards/page-load-apdex.json
fortmarek: Good catch — confirmed by tracing the actual setup. The in-process LokiLoggerHandler is gated on `TUIST_LOKI_URL`, which is unset in our managed envs (the staging values say so explicitly: "Logs follow a different path — the server writes to stdout and a per-node Alloy DaemonSet tails /var/log/pods"). So in production we go through the upstream `feature-pod-logs-via-loki` Alloy chart, whose discovery rules emit `cluster`, `namespace`, `pod`, `container`, `job` plus the `env` extraLabel — no `service_name`.

Fixed in a6cc1caa02 — selector is now `{env="production", container="server"}`, which is unique in the production cluster (only `server-deployment.yaml` declares a `name: server` container).

infra/grafana-dashboards/page-load-apdex.json
fortmarek: Right — `request_kind` lands as text inside the stdout line (the console formatter renders it as `request_kind=<value>` via the metadata config in `config/config.exs`), and Alloy ships pod stdout without promoting Logger metadata to Loki structured metadata. Switched all queries to `| logfmt | request_kind="page_load"` in a6cc1caa02; `| logfmt` extracts the textual `key=value` pairs into queryable labels and the subsequent label filter does the discrimination.

If we ever want to skip the per-query parse, the upstream `feature-pod-logs-via-loki` chart's `structuredMetadata` map + `extraLogProcessingStages` would let us promote `request_kind` to real Loki structured metadata at scrape time. Happy to follow up with that change if the parse cost shows up in query budget — for ~33k page loads/day it should be fine inline.

PR #10715, opened 2026-05-11
feat(cli): add onOutdatedDependencies action to GenerationOptions

freak4pc:
Today, when `tuist generate` notices that `Package.resolved` has changed, it just prints a warning telling you to run `tuist install` and continues. This caused a lot of confusion within our team while adopting, and after talking to @pepicrft and @fortmarek they were on board with this idea :)

This PR adds an `onOutdatedDependencies` option to `Tuist.GenerationOptions` that lets you pick how to react:

- `.warn` — current behavior (default). Graph loader still emits the warning.
- `.install` — run `tuist install` automatically and then continue generation.
- `.fail` — abort generation with an error, forcing the user to run `tuist install`.

```swift
let tuist = Tuist(
    project: .tuist(
        generationOptions: .options(onOutdatedDependencies: .install)
    )
)
```

Behavior is unchanged unless you opt in.

## How it works

Before generation kicks off, `GenerateService` checks the current `Package.resolved` against the one saved under `.build/Derived/Package.resolved` (same check the graph loader uses to emit the existing warning). Then:

- `.warn` skips the pre-check entirely and falls back to the existing graph-loader warning.
- `.install` invokes `InstallService` and continues.
- `.fail` throws `GenerateServiceError.outdatedDependencies`.

If there's no `Package.swift` at all, this is a no-op.

## Test plan

- [x] `GenerateServiceOutdatedDependenciesTests` covers all three actions (install when outdated, no-op when up-to-date, throw on `.fail`, skip when `.warn`)
- [x] Existing `GenerateServiceTests`, `InstallServiceTests`, `DumpServiceIntegrationTests`, and `TuistGeneratedProjectManifestMapperTests` still green


fortmarek (commented):
This looks great!

One thing I wonder could be better a fix is to replace `autoInstalledOutdatedDependencies: Bool` with `onOutdatedDependencies: OnOutdatedDependencies`.

Before:
```swift
.generationOptions(.options(
    autoInstalledOutdatedDependencies: true
))
```

After:
```swift
.generationOptions(.options(
    onOutdatedDependencies: .warning // default
    onOutdatedDependencies: .install // this new option
    onOutdatedDependencies: .error // possibly extension
))
```

What do you think? cc @pepicrft? 👀

freak4pc:
> This looks great!
> 
> One thing I wonder could be better a fix is to replace `autoInstalledOutdatedDependencies: Bool` with `onOutdatedDependencies: OnOutdatedDependencies`.
> 
> Before:
> 
> ```swift
> .generationOptions(.options(
>     autoInstalledOutdatedDependencies: true
> ))
> ```
> 
> After:
> 
> ```swift
> .generationOptions(.options(
>     onOutdatedDependencies: .warning // default
>     onOutdatedDependencies: .install // this new option
>     onOutdatedDependencies: .error // possibly extension
> ))
> ```
> 
> What do you think? cc @pepicrft? 👀

i like it but would go verb - warn/fail/install 
WDYT?

pepicrft:
Love the suggested API change

freak4pc:
ready for you @fortmarek @pepicrft !

cli/Sources/TuistKit/Services/GenerateService.swift
```diff
@@ -69,6 +99,26 @@ public struct GenerateService {
...
                 "The 'tuist generate' command is only available for generated projects and Swift packages."
             )
+
+        if let generatedProject = config.project.generatedProject,
+           generatedProject.generationOptions.onOutdatedDependencies != .warn,
+           try await outdatedDependenciesChecker.packageDependenciesAreOutdated(at: path)
+        {
+            switch generatedProject.generationOptions.onOutdatedDependencies {
+            case .warn:
+                break
```
fortmarek: I'd expect the warning to be logged from here.

cli/Sources/TuistKit/Services/GenerateService.swift
```diff
@@ -101,3 +151,61 @@ public struct GenerateService {
...
+        }
+    }
+}
+
+@Mockable
+protocol OutdatedDependenciesChecking {
+    func packageDependenciesAreOutdated(at path: AbsolutePath) async throws -> Bool
+}
+
+struct OutdatedDependenciesChecker: OutdatedDependenciesChecking {
```
fortmarek: nit: would move this to a separate file.

cli/Tests/TuistKitTests/Services/GenerateServiceTests.swift
```diff
@@ -378,3 +378,159 @@ struct GenerateServiceTests {
         }
     }
 }
+
+struct GenerateServiceOutdatedDependenciesTests {
```
fortmarek, line 385: why are we creating a new suite?

fortmarek (approved):

cli/Sources/TuistKit/Services/GenerateService.swift
freak4pc: fixed

cli/Sources/TuistKit/Services/GenerateService.swift
freak4pc: changed

cli/Tests/TuistKitTests/Services/GenerateServiceTests.swift
freak4pc: trying to combine MockInstallServicing and MockOpening causes some weird testing crash ("outlined destroy of Test"). Couldn't find a way around it... if you have thoughts happy to change

freak4pc:
Hey let me know if this is missing anything? @fortmarek

fortmarek:
Thanks 🙏

PR #10724, opened 2026-05-11
fix(infra): pre-apply tart-kubelet RBAC before helm upgrade

fortmarek:
## Summary

- Adds a workflow step before `helm upgrade --install` that renders `templates/tart-kubelet-rbac.yaml` via `helm template`, `kubectl apply`s it, and stamps the resource with helm ownership annotations so the subsequent upgrade can manage it without `invalid ownership metadata` errors.
- Breaks the chicken-and-egg loop where a regression to the tart-kubelet ClusterRole strands Pods Terminating on Mac minis (hostPort 9091 + max-pods budget), the xcresult-processor rollout never converges, and `helm upgrade --atomic` reverts the fix on timeout.

## Why

Recovered from a stuck production deploy where Pods on Mac minis sat in `Terminating` for 2–3 days because the pre-[a709f96795](https://github.com/tuist/tuist/commit/a709f96795f3b10dd79cc04e9f0f86a000e4f745) ClusterRole was missing `pods/finalizers` update/patch. Every helm rollout attempt applied the fixed RBAC during its `--wait` window, then rolled it back on timeout because the stuck pods kept new pods from scheduling, leaving the cluster in the exact pre-fix state for the next attempt to repeat.

Pre-applying the RBAC out-of-band puts the new permissions in place before helm starts waiting, so the kubelet's reconciler retry loop drains its 403 backlog and clears finalizers within seconds — well inside the `--wait` budget — and the rollout converges.

## Test plan

- [ ] Validate workflow YAML parses (verified locally with `yq`).
- [ ] Confirm `helm template ... --show-only templates/tart-kubelet-rbac.yaml` renders the expected ClusterRole (verified locally — `tuist-tuist-tart-kubelet` with `pods/finalizers` rule).
- [ ] Watch the next production deploy: pre-apply step runs cleanly, helm upgrade proceeds without ownership-metadata errors.
- [ ] Verify that a subsequent re-run is a no-op (helm sees the resource already owned, no diff).

### DIFF
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -346,6 +346,57 @@ jobs:
           # so dependent pods (Pending or otherwise) go with the Job.
           echo "$stale_jobs" | xargs -r kubectl -n "$NAMESPACE" delete job \
             --ignore-not-found --cascade=foreground --timeout=2m
+      - name: Pre-apply tart-kubelet RBAC out-of-band
+        # The tart-kubelet ClusterRole gates whether the per-Mac-mini
+        # kubelet can clear its `tart-kubelet.tuist.dev/vm-cleanup`
+        # finalizer when a Pod terminates (see
+        # infra/tart-kubelet/internal/podagent/reconciler.go). A regression
+        # there leaves Pods stuck Terminating on Mac minis: they hold
+        # hostPort 9091 and count against the host's max-pods budget, so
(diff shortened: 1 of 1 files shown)


fortmarek:
Closing — a709f96795 is sufficient to prevent recurrence once the chart with the fixed RBAC lands successfully. Will handle the current stuck state with a direct kubectl patch on the stuck Terminating pods instead, no workflow change needed.

PR #10729, opened 2026-05-11
fix(cli): coordinate SwiftPM graph reads

pepicrft:
## Summary
- take SwiftPM's own scratch-directory workspace lock while `SwiftPackageManagerGraphLoader` directly reads SwiftPM state
- resolve that scratch directory with SwiftPM's precedence: `SWIFTPM_BUILD_DIR`, `--scratch-path`, `--build-path`, then the package `.build` directory
- use the selected scratch directory for `workspace-state.json`, checkouts, registry downloads, binary artifacts, and the saved `Derived/Package.resolved` marker
- avoid hard-coded `.build/checkouts` assumptions for source-control checkouts under a custom scratch directory
- preserve the legacy `state/logs` directory during analytics cleanup so concurrently running processes do not lose log files

## Root cause
SwiftPM already serializes commands that use the same package scratch directory. I validated this in the SwiftPM source and with parallel `swift package resolve` runs, where later processes wait on the existing scratch-directory lock. Because of that, this PR does not wrap Tuist's `swift package` command invocations.

The evidence-backed Tuist gap is narrower: after SwiftPM has produced scratch-directory state, `SwiftPackageManagerGraphLoader` reads `workspace-state.json`, dependency checkout folders, registry downloads, binary artifact folders, and package manifests directly. Those reads were not coordinated with another process running SwiftPM against the same scratch directory. The fix makes Tuist join SwiftPM's own lock before doing those direct reads.

The log crash has a separate root cause: analytics cleanup removed the legacy `state/logs` directory, which can race with processes still trying to open a log file there.

## What this does not claim
This does not claim that SwiftPM itself lacks process-level locking for `swift package resolve` or `swift package dump-package`. The direct `swift package resolve` failure still needs more data if it reproduces after this change, because SwiftPM should already serialize same-scratch-directory resolver processes unless something is using `--ignore-lock`, a different scratch directory, or mutating the package workspace outside SwiftPM.

## Validation
- cloned `swiftlang/swift-package-manager` and verified `SwiftCommandState.getActiveWorkspace()` acquires an exclusive scratch-directory lock with `FileLock.prepareLock(fileToLock: self.scratchDirectory)`
- stress-ran 8 concurrent `swift package --package-path <package> resolve` processes and observed SwiftPM serializing them with "Another instance of SwiftPM is already running using ... waiting..."
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistLoaderTests/PackageInfoMapperTests -only-testing TuistLoaderTests/SwiftPackageManagerGraphLoaderTests -only-testing TuistLoaderTests/RecursiveManifestLoaderTests/test_loadSPM_Package_ignoresPackagesInCustomSwiftPMCheckouts -only-testing TuistSupportTests/SwiftPackageManagerScratchDirectoryLocatorTests -only-testing TuistKitTests/InstallServiceTests/test_run_when_installing_dependencies_with_scratchPath_savesPackageResolvedInScratchDirectory -only-testing TuistKitTests/AnalyticsStateControllerTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`

### DIFF
--- a/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
+++ b/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
@@ -285,9 +285,17 @@ public struct SwiftPackageManagerGraphLoader: SwiftPackageManagerGraphLoading {
             .reduce(into: [:]) { result, item in
                 let (packageInfo, hash, projectManifest) = item
                 if let projectManifest {
+                    let swiftPackageManagerScratchDirectory: Path? = if Self.isLocalDependencyKind(packageInfo.kind) {
+                        nil
+                    } else {
+                        SwiftPackageManagerPaths
+                            .scratchDirectory(containingCheckout: packageInfo.folder)
+                            .map { Path.path($0.pathString) }
+                    }
                     result[.path(packageInfo.folder.pathString)] = DependenciesGraph.ExternalProject(
                         manifest: projectManifest,
-                        hash: hash
+                        hash: hash,
+                        swiftPackageManagerScratchDirectory: swiftPackageManagerScratchDirectory
                     )
                 }
             }
--- a/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
+++ b/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
@@ -5,6 +5,7 @@ import ProjectDescription
 import TSCUtility
 import TuistConstants
 import TuistCore
+import TuistEnvironment
 import TuistLogging
 import TuistSupport
 import XcodeGraph
@@ -49,7 +50,8 @@ public protocol SwiftPackageManagerGraphLoading {
     func load(
         packagePath: AbsolutePath,
         packageSettings: TuistCore.PackageSettings,
-        disableSandbox: Bool
+        disableSandbox: Bool,
+        swiftPackageManagerArguments: [String]
     ) async throws -> (TuistLoader.DependenciesGraph, [LintingIssue])
 }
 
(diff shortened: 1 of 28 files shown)


cli/Sources/TuistLoader/Models+ManifestMappers/Workspace+ManifestMapper.swift
```diff
@@ -31,7 +31,7 @@ extension XcodeGraph.Workspace {
             )
             .collect()
             .map(\.parentDirectory)
-            .filter { $0.basename != Constants.tuistDirectoryName && !$0.pathString.contains(".build/checkouts") }
+            .filter { $0.basename != Constants.tuistDirectoryName && !$0.pathString.contains("/checkouts/") }
```
fortmarek: This predicate runs on absolute project paths. After broadening it from `.build/checkouts` to any `/checkouts/` segment, a normal repo cloned under `/Users/alice/checkouts/App` will have every discovered `Project.swift`/`Package.swift` parent filtered out. `RecursiveManifestLoader` repeats the same predicate for package projects, so generation can silently lose all projects. Please compare against the actual SwiftPM scratchDirectory/checkouts path instead of the whole absolute path.

cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
```diff
@@ -59,30 +61,56 @@ public struct SwiftPackageManagerGraphLoader: SwiftPackageManagerGraphLoading {
...
+        }
+    }
+
+    private func loadUnsafe(
+        packagePath: AbsolutePath,
+        packageSettings: TuistCore.PackageSettings,
+        disableSandbox: Bool,
+        scratchDirectory: AbsolutePath
+    ) async throws -> (TuistLoader.DependenciesGraph, [LintingIssue]) {
+        let path = scratchDirectory
```
fortmarek, line 113: The PR now reads SwiftPM state from custom scratch/build directories, so external package paths can be `custom-build/checkouts/...`. Some downstream helpers still only recognize `.build/checkouts` (`Project+Core.swift` around `derivedDirectoryPath(for:)`, and `SwiftPackageManagerModuleMapGenerator.swift` around generated module-map placement). Those external targets fall through to `path/Derived`, so packages with resources, generated Info.plists, or generated module maps can write Tuist derived files inside the SPM checkout instead of the scratch `tuist-derived` directory.

fortmarek (commented):
I found two regressions worth addressing before this lands.

fortmarek (approved):

cli/Sources/TuistLoader/Models+ManifestMappers/Workspace+ManifestMapper.swift
pepicrft: Fixed by passing the resolved SwiftPM scratch directory into workspace and recursive package discovery, then filtering only paths under that scratchDirectory/checkouts. I also added regression coverage for a workspace path that contains /checkouts/ but is unrelated to SwiftPM.

cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
pepicrft: Fixed by carrying the SwiftPM scratch directory from the SPM graph into external projects, then using it for external project relocation, target derived directories, and generated module maps. I added custom-scratch and unrelated-checkouts coverage for those paths.

PR #10730, opened 2026-05-11
feat(cli): add --report-path to tuist test for structured summaries

justine-acorns:
Resolves none yet (filing alongside the PR since the maintainer convention is to describe the motivation here).

Adds a new `--report-path <path>` flag (env key `TUIST_TEST_REPORT_PATH`) to `tuist test`. When set, the CLI writes the in-memory `TestSummary` to the given path as JSON after the xcresult is parsed.

### Why

Post-run CI tooling (flake dashboards, CircleCI JUnit transforms, Datadog forwarders, custom summarizers) typically needs the structured test verdict for the run: per-test status, failures, retry repetitions, quarantine state, orphan issues. Today consumers get it via one of two paths:

1. Re-parse the xcresult themselves with `xcresultparser` or similar, losing Tuist's quarantine and orphan enrichment.
2. Query the Tuist API after upload (`tuist test show`, paginated `tuist test case run list --test-run-id`), which forces pagination (~13 pages for a ~6k-test run), races against upload completion, and duplicates parsing that Tuist just did.

Tuist already holds the enriched `TestSummary` in memory during `TestService.run()`. This flag asks Tuist to hand it to downstream tooling directly.

### Behavior

- `--report-path` is opt-in; no change if unset.
- Writes the report regardless of `--inspect-mode`. In local mode the CLI was already parsing the xcresult; in remote mode the CLI now parses it locally for the report side-effect, while the raw bundle still uploads to the server. The upload path is untouched.
- On a parse failure, the CLI logs a warning and skips the report; it does not fail the run. The run's exit code is still driven by xcodebuild.
- Sharded runs write one report per invocation (mirrors existing per-shard behavior). Aggregation is the caller's responsibility.

### JSON shape

The serialized struct is the existing `XCResultParser.TestSummary`, which already has snake-case `Encodable` conformance. No new types.

<details>
<summary>Sample output</summary>

```json
{
  "test_plan_name": "RequiredTests",
  "status": "failure",
  "duration": 399293,
  "run_destinations": [
    { "name": "iPhone 17 Pro", "platform": "iOS Simulator", "os_version": "26.2" }
  ],
  "test_modules": [
    {
      "name": "MoneyManagerUseCasesLiveTests",
      "status": "failure",
      "duration": 28,
      "test_suites": [],
      "test_cases": [
        {
          "name": "executeRebalances_partialFailure",
          "test_suite_name": "ExecuteMoneyManagerRebalancesLiveTests",
          "module": "MoneyManagerUseCasesLiveTests",
          "duration": 28,
          "status": "success",
          "is_quarantined": true,
          "failures": [
            {
              "message": "Expectation failed: (accountId → ...) == (rebalances[0].sourceAccountId → ...)",
              "line_number": 87,
              "issue_type": "assertion_failure"
            }
          ],
          "repetitions": [
            { "repetition_number": 1, "name": "executeRebalances_partialFailure", "status": "failure", "duration": 0, "failures": [/* ... */] },
            { "repetition_number": 2, "name": "executeRebalances_partialFailure", "status": "success", "duration": 28, "failures": [] }
          ],
          "arguments": [],
          "attachments": []
        }
      ]
    }
  ]
}
```

</details>

### Motivating use case

At Acorns we post-process JUnit output after every CI test job to (a) downgrade quarantined failures to `<skipped>` so CircleCI's Tests tab doesn't block the build, and (b) detach retry-pass `<failure>` entries so they don't get counted as real regressions. Today that requires the secondary Tuist API round-trips described above. With this flag the post-processor reads one file.

### How to test locally

```bash
cd path/to/a/tuist/project
tuist test --report-path /tmp/report.json
jq '.status, .test_modules[0].test_cases[0]' /tmp/report.json
```

Flip between `--inspect-mode local` and `--inspect-mode remote` to confirm the report is written in both.

### Tests

- `test_run_writesReportPathInLocalMode` — local-mode upload + report write.
- `test_run_writesReportPathInRemoteMode` — remote-mode upload still produces the report (the interesting one; validates that report emission doesn't depend on the upload mode).

### DIFF
--- a/cli/Sources/TuistEnvKey/EnvKey.swift
+++ b/cli/Sources/TuistEnvKey/EnvKey.swift
@@ -195,6 +195,7 @@ public enum EnvKey: String, CaseIterable {
     case testShardReference = "TUIST_SHARD_REFERENCE"
     case testShardSkipUpload = "TUIST_TEST_SHARD_SKIP_UPLOAD"
     case testShardArchivePath = "TUIST_TEST_SHARD_ARCHIVE_PATH"
+    case testReportPath = "TUIST_TEST_REPORT_PATH"
 
     // TEST SHOW
 
--- a/cli/Sources/TuistKit/Services/TestService.swift
+++ b/cli/Sources/TuistKit/Services/TestService.swift
@@ -244,7 +244,8 @@ public struct TestService { // swiftlint:disable:this type_body_length
         shardIndex: Int? = nil,
         shardSkipUpload: Bool = false,
         shardArchivePath: AbsolutePath? = nil,
-        mode: TestProcessingMode? = nil
+        mode: TestProcessingMode? = nil,
+        reportPath: AbsolutePath? = nil
     ) async throws {
         if validateTestTargetsParameters {
             try Self.validateParameters(
@@ -289,7 +290,8 @@ public struct TestService { // swiftlint:disable:this type_body_length
                 passthroughXcodeBuildArguments: passthroughXcodeBuildArguments,
                 runId: runId,
                 shardArchivePath: shardArchivePath,
-                mode: mode
+                mode: mode,
+                reportPath: reportPath
             )
             return
         }
@@ -313,7 +315,8 @@ public struct TestService { // swiftlint:disable:this type_body_length
                 testPlanConfiguration: testPlanConfiguration,
                 passthroughXcodeBuildArguments: passthroughXcodeBuildArguments,
                 runId: runId,
-                mode: mode
+                mode: mode,
+                reportPath: reportPath
             )
             return
         }
@@ -548,7 +551,8 @@ public struct TestService { // swiftlint:disable:this type_body_length
                 passthroughXcodeBuildArguments: passthroughXcodeBuildArguments,
                 config: config,
                 quarantinedTests: mutedQuarantinedTests,
-                mode: mode
+                mode: mode,
+                reportPath: reportPath
             )
         } catch {
             try await copyResultBundlePathIfNeeded(
@@ -629,7 +633,8 @@ public struct TestService { // swiftlint:disable:this type_body_length
         passthroughXcodeBuildArguments: [String],
         runId: String,
         shardArchivePath: AbsolutePath?,
-        mode: TestProcessingMode
+        mode: TestProcessingMode,
+        reportPath: AbsolutePath?
     ) async throws {
         guard let fullHandle = config.fullHandle else {
             throw TestServiceError.actionInvalid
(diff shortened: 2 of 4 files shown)


justine-acorns:
Quick note on the lint fix: adding one @Option to `TestRunCommand` tipped it over the 350-line `type_body_length` hard error (it was sitting at ~349 on main). I added `// swiftlint:disable:next type_body_length` matching the precedent in [`TuistRunCommand/Services/RunCommandService.swift:125`](https://github.com/tuist/tuist/blob/main/cli/Sources/TuistRunCommand/Services/RunCommandService.swift#L125) and [`TuistShareCommand/Services/ShareCommandService.swift:49`](https://github.com/tuist/tuist/blob/main/cli/Sources/TuistShareCommand/Services/ShareCommandService.swift#L49). Happy to refactor the file instead (e.g. split flag declarations into an extension) if that's preferred — that felt out of scope for this PR.

fortmarek:
> in remote mode the CLI now parses it locally for the report side-effect

This is not ideal; parsing the report locally is expensive for large XCResults, which is what motivated us to move the xcresult processing to remote macOS machines in the first place.

I wonder if a good alternative would be to ingest the data by querying the API or with [webhooks](https://community.tuist.dev/t/rfc-webhooks-for-tuist-events/969) that we're building out soon?

justine-acorns:
> > in remote mode the CLI now parses it locally for the report side-effect
> 
> This is not ideal; parsing the report locally is expensive for large XCResults, which is what motivated us to move the xcresult processing to remote macOS machines in the first place.
> 
> I wonder if a good alternative would be to ingest the data by querying the API or with [webhooks](https://community.tuist.dev/t/rfc-webhooks-for-tuist-events/969) that we're building out soon?

Thanks for the quick feedback on this! Totally hear you on the parse-cost concern. Tried to alleviate that by making it an opt-in thing with the parameter, but if it's straying too much I can just drop this. Maybe I can just share the underlying friction so we can figure out together what the right shape is:
  
The problem that I was trying to solve with this is a mismatch between the CI test results UI and `tuist test` output, resulting from quarantined tests and also orphan issues failing the test but not appearing in the CI's UI.

`tuist test` can exit green (since quarantined-only failures don't block the build) while CircleCI's Tests tab still shows those quarantined failures mixed in with any real ones. When there are real failures alongside, it's hard to tell them apart in the CI UI without some post-processing. We can view the run in Tuist, but not everyone knows to do this, and it is definitely confusing for folks having that misalignment.

Right now we are handling this by fetching the quarantined list separately, which works but feels a bit indirect — the fetch doesn't tell us whether a test was quarantined at the time it ran, just whether it's quarantined now. We are also essentially rewriting logic that tuist already has handled. I also looked at `tuist test case run list --test-run-id`, but it doesn't carry a run-time quarantine signal either, and paginating 5k+ rows to find a handful of failure details is a lot of work for the information we actually need.

I also considered proposing something smaller like a `--status failure` filter on that API, but wondered if there was a more synchronous path that matches what `tuist test` is already computing internally. That's where the flag idea came from. Curious what you think!

justine-acorns:
Closing this one in favor of a more minimal [PR](https://github.com/tuist/tuist/pull/10785) to expose `is_quarantined`. I do still think it would be nice to have a more synchronous and direct output from `tuist test`, but for now this new PR will at least get me the minimal functionality that I need.

PR #10739, opened 2026-05-12
chore(server): harden CI supply chain

pepicrft:
## Summary

Harden CI and release paths against cache poisoning and broad token exposure after reviewing the TanStack npm supply chain incident pattern.

## Changes

- Remove release and deploy cache restores that could influence signed or deployed artifacts.
- Disable Gradle caching in release jobs.
- Scope `id-token: write` to only jobs that exchange OIDC tokens with Tuist.
- Add explicit workflow permissions where they were missing.
- Add CODEOWNERS coverage for workflows, actions, release scripts, and dependency manifests.
- Require the Tuist audience on GitHub Actions and Bitrise OIDC tokens before exchanging them for Tuist access tokens.

## Validation

- `mix test test/tuist/oidc_test.exs test/tuist_web/controllers/api/oidc_controller_test.exs`
- `mix format --check-formatted lib/tuist/oidc.ex lib/tuist_web/controllers/api/oidc_controller.ex test/tuist/oidc_test.exs test/tuist_web/controllers/api/oidc_controller_test.exs`
- Workflow YAML parse with Ruby YAML loader
- `git diff --check`

## Operational Notes

I also tightened repository settings outside this diff:

- Default `GITHUB_TOKEN` permissions are now read-only.
- GitHub Actions can no longer approve pull requests.
- Fork PR approval now applies to all external contributors.

I left repository action allowlisting unchanged because `allowed_actions` is currently `all`, and enabling an allowlist without a complete inventory would likely break CI.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,31 @@
+# CI and release control planes
+/.github/workflows/ @tuist/engineering
+/.github/actions/ @tuist/engineering
+/mise/tasks/ @tuist/engineering
+
+# Dependency manifests and lockfiles
+/mise.toml @tuist/engineering
+/mise.lock @tuist/engineering
+/server/mise.toml @tuist/engineering
+/cache/mise.toml @tuist/engineering
+/noora/.mise.toml @tuist/engineering
+/handbook/mise.toml @tuist/engineering
+/status/mise.toml @tuist/engineering
+/infra/mise.toml @tuist/engineering
+/Package.swift @tuist/engineering
+/Package.resolved @tuist/engineering
+/package.json @tuist/engineering
+/package-lock.json @tuist/engineering
+/server/package.json @tuist/engineering
+/server/pnpm-lock.yaml @tuist/engineering
+/noora/package.json @tuist/engineering
+/noora/aube-lock.yaml @tuist/engineering
+/handbook/package.json @tuist/engineering
+/handbook/aube-lock.yaml @tuist/engineering
+/status/package.json @tuist/engineering
+/status/aube-lock.yaml @tuist/engineering
+/infra/registry-router/package.json @tuist/engineering
+/infra/registry-router/package-lock.json @tuist/engineering
+/gradle/settings.gradle.kts @tuist/engineering
+/gradle/build.gradle.kts @tuist/engineering
+/gradle/gradle.lockfile @tuist/engineering
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -40,7 +40,9 @@ jobs:
     permissions:
       contents: read
       id-token: write
-    if: github.event.pull_request.draft == false || github.event_name != 'pull_request'
+    if: |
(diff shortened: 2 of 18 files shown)


fortmarek (changes requested):
Found two issues:

1. **[P1] Do not expose OIDC to fork PR code** (`.github/workflows/android.yml:40`)

   This job still grants `id-token: write` even when the Tuist auth step is skipped for forks. Because later steps execute checked-out PR code, a fork PR can request a GitHub OIDC JWT directly from the runner env vars and exchange it with Tuist; the server currently accepts GitHub tokens based on `aud` + `repository`, and GitHub pull_request tokens still identify the base repo. Split fork/internal jobs or otherwise ensure jobs running untrusted fork code have no `id-token` permission.

2. **[P2] Validate cacheless release/deploy runtimes** (`.github/workflows/release.yml:795`)

   This removes the release/deploy `.build` caches that were added in #8877/#9132 and #8265 to keep `tuist install` and Swift build work off the critical path. That may be the right supply-chain tradeoff, but the jobs keep their current timeouts and the PR does not include cacheless timing. Please include a cold-run measurement or adjust the affected release/deploy timeouts so hardening does not make releases or canary acceptance flaky.

fortmarek:
In general, I do wonder what the impact of removing the caches will be on the release and whether the trade-off we're making here is the right one. I'm good with the rest 👀

pepicrft:
@fortmarek I benchmarked the affected release/deploy setup paths on a temporary branch with a push-only workflow: https://github.com/tuist/tuist/actions/runs/25749892647

The benchmark only ran local setup/install/generate/submodule steps. It did not save caches, upload artifacts, update changelogs, create tags or releases, upload to App Store Connect, deploy canary/production, or run acceptance tests.

Results:

| Path | Cache mode | Cache result | Total job time | Relevant install/generate steps |
| --- | --- | --- | ---: | --- |
| Release CLI setup | with cache | hit | 1m43s | cache restore 14s, `tuist install` skipped, `cli:ee` 16s |
| Release CLI setup | without cache | disabled | 1m54s | `tuist install` 29s, `cli:ee` 11s |
| Release iOS setup | with cache | miss | 1m45s | `tuist install` 32s, `tuist generate TuistApp --no-binary-cache` 25s |
| Release iOS setup | without cache | disabled | 1m30s | `tuist install` 30s, `tuist generate TuistApp --no-binary-cache` 22s |
| Server acceptance setup | with cache | miss | 1m27s | `tuist install` 32s, `cli:ee` 1s |
| Server acceptance setup | without cache | disabled | 1m22s | `tuist install` 29s, `cli:ee` 1s |

The only observed cache hit was `release-cli`, where removing the `.build` restore added roughly 11 seconds to the benchmarked setup path. The iOS release and server acceptance cache keys missed, so the cacheless path is effectively what those jobs already do for this commit. Based on this, I do not think we need timeout changes for these removals.

fortmarek (approved):

PR #10744, opened 2026-05-12
fix(server): improve Kura cache server readiness

pepicrft:
## Summary

- Allow accounts to add additional Kura regions and keep region labels/provider details out of customer-facing rows.
- Use the latest `kura@...` GitHub release as the displayed version while rolling out the matching Docker tag internally.
- Cache GitHub release page responses so App and Kura release lookups share paginated API data within the release TTL.
- Gate Kura endpoint activation on DNS plus HTTPS `/up` readiness so TLS/DNS propagation does not expose unusable endpoints to the CLI.
- Keep active servers marked `Active` during background rollouts and update the Kura Grafana dashboard to query Kura node metrics instead of legacy cache endpoints.

## Root Cause

The account settings UI was mixing server lifecycle state with deployment state, so active servers with in-flight deployments appeared as `Deploying`. Endpoint activation only waited for DNS, which meant an endpoint could be mirrored into `account_cache_endpoints` before HTTPS/TLS was actually usable.

## Validation

- `mix test test/tuist/github/releases_test.exs`
- `mix test test/tuist/kura_test.exs test/tuist/kura/reconciler_test.exs test/tuist_web/live/account_settings_live_test.exs`
- `mix test test/tuist/kura/regions_test.exs test/tuist_web/live/ops_account_live_test.exs test/tuist_web/live/ops_account_kura_deployment_live_test.exs`
- `git diff --check`
- `jq empty infra/grafana-dashboards/tuist-kura.json`

### DIFF
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -81,7 +81,6 @@ permissions:
 env:
   REGISTRY: ghcr.io
   IMAGE_NAME: tuist/tuist
-  KURA_RUNTIME_IMAGE_NAME: tuist/kura
   KURA_CONTROLLER_IMAGE_NAME: tuist/kura-controller
   HELM_CHART_PATH: infra/helm/tuist
   HELM_RELEASE_NAME: tuist
@@ -132,15 +131,6 @@ jobs:
           push: true
           tags: |
             ${{ env.REGISTRY }}/${{ env.KURA_CONTROLLER_IMAGE_NAME }}:${{ steps.tag.outputs.image_tag }}
-      - uses: docker/build-push-action@v6
-        name: Build and push Kura runtime image
-        with:
-          context: ./kura
-          file: ./kura/Dockerfile
-          push: true
-          tags: |
-            ${{ env.REGISTRY }}/${{ env.KURA_RUNTIME_IMAGE_NAME }}:${{ steps.tag.outputs.image_tag }}
-
   observability-install:
     # Installs / upgrades the grafana/k8s-monitoring wrapper chart ahead of
     # the server rollout so the Alloy OTLP receiver is ready before the
@@ -396,7 +386,6 @@ jobs:
             -f "$HELM_CHART_PATH/values-managed-${{ inputs.environment }}.yaml" \
             --set server.image.tag="$IMAGE_TAG" \
             --set kuraController.image.tag="$IMAGE_TAG" \
-            --set kuraRuntime.image.tag="$IMAGE_TAG" \
             --atomic --timeout 90m --wait
       - name: Deploy Kura regional controllers
         if: inputs.environment == 'production'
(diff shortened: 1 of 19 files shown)


fortmarek (commented):
Two findings from the review:

1. Keep the Kura rollout source of truth in deploy config, not GitHub Releases.

Moving the desired Kura runtime version from deployment-owned configuration to GitHub Releases weakens the current convention where CI/Helm declares the exact image tag running in the cluster. That convention gives us pinning, rollback, auditability, and avoids coupling reconciliation to GitHub availability, rate limits, release ordering, or cache staleness. I’d keep TUIST_KURA_RUNTIME_IMAGE_TAG/Helm as the rollout source of truth and use GitHub release lookup only as a display or discovery hint, if needed.

2. Keep GitHub release failures non-fatal if the lookup stays.

fetch_release_page/1 only handles 200, 5xx, and transport errors. A 403/404/429 response after retries would currently raise a CaseClauseError. If GitHub lookups remain anywhere on a server execution path, please handle all non-200 statuses by logging and returning nil or an error tuple.

pepicrft:
@fortmarek addressed both points.

1. Kura runtime rollout now uses `TUIST_KURA_RUNTIME_IMAGE_TAG` again as the source of truth in every environment. GitHub Releases is no longer used to decide the image tag scheduled by `schedule_runtime_image_deployments/0`.
2. GitHub release lookups now handle every non-200 response by logging and returning `nil`, including 403, 404, and 429.

I also added/updated coverage for both behaviors and ran:

```sh
mix test test/tuist/kura_test.exs test/tuist/github/releases_test.exs test/tuist_web/live/account_settings_live_test.exs test/tuist_web/live/ops_account_live_test.exs test/tuist_web/live/ops_account_kura_deployment_live_test.exs
git diff --check
```

fortmarek (commented):
Follow-up findings after the latest changes:

1. [P1] Keep setting the configured Kura runtime tag

After the latest change, `Tuist.Kura.latest_versions/1` reads only `TUIST_KURA_RUNTIME_IMAGE_TAG`, but `.github/workflows/server-deployment.yml` no longer passes `--set kuraRuntime.image.tag`. Since the Helm value defaults to an empty string and the chart only emits the env var when the tag is set, managed deploys will have no configured Kura version: new Kura server creation will be disabled and active servers will not roll. If the runtime image now comes from the Kura release workflow, this workflow still needs to declare and pin that release tag in Helm.

2. [P3] Remove or wire the Kura release lookup

With rollout moved back to deploy configuration, the new `Tuist.GitHub.Releases.get_latest_kura_release/0` helper has no non-test call sites. The only remaining references are tests, docs, and generic label-stripping helpers, so we are not actually using the GitHub Kura release for display or rollout. Either wire it explicitly as a UI-only discovery hint or drop it to avoid suggesting GitHub is still part of the runtime version path.

fortmarek (approved):

pepicrft:
@fortmarek addressed the follow-ups: server deploys now pass `kuraRuntime.image.tag` to Helm again, resolving the latest `kura@` release into the plain runtime image tag unless the workflow input overrides it. I also removed the unused server-side Kura GitHub release lookup so the app no longer suggests GitHub releases are part of the runtime version path.

PR #10748, opened 2026-05-12
feat(server): add outbound webhooks

fortmarek:
Implements [Webhooks for Tuist Events](https://community.tuist.dev/t/rfc-webhooks-for-tuist-events/969) as a first-class account resource. Owners register HTTPS endpoints from the dashboard, subscribe each one to the Tuist events it cares about, and Tuist takes care of signing, retries, and the per-delivery audit log.

#### Domain

- `Tuist.Webhooks.WebhookEndpoint` schema with `account_id`, `name`, `url`, `signing_secret`, `signing_secret_last_four`, and `event_types`. Both `url` and `signing_secret` use `Tuist.Vault.Binary` so they're Cloak-encrypted at rest on `bytea` columns; `signing_secret_last_four` mirrors the cleartext tail so the dashboard never decrypts the full key just to render the masked preview.
- `Tuist.Webhooks` context: `list_endpoints/1`, `list_endpoints_subscribed_to/2`, `get_account_endpoint/2`, `create_endpoint/2`, `update_endpoint/2`, `rotate_signing_secret/1`, `delete_endpoint/1`, plus the delivery-side helpers `list_deliveries/2`, `delivery_stats/2`, `deliveries_timeseries/2`, `get_delivery_attempt/2`. All dashboard reads go through a narrow `@dashboard_fields` projection that omits `signing_secret`; only the delivery worker's `get_endpoint/1` keeps the full load.
- Event catalog: `test_case.created`, `test_case.updated`, `preview.created`, `preview.deleted`. `event_groups/0` returns them grouped by resource so the create modal can render a "Select all Test cases events" toggle next to the individual checkboxes.
- `Tuist.Webhooks.Dispatcher`: call site for the rest of the codebase. `dispatch_test_case_created/2`, `dispatch_test_case_event/3`, `dispatch_preview_created/1`, `dispatch_preview_deleted/1` — each one resolves the account, looks up subscribed endpoints, builds the envelope, and enqueues a delivery job per match. Single-endpoint enqueue failures are swallowed so one bad subscriber can't block delivery to the others.

#### Signing + delivery

- `Tuist.Webhooks.Signature`: HMAC-SHA256 helpers (`sign/3`, `verify/4`, `generate_secret/0`). Signed string is `\"<unix_ts>.<raw_json>\"`, header is `Tuist-Signature: t=<ts>,v1=<hex_sig>`, with a 5-minute replay tolerance and constant-time comparison. New secrets are minted as `tuist_webhook_<base64url(32 random bytes)>` — matches the `tuist_<scope>_<random>` convention used by project tokens, SCIM tokens, and account tokens.
- `Tuist.Webhooks.Workers.DeliveryWorker`: dedicated Oban worker on its own `:webhooks` queue (concurrency 20). `max_attempts: 7` covers the initial send plus six retries on the 1m → 5m → 30m → 2h → 8h → 24h schedule. 10-second request timeout, redirects disabled (otherwise a 3xx to a private/metadata address would bypass the SSRF guard). Endpoint is re-read on every attempt so URL / secret edits and rotations take effect on the next retry; a deleted endpoint is `:discard` instead of retrying forever.
- `Tuist.OAuth2.SSRFGuard.pin/1` + `connect_options/1` are applied before each request. Loopback, RFC1918, link-local, ULA, and cloud-metadata destinations are rejected before bytes leave the process, and the rejection is still recorded as a failed attempt so it shows up in the dashboard.

#### Delivery audit log (ClickHouse)

- `webhook_delivery_attempts` lives in ClickHouse — append-only with bursty fan-out is exactly the workload CH's MergeTree handles best, and the time-bucketed chart queries get dramatically faster than the equivalent PG aggregates would.
- `MergeTree` engine, `PARTITION BY toYYYYMM(inserted_at)`, `ORDER BY (webhook_endpoint_id, inserted_at)`. Skip indexes on `event_type` and `status` accelerate the dashboard's filter paths. Retention is unbounded, matching other customer-facing audit data (`build_runs`, `test_case_runs`, `runner_jobs`).
- Writes go through `Tuist.Ingestion.Bufferable` — `DeliveryAttempt.Buffer.insert/1` accumulates RowBinary in memory and flushes on size or interval, keeping the worker hot path async and avoiding "too many parts" at fan-out time.
- Each row captures `id`, `event_id`, `event_type`, `attempt`, `status` (`delivered`/`failed`), the request body, request headers (JSON-encoded), response status / headers / body, error string, and `duration_ms`. Response bodies truncated at 64 KiB so a chatty upstream can't bloat the row. `response_status: 0` is the sentinel for "no HTTP response received" (network error, SSRF reject).

#### Dashboard

- **Webhooks** tab on the account settings: lists endpoints with name, URL, masked signing secret (`tuist_webhook_•••••XYZ` suffix only, read from the cleartext `signing_secret_last_four` column), the count of subscribed events, and a kebab with `Rotate secret` / `Delete endpoint`. Whole row navigates to the detail page; a custom `StopRowNavigate` JS hook stops the kebab click from being captured by LiveView's window-level `bindNav` (and replays the menu items' `phx-click` so confirm dialogs and event dispatch still work from inside the row link).
- **Create endpoint** modal: name + HTTPS URL + grouped event checkboxes with per-group select-all. On submit the new signing secret is shown once via a `Clipboard` hook; closing the modal removes it from the LiveView assigns. The form body lives in a shared `TuistWeb.Components.WebhookEndpointForm` component used by both the create and edit modals.
- **Endpoint detail** page:
  - Summary card (URL, masked signing secret, subscribed events, created date) with `Edit endpoint` and a kebab for `Rotate secret` / `Delete endpoint`.
  - Event deliveries chart (ECharts line with Total + Failed series, interactive legend toggles, configurable date range).
  - Events table with cursor pagination (Flop, page size 20), filterable by status, event type, and event-id search; rows navigate to per-event detail.
- **Event detail** page: Summary card (status, response code, attempt, duration, sent timestamp, destination), Request card with headers + pretty-printed JSON body, Response card with status / headers / body.

#### API docs

- Six new `OpenApiSpex` schemas under `TuistWeb.API.Schemas.Webhook` (`WebhookTestCase`, `WebhookPreview`, plus a `WebhookXxxEvent` envelope per supported event type) registered into `components.schemas` via `OpenApiSpex.add_schemas/2`.
- A top-level `Webhook events` OpenAPI tag carries a markdown description with a table of supported events, anchored to each event's schema — Redoc renders it as a sidebar section on `tuist.dev/api/docs`. Description lives on the tag (not `info.description`) so Swift OpenAPI Generator doesn't splat it onto the CLI's `Client.swift` docblock.
- `dispatcher_test.exs` casts every emitted payload against the matching schema so drift between the snapshot builders and the documented contract fails CI.

#### Integrations into existing flows

- `Tuist.Tests` ingestion path calls `Dispatcher.dispatch_test_case_created/2` for newly-observed test cases and `dispatch_test_case_event/3` on flagged transitions (`marked_flaky`, `muted`, `unskipped`, …). Both fan-outs share the same \"first run\" filter as the `first_run` `TestCaseEvent` audit row, so the webhook and the audit log agree by construction.
- `Tuist.AppBuilds` calls `dispatch_preview_created/1` after a preview's app build finishes uploading, and `dispatch_preview_deleted/1` after a preview row is removed.

#### Ops / migrations

- `priv/repo/migrations/20260513120000_add_webhook_endpoints_table.exs` — single consolidated PG migration that creates the table with the final shape (bytea-encrypted url + signing_secret, signing_secret_last_four, event_types, indexed on account_id).
- `priv/ingest_repo/migrations/20260515160000_create_webhook_delivery_attempts_table.exs` — CH MergeTree with the partition / ordering described above.
- `priv/repo/seeds.exs` adds two seeded endpoints (Notion automation, Slack relay) with 1500 attempts each on a 97/3 delivered/failed mix so the chart and table render with realistic data locally.

#### Docs + compliance

- New customer-facing guide at `/en/guides/integrations/webhooks` (linked from the Integrations sidebar): event catalog, payload envelope, header reference, signature verification with a Node example (raw-body + constant-time compare), retry schedule, security notes. References the per-event OpenAPI schemas on `tuist.dev/api/docs` for the payload contract.
- Changelog entry `2026.05.15-webhooks.md` with the endpoint detail screenshot as the hero image.
- `server/data-export.md` documents both new tables. Endpoint URL and signing secret are listed as non-exportable bearer credentials (encrypted at rest, often contain path / query tokens). Delivery attempts are exportable; retained indefinitely on the CH side.
- `dashboard_account.pot` extracted for the new strings; no `.po` changes (handled by `tuistit`).

### Tests

- `Tuist.WebhooksTest` — context-level CRUD, masking, time-series bucketing, rotation, secret encryption round-trip.
- `Tuist.Webhooks.SignatureTest` — sign / verify, tolerance, tampered-payload rejection, malformed-header rejection.
- `Tuist.Webhooks.DispatcherTest` — fan-out, account-scope isolation, no-op when there are zero subscribers, envelope shape per event type, **payload-against-schema conformance** for every event (catches drift between dispatcher and OpenAPI schema at CI time).
- `Tuist.Webhooks.Workers.DeliveryWorkerTest` — 2xx → `:delivered`, 5xx → `:failed` + retry, transport error → `:failed`, SSRF rejection records an attempt without making the request, deleted endpoint discards.
- `TuistWeb.WebhooksLiveTest` — index render, create flow with secret disclosure, rotate, delete.
- `TuistWeb.WebhookLiveTest` — detail render, filters, chart data, event navigation, pagination.

### How to test locally — end-to-end with a real receiver

The full path including SSRF guard, real HMAC, real Oban queue, real retry backoff — five minutes of setup.

**1. Start a public tunnel** in one terminal:

```sh
cloudflared tunnel --url http://localhost:4000
```

Copy the `https://…trycloudflare.com` URL it prints.

**2. Register the endpoint.** Sign in to the dev server (`tuistrocks@tuist.dev` / `tuistrocks`), open `/<account>/webhooks`, click **Add endpoint**, paste the cloudflared URL, tick `preview.created` (easiest to trigger), save. Copy the displayed `tuist_webhook_…` secret — you only see it once.

**3. Start the receiver** in a second terminal — a zero-deps Node script lives at `/tmp/tuist-webhook-receiver.mjs` (see the script body in commit history if you want it in-repo). It captures the raw body, verifies the `Tuist-Signature` header in constant time, and prints the decoded payload:

```sh
TUIST_WEBHOOK_SECRET='tuist_webhook_…' node /tmp/tuist-webhook-receiver.mjs
```

**4. Fire an event** from a project pointing at the dev server:

```sh
tuist share <scheme>   # uploads a build → fires preview.created
```

**5. Verify** in the receiver terminal:

```
──────────────────────────────────────────
2026-05-18T09:04:09Z  preview.created  id=b0a1d0e6-…
✓ signature verified
{ \"id\": \"b0a1d0e6-…\", \"type\": \"preview.created\", \"created\": 1779…, \"object\": { … } }
```

The dashboard chart's Total line ticks up and the events table gains a new row; clicking it shows the exact request body + headers + response we just round-tripped.

**Failure paths worth testing:**
- Kill the receiver mid-trigger → dashboard records a failed delivery, retry schedule (1m → 5m → …) kicks in; bring it back and watch the retry succeed.
- Tamper with the secret env var → receiver returns 401, dashboard records `failed` with status 401.
- Register a `http://127.0.0.1:4000` URL instead of the tunnel → SSRFGuard rejects before sending, attempt recorded with `blocked: private_ip_resolved`.

### Run the test suite

```
mix test test/tuist/webhooks_test.exs test/tuist/webhooks/ test/tuist_web/live/webhooks_live_test.exs test/tuist_web/live/webhook_live_test.exs
```


server/assets/app/css/pages/webhook_event.css
```diff
@@ -0,0 +1,154 @@
+#account-webhook-event {
+  display: flex;
+  flex-direction: column;
+  flex-grow: 1;
+  gap: var(--noora-spacing-7);
+  background: var(--noora-surface-background-primary);
+  /* Outer frame around the page, plus an inner indent that keeps the
```
pepicrft: I'd maybe drop these comments.

server/lib/tuist_web/components/webhook_endpoint_form.ex
```diff
@@ -0,0 +1,149 @@
+defmodule TuistWeb.Components.WebhookEndpointForm do
```
pepicrft: I'd expect `TuistWeb.Components` to be an umbrella for Tuist-wide components. Since this is specific to the webhooks domain, what about placing it inside the LiveView where it's being used, or close to that namespace?

server/lib/tuist_web/components/layouts/headerbar.html.heex
```diff
@@ -0,0 +1,10 @@
+<main class="layout">
```
pepicrft, line 1: What's special about the webhook layout?

server/lib/tuist_web/api/spec.ex
```diff
@@ -46,8 +69,36 @@ defmodule TuistWeb.API.Spec do
...
+  # because Swift OpenAPI Generator splats `info.description` onto
+  # the generated `Client.swift` as a leading docblock — which makes
+  # no sense on an HTTP client struct. Redoc still renders this
+  # block as a sidebar section.
+  defp webhook_events_description do
+    """
+    Tuist can POST event notifications to HTTPS endpoints you register on your
+    account. Each delivery is a JSON envelope with `id`, `type`, `created`, and
+    an event-specific `object` payload, signed via the `Tuist-Signature` header.
+    See the [webhooks integration guide](https://docs.tuist.dev/en/guides/integrations/webhooks)
```
pepicrft: This URL is not correct.

server/lib/tuist_web/api/spec.ex
```diff
@@ -46,8 +69,36 @@ defmodule TuistWeb.API.Spec do
...
+    for the full payload, signature verification, and retry rules.
+
+    The supported event types and their payload shapes:
+
+    | Event type            | When it fires                                                              | Payload schema |
```
pepicrft: This one is a good candidate to become outdated. Can we create a compile-time dependency with the related models such that we can generate this at compile time?

server/lib/tuist/webhooks.ex
```diff
@@ -0,0 +1,395 @@
...
+
+  @doc """
+  Creates an endpoint. The plaintext signing secret is generated server-side
+  and returned alongside the persisted struct so the caller can show it to
+  the user exactly once.
+  """
+  def create_endpoint(account_id, attrs) do
+    plaintext_secret = Signature.generate_secret()
+
+    attrs =
```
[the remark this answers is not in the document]
fortmarek: Rejected — the finding talks itself out of the alleged `IngestRepo`-as-read violation (which doesn't exist; reads go through `ClickHouseRepo`), then mentions an unspecified "Missing `type: :timestamptz`" issue with no location or details. The consolidated migration already uses `timestamps(type: :timestamptz)` and no other timestamp column on the PR omits it. Nothing actionable.

server/lib/tuist/webhooks.ex
```diff
@@ -0,0 +1,395 @@
...
+
+  @doc """
+  Loads a single endpoint by id, regardless of account. Callers that operate
+  inside an account scope must compare `account_id` themselves before acting.
+
+  Returns the full struct, including the decrypted `signing_secret` —
+  this is the read path the delivery worker uses.
+  """
+  def get_endpoint(id) do
+    case Repo.get(WebhookEndpoint, id) do
```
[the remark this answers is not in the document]
fortmarek: Rejected — the function's docstring already marks it as intentionally cross-tenant: "Loads a single endpoint by id, regardless of account. Callers that operate inside an account scope must compare `account_id` themselves before acting … this is the read path the delivery worker uses." Renaming to `get_endpoint_for_worker/1` would falsely imply the worker is the only legitimate caller; the current docstring conveys the same constraint without baking caller identity into the function name.

server/lib/tuist/webhooks/dispatcher.ex
```diff
@@ -0,0 +1,188 @@
...
+           Webhooks.list_endpoints_subscribed_to(account_id, "test_case.updated") do
+      object = test_case_snapshot(test_case)
+
+      Enum.each(endpoints, fn endpoint ->
+        enqueue(endpoint, "test_case.updated", %{
+          "object" => object,
+          "events" => Enum.map(event_types, &Atom.to_string/1),
+          "actor_id" => actor_id,
+          "alert_id" => alert_id
+        })
```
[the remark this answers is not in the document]
fortmarek: Done in 7fdcdbbf4d — each dispatcher entry point now builds a list of `Oban.Job` changesets and calls `Oban.insert_all/1` once per dispatch. The worst-case `dispatch_test_case_created/2` (nested test_cases × endpoints loop) drops from N×M round-trips to a single batch. Per-endpoint error swallowing carries over as a batch-level `try/rescue` since a DB-level failure would have killed every individual insert too.

server/lib/tuist/webhooks.ex
```diff
@@ -0,0 +1,393 @@
...
+  """
+  def list_endpoints(account_id) do
+    Repo.all(
+      from(e in WebhookEndpoint,
+        where: e.account_id == ^account_id,
+        order_by: [asc: e.inserted_at],
+        select: struct(e, ^@dashboard_fields)
+      )
+    )
+  end
```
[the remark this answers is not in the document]
fortmarek: Rejected — re-skin of last round's P2 with the same shape and same answer. The only production caller is `Tuist.Webhooks.Workers.DeliveryWorker.perform/1`, which takes `endpoint_id` from an Oban job's args we minted server-side in the dispatcher from an already-authenticated row. Zero controller / LiveView callers (`grep Webhooks.get_endpoint lib/` confirms). The suggested "use the scoped variant" doesn't help — the worker only has `endpoint_id` in its args, so getting `account_id` would require denormalizing it into the job, which is no safer than what we have. "Make it private" would require collapsing the worker into the context module. The docstring already marks the function as intentionally cross-tenant and names the worker as the read path; that's the right boundary.

server/lib/tuist/webhooks.ex
```diff
@@ -0,0 +1,393 @@
...
+  @doc """
+  Loads a single endpoint by id, regardless of account. Callers that operate
+  inside an account scope must compare `account_id` themselves before acting.
+
+  Returns the full struct, including the decrypted `signing_secret` —
+  this is the read path the delivery worker uses.
+  """
+  def get_endpoint(id) do
+    case Repo.get(WebhookEndpoint, id) do
+      nil -> {:error, :not_found}
```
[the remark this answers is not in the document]
fortmarek: Rejected — third re-skin of the same `get_endpoint/1` cross-tenant finding (P2 → P1 → P1). Same answer as the previous two: only production caller is `DeliveryWorker.perform/1`, which gets `endpoint_id` from server-minted Oban job args; no controller / LiveView caller. The "attacker guesses UUID" framing presumes an HTTP endpoint exposes this function — it doesn't (`grep Webhooks.get_endpoint lib/` lists only `delivery_worker.ex`). The docstring is explicit that the function is cross-tenant by design and names the worker as the read path.

server/lib/tuist/webhooks/dispatcher.ex
```diff
@@ -0,0 +1,190 @@
...
+
+    with %Projects.Project{account_id: account_id} <- Repo.get(Projects.Project, test_case.project_id),
+         [_ | _] = endpoints <-
+           Webhooks.list_endpoints_subscribed_to(account_id, "test_case.updated") do
+      object = test_case_snapshot(test_case)
+
+      body = %{
+        "object" => object,
+        "events" => Enum.map(event_types, &Atom.to_string/1),
+        "actor_id" => actor_id,
```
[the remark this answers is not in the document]
fortmarek: Rejected on two grounds. (1) Same pattern as the other Blick cross-tenant findings on this PR: the dispatcher is called from server-internal write paths (`Tuist.Tests.update_test_case/3` and `Tuist.Tests.create_test_cases/4`), both of which receive a `test_case` / `project_id` that's already been authorized by an upstream controller or LiveView before reaching the ingestion path. There is no user-controllable input here. (2) The suggested fix is mechanically incoherent — the dispatcher doesn't *have* an `account_id` at this point; resolving the project's `account_id` is the entire purpose of the `Repo.get/2` call. Filtering by the value you're trying to derive doesn't add safety, it just changes the failure mode from `nil` to a different `nil`.

pepicrft (commented):
Posting the drafted comments.

server/lib/tuist_web/controllers/api/previews_controller.ex
```diff
@@ -386,6 +386,12 @@ defmodule TuistWeb.API.PreviewsController do
             preload: [:app_builds, :created_by_account]
           )
 
+        # Fire a `preview.created` webhook for any account-scoped endpoint
+        # subscribed to that event. Best-effort: the dispatcher swallows
+        # missing-project or no-subscriber paths so a webhook problem can't
+        # block the upload response.
+        Tuist.Webhooks.Dispatcher.dispatch_preview_created(preview)
```
pepicrft: I validated this locally. `find_or_create_preview/1` reuses an existing preview for identical metadata, so this path can enqueue `preview.created` more than once for the same preview resource. A quick `MIX_ENV=test mix run -e ...` repro gave me `same_preview=true`, `job_count=2`, and the same preview id in both payloads. If the intent is one event per preview, I think this dispatch needs to be gated to the branch where the preview row is actually created.

server/lib/tuist_web/live/webhooks_live.ex
```diff
@@ -0,0 +1,177 @@
...
+           "name" => assigns.create_form_name,
+           "url" => assigns.create_form_url,
+           "event_types" => assigns.create_form_event_types
+         }) do
+      {:ok, endpoint, plaintext_secret} ->
+        socket =
+          socket
+          |> assign_endpoints()
+          |> reset_create_form()
+          |> assign(:disclosure, %{endpoint: endpoint, plaintext_secret: plaintext_secret, mode: :created})
```
pepicrft, line 102: I validated this locally too. `get_account_endpoint/2` gives the dashboard path a struct with `signing_secret=nil`, but `create_endpoint/2` and `rotate_signing_secret/1` return full structs with the decrypted secret, and this assign stores that struct in LiveView state. My local repro printed `scoped_secret=nil` and then `rotated_has_secret=true`. That seems to undercut the defense-in-depth goal here, and the detail page has the same pattern.

pepicrft (commented):
Validated these two concerns locally.

server/lib/tuist_web/components/layouts/headerbar.html.heex
fortmarek: it looks odd to retain the top tab in the account part of the layout ... this will get simpler once we move away from the current top tab pattern to what we have in the project part of the dashboard

server/assets/app/css/pages/webhook_event.css
fortmarek: Dropped — the padding declaration carries the same information.

server/lib/tuist_web/components/webhook_endpoint_form.ex
fortmarek: Moved out of the `TuistWeb.Components` umbrella — file is now `lib/tuist_web/live/webhook_endpoint_form.ex`, module renamed to `TuistWeb.WebhookEndpointForm`, alongside its `WebhookLive` / `WebhooksLive` consumers.

server/lib/tuist_web/components/layouts/headerbar.html.heex
fortmarek: Nothing webhook-specific — it's the minimal authenticated chrome (just the headerbar) for account-scoped pages that drill past the account-settings tab nav. The name was misleading. Renamed the file to `layouts/headerbar.html.heex` and the router binding to `:headerbar` so any future drill-down page can reuse it without the misleading naming.

server/lib/tuist_web/api/spec.ex
fortmarek: Fixed — the docs live at `tuist.dev/<locale>/docs/...` (no `docs.` subdomain, `/docs/` is part of the path). Updated to `https://tuist.dev/en/docs/guides/integrations/webhooks`.

server/lib/tuist_web/api/spec.ex
fortmarek: Done — the table is now built at compile time from `Tuist.Webhooks.WebhookEndpoint.event_groups/0` (new `@webhook_events_table_rows` module attribute that iterates the catalog and derives each schema title from the dotted event type). Touching the catalog forces `WebhookEndpoint` to recompile, which forces `Spec` to recompile via the compile-time call, so the description can't drift from the source of truth.

server/lib/tuist_web/controllers/api/previews_controller.ex
fortmarek: Moved the dispatch to the `start` action and gated it on a new `:created`/`:reused` tag returned by `find_or_create_preview/1` (commit 1873bf604c), so a reused preview no longer redelivers `preview.created`. The dispatcher doc was updated to match the new semantic.

server/lib/tuist_web/live/webhooks_live.ex
fortmarek: Scrubbed `signing_secret` to nil on the structs returned by `Webhooks.create_endpoint/2` and `Webhooks.rotate_signing_secret/1` (commit 1873bf604c). The plaintext is still returned through the separate tuple element for the one-time reveal, and the worker-path `get_endpoint/1` keeps the full decrypted load, but the dashboard assigns no longer carry the plaintext key.

server/lib/tuist/webhooks/workers/delivery_worker.ex
```diff
@@ -0,0 +1,192 @@
...
+              # SSRFGuard.pin/1 only validates the first hop. Following
+              # a 3xx would bypass the guard if the upstream redirects
+              # to a private / metadata address, so we treat redirects
+              # as a failed delivery and let the receiver land on the
+              # canonical URL themselves.
+              redirect: false,
+              connect_options: SSRFGuard.connect_options(hostname)
+            )
+
+          {resp, classify(resp)}
```
pepicrft: I'd make the first element of the tuple the result, and the second one the response, in case the caller needs to extract extra information from it.

server/lib/tuist/webhooks/workers/delivery_worker.ex
```diff
@@ -0,0 +1,192 @@
...
+      response_status: response_status(response),
+      response_headers: encode_headers(response_headers_map(response)),
+      response_body: response_body(response),
+      error: error_message(result),
+      duration_ms: fields.duration_ms,
+      inserted_at: DateTime.utc_now()
+    }
+
+    DeliveryAttempt.Buffer.insert(row)
+  rescue
```
pepicrft: This reads very defensive. Isn't the buffer holding changesets in memory to flush its persistence into CH asynchronously? If so, I don't see why we'd need to be defensive here. Also, we should have visibility over these errors, we won't notice when things are not working as expected.

server/test/tuist/webhooks/workers/delivery_worker_test.exs
```diff
@@ -0,0 +1,120 @@
...
+      Webhooks.create_endpoint(account.id, %{
+        "name" => "Hook",
+        "url" => "https://example.com/hook",
+        "event_types" => ["test_case.updated"]
+      })
+
+    %{endpoint: endpoint, plaintext: plaintext}
+  end
+
+  defp job_args(endpoint, opts \\ []) do
```
pepicrft: Many tests in this file create an account, create an endpoint, and get the job args, which makes me think we might want to move them to `setup` and return them there such that test scenarios can take them from the map passed by `ExUnit`

server/lib/tuist/webhooks/workers/delivery_worker.ex
```diff
@@ -0,0 +1,192 @@
...
+    # response received", which the dashboard already branches on.
+    row = %{
+      id: Ecto.UUID.generate(),
+      webhook_endpoint_id: endpoint.id,
+      event_id: fields.event_id,
+      event_type: fields.event_type,
+      attempt: fields.attempt,
+      status: if(result == :ok, do: "delivered", else: "failed"),
+      request_body: fields.request_body,
+      request_headers: encode_headers(fields.request_headers),
```
pepicrft: This contains `Tuist-Signature`, which is replayable, so I'd consider omitting it.

server/lib/tuist/webhooks/workers/delivery_worker.ex
```diff
@@ -0,0 +1,192 @@
...
+      response_status: response_status(response),
+      response_headers: encode_headers(response_headers_map(response)),
+      response_body: response_body(response),
```
pepicrft: These come from the customer and might contain sensitive information. Would it make sense to redact the headers that might contain sensitive information?

server/test/tuist/webhooks/workers/delivery_worker_test.exs
```diff
@@ -0,0 +1,120 @@
+defmodule Tuist.Webhooks.Workers.DeliveryWorkerTest do
```
pepicrft, line 1: I'm missing some tests here that check that the retries are persisted with the right information in them.

server/priv/docs/en/guides/integrations/webhooks.md
```diff
@@ -0,0 +1,174 @@
...
+## Event types {#event-types}
+
+Tuist groups events by the resource they describe. The `type` field in the envelope uses dotted notation (`{resource}.{action}`) — same as the value you'd subscribe to in the dashboard.
+
+| Type                  | When it fires                                                                                  |
+| --------------------- | ---------------------------------------------------------------------------------------------- |
+| `test_case.created`   | A test case was observed for the first time in this account.                                   |
+| `test_case.updated`   | A test case's attributes changed — flakiness flag, state transition (muted, skipped), etc.     |
+| `preview.created`     | A new preview was created (fires once the app build finishes uploading).                       |
+| `preview.deleted`     | A preview was deleted from this account.                                                       |
```
pepicrft: Can we generate this one with the compile-time data that we have? Otherwise, it's prone to becoming out of sync.

server/priv/docs/en/guides/integrations/webhooks.md
```diff
@@ -0,0 +1,174 @@
...
+| ------------------- | --------------------------------------------------------------------------------------------- |
+| `Content-Type`      | Always `application/json`.                                                                    |
+| `User-Agent`        | `Tuist-Webhooks/1.0`.                                                                         |
+| `Tuist-Event-Id`    | Same UUID as `id` in the body — convenient if you log headers before parsing the payload.     |
+| `Tuist-Event-Type`  | Same string as `type` in the body.                                                            |
+| `Tuist-Signature`   | The HMAC-SHA256 signature, formatted as `t={timestamp},v1={hex_digest}` — see below.          |
+
+## Responding to a webhook {#responding}
+
+To acknowledge a delivery, your endpoint should return any `2xx` status code. Tuist treats anything else (3xx, 4xx, 5xx) — along with connection failures and timeouts — as a delivery failure and will retry.
```
pepicrft: 3xx are not failures, are they?

server/priv/docs/images/guides/integrations/webhooks/event-detail.png
pepicrft, line 1: Just looking at this image, I'm missing a button to copy the payload.

pepicrft: And another one to copy the whole request, potentially as a `.har` file that I can then throw into an agent to investigate issues in my server handling Tuist requests.

server/priv/docs/en/guides/integrations/webhooks.md
```diff
@@ -0,0 +1,174 @@
...
+| `preview.deleted`     | A preview was deleted from this account.                                                       |
+
+More event types will follow. If you need one we don't yet emit, let us know and we'll add it to the catalog.
+
+## Create an endpoint {#create-an-endpoint}
+
+Endpoints are configured per account. Open **Webhooks** from your account settings, click **Add endpoint**, and provide:
+
+- **Name** — how the endpoint appears in the listing.
+- **Endpoint URL** — must use `https://`. Private network destinations (loopback, RFC1918, link-local, cloud metadata addresses) are rejected at delivery time so a misconfigured webhook can't be used to probe internal services.
```
pepicrft: Do other platforms mention anything about safety measures like these? I'd have expected that to be an implementation detail.

server/lib/tuist/webhooks/signature.ex
```diff
@@ -0,0 +1,99 @@
...
+  Returns the signature header value for `payload` signed with `secret` at
+  `timestamp` (Unix seconds).
+  """
+  def sign(payload, secret, timestamp) when is_binary(payload) and is_binary(secret) and is_integer(timestamp) do
+    signed_payload = "#{timestamp}.#{payload}"
+    digest = :crypto.mac(:hmac, :sha256, secret, signed_payload)
+    "t=#{timestamp},#{@scheme}=#{Base.encode16(digest, case: :lower)}"
+  end
+
+  @doc """
```
pepicrft, line 30: Small security nit: the timestamp check here does not fully prevent replay, it only bounds the replay window. If someone captures a valid request, they can resend it within the tolerance and `verify/4` will still return `:ok`. Since the payload already carries a stable id and retries reuse it, I think the docs/comments should say receivers must dedupe on id or Tuist-Event-Id, not just check timestamp.

I’d reword “prevents replay attacks” to “limits replay window; combine with idempotency on delivery id for full replay defense.

server/lib/tuist/webhooks/webhook_endpoint.ex
```diff
@@ -0,0 +1,153 @@
...
+    field :signing_secret, Binary
+    field :signing_secret_last_four, :string
+    field :event_types, {:array, :string}, default: []
+
+    belongs_to :account, Account
+
+    timestamps(type: :utc_datetime)
+  end
+
+  def create_changeset(endpoint, attrs) do
```
pepicrft, line 86: I'm missing some tests for this one.

server/lib/tuist/webhooks/webhook_endpoint.ex
```diff
@@ -0,0 +1,153 @@
...
+      :signing_secret_last_four,
+      :event_types
+    ])
+    |> validate_required([:account_id, :name, :url, :signing_secret, :signing_secret_last_four])
+    |> validate_length(:name, min: 1, max: 100)
+    |> validate_https_url()
+    |> validate_event_types()
+  end
+
+  def update_changeset(endpoint, attrs) do
```
pepicrft, line 102: Same here, I'd add some tests.

server/lib/tuist/tests.ex
```diff
@@ -1649,29 +1674,49 @@ defmodule Tuist.Tests do
...
-          }
-        end)
+  # Webhook fan-out for the same set of test cases that get a `first_run`
+  # audit-log row. Best-effort: failures don't abort the write that
+  # produced the event.
+  defp dispatch_test_case_created_webhooks(_project_id, _test_cases, []), do: :ok
 
-      TestCaseEvent.Buffer.insert_all(events)
-    end
+  defp dispatch_test_case_created_webhooks(project_id, test_cases, first_run_test_case_runs) do
```
pepicrft, line 1712: Where are we testing that  some logic in `Tuist.Tests` has dispatching of webhooks as a side-effect feature.

server/lib/tuist/tests.ex
```diff
@@ -1649,29 +1674,49 @@ defmodule Tuist.Tests do
...
+    first_run_ids = MapSet.new(first_run_test_case_runs, & &1.test_case_id)
+    new_test_cases = Enum.filter(test_cases, &MapSet.member?(first_run_ids, &1.id))
+    Dispatcher.dispatch_test_case_created(project_id, new_test_cases)
+    :ok
+  rescue
```
pepicrft: What are we rescuing from? This reads coding agent trying to be too defensive.

pepicrft (approved):
When clicking "select all", there's a slight delay, likely because we are going to the server and back to the client to update the state. I'd investigate what it'd take to update those optimistically somehow.

Image

server/lib/tuist/webhooks/workers/delivery_worker.ex
fortmarek: Swapped the destructuring to `{result, response}` (commit 3548b7c190).

server/lib/tuist/webhooks/workers/delivery_worker.ex
fortmarek: Dropped the rescue around `DeliveryAttempt.Buffer.insert/1` (commit 3548b7c190). The buffer is in-memory and a real persistence failure should surface to Oban rather than be silently logged.

server/test/tuist/webhooks/workers/delivery_worker_test.exs
fortmarek: Refactored to a single `setup` block that builds the endpoint and base job args, with scenarios destructuring from the context map (commit 3548b7c190).

server/lib/tuist/webhooks/workers/delivery_worker.ex
fortmarek: `Tuist-Signature` is now redacted in the persisted request-headers JSON; the underlying `headers` list still carries it for the outbound POST so receivers continue to see the real signature (commit 3548b7c190).

server/lib/tuist/webhooks/workers/delivery_worker.ex
fortmarek: Added a case-insensitive redaction pass for sensitive response headers — `set-cookie`, `cookie`, `authorization`, `proxy-authorization`, `x-auth-token` — before the row hits ClickHouse (commit 3548b7c190).

server/test/tuist/webhooks/workers/delivery_worker_test.exs
fortmarek: Added an `attempt persistence` describe block with three tests covering the retry attempt number on the persisted row, the upstream-status-and-error pair on a non-2xx, and the `response_status: 0` sentinel on a network error — plus two redaction tests that assert the new header redaction. (commit 3548b7c190)

server/priv/docs/en/guides/integrations/webhooks.md
fortmarek: Auto-generated now — added a `{{webhook_events_table}}` macro to `Tuist.Docs.Loader` that interpolates rows built from `WebhookEndpoint.event_groups/0` at compile time, and replaced the hand-written table in the guide with the placeholder. Touching the catalog forces the docs page to recompile (commit 3548b7c190).

server/priv/docs/en/guides/integrations/webhooks.md
fortmarek: Reworded — the guide now says "anything outside the 2xx range" and explicitly notes Tuist doesn't follow 3xx, so receivers should respond at the canonical URL rather than redirect (commit 3548b7c190).

server/priv/docs/images/guides/integrations/webhooks/event-detail.png
fortmarek: Added "Copy payload" and "Copy body" buttons to the request and response cards on the event detail page (commit 3548b7c190). The `.har` export is a bigger piece of work — happy to do it, but I'd rather land it as a separate PR so this one doesn't grow further; I can open a follow-up if you want to track it.

server/priv/docs/en/guides/integrations/webhooks.md
fortmarek: Removed the SSRF detail from the "Endpoint URL" bullet; the same point already lives in the Security notes section where it fits naturally (commit 3548b7c190).

server/lib/tuist/webhooks/signature.ex
fortmarek: Updated the moduledoc to say the timestamp check only "bounds the replay window" and explicitly direct receivers to dedupe on the payload `id` (`Tuist-Event-Id`). Mirrored the wording in the integration guide and added a dedicated step #4 in the verification recipe (commit 3548b7c190).

server/lib/tuist/webhooks/webhook_endpoint.ex
fortmarek: Added a new `test/tuist/webhooks/webhook_endpoint_test.exs` with focused coverage for `create_changeset/2` (full-attrs happy path, required-field errors, blank/long-name rejection, non-HTTPS rejection, empty/unsupported event-type rejection, every supported type accepted) — commit 3548b7c190.

server/lib/tuist/webhooks/webhook_endpoint.ex
fortmarek: Same file — added coverage for `update_changeset/2` covering name/url/event_types updates, that `signing_secret` stays untouched (rotation is the only way to change it), blank-name rejection, non-HTTPS rejection, and empty/unsupported event-type rejection (commit 3548b7c190).

server/lib/tuist/tests.ex
fortmarek: Added a test under `describe "create_test/1"` in `tests_test.exs` that creates an endpoint subscribed to `test_case.created`, runs a full `Tests.create_test/1` with two new test cases, and asserts both Oban delivery jobs were enqueued with the right `webhook_endpoint_id`, `event_type`, and payload — plus that an endpoint subscribed to a *different* event type doesn't receive one (commit 3548b7c190).

server/lib/tuist/tests.ex
fortmarek: Dropped the rescue — `Dispatcher.dispatch_test_case_created/2` already wraps the `Oban.insert_all/1` call in its own try/rescue with logging, so the outer one was strictly defensive and was just hiding real failures. (commit 3548b7c190)

fortmarek:
> When clicking "select all", there's a slight delay, likely because we are going to the server and back to the client to update the state. I'd investigate what it'd take to update those optimistically somehow.

This is a local server issue, you can observe that across all Noora components.

server/lib/tuist/webhooks.ex
```diff
@@ -0,0 +1,398 @@
...
+      endpoint -> {:ok, endpoint}
+    end
+  end
+
+  @doc """
```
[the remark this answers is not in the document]
fortmarek: Rejecting — this is the documented cross-tenant load path for the delivery worker, and the docstring on `get_endpoint/1` (lines 64-70) already says exactly that. The endpoint_id reaches the worker via an Oban job enqueued by the dispatcher, which loads endpoints through `list_endpoints_subscribed_to/2` (account-scoped). The dashboard / LiveView paths go through `get_account_endpoint/2`. Same finding has been raised and rejected several times against this PR; the structure is intentional.

PR #10751, opened 2026-05-12
fix(server): optimize clickhouse dashboard queries

fortmarek:
## Summary

This PR reduces ClickHouse CPU and memory pressure from several high-frequency marketing stats, dashboard, and analytics queries.

It keeps the previous `test_case_runs` materialized-view rehydration optimization intact, then narrows the remaining expensive `FINAL` lookup with a `ran_at` range, adds short-lived application caching for polling-style marketing/dashboard queries, adds a partition-pruned fast path for build-run detail lookups, prevents an unscoped `test_suite_runs` full-table sort, and rebuilds the active daily test-case stats table as exact daily presence rows so we avoid `uniqExactMerge` memory pressure without returning approximate counts.

## Context

Recent production query analysis showed four main sources of avoidable ClickHouse work:

- High-frequency marketing stats polling queries repeatedly counted recent `test_runs`, `build_runs`, test case runs, and cache artifacts globally with no application-level caching.
- The 24h `test_runs` count used `FINAL`, forcing a runtime dedupe merge even though the query only needs a recent aggregate count.
- Build detail lookups queried `build_runs` by `project_id + id` but did not include an `inserted_at` predicate, so ClickHouse could not prune monthly partitions.
- `test_case_runs FINAL` rehydration was still scanning all historical rows for selected test-case IDs after the slim MV page was selected.
- `test_case_runs_active_daily_stats` used `uniqExactMerge`, which can allocate large UUID sets for large projects.
- An unscoped `test_suite_runs ORDER BY inserted_at DESC LIMIT 5` path could scan/sort the whole table.

## Queries Addressed

The changes in this PR map to the following production queries from the investigation.

### High-frequency global marketing stats polling

These global count queries were part of the high-frequency query group that totaled ~228 CPU-sec over the 3-hour analysis window. They run thousands of times per hour from the global `Tuist.Marketing.Stats` poller and are good candidates for a very short application cache.

```sql
SELECT count(*)
FROM test_runs AS t0 FINAL
WHERE t0.inserted_at >= ?
```

- Observed cost: 4,729 executions over 3 hours.
- Read volume: 10.3 GiB.
- Scope: global across all projects; this feeds the marketing stats counters, not a per-project dashboard.
- Problem: `FINAL` is unnecessary for this recent global marketing count and forces runtime dedupe work.
- PR change: remove `FINAL` and cache `last_24h_test_run_count/0` for 10 seconds.

```sql
SELECT count(*)
FROM build_runs AS b0
WHERE b0.inserted_at >= ?
```

- Observed cost: 6,389 executions over 3 hours.
- Read volume: 7.3 GiB.
- Rows scanned: 973M.
- Scope: global across all projects; this feeds the marketing stats counters.
- Problem: repeated polling without caching.
- PR change: cache `last_24h_build_count/0` for 10 seconds.

### Build-run detail lookup

```sql
SELECT b0.id, b0.duration, b0.project_id, ...
FROM build_runs AS b0
WHERE (b0.project_id = ?)
  AND (b0.id = '<single-uuid>')
ORDER BY b0.inserted_at DESC
LIMIT 1
```

- Observed cost: 1,505 executions over 3 hours.
- Observed latency: p90 17ms.
- Rows scanned: 119,627 average, 152,664 max.
- Read volume: 2.12 MiB average.
- Projection usage: `projections_used: []`.
- Problem: this is a point lookup for a single build-run detail page/callback, but `build_runs` is partitioned by `toYYYYMM(inserted_at)`. Without an `inserted_at` predicate, ClickHouse checks every monthly partition even though the primary key is `(project_id, id)`.
- Additional finding: an id-only shape did use `default.build_runs.proj_by_id`, but still read 168,607 rows on average, so fixing projection usage alone does not solve the cross-partition scan.
- PR change: `Builds.get_build/2` first tries a 90-day `inserted_at`-bounded lookup so ClickHouse can prune recent monthly partitions, then falls back to the previous unbounded lookup to preserve old build URLs.

### `test_case_runs FINAL` rehydration

```sql
SELECT t0.id, t0.name, t0.test_run_id, ...
FROM test_case_runs AS t0 FINAL
WHERE t0.project_id IN ?
  AND t0.test_case_id IN (uuid1, uuid2, ... ~100+ UUIDs)
```

- Observed latency: p90 5,250ms in the slow-query analysis.
- Observed CPU: 78 CPU-sec over 3 hours.
- Per-query profile: ~1.0s average CPU / 485ms average wall time.
- Read volume: 37 GiB total in the CPU window.
- Table context: ~1.1B rows, 70 GiB compressed, `SharedReplacingMergeTree`, sort key `(project_id, test_case_id, ran_at, id)`.
- Problem: even after using the slim MV to choose the page, the final rehydration query could still scan all historical runs for the selected test-case IDs.
- PR change: keep the previous MV optimization, but add a `ran_at` min/max range from the selected slim results so the `FINAL` lookup is bounded by time and aligned with the sort key.

### Unscoped `test_suite_runs` listing

```sql
SELECT id, name, test_run_id, ..., inserted_at
FROM test_suite_runs
ORDER BY inserted_at DESC
LIMIT 5
```

- Observed cost: 33 CPU-sec from a single execution.
- Observed latency: 36 seconds wall time.
- Rows scanned: 150M.
- Read volume: 24.8 GiB.
- Problem: no `WHERE` predicate means ClickHouse has to scan/sort the full table.
- PR change: add a default 7-day `inserted_at` lower bound for unscoped `list_test_suite_runs/1` calls while preserving scoped queries by `test_run_id`, `test_module_run_id`, or `shard_id`.

### Active test-case distinct count

```sql
SELECT uniqExactMerge(test_case_ids_state)
FROM test_case_runs_active_daily_stats AS t0
WHERE t0.project_id = 1227
  AND t0.date BETWEEN ? AND ?
```

- Observed latency: p90 2,330ms.
- Observed memory: ~1.26 GiB per execution.
- Table context: project 1227 accounted for 9,442 of 10,229 total rows (~92%).
- Existing table shape: `SharedAggregatingMergeTree`, sort key `(project_id, date, is_ci)`, no partition key.
- Problem: `uniqExactMerge` has to hold exact UUID sets in memory, and the table lived in one logical partition.
- PR change: rebuild the stats table as exact daily presence rows keyed by `(project_id, date, is_ci, test_case_id)`, query it by grouping `test_case_id` over the 14-day window, and partition the table monthly with `PARTITION BY toYYYYMM(date)`.

## Changes

- Added 10-second app-level caching for global marketing stats counts:
  - `Tuist.Tests.last_24h_test_run_count/0`
  - `Tuist.Tests.last_24h_test_case_run_count/0`
  - `Tuist.Tests.last_24h_flaky_test_case_run_count/0`
  - `Tuist.Builds.last_24h_build_count/0`
  - `Tuist.Cache.last_24h_artifacts_count/0`
- Removed unnecessary `FINAL` from the 24h `test_runs` count.
- Cached `Tuist.Builds.recent_build_status_counts/2` for 10 seconds.
- Added a 90-day partition-pruned fast path to `Tuist.Builds.get_build/2`, with an unbounded fallback for older builds.
- Added a 7-day `inserted_at` floor to unscoped `list_test_suite_runs/1` calls while preserving scoped queries by `test_run_id`, `test_module_run_id`, or `shard_id`.
- Added `ran_at` min/max bounds to the `test_case_runs FINAL` rehydration query, using the already-selected slim MV results as the source of truth.
- Switched active daily test-case analytics reads from `uniqExactMerge(test_case_ids_state)` to an exact grouped count over daily presence rows.
- Added a ClickHouse ingest migration that rebuilds `test_case_runs_active_daily_stats` as:
  - exact rows containing `project_id`, `date`, `is_ci`, and `test_case_id`
  - `PARTITION BY toYYYYMM(date)`
- Updated the active daily stats schema docs and data export docs to describe exact daily presence rows.

## Migration Details

The new ingest migration rebuilds `test_case_runs_active_daily_stats` through a replacement table:

1. Creates `test_case_runs_active_daily_stats_replacement` with exact daily presence columns and monthly partitioning.
2. Creates a replacement materialized view to capture live writes during the initial backfill.
3. Backfills from active `test_case_runs` source partitions.
4. Drops the old canonical MV.
5. Drops the replacement MV before the table swap, because ClickHouse materialized views keep their `TO` target table in the definition.
6. Exchanges the replacement table into the canonical table name.
7. Recreates the canonical MV targeting the canonical table.
8. Runs a small catch-up backfill from the migration start timestamp to cover the short DDL window.
9. Drops the old table now held under the replacement name.

The down migration rebuilds the old `uniqExact` aggregate shape without the monthly partition key.

## Expected Impact

- The high-frequency global marketing stats polling paths should stop issuing a fresh ClickHouse query on every 5-second poll and instead coalesce within a 10-second window.
- The `test_runs` count avoids `FINAL` runtime dedupe work.
- Recent build detail lookups should read fewer monthly partitions while old build detail URLs continue to work through the fallback.
- The remaining `test_case_runs FINAL` rehydration query keeps its exact/deduped behavior but gets a `ran_at` range predicate aligned with the existing sort key.
- Active test-case analytics remain exact, but avoid merging large per-day exact UUID-set states. Instead, the query scans precomputed daily presence rows for the 14-day window and groups by `test_case_id`.
- Unscoped test-suite-run listings no longer trigger an all-history sort.

## Validation

- Ran formatting:
  - `mix format`
- Ran focused tests:
  - `mix test test/tuist/builds_test.exs:625 test/tuist/builds_test.exs:1486 test/tuist/cache_test.exs test/tuist/tests_test.exs:658 test/tuist/tests_test.exs:1036 test/tuist/tests_test.exs:8061 test/tuist/tests_test.exs:8076 test/tuist/tests_test.exs:8092 test/tuist/tests/analytics_test.exs:2547`
  - Result: `18 tests, 0 failures`
  - `mix test test/tuist/builds_test.exs:200 test/tuist/builds_test.exs:214 test/tuist/builds_test.exs:231`
  - Result: `3 tests, 0 failures`
- Ran local ClickHouse migration validation:
  - `mix ecto.migrate -r Tuist.IngestRepo`
- Verified local ClickHouse schema and query behavior:
  - `test_case_runs_active_daily_stats` has exact presence columns: `project_id`, `date`, `is_ci`, `test_case_id`.
  - `test_case_runs_active_daily_stats` is partitioned by `toYYYYMM(date)`.
  - `test_case_runs_active_daily_stats_mv` targets the canonical stats table.
  - Exact presence count matched exact raw-table distinct count in local ClickHouse: `474` vs `474`.
  - Local `EXPLAIN indexes = 1, projections = 1` for the new bounded `build_runs` lookup pruned `build_runs` from 14/14 parts to 4/14 parts in the local database, while preserving the `(project_id, id)` primary-key lookup.
- Ran:
  - `git diff --check`

## Notes And Trade-Offs

- Active test-case counts stay exact. The trade-off is that the stats table now stores one daily presence row per active test case/environment/day instead of one aggregate-state row per project/environment/day.
- I did not change the `build_runs` or `test_runs` sort keys here. The current ordering appears tied to previous ReplacingMergeTree dedupe optimizations, so this PR focuses on caching, removing unnecessary `FINAL`, and adding missing predicates.
- The 10-second cache TTL is intentionally short to reduce polling pressure while keeping marketing stats and dashboards fresh enough for interactive use.
- The investigation also surfaced a recent-build chart query with 2,426 executions / 4.7 GiB read over 3 hours. I did not add caching to `Builds.list_build_runs/2` for that path because the code is page-load / query-param driven rather than a poller; a per-project 10-second cache would only help duplicate same-project overview loads inside a very small window while adding freshness ambiguity.


server/lib/tuist/tests.ex
```diff
@@ -132,11 +144,14 @@ defmodule Tuist.Tests do
   end
 
   def last_24h_test_run_count do
+    cached_count(:last_24h_test_run_count, &last_24h_test_run_count_query/0)
+  end
+
+  defp last_24h_test_run_count_query do
     twenty_four_hours_ago = DateTime.add(DateTime.utc_now(), -24, :hour)
 
     ClickHouseRepo.one(
```
pepicrft, line 153: Removing `FINAL` here changes the metric from logical test runs to raw ReplacingMergeTree row versions. `test_runs` gets new versions inserted for sharded runs when the aggregate status changes, when a run is marked flaky, and when stale in-progress runs are expired, so this will overcount any run that receives an update inside the 24h window. If we want to avoid `FINAL`, this still needs to dedupe by `id`, for example by grouping/counting distinct IDs over the recent window, otherwise the marketing stat regresses in correctness.

pepicrft (approved):
Looks good overall. I left one inline note to address before merging.

server/lib/tuist/tests.ex
fortmarek: Since this is only for the marketing page, we can live with some overcounting, rather than making the query less performant. Being fully correct here doesn't matter that much.

PR #10762, opened 2026-05-13
fix(kura): repair public load balancer exposure

pepicrft:
## Summary

- Switch Kura Hetzner LoadBalancer health checks to TCP for the TCP passthrough HTTPS and gRPC services.
- Allow public HTTPS and gRPC traffic in the Kura NetworkPolicy while keeping plain HTTP limited to in-cluster peers.
- Make workload cluster bootstrap set a cluster-specific external-dns TXT owner id so managed clusters do not fight over records under `tuist.dev`.

## Root cause

Kura pods and cluster-internal routing were healthy, but Hetzner marked the public LoadBalancer targets unhealthy because it was probing HTTP `/ready` against the TLS passthrough NodePort. The NetworkPolicy also did not allow external-source traffic to the public TLS ports when `externalTrafficPolicy: Local` preserved the client or load balancer source IP.

## Validation

- `mise exec go@1.25.10 -- go test ./...` from `infra/kura-controller`
- Verified the live production workaround returns `HTTP 200` from `https://tuist-eu-central-1.kura.tuist.dev/up` and Hetzner marks the Kura HTTPS and gRPC targets healthy.

### DIFF
--- a/infra/kura-controller/controllers/kurainstance_controller.go
+++ b/infra/kura-controller/controllers/kurainstance_controller.go
@@ -29,11 +29,11 @@ import (
 const (
 	KuraInstanceFinalizer = "kurainstances.kura.tuist.dev/finalizer"
 
-	httpPort          int32 = 4000
-	httpsPort         int32 = 443
-	httpsTargetPort   int32 = 4443
-	grpcPort          int32 = 50051
-	peerPort          int32 = 7443
+	httpPort        int32 = 4000
+	httpsPort       int32 = 443
+	httpsTargetPort int32 = 4443
+	grpcPort        int32 = 50051
+	peerPort        int32 = 7443
 
 	// drainCompletionTimeoutMs and preStopDelaySeconds together set how
 	// long a Kura pod is given to bleed connections off before SIGTERM.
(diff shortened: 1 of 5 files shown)


infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -476,15 +476,15 @@ func podTemplate(instance *kurav1alpha1.KuraInstance) corev1.PodTemplateSpec {
...
-				Ports: containerPorts(instance),
-				Env:            append(baseEnv(instance), instance.Spec.ExtraEnv...),
-				EnvFrom:        sharedSecretsEnvFrom(),
-				Resources:      defaultResources(),
-				VolumeMounts:   volumeMounts(instance),
-				Lifecycle:      preStopLifecycle(),
-				ReadinessProbe: httpProbe("/ready", 5, 10),
-				LivenessProbe:  httpProbe("/up", 20, 20),
-				StartupProbe:   httpProbe("/up", 0, 10),
+				Ports:           containerPorts(instance),
```
fortmarek, line 479: I see a bunch of formatting changes often landing in the `.go` files ... should we consider bringing in a formatter to make the changes more consistent?

fortmarek (approved):

infra/kura-controller/controllers/kurainstance_controller.go
pepicrft: Added a Kura controller format job that runs `gofmt -l .` and fails when files are not formatted.

PR #10763, opened 2026-05-13
feat(cli): add fast package resolution opt-in

pepicrft:
## Summary
- Adds an opt-in `TUIST_USE_FAST_PACKAGE_RESOLUTION` package installation path.
- Treats SwifterPM as Tuist's internal resolver implementation: release bundles invoke the `swifterpm` binary next to `tuist`, while development falls back to `mise x -- swifterpm`.
- Uses `Environment.current.isVariableTruthy` and `Environment.current.variables` at runtime, so tests can keep mocking the current environment without resolver-specific dependency injection.
- Lets Tuist load SwifterPM's persisted package-info JSON from `.build/swifterpm/package-info`, which skips recompiling package manifests when the cache is present.
- Pins SwifterPM `0.2.9` in `mise.toml` and `mise.lock`, and bundles, signs, and notarizes it next to `tuist` in `tuist.zip`.
- Restores the registry pins in `Package.resolved` and `server/native/xcresult_nif/Package.resolved` (revert of the earlier removal) now that SwifterPM `0.2.8` resolves Swift Package Registry pins natively, so Tuist keeps resolving `tuist.Command/FileSystem/Path`, `kolos65.Mockable`, `MobileNativeFoundation.XCLogParser`, `stephencelis.SQLite_swift`, and the `xcresult_nif` dependencies from the registry.
- Keeps the new SwifterPM cache-loader error type in its own Swift file and migrates the touched controller tests to Swift Testing.
- Declares `TuistSupportTests`' `TuistConstants` dependency explicitly so `tuist inspect dependencies --only implicit` stays green.

> [!NOTE]
> SwifterPM is intentionally a small standalone Rust executable for this first iteration instead of new resolver code inside Tuist. The resolver and cache logic is useful outside Tuist too, including as a library for the build system I am building, and keeping it behind an opt-in executable gives Tuist a narrow integration contract: invoke the tool, then consume SwiftPM-compatible state and package-info JSON. I do not expect this to be high maintenance because the Tuist side stays a thin adapter and the executable is versioned, signed, notarized, and tested separately. If that assumption turns out wrong or the integration starts creating friction, I am not opposed to rewriting it in Swift and moving it into this codebase.

## Submodules
- SwiftPM recursively initializes package submodules, so SwifterPM preserves that behavior instead of treating GitHub source archives as universally equivalent.
- SwifterPM uses source archives only for packages that do not declare `.gitmodules`.
- When a package declares `.gitmodules`, SwifterPM falls back to a git checkout and runs recursive submodule initialization, which fixes packages such as `swift-protobuf` without making every package pay the git-clone cost.

Measured after rebasing onto `origin/main` at `97add490a2` on macOS 26.4.1, Apple M5 Pro arm64. I used a temporary Git worktree of this branch and ran the freshly built Tuist binary directly with `TUIST_FRAMEWORK_SEARCH_PATHS` pointing at the build output, so the numbers do not include `swift run` launcher overhead. Each number is one `hyperfine --runs 1` measurement.

| Scenario | Normal Tuist install | `TUIST_USE_FAST_PACKAGE_RESOLUTION=1` | Improvement |
| --- | ---: | ---: | ---: |
| Cold | 82.961s | 34.092s | 2.43x faster, 58.9% less wall time |
| Worktree-warm | 28.268s | 2.592s | 10.90x faster, 90.8% less wall time |

Cold removes the package-local `.build` directory before each run. The fast cold run also removes the isolated `XDG_CACHE_HOME/swifterpm` cache. Worktree-warm removes only the package-local `.build` directory and keeps resolver-level caches warm, which models switching to another clean worktree.

## Resolver benchmark — full resolution from scratch (SwifterPM 0.2.9 vs SwiftPM)

Re-measured with SwifterPM `0.2.9` (release build; native PubGrub transitive resolution + native Swift Package Registry resolution, [tuist/swifterpm#2](https://github.com/tuist/swifterpm/pull/2) merged). `Package.resolved` is deleted first, so this exercises **full resolution, not lockfile restore** — the path @fortmarek's P1 flagged as incomplete on `0.2.7`. SwifterPM runs `resolve`; SwiftPM runs `swift package resolve`. Registry transformation is disabled on both sides so the lockfiles are directly comparable. Tuist fixtures are detached git worktrees; real-world fixtures are the pinned Pocket Casts iOS / Firefox iOS manifests. Apple M5 Pro, macOS 26.4.x, one `hyperfine --runs 1` measurement each. Cold wipes the resolver cache + scratch; warm keeps the cache primed.

| Fixture | Cold — SwifterPM / SwiftPM | Warm — SwifterPM / SwiftPM | Pins — SwifterPM / SwiftPM | SwiftPM accepts SwifterPM lockfile |
| --- | ---: | ---: | ---: | :--- |
| Firefox iOS | 8.54s / 14.33s — **1.68x** | 0.013s / 4.12s — **317x** | 6 / 6 — identical | ✅ |
| Pocket Casts iOS | 43.96s / 267.96s — **6.09x** | 3.32s / 40.33s — **12.1x** | 42 / 36 — superset¹ | ✅ |
| Tuist root manifest | 264.25s / 285.71s — **1.08x** | 0.75s / 68.70s — **91.6x** | 85 / 83 — superset¹ | ✅ |

¹ SwifterPM's set is a superset of SwiftPM's pruned set — it keeps transitive dependencies SwiftPM prunes by product/target reachability (e.g. the unused SQLCipher path `csqlite`/`sqlcipher.swift` on the Tuist manifest). SwiftPM accepts the superset under `--force-resolved-versions`.

> [!NOTE]
> The first run of this benchmark caught a real SwifterPM bug: fresh resolution of the Tuist manifest emitted **two** `difference` pins — `krzysztofzablocki/Difference` @ `1.1.0` (Tuist's direct dependency) and `krzysztofzablocki/Difference.git` @ branch `master` (pulled in by a transitive dependency) — because identity was not canonicalized across the `.git` suffix / URL case, so `swift package resolve --force-resolved-versions` rejected the lockfile as `duplicated entry for package "difference"`. Fixed in SwifterPM `0.2.9` ([`fix: dedupe resolved pins by canonical package identity`](https://github.com/tuist/swifterpm/commit/67e2c7b)): pins are collapsed to one per canonical identity, preferring the version-resolved entry SwiftPM keeps. The table above reflects `0.2.9` (resolver timings are unchanged — the dedupe is a constant-time post-processing step; correctness re-verified on the fixed binary). Covered by a unit regression test in `src/resolve.rs`.

## Resolver benchmark — lockfile restore (SwifterPM 0.2.9, `--force-resolved-versions`)

This is the path the fast integration actually uses: `SwiftPackageManagerController.swifterPMArguments` appends `--force-resolved-versions` **only when a `Package.resolved` exists**, so an authoritative lockfile is restored deterministically and fresh resolution is only the no-lockfile fallback. The restored lockfile is this PR's mixed registry+source-control state (`6 registry`, `83 remoteSourceControl`); the all-source-control lockfile is SwifterPM's own SCM resolution of the same manifest. Both run `swifterpm --force-resolved-versions resolve` on SwifterPM `0.2.9`. Cold wipes scratch + restore cache; warm wipes only scratch.

| Lockfile | Cold restore | Warm restore | Pins |
| --- | ---: | ---: | ---: |
| Mixed registry + source-control (restored) | 59.32s | 0.10s | 89 |
| All source-control | 30.05s | 0.10s | 86 |

Restoring registry pins via SwifterPM's native registry download is ~2x slower cold than git source-control restore (archive fetch + extract vs. cached git), and indistinguishable warm (~0.1s). Both restore deterministically from the authoritative lockfile, which is the contract the integration relies on.

## Validation
- `hyperfine --runs 1` comparing `tuist install --path <temporary-tuist-worktree>` with and without `TUIST_USE_FAST_PACKAGE_RESOLUTION=1` for cold and worktree-warm resolution.
- Resolver benchmarks above: SwifterPM `0.2.9` vs `swift package resolve`, full resolution from scratch and `--force-resolved-versions` restore, against detached Tuist worktrees and the pinned Pocket Casts iOS / Firefox iOS manifests. `swift package resolve --force-resolved-versions` accepts SwifterPM's fresh lockfile for all three fixtures.
- `cargo test` in SwifterPM is green, including the new `dedupe_pins_by_identity` regression tests.
- `mise run cli:lint --fix`
- `git diff --check`
- `jq empty Package.resolved server/native/xcresult_nif/Package.resolved`
- `Package.resolved` and `server/native/xcresult_nif/Package.resolved` restore their registry pins (`6`/`11` `"kind" : "registry"` entries) and `mise x -- swifterpm --force-resolved-versions resolve` materializes them from the registry.
- `swift package dump-package --package-path . --disable-scm-to-registry-transformation`
- `tuist inspect dependencies --only implicit`
- `mise x github:tuist/swifterpm@0.2.9 -- swifterpm --version` prints `swifterpm 0.2.9`; `mise install` validates the `0.2.9` checksums for every platform slot.
- `mise x -- swifterpm --package-path . --cache-path /tmp/swifterpm-tuist-cache --scratch-path /tmp/swifterpm-tuist-swifterpm-scratch resolve --force-resolved-versions --quiet`
- `swift build --package-path . --product tuist --disable-scm-to-registry-transformation --scratch-path /tmp/swifterpm-tuist-clean-test-scratch --skip-update`
- `TUIST_USE_FAST_PACKAGE_RESOLUTION=1 /tmp/swifterpm-tuist-clean-test-scratch/arm64-apple-macosx/debug/tuist inspect dependencies --only implicit`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistSupportTests/SwiftPackageManagerControllerTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`

## SwifterPM release
- Released `0.2.9`: https://github.com/tuist/swifterpm/releases/tag/0.2.9
- Native PubGrub transitive resolution and native Swift Package Registry resolution landed in [tuist/swifterpm#2](https://github.com/tuist/swifterpm/pull/2) ("Resolve packages natively"), shipped in `0.2.8`.
- `0.2.9` fixes the package-identity canonicalization bug surfaced by the resolver benchmark (duplicate `difference` pin → SwiftPM rejection).
- The release includes macOS, Linux, Windows archives plus `SHA256.txt` and `SHA512.txt`.
- The release is signed and notarized for macOS through the release pipeline (the `0.2.8` pipeline also fixed the `aarch64-apple-darwin` build by installing Rust via `dtolnay/rust-toolchain`).


fortmarek:
I think we should frame this as a P1 around the intended contract for swifterpm rather than only the auto-added flag:

**P1: Treat swifterpm as lockfile restore, not full resolution**

The fast path is only correct when `Package.resolved` already exists and is authoritative. In swifterpm 0.2.7, the unlocked `resolve` path only inspects the root manifest’s direct source-control dependencies and does not compute SwiftPM’s full transitive graph. If `Package.resolved` is missing or outdated, `tuist install` can generate an incomplete lockfile/workspace state/package-info cache. We should either fall back to `swift package resolve` when the lockfile is missing/outdated, or make the fast path explicitly require locked resolution from an existing `Package.resolved`.

Related suggestion: if we intentionally constrain swifterpm to “materialize an existing `Package.resolved`” rather than “perform SwiftPM-equivalent resolution,” adding Tuist Registry support should be much more tractable. It would mostly mean supporting `kind: "registry"` pins from the lockfile: read the registry config, download the exact release archive, extract it to SwiftPM-compatible `.build/registry/downloads/...`, emit matching `workspace-state.json` entries, and include those packages in the package-info cache. The hard version is supporting missing lockfiles or `--replace-scm-with-registry`, because that requires identifier lookup, release queries, constraint solving, and writing correct registry pins.

pepicrft:
@fortmarek This P1 was exactly right for `0.2.7`. Since you wrote it, swifterpm [#2 "Resolve packages natively"](https://github.com/tuist/swifterpm/pull/2) merged (shipped in `0.2.8`), I extended the PR's benchmarks to verify the contract you described, that run **caught a real swifterpm bug**, and the fix shipped in `0.2.9` — which this branch now pins. Walking through both halves of your comment:

### 1. "Lockfile restore, not full resolution"

The architectural concern is fixed. `0.2.8`+ no longer "only inspects the root manifest's direct source-control dependencies": the unlocked `resolve` path is a real PubGrub solve over the complete transitive graph (`pubgrub::resolve` in `src/resolve.rs`, fed by a parallel BFS that recursively discovers transitives). The new **"full resolution from scratch"** table deletes `Package.resolved` and measures it computing the whole graph:

| Fixture | SwiftPM accepts swifterpm's fresh lockfile | Speed vs `swift package resolve` |
| --- | --- | --- |
| Firefox iOS | ✅ identical 6-pin set | 1.68x cold / 317x warm |
| Pocket Casts iOS | ✅ | 6.09x cold / 12.1x warm |
| Tuist root manifest | ✅ (after the `0.2.9` fix below) | 1.08x cold / 91.6x warm |

Extending the benchmark to Tuist's own manifest is what exposed the bug: fresh resolution emitted **two** `difference` pins — `krzysztofzablocki/Difference` @ `1.1.0` (Tuist's direct dep) and `krzysztofzablocki/Difference.git` @ `master` (a transitive dep) — because identity wasn't canonicalized across the `.git`/URL-case difference, so `swift package resolve --force-resolved-versions` rejected the lockfile as `duplicated entry for package "difference"`. Fixed in swifterpm `0.2.9` ([`fix: dedupe resolved pins by canonical package identity`](https://github.com/tuist/swifterpm/commit/67e2c7b), with a regression test): pins collapse to one per canonical identity, preferring the version-resolved entry SwiftPM keeps. Tuist now resolves to a single `difference` @ `1.1.0` and SwiftPM accepts the lockfile. (The remaining 85-vs-83 pin gap is the unused SQLCipher path SwiftPM prunes by reachability — a harmless superset SwiftPM still accepts, same as Pocket Casts.)

Your conclusion holds and the integration already encodes it: `SwiftPackageManagerController.swifterPMArguments` appends `--force-resolved-versions` **only when a `Package.resolved` exists**, so authoritative lockfile restore is the primary path (the "lockfile restore" table: warm ~0.1s, registry pins materialized natively) and fresh resolution is the no-lockfile fallback. I'd still like to add the explicit guard you suggested — when `Package.resolved` is missing **or** swifterpm's output is inconsistent with the manifest, fall back to `swift package resolve` — as defense-in-depth so the contract is enforced rather than implicit. Happy to wire that into the controller in this PR if you agree.

### 2. Registry support

`0.2.8`+ implements precisely the "tractable" version you outlined: `src/registry.rs` reads `.swiftpm/configuration/registries.json` and queries releases; `download_registry_archive` / `restore.rs::ensure_registry_source` fetch and extract the exact release archive; `resolve.rs` emits `kind: "registry"` pins and `restore.rs` materializes them. This branch's `revert: restore registry pins from package locks` puts Tuist back on the registry for `tuist.Command/FileSystem/Path`, `kolos65.Mockable`, `MobileNativeFoundation.XCLogParser`, `stephencelis.SQLite_swift` (root) plus the `xcresult_nif` deps, restored through the registry path (benchmarked in the "lockfile restore" table). The boundary stays honest about your "hard version": swifterpm does **not** do SCM→registry transformation during a *fresh* resolution, so registry usage comes via authoritative `kind: "registry"` pins in the lockfile — exactly the framing you proposed.

---

Separately: the Blick P1 about `import XCTest` in the migrated Swift Testing controller test was already addressed by a later commit on this PR — both touched test files (`SwiftPackageManagerControllerTests`, `SwiftPackageManagerGraphLoaderTests`) are 100% Swift Testing now (`@Test`/`#expect`, zero `XCTAssert`). Resolving that thread.

cli/Tests/TuistSupportTests/SwiftPackageManager/SwiftPackageManagerControllerTests.swift
```diff
@@ -1,33 +1,36 @@
-import Command
-import TSCUtility
-import TuistCore
+import FileSystem
```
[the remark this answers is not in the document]
pepicrft: Already addressed by a later commit on this PR: `SwiftPackageManagerControllerTests.swift` (and `SwiftPackageManagerGraphLoaderTests.swift`) are now 100% Swift Testing — no `import XCTest`, no `XCTAssert*`; `package_registry_login`/`package_registry_logout` use `#expect(commandRunner.called(command))`. Verified: `grep -c "import XCTest|XCTAssert"` → 0 in both files. Resolving.

cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
```diff
@@ -121,25 +157,32 @@ public struct SwiftPackageManagerGraphLoader: SwiftPackageManagerGraphLoading {
...
+        )
 
-        let rootPackage = try await manifestLoader.loadPackage(at: packagePath.parentDirectory, disableSandbox: disableSandbox)
-
-        var packageInfos: [
-            // swiftlint:disable:next large_tuple
-            (
-                id: String,
-                name: String,
-                folder: AbsolutePath,
```
fortmarek, line 136: When TUIST_USE_FAST_PACKAGE_RESOLUTION is set, this trusts the persisted package-info JSON whenever index.json exists. The only freshness check added here compares Package.resolved, so edits to Package.swift that do not rewrite the lockfile, such as target/product/trait changes or local package manifest edits, reuse stale PackageInfo and generate an old graph. The previous path reloaded manifests on every load. Please include manifest/package-info invalidation, or fall back to ManifestLoader for root/local packages.

cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift
```diff
@@ -305,18 +345,88 @@ public struct SwiftPackageManagerGraphLoader: SwiftPackageManagerGraphLoading {
...
+            }
+    }
+
+    private static func identity(for packageInfo: SwiftPackageInfoEntry) -> String {
+        if packageInfo.kind == "registry" {
+            // A package is uniquely identified by a scoped identifier in the form scope.package-name.
+            return String(packageInfo.name.split(separator: ".").last ?? "").lowercased()
+        } else {
+            return packageInfo.name.lowercased()
+        }
```
fortmarek, line 425: The deduper assumes a registry identity can be matched to its SCM counterpart by taking text after the last dot. This PR restores a current lockfile pair stephencelis.SQLite_swift and sqlite.swift; these normalize to sqlite_swift vs sqlite.swift, so both survive deduplication and Tuist maps two external SQLite packages with duplicate products/targets. The canonicalization needs to handle registry escape rules or compare against a resolver-provided canonical identity before building the package dictionaries.

.github/workflows/cli.yml
```diff
@@ -140,11 +135,11 @@ jobs:
       - name: Setup Tuist
         if: github.event.pull_request.head.repo.fork != true
         run: tuist setup
-      - name: Install Tuist dependencies
+      - name: Resolve SwiftPM dependencies
         if: steps.cache.outputs.cache-hit != 'true'
-        run: tuist install
+        run: swift package resolve --disable-scm-to-registry-transformation
```
fortmarek, line 140: I assume this is not using swifterpm then?

.github/workflows/release.yml
```diff
@@ -86,11 +86,11 @@ jobs:
           LATEST_VERSION=$(git tag -l | grep -E '^[0-9]+\.[0-9]+\.[0-9]+$' | sort -V | tail -n1 || echo "0.0.0")
 
           # Check if there are any releasable changes by generating changelog
-          CHANGELOG=$(git cliff --include-path "cli/**/*" --include-path ".xcode-version-releases" --include-path "Package.swift" --include-path "Package.resolved" --config cliff.toml --repository "../" 2>/dev/null -- ${LATEST_VERSION}..HEAD || echo "")
+          CHANGELOG=$(git cliff --include-path "cli/**/*" --include-path "mise/tasks/cli/**/*" --include-path "mise.toml" --include-path ".xcode-version-releases" --include-path "Package.swift" --include-path "Package.resolved" --config cliff.toml --repository "../" 2>/dev/null -- ${LATEST_VERSION}..HEAD || echo "")
```
fortmarek, line 89: why are we releasing the cli when `mise.toml` changes? That doesn't feel right.

.github/workflows/release.yml
```diff
@@ -916,7 +916,7 @@ jobs:
       - uses: jdx/mise-action@v4.0.1
         with:
           version: ${{ env.MISE_VERSION }}
-          install_args: "tuist git-cliff 1password-cli"
+          install_args: "tuist git-cliff 1password-cli github:tuist/swifterpm"
```
fortmarek, line 919: why are we installing `swifterpm` through mise?

cli/Sources/TuistConstants/Constants.swift
```diff
@@ -71,6 +71,7 @@ public enum Constants {
         public static let token = "TUIST_TOKEN"
         public static let deprecatedToken = "TUIST_CONFIG_TOKEN"
         public static let quiet = "TUIST_CONFIG_QUIET"
+        public static let useFastPackageResolution = "TUIST_USE_FAST_PACKAGE_RESOLUTION"
```
fortmarek, line 74: `TUIST_SWIFTERPM` or `TUIST_USE_SWIFTERPM` sounds a bit better if we are to popularize the new CLI itself

cli/Sources/TuistLoader/Loaders/SwifterPMPackageInfoCacheLoader.swift
```diff
@@ -0,0 +1,102 @@
...
+import Foundation
+import Path
+import TuistConstants
+import TuistEnvironment
+import XcodeGraph
+
+struct SwifterPMPackageInfoCacheLoader {
+    private let fileSystem: FileSysteming
+
+    init(fileSystem: FileSysteming) {
```
fortmarek, line 11:
```suggestion
    init(fileSystem: FileSysteming = FileSystem()) {
```

cli/Sources/TuistSupport/SwiftPackageManager/SwiftPackageManagerController.swift
```diff
@@ -206,4 +229,63 @@ public struct SwiftPackageManagerController: SwiftPackageManagerControlling {
...
+
+    private func swifterPMCommandPrefix() -> [String] {
+        if let currentExecutablePath = Environment.current.currentExecutablePath() {
+            let bundledPath = currentExecutablePath.parentDirectory.appending(component: "swifterpm")
+            if FileManager.default.isExecutableFile(atPath: bundledPath.pathString) {
+                return [bundledPath.pathString]
+            }
+        }
+
+        return ["mise", "x", "--", "swifterpm"]
```
fortmarek, line 262: requiring mise here doesn't feel right.

fortmarek (commented):
Why did we go with Rust for SwifterPM? Do we actually benefit from using Rust vs using Swift that we can directly integrate as a library instead of vendoring it as a CLI? Fwiw, I don't fully think using Rust is the right choice for something like SwifterPM if the goal is to make it a popular tool in the Swift ecosystem.

Additionally, while this experiment is impressive, I have to wonder if investing into the mise cache would have offered the same improvement for worktrees without having to align with whatever SwiftPM is doing and binding us to maintain this layer for a long time to come, and thus a better investment for Tuist itself. Especially since the primary improvements come from the warm cache.

PR #10764, opened 2026-05-13
fix(noora): pin aube version in storybook Dockerfile

asmitbm:
The Dockerfile installed @endevco/aube without a version pin, so each CI run picked up whatever was latest on npm. Recent aube releases (1.10+) execute esbuild's bin shim through Node, which fails with "SyntaxError: Invalid or unexpected token" because esbuild's postinstall replaces the shim with the platform-native binary.

Pin to 1.6.2 to match the version pinned in mise.lock for local dev.


fortmarek (approved):
Thanks!

PR #10769, opened 2026-05-13
fix(server): expose Kura controls to ops

pepicrft:
## Summary

- Allow ops users to access settings for any account.
- Gate the Kura account settings section behind ops access only.
- Show undeployed Kura regions in the servers table with per-region deploy actions.

## Context

The Kura settings page only showed deployed servers, so there was no direct way to create a server in another available region once one region already existed. The section is also intended to stay ops-only until it is exposed to regular users later.

## Testing

- `git diff --check`

Could not complete the focused LiveView tests locally because the test database setup is blocked by duplicate migration version `20260512120000`. `mix deps.get --locked` also rewrites unrelated `cowboy` and `cowlib` lock entries in this worktree, so I left `server/mix.lock` unchanged.

### DIFF
--- a/server/lib/tuist/authorization.ex
+++ b/server/lib/tuist/authorization.ex
@@ -160,8 +160,8 @@ defmodule Tuist.Authorization do
       desc("Allows the admin of an account to update its settings.")
       allow([:authenticated_as_user, user_role: :admin])
 
-      desc("Allows users with ops access to update any account's settings.")
-      allow([:authenticated_as_user, :ops_access])
+      desc("Allows users with ops write access to update any account's settings.")
+      allow([:authenticated_as_user, :ops_write_access])
     end
 
     action :delete do
--- a/server/lib/tuist/authorization.ex
+++ b/server/lib/tuist/authorization.ex
@@ -159,6 +159,9 @@ defmodule Tuist.Authorization do
     action :update do
       desc("Allows the admin of an account to update its settings.")
       allow([:authenticated_as_user, user_role: :admin])
+
+      desc("Allows users with ops access to update any account's settings.")
+      allow([:authenticated_as_user, :ops_access])
     end
 
     action :delete do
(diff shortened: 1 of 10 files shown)


server/lib/tuist/authorization.ex
```diff
@@ -159,6 +159,9 @@ defmodule Tuist.Authorization do
     action :update do
       desc("Allows the admin of an account to update its settings.")
       allow([:authenticated_as_user, user_role: :admin])
+
+      desc("Allows users with ops access to update any account's settings.")
+      allow([:authenticated_as_user, :ops_access])
```
fortmarek: we can allow this blanket permission for now, but long-term, we really need to default to read-only access to the user's dashboard with an admin/write access only allowed after explicitly asking for a permission (ideally, with an included reason).

This is fine as long as we are the only ones accessing, but since `/ops` is accessed by anyone at Tuist, the company, this is not great long-term.

server/lib/tuist_web/live/account_settings_live.ex
```diff
@@ -328,8 +327,8 @@ defmodule TuistWeb.AccountSettingsLive do
     |> assign(:add_kura_server_form, default_kura_server_form(available_regions))
   end
 
-  defp kura_enabled?(account) do
-    Environment.dev?() or FunWithFlags.enabled?(:kura, for: account)
+  defp kura_enabled?(current_user) do
+    Authorization.authorize(:ops_read, current_user) == :ok
   end
```
fortmarek, line 333: I think this should probably go via a feature flag and be gated per org instead of gated by `/ops` access

fortmarek (approved):

server/priv/marketing/changelog/2026.05.13-kura-cache-server-regions.md
```diff
@@ -0,0 +1,7 @@
+---
+title: Manage Kura cache server regions
+category: "Product"
+pull_request: 10769
+---
+
+Ops users can now manage Kura cache servers from account settings across every account. The Kura section shows deployed servers alongside regions that are not deployed yet, with a direct deploy action for each available region so expanding Kura coverage no longer requires leaving the settings page.
```
fortmarek: @pepicrft This is an example of something that definitely should _not_ go into the product changelog.

PR #10771, opened 2026-05-13
fix(cli): ignore embeddable watch apps in redundant dependency inspection

shgew:
This fixes a false positive in `tuist inspect dependencies --only redundant` for modern watchOS apps represented as watchOS `.app` targets.

An iOS app target can intentionally depend on an embeddable watch application so the generated project includes the watch embed relationship. The redundant dependency linter already ignored extension-like products and legacy `.watch2App` dependencies, but it still reported modern embeddable watch apps as redundant. This change reuses `Target.isEmbeddableWatchApplication()` so inspection matches the generator's watch app embedding semantics.

One existing reproduction is `examples/xcode/generated_ios_app_with_watch_application`, where `App` depends on `WatchApp` and `WatchApp` is declared as:

```swift
.target(
    name: "WatchApp",
    destinations: [.appleWatch],
    product: .app,
    ...
)
```

On current `main`, this command fails:

```console
$ tuist inspect dependencies --path examples/xcode/generated_ios_app_with_watch_application --only redundant
The following redundant dependencies were found:
 - App redundantly depends on: FrameworkA, WatchApp
 - WatchApp redundantly depends on: FrameworkA
 - WatchWidgetExtension redundantly depends on: FrameworkA
```

The `FrameworkA` entries are separate findings from that example. The false positive fixed here is `App redundantly depends on: WatchApp`, because that dependency is required for embedding the watch application.

### How to test locally

Run `TuistUnitTests`.

### DIFF
--- a/cli/Sources/TuistInspectCommand/Services/GraphImportsLinter.swift
+++ b/cli/Sources/TuistInspectCommand/Services/GraphImportsLinter.swift
@@ -116,6 +116,10 @@
                 .filter { dependency in
                     switch target.target.product {
                     case .app, .watch2AppContainer:
+                        if dependency.target.isEmbeddableWatchApplication() {
+                            return false
+                        }
+
                         switch dependency.target.product {
                         case .appExtension, .stickerPackExtension, .messagesExtension, .extensionKitExtension, .watch2App:
                             return false
--- a/cli/Sources/TuistInspectCommand/Services/GraphImportsLinter.swift
+++ b/cli/Sources/TuistInspectCommand/Services/GraphImportsLinter.swift
@@ -75,7 +75,8 @@
                     graphTraverser: graphTraverser,
                     target: target,
                     includeExternalDependencies: inspectType == .implicit,
-                    excludeAppDependenciesForTests: inspectType == .redundant
+                    excludeAppDependenciesForTests: inspectType == .redundant,
+                    excludeEmbeddableWatchApps: inspectType == .redundant
                 )
 
                 let observedImports = switch inspectType {
@@ -96,7 +97,8 @@
             graphTraverser: GraphTraverser,
             target: GraphTarget,
             includeExternalDependencies: Bool,
-            excludeAppDependenciesForTests: Bool
+            excludeAppDependenciesForTests: Bool,
+            excludeEmbeddableWatchApps: Bool
         ) -> Set<String> {
             let targetDependencies = if includeExternalDependencies {
                 graphTraverser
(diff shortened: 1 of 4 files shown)


pepicrft (approved):
Looks good overall. The redundant dependency fix is heading in the right direction, and the focused local test run plus format check passed here. I left one follow-up comment about implicit mode and `watch2AppContainer` coverage.

cli/Sources/TuistInspectCommand/Services/GraphImportsLinter.swift
```diff
@@ -116,6 +116,10 @@
                 .filter { dependency in
                     switch target.target.product {
                     case .app, .watch2AppContainer:
+                        if dependency.target.isEmbeddableWatchApplication() {
```
pepicrft: One thing to double check here: this helper is shared by both `.redundant` and `.implicit`, so this exclusion also changes implicit inspection. With the current `canImport` handling, a host app that does `#if canImport(WatchApp) import WatchApp #endif` can now be reported as implicitly importing `WatchApp` even when the dependency is declared. I think we either want to scope this to redundant mode only or add an implicit regression that proves this stays quiet.

cli/Tests/TuistKitTests/Services/Inspect/InspectDependenciesCommandServiceTests.swift
```diff
@@ -602,6 +602,36 @@ struct InspectDependenciesCommandServiceTests {
         try await subject.run(path: path.pathString, inspectionTypes: [.redundant])
     }
 
+    @Test
+    func runRedundantOnlyDoesntFlagModernWatchAppDependencies() async throws {
```
pepicrft, line 606: Can we cover the `watch2AppContainer` host path too? The production change applies to both `.app` and `.watch2AppContainer`, but this regression only exercises the `.app` host case.

pepicrft (approved):
Looks good overall. The focused local test run and format check passed here. I left two non-blocking follow-ups inline.

shgew:
@pepicrft Thank you for the feedback!
Addressed in https://github.com/tuist/tuist/pull/10771/changes/ed3408af70e546a878eb3bba9fa7bda73880800f.

pepicrft:
Thanks @shgew. Let's merge once CI is green.

shgew:
@pepicrft looks like something went wrong on CI, but I can't understand if it's my changes or just something's flaky. Can you please advise if I should fix anything?

pepicrft:
> @pepicrft looks like something went wrong on CI, but I can't understand if it's my changes or just something's flaky. Can you please advise if I should fix anything?

It's likely flakiness. I'm looking into it myself and I'll report once I have it figured out.

PR #10777, opened 2026-05-13
fix(infra): restore CAPI core self-heal in server deploy

fortmarek:
### Summary

The runners PR (#10653) folded `clusterctl init` into `infra/mise/tasks/k8s/bootstrap-workload.sh` step 8 on the assumption every workload cluster had been re-bootstrapped with the new script. The canary cluster hadn't, so post-merge deploys failed with:

```
UPGRADE FAILED: resource mapping not found for name: "tuist-tuist-capi" namespace: "tuist-canary": no matches for kind "Cluster" in version "cluster.x-k8s.io/v1beta1"
```

when the chart tried to render the new `infra/helm/tuist/templates/capi-cluster.yaml` (`kind: Cluster` from CAPI core).

This restores the deploy-time `clusterctl init` self-heal step in `server-deployment.yml`. It is idempotent (no-op once CAPI core is present at the pinned version) and covers any workload cluster bootstrapped before the new step in `bootstrap-workload.sh` existed.

Failing run: https://github.com/tuist/tuist/actions/runs/25803781264/job/75804969011

### How to test locally

No straightforward local test path: this is a workflow-only change that runs `clusterctl init --core cluster-api:v1.10.4` against a workload cluster. The CI run on this PR exercises the new step against canary, which is the cluster that was missing CAPI core.

### DIFF
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -323,6 +323,51 @@ jobs:
               echo "Release $HELM_RELEASE_NAME status is ${status:-not-installed}; no recovery needed."
               ;;
           esac
+      - name: Ensure CAPI core in workload cluster
+        # The chart renders nested CAPI CRs (Cluster + MachineDeployment +
+        # Machine) into the workload cluster for Mac mini fleet management
+        # (see infra/helm/tuist/templates/capi-cluster.yaml). CAPI core
+        # itself ships outside the chart (the ~7000-line components
+        # manifest doesn't carry as a subchart) and is installed once
+        # per workload cluster by infra/mise/tasks/k8s/bootstrap-workload.sh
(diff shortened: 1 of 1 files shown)


fortmarek:
Closing — installed CAPI core (`cluster-api:v1.10.4`) manually on canary and production instead.

Ran on both clusters:

```
clusterctl init --core cluster-api:v1.10.4
kubectl -n capi-system wait --for=condition=Available deploy/capi-controller-manager --timeout=10m
```

`capi-controller-manager` reached `Available` and the `clusters.cluster.x-k8s.io` / `machinedeployments.cluster.x-k8s.io` CRDs are registered, so the chart's `capi-cluster.yaml` template should now apply cleanly. Re-adding the deploy-time self-heal step isn't needed — bootstrap-workload.sh step 8 covers any future cluster.

PR #10793, opened 2026-05-14
ci(cli, server, app): route macOS jobs to managed Tuist runner fleet

fortmarek:
Migrate every macOS pipeline in the repo — both Namespace Labs profiles and GitHub-hosted `macos-26` — to the Tuist-managed warm pool (`runs-on: tuist-macos`), backed by the Mac mini fleet provisioned through the runners-controller / runner-image / tart-kubelet stack.

## Jobs moved

| Workflow | Job | From → To |
| -- | -- | -- |
| `cli.yml` | Lint | `macos-26` → `tuist-macos` |
| `cli.yml` | SwiftPM Build | `macos-26` → `tuist-macos` |
| `cli.yml` | Cache | `macos-26` → `tuist-macos` |
| `cli.yml` | Unit Tests | `namespace-profile-macos-xcode-26` → `tuist-macos` |
| `cli.yml` | Build Acceptance Tests | `namespace-profile-macos-xcode-26` → `tuist-macos` |
| `cli.yml` | Acceptance Tests | `namespace-profile-macos-xcode-26` → `tuist-macos` |
| `cli.yml` | Acceptance Tests (Fork) | `macos-26` → `tuist-macos` |
| `server.yml` | Swift Tests (XCActivityLogParser) | `namespace-profile-macos-xcode-26` → `tuist-macos` |
| `server-production-deployment.yml` | Acceptance Tests | `namespace-profile-macos-xcode-26` → `tuist-macos` |
| `release.yml` | release-tuist (macOS variant) | `namespace-profile-default-macos` → `tuist-macos` |
| `release.yml` | release-tuist tests | `namespace-profile-macos-xcode-26` → `tuist-macos` |
| `release.yml` | release-app | `macos-26` → `tuist-macos` |
| `release.yml` | release-kura-binary (Darwin x86_64) | `namespace-profile-macos-xcode-26` → `tuist-macos` |
| `release.yml` | release-kura-binary (Darwin aarch64) | `namespace-profile-macos-xcode-26` → `tuist-macos` |
| `app.yml` | Build | `macos-26` → `tuist-macos` |
| `app.yml` | Test | `macos-26` → `tuist-macos` |
| `app.yml` | Device Build | `macos-26` → `tuist-macos` |

Linux jobs on `namespace-profile-default-with-volume` are unaffected — they continue to use Namespace for the BuildKit-backed Docker image builds.

## Dropped Namespace-specific actions

[...]

PR #10798, opened 2026-05-14
fix(infra): repair runner-fleet bootstrap regressions

fortmarek:
> Three independent regressions in the CAPI Scaleway provider and macOS host bootstrap kept the canary/production runner fleet from coming up healthy. Fix them together because they were all introduced in the same window and they all live on the bootstrap path.

## Symptoms

When validating the runner fleet roll on PR #10793, every Pod on every Mac mini in production was stuck in `TartCreateFailed` with:

```
tart run: ensure gui session: Aqua session for uid 501 did not come up within 30s after SIGHUP loginwindow
```

…even after a manual reboot of the host. CAPI also reported 9 Machine CRs in the runner MachineDeployment but only 5 unique Scaleway servers backed them; the remaining 4 were duplicates and the operator-pre-ordered `tuist-pool-*` hosts sat unclaimed. The controller was looping on bootstrap retries (192 attempts in 4h55m on a single Machine).

Root-causing surfaced three independent bugs.

## Bug 1 — Empty kcpassword on adopted hosts

**Root cause.** Scaleway's Apple Silicon API only populates the `sudo_password` field on the **CreateServer** response. Every subsequent GET/list returns it empty. Hosts adopted from the pool prefix never go through CreateServer, so [`scalewayServerToServer`](infra/cluster-api-provider-scaleway-applesilicon/internal/scaleway/client.go) staged an empty password into the bootstrap Secret. Downstream, [`enableAutoLogin`](infra/macos-host-bootstrap/bootstrap.go) either silently skipped (post-#10653) or wrote a kcpassword whose plaintext was just the 11-byte cipher key padding (older controller image). Either way `loginwindow` rejected the auto-login, Aqua never came up on boot, and `tart run` failed forever.

**Fix.** Scaleway's `vnc_url` field embeds the same OS-default credentials as `vnc://<ssh_username>:<password>@<host>:<port>` and IS surfaced on every GET. `passwordFromVncURL` parses the password out of that URL (handling percent-encoded special characters), and `scalewayServerToServer` uses it as a fallback when `SudoPassword` is empty. Verified out-of-band that the VNC password authenticates the m1 macOS user.

Defense-in-depth: the controller refuses to bootstrap a Machine whose bootstrap Secret has no sudo password. Machines that already hit the failure mode have `Status.ServerID` set and an existing bootstrap Secret with an empty sudo-password, so this reconcile pass would skip Stage 1 (where the vnc_url fallback runs) and loop on `MissingSudoPassword` forever. Before refusing, the controller now tries a fresh `ScalewayClient.GetServer`; if it returns a non-empty `SudoPassword` (from the vnc_url fallback), the Secret is updated in place and bootstrap resumes. `MissingSudoPassword` is surfaced only when even the refresh comes back empty.

## Bug 2 — Non-idempotent pf install

**Root cause.** [`installVMEgressFirewall`](infra/macos-host-bootstrap/bootstrap.go) validates `/etc/pf.conf` with `pfctl -nf` before loading. The anchor defines two `persist` tables (`vm_sources`, `blocked_dst`); once loaded they survive subsequent `pfctl -f` invocations. A re-run on an already-bootstrapped host fails the validate step with `cannot define table vm_sources: Resource busy`, the whole bootstrap step errors, and the controller retries forever.

**Fix.** Flush the anchor's state before validating:

```bash
sudo pfctl -a tuist.runners -F all 2>/dev/null || true
```

Scoped to the named anchor so the system's main ruleset stays untouched. First-run hosts (where the anchor doesn't exist yet) tolerate the non-zero exit via `|| true`.

## Bug 3 — Duplicate adoption of the same pool host

**Root cause.** [`AdoptByPrefix`](infra/cluster-api-provider-scaleway-applesilicon/internal/scaleway/client.go) assumed that when two reconciles race the same pool host, Scaleway would return a name conflict on the losing UpdateServer call. **It does not** — every UpdateServer with a *different* target name succeeds. Both reconciles walked away with the same ServerID. Production state: 3 Machine CRs pointing at one Scaleway server, plus 2 unclaimed pool hosts left over because the duplicates were burning the controller's reconciliation budget.

**Fix — serialize adoption with a process-local mutex.**

Scaleway's `UpdateServer` is unconditional (no CAS / etag), so optimistic read-after-write verification alone can't establish "we own this server" — two interleaved phase-1+verify rounds can each observe their own pending marker and both proceed to phase 2, leaving both callers with the same ServerID. The controller is single-replica via leader election, so a `sync.Mutex` on `Client` keyed at `AdoptByPrefix` entry is sufficient: only the leader runs, and the leader doesn't race itself once adoption is serialized.

A two-phase rename runs *inside* the mutex as defense-in-depth against external mutation (operator console rename mid-adoption, or a future multi-replica deployment that loses leader election between reconciles):

1. **Phase 1.** Rename the candidate `tuist-pool-001` → `tuist-claim-pending-<uuid>`. GetServer the candidate; verify `Name` reflects OUR marker.
2. **Phase 2.** Rename the marker to the final `claimName`.

A crash between phases leaves the host outside any pool prefix as a `tuist-claim-pending-*` orphan. The adoption scan matches that prefix too, so the next reconcile re-adopts it through the same two-phase dance.

### Scaleway error handling

The previous code swallowed every `UpdateServer` / `GetServer` failure in phase 1 / verify as "race lost." That meant a 403 (auth), 5xx (outage), or validation error across all candidates was silently downgraded to `ErrNoAvailableHost` and the operator was told to pre-order capacity they already had. Now only `scw.ResourceNotFoundError` is treated as race-lost; everything else surfaces as `ScalewayAdoptFailed`.

## Tests

A small `AppleSiliconAPI` interface in `internal/scaleway/client.go` is the seam between `Client` and the SDK's concrete `*applesilicon.API`. Production code is unaffected — the SDK type satisfies the interface via Go's structural typing. Tests drive an in-memory fake that simulates Scaleway's rename-store semantics, with hooks for injecting concurrent renames and forcing per-call errors on `UpdateServer` / `GetServer`.

Coverage in `internal/scaleway/client_test.go`:

- `passwordFromVncURL` — empty input, alphanumeric password, percent-encoded special chars, no userinfo, user-only, malformed URL.
- `scalewayServerToServer` — VNC fallback when `SudoPassword` is empty, primary `SudoPassword` wins when both sources are set, both-empty stays empty.
- `AdoptByPrefix`:
  - Happy path (single pool host, both phases run, server renamed to claimName).
  - Idempotent rediscovery (host already named claimName from a prior reconcile; zero UpdateServer calls).
  - Race-lost via external rename overwriting the pending marker between phase 1 and verify; AdoptByPrefix skips the candidate and picks another.
  - Orphan re-adoption (stale `tuist-claim-pending-*` host from a crashed reconcile is treated as eligible and re-adopted via the same two-phase path).
  - Eligibility filters (wrong prefix / not delivered / not ready / wrong type / wrong OS → ErrNoAvailableHost, zero UpdateServer calls).
  - Phase 1 UpdateServer **404** (server deleted under us) — skips candidate, falls through to next.
  - Phase 1 UpdateServer **non-404** (auth / outage / validation) — surfaces as a real error, NOT ErrNoAvailableHost.
  - Phase 1 verify GetServer **404** — skips candidate.
  - Phase 1 verify GetServer **non-404** — surfaces as a real error.
  - Phase-2 promotion failure (error surfaced, server left in claim-pending state so the next reconcile re-adopts it).
  - Empty `poolPrefix` rejected.

## How to test locally

This is the kind of change that's most meaningfully validated against the real Scaleway API + a real Mac mini. Test plan:

- [ ] `go build ./...` and `go test ./...` clean in both `infra/cluster-api-provider-scaleway-applesilicon` and `infra/macos-host-bootstrap`.
- [ ] On merge, the controller image rebuilds. On the next reconcile pass against the production cluster:
  - The 4 existing duplicate Machine CRs (`mndbc-m7r47`, `mndbc-xs6fx`, `mndbc-bzd26`, `mndbc-ffjpx`) need to be deleted by hand so CAPI re-creates Machines that actually adopt the pre-ordered `tuist-pool-005` and `tuist-pool-007` hosts via the two-phase path.
  - Newly-bootstrapped hosts should reach `Ready` with kcpassword=22 bytes (vs the broken 11 bytes on the current 4 hosts) and `stat -f "%Su" /dev/console` returning `m1`.
  - Existing Machines whose Secret holds an empty sudo-password should refresh on the next reconcile (a `CredentialsReclaimed` event will be emitted) and proceed to bootstrap rather than stalling on `MissingSudoPassword`.
- [ ] The pf install becomes idempotent on a re-bootstrap; can be verified by deleting + recreating a Machine and watching the controller's retry log.

## Follow-ups (out of scope here; tracked for separate PRs)

- **Operator runbook for cleaning up the existing duplicate Machine CRs** on the production cluster, plus the manual kcpassword fix that's already been applied to the 4 stuck hosts (sblfl, m789d, hz6vv, k6kwt) via SSH out-of-band. Captured in the runner-feature project notes.
- **Runner image \$HOME parity** is tracked in #10797 (independent fix).

### DIFF
--- a/infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
+++ b/infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
@@ -293,20 +293,32 @@ func (r *ScalewayAppleSiliconMachineReconciler) reconcileNormal(
 	// Secret and proceed with bootstrap. Only surface
 	// MissingSudoPassword when even the refresh comes back empty (the
 	// host genuinely has no recoverable credentials).
-	if bootstrapCreds.SudoPassword == "" {
-		if machine.Status.ServerID != "" {
-			srv, refreshErr := r.ScalewayClient.GetServer(ctx, machine.Status.ServerID, machine.Spec.Zone)
-			if refreshErr == nil && srv != nil && srv.SudoPassword != "" {
-				if writeErr := r.CredentialsManager.SetMachineCredentials(ctx, machine.Name, srv.SudoPassword, srv.SSHUsername); writeErr != nil {
-					logger.Error(writeErr, "refresh bootstrap secret from Scaleway")
-					return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
-				}
-				r.Recorder.Eventf(machine, corev1.EventTypeNormal, "CredentialsReclaimed",
-					"Recovered sudo password for %s from Scaleway vnc_url; bootstrap will resume",
-					machine.Name)
-				bootstrapCreds.SudoPassword = srv.SudoPassword
-				bootstrapCreds.SSHUsername = srv.SSHUsername
+	if bootstrapCreds.SudoPassword == "" && machine.Status.ServerID != "" {
+		srv, refreshErr := r.ScalewayClient.GetServer(ctx, machine.Status.ServerID, machine.Spec.Zone)
+		switch {
+		case refreshErr != nil:
+			// Transient Scaleway error (network blip, 5xx, throttle).
+			// Don't bury it as `MissingSudoPassword` with a 5-minute
+			// backoff — that condition is reserved for a definitive
+			// "Scaleway says there's no password". Surface a
+			// retryable condition and requeue soon so the refresh
+			// gets another chance on the next reconcile tick.
+			conditions.MarkFalse(machine, BootstrappedCondition, "CredentialsRefreshFailed",
+				clusterv1.ConditionSeverityWarning, "%v", refreshErr)
+			r.Recorder.Eventf(machine, corev1.EventTypeWarning, "CredentialsRefreshFailed",
+				"Scaleway GetServer for %s failed while trying to recover sudo password: %v",
+				machine.Name, refreshErr)
+			return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+		case srv != nil && srv.SudoPassword != "":
+			if writeErr := r.CredentialsManager.SetMachineCredentials(ctx, machine.Name, srv.SudoPassword, srv.SSHUsername); writeErr != nil {
+				logger.Error(writeErr, "refresh bootstrap secret from Scaleway")
+				return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
 			}
+			r.Recorder.Eventf(machine, corev1.EventTypeNormal, "CredentialsReclaimed",
+				"Recovered sudo password for %s from Scaleway vnc_url; bootstrap will resume",
+				machine.Name)
+			bootstrapCreds.SudoPassword = srv.SudoPassword
+			bootstrapCreds.SSHUsername = srv.SSHUsername
 		}
 	}
 	if bootstrapCreds.SudoPassword == "" {
--- a/infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
+++ b/infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
@@ -275,6 +275,27 @@ func (r *ScalewayAppleSiliconMachineReconciler) reconcileNormal(
 		return ctrl.Result{RequeueAfter: 10 * time.Second}, nil
 	}
 
+	// Refuse to bootstrap with an empty sudo password. The bootstrap
+	// path stages `/etc/kcpassword` from `UserPassword`; an empty
+	// value XORs to just the cipher key padding, loginwindow rejects
+	// the auto-login, and Aqua never comes up — `tart run` then fails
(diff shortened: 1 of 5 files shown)


pepicrft (approved):
Validated the empty-password recovery path locally. The one edge case I could still reproduce is a transient `GetServer` failure being surfaced as `MissingSudoPassword` with a 5 minute backoff instead of retrying the Scaleway refresh path. Worth a small follow-up, but not enough to hold this fix.

fortmarek:
Thanks for the catch — addressed in 0a69bc5d98 rather than as a follow-up since the diff is small. Split the refresh-failed branch out: a transient `GetServer` error now marks a `CredentialsRefreshFailed` warning condition and requeues after 30s, leaving `MissingSudoPassword` (and its 5-min backoff) reserved for the definitive "Scaleway returned a clean response with no password" case.

PR #10799, opened 2026-05-14
feat(kura): geographic attribution (country + subdivision) on metrics and traces

pepicrft:
## Summary

Kura now enriches telemetry with the geographic location of both the request client and the serving node, so we can decide where a customer would benefit from a node in a new region directly off Tempo/Prometheus data.

- The container image vendors [DB-IP Lite **City**](https://db-ip.com/db/download/ip-to-city-lite) (CC BY 4.0) at `/opt/geoip/dbip-city-lite.mmdb`, so attribution is on by default with no network access at cold start. It soft-fails (logs a warning, runs without attribution) when the file is absent on custom builds.
- Client IP (read from `X-Forwarded-For` / `X-Real-IP`) is resolved to two granularities:
  - **Country** (ISO 3166-1): `client_country` Prometheus label on `kura_http_requests_total` and `geo.country.iso_code` OTel span attribute. Falls back to `unknown` when the header is missing, the IP is private, or the DB is absent.
  - **Subdivision** (ISO 3166-2, e.g. `US-CA`): `geo.region.iso_code` OTel span attribute only.
- The serving node resolves its own country and subdivision once at startup from a single egress-IP probe and stamps them as `geo.country.iso_code` / `geo.region.iso_code` OTel Resource attributes. `KURA_NODE_COUNTRY` / `KURA_NODE_SUBDIVISION` override the probe; country additionally falls back to the `KURA_REGION` prefix.
- A background tokio task re-downloads the monthly City dump every `KURA_GEOIP_REFRESH_INTERVAL_SECS` (default `86400`, `0` disables) and swaps the in-process `RwLock<Reader<Vec<u8>>>` under a microsecond write guard, so concurrent lookups never observe a partial state. Outcomes are tracked in `kura_geoip_refresh_total{result="ok|http_error|parse_error"}`.

## Naming

Span and Resource attributes use the OpenTelemetry [`geo.*` semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/geo/) (`geo.country.iso_code`, `geo.region.iso_code`) so standard geo tooling understands them. The `client_country` Prometheus label is kept as-is (short, Prometheus-idiomatic; semconv covers spans/logs/resource, not metric label names). The existing `kura.region` resource attribute (cloud deployment region, e.g. `fr-par`) is untouched and does not collide thanks to the `geo.` namespace.

## Why subdivision is traces-only

ISO 3166-2 has thousands of codes. Adding it as a Prometheus label, crossed with route × method × status, would inflate the active series on the busiest counter. Country stays on metrics because it is bounded (~250 codes); subdivision lives on sampled traces, which is enough to compute per-request geographic distance.

## Resource bounds

The City dump is ~60 MiB compressed / ~125 MiB decompressed today. The download streams and aborts the moment it crosses the compressed cap; refresh is bounded to 128 MiB compressed / 256 MiB decompressed with a 60-second timeout. The vendored reader is plain process memory (not charged against Kura's internal cache-accounting budget) and sits comfortably within the pod memory limit. Per request the only added cost is one MMDB tree walk under a read lock.

## Privacy

Raw client IPs are never persisted: the address is resolved to a coarse country/subdivision in process and immediately discarded, so there is no per-account or per-user record. The emitted attributes and labels are aggregate operational signals with no account scope, so this is intentionally not added to `server/data-export.md` (that document is scoped to account-specific data exportable on legal request).

## Testing

- `mise exec -- cargo fmt`
- `mise exec -- cargo clippy --all-targets -- -D warnings` (from `kura/`, clean)
- `mise exec -- cargo test` (151 passed, 0 failed)


fortmarek (commented):
I found two issues worth addressing before merge:

1. In `kura/src/geoip.rs`, the compressed download cap is checked after `response.bytes().await`, which means `reqwest` has already buffered the full response into memory. That leaves the GeoIP refresh path effectively unbounded against an unexpectedly large DB-IP response. We should switch this to the streaming body API and stop once `MAX_COMPRESSED_BYTES` is exceeded.

2. This introduces a new telemetry surface (`client_country` in Prometheus and `client.country` in OTel spans), but `server/data-export.md` is not updated. Per the repo guidance, we should document this either as exportable customer analytics data or explicitly as non-exportable operational telemetry.

One product/infra suggestion while we are shaping this: I would consider starting with subdivision-level attribution instead of country-only. Country is useful directionally, but it is too coarse for large markets like the US, Canada, Brazil, India, or Australia. A `client_subdivision` value such as `US-CA`, `US-VA`, `US-TX`, `CA-ON`, or `AU-NSW` would be much more actionable for deciding where to place new Kura nodes, while still keeping cardinality and privacy risk much lower than city-level coordinates or geohashes. We can still fall back to country when subdivision is missing.

pepicrft:
Thanks Marek. Consolidating my updates into one comment. On your three points:

1. Streaming download: fixed in 4a3e8b38a2. It streams the body and bails the moment it crosses the compressed cap, so reqwest never buffers an oversized response.

2. Subdivision: took your suggestion and ran with it. Switched the vendored DB to DB-IP City Lite, so we now emit ISO 3166-2 (US-CA, CA-ON, AU-NSW, etc.) for both the client and the serving node, falling back to country when subdivision is missing. It's traces-only and deliberately off the Prometheus labels, since ISO 3166-2 is thousands of codes and would blow up cardinality on the busiest counter; country stays on metrics since it's bounded (~250 codes).

Two follow-on decisions baked in:

- Naming follows the OTel geo.* semantic conventions (geo.country.iso_code / geo.region.iso_code) on spans and the node resource, so Grafana/Tempo read them directly. The client_country Prometheus label is kept as-is and kura.region (the cloud deployment region) is untouched; the geo. namespace keeps them from clashing.
- Resource bounds: measured the City dump at ~60MiB compressed / ~125MiB decompressed, refresh capped at 128/256MiB. For the record, my earlier memory-budget worry was off: the 768MiB is Kura's internal cache-accounting Helm value, not an OS ceiling, and the real limit is the ~2Gi pod, which we're comfortably within.

PR title and description are updated to match. Good call starting at subdivision, much more actionable for the big markets you flagged.

fortmarek (commented):
I re-reviewed the latest PR state. The original streaming-size issue looks fixed, but I found three follow-up issues:

1. **[P2] Geo telemetry is still tenant-scoped in traces**

`server/data-export.md` was removed again with the rationale that the geo telemetry has no account scope. However, every exported span resource still includes `kura.tenant_id`, and `http.request` spans now record client `geo.country.iso_code` / `geo.region.iso_code`. That makes the retained Tempo data tenant-scoped in practice, so `server/data-export.md` should still document this telemetry or explicitly exclude retained operational traces.

Location: `kura/src/telemetry.rs` around the resource attributes where `kura.tenant_id`, `geo.country.iso_code`, and `geo.region.iso_code` are emitted.

2. **[P2] Region fallback emits invalid country codes**

`geo.country.iso_code` is documented as ISO 3166-1, but the fallback blindly uppercases the first two letters of `KURA_REGION`. Common region labels in this repo include `eu-west`, `ap-south`, and `local`, which become `EU`, `AP`, and `LO`; those are not valid country locations for Grafana/Tempo distance analysis. I would prefer an explicit mapping or omitting the attribute unless the prefix is known to be a real deployment country.

Location: `kura/src/node_location.rs`, `iso_prefix_from_region`.

3. **[P3] Node overrides still perform the egress probe**

`resolve_node_location` fetches `api.ipify.org` before applying overrides, so setting both `KURA_NODE_COUNTRY` and `KURA_NODE_SUBDIVISION` still incurs external egress and can add up to the 3-second startup timeout. It would be better to skip the probe when configured overrides already provide the fields being resolved.

Location: `kura/src/node_location.rs`, the initial `probed` resolution in `resolve_node_location`.

fortmarek (approved):
just a couple of extra codex-driven suggestion, but otherwise, looks good to me 👍

PR #10827, opened 2026-05-16
feat(cli): align Settings API surface, use ConfigurationName

alexanderwe:
### Summary 

This PR aligns the API surface of the `ProjectDescription.Settings` initialiser to accept `ConfigurationName` instead of `String` types. 

`ConfigurationName` is used consistently across the API surface of `ProjectDescription` but the `defaultSettings` init parameter was missing it. With this change consumers can rely on using a type safe API which does not necessarily rely on String type to the passed in the initialiser. 

No change to any other package was made and the mapping to `XcodeGraph` is then utilising the `rawValue` of the `ConfigurationName`, essentially the same as before. 

### How to test locally

Run the unit tests.

### Notes 

This should not have any breaking change impact on existing `defaultSettings` declaration since `ConfigurationName` conforms to `ExpressibleByStringLiteral`.

### DIFF
--- a/cli/Sources/ProjectDescription/Settings.swift
+++ b/cli/Sources/ProjectDescription/Settings.swift
@@ -115,7 +115,7 @@ public struct Settings: Equatable, Codable, Sendable {
     public var base: SettingsDictionary
     public var configurations: [Configuration]
     public var defaultSettings: DefaultSettings
-    public var defaultConfiguration: String?
+    public var defaultConfiguration: ConfigurationName?
 
     /// Creates settings with default.configurations `Debug` and `Release`
     ///
@@ -137,7 +137,7 @@ public struct Settings: Equatable, Codable, Sendable {
         debug: SettingsDictionary = [:],
         release: SettingsDictionary = [:],
         defaultSettings: DefaultSettings = .recommended,
-        defaultConfiguration: String? = nil
+        defaultConfiguration: ConfigurationName? = nil
     ) -> Settings {
         Settings(
             base: base,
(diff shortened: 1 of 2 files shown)


pepicrft (changes requested):
Just checked `ConfigurationName` since it's been a while since last time I interacted with it, and I don't think the value of doing `.debug` & `.release` is worth this:

> ConfigurationName is used consistently across the API surface of ProjectDescription but the defaultSettings init parameter was missing it. With this change consumers can rely on using a type safe API which does not necessarily rely on String type to the passed in the initialiser.

Have you run into an scenario where the usage of `ConfigurationName` would have prevented a typo with `Debug` or `Release`?

alexanderwe:
The problem that I faced was if your project is not using any of the "default" configurations ("Debug" or "Release") you can run into typos here. 

So the idea is that users can declare:

```swift
extension ConfigurationName {
  static let custom: Self = "Custom"
}
```

Then users can use the exact same type safe name everywhere where a configuration name needs to be passed. 

In addition everywhere where a configuration is referenced the API is accepting a `ConfigurationName` type instead of a `String` type. So I thought it would make sense to add this to the `defaultConfiguration` parameter as well. Otherwise you only have the option to use a `String`.

I hope this makes my intention clearer.

pepicrft:
Understood. Thanks @alexanderwe.
This looks good tome then, thoughts @fortmarek ?

alexanderwe:
@fortmarek @pepicrft, not urgent, but any thoughts about the next steps on this one ?

PR #10835, opened 2026-05-18
fix(infra): surface scheduling-failure taints in install-platform diagnostics

fortmarek:
## Summary

Follow-up to [#10831](https://github.com/tuist/tuist/pull/10831), narrowed to just the diagnostics improvement after we identified that the actual `tuist-kura-us-east` breakage was 4 CAPI Machines stuck for 9 days in \`handleBootStateUnset\` with MHC's circuit-breaker correctly blocking remediation — an upstream-of-deploy problem that needs fixing at the cluster level, not by relaxing the deploy's fail/no-fail semantics.

The deploy-time failure mode that misled us was: \`platform-ingress-nginx-admission-create\` pod \`Pending\` with \`0/1 nodes are available: 1 node(s) had untolerated taint(s)\`. \`install-platform.sh\`'s existing on-failure dump showed the stuck pod but not which taint was blocking it — making this look like a generic helm hook timeout. This PR adds two cheap commands to that dump so the next stuck regional cluster's deploy log says what's wrong directly:

- \`kubectl get nodes -o wide\` — full node list
- \`kubectl get nodes -o jsonpath='...{.spec.taints}...'\` — per-node taints, which \`-o wide\` hides

No behavioral change. The Kura regional deploy still hard-fails the production job when a region is broken; that gating is intentional (silently shipping past a broken regional cluster would mask exactly the kind of infra rot that just bit us).

## Test plan

- [ ] Next time \`mise k8s:install-platform\` fails on a regional cluster, confirm the \`--- nodes ---\` and \`--- node taints ---\` blocks show up in the install-platform diagnostics group in the deploy log.

### DIFF
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -427,20 +427,10 @@ jobs:
             --atomic --timeout 30m --wait
       - name: Deploy Kura regional controllers
         if: inputs.environment == 'production'
-        # A broken regional Kura cluster (e.g. workers that never joined)
-        # must not gate the main tuist deploy: tuist.dev serves customers
-        # from the main cluster, while regional Kura controllers serve an
-        # opt-in cache feature that can run on its previously deployed
-        # image until the regional cluster is recovered. Per-region
-        # failures are captured below and reported as a workflow warning
-        # so the cascade (canary -> acceptance -> production) can still
-        # promote the main service. The step itself stays green; the
-        # ::warning:: annotations surface stuck regions for follow-up.
-        continue-on-error: true
         env:
           OP_SERVICE_ACCOUNT_TOKEN: ${{ secrets.OP_SERVICE_ACCOUNT_TOKEN }}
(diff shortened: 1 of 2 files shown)


pepicrft (approved):
Looks good to me.

One follow-up I would suggest is surfacing regional Kura deploy failures as a degraded-success signal instead of leaving them only in the logs. The current change makes sense because it unblocks the main production deploy, but we should still emit something obvious when a region fails, for example a dedicated Slack warning and a clear step summary that lists the failed regions.

That would preserve the non-blocking behavior while making the partial failure state visible.

fortmarek:
> One follow-up I would suggest is surfacing regional Kura deploy failures as a degraded-success signal instead of leaving them only in the logs. The current change makes sense because it unblocks the main production deploy, but we should still emit something obvious when a region fails, for example a dedicated Slack warning and a clear step summary that lists the failed regions.

I'll add a grafana rule

PR #7200, opened 2024-12-26
Add TestableTarget.parallelization property

fortmarek:
### Short description 📝

Resolves: https://github.com/tuist/tuist/issues/6827
Supersedes: https://github.com/tuist/tuist/pull/7185

This PR adds new API in `ProjectDescription` for specifying the new options for running testable targets in parallel:
- `disabled`
- `swiftTestingOnly`
- `enabled`

The current property `TestableTarget.parallelizable` is now deprecated to the `parallelization` enum with the above cases.

The initial piece was implemented in `XcodeProj` in: https://github.com/tuist/XcodeProj/pull/871

The prerequisite of this PR is: https://github.com/tuist/XcodeGraph/pull/92

### How to test the changes locally 🧐

> Include a set of steps for the reviewer to test the changes locally (see [the documentation](https://docs.tuist.io/contributors/get-started) for reference).

### Contributor checklist ✅

- [x] The code has been linted using run `mise run lint-fix`
- [x] The change is tested via unit testing or acceptance testing, or both
- [x] The title of the PR is formulated in a way that is usable as a changelog entry
- [x] In case the PR introduces changes that affect users, the documentation has been updated

### Reviewer checklist ✅

- [ ] The code architecture and patterns are consistent with the rest of the codebase
- [ ] Reviewer has checked that, if needed, the PR includes the label `changelog:added`, `changelog:fixed`, or `changelog:changed`, and the title is usable as a changelog entry

PR #10838, opened 2026-05-18
feat(cli): add tri-state parallelization to autogen schemes

shgew:
## Why

[#7200](https://github.com/tuist/tuist/pull/7200) gave `TestableTarget` a `.swiftTestingOnly` parallelization mode for hand-rolled schemes: parallelize the Swift Testing tests in a target while running its XCTest tests serially. Autogenerated schemes never got it. The options that drive autogen schemes - `Project.Options.AutomaticSchemesOptions.enabled` and the workspace's `AutogeneratedWorkspaceSchemes.enabled` - carry a `TestingOptions` `OptionSet` whose only parallelization flag, `.parallelizable`, is all-or-nothing.

## What

Add a third `TestingOptions` flag, `swiftTestingOnlyParallelizable`, so autogenerated schemes can request the Swift-Testing-only mode. The public manifest API is unchanged: you keep configuring it through the existing `testingOptions:` option set.

```swift
let project = Project(
    name: "App",
    options: .options(
        automaticSchemesOptions: .enabled(
            testingOptions: [.swiftTestingOnlyParallelizable]
        )
    ),
    targets: [
        .target(name: "App", destinations: .iOS, product: .app, bundleId: "dev.tuist.App", sources: ["App/**"]),
        .target(name: "AppTests", destinations: .iOS, product: .unitTests, bundleId: "dev.tuist.AppTests", sources: ["AppTests/**"], dependencies: [.target(name: "App")]),
    ]
)
```

The flag flows through the existing manifest mappers into the graph unchanged. The autogenerated-scheme generators (`AutogeneratedSchemesProjectMapper`, `AutogeneratedWorkspaceSchemeGraphMapper`) resolve the option set into a `TestableTarget.Parallelization` through a new `TestingOptions.parallelization` accessor, with this priority: `swiftTestingOnlyParallelizable` → `.swiftTestingOnly`, else `.parallelizable` → `.all`, else `.none`.

## Why this approach

An earlier revision of this PR reshaped the `.enabled` case payloads into flat `parallelization` + `randomExecutionOrdering` parameters. That was source-breaking for anyone pattern-matching `AutomaticSchemesOptions.enabled` / `AutogeneratedWorkspaceSchemes.enabled` (a static-func overload preserves construction but not destructuring), and it changed the `tuist dump` output.

Adding a flag to the existing `OptionSet` instead keeps the change minimal and non-breaking:

- The stored enum payloads are untouched, so pattern matching still compiles.
- `tuist dump` keeps emitting `testingOptions`; existing values are unchanged and the new mode is just an additional bit.

## Impact

- New `TestingOptions.swiftTestingOnlyParallelizable` option for autogenerated schemes (project- and workspace-level). It takes precedence over `.parallelizable` when both are set.
- No breaking changes to the manifest API or to `tuist dump` output.

## How to test

```bash
  -only-testing:TuistLoaderTests/ProjectManifestMapperTests \
  -only-testing:TuistLoaderTests/WorkspaceGenerationOptionsManifestMapperTests \
  -only-testing:TuistGeneratorTests/AutogeneratedSchemesProjectMapperTests \
  -only-testing:TuistGeneratorTests/AutogeneratedWorkspaceSchemeWorkspaceMapperTests \
  -only-testing:TuistGeneratorTests/SchemeDescriptorsGeneratorTests \
  -only-testing:TuistKitTests/DumpServiceTests \
```

End to end: drop the manifest above into a fresh project, run `tuist generate --no-open`, and inspect the generated `App.xcscheme`. The `TestableReference` for `AppTests` will have no `parallelizable` attribute - that absence is how Xcode encodes Swift Testing-only parallelization.


cli/Tests/TuistKitTests/Commands/DumpServiceIntegrationTests.swift
```diff
@@ -66,7 +66,8 @@ final class DumpServiceTests: TuistTestCase {
                         ]
                       }
                     },
-                    "testingOptions": 0
+                    "parallelization": "disabled",
```
shgew: This potentially can break the consumers of `tuist dump`, which is undesirable. I couldn't find a perfect way to solve it, since the new state can't really be represented using old options. I will gladly address this though with some guidance :)

shgew:
@pepicrft I hate to tag you like this, but maybe you would have a chance to look into this in a near future?

cli/Sources/ProjectDescription/ProjectOptions.swift
```diff
@@ -81,6 +82,40 @@ extension Project {
 
             /// Disable autogenerated schemes
             case disabled
+
+            /// Deprecated overload preserving the legacy `testingOptions:` signature.
+            /// Forwards to the new case shape so existing manifests keep compiling.
+            @_disfavoredOverload
+            @available(
```
pepicrft: This keeps `.enabled(testingOptions:)` construction working through the deprecated overload, but it still breaks users who pattern-match the enum case or bind its associated values in helper code.

Could we avoid changing this case payload for now? One option is to keep `testingOptions: TestingOptions` as the stored associated value, add a new `TestingOptions` bit for Swift Testing only, and expose the new `parallelization` / `randomExecutionOrdering` API as convenience factories or computed accessors. That would preserve source compatibility and still unblock the tri-state behavior.

cli/Sources/ProjectDescription/WorkspaceGenerationOptions.swift
```diff
@@ -17,14 +17,40 @@ extension Workspace {
...
+                randomExecutionOrdering: Bool = false,
                 testLanguage: SchemeLanguage? = nil,
                 testRegion: String? = nil,
                 testScreenCaptureFormat: ScreenCaptureFormat? = nil
             )
+
+            /// Deprecated overload preserving the legacy `testingOptions:` signature.
+            /// Forwards to the new case shape so existing manifests keep compiling.
+            @_disfavoredOverload
+            @available(
```
pepicrft: Same compatibility concern as the project options case: changing the associated values makes direct enum pattern matching source-breaking, even though the deprecated factory preserves labeled construction.

I’d prefer keeping the stored enum shape stable and mapping the new tri-state API into the legacy `TestingOptions` representation until a major version.

cli/Tests/TuistKitTests/Commands/DumpServiceIntegrationTests.swift
pepicrft: Can we keep emitting `testingOptions` for compatibility, or emit both fields for one release? If we keep the old enum payload internally, this would fall out naturally and avoid making the dump format part of the breaking change.

shgew:
@pepicrft thank you for the guidance! I have updated the PR, please let me know if this closely aligns with your vision for this feature.

cli/Sources/ProjectDescription/ProjectOptions.swift
shgew: Updated

cli/Sources/ProjectDescription/WorkspaceGenerationOptions.swift
shgew: Updated

cli/Tests/TuistKitTests/Commands/DumpServiceIntegrationTests.swift
shgew: I think having both is too much... Like you say, this change seems like a better fit for the next major release.

shgew:
Updated PR description

shgew:
@pepicrft maybe you have some time to take a look?

pepicrft (approved):

shgew:
thank you for finding the time @pepicrft !

Issue #2, opened 2018-04-16
Add install/uninstall scripts like Homebrew


pepibumur:

pepibumur:
This is already done 🎉

PR #10858, opened 2026-05-19
fix(infra, capi-scaleway): unstick runners-fleet (pf tables + manual-cleanup race + prep-script simplification)

### LINKED ISSUE #1: Setup continuous integration with Travis

fortmarek:
## Summary

Three related fixes uncovered while diagnosing a stuck production runners-fleet on 2026-05-19.

### 1. `installVMEgressFirewall` couldn't recover hosts with persist'd pf tables

Three CRs (`bzd26`, `ffjpx`, `k6kwt`) all bound to the same host had been failing bootstrap for 5 days with:

```
/etc/pf.anchors/tuist.runners:11: cannot define table vm_sources: Resource busy
/etc/pf.anchors/tuist.runners:12: cannot define table blocked_dst: Resource busy
pfctl: Syntax error in config file: pf rules not loaded
```

`pfctl(8)` defines `-T kill` as "Kill a table" — i.e. destroy it from kernel state — and that's the only command that removes a `persist`'d table. `-F all` / `-F Tables` explicitly preserve persist tables (they only flush addresses).

The original script's `-T kill` invocation was likely failing because older versions of the script had at various times placed these tables in either anchor scope or the main ruleset, and the cleanup only operated on one of them with the other silently swallowed by `2>/dev/null || true`. Fix: kill at *both* scopes, then load an empty anchor as belt-and-suspenders for stale anchor-level rule references.

### 2. Operator-driven CR cleanup raced `AdoptByPrefix`

When recovering from a duplicate-claim state (or hand-rolling a CR off a shared host), the operator clears `status.serverID` so `reconcileDelete` skips Scaleway release, then deletes the CR. There's an unavoidable race today: between the patch and the delete, the reconcile loop can see the empty `serverID`, run `AdoptByPrefix` against the pool, rename a `tuist-pool-*` host to the doomed CR's claim name, and persist the bind. Then `reconcileDelete` tries to `DeleteServer` against the freshly-claimed pool host and gets stuck on the Apple 24h licensing floor.

Switch the pause gate to honor the standard CAPI `cluster.x-k8s.io/paused` annotation on the infra CR itself (in addition to `Cluster.Spec.Paused`). Operators set the annotation *first*, then patch, then delete; the annotation latches `reconcileNormal` off until `DeletionTimestamp` lands. `reconcileDelete` is intentionally above the pause gate so teardown still progresses on paused CRs.

Operator recipe captured in `AGENTS.md`.

### 3. `prepare-fleet-host` dropped sshpass dependency

Discovered while recovering pool hosts whose CAPI bootstrap was failing on sudo: the script's sshpass fallback path is the source of every operator friction in this flow (PTY exhaustion, heredoc-quoting bugs, silent EXIT=5, zsh/bash `read -rsp` portability). It's also unnecessary — Scaleway Apple Silicon hosts auto-inject all project-level SSH keys into `~/.ssh/scw_authorized_keys` at first boot, and sshd reads both `authorized_keys` and `scw_authorized_keys`. Once `EnsureFleetSSHKey` has registered the fleet pubkey at the project level, every newly-ordered pool host accepts the fleet private key out of the box.

Drop the sshpass path. The script now uses SSH-key transport exclusively, installs `/etc/sudoers.d/m1-nopasswd` (one-time `sudo -S` use of the password), then installs `/etc/kcpassword` + `autoLoginUser` via the now-passwordless sudo. On SSH-key probe failure, errors out with a clear recovery path (VNC + manual paste into `scw_authorized_keys`, or reinstall host).

## Test plan

- [x] `go test ./controllers/...` — adds `TestReconcile_PausedAnnotationSkipsAdoption` covering the new pause gate.
- [x] Script syntax validated (`bash -n`).
- [x] Pause-annotation cleanup recipe used in production on 2026-05-19 to recover 4 duplicate CRs; worked as designed.
- [x] Simplified `prepare-fleet-host` used in production on 2026-05-19 to recover `h5dkz` and `njpkv` (both successfully bootstrapped after the operator-side sudoers install).
- [ ] **(blocking draft)** Build operator image with this change, deploy to production cluster, observe `tuist-tuist-runners-fleet-mndbc-k6kwt` flip from `BootstrappedCondition=False` to `True` (the host with stuck persist tables — proves fix #1 against the real failure).

## Context

Full diagnosis snapshot in `/tmp/capi-rescue/diagnosis.md` (off-tree); cluster recovery walked from 3/9 Ready hosts to 8/9 over the course of the day, with `k6kwt` blocked on this PR being deployed.

### DIFF
--- a/infra/cluster-api-provider-scaleway-applesilicon/AGENTS.md
+++ b/infra/cluster-api-provider-scaleway-applesilicon/AGENTS.md
@@ -192,6 +192,48 @@ kubectl describe scalewayapplesiliconmachine <name>
 kubectl get events --field-selector involvedObject.kind=ScalewayAppleSiliconMachine
 ```
 
+### Unstick a host whose CAPI bootstrap is failing on sudo
+
+Symptom: `kubectl describe scalewayapplesiliconmachine <name>` shows
+`BootstrappedCondition=False` with a message containing `sudo:`
+errors, or the bootstrap looping on early SSH steps after Stage 1
+provisioning succeeded.
+
+Root cause is almost always: the operator-stored `m1` password in
+the bootstrap Secret has drifted from what's actually set on the
+host (Scaleway-issued password rotated, host got reinstalled,
+controller crashed mid-store, etc.). CAPI's bootstrap can SSH in
+(fleet key works) but can't `sudo -S` to install
+`/etc/sudoers.d/m1-nopasswd`, so every subsequent step fails.
+
+Recovery: run `prepare-fleet-host` to install the sudoers entry
+out-of-band using the operator-provided current password.
+
+```bash
+# Get the live m1 password from Scaleway:
+scw apple-silicon server get <server-id> zone=<zone> -o json \
+  | jq -r .vnc_url
+# (Password is between `m1:` and `@` in the vnc:// URL.)
+
+# Then:
+mise run k8s:prepare-fleet-host <env> <fleet-name> <host-ip>
+```
+
+The script SSHes in with the fleet key (which Scaleway auto-injects
+at first boot via project-level keys), prompts for the password,
+installs `/etc/sudoers.d/m1-nopasswd` and `/etc/kcpassword` /
+`autoLoginUser`. After that, CAPI's bootstrap proceeds without ever
+needing a correct password in its Secret.
+
+If the fleet pubkey isn't on the host (rare — Scaleway didn't
+inject), the script's SSH probe fails with `Permission denied`.
+Recover by VNC'ing into the host and pasting the pubkey into
+`~/.ssh/scw_authorized_keys`, then re-running the script. Don't
+bother with `~/.ssh/authorized_keys` — Scaleway's `sshd_config`
+reads both files, but `scw_authorized_keys` is the one the
+first-boot injection writes to, so anything you add there
+mirrors the auto-inject convention.
+
 ### Detach a CR without releasing its Scaleway host
 
 Reserved for recovering from a duplicate-claim state (multiple CRs
--- a/infra/cluster-api-provider-scaleway-applesilicon/AGENTS.md
+++ b/infra/cluster-api-provider-scaleway-applesilicon/AGENTS.md
@@ -191,3 +191,55 @@ kubectl describe scalewayapplesiliconmachine <name>
 # transitions, drift-loop attempts, terminal-failure transitions)
 kubectl get events --field-selector involvedObject.kind=ScalewayAppleSiliconMachine
 ```
+
+### Detach a CR without releasing its Scaleway host
+
+Reserved for recovering from a duplicate-claim state (multiple CRs
+ended up bound to the same Scaleway server) or for hand-rolling a CR
+off a host that's actively serving traffic. The standard
+`kubectl delete machine <name>` path always calls Scaleway's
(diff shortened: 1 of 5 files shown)


pepicrft (approved):
Reviewed all 4 commits. Three independent infra fixes from one incident. Solid direction, approving. A few notes for before un-drafting.

## Fix #2: controller pause gate

Right call honoring `cluster.x-k8s.io/paused` on the CR. Design is sound: `reconcileDelete` runs above the pause gate, the annotated CR tears down, and the MachineSet recreates a fresh (unannotated) CR that adopts normally. The `HasPaused` vs `IsPaused` split is correctly reasoned.

- Minor ordering regression: the refactor moved the `InfrastructureReady` check before the pause evaluation. Previously `Spec.Paused` returned a quiet stop; now a paused owned CR whose parent Cluster is `InfrastructureReady=false` requeues every 30s instead of going silent. Not a correctness bug for the production fleet (its cluster is infra-ready), but the cleanest design evaluates both pause signals before any requeue. Consider hoisting the pause checks above the infra-ready gate.
- `HasPaused` is presence-based, so `paused: "false"` also pauses (standard CAPI). Recipe uses `=true`, so fine, just noting.

## Fix #1: pf persist-table eviction

Logic is sound. `-T kill` is the documented table-destroy primitive and `-F all` does preserve persist tables, so the old code genuinely could not recover. Dual-scope kill plus empty-anchor load is reasonable, and the `2>/dev/null || true` is acceptable because the `pfctl -nf` validate surfaces a clear error if still wedged.

- Primary risk is already acknowledged: unverified against the real stuck host. Correctly gated as draft.
- Commit narrative is self-contradicting: commit 1 asserts `-T kill` does not destroy tables, commit 3 reverts that claim. Please squash commits 1+3 before merge so history does not carry the wrong rationale.

## Fix #3: drop sshpass

Good simplification, fewer failure modes, premise is correct (Scaleway auto-injects project keys into `scw_authorized_keys` at first boot, sshd reads both files). Clear probe-fail recovery path.

- Minor regression risk: `printf '%s' "$SUDO_PW" | sudo -S` dropped the trailing newline the old code had. `sudo -S` expects a newline-terminated password. EOF usually works but the `\n` is the standard idiom and there is no upside to dropping it. Recommend restoring it.
- Pre-existing (optional): `SUDO_PW='$SUDO_PW'` interpolated single-quoted onto the remote command line breaks on a single-quote password and is visible in `ps`. Since you are already simplifying this path, passing via stdin/env-file would harden it. Low priority given Scaleway passwords are alphanumeric.

## Tests

`TestReconcile_PausedAnnotationSkipsAdoption` is good (the nil-client trick is a clean failure signal). Gaps:

- The "pause must not block teardown" contract is asserted in a comment but not tested. Worth a sibling test: paused CR with `DeletionTimestamp` still reaches `reconcileDelete` and skips Scaleway release when `serverID` is empty. That is the exact scenario this PR exists for.
- Owned-CR-with-annotation path (cluster present and ready) is not covered, only the standalone path.

## Recommendation

Approving the direction. Before un-drafting: complete the production verification of fix #1, squash the contradictory pf commits, restore the `\n` in the `sudo -S` pipe. Consider also hoisting the pause checks above the infra-ready requeue and adding the reconcileDelete-while-paused test.

PR #10872, opened 2026-05-19
fix(cli): re-run foreign build script when inputs cannot be tracked

fortmarek:
### Purpose

Reported in the Tuist Community Slack: a user wired up `foreignBuild()` for a KMP module, but editing the Kotlin sources didn't rebuild the framework — the iOS app kept linking the stale xcframework.

The root cause is a silent failure in input expansion. When a `.glob` or `.folder` input resolves to zero files at generation time (most commonly because `.relativeToRoot("../...")` aims at the wrong path), the Xcode shell script build phase is written with an empty `inputPaths`. Combined with an existing `outputPaths` xcframework, Xcode's dependency analysis decides the script is up-to-date forever — it runs once on the first build, then never again. Source changes on the non-Xcode side are dropped on the floor without any feedback.

Reproduced statically by inspecting the generated `project.pbxproj` against a fixture whose `.glob` aims at a non-existent path:

```
inputPaths = (
);
outputPaths = (
    ../../../../myAwesomeKmpProject/build/SharedKMP.xcframework,
);
```

No `alwaysOutOfDate`, no warning, no way for the user to notice.

### Fix

- Warn when a `.glob` input matches zero files (in the manifest mapper).
- Warn when a `.folder` input expands to no files (in `ForeignBuildGraphMapper`).
- When the final `inputPaths` for a foreign build script is empty, set `basedOnDependencyAnalysis = false`, which the build-phase generator translates to `alwaysOutOfDate = 1` in the pbxproj. The script will then re-run on every build instead of caching the stale output indefinitely.
- Document the failure mode on `ForeignBuild.Input` so it's discoverable from autocomplete.

When inputs do resolve correctly, behavior is unchanged: `inputPaths` is populated, `alwaysOutOfDate` is not set, and Xcode does its usual incremental analysis.

### How to test locally

1. Set up a fixture where the foreign build's `.glob` / `.folder` input resolves to no files (e.g. a wrong `..` count in `.relativeToRoot`):

```swift
.foreignBuild(
    name: "SharedKMP",
    destinations: .iOS,
    script: "...",
    inputs: [.glob(.relativeToRoot("does-not-exist/**/*.kt"))],
    output: .xcframework(path: ..., linking: .dynamic)
)
```

2. `tuist generate` — observe the warning:

```
Foreign build glob input '.../does-not-exist/**/*.kt' matched no files. ...
```

3. Inspect the generated `project.pbxproj` and confirm the script phase has `alwaysOutOfDate = 1`.

4. Build, then build again without changing anything — the script should re-run (whereas before this fix it would have been silently skipped).

5. Run `ForeignBuildGraphMapperTests` for unit coverage.

### DIFF
--- a/cli/Sources/ProjectDescription/Target.swift
+++ b/cli/Sources/ProjectDescription/Target.swift
@@ -124,23 +124,18 @@ public struct Target: Codable, Equatable, Sendable {
         /// - **Content hashing**: all inputs (including scripts) are used to compute a content hash for Tuist's
         ///   binary caching and selective testing, so that the foreign build step can be skipped when inputs are unchanged.
         ///
-        /// > Important: If folder/glob inputs resolve to no files at generation time, Xcode has nothing to track and
-        /// > the script will be marked as always out-of-date so it re-runs on every build. Tuist emits a warning in
-        /// > that case — double-check the configured paths.
+        /// > Important: If folder/glob inputs resolve to no files, Xcode has nothing to track for incremental
+        /// > builds, and the foreign build script is marked as always out-of-date so it re-runs on every build.
+        /// > This is fine when the underlying build system (Gradle, Cargo, CMake, …) handles its own
+        /// > incrementality; otherwise, double-check the configured paths.
         public enum Input: Codable, Hashable, Sendable {
             /// A single file whose contents affect the build output.
             case file(Path)
 
             /// A folder whose contents (recursively) affect the build output.
-            ///
-            /// The folder is expanded at generation time. If the folder doesn't exist or is empty, Tuist emits
-            /// a warning since Xcode would otherwise have no inputs to track for incremental builds.
             case folder(Path)
 
             /// A glob pattern that resolves to files whose contents affect the build output.
-            ///
-            /// The pattern is expanded at generation time. If the pattern matches no files, Tuist emits a
-            /// warning since Xcode would otherwise have no inputs to track for incremental builds.
             case glob(Path)
 
             /// A shell script whose stdout produces a cache key component (e.g. `"git rev-parse HEAD"`).
--- a/cli/Sources/ProjectDescription/Target.swift
+++ b/cli/Sources/ProjectDescription/Target.swift
@@ -123,14 +123,24 @@ public struct Target: Codable, Equatable, Sendable {
         ///   build phase so that Xcode can skip re-running the script when inputs haven't changed.
         /// - **Content hashing**: all inputs (including scripts) are used to compute a content hash for Tuist's
         ///   binary caching and selective testing, so that the foreign build step can be skipped when inputs are unchanged.
+        ///
+        /// > Important: If folder/glob inputs resolve to no files at generation time, Xcode has nothing to track and
+        /// > the script will be marked as always out-of-date so it re-runs on every build. Tuist emits a warning in
+        /// > that case — double-check the configured paths.
         public enum Input: Codable, Hashable, Sendable {
             /// A single file whose contents affect the build output.
             case file(Path)
 
             /// A folder whose contents (recursively) affect the build output.
+            ///
+            /// The folder is expanded at generation time. If the folder doesn't exist or is empty, Tuist emits
+            /// a warning since Xcode would otherwise have no inputs to track for incremental builds.
             case folder(Path)
 
             /// A glob pattern that resolves to files whose contents affect the build output.
+            ///
+            /// The pattern is expanded at generation time. If the pattern matches no files, Tuist emits a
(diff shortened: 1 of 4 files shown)


cli/Sources/ProjectDescription/Target.swift
```diff
@@ -123,14 +123,24 @@ public struct Target: Codable, Equatable, Sendable {
         ///   build phase so that Xcode can skip re-running the script when inputs haven't changed.
         /// - **Content hashing**: all inputs (including scripts) are used to compute a content hash for Tuist's
         ///   binary caching and selective testing, so that the foreign build step can be skipped when inputs are unchanged.
+        ///
+        /// > Important: If folder/glob inputs resolve to no files at generation time, Xcode has nothing to track and
```
pepicrft: This comments about warnings are specific to the foreign build scenario, no? If so, I'd drop them.

cli/Sources/TuistGenerator/Mappers/ForeignBuildGraphMapper.swift
```diff
@@ -92,12 +94,23 @@ public struct ForeignBuildGraphMapper: GraphMapping {
                         try await fileSystem.exists($0, isDirectory: false)
                     }
                     .map(\.pathString)
+                if filePathStrings.isEmpty {
+                    Logger.current.warning(
```
pepicrft: Let's use `AlertController` for the warning. Worth codifying with swiftlint.

cli/Sources/TuistGenerator/Mappers/ForeignBuildGraphMapper.swift
```diff
@@ -92,12 +94,23 @@ public struct ForeignBuildGraphMapper: GraphMapping {
...
+                }
                 pathStrings.append(contentsOf: filePathStrings)
             case .script:
                 break
             }
         }
 
+        if pathStrings.isEmpty {
+            Logger.current.warning(
+                "Foreign build target '\(targetName)' has no resolvable input file paths. The foreign build script will run on every build because Xcode has nothing to track for incremental builds."
```
pepicrft: And this is not necessarily bad if the underlying build system handles the incrementality, so worth mentioning.

cli/Sources/TuistLoader/Models+ManifestMappers/TargetDependency+ManifestMapper.swift
```diff
@@ -133,6 +133,11 @@ extension XcodeGraph.ForeignBuild.Input {
             let resolvedPath = try generatorPaths.resolve(path: path)
             let pattern = String(resolvedPath.pathString.dropFirst())
             let matchedPaths = try await fileSystem.glob(directory: AbsolutePath.root, include: [pattern]).collect().sorted()
+            if matchedPaths.isEmpty {
+                Logger.current.warning(
```
pepicrft: Let's use `AlertController`.

pepicrft (approved):

cli/Sources/ProjectDescription/Target.swift
fortmarek: Dropped the per-case warning notes from `.folder` and `.glob` and tightened the enum-level callout to focus on the user-visible behavior (always out-of-date) and when that's fine (underlying build system handles incrementality).

cli/Sources/TuistGenerator/Mappers/ForeignBuildGraphMapper.swift
fortmarek: Switched both foreign-build warnings to `AlertController.current.warning(.alert(...))`. Holding off on a swiftlint rule in this PR — `Logger.current.warning` is still used in many other call sites (e.g. `FileElement+ManifestMapper.swift`, `CopyFileElement+ManifestMapper.swift`) and a sweep across them feels like its own change.

cli/Sources/TuistGenerator/Mappers/ForeignBuildGraphMapper.swift
fortmarek: Reworded to acknowledge that this is fine when the underlying build system (Gradle, Cargo, CMake, …) handles its own incrementality, and only suggest declaring inputs when that isn't the case.

cli/Sources/TuistLoader/Models+ManifestMappers/TargetDependency+ManifestMapper.swift
fortmarek: Switched the glob warning to `AlertController.current.warning(.alert(...))` as well.

PR #10878, opened 2026-05-20
fix(server): harden GitHub webhook dispatch + add deliveries inspector task

fortmarek:
Hardens the `workflow_job` webhook dispatch path against the failure mode that caused the 2026-05-19 incident, and adds a small operator tool to inspect (and redeliver) failed webhooks.

## What broke on 2026-05-19

PostgreSQL SSL disconnects → request handlers hung on PG checkouts → HTTP body-read timeouts spiked 77x → backend connections reset to ingress → 3.78 eps of 502s for ~30 min. Some `workflow_job.queued` deliveries never reached the controller (visible only in the GitHub App's delivery log); the ones that did were uniformly marked `outcome=ignored` because the synchronous K8s LIST inside `match_pool` stalled with everything else and returned `{:error, :no_pools}`. Net effect: jobs queued in GitHub stayed queued because Tuist never persisted them.

## Three changes

1. **Async dispatch via Oban.** Controller now enqueues a `Tuist.Runners.Workers.DispatchWorker` onto the existing `:webhooks` queue and 200s the moment the signature is verified. PG + K8s + CH are off the request path. `unique` by `delivery_guid` collapses GitHub retries of the same payload.
2. **Cache the K8s LIST and account lookup.** RunnerPool CRs are operator-managed and `runner_max_concurrent` is re-checked in PG at claim time, so a short cache is safe. Pools: 30s TTL, success-only. Accounts: 60s TTL, successful lookups only.
3. **Split the `:ignored` telemetry outcome.** Each ignore branch carries its own atom (`no_account` / `runners_disabled` / `no_matching_pool` / `no_pools` / `ambiguous_pool`) and that atom flows through to `tuist_runners_webhook_count_total`. A `no_pools` spike will now be a distinct, alertable signal — the previous flat `ignored` masked the apiserver outage as user-caused.

## Mise task: `runner:recent-deliveries`

`mise run runner:recent-deliveries -- --env production --since 2026-05-19T15:20:00Z` pulls the App credentials from 1Password, mints an RS256 JWT, and queries `GET /app/hook/deliveries`. Filters by `--since` / `--workflow-job-id`; `--redeliver <id>` reissues a failed delivery without going to the GitHub UI. Used during the incident investigation; checked in so it's reusable.

## How to test locally

Server changes:

```bash
cd server
mix test test/tuist/runners/dispatch_test.exs \
         test/tuist/runners/workers/dispatch_worker_test.exs \
         test/tuist_web/controllers/webhooks/github_controller_test.exs
mix credo lib/tuist/runners/dispatch.ex lib/tuist/runners/workers/dispatch_worker.ex lib/tuist_web/controllers/webhooks/github_controller.ex
```

Mise task (requires `op` signed into `tuist.1password.com`):

```bash
mise run runner:recent-deliveries -- --env production --since 2026-05-19T15:20:00Z --limit 500 \
| jq 'group_by({event, action, status_code}) | map({event: .[0].event, action: .[0].action, status_code: .[0].status_code, n: length}) | sort_by(-.n)'
```

Prints a histogram of webhook outcomes for the given window.


server/test/tuist/runners/dispatch_test.exs
```diff
@@ -1,10 +1,175 @@
...
+    # `KeyValueStore` writes to the application-wide `:tuist`
+    # Cachex, so tests run in parallel without colliding only as
+    # long as each one's cache keys are unique. Account keys are
+    # already keyed on the per-test org name (`organization_fixture`
+    # uses a unique integer); pinning the namespace to a fresh
+    # value here makes the pool-cache key unique too. The stub
+    # closes over a single string so multiple invocations inside
+    # one test return the same namespace (a `unique_integer/1`
+    # call inside the stub would yield a different namespace on
+    # each call and the cache would never hit).
```
pepicrft: When I first had to solve this, I leaned on scoping a cache to the test's lifecycle and passing that instead. Both achieve parallelization, but I find that the other one is cleaner.
I'd also drop comments like this

server/test/tuist/runners/dispatch_test.exs
```diff
@@ -1,10 +1,175 @@
...
+
+      stub(Accounts, :get_account_by_handle, fn _ -> account end)
+      stub(Client, :list_runner_pools, fn _ns -> {:ok, []} end)
+
+      assert {:ignored, :no_pools} =
+               Dispatch.handle_webhook(queued_payload(owner: account.name), 1)
+    end
+  end
+
+  describe "caching" do
```
pepicrft: I'd align with the convention of using `describe` to describe the function, under test, `list_runner_pools_cached` or `queued_payload`, and then have particular test cenarios that ensure that caching is done as expected in the implementation.

server/test/tuist_web/controllers/webhooks/github_controller_test.exs
```diff
@@ -608,4 +608,56 @@ defmodule TuistWeb.Webhooks.GitHubControllerTest do
       assert GitHubController.resolve_webhook_secret(conn) == "github-com-env-secret"
     end
   end
+
+  describe "handle/2 workflow_job" do
+    alias Tuist.Runners.Workers.DispatchWorker
```
pepicrft: Let's not alias inside `describe` bodies.

server/test/tuist_web/controllers/webhooks/github_controller_test.exs
```diff
@@ -608,4 +608,56 @@ defmodule TuistWeb.Webhooks.GitHubControllerTest do
...
+
+    test "200s immediately and enqueues a DispatchWorker job with the payload + GUID", %{conn: conn} do
+      installation_id = 111_169_140
```
pepicrft: Let's generate a random number here.

pepicrft (approved):

server/test/tuist/runners/dispatch_test.exs
fortmarek: Done. Dispatch tests now `start_supervised!` a per-test Cachex and the production code reads the cache name from the process dictionary (defaulting to `:tuist` in prod). Dropped the long comment.

server/test/tuist/runners/dispatch_test.exs
fortmarek: Reorganized: caching scenarios now live under `describe "handle_webhook/2"` and `describe "match_pool/1"` alongside the public functions they exercise.

server/test/tuist_web/controllers/webhooks/github_controller_test.exs
fortmarek: Moved the `DispatchWorker` alias to the module-level aliases.

server/test/tuist_web/controllers/webhooks/github_controller_test.exs
fortmarek: Switched to `System.unique_integer([:positive])`.

PR #10890, opened 2026-05-20
perf(ci): unblock Release workflow by skipping cliff's GitHub API enrichment

fortmarek:
### Summary

The Release workflow's _Check for releasable changes_ gate was hitting its 15-minute job timeout and blocking releases ([example run](https://github.com/tuist/tuist/actions/runs/26156564462/job/76937391265)). The job ran 15 sequential component checks; the last two had not even started when the runner was killed.

Two compounding causes:

1. **cliff was hammering the GitHub API.** With `GITHUB_TOKEN` set, `git-cliff` enriches each commit with PR title + username via the API. For `--bumped-version` that metadata is irrelevant, but cliff still made the calls — 60-180s per component check.
2. **Each component invoked cliff twice** (full changelog _and_ `--bumped-version`).

### Changes

- `GITHUB_TOKEN` and `GH_TOKEN` are unset before cliff runs in the check step. Per-component cliff calls drop from ~120s to <1s.
- A single `--bumped-version` call per component replaces the previous _full changelog + bumped-version_ pair; release detection compares the bumped version against the latest tag with `sort -V`. Correctness is preserved — every `cliff.toml` config has `filter_commits = true`, so skipped commits don't contribute to either the changelog or the bump.
- The 15 nearly-identical per-component check steps are collapsed into a single step that calls a `check_component` shell helper. Per-component log sections are preserved via `::group::` blocks.

Net diff: 980 lines deleted, 186 added in `.github/workflows/release.yml`.

### How to test locally

The release workflow only runs on push to `main`, so the verification done here was:

- [x] YAML parses cleanly; `bash -n` on the helper script.
- [x] All 46 outputs the downstream `release-*` jobs need are defined; none are orphaned.
- [x] `actionlint` count is unchanged from `main` (no new error categories).
- [x] End-to-end execution of the bash helper across all 15 components locally: completes in ~35s, computes `should-release` / `next-version` values consistent with the previous workflow (no spurious bumps; new components without prior tags correctly resolve to their initial version).


.github/workflows/release.yml
```diff
@@ -83,940 +83,139 @@ jobs:
...
-              NEXT_VERSION=$(git cliff "${INCLUDE_PATHS[@]}" --config infra/linux-runner-image/cliff.toml --repository "." --bumped-version 2>/dev/null)
-            fi
-          else
-            if [ -n "$LATEST_VERSION" ]; then
-              NEXT_VERSION="${LATEST_VERSION#linux-runner-image@}"
+          set -euo pipefail
+          ANY=false
+
+          # Args: component tag_prefix tag_regex cliff_config initial_version [include_path ...]
+          check_component() {
```
pepicrft: Around 50 lines of bash sitting inside a YAML `run:` block, so we can't shellcheck it or run it locally to sanity-check a release before pushing. Worth considering a move to a mise task (e.g. `release:check`) driven by a small components config, as a follow-up.

pepicrft (approved):
LGTM on the perf fix, unblocking releases is the right call.

Two more observations for a follow-up (the relevant lines aren't in this PR's diff, so noting them here rather than inline):

1. The `git tag -l | grep -E '^X@...$' | sort -V | tail -n1` pattern is repeated 30 times across the file, once per Get release notes step and once per Update CHANGELOG.md step. The check-releases job already computes each component's latest tag but doesn't expose it as an output. Surfacing it once and reusing across release-* jobs would kill the other 29 copies.

2. check-releases already exports `app-next-version-number` pre-stripped (line 35), so the manual `${VERSION_NUMBER#app@}` strip in release-app (line 400), release-ios (line 479), and release-android (line 520) is redundant.

Happy to take those on separately if useful.

.github/workflows/release.yml
fortmarek: Moved the helper to [mise/tasks/release/check.sh](https://github.com/tuist/tuist/blob/77f01e3471/mise/tasks/release/check.sh) driven by [mise/tasks/release/components.json](https://github.com/tuist/tuist/blob/77f01e3471/mise/tasks/release/components.json). Workflow step is now a one-liner (`mise run release:check`); locally `mise run release:check [component...]` runs the same logic, optionally filtered to one or more components. shellcheck passes clean.

PR #10902, opened 2026-05-20
feat(server): make the GitHub webhook hot path leave Postgres alone

fortmarek:
## Summary

Pulls two PG round-trips out of the `/webhooks/github` request handler so a burst of inbound `workflow_job` deliveries can't saturate the Ecto pool and start returning 502 to GitHub (which we observed during the [#10886](https://github.com/tuist/tuist/pull/10886) bring-up when ~30 deliveries hit at once).

Diagnosis traced the slow path to two synchronous Postgres calls every webhook makes before returning 200:

1. `VCS.list_github_app_installations_for_webhook/2` — uncached `Repo.all` for HMAC candidate-set lookup, runs before signature verification.
2. `Oban.insert/1` — even for `workflow_job.action` values the dispatch worker treats as `:ignored` (most notably `in_progress`, which is ~33 % of `workflow_job` traffic).

Both write through the same per-replica pool. With Ecto's 15 s checkout timeout, a burst that overflows the pool produces the exact 9 s upstream-timeout 502 we caught in GitHub's hookshot delivery log.

## What changed

**Controller: skip the `Oban.insert` when the action isn't dispatchable.** `Tuist.Runners.Dispatch.handle_webhook/2` only persists `queued` and `completed`; every other action falls through its catch-all `:ignored` branch. The controller now mirrors that gate and 200s without enqueueing, removing the PG write for ~33 % of inbound `workflow_job` deliveries.

**VCS: cache `list_github_app_installations_for_webhook/2`.** 60 s TTL, keyed by `(installation_id, app_id)`. Steady-state webhook traffic skips the `Repo.all/1` entirely. The TTL is the sole invalidation mechanism: the only field a stale entry could mis-serve (`webhook_secret`) only rotates through `replace_github_app_installation/2`, whose manifest-re-registration flow takes longer than 60 s anyway. The other three mutators (`create`, `update`, `delete`) can't produce a stale-cache scenario worth instrumenting:

* `update_github_app_installation/2`'s changeset whitelists only `:html_url`, `:installation_id`, `:app_slug` — none affect HMAC verification.
* `delete_github_app_installation/1` removes a row, but GitHub stops dispatching to an uninstalled App, so any in-flight webhook is signed by the still-valid (about-to-be-deleted) secret.
* `create_github_app_installation/1` has no entry to evict — the first lookup is a cold miss by definition.

## Test plan

- [x] `mix test test/tuist/vcs_test.exs test/tuist_web/controllers/webhooks/github_controller_test.exs` — 125 tests, 0 failures
- [x] New cache-behaviour tests in `vcs_test.exs` cover: cache hit on repeated query, distinct cache slots for `(iid, nil)` vs `(iid, aid)`, no-op when both filters are nil
- [x] New controller tests in `github_controller_test.exs` cover: `in_progress` → no enqueue, unknown action → no enqueue, `completed` → enqueue, plus the existing `queued` and `no installation.id` paths
- [x] `mix credo` clean on touched files
- [x] `mix format --check-formatted` clean

### DIFF
--- a/server/lib/tuist/vcs.ex
+++ b/server/lib/tuist/vcs.ex
@@ -34,12 +34,11 @@ defmodule Tuist.VCS do
   # Per-webhook lookup cache: every inbound GitHub webhook calls
   # `list_github_app_installations_for_webhook/2` once before HMAC
   # verification, so memoising the result keeps that PG read off the
-  # request path under burst load. The TTL is short enough that an
-  # in-flight rotation of `webhook_secret` heals within a minute, and
-  # mutating callers (`update_github_app_installation` /
-  # `replace_github_app_installation` / `create_github_app_installation`
-  # / `delete_github_app_installation`) evict explicitly so the cache
-  # never carries a deleted or rewritten row past the next mutation.
+  # request path under burst load. The TTL is short enough that any
+  # install / uninstall / re-register flow heals within a minute,
+  # and the only field a stale entry could mis-serve (`webhook_secret`)
+  # is rotated through `replace_github_app_installation/2` whose flow
+  # already spans longer than the TTL.
   @webhook_installations_cache_ttl_ms 60_000
 
   @doc """
@@ -1305,11 +1304,10 @@ defmodule Tuist.VCS do
   this path before HMAC verification, and the underlying table only
   changes on App install / update / uninstall (rare relative to
   webhook traffic), so memoising the lookup removes one Postgres
-  round-trip per webhook on the hot path. Mutating functions
-  (`update_github_app_installation`, `replace_github_app_installation`,
-  `create_github_app_installation`, `delete_github_app_installation`)
-  call `invalidate_webhook_installation_cache/1` so the cache never
-  outlives the row it points at by more than the TTL window.
+  round-trip per webhook on the hot path. The TTL is the sole
+  invalidation mechanism: mutations don't bust the cache because
+  the only field a stale entry could mis-serve (`webhook_secret`)
+  rotates through a flow that takes longer than the TTL anyway.
   """
   def list_github_app_installations_for_webhook(installation_id, app_id) do
     installation_id = maybe_to_string(installation_id)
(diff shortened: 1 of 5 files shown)


server/lib/tuist/vcs.ex
```diff
@@ -1312,6 +1345,24 @@ defmodule Tuist.VCS do
...
   end
 
+  @doc """
+  Name of the Cachex instance backing the webhook installations cache.
+  Returns the application-wide `:tuist` cache in prod; tests can
+  `stub/3` this to point at a per-test Cachex started via
+  `start_supervised!` so cache state never leaks across parallel
+  cases.
+  """
+  def cache_name, do: :tuist
```
pepicrft: I'd lean on passing this as `opts` in the function that needs it. Feels cleaner and we can skip the additional comment explaining why this is needed.

server/lib/tuist/vcs.ex
```diff
@@ -1312,6 +1345,24 @@ defmodule Tuist.VCS do
...
+
+  # Cachex's key serialiser (`Tuist.KeyValueStore.cache_key/1`) joins
+  # list elements with `-`, so the key needs to be a flat list of
+  # `to_string/1`-friendly terms. The two `_` slots reserve dedicated
+  # positions for `installation_id` and `app_id` (empty string when
+  # absent) so a lookup keyed by just one of them can't collide with a
+  # lookup keyed by the other.
```
pepicrft: A coding agent should infer this by looking at the implementation of the function.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
```diff
@@ -226,6 +226,16 @@ defmodule TuistWeb.Webhooks.GitHubController do
...
+  # `Tuist.Runners.Dispatch.handle_webhook/2` only branches on these
+  # two — every other action (notably `in_progress`, ~33 % of
+  # `workflow_job` traffic) falls through to the catch-all
+  # `:ignored` branch in the worker. Short-circuiting here removes
+  # the Oban.insert (and its Postgres write) for those events,
+  # which under burst load was costing us ~one PG checkout per
+  # webhook for work the worker was going to discard anyway.
+  @dispatchable_workflow_job_actions ~w(queued completed)
+
   defp handle_workflow_job(conn, params) do
```
pepicrft, line 239: As a follow-up improvement, you can batch the inserts using something like Broadway, such that we do 1 query for multiple job inserts.

server/test/tuist/vcs_test.exs
```diff
@@ -3658,4 +3658,63 @@ defmodule Tuist.VCSTest do
...
     end
   end
+
+  describe "list_github_app_installations_for_webhook/2 caching" do
+    setup do
+      cache = :"vcs_webhook_#{System.unique_integer([:positive])}"
+      start_supervised!({Cachex, name: cache})
+      stub(VCS, :cache_name, fn -> cache end)
+      :ok
+    end
```
pepicrft, line 3667: This pattern is ugly and hard to reason about for a future agent. I'd return the cache in a map here, `%{cache: cache}` and then use it when calling `VCS.create_github_app_installation(%{}, cache: cache)`

pepicrft (approved):

server/lib/tuist/vcs.ex
fortmarek: Done in d67c935 — dropped `cache_name/0` and the comment, added a `:cache` opt on `list_github_app_installations_for_webhook/3` that the test passes through directly.

server/lib/tuist/vcs.ex
fortmarek: Comment removed in d67c935.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
fortmarek: Agreed. Filing as a follow-up — out of scope here, but the `:webhooks` queue is a good fit for Broadway batching once we hit the next throughput ceiling.

server/test/tuist/vcs_test.exs
fortmarek: Done in d67c935 — setup now returns `%{cache: cache}` and each test destructures it and passes `cache: cache` through to the VCS call. No more `stub(VCS, :cache_name, ...)`.

PR #10916, opened 2026-05-21
fix(infra): unblock HPA and stop chronic ingress 502s

fortmarek:
## What's broken

Grafana flagged two production-impacting issues on `tuist-tuist-server`:

1. **HPA broken for ~15 days.** `FailedGetResourceMetric` had fired ~90,000 times (every ~15s). `pods.metrics.k8s.io` was unavailable, so the autoscaler could not read CPU/memory and was stuck at 2 pods with no scale-out relief during load spikes.
2. **Chronic `protocol_error` baseline at ~0.18 rps.** GitHub webhooks intermittently received 502s; the dropped requests never appeared in application logs because Bandit terminated the TCP connection before Phoenix could assign a `request_id`.

## Root causes

**HPA:** `metrics-server` was never installed on the workload clusters. Nothing in the platform stack provided the `metrics.k8s.io` aggregated API.

**`protocol_error`:** classic backend-shorter-than-proxy keepalive race. Bandit's `thousand_island_options.read_timeout` is 15s (set deliberately in [b427dde5c7](https://github.com/tuist/tuist/commit/b427dde5c7) to fail straggler connections fast). ingress-nginx's default `upstream-keepalive-timeout` is 60s, so nginx repeatedly pulled a "warm" connection from its keepalive pool that the backend had already FIN'd. The next request hit a closed socket → ingress returned 502 → Bandit logged `reason=protocol_error`.

### `metrics-server`

Added as a chart dependency on `infra/helm/platform/`, enabled by default with `--kubelet-insecure-tls` (kubeadm kubelets serve a self-signed cert the apiserver CA cannot validate; CSR auto-approval is opt-in). Once the chart rolls out, the `v1beta1.metrics.k8s.io` APIService registers, HPAs see resource metrics, and `tuist-tuist-server` can scale between its existing 2–5 replica bounds again.

### ingress-nginx upstream keepalive

Added `upstream-keepalive-timeout: \"10\"` to the platform chart's ingress-nginx config so nginx closes idle upstream connections before Bandit does. The 15s Bandit timeout is left untouched, preserving the property from b427dde5c7 that straggler connections fail in 15s instead of 60s.

### Rollout plumbing

Wired a new `platform-install` job into `server-deployment.yml`, mirroring the existing `observability-install` and `tailscale-operator-install` pattern. The job runs `mise -C infra run k8s:install-platform` on every deploy so platform-chart edits self-heal onto existing clusters without a manual re-bootstrap. The script header already advertised it as \"Idempotent — safe to run on every deploy\"; this just wires it in.

The job derives `CLOUDFLARE_API_TOKEN` from the in-cluster `platform/cloudflare-api-token` Secret (the CI ServiceAccount is cluster-admin) and the cluster name from the existing release's `external-dns.txtOwnerId`, so external-dns ownership stays identical to bootstrap.

## Why not the obvious alternatives

- **\"Just bump Bandit's read_timeout back to 60s.\"** Reintroduces the bug b427dde5c7 fixed — CLI requests blocked for 60s waiting for stranded straggler connections to time out.
- **\"Install metrics-server as a separate one-off chart.\"** The platform chart already exists to bundle cluster-wide infrastructure (cert-manager, ingress-nginx, external-dns, ESO). metrics-server fits cleanly in the same boundary; a new wrapper chart would be duplicate scaffolding.
- **\"Patch the ingress-nginx ConfigMap directly with kubectl.\"** Drifts the live state from the chart-managed source of truth; the next platform reconcile would fight it.

## Validation

Locally:

- `helm dependency update infra/helm/platform` pulls metrics-server-3.12.2
- `helm lint infra/helm/platform -f values-hetzner.yaml` and `helm lint … -f values-kura-region.yaml` both pass
- `helm template …` renders the metrics-server Deployment with the correct args, the `v1beta1.metrics.k8s.io` APIService, and `upstream-keepalive-timeout: \"10\"` in the ingress-nginx ConfigMap
- `actionlint .github/workflows/server-deployment.yml` produces only the pre-existing self-hosted runner-label warnings shared with the other jobs

In production, the platform-install job will run as part of the canary → production cascade on merge; expect `kubectl get apiservice v1beta1.metrics.k8s.io` to flip to `Available=True` and the `protocol_error` baseline to fall to zero shortly after rollout.

### How to test locally

```bash
cd infra/helm/platform
helm dependency update
helm lint . -f values-hetzner.yaml
helm template platform . -f values-hetzner.yaml --namespace platform \
  | grep -E 'kind: APIService|upstream-keepalive-timeout|metrics-server-3.12.2'
```

### DIFF
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -315,7 +315,7 @@ jobs:
           ref: ${{ env.DEPLOY_SHA }}
       - uses: jdx/mise-action@v3.2.0
         with:
-          install_args: "helm kubectl jq"
+          install_args: "helm kubectl"
           cache: "false"
       - uses: 1password/install-cli-action@v2
       - name: Load kubeconfig from 1Password
@@ -333,24 +333,19 @@ jobs:
           echo "Reaching apiserver at $API"
           kubectl --request-timeout=10s get --raw /version >/dev/null
       - name: Run k8s:install-platform
-        # The script needs CLOUDFLARE_API_TOKEN and a cluster name. We avoid
-        # adding a new GitHub secret for the token by reading it back from
-        # the in-cluster Secret that the bootstrap script created (CI runs
-        # cluster-admin). The cluster name is taken from the existing
-        # platform release's stored values so the txtOwnerId external-dns
-        # uses on every reconcile stays identical to bootstrap.
+        # CLOUDFLARE_API_TOKEN is read from the in-cluster Secret the bootstrap
+        # script created (CI is cluster-admin) instead of being added as a
+        # second GitHub secret. ::add-mask:: registers the value with Actions
+        # so it's redacted in subsequent step output (GHA only auto-masks
+        # values it knows are secrets; values derived at runtime aren't).
+        env:
+          CLUSTER_NAME: ${{ inputs.environment == 'production' && 'tuist' || format('tuist-{0}', inputs.environment) }}
         run: |
           set -euo pipefail
           CLOUDFLARE_API_TOKEN="$(kubectl -n platform get secret cloudflare-api-token \
             -o jsonpath='{.data.api-token}' | base64 -d)"
+          echo "::add-mask::$CLOUDFLARE_API_TOKEN"
           export CLOUDFLARE_API_TOKEN
-          OWNER="$(helm get values platform -n platform -o json \
-            | jq -r '."external-dns".txtOwnerId // empty')"
-          CLUSTER_NAME="${OWNER%-platform}"
-          if [ -z "$CLUSTER_NAME" ]; then
-            echo "ERROR: could not derive cluster name from existing platform release" >&2
-            exit 1
-          fi
           echo "Reconciling platform chart for cluster $CLUSTER_NAME"
           mise -C infra run k8s:install-platform "$KUBECONFIG" "$CLUSTER_NAME"
 
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -285,15 +285,87 @@ jobs:
             -f "$TAILSCALE_CHART_PATH/values-${{ inputs.environment }}.yaml" \
             --timeout 10m --wait
 
+  platform-install:
+    # Installs / upgrades the platform chart (cert-manager, ingress-nginx,
+    # external-dns, ESO, metrics-server) ahead of the server rollout. The
+    # bootstrap script lays this down once per cluster, but app-impacting
+    # platform settings — ingress-nginx upstream-keepalive-timeout,
+    # metrics-server presence — must self-heal on every deploy so existing
+    # clusters pick up chart edits without a manual re-bootstrap. Idempotent:
+    # steady-state finishes in under a minute when nothing has drifted.
+    #
+    # Shares the `server-deploy-<env>` concurrency lane with the deploy job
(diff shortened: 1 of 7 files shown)


infra/helm/platform/values.yaml
```diff
@@ -50,10 +50,16 @@ ingress-nginx:
...
+    #
+    # upstream-keepalive-timeout must stay below Bandit's
+    # thousand_island_options.read_timeout (15s in server/config/runtime.exs).
+    # Otherwise nginx reuses a connection the backend has already FIN'd,
+    # producing 502s with Bandit `reason=protocol_error`.
     config:
       use-forwarded-headers: "true"
       use-proxy-protocol: "true"
       compute-full-forwarded-for: "true"
+      upstream-keepalive-timeout: "10"
```
pepicrft, line 62: Validated the 15s ceiling at `server/config/runtime.exs:300` (`thousand_island_options.read_timeout: to_timeout(second: 15)`). 10s leaves a healthy gap for the FIN to land before nginx reuses the socket. Good fix.

infra/helm/platform/values.yaml
```diff
@@ -88,6 +94,22 @@ external-secrets:
...
     port: 9443
 
+# Backs the pods.metrics.k8s.io API the HPA reads. Without it, HPAs report
+# `FailedGetResourceMetric` and refuse to scale.
+# --kubelet-insecure-tls because kubeadm kubelets serve a self-signed cert
+# the kube-apiserver CA can't validate (CSR auto-approval is opt-in).
+metrics-server:
+  enabled: true
+  args:
+    - --kubelet-insecure-tls
```
pepicrft, line 104: `--kubelet-insecure-tls` is the right call for kubeadm. The alternative (CSR auto-approval) widens the trust path more than this does. The risk here is bounded to the in-cluster metrics-server → kubelet hop, not external.

.github/workflows/server-deployment.yml
```diff
@@ -285,15 +285,82 @@ jobs:
...
+        # second GitHub secret. ::add-mask:: registers the value with Actions
+        # so it's redacted in subsequent step output (GHA only auto-masks
+        # values it knows are secrets; values derived at runtime aren't).
+        env:
+          CLUSTER_NAME: ${{ inputs.environment == 'production' && 'tuist' || format('tuist-{0}', inputs.environment) }}
+        run: |
+          set -euo pipefail
+          CLOUDFLARE_API_TOKEN="$(kubectl -n platform get secret cloudflare-api-token \
+            -o jsonpath='{.data.api-token}' | base64 -d)"
+          echo "::add-mask::$CLOUDFLARE_API_TOKEN"
```
pepicrft, line 347: Checked the downstream path: `install-platform.sh` only flows the token through `kubectl create secret --from-literal=… | kubectl apply -f -` and `unset`s it after use. With `set -euo pipefail` and no `set -x`, the assignment-before-mask window is safe. Nice.

pepicrft (approved):
Read the full diff end-to-end and validated the two root-cause claims against the codebase. Both fixes match the diagnosis, the rollout plumbing is consistent with the existing observability-install / tailscale-operator-install pattern, and the Cloudflare-token handling is appropriately defensive. Leaving a few inline notes on the spots I checked. Ship it; watch `kubectl get apiservice v1beta1.metrics.k8s.io` flip Available=True and the `protocol_error` baseline collapse on canary.

PR #10925, opened 2026-05-21
[codex] fix(infra): stop Kura global endpoint reconciliation in regionals

pepicrft:
## What changed

This stops Kura regional controllers from reconciling Cloudflare global endpoints while keeping the existing Hetzner regional load balancer path intact.

The chart now exposes a dedicated `kuraGlobalEndpointsEnabled` flag, the controller accepts a matching `--reconcile-global-endpoints` argument, and the regional overlay disables both global endpoint reconciliation and the readiness requirement on global endpoints.

## Why this changed

Regional Kura deployments were getting stuck in `Deploying` because the controller still tried to reconcile Cloudflare global endpoints and treated failures there as blocking. We are currently hitting the Cloudflare pool origin quota, so new regional tenants can stay healthy at the pod level but never publish a ready status.

The obvious alternative was turning off Cloudflare load balancing entirely in the regional overlay, but that would also disable cleanup for tenants that already have global endpoints provisioned. The dedicated reconciliation flag is narrower: it stops creating or updating global endpoints in regionals while preserving destroy-path cleanup.

## Impact

Once this rolls out to the regional clusters, Kura tenants in those regions can become ready without being blocked by Cloudflare global endpoint reconciliation. Regional public and gRPC traffic still use the existing Hetzner load balancers.

## Root cause

The controller previously had only one Cloudflare switch. If Cloudflare load balancing was enabled at all, reconciliation always attempted to create or update global endpoints. In the current quota-constrained state, those calls fail and prevent status from progressing.

## Validation

- Rendered the regional Helm chart and confirmed the controller receives `--reconcile-global-endpoints=false` and `--require-global-endpoints=false`
- Confirmed the rendered manifest still wires the Cloudflare secret so cleanup remains available
- Ran `go test ./...` in `infra/kura-controller` with Go 1.25 in Docker

### DIFF
--- a/infra/helm/tuist/templates/kura-controller.yaml
+++ b/infra/helm/tuist/templates/kura-controller.yaml
@@ -172,6 +172,7 @@ spec:
             {{- end }}
             - --cloudflare-api-token=$(CLOUDFLARE_API_TOKEN)
             {{- end }}
+            - --reconcile-global-endpoints={{ ternary "true" "false" .Values.kuraGlobalEndpointsEnabled }}
             - --require-global-endpoints={{ ternary "true" "false" .Values.kuraGlobalEndpointsRequired }}
           ports:
             - name: metrics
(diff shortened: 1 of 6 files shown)


pepicrft:
Closing as duplicate of #10923. The same change is already merged into main.

PR #10927, opened 2026-05-21
refactor(capi-scaleway): drop auto-order + terminate paths, pool only

fortmarek:
### Purpose

Every managed environment already runs the CAPI Scaleway provider with `Spec.AdoptPoolPrefix` set, and the operator-owned pre-order workflow is the only sane path on Scaleway Apple Silicon (24h Apple licensing floor, flaky inventory). The companion `tuist.dev/release-policy: terminate` opt-out that #10922 added was extra surface nobody used, plus a `shouldReleaseToPool` decision that had to be re-derived on every delete. Collapse it to the pool path.

### What changed

- **`acquireServer`** always calls `AdoptByPrefix`. An empty `Spec.AdoptPoolPrefix` now surfaces a `MissingAdoptPoolPrefix` condition + event instead of silently falling through to `CreateServer`.
- **`reconcileDelete`** always calls `ReleaseToPool`. The `ReleasePolicyAnnotation` / `ReleasePolicyPool` / `ReleasePolicyTerminate` constants and the `shouldReleaseToPool` helper are gone. Operators retiring a broken host do so via the Scaleway console.
- **`scaleway.Client`** drops `CreateServer`, `DeleteServer`, `resolveOSID`, and `isPreconditionFailed`. The `AppleSiliconAPI` interface loses the four SDK methods nothing calls anymore (`CreateServer`, `DeleteServer`, `WaitForServer`, `ListOS`).
- **`Spec.AdoptPoolPrefix`** is now required: no `omitempty` on the type, `required: [adoptPoolPrefix]` in both CRDs, and chart templates emit the field unconditionally guarded by a `required` helper. Chart defaults flip to `"tuist-pool-"` so a greenfield install matches the managed envs.
- **Docs**: AGENTS.md, CRDs, `values.yaml`, `values-managed-production.yaml`, and the controller's surrounding prose stop referencing the removed paths. The detach-without-release runbook now points at `ReleaseToPool` rather than `DeleteServer`.
- **Tests**: removed the `TestShouldReleaseToPool_*`, `TestReconcileDelete_TerminateAnnotationFallsBackToDeleteServer`, `TestReconcileDelete_AutoOrderUsesDeleteServer`, and the four `TestDeleteServer_*` cases. The remaining delete-path test is `TestReconcileDelete_ReleasesToPool`.

### How to test locally

From the provider module:

\`\`\`bash
cd infra/cluster-api-provider-scaleway-applesilicon
go build ./...
go vet ./...
go test ./...
\`\`\`

Render the chart against each managed env to confirm the new \`adoptPoolPrefix\` field is emitted everywhere:

\`\`\`bash
helm template tuist infra/helm/tuist -f infra/helm/tuist/values-managed-canary.yaml | grep adoptPoolPrefix
helm template tuist infra/helm/tuist -f infra/helm/tuist/values-managed-staging.yaml | grep adoptPoolPrefix
helm template tuist infra/helm/tuist -f infra/helm/tuist/values-managed-production.yaml | grep adoptPoolPrefix
\`\`\`

### DIFF
--- a/infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
+++ b/infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
@@ -107,6 +107,7 @@ type ScalewayAppleSiliconMachineSpec struct {
 	// scan once Scaleway flips it back to `Delivered + Ready`.
 	// Physical destruction is operator-owned via the Scaleway
 	// console so the 24h billing floor doesn't leak into deploy flows.
+	// +kubebuilder:validation:MinLength=1
 	AdoptPoolPrefix string `json:"adoptPoolPrefix"`
 }
 
--- a/infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
+++ b/infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
@@ -103,7 +103,7 @@ type ScalewayAppleSiliconMachineSpec struct {
 	// the host back into the pool namespace (with a fresh UUID
 	// suffix) and triggers a Scaleway OS reinstall, then drops the
 	// k8s-side state. The host stays alive, is reset to factory-
-	// default state, and becomes eligible for the next AdoptByPrefix
+	// default state, and becomes eligible for the next AdoptFromPool
 	// scan once Scaleway flips it back to `Delivered + Ready`.
 	// Physical destruction is operator-owned via the Scaleway
 	// console so the 24h billing floor doesn't leak into deploy flows.
--- a/infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
+++ b/infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
@@ -76,13 +76,12 @@ type ScalewayAppleSiliconMachineSpec struct {
 	// +optional
 	HostMemoryMB int `json:"hostMemoryMB,omitempty"`
 
-	// AdoptPoolPrefix switches the controller from `Scaleway CreateServer`
-	// (auto-order) to "claim a pre-ordered host whose Scaleway-side
-	// name starts with this prefix." Recommended for the customer-
-	// runner fleet because Scaleway Mac mini inventory is frequently
-	// out of stock and Apple's 24h licensing floor makes speculative
-	// auto-ordering expensive — pre-ordering days in advance and
-	// letting the controller adopt is the operationally sane path.
+	// AdoptPoolPrefix is the Scaleway-side name prefix the controller
+	// scans when claiming a Mac mini for this Machine. Required:
+	// the controller has no auto-order path. Operators pre-order
+	// capacity in the Scaleway console because Mac mini inventory is
+	// frequently out of stock and Apple's 24h licensing floor makes
+	// speculative ordering expensive.
 	//
 	// Operator workflow:
 	//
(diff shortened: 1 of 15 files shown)


infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
```diff
@@ -691,40 +677,22 @@ func (r *ScalewayAppleSiliconMachineReconciler) reconcileDelete(
...
-					"Scaleway DeleteServer: %v (will retry)", err)
-				return ctrl.Result{RequeueAfter: 60 * time.Second}, nil
-			}
-			machine.Status.ServerID = ""
-			r.Recorder.Eventf(machine, corev1.EventTypeNormal, "Deleted",
-				"Scaleway server released")
+		r.Recorder.Eventf(machine, corev1.EventTypeNormal, "Releasing",
+			"Returning Scaleway server %s to pool %q (with reinstall)",
+			machine.Status.ServerID, machine.Spec.AdoptPoolPrefix)
+		if err := r.ScalewayClient.ReleaseToPool(ctx, machine.Status.ServerID, machine.Spec.Zone, machine.Spec.AdoptPoolPrefix); err != nil {
```
pepicrft: Validated this edge case locally. Older `macosFleet` objects could be created without `adoptPoolPrefix` because the previous chart only rendered the field when the value was set, and canary did not set it. For those legacy objects this unconditional `ReleaseToPool` path loops on empty `poolPrefix` during deletion. I do not see it as blocking for this refactor, but it looks worth a small compatibility follow-up.

infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
```diff
@@ -98,27 +97,17 @@ type ScalewayAppleSiliconMachineSpec struct {
...
-	AdoptPoolPrefix string `json:"adoptPoolPrefix,omitempty"`
+	// Delete semantics: on Machine deletion the controller renames
+	// the host back into the pool namespace (with a fresh UUID
+	// suffix) and triggers a Scaleway OS reinstall, then drops the
+	// k8s-side state. The host stays alive, is reset to factory-
+	// default state, and becomes eligible for the next AdoptFromPool
+	// scan once Scaleway flips it back to `Delivered + Ready`.
+	// Physical destruction is operator-owned via the Scaleway
+	// console so the 24h billing floor doesn't leak into deploy flows.
+	AdoptPoolPrefix string `json:"adoptPoolPrefix"`
```
pepicrft, line 111: Validated this too. The new contract says `adoptPoolPrefix` is required, but the schema still accepts `""` because it only requires presence. I verified that with a server-side dry run against the generated CRD: omitting the field is rejected, `adoptPoolPrefix: ""` is accepted. Non-blocking here, but I would tighten the validation so the API fails fast.

pepicrft (approved):
Validated the two notes locally. They are real, but I do not see them as blocking for this refactor. Approving, and I would follow up with a small compatibility and validation patch.

infra/cluster-api-provider-scaleway-applesilicon/controllers/scalewayapplesiliconmachine_controller.go
fortmarek: Fixed in 5c8dac62ff: `reconcileDelete` now skips the Scaleway release stage when `AdoptPoolPrefix` is empty, surfaces a `ReleaseSkipped` warning event, and proceeds with the rest of teardown so legacy CRs unblock. Added a controller-level test pinning the path.

infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
fortmarek: Fixed in 5c8dac62ff: added `+kubebuilder:validation:MinLength=1` on the Go type and mirrored `minLength: 1` into both CRD YAMLs, so the apiserver now rejects `adoptPoolPrefix: ""` at admission time.

PR #10940, opened 2026-05-25
feat(server): add korean specific rule for ai translation

2sem:
Resolves <https://github.com/tuist/tuist/issues/YYY>

> Add korean translation rule for seamless korean texts

### How to test locally

> i don't know how to test locally this translation


2sem:
@pepicrft 🙏

PR #10951, opened 2026-05-26
fix(kura): reduce HTTP request metric cardinality

pepicrft:
## Summary
- Route normalization already collapses scanner traffic to `/_unmatched`, but `kura_http_requests_total` still multiplied route and status series by `method` and `client_country`, which kept Grafana cardinality higher than needed.
- Keep `kura_http_requests_total` limited to route and status, and move country attribution into a dedicated `kura_http_requests_by_country_total` metric so the geography dashboard still works without paying the route-by-country cross-product.
- Skip probe, internal, and unmatched traffic in the country metric, and update the Kura dashboard and docs to use the new metric shape.

## Testing
- `mise exec -- cargo fmt`
- `mise exec -- cargo test render_includes_recorded_metrics`
- `mise exec -- cargo test unknown_paths_use_a_stable_unmatched_route_metric_label`
- `mise exec -- cargo clippy --all-targets -- -D warnings`


pepicrft:
Superseded by #11025. I'll open a smaller follow-up for the remaining method-label cardinality reduction.

PR #10955, opened 2026-05-26
fix(server): send large test case lookups as multipart

pepicrft:
## Summary
- Fix the self-hosted test ingestion path where large existing-test-case lookups can exceed request header limits.
- Keep the lookup on Ecto primitives and pass `multipart: true` to `ClickHouseRepo.all/2`, so `ecto_ch` keeps handling UUID, boolean, datetime, and array decoding.
- Add a regression test that builds a 5,001-test-case payload and asserts the lookup uses multipart transport for both chunks.

## Testing
- `mix format lib/tuist/tests.ex test/tuist/tests_test.exs`
- `MIX_ENV=test mix compile`
- `mix test test/tuist/tests_test.exs:1175` (blocked during test DB setup: `20240415140351_create_users_auth_tables.exs` tries to create `users_tokens` before a `users` table exists, raising `ERROR 42P01 (undefined_table) relation "users" does not exist`)

### DIFF
--- a/server/lib/tuist/tests.ex
+++ b/server/lib/tuist/tests.ex
@@ -784,16 +784,9 @@ defmodule Tuist.Tests do
     |> Enum.uniq()
   end
 
-  # Batch size for `id IN (...)` lookups. Each UUID is ~38 bytes encoded in the
-  # SQL body, so 5_000 IDs is ~190 KB, well below ClickHouse's default
-  # `max_query_size` of 256 KB even with surrounding query text. Larger batches
-  # would risk a `TOO_LARGE_QUERY` rejection on big test reports.
-  #
-  # This lookup intentionally uses a raw multipart query instead of the Ecto
-  # `Repo.all` path. The ClickHouse client backing Ecto sends bound params as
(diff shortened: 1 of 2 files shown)


fortmarek (commented):
Requesting changes on two points:

1. The new regression test writes `nil` into `projects.default_branch`, but that column is `null: false`. CI is already failing with `null value in column "default_branch" ... violates not-null constraint`, so the test fails before it reaches the multipart lookup assertion. Could we leave the fixture default branch intact, or stub `Projects.get_project_by_id/1` if the intent is only to skip the branch-presence lookup?

2. Can we keep the lookup using Ecto primitives and pass `multipart: true` to `ClickHouseRepo.all/2` first instead of the raw query? For example:

   `existing_test_cases_chunk_query(project_id, ids_chunk) |> ClickHouseRepo.all(multipart: true)`

   That should preserve ecto_ch decoding for UUIDs, booleans, datetimes, and arrays while still moving the large params into the POST body, avoiding the manual `decode_clickhouse_*` layer.

fortmarek (approved):

PR #10958, opened 2026-05-27
fix(server): attach existing Tuist users on SCIM POST instead of returning 409

fortmarek:
## What changed

`Tuist.SCIM.provision_user/2` no longer returns `{:error, :email_taken}` (HTTP 409) when an Okta SCIM `POST /Users` targets an email that matches an existing Tuist user not yet a member of the calling org. Instead the user is attached to the org with the requested role and receives an email notification so they can audit or leave if the attachment was unexpected.

- `server/lib/tuist/scim.ex` — `provisionable_user/1` now returns `{:ok, user, :existing | :created}` so the caller can distinguish "user we just created" from "existing user we just attached." `provision_user/2` was split into `apply_provision/4` + `attach_active_user/4` to keep cyclomatic complexity within credo's bounds. The notification is dispatched after the transaction commits.
- `server/lib/tuist/accounts/user_notifier.ex` — new `deliver_scim_organization_attachment/2`.
- `server/test/tuist/scim_test.exs` and `server/test/tuist_web/controllers/scim/users_controller_test.exs` — the tests that codified the rejection (`"rejects existing users outside the organization"` / `"returns conflict when the email belongs to a user outside the organization"`) were updated to assert the new attach behavior at both the context and controller layers.
- `server/priv/gettext/dashboard_account.pot` — new gettext strings for the email body.

## Why

Customers who had members in Tuist before enabling SCIM hit 409 on every Okta `POST /Users` for a pre-existing email, blocking IdP rollouts. Recovery currently requires either pre-seeding Okta with internal Tuist user IDs (not practical) or manually attaching members through the dashboard. This was observed in production via repeated SCIM 409 bursts on a customer org immediately after enabling SCIM.

## Root cause

`provisionable_user/2` had a defensive branch: if `get_user_by_email/1` matched a user that didn't already `belongs_to_organization?`, it returned `{:error, :email_taken}`. The defense was meant to prevent one tenant's Okta from claiming another tenant's users. In practice it broke the *only* path that lets a customer roll SCIM out after their team already exists in Tuist.

## Why attach rather than the alternatives

Two alternatives were considered and rejected:

- **Attach only if the user is in no other orgs** — defensible, but breaks the contractor-in-two-orgs case which is real for enterprise customers.
- **Attach only if the user already authenticated via this org's SSO** — defeats the point of SCIM, which exists to provision *before* first login. Would not have fixed customers' current state.

The SCIM token is org-scoped (`account:scim:write` on an org-owned `AccountToken`), so the IdP is already authoritative for the calling org. The notification email gives the attached user a clear signal and an account-settings link to leave, matching how comparable SCIM service providers handle the same case.

## Impact

- Customers enabling SCIM after Tuist adoption see their existing members attached on the first Okta sync instead of repeated 409s.
- Existing Tuist users whose email is POSTed by a SCIM-enabled org receive an email so they can audit the attachment.
- No change for the "user already in org" and "brand-new user" paths — both remain idempotent / creating respectively.
- The `:email_taken` error path is still reachable from `replace_user`/`patch_user` (email-change uniqueness conflicts) and from the create-user race in `provisionable_user/1`, so the controller's 409 handler stays.

## How to test locally

1. `mise run install` if dependencies are not installed.
2. `mix test test/tuist/scim_test.exs test/tuist_web/controllers/scim/users_controller_test.exs` — 38 tests, 0 failures.
3. Manual flow: create an org with a SCIM token via `Tuist.SCIM.create_token/2`, create a standalone Tuist user, then `POST /scim/v2/Users` with that user's email. Pre-fix: HTTP 409. Post-fix: HTTP 201, user attached, notification email sent.


server/lib/tuist/scim/workers/attachment_notifier_worker.ex
```diff
@@ -0,0 +1,24 @@
...
+  alias Tuist.Accounts.User
+  alias Tuist.Accounts.UserNotifier
+  alias Tuist.Repo
+
+  @impl Oban.Worker
+  def perform(%Oban.Job{args: %{"user_id" => user_id, "organization_id" => organization_id}}) do
+    user = Repo.get!(User, user_id)
+    organization = Organization |> Repo.get!(organization_id) |> Repo.preload(:account)
+    UserNotifier.deliver_scim_organization_attachment!(user, organization)
+    :ok
```
pepicrft: Should we bind the result of `UserNotifier.deliver_scim_organization_attachment!` to the result of `perform`? That way we can lean on Oban mechanisms to do things like retry in scenarios like emails not being delivered due to network flakiness.

pepicrft (approved):

server/lib/tuist/scim/workers/attachment_notifier_worker.ex
fortmarek: Done in 7c64d8ca94 — dropped the bang on `deliver_scim_organization_attachment` so it now returns `{:ok, email} | {:error, reason}` via Bamboo's `Mailer.deliver_now/1`, and `perform/1` case-matches the tuple so Oban handles transient failures via its standard `{:error, _}` retry path. (Worth noting: the bang version was already getting retried because Oban catches raised exceptions, but the explicit return is the canonical worker shape — agreed.) Added a test that asserts `{:error, _}` is forwarded so the transient-failure retry can't silently regress.

server/lib/tuist/scim.ex
```diff
@@ -212,54 +215,72 @@ defmodule Tuist.SCIM do
...
+    {user, provenance} =
+      case provisionable_user(email) do
+        {:ok, %User{} = u, p} -> {u, p}
+        {:error, reason} -> Repo.rollback(reason)
       end
+
+    if active do
+      attach_active_user(user, organization, role, provenance)
+    else
+      :ok = remove_user_role_from_organization(user, organization)
```
pepicrft: This makes `POST /Users` inconsistent for `active: false` when the email already belongs to an existing Tuist user outside the org. We return success here without ever attaching the user, so the controller still answers `201` and emits a `Location`, but `GET /Users/:id` immediately 404s because membership was never created. I think we should either keep returning a conflict for this case, or attach first and then deactivate so the returned SCIM resource actually exists.

server/lib/tuist/scim.ex
```diff
@@ -212,54 +215,72 @@ defmodule Tuist.SCIM do
...
-      {:ok, user} -> {:ok, Repo.preload(user, :account)}
-      {:error, reason} -> {:error, reason}
+  end
+
+  defp attach_active_user(user, organization, role, provenance) do
+    if provenance == :existing and not Accounts.belongs_to_organization?(user, organization) do
+      enqueue_attachment_notification(user, organization)
     end
+
+    :ok = Accounts.add_user_to_organization(user, organization, role: role)
```
pepicrft, line 276: This opens a race for existing users. Two concurrent SCIM POSTs for the same email can both pass the membership check above and then both call `add_user_to_organization/3`. That helper still does a read then insert, and the only uniqueness we enforce here is on `users_roles.role_id`, so we can end up with duplicate org memberships for the same user and organization. Once that happens, the later `Repo.one(...)` lookups in `get_user_role_in_organization/2` and `organization_role_membership/2` start raising `Ecto.MultipleResultsError`.

server/lib/tuist/scim.ex
fortmarek: Fixed in 297a0f2cc4. `active: false` for an existing user outside the org now rolls back with `:email_taken` so the controller answers 409 instead of 201+Location. Considered "attach then deactivate" — but Tuist memberships are binary (no inactive-member representation), so the end state would still be a non-member and GET would still 404. Rejecting forces the IdP to use PATCH, which is the verb actually meant for deactivation. Added a context test and a controller test that confirms the 409 + uniqueness scimType.

server/lib/tuist/scim.ex
fortmarek: Fixed in 297a0f2cc4. Added a Postgres advisory xact lock keyed on `(user_id, organization_id)` at the start of `attach_active_user/4`, so concurrent SCIM POSTs for the same email serialize through the membership check + insert. Second caller sees the first commit and `add_user_to_organization/3` is a no-op. Lock releases at transaction end. (Underlying `add_user_to_organization` race still exists for non-SCIM callers — happy to file a follow-up to push the lock into the helper.)

PR #10961, opened 2026-05-27
Add negative test run branch filter

pepicrft:
Resolves N/A

Adds support for excluding test runs by branch substring so merge queue runs such as `gh-readonly-queue` can be filtered out.

- Adds a per-filter operator override in Noora and enables `does not contain` for the test-runs branch filter.
- Maps the dashboard operator to Flop's `:not_ilike` filter for case-insensitive branch exclusion.
- Adds `git_branch_not_contains` to the test-runs API and covers the context, API, and LiveView behavior with regression tests.

### How to test locally

- `mix deps.get` from `noora/`
- `mix deps.get` from `server/`
- `mix format lib/noora/filter.ex` from `noora/`
- `mix format lib/tuist_web/controllers/api/tests_controller.ex lib/tuist_web/live/test_runs_live.ex test/tuist/tests_test.exs test/tuist_web/controllers/api/tests_controller_test.exs test/tuist_web/live/test_runs_live_test.exs` from `server/`
- `MIX_ENV=test mix ecto.reset` from `server/`
- `MIX_ENV=test mix test test/tuist/tests_test.exs` from `server/`
- `MIX_ENV=test mix test test/tuist_web/controllers/api/tests_controller_test.exs` from `server/`
- `MIX_ENV=test mix test test/tuist_web/live/test_runs_live_test.exs` from `server/`


noora/lib/noora/filter.ex
```diff
@@ -17,12 +17,13 @@ defmodule Noora.Filter do
...
   - `:options` - List of available options (for `:option` type only)
   - `:options_display_names` - Map of option values to display names
+  - `:operators` - Optional operator override list for this filter
   - `:operator` - Comparison operator (e.g., `:==`, `:=~`, `:<`)
   - `:value` - Current filter value
 
   ### Filter Types and Operators
 
-  - **Text filters** (`:text`) - Support operators: `:==` (is), `:=~` (contains)
+  - **Text filters** (`:text`) - Default operators: `:==` (is), `:=~` (contains). Individual filters can override this list.
```
fortmarek: any reason why not to include not equals by default? I'd consider doing that personally

pepicrft: Included this by default now. `noora/lib/noora/filter.ex` exposes `does not contain` for all text filters, and `server/lib/tuist_web/live/test_runs_live.ex` no longer needs a branch-specific operator override. The existing LiveView regression test still covers the branch filter behavior.

server/lib/tuist_web/controllers/api/tests_controller.ex
```diff
@@ -41,6 +41,11 @@ defmodule TuistWeb.API.TestsController do
         type: :string,
         description: "Filter by git branch."
       ],
+      git_branch_not_contains: [
```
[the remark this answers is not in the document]
pepicrft: Added the Product changelog entry in `server/priv/marketing/changelog/2026.05.27-test-run-branch-exclusion-filter.md` and included the provided screenshot at `server/priv/static/marketing/images/changelog/2026.05.27-test-run-branch-exclusion-filter.png`. I also ran `git diff --check`; the earlier focused server tests for the API, context, and LiveView filter behavior are still part of this PR validation.

fortmarek: I would avoid growing this as a one-off operator encoded in the parameter name. `git_branch_not_contains` is clear in isolation, but if we later need `scheme_not_contains`, `git_branch_contains`, `git_branch_not_equals`, etc., the API surface will fan out pretty quickly.

Stripe has a useful split here: normal list endpoints keep simple explicit params, while richer filtering moves to dedicated search endpoints with a documented query language like `query=-currency:"usd"` or `query=email~"amy"`. I would recommend either keeping this API to exact-match filters only, or adopting a Stripe-style search query language for test runs instead of adding a bespoke negative substring param. That keeps operators in one extensible grammar and avoids leaking implementation details like Flop operators into public API shape.

References: https://docs.stripe.com/api/customers/list and https://docs.stripe.com/search

fortmarek (approved):
We can go ahead with the current API shape, but I do wonder if the rich search query would be good to adopt now or soon-ish before the query API includes a ton of operator-value pairs (with more operators, the matrix could explode)

server/lib/tuist_web/controllers/api/tests_controller.ex
pepicrft: Changed the API shape to avoid the one-off `git_branch_not_contains` parameter. The test-runs list endpoint now has a single richer `query` parameter with Stripe-style expressions, so this use case is `query=-git_branch~"gh-readonly-queue"`. Existing exact-match parameters stay in place, and the query parser maps supported fields/operators to internal filters without exposing Flop operator names. Updated the API controller, controller tests, and changelog copy accordingly. Validation: `MIX_ENV=test mix test test/tuist_web/controllers/api/tests_controller_test.exs`.

PR #10965, opened 2026-05-27
feat(server): Support self-hosted Kura endpoints

pepicrft:
Resolves N/A

Adds self-hosted Kura endpoint support for on-premise Tuist deployments. The server can now read Kura endpoints from runtime configuration, the Tuist Helm chart can render those endpoints into the server pod, and the Kura Helm chart is included in Helm validation and release publishing.

The self-hosting docs now split the old installation page into a Control plane page, remove the legacy cache-node pages from the sidebar, point existing cache-node users to the deprecated source docs, and add a Kura page that explains why Kura is decentralized, how to deploy it with or without Kubernetes, and which runtime variables operators can configure.

### How to test locally

- `helm lint kura/ops/helm/kura`
- `helm template kura kura/ops/helm/kura > /tmp/kura-rendered.yaml && wc -l /tmp/kura-rendered.yaml`
- `helm package kura/ops/helm/kura --destination /tmp`
- `helm lint infra/helm/tuist --set 'server.kuraEndpointUrls[0]=http://kura:4000'`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release.yml"); YAML.load_file(".github/workflows/helm.yml"); puts "ok"'`
- `mix test test/tuist/environment_test.exs test/tuist/accounts_test.exs test/tuist/docs/redirects_test.exs` (fails during test database setup with `Postgrex.Error` because relation `users` does not exist)

### DIFF
--- a/server/lib/tuist/docs/redirects.ex
+++ b/server/lib/tuist/docs/redirects.ex
@@ -39,10 +39,10 @@ defmodule Tuist.Docs.Redirects do
     {:exact, "/tutorials/tuist/install", "/tutorials/xcode/create-a-generated-project"},
     {:exact, "/tutorials/tuist/create-project", "/tutorials/xcode/create-a-generated-project"},
     {:exact, "/tutorials/tuist/external-dependencies", "/tutorials/xcode/create-a-generated-project"},
-    {:exact, "/tutorials/tuist-cloud-tutorials", "/guides/server/self-host/install"},
-    {:exact, "/tutorials/tuist/enterprise-infrastructure-requirements", "/guides/server/self-host/install"},
-    {:exact, "/tutorials/tuist/enterprise-environment", "/guides/server/self-host/install"},
-    {:exact, "/tutorials/tuist/enterprise-deployment", "/guides/server/self-host/install"},
+    {:exact, "/tutorials/tuist-cloud-tutorials", "/guides/server/self-host/control-plane"},
+    {:exact, "/tutorials/tuist/enterprise-infrastructure-requirements", "/guides/server/self-host/control-plane"},
+    {:exact, "/tutorials/tuist/enterprise-environment", "/guides/server/self-host/control-plane"},
+    {:exact, "/tutorials/tuist/enterprise-deployment", "/guides/server/self-host/control-plane"},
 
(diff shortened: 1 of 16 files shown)


server/lib/tuist/docs/redirects.ex
```diff
@@ -39,10 +39,10 @@ defmodule Tuist.Docs.Redirects do
     {:exact, "/tutorials/tuist/install", "/tutorials/xcode/create-a-generated-project"},
     {:exact, "/tutorials/tuist/create-project", "/tutorials/xcode/create-a-generated-project"},
     {:exact, "/tutorials/tuist/external-dependencies", "/tutorials/xcode/create-a-generated-project"},
-    {:exact, "/tutorials/tuist-cloud-tutorials", "/guides/server/self-host/install"},
-    {:exact, "/tutorials/tuist/enterprise-infrastructure-requirements", "/guides/server/self-host/install"},
-    {:exact, "/tutorials/tuist/enterprise-environment", "/guides/server/self-host/install"},
-    {:exact, "/tutorials/tuist/enterprise-deployment", "/guides/server/self-host/install"},
+    {:exact, "/tutorials/tuist-cloud-tutorials", "/guides/server/self-host/control-plane"},
```
pepicrft, line 42: @fortmarek I went with "Control plane" here for the main server, cause "Server" seems wrong (Kura is also a server). Let me know what do you think:

Image

server/priv/docs/en/guides/server/self-host/control-plane.md
```diff
@@ -14,6 +14,12 @@ We offer a self-hosted version of the Tuist server for organizations that requir
 >
 > Self-hosting Tuist requires a legally valid paid license. The on-premise version of Tuist is available only for organizations on the Enterprise plan. If you are interested in this version, please reach out to [contact@tuist.dev](mailto:contact@tuist.dev).
 
+> [!WARNING]
```
pepicrft, line 17: I'm removing it from our docs, but keeping a reference to the markdown files so people can still check out (we have on-premise customers that already deployed this)

fortmarek (approved):

PR #10982, opened 2026-05-28
feat(kura): co-locate runner-cache Kura nodes with the runner fleet

fortmarek:
## Summary

Adds the private-region concept on top of the umbrella-cluster topology (now on `main`) so a single-replica Kura node can be co-located with a customer's runners-as-a-service pool inside the cluster. The runner dispatch chain hands the runner a `cache_endpoint_url` it exports as `TUIST_CACHE_ENDPOINT`; the Gradle plugin / CLI honor that as a cache-endpoint override, so build cache traffic stays on the cluster's internal Service DNS instead of going through the public Kura ingress.

**Validated end-to-end in staging** with real Gradle build-cache traffic on a kata runner (see "Staging E2E results" below).

## What's in this PR

**Regions** (`Tuist.Kura.Regions`)
- `Regions.private?/1` + `Regions.selectable/0`. Private regions stay in `available/0` (so `create_server/1` accepts them) but are hidden from the customer picker — they are control-plane-provisioned, not customer-picked.
- Two private runner-cache regions in a `@private_region_specs` catalog (mirroring main's spec-driven `@managed_region_specs`):
  - `scw-fr-par-runners` — production target, pinned to the `kura-scw-fr-par` node pool, `scw-bssd`.
  - `hetzner-staging-runners` — staging-only, pinned to the umbrella cluster's `kura` node pool. Enabled in `values-managed-staging.yaml`.

**KuraInstance CRD**
- New `spec.private` boolean — the explicit marker for a no-public-endpoint instance. The controller's topology already does the right thing when public hosts are absent (`ClusterIP` Service, no Ingress, no Certificate). **The hand-maintained CRD yaml in `infra/helm/tuist/crds/` now carries the field; Helm never upgrades `crds/`, so each environment needs a one-time `kubectl apply` of the CRD (done on staging).**

**Activation** (`Tuist.Kura.activate_server/2`)
- Private regions skip the public DNS + HTTPS `/up` probe and the `account_cache_endpoints` mirror (CLI-facing; developer machines can't reach the in-cluster endpoint).
- `Provisioner.public_url/3` interpolates `private_url_template` — e.g. `http://kura-tuist-staging.kura.svc.cluster.local:4000`.

**Lifecycle** (`Tuist.Kura.RunnerCache`, runs in the reconciler tick)
- Identity rule: an account with ≥1 Runner Profile and the `:runners` feature flag enabled has exactly one non-destroyed node in the active private region; accounts without profiles (or flag off) have none. (Rebased onto main's profile+flag model after `accounts.runner_max_concurrent` was dropped.)
- Self-heals: failed nodes that never had a successful deployment are retried each tick — a transient apiserver error or missing CRD field can't strand an account.

**Runner routing**
- `Tuist.Kura.runner_cache_endpoint_url/1`: in-cluster URL for the account's active private node, or `nil`.
- `RunnersController.dispatch` returns `cache_endpoint_url` in the claim response.
- Linux runner image: split-container shape stages the URL at `<JIT_PATH>.cache-endpoint`; `run-job.sh` exports `TUIST_CACHE_ENDPOINT` before `exec`. Rollout-bridge + macOS images export in place.
- Cache routing is a soft optimization — a staging failure logs and falls back to the CLI's default resolution.

**NetworkPolicy**
- Runner-namespace egress carve-out to the `kura` namespace on 4000/50051.
- **Fixes a dispatch-egress livelock**: the policy's idle-only scoping (`runner-pool-owner DoesNotExist`) revoked dispatch egress mid-claim — Cilium recomputes the identity when the server stamps the owner label, and when revocation won the race the claim response was lost, leaving the poller stuck retrying against `POLICY_DENIED` forever (observed via Hubble; both staging warm pollers were zombies, so every claim fell through to other capacity). Dispatch egress now selects all runner Pods; the audience-scoped SA token (poller-only) is the access control.

## Staging E2E results

<details>
<summary><b>Reference: the smoke workflow that produced these numbers</b> (<code>kura-runner-cache-routing-smoke.yml</code> — E2E scaffolding, intentionally not committed; removed in 410112d10a. Durable benchmark tooling lives in tuist/cache-benchmark#2.)</summary>

```yaml
name: Kura Runner Cache Routing Smoke

# A/B measurement of the runner-local Kura build cache using REAL
# Gradle build-cache traffic, exercising this PR's product flow
# end-to-end (fire: runner image pinned, node active):
#
#   1. The RunnerCache reconciler provisions a private (in-cluster
#      only) Kura node for every account with Runner Profiles.
#   2. Runner dispatch includes `cache_endpoint_url` in the claim
#      response once that node is active.
#   3. The runner image (`run-job.sh`) exports it as
#      `TUIST_CACHE_ENDPOINT`, which the Tuist Gradle plugin honours
#      (TuistConfigReader.kt) for every remote build-cache request.
#
# Phases:
#   * routed   — the build uses the dispatch-provided endpoint as-is
#                (the PR's actual routing; in-cluster Service DNS).
#   * baseline — `TUIST_CACHE_ENDPOINT` is unset, so the plugin falls
#                back to its default cache resolution through
#                staging.tuist.dev (the status quo this PR improves).
#
# Each phase runs `ITERATIONS` cold (cache-miss: all local caches
# purged, artifacts pulled from the remote) and warm (cache-hit)
# builds of the Tuist Gradle plugin and reports wall-clock stats.
#
# Auth: `tuist auth login` in CI uses GitHub OIDC; staging maps the
# repo claim (tuist/tuist) to the staging `tuist/tuist` project via
# its `vcs_connections` row. Requires `id-token: write`.

on:
  workflow_dispatch:
    inputs:
      project_handle:
        description: Staging project handle the build is attributed to
        required: false
        default: tuist/tuist
        type: string
      cache_endpoint_override:
        description: Optional explicit cache endpoint for the routed phase (defaults to the dispatch-provided TUIST_CACHE_ENDPOINT)
        required: false
        type: string
      iterations:
        description: Repetitions per phase (each is a clean build that round-trips Gradle's cache)
        required: false
        default: '3'
        type: string
  push:
    branches:
      - feat/kura-scaleway-runner-cache
    paths:
      - .github/workflows/kura-runner-cache-routing-smoke.yml

permissions:
  contents: read
  # `tuist auth login` detects CI and mints a GitHub OIDC id-token to
  # exchange with staging.tuist.dev for an access token.
  id-token: write

env:
  MISE_VERSION: "2026.4.18"
  MISE_GITHUB_TOKEN: ${{ github.token }}
  MISE_TASK_RUN_AUTO_INSTALL: 0
  MISE_GITHUB_ATTESTATIONS: 0
  # Force the CLI + Gradle plugin to talk to staging, not the local
  # acceptance-test server some runner shapes bake in via
  # TUIST_SERVER_URL.
  TUIST_URL: https://staging.tuist.dev
  PROJECT_HANDLE: ${{ inputs.project_handle || 'tuist/tuist' }}
  CACHE_ENDPOINT_OVERRIDE: ${{ inputs.cache_endpoint_override || '' }}
  ITERATIONS: ${{ inputs.iterations || '3' }}

jobs:
  routing-smoke:
    # MUST be the staging-scoped profile label, not a `shape-*` label.
    # Both production and staging receive the tuist/tuist webhooks and
    # advertise identical `shape-linux-*` dispatch labels (same chart),
    # so a shape-labeled job races both environments' pollers and prod's
    # warm fleet wins — the job then runs in the production cluster
    # where staging's in-cluster Kura Service doesn't exist (NXDOMAIN /
    # dead ClusterIPs). `tuist-staging-<profile>` labels resolve only on
    # staging, pinning the job to staging kata runners.
    runs-on: tuist-staging-linux
    timeout-minutes: 30
    steps:
      - name: Resolve cache endpoint
        shell: bash {0}
        run: |
          # `TUIST_CACHE_ENDPOINT` is exported by the runner image's
          # run-job.sh when the dispatch response carried
          # `cache_endpoint_url` — i.e. when this account's private
          # runner-cache Kura node is active. That inherited value IS
          # the system under test.
          echo "pod(hostname)=$(hostname) at $(date -u +%FT%TZ)"
          endpoint="${CACHE_ENDPOINT_OVERRIDE:-${TUIST_CACHE_ENDPOINT:-}}"
          if [ -z "$endpoint" ]; then
            echo "::error::no cache endpoint: dispatch did not include cache_endpoint_url (private Kura node not active yet?) and no override input was provided"
            exit 1
          fi
          echo "routed endpoint: $endpoint"
          echo "ROUTED_ENDPOINT=$endpoint" >> "$GITHUB_ENV"
          code=$(curl -sS -o /dev/null -w '%{http_code}' --max-time 8 "$endpoint/up" 2>/dev/null) || code=000
          if [ "$code" = "000" ]; then
            echo "::error::routed endpoint unreachable: $endpoint/up"
            exit 1
          fi
          echo "reachability: $endpoint/up -> HTTP $code"

      - uses: actions/checkout@v4

      - uses: jdx/mise-action@v4.0.1
        with:
          version: ${{ env.MISE_VERSION }}
          install_args: "java gradle tuist"
          cache: "true"

      - uses: actions/setup-java@v4
        with:
          distribution: temurin
          java-version: '21'

      - name: Show environment
        run: |
          set -euo pipefail
          echo "Hostname:   $(hostname)"
          echo "Kernel:     $(uname -a)"
          echo "Project:    $PROJECT_HANDLE"
          echo "Server:     $TUIST_URL"
          echo "Routed URL: $ROUTED_ENDPOINT"
          echo "Iterations: $ITERATIONS"

      - name: Authenticate with staging Tuist
        run: |
          # CI mode → GitHub OIDC: the CLI requests an id-token from
          # the workflow and exchanges it with staging.tuist.dev for an
          # access token scoped to the projects connected to this repo
          # (staging project tuist/tuist). The token lands in the local
          # credential store keyed by TUIST_URL, where the Gradle
          # plugin's TokenProvider picks it up. Clear the acceptance
          # shape's localhost server env first so the token is keyed
          # against staging.
          unset TUIST_SERVER_URL TUIST_HOSTED
          if ! tuist auth login; then
            echo "::group::tuist session logs"
            cat "$HOME"/.local/state/tuist/sessions/*/logs.txt 2>/dev/null | tail -80
            echo "::endgroup::"
            exit 1
          fi

      - name: Configure Gradle project handle
        run: |
          set -euo pipefail
          # Attribute the build to the staging project the OIDC token
          # covers.
          cat > gradle/tuist.toml <<TOML
          project = "${PROJECT_HANDLE}"
          TOML

      - name: Run A/B benchmark
        shell: bash {0}
        run: |
          IT="$ITERATIONS"

          # Clears every layer Gradle would hit before talking to the
          # remote, so each "miss" build round-trips both the local
          # build-cache and the remote Tuist build-cache.
          purge_gradle_caches() {
            rm -rf gradle/build gradle/.gradle ~/.gradle/caches/build-cache-1 || true
          }

          # Runs `./gradlew assemble --build-cache` and prints total
          # wall-clock seconds. `assemble` not `build`: keeps the run
          # cache-bound rather than test-bound. For the routed phase
          # the dispatch-provided endpoint is exported; for baseline
          # the variable is removed so the plugin uses its default
          # cache resolution.
          # Build failures must fail the smoke run — a timing for a
          # build that errored (or never reached the cache) is worse
          # than no timing, because the job would go green while the
          # routed path is broken. measure_one runs in a command
          # substitution (subshell), so failures are recorded in a
          # sentinel file rather than a shell variable, and the step
          # exits non-zero after the timing table is printed.
          rm -f /tmp/gradle-build-failures
          measure_one() {
            local path=$1
            local kind=$2   # "miss" (purge first) or "hit" (warm cache)
            if [ "$kind" = "miss" ]; then purge_gradle_caches; fi
            local started ended status=0
            started=$(date +%s.%N)
            if [ "$path" = routed ]; then
              ( cd gradle && TUIST_CACHE_ENDPOINT="$ROUTED_ENDPOINT" \
                  ./gradlew --no-daemon -Dorg.gradle.caching=true \
                            --build-cache assemble \
                            >>/tmp/gradle-${path}-${kind}.log 2>&1 ) || status=$?
            else
              ( cd gradle && env -u TUIST_CACHE_ENDPOINT \
                  ./gradlew --no-daemon -Dorg.gradle.caching=true \
                            --build-cache assemble \
                            >>/tmp/gradle-${path}-${kind}.log 2>&1 ) || status=$?
            fi
            ended=$(date +%s.%N)
            if [ "$status" -ne 0 ]; then
              echo "${path}/${kind} exit=${status}" >> /tmp/gradle-build-failures
              echo "BUILD FAILED (${path}/${kind}, exit ${status}) — tail of log:" >&2
              tail -20 "/tmp/gradle-${path}-${kind}.log" >&2
            fi
            python3 -c "import sys; print(f'{float(sys.argv[1]) - float(sys.argv[2]):.3f}')" "$ended" "$started"
          }

          declare -A results
          for kind in miss hit; do
            for path in routed baseline; do
              for i in $(seq 1 "$IT"); do
                echo "::group::$path / $kind / iteration $i"
                t=$(measure_one "$path" "$kind")
                results["$path|$kind|$i"]=$t
                echo "wall-clock: ${t}s"
                tail -30 "/tmp/gradle-${path}-${kind}.log"
                echo "::endgroup::"
              done
            done
          done

          {
            echo "path,kind,iter,seconds"
            for key in "${!results[@]}"; do
              IFS='|' read -r path kind iter <<<"$key"
              echo "$path,$kind,$iter,${results[$key]}"
            done
          } | tee /tmp/results.csv

          if [ -s /tmp/gradle-build-failures ]; then
            echo "::error::Gradle build(s) failed — timings above are not trustworthy"
            cat /tmp/gradle-build-failures
            exit 1
          fi

      - name: Summarize
        run: |
          set -euo pipefail
          python3 - <<'PY'
          import csv, statistics
          rows = []
          with open('/tmp/results.csv') as f:
              for r in csv.DictReader(f):
                  rows.append({"path": r["path"], "kind": r["kind"], "iter": int(r["iter"]), "seconds": float(r["seconds"])})

          def col(kind, path):
              return [r["seconds"] for r in rows if r["kind"] == kind and r["path"] == path]

          def stats(vs):
              if not vs:
                  return None
              sv = sorted(vs)
              return {
                  "n": len(sv),
                  "min": sv[0],
                  "max": sv[-1],
                  "mean": statistics.mean(sv),
                  "median": statistics.median(sv),
              }

          print()
          print(f"{'phase':<10} {'path':<10} {'n':>3} {'min':>8} {'median':>8} {'mean':>8} {'max':>8}   (seconds wall-clock)")
          print("-" * 66)
          for kind in ["miss", "hit"]:
              for path in ["baseline", "routed"]:
                  s = stats(col(kind, path))
                  if s is None:
                      print(f"{kind:<10} {path:<10} {'-':>3}")
                      continue
                  print(f"{kind:<10} {path:<10} {s['n']:>3} {s['min']:>8.2f} {s['median']:>8.2f} {s['mean']:>8.2f} {s['max']:>8.2f}")
              print()

          for kind in ["miss", "hit"]:
              base = stats(col(kind, "baseline"))
              routed = stats(col(kind, "routed"))
              if base and routed and routed["median"] > 0:
                  saved = base["median"] - routed["median"]
                  ratio = base["median"] / routed["median"]
                  print(f"{kind}: routed median is {ratio:.2f}x the speed of baseline ({saved:+.2f}s saved per build)")
          PY

      - name: Upload Gradle logs
        if: always()
        uses: actions/upload-artifact@v4
        with:
          name: gradle-logs
          path: /tmp/gradle-*.log
          if-no-files-found: ignore
          retention-days: 7
```

</details>

Real `./gradlew assemble --build-cache` runs of the Gradle plugin on a staging kata runner (3 iterations per cell), claim → JIT → `TUIST_CACHE_ENDPOINT` exported by the runner image from the dispatch response:

| phase | path | median wall-clock |
|---|---|---|
| cache-miss (cold) | no remote cache | 31.7 s |
| cache-miss (cold) | **routed (in-cluster Kura)** | **7.45 s** |
| cache-hit (warm) | no remote cache | 11.5 s |
| cache-hit (warm) | **routed** | **6.0 s** |

Cold builds restore from the runner-local node **4.3× faster than recompiling** (24s saved per build). Caveat: the "baseline" degraded to no-remote-cache because staging's legacy public cache endpoints (`cache-*-staging.tuist.dev`) are down (separate issue — kamal app hung behind a healthy proxy/TLS). The apples-to-apples same-node comparison was completed afterwards — see the next section.

## Same-node path A/B: in-cluster vs public ingress (completed)

Measured from a pod on a staging bare-metal runner node under the runner fleet's exact NetworkPolicies, against the **same Kura node** over both paths (public leg via a temporary ingress, removed after the run).

**Request latency** (`/up`, 40 requests per cell):

| | public (LB + TLS + nginx) | private (in-cluster svc) | delta |
|---|---|---|---|
| cold connection p50 | 21.0 ms | **5.0 ms** | ~4× |
| cold connection p90 | 24.6 ms | **6.0 ms** | |
| warm (reused) | 1.8 ms | **0.5 ms** | ~3.5× |

~13.7 ms of the public cold path is TLS handshake alone. Cache clients open fresh connections per job step, so the cold row is representative.

**Bulk throughput** (200 MB artifact, module multipart flow, 3 iterations per cell):

| | public | private |
|---|---|---|
| upload (20 × 10 MB parts) | 40–42 MB/s | **88–99 MB/s** |
| download (single GET) | 107 MB/s | **107–108 MB/s** |

Interpretation:
- **Uploads pay ~2.3× through the gateway** — with request buffering off, every byte streams synchronously through the nginx workers (TLS decrypt + proxy), which are CPU-bound well before the link is.
- **Downloads are path-identical same-DC** (~1 Gbit ceiling on both), so the gateway is not a download bottleneck *when there is no WAN in the path*.
- **The WAN is the real divider**: this A/B has no WAN leg (runner and LB share a DC). External runners add ~20 ms RTT, where single-stream TCP becomes window-limited (roughly 12–50 MB/s at common window sizes) plus slow start on every fresh connection — that penalty applies to downloads too. Co-located runners keep the full ~1 Gbit/s with sub-percent variance.

Benchmark tooling for future runs (e.g. a true WAN measurement from a GitHub-hosted runner): tuist/cache-benchmark#2.

## Findings & follow-ups surfaced by the E2E (separate issues to file)
- **Cross-env claim leak**: production and staging both receive `tuist/tuist` webhooks and advertise identical `shape-linux-*` dispatch labels, so a shape-labeled job races both environments — all early smoke runs silently executed on production runners. Staging-targeting jobs must use `tuist-staging-*` profile labels.
- **Legacy staging cache down**: `cache-eu-central-staging.tuist.dev` / `cache-us-east-staging.tuist.dev` accept TCP/TLS but the app behind kamal-proxy hangs; staging users currently have no default remote cache.
- **Staging ClickHouse schema drift**: `runner_jobs.log_archived_at` was missing despite `schema_migrations` claiming the migration ran (applied manually).
- Cilium `enable-endpoint-routes=true` set on staging (documented kata-compat setting; to be persisted in the cilium chart values).

## Phase 3 (follow-up PR)
Retry backoff for self-healed nodes, cluster networking for macOS Tart VMs so the Scaleway Apple-Silicon fleet can consume its local nodes (dispatch already gates on `Catalog.fleet_on_cluster_network?/1`), **cluster-locality check in dispatch** — `cache_endpoint_url` must only be handed to runners in the same cluster as the node (svc DNS doesn't cross clusters; staging hides this because runners and Kura share one cluster, but prod Linux runners live in Hetzner while `scw-fr-par-runners` nodes would live in Scaleway — this is a hard precondition for activating the Scaleway region), RunnerCache test coverage, residency gating, observability.


infra/linux-runner-image/dispatch-poll.sh
```diff
@@ -119,9 +124,35 @@ while true; do
...
       fi
+      # Rollout-bridge single-container exec path: route the job's Tuist
+      # cache at the account's private runner-cache Kura node when the
+      # server includes it. Exported before exec so the GitHub Actions
+      # runner — and every job step — inherits it; the Tuist CLI honors
+      # TUIST_CACHE_ENDPOINT as a cache-endpoint override.
+      cache_endpoint=$(jq -r '.cache_endpoint_url // empty' /tmp/dispatch.json)
+      if [ -n "${cache_endpoint}" ]; then
+        echo "$(date -u +%FT%TZ) dispatch-poll: routing cache to runner-local endpoint ${cache_endpoint}"
+        export TUIST_CACHE_ENDPOINT="${cache_endpoint}"
```
fortmarek, line 154: alternative would be a proxy, but this model feels simpler. But can change to a proxy if this proves to be problematic for whatever reason

infra/helm/tuist/templates/runners-namespace.yaml
```diff
@@ -294,9 +315,6 @@ spec:
   podSelector:
     matchLabels:
       tuist.dev/runner: "true"
```
pepicrft, line 333: Dropping the `runner-pool-owner DoesNotExist` selector seems to bring back the failure mode the deleted comment was set up to prevent: this policy's peer set is `component=server`, server's image tag is `sha-<commit>` so identity churns on every Helm upgrade, and claimed runner Pods are now selected, so their egress datapath regenerates on each deploy. The old note said that drops the runner's GitHub WebSocket and surfaces as 'runner lost communication'. The new comment addresses the stamping live-lock but doesn't say why the rollout disruption is acceptable now. Worth watching a real staging deploy with a mid-job runner before assuming this is benign.

server/lib/tuist/kura/runner_cache.ex
```diff
@@ -0,0 +1,221 @@
...
+    end
+
+    :ok
+  end
+
+  # The first private (runner-cache) region available in this runtime,
+  # or nil. `available/0` is env-gated, so this stays nil until a
+  # private region is wired into `TUIST_KURA_AVAILABLE_REGIONS`.
+  defp runner_cache_region do
+    Enum.find(Regions.available(), &Regions.private?/1)
```
pepicrft: If two private regions ever end up in `TUIST_KURA_AVAILABLE_REGIONS` at once (eg during a migration from one to another), `Enum.find` picks one and the other is silently not reconciled at all: no provisioning, no teardown. Not blocking since today only one is wired per env, but the failure shape is silent, which is the part worth flagging.

server/lib/tuist/kura/runner_cache.ex
```diff
@@ -0,0 +1,221 @@
...
+          "" -> nil
+          trimmed -> trimmed
+        end
+
+      _ ->
+        nil
+    end
+  end
+
+  defp accounts_needing_node(region_id) do
```
pepicrft, line 124: The 'candidate set is tiny' premise holds in prod, but in any env where `FeatureFlags.runners_enabled?` short-circuits to `true` (non-prod, where auto-profiles can spread), this loads every Account with a profile into BEAM every 30s and runs a FunWithFlags lookup per account. Folding the flag check into SQL (or a hard cap on the candidate count) would make the load bounded in those envs.

server/lib/tuist/kura.ex
```diff
@@ -311,6 +326,86 @@ defmodule Tuist.Kura do
...
+  traffic stays inside the cluster, next to the runners, instead of
+  crossing the public ingress dataplane. Auth is unchanged: the same
+  Guardian JWT is verified by the same `tuist.lua` hook on the private
+  node, which carries the same `tenantID`. This reads the active
+  private `Server` row directly — `account_cache_endpoints` is
+  CLI-facing and a developer machine can't reach the in-cluster
+  endpoint.
+  """
+  def runner_cache_endpoint_url(%Account{id: account_id}) do
+    private_region_ids = Enum.map(Enum.filter(Regions.all(), &Regions.private?/1), & &1.id)
```
pepicrft: Small one: this reads from `Regions.all()` rather than `Regions.available()`. If a private region is dropped from `TUIST_KURA_AVAILABLE_REGIONS`, the controller stops reconciling it, but the stale `url` on the Server row keeps getting returned to runner dispatch, so jobs export a `TUIST_CACHE_ENDPOINT` they can't reach. Gating on `available()` would make config drift self-heal.

server/lib/tuist/kura.ex
```diff
@@ -311,6 +326,86 @@ defmodule Tuist.Kura do
...
+
+    if private_region_ids == [] do
+      nil
+    else
+      Server
+      |> where([s], s.account_id == ^account_id and s.status == :active and s.region in ^private_region_ids)
+      |> order_by([s], asc: s.region)
+      |> limit(1)
```
pepicrft: The 'at most one active private node per account' invariant is enforced by the reconciler but not by the DB, so a race could produce two rows. `limit(1)` + `order_by(:region)` then quietly picks one alphabetically. Using `Repo.one/2` without the limit would raise on duplicates and surface the regression at the dispatch path, which feels healthier than silently routing to one of them.

server/lib/tuist_web/controllers/runners_controller.ex
```diff
@@ -27,13 +27,12 @@ defmodule TuistWeb.RunnersController do
...
+         {:ok, %{jit: jit, account: account, workflow_job_id: workflow_job_id} = claim} <-
            Runners.dispatch_for_sa(ns, sa_name) do
-      json(conn, %{
-        encoded_jit_config: jit,
-        owner: account.name,
-        workflow_job_id: workflow_job_id
-      })
+      json(
+        conn,
+        dispatch_response(jit, account, workflow_job_id, Map.get(claim, :fleet_on_cluster_network, false))
```
pepicrft: Minor: `Runners.dispatch_for_sa` always sets `:fleet_on_cluster_network`, so the `false` default papers over a future code path that forgets to set it (silently disables in-cluster routing instead of crashing). `Map.fetch!/2`, or moving the key into the `with` head match, would assert the contract.

pepicrft (approved):
Did a careful read pass over the diff. Approving so this can land. Left a few inline observations: the network-policy one is the one I'd most want a second look at before/after a staging deploy, the rest are quieter shape concerns.

server/lib/tuist/kura/runner_cache.ex
fortmarek: Fixed in 5facde9467: `runner_cache_region/0` now logs an error listing every configured private region when more than one is wired, then reconciles the first deterministically — no longer a silent no-op for the others.

server/lib/tuist/kura/runner_cache.ex
fortmarek: Fixed in 5facde9467: the candidate query now carries a per-tick `LIMIT` so the BEAM load is bounded, and logs when the cap is hit (convergent provisioning picks up any overflow next tick). A cap rather than folding the flag into SQL because the `:runners` gate is often a global boolean gate, so an actor-gate pre-join wouldn't shrink the set. (Side note: the `FeatureFlags.runners_enabled?` short-circuit was already replaced earlier — this gates on `FunWithFlags.enabled?/2` directly.)

server/lib/tuist/kura.ex
fortmarek: Fixed in 5facde9467: switched `Regions.all/0` → `Regions.available/0`, so a private region dropped from `TUIST_KURA_AVAILABLE_REGIONS` stops being handed to dispatch — config drift now self-heals.

server/lib/tuist/kura.ex
fortmarek: Addressed in 5facde9467 with a variant: instead of `Repo.one` raising (which would 500 the dispatch hot path over a benign duplicate — both rows are the same account's valid caches, so either URL works), it fetches `limit(2)`, logs an error with both URLs when a duplicate exists, and routes deterministically. Surfaces the regression loudly without converting it into an availability hit. Happy to switch to raising if you'd rather fail hard here.

server/lib/tuist_web/controllers/runners_controller.ex
fortmarek: Fixed in 5facde9467: `:fleet_on_cluster_network` is now matched in the `with` head (no `Map.get` default), so a future path that forgets to set it fails the clause loudly instead of silently disabling in-cluster routing. Updated the existing test stub to honor the contract and added coverage asserting the URL is included for cluster-networked fleets and withheld otherwise.

infra/helm/tuist/templates/runners-namespace.yaml
fortmarek: Investigated rather than reverted (reverting reintroduces the livelock the new code fixes). The rollout-disruption mechanism requires the server's Cilium identity to churn on each deploy — and on staging it doesn't:

- The server Pod's identity-relevant labels are `app.kubernetes.io/{name,instance,component}`, all deploy-stable. The image tag (`sha-<commit>`) is a container-spec field, **not a label**; `pod-template-hash` is excluded from Cilium identity by default.
- Verified live: the running server endpoint's identity is **6770**, with security-labels `component=server, instance=tuist` (+ name/namespace) only — no chart/version/sha. A rolling deploy gives the new Pod the same identity, so no owned-runner egress map regenerates.
- Separately, the GitHub control-plane WebSocket is `runner → github.com:443`, governed by the `runners-default-deny` `0.0.0.0/0` ipBlock rule — **not** this server-peer rule — so selecting mid-job Pods here can't gate it. The only regeneration that does happen (the runner's *own* identity churn when the owner label is stamped) occurs once at claim time, before the job's WebSocket exists.

Recorded all of this in the policy comment in 5facde9467. I'll still watch a real staging deploy with a mid-job runner and report back before treating it as fully settled — agree the empirical check is worth doing.

PR #10988, opened 2026-05-29
feat: Tailscale-based human kubectl access + JIT elevation bot

fortmarek:
## Summary

Per-person identity-based `kubectl` access to every Tuist cluster, with default-deny tiers (`view` for everyone, including founders) and a Slack-approved JIT elevation flow as the single write path. Built on Pomerium per workload cluster + a Go sidecar (`kube-impersonator`) that injects per-request impersonation headers + a new standalone internal-ops service (`tuist-ops/`) that owns the Slack bot and the policy endpoint.

Today the only access boundary is 1Password vault membership and every kubeconfig behind it is `cluster-admin`. There is no per-person identity, no graduated permission, and no friction in the way of an agent acting on a founder's behalf wiping production. This PR fixes all three and, deliberately, keeps internal cluster access on a separate failure domain from the customer-facing Tuist server.

## Decoupling: why a new `tuist-ops/` service

Earlier iterations of this PR put the JIT bot inside the customer-facing `server/` and mutated the tailnet ACL at runtime. Both were wrong:

- **Same failure domain as customers.** A customer-facing outage would take down internal cluster access right when we needed it to fix the outage.
- **The HuJSON splice was the worst code in the PR.** Mutating the tailnet ACL document at runtime required a hand-rolled HuJSON splicer with 12 golden-file tests defending byte-preservation; a subtle bug could rewrite production tailnet policy and look like a normal elevation in every log.
- **Conflated audit surfaces.** Internal access events lived next to customer auth events in the same DB.

This PR instead:

- Stands up `tuist-ops/` as a standalone Phoenix app deployed to the production cluster (same pattern as `slack/` — internal Tuist-team tooling lives in its own namespace alongside customer workload). JIT is its first feature module; future `/ops/db` and other internal tooling land here too (the URL is `ops.tuist.dev`, equivalent of today's `tuist.dev/ops/*`).
- Uses **Pomerium** as the kubectl gateway. Pomerium authenticates humans via **Google Workspace OIDC** (not via Tuist's customer OIDC), and a per-pod **`kube-impersonator` Go sidecar** calls `tuist-ops`'s policy endpoint per request to resolve impersonation headers.
- Stores elevation state as a row in tuist-ops's own CNPG Postgres, not a mutation to the tailnet ACL. The tailnet ACL is now static and code-reviewed; `tuist-ops` never writes to it.

### New: `tuist-ops/` Phoenix app
- `lib/tuist_ops/jit/` — JIT bot business logic (`Approvals`, `Elevation`, `Policy`, `Request`, `SlackBlocks`, `SlackClient`, `TailscaleClient`, `Workers.RevertWorker`). Role-based policy (Owner/Admin/Member tailnet roles drive self-approve + approver-allowed decisions; no email lists in code).
- `lib/tuist_ops_web/controllers/slack_controller.ex` — `/webhooks/slack/{slash,interactive}` Slack-signed endpoints.
- `lib/tuist_ops_web/controllers/policy_controller.ex` — impersonation policy endpoint (`/api/v1/policy`) dialled per kubectl call by each env's `kube-impersonator` sidecar. Derives env from the Host header (`kube-staging.tuist.dev` → `staging`), reads `x-pomerium-claim-email` for subject, queries the Elevation table, returns `Impersonate-User` + `Impersonate-Group` response headers (HTTP 200) or 403.
- `lib/tuist_ops_web/controllers/health_controller.ex` — `/api/v1/healthz` for k8s probes.
- `Dockerfile` + `rel/overlays/bin/server` + standard Phoenix scaffolding (mix.exs, config/, application, repo, endpoint, router).
- `test/jit/` — three unit test suites (policy, approvals, slack_blocks). 30 tests, all passing.

### New: `kube-impersonator/` Go sidecar
- Tiny stdlib-only Go reverse proxy (~250 lines, distroless image ~8MB), deployed alongside Pomerium in each env's `pomerium` namespace.
- Per kubectl request: reads `X-Pomerium-Claim-Email`, calls `tuist-ops/api/v1/policy` over the tailnet egress, reads back `Impersonate-User` + `Impersonate-Group(s)` response headers, strips the inbound bearer, attaches the pod's ServiceAccount token, and reverse-proxies to `kubernetes.default.svc:443`.
- **Fails closed.** Policy call failure → 502 to kubectl, never forwards to the apiserver with no impersonation decision.
- `Dockerfile` + `.github/workflows/kube-impersonator-image.yml` (builds + pushes `ghcr.io/tuist/kube-impersonator:sha-<short>` + branch tag on every push touching `kube-impersonator/**`).
- Why a sidecar at all: Pomerium can set static request headers but can't make per-request external lookups for dynamic header injection — the elevation→edit-tier decision lives in tuist-ops's DB and has to be resolved per request. Pomerium's `extra_envoy_filters` (the path the original design assumed) isn't actually a supported config key in any Pomerium version we tested.

### New: Pomerium chart (`infra/helm/pomerium/`)
- **Self-contained** chart, not a wrapper around upstream's `pomerium/pomerium`. Upstream is multi-component split-mode (separate `authenticate` / `authorize` / `databroker` / `proxy` workloads) — too heavy for our scale. We deploy a single all-in-one Pomerium pod with the impersonator sidecar in the same pod.
- One Helm release per workload env (staging / canary / production), with per-env values overlays.
- `kube-<env>.tuist.dev` is the public kubectl endpoint; `authenticate.kube-<env>.tuist.dev` is the OIDC callback host. cert-manager DNS-01 issues both.
- `templates/configmap.yaml` — Pomerium config (Google OIDC, route `to: 127.0.0.1:8081` pointing at the sidecar, `jwt_claims_headers: { X-Pomerium-Claim-Email: email }` so the sidecar can identify the user, memory-backed databroker).
- `templates/deployment.yaml` — two containers, Pomerium + kube-impersonator, sharing the pod ServiceAccount.
- `templates/access-tiers.yaml` — ClusterRoleBindings for `tuist-admins → view`, `tuist-eng → view`, `tuist-<env>-write → edit`. This chart owns the impersonation-tier RBAC.
- `templates/rbac.yaml` — grants the Pomerium pod's SA the `impersonate` verb on users/groups/serviceaccounts.
- `templates/external-secret.yaml` — ESO syncing Pomerium credentials (`shared_secret`, `cookie_secret`, IdP client id/secret) from per-env `POMERIUM_<ENV>` 1P items.
- `templates/tailscale-egress.yaml` — `ExternalName` Service `tuist-ops-egress` annotated for the tailscale operator so the sidecar reaches `ops.<tailnet>.ts.net` without joining the tailnet itself. Tagged `tag:tuist-k8s-<env>` so the operator can mint the egress device (default `tag:k8s` isn't authorised on the OAuth client).
- `.github/workflows/pomerium-deployment.yml` — matrix per-env deploy via `workflow_dispatch`, scoped per `server-k8s-<env>` GitHub environment.

### New: `tuist-ops` chart (`infra/helm/tuist-ops/`)
- Single-instance deploy targeted at the **production cluster** (its own namespace `tuist-ops`, isolated from `server/` at every layer except cluster).
- `templates/deployment.yaml` — Recreate strategy (single-instance Postgres pool).
- `templates/service.yaml` — `ClusterIP` + `tailscale.com/expose: true` + `tailscale.com/tags: tag:tuist-k8s-production` so the tailscale-operator provisions a tailnet device at `ops.<tailnet>.ts.net`. Pomerium pods in all envs reach it via this device.
- `templates/ingress.yaml` — public ingress on `ops.tuist.dev` that **only routes `/webhooks/slack/*`** to the upstream; `/api/v1/policy` stays reachable only on the tailnet.
- `templates/cnpg-cluster.yaml` — own CNPG Cluster (single instance, 5Gi storage, daily backups to Tigris).
- `templates/externalsecret.yaml` + `templates/backup-credentials.yaml` — three ExternalSecrets (runtime credentials, SECRET_KEY_BASE, S3 backup keys).
- `.github/workflows/tuist-ops-deployment.yml` — build + push image to ghcr.io, deploy to production via the same `server-k8s-production` GitHub environment used by the main server.

### `mise.toml`
- Pinned `github:pomerium/cli @ 0.32.2` — the kubeconfig exec credential plugin. `mise install` picks it up; no `brew install` step.

### Removed from `server/`
- `lib/tuist/tailscale_jit/` — entire module tree (relocated to `tuist-ops/lib/tuist_ops/jit/`).
- `lib/tuist_web/controllers/webhooks/tailscale_jit_controller.ex`.
- `lib/tuist_web/plugs/slack_webhook_plug.ex` (no other callers).
- `priv/repo/migrations/...create_tailscale_jit_tables.exs` (relocated to tuist-ops's migrations).
- Boundary exports in `lib/tuist.ex`.
- `TAILSCALE_JIT_*` env vars in `lib/tuist/environment.ex`.
- `tailscale-jit-externalsecret.yaml` + `tailscaleJIT` block in the server's Helm chart.
- Mimic.copy lines in `server/test/test_helper.exs` for the removed modules.
- Drift reconciler + reconciler-on-boot + Application supervisor (no drift possible with DB-flag elevation).

### Removed from `infra/helm/tailscale-operator/`
- `apiServerProxyConfig` from values + overlays — the operator's in-process apiserver proxy was the v1 kubectl gateway; Pomerium replaces it.
- `accessBindings` from values + overlays — the ClusterRoleBindings that mapped impersonated groups to ClusterRoles now live in the Pomerium chart.
- `templates/rbac.yaml` deleted (rendered nothing once `accessBindings: []`).

### Tailnet ACL (`infra/tailscale/acls.json`)
- Break-glass groups (`group:tuist-{staging,canary,prod}-write`) and their impersonation grants deleted. No more runtime ACL mutation; the ACL is humans-only static.
- The `autogroup:member` → `tag:tuist-k8s-<env>` kubectl-via-operator-proxy grant deleted (operator proxy retired; Pomerium is the gateway).
- Cluster-proxy `src` selectors use `autogroup:owner` + `autogroup:admin` (cluster admins) and `autogroup:member` (engineers). Role changes in the Tailscale admin console propagate without an ACL edit.

### Documentation
- `kube-impersonator/AGENTS.md` (new) — sidecar architecture, failure mode, config knobs.
- `tuist-ops/AGENTS.md` (new) — what runs here and how to extend it (routing convention: never `/ops` prefix, the subdomain is the prefix).
- `AGENTS.md` (root) — repo map updated with `tuist-ops/` and `kube-impersonator/` entries.
- `infra/AGENTS.md` updated to describe the new Pomerium + sidecar + tuist-ops architecture and the post-pivot agent rules.
- `infra/k8s/jit-elevation-audit.md` rewritten for the three-trail audit story (Slack + tuist-ops Postgres + Pomerium access log; the "Tailscale ACL audit log" trail no longer applies).
- `server/data-export.md` — JIT tables removed from the inventory (they're not in this server anymore).

### Default read access (no elevation)

```
kubectl --context kube-staging.tuist.dev get pods
  → pomerium-cli exec plugin supplies bearer (session cached from Google OIDC, no creds on disk)
  → Pomerium (in workload cluster) validates session
  → forwards to kube-impersonator sidecar on 127.0.0.1:8081
  → sidecar: GET ops.<tailnet>.ts.net/api/v1/policy
       Host: kube-staging.tuist.dev
       X-Pomerium-Claim-Email: marek@tuist.dev
       tuist-ops: TailscaleClient.user_role(marek@tuist.dev) → :owner
                  No active elevation row for (marek, staging)
                  → 200 OK
                    Impersonate-User: marek@tuist.dev
                    Impersonate-Group: tuist-admins
  → sidecar strips inbound bearer, attaches pod SA token, copies Impersonate-* onto upstream request
  → kubernetes.default.svc:443
  → apiserver impersonates marek as tuist-admins → ClusterRole `view` → reads allowed
```

### Elevating to write

```
1. In Slack: /elevate prod 30m "restart deploy/tuist-server"
2. tuist-ops posts approval card in #tailscale-jit-approvals
3. Second human clicks Approve (or Marek self-approves on non-prod)
4. tuist-ops writes Elevation row: { requester: marek, env: production,
                                     status: active, expires_at: now+30m }
5. Slack card flips to "Active until ..."
6. Marek goes back to terminal, no kubeconfig change:
   kubectl --context kube-prod.tuist.dev delete pod tuist-server-abc
     → same flow as above, but the policy call now finds the active row
     → returns: Impersonate-Group: tuist-admins
                Impersonate-Group: tuist-prod-write
     → apiserver RBAC: tuist-prod-write → edit → delete allowed
7. At 30m mark, RevertWorker flips status=reverted, Slack card updates
8. Next call after that: back to view tier
```

The transition view → edit → view is **per HTTP request** at the sidecar, driven by the DB row's existence. Revocation takes effect on the next call after the row flips, not on a propagation delay.

### Decoupled failure domains
The customer-facing Tuist server can be down without affecting cluster access (Pomerium + sidecar + tuist-ops run independently of `server/`). The reverse is also true.

### No bearer in the kubeconfig
`pomerium-cli` is an exec credential plugin: the bearer is fetched on demand per-call, never on disk.

### Default-deny tiers; only Slack-approved sessions can write
View tier is `view` (no Secrets readable). Write tier requires an active Elevation row that resolves at request time. Pomerium gates session identity at the network boundary; the sidecar gates per-request impersonation; RBAC gates verbs at the apiserver; the elevation row is checked by tuist-ops between them.

### Agent containment
The agent inherits the human's session via the same `pomerium-cli` exec plugin. It **cannot**:
- Trigger `/elevate` autonomously (no Slack identity, no signing secret).
- Click Approve (same gate).
- Promote its own tailnet role (admin-console access via Google account is needed).

So the agent's effective tier is bounded by what the human has at this exact moment. If the human elevates for "restart deploy", the agent inherits that for the window; this is what ClawPatrol (a separate follow-up) would gate at the per-command level.

### Authorization policy

| Approver's tailnet role | staging | canary | production |
|---|---|---|---|
| **Owner** | self-approve / approve ✓ | ✓ | ✓ |
| **Admin** | ✓ | ✓ | ✓ |
| **Member** (engineers) | ✓ | ✓ | ✗ |
| **Auditor / Billing admin / off-tailnet** | ✗ | ✗ | ✗ |

The same matrix gates both `self_approval_allowed?` (requester clicks Approve on their own request) and `approver_allowed?` (second human clicks). Production approval requires Owner/Admin role even when the requester is different — closes the "two engineers agreeing among themselves" hole.

### Fail-closed sidecar
If tuist-ops is unreachable or the policy call fails, the sidecar returns 502 to kubectl. Better to interrupt access than forward to the apiserver with no impersonation decision (which would either fail RBAC anyway or, worse, succeed as the Pomerium pod's SA).

### Per-user, per-call audit
Three independent trails join on the elevation id (see `infra/k8s/jit-elevation-audit.md`):

| Trail | Granularity | Where |
|---|---|---|
| Slack thread | Per-elevation | Slack (off-platform) |
| tuist-ops Postgres | Per-elevation lifecycle | production cluster CNPG (own cluster, not the main server's) |
| Pomerium access log | **Per kubectl call** with user email + verb + path + response code | Grafana Cloud Loki via k8s-monitoring |

Apiserver-level audit (full request bodies, resource diffs) is a separate follow-up — `ClusterClass` flag change with different blast radius.

## Manual steps on merge (not automated)

Staging is **already wired and validated end-to-end** on this branch (see Validation below). To finish the rollout to canary + production:

1. **Mirror `infra/tailscale/acls.json` into the Tailscale admin console.** Tailscale has no public ACL sync API; the file is the source of truth and humans copy it across. (Already done for the staging-validating state.)
2. **Per-env Pomerium 1P items** for canary + production (`POMERIUM_CANARY`, `POMERIUM_PRODUCTION`): `shared_secret`, `cookie_secret`, `idp_client_id`, `idp_client_secret`. Staging item already populated.
3. **Shared Google Workspace OAuth client** covers all three envs (intentional simplification at our scale; trade-off documented in `infra/helm/pomerium/NOTES.md`). Verify in Google Cloud Console that the client's "Authorized redirect URIs" list includes `https://authenticate.kube-staging.tuist.dev/oauth2/callback`, `https://authenticate.kube-canary.tuist.dev/oauth2/callback`, and `https://authenticate.kube-prod.tuist.dev/oauth2/callback` — Pomerium fails OIDC with `redirect_uri_mismatch` otherwise.
4. **Deploy:** `gh workflow run pomerium-deployment.yml -f environment=canary --ref main` then the same for `production`.
5. **Each person runs**: kubeconfig set-cluster / set-credentials / set-context once per env using `pomerium-cli k8s exec-credential` as the exec command. (Snippet in `infra/AGENTS.md`.) First kubectl call opens a browser for the OIDC dance; session caches for ~24h.

## Follow-ups (tracked separately, not in this PR)

1. **Apiserver audit logging** (`ClusterClass` flag). Adds full request/response audit on top of Pomerium's per-call log.
2. **Tailscale-only lockdown** of the apiserver (firewall the public LB to tailnet source IPs). The 1P break-glass kubeconfig stays as the fallback path.
3. **ClawPatrol as the per-command agent gate.** Composes with Pomerium: Pomerium gates session identity, ClawPatrol gates individual commands across all tools an agent might use (kubectl, gh, op, terraform). Wait for ClawPatrol 1.0.
4. **SCIM / Workspace-group sync.** When team grows past ~15 people across multiple ops domains, sync Workspace groups into the tailnet for richer role decomposition. Same `TailscaleClient.user_role/1` shape, broader source of truth.
5. **Migrate `/ops/db` into tuist-ops.** First user of the new ops Phoenix app beyond the JIT bot. Adds LiveView to the tuist-ops deps + lands at `ops.tuist.dev/db`.
6. **Pin Pomerium image to sha-<short>** in `infra/helm/pomerium/values.yaml` rather than the branch tag, once this lands on main.

### Static / build-time
- **tuist-ops** compiles clean (`mix compile --warnings-as-errors`); 30 unit/integration tests pass.
- **server** compiles clean after JIT excision.
- **kube-impersonator** Go binary builds clean (`go build ./...`); distroless image builds + pushes via `.github/workflows/kube-impersonator-image.yml`.
- Helm chart renders clean for all three charts (`helm template` against staging values).

### End-to-end on staging — validated

The full chain ran against the live staging cluster on this branch:

1. ✅ `pomerium-cli k8s exec-credential https://kube-staging.tuist.dev` opens the browser, the Google sign-in completes, session caches locally.
2. ✅ `kubectl --context kube-staging.tuist.dev auth whoami` returns `Username: marek@tuist.dev`, `Groups: [tuist-admins system:authenticated]`. The chain (Pomerium → sidecar → tuist-ops PolicyController → apiserver) is proven by the impersonation showing through.
3. ✅ View-tier ceiling enforced: `create configmap` → 403 forbidden, `get secrets` → 403 forbidden.
4. ✅ `kubectl get pods -n tuist`, `get ns`, `describe`, `logs` all work (view-tier reads).
5. ✅ `/elevate staging 5m phase3-validation` posts the approval card in the configured Slack channel; self-approve flips it active and writes the `tailscale_jit_elevations` row.
6. ✅ Same kubectl context immediately allows `create configmap` (no kubeconfig refresh): `Groups: [tuist-admins tuist-staging-write system:authenticated]`. The dynamic edit-tier comes from the policy response to that very kubectl call.
7. ✅ RevertWorker logs visible at TTL boundary. Next kubectl call after that returns to view tier.

Canary + production rollout is one `gh workflow run` per env away, then the same e2e on each.


tuist-ops/lib/tuist_ops/jit/tailscale_client.ex
fortmarek, line 1: we can potentially replace this by Google groups but for now, piggybacked on Tailscale roles

kube-impersonator/main.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+}
+
+// apiserverTransport: the apiserver's serving cert is signed by
+// the cluster CA. We don't bundle a trust store here; the network
+// hop is pod-local to in-cluster (`kubernetes.default.svc`) and
+// the threat model is already constrained by pod-network policy.
+func apiserverTransport() http.RoundTripper {
+	return &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, //nolint:gosec
```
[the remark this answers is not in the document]
fortmarek: Fixed by pinning the cluster CA instead of disabling verification — `kube-impersonator` now reads `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt` at startup, builds an x509 pool, and sets `RootCAs` on the upstream transport with `MinVersion: TLS 1.2`. No more `InsecureSkipVerify`. Commit at HEAD.

infra/helm/pomerium/templates/deployment.yaml
```diff
@@ -0,0 +1,122 @@
...
+        # kubectl request flows through here on the way to the
+        # apiserver. SA token is auto-mounted by Kubernetes via the
+        # ServiceAccount; the sidecar reads it for the upstream
+        # Authorization header.
+        - name: kube-impersonator
+          image: "{{ .Values.kubeImpersonator.image.repository }}:{{ .Values.kubeImpersonator.image.tag }}"
+          imagePullPolicy: {{ .Values.kubeImpersonator.image.pullPolicy }}
+          env:
+            - name: LISTEN_ADDR
+              value: ":{{ .Values.kubeImpersonator.port }}"
```
pepicrft: High: this sidecar should not listen on all pod interfaces. `LISTEN_ADDR=:8081` binds the impersonator to the pod IP, so any pod that can reach the Pomerium pod IP can bypass Pomerium, spoof `Host` and `X-Pomerium-Claim-Email`, and get the sidecar to forward with the Pomerium SA bearer plus policy-resolved impersonation groups. I validated the production render: it emits this listener and no NetworkPolicy for the Pomerium chart. Bind this to `127.0.0.1:8081` and add a NetworkPolicy, or split health checks onto a separate listener if kubelet probes need pod-IP reachability.

infra/helm/pomerium/templates/access-tiers.yaml
```diff
@@ -0,0 +1,66 @@
...
+  name: tuist-{{ .Values.tuistEnv }}-write-edit
+  labels:
+    {{- include "pomerium.labels" . | nindent 4 }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: edit
+subjects:
+  - kind: Group
+    name: tuist-{{ .Values.tuistEnv }}-write
```
pepicrft, line 65: High: production write access is bound to a different group than the policy emits. With `tuistEnv: production`, this template renders `tuist-production-write`, but `PolicyController` emits `tuist-prod-write` for production and the impersonator ClusterRole only allows `tuist-prod-write`. I validated the production render and the rendered `edit` binding is for `tuist-production-write`, so a production JIT elevation will look active but will not receive `edit`. Use one shared helper or value for the write group name and add a rendered chart assertion for all envs.

tuist-ops/priv/repo/migrations/20260603120000_create_tailscale_jit_tables.exs
```diff
@@ -0,0 +1,54 @@
...
+      timestamps(type: :timestamptz)
+    end
+
+    # Drift reconciler scans active elevations to compare DB state
+    # against actual ACL membership; covers the (status, expires_at)
+    # window query.
+    # excellent_migrations:safety-assured-for-next-line index_not_concurrently
+    create index(:tailscale_jit_elevations, [:status, :expires_at])
+    # excellent_migrations:safety-assured-for-next-line index_not_concurrently
+    create index(:tailscale_jit_elevations, [:request_id])
```
pepicrft, line 52: Medium: this needs to be unique if one request is supposed to produce one elevation. `approve/2` reads the pending request without a row lock and then inserts an elevation, while the Slack revoke button carries only one `elevation.id`. Two near-simultaneous Approve clicks can persist multiple active rows for the same request, and revoking the visible card only pulls one revert job forward. I validated the test schema accepts two active elevations for a single request. Add a unique constraint on `request_id` or otherwise lock the request during approval so revoke and idempotency match the documented behavior.

pepicrft (changes requested):
Security review found blocking issues that should be fixed before merge.

infra/helm/pomerium/templates/deployment.yaml
fortmarek: Split kube-impersonator into two listeners. Proxy is now on `127.0.0.1:8081` (only same-pod Pomerium can reach it — loopback bind is the auth); kubelet probes go to a new `:8082` health listener that only serves `/healthz` and has no proxy mounted. Deployment + values updated to match. NetworkPolicy left as separate hardening.

infra/helm/pomerium/templates/access-tiers.yaml
fortmarek: Renamed everywhere to `tuist-production-write` to match what the chart's `tuist-{{ .Values.tuistEnv }}-write` template renders for `tuistEnv: production`. 11 sites: PolicyController (env_write_group + group_for_env), Policy (@group_to_env), SlackController (@env_to_group), the impersonator ClusterRole resourceNames, and 5 test fixtures.

tuist-ops/priv/repo/migrations/20260603120000_create_tailscale_jit_tables.exs
fortmarek: Two layers: (1) new migration replaces the non-unique index on `tailscale_jit_elevations.request_id` with a unique one + Elevation.create_changeset declares `unique_constraint(:request_id)` — DB-level safety net; (2) `Approvals.approve/2` takes `SELECT ... FOR UPDATE` on the Request row at the top of the transaction, so concurrent approves serialise on the lock and only the first one proceeds. The second re-reads, sees `status: "approved"`, falls through to the existing already-approved idempotent branch. New test in `changesets_test.exs` asserts the DB rejects the duplicate insert.

pepicrft (approved):

PR #11003, opened 2026-06-01
fix(cli): skip invalid activity logs during cache warm

pepicrft:
Resolves no GitHub issue.

This fixes cache warming failures caused by invalid `.xcactivitylog` files collected during `tuist cache warm`.

The activity log parsing in this path is only used to attach build-time metadata before storing cache artifacts. If Xcode leaves behind an invalid or incomplete log, failing the whole cache warm is stricter than necessary because the artifact has already been built and can still be stored without timing metadata.

This change makes build-time extraction best-effort: invalid logs are skipped with a warning, while valid logs continue to contribute target build times.

### How to test locally

- `tuist generate TuistXCActivityLogTests --no-open`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme TuistUnitTests -only-testing TuistXCActivityLogTests/XCActivityLogControllerTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`

### DIFF
--- a/cli/Sources/TuistXCActivityLog/XCActivityLogController.swift
+++ b/cli/Sources/TuistXCActivityLog/XCActivityLogController.swift
@@ -106,7 +106,8 @@ public struct XCActivityLogController: XCActivityLogControlling {
                     withoutBuildSpecificInformation: false
                 )
             } catch {
-                throw XCActivityLogControllerError.wrap(error, path: activityLogPath)
+                logSkippedBuildTimeExtraction(error: error, path: activityLogPath)
+                continue
             }
             let buildStep: BuildStep
             do {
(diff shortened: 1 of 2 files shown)


ZeddS2:
Hi @pepicrft , may I ask why this was closed without merging?

We are seeing the same issue with Tuist 4.191.3 + Xcode 26.2.

`tuist cache warm` succeeds through build and XCFramework creation, then fails at `Storing binaries to speed up workflows` because one `.xcactivitylog` is invalid.

The behavior in this PR seems like exactly what we need, so I’m curious if this approach was abandoned for a specific reason.

PR #11008, opened 2026-06-01
fix(server): fall back from empty Kura endpoints

pepicrft:
## What changed

- Canary Kura acceptance tests now reuse the existing `tuist` account and create random projects under it, instead of creating random temporary organizations.
- The production deployment workflow no longer strips `TUIST_FEATURE_FLAG_KURA` from the acceptance test step, so the canary leg exercises Kura again.
- Server Kura endpoint resolution now falls back to the existing cache endpoints when the account-scoped Kura endpoint list is empty or the configured global Kura endpoint list only contains blanks.
- Canary cluster config now declares a `kura` worker pool labeled with `node.cluster.x-k8s.io/pool=kura`, matching the scheduler selector used by Kura pods.

## Why

Kura servers are scoped to an account. In canary, the Kura servers are bound to `tuist`, but the acceptance tests were creating temporary organizations. With `TUIST_FEATURE_FLAG_KURA=1`, those temporary accounts requested Kura endpoints and got no usable endpoint back, which led the CLI into a bad cache URL path.

There was also an infrastructure issue in canary: the Kura StatefulSet pods were pending because they select `node.cluster.x-k8s.io/pool=kura`, but the cluster had no nodes with that pool label.

## Impact

The acceptance tests create isolated random projects while using the account that owns the canary Kura nodes. When Kura is unavailable or has no endpoints, clients still receive the normal cache endpoints instead of an empty list.

## Validation

- `swiftformat cli/Sources/TuistAcceptanceTesting/TuistAcceptanceTestFixtureTestingTrait.swift cli/Tests/TuistCacheEEAcceptanceTests/TuistCacheEECanaryAcceptanceTests.swift`
- `tuist install`
- `mise run cli:ee`
- `TUIST_EE=1 tuist generate TuistAcceptanceTesting TuistCacheEEAcceptanceTests --no-open`
- `xcodebuild build -workspace Tuist.xcworkspace -scheme TuistCacheEEAcceptanceTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `mix format --check-formatted lib/tuist/accounts.ex test/tuist_web/controllers/api/cache_controller_test.exs`
- `mix test test/tuist_web/controllers/api/cache_controller_test.exs`
- Canary diagnosis with `kubectl`: `kura-tuist-eu-central-1-0` was pending because no node matched `node.cluster.x-k8s.io/pool=kura`.

### DIFF
--- a/cli/Tests/TuistCacheEEAcceptanceTests/TuistCacheEECanaryAcceptanceTests.swift
+++ b/cli/Tests/TuistCacheEEAcceptanceTests/TuistCacheEECanaryAcceptanceTests.swift
@@ -28,7 +28,7 @@ struct TuistCacheEECanaryAcceptanceTests {
         .withMockedEnvironment(inheritingVariables: ["PATH"]),
         .withMockedNoora,
         .withMockedLogger(forwardLogs: true),
-        .withFixtureConnectedToCanary("generated_project_with_caching_enabled")
+        .withFixtureConnectedToCanary("generated_project_with_caching_enabled", accountHandle: "tuist")
     ) func generated_project_with_caching_enabled() async throws {
         let fixtureDirectory = try #require(TuistTest.fixtureDirectory)
         let xcodeprojPath = fixtureDirectory.appending(component: "App.xcodeproj")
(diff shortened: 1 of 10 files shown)


cli/Tests/TuistCacheEEAcceptanceTests/TuistCacheEECanaryAcceptanceTests.swift
```diff
@@ -146,7 +146,7 @@ struct TuistCacheEECanaryAcceptanceTests {
         .withMockedEnvironment(inheritingVariables: ["PATH"]),
         .withMockedNoora,
         .withMockedLogger(forwardLogs: true),
-        .withFixtureConnectedToCanary("generated_multiplatform_app")
+        .withFixtureConnectedToCanary("generated_multiplatform_app", accountHandle: "tuist")
```
pepicrft, line 149: @fortmarek, since cache nodes are bound to the account, we can't create an organization for each test case. I leaned on reusing the existing `tuist` one, but I can also create one for acceptance tests, like `tuist-acceptance-tests` since there's a risk of acceptance test temporary projects remaining in the account (e.g. when the execution is interrupted). Alternatively, I can keep "tuist", and have some job there that periodically cleans those projects (e.g. establishing a convention)

fortmarek (approved):

PR #11015, opened 2026-06-01
fix(cli): infer DerivedData location for inspect build and tests

natanrolnik:
## What

When no `--derived-data-path` is passed, `tuist inspect build` and `tuist inspect tests` now infer the DerivedData location the same way Xcode does, instead of always appending `<name>-<hash>` to the default directory.

## Why

The inference only handled Xcode's default DerivedData location. A custom `IDECustomDerivedDataLocation` (Xcode > Settings > Locations > Derived Data) was not supported:

- A **relative** location (e.g. `.derived-data`) couldn't be validated as an absolute path, so it fell back to `~/Library/Developer/Xcode/DerivedData` and the `Logs` directory was never found.
- A **relative** location uses a different per-project folder than the default: just `<workspaceName>` with **no** hash (the folder sits next to the workspace).

Xcode infers the mode from the shape of the string, with no companion key. The supported rule is now:

| `IDECustomDerivedDataLocation` | Root | Per-project folder |
|---|---|---|
| missing (Default) | `~/Library/Developer/Xcode/DerivedData` | `<name>-<hash>` |
| absolute (Custom) | the value | `<name>-<hash>` |
| relative (Relative) | `<workspace dir>/<value>` | `<workspaceName>` (no hash) |

## How

- New `DerivedDataLocation` enum in `TuistEnvironment` with a pure `classify(rawLocation:)`.
- New `Environment.derivedDataLocation()` seam (reads `IDEDerivedDataPathOverride` / `IDECustomDerivedDataLocation`); `derivedDataDirectory()` now returns only the global default.
- `DerivedDataLocator.locate` resolves root + whether to hash from the mode, keeping `DERIVED_DATA_DIR` / `BUILD_DIR` precedence (set by Xcode during post-actions).

Both `inspect build` and `inspect tests` benefit, since both go through `DerivedDataLocator.locate`.

## Validation

`xcodebuild test ... -only-testing TuistXcodeBuildProductsTests` passes, including new tests for the pure classifier and for the relative/custom/precedence cases. (The one unrelated failure in that target is a pre-existing locale-dependent `AppBundleLoaderTests` assertion comparing a Foundation error string in English.)

### DIFF
--- a/cli/Sources/TuistEnvironment/DerivedDataLocation.swift
+++ b/cli/Sources/TuistEnvironment/DerivedDataLocation.swift
@@ -20,18 +20,19 @@ public enum DerivedDataLocation: Equatable, Sendable {
     /// `<workspaceName>` without a hash, since the directory lives next to the workspace.
     case relativeToWorkspace(RelativePath)
 
-    /// Classifies the raw value of `IDECustomDerivedDataLocation` (or `IDEDerivedDataPathOverride`)
-    /// into a location, mirroring how Xcode interprets the preference string.
-    public static func classify(rawLocation: String?) -> DerivedDataLocation {
-        guard let rawLocation, !rawLocation.isEmpty else {
-            return .default
+    /// Builds a location from the raw value of `IDECustomDerivedDataLocation` (or
+    /// `IDEDerivedDataPathOverride`), mirroring how Xcode interprets the preference string:
+    /// an empty value means the default, an absolute path a custom location, and a relative
+    /// path a location resolved against the workspace directory.
+    public init(_ location: String) {
+        if location.isEmpty {
+            self = .default
+        } else if let absolute = try? AbsolutePath(validating: location) {
+            self = .custom(absolute)
+        } else if let relative = try? RelativePath(validating: location) {
+            self = .relativeToWorkspace(relative)
+        } else {
+            self = .default
         }
-        if let absolute = try? AbsolutePath(validating: rawLocation) {
-            return .custom(absolute)
-        }
-        if let relative = try? RelativePath(validating: rawLocation) {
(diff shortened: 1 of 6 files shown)


cli/Sources/TuistEnvironment/DerivedDataLocation.swift
```diff
@@ -0,0 +1,37 @@
...
+    /// same hashing as the default location, so a shared directory can host several projects.
+    case custom(AbsolutePath)
+
+    /// A location relative to the workspace directory. Per-project folders are named
+    /// `<workspaceName>` without a hash, since the directory lives next to the workspace.
+    case relativeToWorkspace(RelativePath)
+
+    /// Classifies the raw value of `IDECustomDerivedDataLocation` (or `IDEDerivedDataPathOverride`)
+    /// into a location, mirroring how Xcode interprets the preference string.
+    public static func classify(rawLocation: String?) -> DerivedDataLocation {
```
fortmarek: what do you think of having a custom `init(_ location: String)` or have a `rawValue: String`?

fortmarek (approved):
Just one minor nit, otherwise looks good, thanks!

cli/Sources/TuistEnvironment/DerivedDataLocation.swift
natanrolnik: Sounds good! Will try.

natanrolnik: I think that `rawValue` isn't the best fit here because the enum has associated values, but the init is indeed better 👍🏻

fortmarek (approved):
Amazing, thanks!

PR #11037, opened 2026-06-02
feat(noora, server): improve breadcrumbs, add animations and polish the component

asmitbm:
https://github.com/user-attachments/assets/40cac699-09f2-41ba-bb0e-160bf01667ee

## Breadcrumb project switcher: polish + animations

Improvements to the breadcrumb dropdown (the project switcher) and a few supporting additions to Noora.

What changed:

### Project switcher dropdown

* The selected project now shows a checkmark. Space for it is always reserved, so badges stay neatly aligned whether or not a row is selected.
* Gave the dropdown a wider minimum width so it doesn't feel cramped.
* The selector icon now morphs (smoothly animates) between two states when you open/close the dropdown.
* The dropdown now fades and slides in when opening, and out when closing, instead of appearing/disappearing instantly.

### Dropdown item structure

* Grouped the item's text content separately from the trailing icon, so the layout is cleaner and easier to extend (e.g. badges/buttons next to the label).
* Reusable additions to Noora

### New easing tokens (e.g. --ease-out-cubic) available globally for animations.

* A new morphing_icon component that smoothly transitions one icon into another based on state.
* A new selector-2 icon.

### Developer experience

* Local dev now hot-reloads when Noora's JS/CSS changes, so you don't need a manual rebuild while iterating.
Notes
* Animations respect prefers-reduced-motion and only animate opacity/transform (GPU-friendly).


noora/lib/noora/breadcrumbs.ex
```diff
@@ -101,10 +102,11 @@ defmodule Noora.Breadcrumbs do
       data-on-pointer-down-outside={@on_pointer_down_outside}
       data-on-focus-outside={@on_focus_outside}
       data-on-interact-outside={@on_interact_outside}
-      data-positioning-offset-main-axis={6}
+      data-positioning-offset-main-axis={8}
+      data-positioning-anchor="selector"
```
pepicrft: This points the dropdown positioning at `data-part="selector"`, but this component no longer renders a `data-part="selector"` element. The selector icon is now a `Noora.Icon.morphing_icon` inside the `data-part="trigger"` button, and that SVG does not set `data-part`. Because of that, `Dropdown.js` falls back to `null` from `getAnchorRect`, so the custom anchor never applies and the menu is not anchored to the split trigger as intended. This should either use `data-positioning-anchor="trigger"` or restore a real `data-part="selector"` wrapper around the icon.

noora/js/IconMorph.js
```diff
@@ -0,0 +1,324 @@
+// Filled-path icon morphing, ported from the flubber-style sampling approach.
```
pepicrft: This feels too heavy for the breadcrumb selector animation. The hook adds a custom path parser, path sampling, ring matching, spring integration, a global measuring SVG, and mutation observation for a two-state icon change, with no tests around the geometry or lifecycle. That makes the component harder to maintain than the rest of the Noora hooks, which mostly delegate state to Zag and keep local JS small. Can we simplify this to a CSS/SVG transition, for example two stacked icons with opacity/transform, or add focused tests if we really need path morphing as a reusable primitive?

server/mix.lock
```diff
@@ -144,7 +144,7 @@
...
+  "phoenix_live_view": {:hex, :phoenix_live_view, "1.1.31", "c45c85df509dd79c917bc530e26c71299e3920850f65ea52ab6a19ccee66875a", [:mix], [{:igniter, ">= 0.6.16 and < 1.0.0-0", [hex: :igniter, repo: "hexpm", optional: true]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: true]}, {:lazy_html, "~> 0.1.0", [hex: :lazy_html, repo: "hexpm", optional: true]}, {:phoenix, "~> 1.6.15 or ~> 1.7.0 or ~> 1.8.0-rc", [hex: :phoenix, repo: "hexpm", optional: false]}, {:phoenix_html, "~> 3.3 or ~> 4.0", [hex: :phoenix_html, repo: "hexpm", optional: false]}, {:phoenix_template, "~> 1.0", [hex: :phoenix_template, repo: "hexpm", optional: false]}, {:phoenix_view, "~> 2.0", [hex: :phoenix_view, repo: "hexpm", optional: true]}, {:plug, "~> 1.15", [hex: :plug, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4.2 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}], "hexpm", "2f53cc6a9e149f30449341c2775990819d97e3b22338fe719c4d30342e6f9638"},
```
pepicrft, line 147: This lockfile bump looks unrelated to the breadcrumb/Noora UI work. It updates `phoenix_live_view` and `thousand_island` without any matching `mix.exs` change or explanation in the PR, which widens the review and deployment risk for a visual component change. Can we revert these lockfile changes unless they are intentionally required, and if they are required, split or explain that dependency update explicitly?

pepicrft (approved):
Approved with a few cleanup comments around the breadcrumb anchoring, animation complexity, and an unrelated lockfile bump.

noora/js/IconMorph.js
asmitbm: I’ll be using this hook with other icons across the dashboard as well. The breadcrumbs component was simply the first place where I used it.

fortmarek:
A couple of concrete things I'd address before merging:

- `noora/lib/noora/breadcrumbs.ex:124`: the dropdown trigger is now an icon-only `<button>` with no accessible name. Previously the trigger contained the visible breadcrumb label, so screen readers had something to announce. We should add an `aria-label` or visually hidden label for the switcher button.

- `noora/lib/noora/icon.ex:63` / `noora/lib/noora/icon.ex:117`: `morphing_icon` looks like a general-purpose Noora icon primitive, but the implementation only seems safe for compatible filled-path icons. Some existing icons are stroked and/or multi-path, so using this broadly could render incorrect filled shapes. If this is intended for multiple dashboard interactions, I think we should make the support contract explicit or validate unsupported icons instead of letting them fail visually.

On the API shape: since this is planned for more than the breadcrumb selector, I'd prefer folding this into the existing `.icon` component rather than introducing a separate `morphing_icon` component. I would still keep animation opt-in rather than making morphing the default behavior.

For example:

```elixir
<.icon name="check" />

<.icon
  id="project-switcher-icon"
  name="selector"
  active_name="selector_2"
  transition="morph"
  watch="[data-part='trigger']"
  active_state="open"
/>
```

Or even allow cheaper strategies for most UI cases:

```elixir
<.icon name="selector" active_name="selector_2" transition="crossfade_rotate" />
<.icon name="selector" active_name="selector_2" transition="morph" />
```

That keeps `.icon` as the entrypoint, makes animation a first-class capability of icons, and lets Noora choose or validate the right strategy. We could also support `transition="auto"`, where known-compatible pairs use morphing and everything else falls back to a crossfade/rotate transition.

So my take is: yes to a reusable primitive if there are multiple planned uses, but I'd make it a deliberate `.icon` API with an explicit transition strategy and compatibility checks rather than a general `morphing_icon` helper.

pepicrft (approved):

fortmarek (approved):

PR #11043, opened 2026-06-02
chore(kura): repair e2e shellspec suite and add test runner tasks

esnunes:
## What changed

Repairs the Kura `shellspec` e2e suite (which **hung indefinitely** and reported failures even though every spec passed in isolation), and then hardens the harness so the suites are isolated and can run in parallel. Concretely:

- **Suite repair** — fix a container/port leak in `memory_pressure` teardown and a reporter crash caused by `docker compose exec` stealing stdin.
- **Test runner tasks** — add `test-unit` / `test-e2e` as file-based `mise` tasks.
- **Per-suite env scoping** (review feedback) — replace `export`/`unset` with a per-suite `--env-file`.
- **Parallelizable suite** — publish every container on an ephemeral host port (no manual port bookkeeping) and make `test-e2e` build once then fan out with `-j`.

## Suite repair

Running each spec individually passed, so the failures were *interactions* that only surfaced in the full run (shellspec runs files alphabetically). There were **two independent root causes**, both in the test harness:

**1. Container/port leak from `memory_pressure` teardown.**
`teardown_suite` unset `COMPOSE_PROJECT_NAME` and the compose-file interpolation vars *before* calling `compose_teardown`, so `docker compose down` targeted the wrong project (the directory-default `kura`) and left its 3 containers running. Those containers held the **default gRPC host ports 5101–5103** — this spec didn't override them, and most specs shared those defaults. Every spec that ran afterward then broke:
- **mtls** and **tenant_scope** → `setup_suite` failed fast on `Bind for 0.0.0.0:5101 failed: port is already allocated`
- **rollout** → spun in 180s `wait_for_http` retry loops (per example, twice), which is what presented as the suite "never ending"

Fix: reorder `teardown_suite` so `compose_teardown` runs *before* the env is torn down, letting the `down` target the correct, still-named project. (The env scoping below later removes the unsets entirely.)

**2. shellspec reporter crash from `docker compose exec`.**
Even after fixing #1, the suite ran green but aborted with `[reporter: 1]` → fatal exit 1. `docker compose exec` attaches the caller's **stdin** by default (unlike raw `docker exec`, which needs `-i`). Under shellspec, that inherited stdin is wired into the executor→reporter pipeline, so an attached `exec` corrupts the descriptor the reporter reads from and crashes it — even when the example passes. Triggered by rollout's three `dc exec` calls.

Fix: detach stdin (`</dev/null`) in the central `dc()` wrapper rather than per call site, since the behavior is intrinsic to `docker compose exec` regardless of where it's called. Verified no `dc` subcommand in the suite reads stdin, so it's safe.

## Test runner tasks

`test-unit` and `test-e2e` are defined as scripts under `kura/mise/tasks/` (the repo-wide convention used by `cache/`, `slack/`, `handbook/`, and the root `mise/tasks/`) rather than inline `[tasks.*]` in `mise.toml`.

- `test-unit` — runs `cargo test` with a raised file-descriptor limit; the RocksDB-backed unit tests open many fds in parallel and hit `EMFILE` under the default macOS `ulimit -n` of 256.
- `test-e2e` — builds the image once, then runs the shellspec suite (see parallelization below).

## Per-suite env scoping (review feedback)

Each spec used to `export` its compose configuration (project name, ports, resource limits) into the shell, and the `memory_pressure` spec had to `unset` it all again in teardown. Because shellspec runs every spec file in **one shared shell process**, that exported state leaked across specs — which is also what made the original teardown bug so destructive.

Now `dc()` passes `docker compose --env-file <per-suite file>`, and a new `suite_env NAME VALUE` helper sets a **non-exported** shell variable (for in-shell use such as building URLs) while recording `NAME=VALUE` in that per-suite file (created under the suite's temp dir). Values reach docker compose and the spec body without ever entering the process environment, so nothing leaks between spec files and the `export`/`unset` pattern is gone.

`local` isn't an option here: shellspec invokes `setup_suite`, each example, and `teardown_suite` as separate shell calls, so the values must persist beyond a single function scope — but they must not become process-wide `export`s. The `--env-file` indirection is what satisfies both.

## Parallelizable suite

shellspec parallelizes across spec **files** (`shellspec -j N`; within a file, examples stay serial). Concurrent suites would otherwise clash on published host ports — the base compose publishes a gRPC host port per node defaulting to `5101-5103` that most specs never overrode, plus several overlapping HTTP ports.

Rather than hand-assign a disjoint port range per spec (which has to stay unique as specs are added and can still collide with unrelated processes on the host), each spec **requests ephemeral host ports**: it records the relevant compose port variables as `0` (`ephemeral_ports`), so docker compose publishes them on a free host port chosen by the kernel, and reads the actual ports back after startup (`resolve_http_node` / `resolve_host_port`). The kernel guarantees uniqueness, so parallel suites can never clash. The compose files are **untouched**, so local `docker compose up` keeps its stable default ports; only the e2e env files set `0`.

An ephemeral host port is re-allocated on **every container start** — recreate, restart, and stop+start all change it — so specs re-resolve after each such operation. `cluster`, whose stateful examples share one stack and restart a node mid-run, re-resolves before every example via a `Before` hook, because shellspec isolates per-example variable state in subshells (a value re-resolved inside one example would not survive to the next).

The other thing that would serialize workers is each suite building its own image — compose names a built image `<project>-<service>`, and every suite uses a distinct `COMPOSE_PROJECT_NAME`. So the kura services declare `image: ${KURA_IMAGE:-}` (empty by default, so local `docker compose up` keeps its per-project naming), and the `test-e2e` task exports a fixed `KURA_IMAGE` and builds it once; every suite then resolves to that single prebuilt image and reuses it under `KURA_E2E_SKIP_BUILD=1`, with no per-project rebuilds. The task forwards arguments rather than hardcoding a job count:

```bash
mise run test-e2e            # serial
mise run test-e2e -- -j 8    # parallel
```

## Validation

- Full `shellspec` run **green serially (18 examples, 0 failures)** and under **`shellspec -j 4` (18/18, ~3.5x faster: 266s → 75s, single shared image, no per-project rebuilds, 0 leftover containers)**.
- Original suite repair confirmed with **0 leftover containers**; each previously-failing spec (mtls, rollout, tenant_scope) also passing in a clean environment.
- `suite_env` + `dc --env-file` verified end-to-end, including a check that a non-exported value from one suite does **not** leak into another suite's `docker compose`.
- Ephemeral allocation + `docker compose port` resolution verified against a live container; `bash -n` clean on all specs and tasks; `mise` argument forwarding confirmed (`-- -j 8` → `shellspec -j 8`).

### DIFF
--- a/kura/mise.toml
+++ b/kura/mise.toml
@@ -11,4 +11,4 @@ run = "ulimit -n 65536 && cargo test"
 
 [tasks.test-e2e]
 description = "Run the end-to-end shellspec suite against the Dockerized service"
-run = ["docker compose up --build -d", "shellspec"]
+run = "shellspec"
--- a/kura/mise.toml
+++ b/kura/mise.toml
@@ -4,3 +4,11 @@ shellspec = "0.28.1"
 bazel = "9.1.0"
 "aqua:facebook/buck2" = "2026-04-15"
 node = "25.9.0"
+
+[tasks.test-unit]
+description = "Run the Rust unit tests with a raised file-descriptor limit (RocksDB-backed tests open many fds in parallel)"
+run = "ulimit -n 65536 && cargo test"
+
+[tasks.test-e2e]
+description = "Run the end-to-end shellspec suite against the Dockerized service"
+run = ["docker compose up --build -d", "shellspec"]
--- a/kura/mise.toml
+++ b/kura/mise.toml
@@ -4,11 +4,3 @@ shellspec = "0.28.1"
 bazel = "9.1.0"
 "aqua:facebook/buck2" = "2026-04-15"
 node = "25.9.0"
-
-[tasks.test-unit]
-description = "Run the Rust unit tests with a raised file-descriptor limit (RocksDB-backed tests open many fds in parallel)"
-run = "ulimit -n 65536 && cargo test"
-
-[tasks.test-e2e]
-description = "Run the end-to-end shellspec suite against the Dockerized service"
-run = "shellspec"
--- a/kura/mise/tasks/test-unit.sh
+++ b/kura/mise/tasks/test-unit.sh
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+#MISE description="Run the Rust unit tests with a raised file-descriptor limit (RocksDB-backed tests open many fds in parallel)"
+set -euo pipefail
+
+ulimit -n 65536
+cargo test
--- a/kura/spec/e2e/memory_pressure_spec.sh
+++ b/kura/spec/e2e/memory_pressure_spec.sh
@@ -41,6 +41,7 @@ Describe 'memory pressure resilience'
   }
 
   teardown_suite() {
+    compose_teardown
     unset COMPOSE_PROJECT_NAME
     unset KURA_US_PORT KURA_EU_PORT KURA_AP_PORT
     unset KURA_US_URL KURA_EU_URL KURA_AP_URL
@@ -50,7 +51,6 @@ Describe 'memory pressure resilience'
     unset KURA_E2E_METADATA_STORE_READ_CACHE_BYTES
     unset KURA_E2E_METADATA_STORE_WRITE_BUFFER_POOL_BYTES
     unset KURA_E2E_METADATA_STORE_WRITE_BUFFER_BYTES
-    compose_teardown
   }
 
   BeforeAll 'setup_suite'
--- a/kura/spec/e2e/memory_pressure_spec.sh
+++ b/kura/spec/e2e/memory_pressure_spec.sh
@@ -4,30 +4,30 @@ Describe 'memory pressure resilience'
   Include spec/e2e/support.sh
 
   setup_suite() {
-    export COMPOSE_PROJECT_NAME="kura-memory-pressure"
-    export KURA_US_PORT=4701
-    export KURA_EU_PORT=4702
-    export KURA_AP_PORT=4703
-    export KURA_US_URL="http://localhost:${KURA_US_PORT}"
-    export KURA_EU_URL="http://localhost:${KURA_EU_PORT}"
-    export KURA_AP_URL="http://localhost:${KURA_AP_PORT}"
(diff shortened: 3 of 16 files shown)


kura/spec/e2e/memory_pressure_spec.sh
```diff
@@ -41,6 +41,7 @@ Describe 'memory pressure resilience'
   }
 
   teardown_suite() {
+    compose_teardown
     unset COMPOSE_PROJECT_NAME
```
pepicrft: I'd investigate if we can scope the variables to the test instead of using `set/unset`, which mutate a global state. I don't think we are parallelizing the test execution today, so mutating the global state is not an issue, but as we add more tests, that might be a thing down the line.

kura/mise.toml
```diff
@@ -4,3 +4,11 @@ shellspec = "0.28.1"
 bazel = "9.1.0"
 "aqua:facebook/buck2" = "2026-04-15"
 node = "25.9.0"
+
+[tasks.test-unit]
```
pepicrft: Mise supports defining tasks as scripts, so I'm inclined to do that instead. I believe they are defined un `kura/.mise/tasks`

pepicrft (approved):

kura/mise/tasks/test-unit.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+#MISE description="Run the Rust unit tests with a raised file-descriptor limit (RocksDB-backed tests open many fds in parallel)"
+set -euo pipefail
+
+ulimit -n 65536
```
pepicrft, line 5: why is this needed?

pepicrft (approved):

kura/mise/tasks/test-unit.sh
esnunes: on macos the number of files a process can open by default is 256, this was causing some tests (RocksDB-backed) to fail

PR #11045, opened 2026-06-02
fix(server): add id tiebreaker to test case runs pagination

justine-acorns:
## What

Add `:id` as a secondary sort key to the test case runs list endpoint pagination.

## Why

The endpoint orders by `ran_at` only. When multiple rows share the same `ran_at` timestamp (common in parallel test runs), ClickHouse reshuffles their relative order between page fetches. This causes:

- Duplicate rows appearing on adjacent pages
- Rows that never appear on any page (skipped during the reshuffle)

In practice, a test run with 5681 test cases only returns ~4254 unique rows through pagination, with ~1427 rows lost. This means downstream consumers (like our CI test reporting) miss quarantined test failures that should be filtered from results.

## How

Add `:id` to `order_by` and `:asc` to `order_directions` in `render_test_case_runs`, matching the pattern already used in `apply_quarantined_order` which has a comment explaining the exact same issue (lines 2396-2399 of `tests.ex`).

Also added `:id` to the `sortable` list in `TestCaseRunByTestRun` and `TestCaseRunByProject` schemas so Flop accepts it.

### DIFF
--- a/server/lib/tuist/tests/test_case_run.ex
+++ b/server/lib/tuist/tests/test_case_run.ex
@@ -28,7 +28,7 @@ defmodule Tuist.Tests.TestCaseRun do
       :shard_id,
       :shard_index
     ],
-    sortable: [:inserted_at, :duration, :name, :ran_at]
+    sortable: [:inserted_at, :duration, :name, :ran_at, :id]
   }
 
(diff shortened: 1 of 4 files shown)


pepicrft (approved):

fortmarek (approved):
This is great, thanks!

PR #11047, opened 2026-06-03
feat(server): support OAuth service tokens

pepicrft:
## What

Adds first-class support for OAuth client-credentials service tokens in the server authentication pipeline.

The change introduces an `AuthenticatedService` subject, teaches Guardian how to sign and resolve it, updates the Boruta token generator to distinguish user tokens from service tokens, and lets plugs and authorization checks handle service-authenticated requests.

## Why

Internal services should be able to authenticate to Tuist with short-lived OAuth tokens without minting or storing one token per customer account. This PR keeps the scope to authentication plumbing only, so a follow-up can add whichever API surface needs to consume the service subject.

## Notes

Service tokens are only issued for explicitly configured static OAuth service clients. Dynamically registered clients and the regular Tuist OAuth client do not become service subjects just because they use client credentials.

The implementation is generic: configured service clients carry a client id, secret, optional display name, optional TTL, and explicit scopes in the generated token. No downstream service name is encoded in the implementation.

> [!NOTE]
> Scopes should describe the resource and action, for example `account:usage:read`. The account boundary comes from the authenticated subject: an account token with that scope can be limited to its own account with `accounts_match`, while a statically configured service token with the same scope can be authorized by policy to read that resource across accounts. This avoids introducing separate `:any` scopes while keeping cross-account access explicit in authorization rules.

> [!IMPORTANT]
> Service-token usage is intended to be cluster-internal. The follow-up API surface should be protected with network rules so only workloads inside the cluster can reach it. OAuth client credentials and short-lived tokens remain the application-level control, while the network boundary limits exposure if a token or client secret is leaked.

## Validation

- `mise exec -- mix format --check-formatted`
- `mise exec -- mix test test/tuist/oauth/clients_test.exs test/tuist/oauth/token_generator_test.exs test/tuist_web/authentication_test.exs test/tuist/oauth/introspection_test.exs test/tuist_web/rate_limit/mcp_test.exs test/tuist_web/plugs/sentry_context_plug_test.exs test/tuist/analytics_test.exs test/tuist_web/plugs/api/authorization/authorization_plug_test.exs`


fortmarek:
## Findings

[P1] Service-token issuance should be constrained to the management-cluster boundary.

This PR adds bearer service tokens that are intentionally cross-account once authorized by policy. Since bearer tokens are portable after minting, OAuth client credentials alone do not prove that the caller is still Atlas, Tuist Server, or even running inside the same management cluster. For defense in depth, both service-token issuance and any API surface that accepts `AuthenticatedService` should be reachable only through the internal management-cluster path. Concretely, I would keep these routes off the public ingress and enforce the boundary with Kubernetes `NetworkPolicy` and/or an internal Gateway route. The stronger version is to require workload identity too, for example mTLS/service-mesh identity or a projected Kubernetes service-account JWT with an audience for Tuist Server, so the server verifies the expected namespace/service account in addition to the OAuth token.

[P1] Service detection can diverge from the client Boruta actually authenticated.

`TokenGenerator.generate/2` decides to issue a service token with `Clients.service_client?(client_id)`, but `Clients.get_client/1` resolves built-in static clients before service clients. If the configured service-client list accidentally reuses the Tuist OAuth client id, Boruta authenticates the Tuist client while the token generator still signs an `AuthenticatedService` token. That would bypass the intended service-client secret, TTL, and scope configuration. Please either reject duplicate service-client ids at config parsing time or make service-token generation depend on the same resolved client identity/order that Boruta authenticated.

[P2] Service tokens can pass `require_authentication` and then crash account-owned endpoints.

`authenticated_subject_account/1` now returns `nil` for `AuthenticatedService`, but several authenticated API paths still assume every authenticated non-user subject has an owning account. For example, `OrganizationsController.index/2` preloads `authenticated_subject_account(conn)` and then dereferences `account.organization`; a valid service token hitting `GET /api/organizations` would satisfy authentication and then raise instead of returning a controlled 403/401. Please explicitly reject service subjects on account-required routes, or update those callers to handle `nil` before dereferencing/storing the account.

pepicrft:
Thanks Marek, all three are good catches.

**Service detection diverging from Boruta** (P1): fixed in fca511f61b. The root issue was that `static_client/1` resolves the built-in CLI/Kura clients before service clients, so a service-client config reusing one of those ids would have Boruta authenticate the built-in client while `service_client?/1` still treated it as a service client. I now drop any service-client config whose id collides with a reserved built-in id, so client resolution, `service_client?/1`, and `authorized_scopes/1` all agree with what Boruta authenticated.

**Service tokens crashing account-owned endpoints** (P2): also fixed in fca511f61b. The `authenticated_api` pipeline now rejects service subjects with a 403 right after `require_authentication`. No route on that surface is meant for service tokens today, and their authorization goes through the `let_me` policy layer separately, so this kills the whole nil-account crash class at one chokepoint instead of guarding each caller.

**Management-cluster boundary** (P1): I agree, but it's an infra change rather than something for this PR. Keeping these routes off the public ingress plus a `NetworkPolicy`/internal Gateway route, and ideally workload identity on top, lives in the Helm/infra layer. I'll follow up with that separately so we don't block this on it.

server/lib/tuist/accounts/authenticated_service.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Tuist.Accounts.AuthenticatedService do
```
fortmarek: [P2] Avoid making Atlas a generic accountless service subject

I am worried this introduces a new first-class Guardian subject without a crisp resource boundary. If the immediate motivation is Atlas/customer-experience context, a static OAuth service client feels broader than the thing we actually need:

- It lives in the same auth path as user/account tokens, so every future `current_subject`/policy path has to remember to reject or special-case service subjects.
- Its scopes are not tied to deployment identity, namespace, audience, pod lifetime, or a resource grant.
- The client secret becomes a long-lived credential to distribute and rotate, while token minting sits behind the public OAuth token endpoint unless separately fenced.
- Tenant safety stays implicit: a later tenant-scoped service permission could accidentally become cross-tenant unless every check reasserts tenancy.

I would prefer modeling Atlas like a workload identity rather than a product OAuth client. Concretely, we could expose a curated internal Atlas endpoint/read model, accept a projected Kubernetes service-account token with an audience such as `tuist-server`, validate it through TokenReview or the cluster OIDC JWKS, assert the expected namespace/service account (and optionally pod UID), then map that to a single capability like `atlas:customer_context:read`. A NetworkPolicy can also restrict the route so only the Atlas workload reaches it. That matches the k8s/GKE/EKS/Vault/SPIFFE direction: short-lived, deployment-bound identity; no static OAuth client secret; and authorization expressed as a narrow workload capability.

An even tighter option would be a push/read-model flow where Tuist publishes the customer context Atlas needs, instead of giving Atlas a broadly reusable API principal. If we still want a reusable abstraction, I would name/model it as `AuthenticatedWorkload` or `InternalPrincipal` and keep it on an internal auth pipeline. The important bit is that it should not look like a normal generic OAuth subject unless we really intend third-party machine clients to use it.

pepicrft: You're right. I was treating this like a reusable OAuth M2M client, but Atlas is an internal workload, so it should use workload identity instead of a generic accountless Guardian subject.

I pushed ff9a584d98, 05847f1d93, and the follow-up 0ad1401096 after checking the Atlas production cluster:

- Removed `AuthenticatedService` and the configured OAuth service-client path, so `/oauth2/token` no longer mints service principals.
- Added a curated internal Atlas usage read model at `GET /api/internal/atlas/accounts/:account_handle/usage`.
- Atlas runs in a separate Kubernetes cluster, so Tuist cannot validate Atlas tokens with Tuist's own TokenReview API. The endpoint now verifies Atlas projected ServiceAccount JWTs offline against the pinned Atlas Kubernetes JWKS, checks the `tuist-server` audience, and gates the subject to the `atlas-production/atlas` ServiceAccount.
- Pinned the current Atlas public JWKS in the managed production values. It is public verification material, not a secret, and avoids trusting the generic kubeadm issuer string by itself.
- Kept this outside the public authenticated API pipeline, so there is no generic cross-tenant service subject for future policies to accidentally accept.

Validation: `mix test test/tuist/atlas_workload_identity_test.exs test/tuist_web/controllers/internal/atlas_usage_controller_test.exs test/tuist/kubernetes/client_test.exs` passed with 24 tests, 0 failures. I also rendered the production Helm chart with dummy release-time image pins to verify the Atlas env block renders correctly.

pepicrft: Closing this out. The final shape removes the generic OAuth service subject entirely and narrows this to the Atlas read model endpoint.

The important follow-up from checking the Atlas cluster is that Atlas runs in a separate Kubernetes cluster, so Tuist cannot use its own TokenReview API for Atlas tokens. The endpoint now verifies Atlas projected ServiceAccount JWTs against the pinned Atlas public JWKS, checks the `tuist-server` audience, and gates the subject to `system:serviceaccount:atlas-production:atlas`.

This leaves the public ingress question explicit: the route is not network-private today, but it is no longer backed by a static OAuth client secret or a reusable accountless Guardian subject. The next Atlas-side hardening is to mount a dedicated short-lived projected token with `audience: tuist-server` and have Atlas use that for this call.

pepicrft:
@fortmarek quick update on the Atlas/service-token feedback.

I ended up taking your suggestion and narrowed this away from generic OAuth M2M/service subjects:

- Removed `AuthenticatedService` and the configured OAuth service-client token path, so `/oauth2/token` does not mint accountless service principals.
- Added a curated Atlas read model at `GET /api/internal/atlas/accounts/:account_handle/usage`. It takes the account handle because Atlas already has that handle.
- Kept the Atlas endpoint outside the normal authenticated API pipeline, so user/account policy code does not need to special-case a reusable cross-tenant service subject.
- Checked the Atlas production cluster. Atlas runs in a separate Kubernetes cluster, so Tuist cannot call its own TokenReview API for these tokens. The server now verifies Atlas projected ServiceAccount JWTs against the pinned Atlas public JWKS, checks the `tuist-server` audience, and gates access to `system:serviceaccount:atlas-production:atlas`.
- Pinned the current Atlas public JWKS in managed production values. This is public verification material, not a secret.
- The route is still not network-private at the ingress layer today. The important change here is that it is no longer protected by a static OAuth client secret or generic accountless principal. The Atlas-side hardening that remains is mounting a dedicated short-lived projected token with `audience: tuist-server` for this call.

Validation: `mix test test/tuist/atlas_workload_identity_test.exs test/tuist_web/controllers/internal/atlas_usage_controller_test.exs test/tuist/kubernetes/client_test.exs` passed with 24 tests, 0 failures, and I rendered the production Helm chart with dummy release-time image pins to verify the Atlas env block.

server/lib/tuist/atlas_workload_identity.ex
```diff
@@ -0,0 +1,142 @@
...
+  TokenReview API to authenticate Atlas tokens. Instead, Tuist verifies the
+  JWT signature against the pinned Atlas Kubernetes JWKS, checks the configured
+  issuer and audience, and returns the ServiceAccount principal from `sub`.
+  """
+
+  alias Tuist.Environment
+
+  @clock_skew_seconds 60
+
+  def verify(token) when is_binary(token) and token != "" do
```
fortmarek: [P2] Make Atlas workload auth an explicit internal pipeline/trust policy

This is a much better direction than the generic `AuthenticatedService`/OAuth service-client shape, and it matches the reality that Atlas and Tuist are workload identities rather than Tuist account subjects. I would still tighten the design a bit so this becomes a reusable cross-cluster workload boundary instead of controller-local token verification.

Because Atlas and Tuist are under the same management plane but not the same workload cluster, Tuist cannot rely on its own apiserver TokenReview for Atlas tokens. Offline JWKS validation is reasonable, but I would model it as an explicit trust policy: issuer, audience, JWKS, expected namespace, expected ServiceAccount, and max token TTL. Ideally that trust bundle is rendered/reconciled by the management plane from Atlas cluster state rather than being a set of unrelated env vars in the server deployment.

Concretely, I would reshape this as:

- `Environment.atlas_workload_identity_policy/0` returning one policy object, including `max_token_ttl_seconds`.
- `AtlasWorkloadIdentity.verify/2` validating signature, issuer, audience, `sub`, Kubernetes private claims, required `iat`, and `exp - iat <= max_token_ttl_seconds`.
- A small `InternalAtlasAuthPlug` that extracts the bearer token, verifies the workload, assigns `:atlas_principal`, and centralizes 401/503 handling.
- A dedicated router pipeline such as `:atlas_internal_api` with JSON accepts, Atlas auth, and an endpoint-specific rate-limit plug, rather than plain `:non_authenticated_api`.
- A small domain/read-model function like `Tuist.Atlas.customer_context(account_handle)` so the controller stays a thin HTTP adapter as Atlas asks for more customer context.

That keeps the current k8s-native direction, but makes the boundary more explicit: cross-cluster workload trust plus a curated internal read model, not a normal product API subject and not one-off auth logic embedded in a controller.

fortmarek (approved):

server/lib/tuist/atlas_workload_identity.ex
pepicrft: Addressed in 58121fad25.

I reshaped this into an explicit Atlas internal workload boundary:

- `Environment.atlas_workload_identity_policy/0` now returns one trust policy with issuer, audience, JWKS, expected namespace, expected ServiceAccount, and max token TTL.
- `AtlasWorkloadIdentity.verify/2` now takes that policy and validates the signature, issuer, audience, required `iat`, expiration, `exp - iat <= max_token_ttl_seconds`, `sub`, and the Kubernetes private ServiceAccount claims.
- Added `TuistWeb.Plugs.InternalAtlasAuthPlug` and a dedicated `:atlas_internal_api` router pipeline, so bearer extraction, 401/503 handling, and `:atlas_principal` assignment are centralized outside the controller.
- Added `TuistWeb.Plugs.AtlasRateLimitPlug` plus `TuistWeb.RateLimit.Atlas` for an Atlas-specific internal endpoint rate limit.
- Added `Tuist.Atlas.customer_context/1`, keeping `AtlasUsageController` as a thin HTTP adapter over the curated read model.
- Set `TUIST_ATLAS_TOKEN_MAX_TTL_SECONDS=3600` in the production Helm values to make the current trust policy explicit.

Validation: `mix test test/tuist/atlas_workload_identity_test.exs test/tuist/atlas_test.exs test/tuist_web/controllers/internal/atlas_usage_controller_test.exs test/tuist_web/rate_limit/atlas_test.exs` passed with 21 tests. I also ran focused `mix format --check-formatted`, `git diff --check`, and rendered the production chart with `helm template tuist infra/helm/tuist -f infra/helm/tuist/values-managed-common.yaml -f infra/helm/tuist/values-managed-production.yaml -f infra/helm/tuist/values-ci.yaml`.

PR #11079, opened 2026-06-04
docs(server): update self-host Docker Compose to use Kura

pepicrft:
Resolves N/A

This PR updates the self-host Docker Compose quick start so it runs the Tuist server with Kura instead of the legacy cache-node path. The compose bundle has no Redis or legacy cache-server service. The local cache-node service is Kura only.

It adds a downloadable self-host bundle under `server/priv/static/server/self-host/` with Postgres, ClickHouse, ClickHouse Keeper, MinIO, Kura, pgweb, and Tuist wired together. The Tuist container points at Kura through the Docker network via `http://kura:4000`, while the docs still expose the host-facing Kura HTTP and gRPC ports for local inspection. It also exposes the `server/` static download path through `TuistWeb.static_paths/0` and updates the self-host control-plane guide to download the new files.

The validation surfaced that the release start script fails on a fresh database because `schema_migrations` does not exist yet. This PR makes that lookup tolerant so first-time self-hosted compose runs can reach the migration step.

### How to test locally

- `docker --context=default buildx build --builder tuist-serial --platform linux/arm64 --load -f server/Dockerfile --build-arg TUIST_HOSTED=0 --build-arg TUIST_VERSION=selfhost-test -t tuist-selfhost:test .`
- `docker --context=default compose -p "$project" --env-file "$tmpdir/.env" -f "$tmpdir/docker-compose.yml" config --services`, confirming the services are Postgres, ClickHouse Keeper, ClickHouse, Kura, MinIO, MinIO init, Tuist, and pgweb
- `docker --context=default compose -p "$project" --env-file "$tmpdir/.env" -f "$tmpdir/docker-compose.yml" up -d --wait` with a valid self-host license and the locally built Tuist image
- Health checks passed for Tuist `/ready`, Kura `/up`, ClickHouse `/ping`, MinIO readiness, and Postgres `pg_isready`
- `sh -n server/rel/overlays/bin/start`
- `git diff --check`

### DIFF
--- a/server/lib/tuist_web.ex
+++ b/server/lib/tuist_web.ex
@@ -23,7 +23,7 @@ defmodule TuistWeb do
 
   def static_paths,
     do:
-      ~w(assets fonts images favicon.ico favicon-16x16.png favicon-32x32.png apple-touch-icon.png android-chrome-192x192.png android-chrome-512x512.png site.webmanifest browserconfig.xml mstile-150x150.png js css .well-known marketing app apidocs skills docs)
+      ~w(assets fonts images favicon.ico favicon-16x16.png favicon-32x32.png apple-touch-icon.png android-chrome-192x192.png android-chrome-512x512.png site.webmanifest browserconfig.xml mstile-150x150.png js css .well-known marketing app apidocs skills docs server)
 
   def router do
     quote do
--- a/.github/workflows/gradle-cache-acceptance.yml
+++ b/.github/workflows/gradle-cache-acceptance.yml
@@ -44,7 +44,7 @@ jobs:
     if: github.event.pull_request.draft == false || github.event_name != 'pull_request'
     services:
       postgres:
-        image: postgres:16
+        image: public.ecr.aws/docker/library/postgres:16
         env:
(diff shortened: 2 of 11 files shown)


server/lib/tuist_web.ex
```diff
@@ -23,7 +23,7 @@ defmodule TuistWeb do
 
   def static_paths,
     do:
-      ~w(assets fonts images favicon.ico favicon-16x16.png favicon-32x32.png apple-touch-icon.png android-chrome-192x192.png android-chrome-512x512.png site.webmanifest browserconfig.xml mstile-150x150.png js css .well-known marketing app apidocs skills docs)
+      ~w(assets fonts images favicon.ico favicon-16x16.png favicon-32x32.png apple-touch-icon.png android-chrome-192x192.png android-chrome-512x512.png site.webmanifest browserconfig.xml mstile-150x150.png js css .well-known marketing app apidocs skills docs server)
```
pepicrft, line 26: The docker-composer configuration files live at `static/server/...`

fortmarek (approved):

PR #11082, opened 2026-06-04
feat(kura): Bazel build foundation — cross-compiled Linux binaries + shadow CI

esnunes:
## Summary

Lays the Bazel foundation for building Kura, producing both Linux binaries
(`x86_64` and `aarch64`) from a single host with no QEMU, and wires it into CI as a
**non-gating shadow** alongside the existing Cargo/Docker release path. **Nothing in
the release path changes** — this ships no user-facing change; it is the groundwork
the headline win (a Bazel-built multi-arch OCI image, Phase 3) builds on.

Full design + decisions: `kura/docs/bazel-migration-plan.md`.

## Why

`release-kura-docker` builds `kura/Dockerfile` for `linux/amd64,linux/arm64` on the
x86 `tuist-linux` runner, which means **arm64 is built under QEMU emulation** — slow
and a recurring source of flakiness. The end goal is to assemble a multi-arch image
from two natively-cross-compiled binaries (Bazel → `rules_oci`), deleting QEMU/Buildx.
That requires first proving every native dependency builds hermetically under Bazel
and that one host can emit both arches. This PR does exactly that.

## What changed

All additions under `kura/` (plus one new workflow); no existing build files modified.

- **Workspace + toolchain** — `MODULE.bazel`, `.bazelrc`, `.bazelversion` (Bazel 9.1.0),
  `rules_rust` 0.70.0 pinned to Rust 1.94.1 / edition 2024, `crate_universe`
  (`from_cargo`) generating `@crates` from the existing `Cargo.toml`/`Cargo.lock` with a
  committed `Cargo.Bazel.lock`. `BUILD.bazel` wires `//:kura` (lib + bin).
- **Native `-sys` crates build hermetically** — `librocksdb-sys` (C++ + bindgen + cmake
  + bundled lz4), `tikv-jemalloc-sys`, `ring`, `aws-lc-sys` all compile in-sandbox; lz4
  /rocksdb/jemalloc/lua link statically, so the runtime needs only
  libc/libm/libstdc++/libgcc_s.
- **Cross-compilation, no QEMU** — `//bazel/toolchains/cc:{x86_64,arm64}_linux_toolchain`
  over the Debian Bookworm cross GCC (gcc 12, glibc 2.36 — matching the runtime image
  exactly). Each cross toolchain is scoped by **both** `exec_compatible_with` and
  `target_compatible_with` so it is selected only when host arch ≠ target arch; native
  builds keep the autodetected host cc.
- **Shadow CI** — `.github/workflows/kura-bazel.yml` + `kura/bazel/ci-validate.sh`: a
  2-runner matrix (`ubuntu-latest`, `ubuntu-24.04-arm`) each builds **both** arches in
  the Bookworm dev container, validating ELF machine type, a glibc floor ≤ 2.36, and a
  native empty-env smoke run; binaries are uploaded as artifacts.
- **Local dev parity** — `bazel/linux-dev.Dockerfile` + `bazel/linux-build.sh` reproduce
  the exact CI environment from a macOS host via Docker.

## Key decisions (and why, over the obvious alternatives)

- **Lua via a `cc_library`, not a patch.** `lua-src`'s build script bakes
  `env!("CARGO_MANIFEST_DIR")` (a compile-time sandbox path) that no longer exists when
  `mlua-sys` runs, so it can't find the vendored `lua-5.4.8/`. Rather than patch
  third-party source, we expose those sources as filegroups and compile Lua as
  `//bazel/third_party/lua:lua`, disabling only the now-redundant `mlua-sys` build
  script. **No third-party source is modified.**
- **Debian cross GCC, not `toolchains_llvm`.** LLVM would pull clang/libc++ and an
  external sysroot whose libstdc++ could be too old for rocksdb's C++17 — diverging from
  the gcc/libstdc++ native build. The Debian cross GCC is the same gcc/glibc as native,
  matching the Bookworm runtime.
- **`bookworm` floor only.** Per the runtime decision, we don't build against an older
  glibc; build-glibc (2.36) == runtime-glibc.

## Impact

- **Developer:** a reproducible, hermetic Bazel build for Kura and the binaries the OCI
  phase will consume. Local dev mirrors CI via the dev container.
- **Release / users:** none. The shadow workflow is non-gating and the Cargo/Docker
  release path is untouched — rollback is deleting one workflow file.

## Validation

- Ran `bazel/ci-validate.sh` end-to-end in the Bookworm container locally (arm64 host):
  both binaries built, correct ELF types, glibc floor 2.34, native smoke run clean.
- **CI run [26933769463](https://github.com/tuist/tuist/actions/runs/26933769463) — both
  jobs green.** This is the first validation of the **arm64-from-x86_64 cross build on
  real x86_64 hardware** (it can't run on an Apple-silicon dev machine). Per job, both
  binaries: `ELF64` with the right machine type, glibc floor **2.34 ≤ 2.36**, native
  binary depends only on base libs and exits cleanly with the expected missing-config
  error.

## Out of scope (follow-ups)

macOS tarballs (Phase 2), the **OCI image that deletes QEMU** (Phase 3 — the actual
win), wiring into `release-kura-docker`, and a remote/disk cache to cut the ~45 min cold
build (Phase 4). See the plan doc.


esnunes:

PR #11091, opened 2026-06-04
feat(infra): rank queued load above idle warm floors in runner fleet allocation

fortmarek:
## What changed

Re-tiers the Linux runner fleet allocator (`AllocateFleet` in `infra/runners-controller/internal/scaling/allocate.go`) so a pool's real queued load takes precedence over other pools' idle warm-pool floors when they contend for the shared bare-metal node pool's memory.

## Why

Linux runner shapes (`2vcpu-8gb`, `4vcpu-16gb`, ...) all run as kata microVMs on the same bare-metal node pool and bin-pack on memory. The allocator distributes that memory across shapes, but it treated each pool's `minWarmPoolFloor` as an inviolable tier:

```
old: (1) floor  (2) load  (3) p95 headroom    -- tiers 1+2 always granted
```

So every shape's floor was funded before any shape's real queued work. In production this produced the exact pathology the change targets: a small shape pinned at a large warm floor (mostly idle pods) held a big chunk of fleet memory, while a larger shape's real queued jobs sat `Pending` because no memory was left. Observed live: a small shape sitting at floor with ~20 idle pods (no jobs), next to a larger shape with real queued load and ~10 pods Pending. Idle warm capacity was outranking another shape's actual queued work.

## The change

```
new: (1) load  (2) floor above load  (3) p95 headroom
```

Only tier 1 (real load = `claimed + queued`) is inviolable and granted in full, even past capacity (the excess goes Pending, the operator's "add a host" signal). Floors and the p95 buffer are idle warm capacity and now yield under contention, headroom first then floor, to admit another pool's queued work. An idle shape's `desired` can drop below its configured floor; its idle pods are then reaped to free memory for the starved shape.

- Raising per-shape floors / shuffling helm values is whack-a-mole across 9 shapes and does not encode the policy (demand beats speculation).
- Native k8s priority + preemption reaches the same outcome but adds sharp edges: pod priority is immutable (cannot promote idle to busy on claim) and a claim-during-preemption race. The allocator already owns cross-pool sizing, so expressing the policy there is simpler and race-free.
- The fleet allocator and its node-memory gathering already exist; this is a focused re-tiering of the pure allocation function, no new inputs, CRD, RBAC, or config knob.

## Busy-pod safety

The reaper is unaffected. `RunnerPoolReconciler` scale-down iterates only `idleAlive`, built solely from pods where `isIdle()` is true (no `tuist.dev/runner-pool-owner` label). A mid-job pod always carries that label and is never eligible for deletion, regardless of how far below floor `desired` drops. Lowering `desired` only increases the count of idle warm pods reaped; it can never reap a running job.

## Tradeoff

Under sustained load on one shape, other shapes' warm pools shrink toward their real load, so a returning spike on a squeezed shape pays cold-start instead of landing on a warm pod. This is deliberate: a job queued now beats a warm pod held for a job that might arrive. The existing per-pool scale-down cooldown damps the reap so it does not thrash.

## Validation

- `go build ./...`, `go test ./...`, `go vet ./...`, `gofmt` all clean.
- Added `TestAllocateFleet_SqueezesIdleFloorForAnotherPoolsQueuedLoad`, a regression test mirroring the production case (a 20-floor idle shape squeezed to fit a queued-load shape).
- Updated the zero-capacity test for the new semantics: real load is still honored (goes Pending, the add-a-host signal), a speculative floor with no load behind it is no longer manufactured into Pending pods.
- The existing cross-pool reclaim controller test still passes unchanged (when the idle floor fits, only headroom squeezes, same result).

## Out of scope

Draft because it is part of ongoing runner-fleet work. A separate investigation into runners dying mid-job ("self-hosted runner lost communication") is in progress; that is a VM-sizing / observability problem (guest-side resource exhaustion), orthogonal to this allocation policy.

### DIFF
--- a/infra/helm/tuist/crds/tuist.dev_runnerpools.yaml
+++ b/infra/helm/tuist/crds/tuist.dev_runnerpools.yaml
@@ -131,10 +131,17 @@ spec:
                       default: 1
                       minimum: 0
                       description: |
-                        Lower bound for the warm pool size. The server-side
-                        rolling p95 of concurrent claims over the last hour
-                        can lift the effective floor higher; this value
-                        only floors it, never caps it.
+                        Preferred lower bound (target) for the warm pool
+                        size. The server-side rolling p95 of concurrent
+                        claims over the last hour can lift the effective
+                        target higher. Not a hard floor: in a shared Linux
+                        fleet (multiple shape pools bin-packing on one
+                        bare-metal node pool), the allocator can squeeze an
+                        idle pool's desired below this value under memory
+                        contention to admit another shape's real queued
+                        work. macOS and uncontended pools honor it as a
+                        floor; real load (claimed + queued) is always funded
+                        above it.
                     maxReplicas:
                       type: integer
                       default: 0
--- a/infra/runners-controller/AGENTS.md
+++ b/infra/runners-controller/AGENTS.md
@@ -25,23 +25,30 @@ independent workqueues:
   `RunnerPool` spec, so a tuning change is helm-only.
 
   **Fleet-capacity awareness (Linux).** Linux shape pools share one
-  bare-metal node pool, so their speculative warm headroom competes for
-  the same memory. For `os: linux` pools the reconciler runs the
-  per-pool target through `internal/scaling/allocate.go`'s
-  `AllocateFleet`, a three-tier waterfall over the pools sharing a
-  `FleetSelector`: (1) every pool's `minWarmPoolFloor`, (2) real load
-  (`claimed + queued`), then (3) the speculative p95 buffer from
-  whatever memory is left. Tiers 1+2 are always honored (excess goes
-  Pending — the "add a host" signal); tier 3 is squeezed under
-  contention, so an idle shape's warm Pods fall back toward its floor to
-  admit another shape's real queued work. Memory is the only dimension
-  (kata pins it per microVM; CPU is oversubscribed). Fleet allocatable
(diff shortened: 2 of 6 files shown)


pepicrft (approved):
Thanks for tightening this policy. The main allocation change reads well to me.

One non-blocking note: `AllocateFleet` now documents and tests zero-capacity behavior where real load is preserved but speculative floor is not. The reconciler still falls back to the per-pool target on `fleetMem <= 0`, so a successful zero-memory fleet read keeps the old production behavior. I would either let the allocator handle `fleetMem == 0` when `err == nil`, or keep the fallback and update the zero-capacity test/comment to avoid documenting behavior the reconciler does not use.

Approved.

fortmarek:
Thanks! Addressed in 9312641.

I went with keeping the `fleetMem <= 0` fallback rather than routing zero into the allocator. A zero sum with `err == nil` is almost always a transient empty node-list read (informer blip, or no Ready nodes mid-roll), not a genuine zero-memory fleet. If `AllocateFleet` handled it, every pool's floor would squeeze to zero and reap warm Pods fleet-wide on that blip, which is the exact mass scale-down the `node-read blip must never trigger a mass scale-down` fallback exists to prevent. So the fallback is deliberate.

The real fix was the docs: added a comment at the guard explaining why zero is treated like an error, and reframed the zero-capacity test as `AllocateFleet`'s pure-function contract with a note that the reconciler never calls it with `fleetMem <= 0`. No logic change.

PR #11099, opened 2026-06-04
feat(cli): add opt-in SwifterPM install resolution

pepicrft:
No linked issue.

Adds an opt-in SwifterPM-backed path for `tuist install` dependency resolution. When `TUIST_USE_SWIFTERPM` is present, Tuist shells out to SwifterPM for `resolve` and `update`, preferring a bundled `vendor/swifterpm` next to the Tuist executable and falling back to `swifterpm` on `PATH`. The default behavior continues to use `swift package`.

This also pins SwifterPM 0.5.1 in mise, installs it in the CLI release workflow, bundles it into `tuist.zip`, and codesigns/notarizes it with the rest of the CLI release payload.

`tuist generate` now reuses the package-info JSON cache that SwifterPM writes under `.build/swifterpm/package-info` when the cache is present and fresh. That avoids shelling out to load every package manifest after a SwifterPM-backed install. If the cache is missing, has an unsupported schema version, cannot be decoded, or an entry is older than its `Package.swift`, generation falls back to the existing manifest loader.

### How to test locally

- `mise lock github:tuist/swifterpm --platform linux-x64,macos-arm64,macos-x64`
- `swiftformat cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift cli/Tests/TuistLoaderTests/Loaders/SwiftPackageManagerGraphLoaderTests.swift`
- `PATH="$(dirname "$(mise which swifterpm)"):$PATH" TUIST_USE_SWIFTERPM=1 <built-tuist> install`
- `PATH="$(dirname "$(mise which swifterpm)"):$PATH" TUIST_USE_SWIFTERPM=1 <built-tuist> generate --no-open`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme TuistUnitTests -only-testing TuistSupportTests/SwiftPackageManagerControllerTests -only-testing TuistLoaderTests/SwiftPackageManagerGraphLoaderTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -286,6 +286,8 @@ jobs:
       - name: Build tests
         id: shard
         run: tuist test --build-only --shard-total 2 --shard-granularity suite TuistAcceptanceTests
+        env:
+          TUIST_SHARD_REFERENCE: github-${{ github.run_id }}-cli-acceptance-tests
 
   cli-acceptance-tests:
     name: Acceptance Tests
@@ -347,6 +349,7 @@ jobs:
         run: tuist test --without-building --platform macOS TuistAcceptanceTests -- -retry-tests-on-failure -test-iterations 2
         env:
           TUIST_SHARD_INDEX: ${{ matrix.shard }}
+          TUIST_SHARD_REFERENCE: github-${{ github.run_id }}-cli-acceptance-tests
 
   cli-acceptance-tests-fork:
     name: Acceptance Tests (Fork)
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -197,7 +197,7 @@ jobs:
   cli-unit-tests:
     name: Unit Tests
     runs-on: tuist-macos
-    timeout-minutes: 30
+    timeout-minutes: 60
     permissions:
       contents: read
       id-token: write
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -197,7 +197,7 @@ jobs:
   cli-unit-tests:
     name: Unit Tests
     runs-on: tuist-macos
-    timeout-minutes: 60
+    timeout-minutes: 30
     permissions:
       contents: read
       id-token: write
@@ -286,8 +286,6 @@ jobs:
       - name: Build tests
         id: shard
         run: tuist test --build-only --shard-total 2 --shard-granularity suite TuistAcceptanceTests
-        env:
-          TUIST_SHARD_REFERENCE: github-${{ github.run_id }}-cli-acceptance-tests
 
   cli-acceptance-tests:
     name: Acceptance Tests
@@ -349,7 +347,6 @@ jobs:
         run: tuist test --without-building --platform macOS TuistAcceptanceTests -- -retry-tests-on-failure -test-iterations 2
         env:
           TUIST_SHARD_INDEX: ${{ matrix.shard }}
-          TUIST_SHARD_REFERENCE: github-${{ github.run_id }}-cli-acceptance-tests
 
   cli-acceptance-tests-fork:
     name: Acceptance Tests (Fork)
--- a/cli/Sources/TuistSupport/SwiftPackageManager/SwiftPackageManagerController.swift
+++ b/cli/Sources/TuistSupport/SwiftPackageManager/SwiftPackageManagerController.swift
@@ -3,9 +3,11 @@ import FileSystem
 import Foundation
 import Mockable
 import Path
+import SwifterPMCore
 import TSCUtility
 import TuistConstants
 import TuistEnvironment
+import TuistLogging
 
 /// Protocol that defines an interface to interact with the Swift Package Manager.
 @Mockable
@@ -64,32 +66,79 @@ public protocol SwiftPackageManagerControlling {
     ) async throws
 }
 
+protocol SwifterPMResolving: Sendable {
+    func resolveDependencies(_ request: SwifterPMResolutionRequest) async throws
+    func updateDependencies(_ request: SwifterPMResolutionRequest) async throws
+}
+
+extension SwifterPM: SwifterPMResolving {
+    func resolveDependencies(_ request: SwifterPMResolutionRequest) async throws {
+        _ = try await resolve(request)
+    }
+
+    func updateDependencies(_ request: SwifterPMResolutionRequest) async throws {
+        _ = try await update(request)
+    }
+}
+
+public enum SwiftPackageManagerControllerError: FatalError, Equatable {
+    case missingSwifterPMArgumentValue(String)
+    case unsupportedSwifterPMArguments([String])
+    case conflictingSCMToRegistryTransformationArguments
+
+    public var type: ErrorType {
+        .abort
+    }
+
+    public var description: String {
+        switch self {
+        case let .missingSwifterPMArgumentValue(argument):
+            return "The SwifterPM argument \(argument) requires a value."
+        case let .unsupportedSwifterPMArguments(arguments):
+            return "The following Swift Package Manager arguments are not supported by SwifterPM: \(arguments.joined(separator: ", "))"
+        case .conflictingSCMToRegistryTransformationArguments:
+            return "The source-control to registry transformation arguments conflict."
+        }
+    }
+}
+
 public struct SwiftPackageManagerController: SwiftPackageManagerControlling {
     private let fileSystem: FileSysteming
     private let commandRunner: () -> CommandRunning
+    private let swifterPM: SwifterPMResolving
     private let environmentVariables: @Sendable () -> [String: String]
-    private let currentExecutablePath: @Sendable () -> AbsolutePath?
 
     public init() {
         self.init(
             fileSystem: FileSystem(),
-            commandRunner: { CommandRunner() }
+            commandRunner: { CommandRunner() },
+            swifterPM: SwifterPM()
         )
     }
 
     init(
         fileSystem: FileSysteming,
         commandRunner: @escaping () -> CommandRunning,
-        environmentVariables: @escaping @Sendable () -> [String: String] = { Environment.current.variables },
-        currentExecutablePath: @escaping @Sendable () -> AbsolutePath? = { Environment.current.currentExecutablePath() }
+        swifterPM: SwifterPMResolving = SwifterPM(),
+        environmentVariables: @escaping @Sendable () -> [String: String] = { Environment.current.variables }
     ) {
         self.fileSystem = fileSystem
         self.commandRunner = commandRunner
+        self.swifterPM = swifterPM
         self.environmentVariables = environmentVariables
-        self.currentExecutablePath = currentExecutablePath
     }
 
     public func resolve(at path: AbsolutePath, arguments: [String], printOutput: Bool) async throws {
+        if swifterPMIsEnabled {
+            let request = try await swifterPMRequest(
+                packagePath: path,
+                arguments: arguments,
+                printOutput: printOutput
+            )
+            try await swifterPM.resolveDependencies(request)
+            return
+        }
+
         let command = try await buildResolveOrUpdateCommand(
             packagePath: path,
             extraArguments: arguments + ["resolve"]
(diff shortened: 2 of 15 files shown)


fortmarek (changes requested):
I found a few issues that should be addressed before we merge this.

[P1] Registry cache can serve the wrong package across registries.

SwifterPM keys registry sources and archives by identity and version only, then returns the cached source before checking the active registry or checksum. If two registries publish the same scoped identity/version, or a registry mapping changes, SwifterPM can reuse stale or wrong source code. The relevant paths are `Sources/swifterpm/Cache.swift` for `sourcePath(pin:)` and `registryArchivePath(identity:version:)`, `Sources/swifterpm/Restore.swift` for `ensureRegistrySource`, and `Sources/swifterpm/Registry.swift` for `downloadArchive`. The cache key should include the registry URL and preferably checksum, and cached archives should still be validated.

[P1] `--replace-scm-with-registry` is silently ignored on the SwifterPM path.

Tuist forwards passthrough SwiftPM args into SwifterPM, including existing configs that use `--replace-scm-with-registry`, but SwifterPM only parses/stores that flag and never applies a source-control-to-registry transformation. This makes opt-in installs resolve a different graph than `swift package` for users relying on that flag. The Tuist side forwards it through `SwiftPackageManagerController`, and the existing install tests assert this passthrough behavior. On the SwifterPM side, `Sources/swifterpm/CLI.swift` parses `replaceSCMWithRegistry`, `useRegistryIdentityForSCM`, and `disableSCMToRegistryTransformation`, but those values are not used by the resolver.

[P2] The Tuist mise integration asks for Linux, but SwifterPM only releases macOS assets.

This PR adds `github:tuist/swifterpm` globally in `mise.toml`, and release automation runs `mise lock tuist github:tuist/swifterpm --platform linux-x64,macos-arm64,macos-x64`. The checked-in `mise.lock` only has macOS SwifterPM entries, and SwifterPM’s own release workflow only builds `x86_64-apple-darwin` and `aarch64-apple-darwin`. This likely breaks Linux `mise install` or the release lock update unless SwifterPM also publishes Linux assets.

[P2] SwifterPM writes manifest cache files into user package directories.

`ManifestLoader.dumpPackageJSON` writes `.swifterpm-manifest.json` directly beside each `Package.swift`. Through this PR, `TUIST_USE_SWIFTERPM=1 tuist install` can dirty the root package and local package dependency checkouts with untracked files. The manifest cache should live under `.build/swifterpm` or the global SwifterPM cache instead.

[P2] Transitive local package dependencies are not represented.

The resolver’s main dependency parser only handles source-control and registry dependencies, while filesystem dependencies are handled separately and only appended from the root package when writing workspace state. A local package that itself depends on another local package will be omitted from resolution/workspace state.

[P3] SwifterPM’s user-visible version defaults are stale.

The CLI reports `0.1.0`, and the Bazel extension default is also `0.1.0`, while Tuist pins `0.5.1`. This will confuse debugging and release verification.

I validated this by fetching the PR locally, cloning `tuist/swifterpm`, and reviewing the Tuist install/release integration together with SwifterPM’s resolver, registry cache, restore, manifest cache, and release workflow paths.

fortmarek:
[P2] Consider linking SwifterPM as a library instead of bundling and shelling out to its CLI.

Because Tuist controls SwifterPM, packaging a second signed binary and invoking it through string arguments feels heavier than necessary for a supported integration. It also makes the contract harder to test and easier to drift: Tuist can pass flags that SwifterPM parses differently or ignores, errors are only process-level, and release packaging now has to keep Tuist and the bundled SwifterPM binary aligned across platforms. I’d consider exposing a `SwifterPMCore` library target, keeping the `swifterpm` executable as a thin wrapper around it, and having Tuist call the typed resolver/restorer APIs directly. Shelling out is fine for an opt-in spike, but I would not make that the long-term integration shape for a tool we own.

pepicrft:
@fortmarek I agree. I changed the integration to link `SwifterPMCore` directly and removed the `vendor/swifterpm` release packaging path. `tuist install` now stays on the existing `swift package` path unless `TUIST_USE_SWIFTERPM` is present, and the opt-in path builds a typed `SwifterPMResolutionRequest` instead of passing a CLI string contract through a subprocess.

While validating the generated Xcode project I hit a SwifterPMCore compile issue with `swift-subprocess` trait-gated `.data(limit:)` helpers in generated projects. I opened https://github.com/tuist/swifterpm/pull/13 to make the core target use the non-trait `.bytes(limit:)` API. Once that is released, I’ll move this PR to the new tag and finish the `tuist install` validation here.

pepicrft:
@fortmarek Follow-up on this: SwifterPM `0.6.1` is out and this PR now depends on `SwifterPMCore` from that release, so there is no bundled `swifterpm` binary and no subprocess boundary for the opt-in path.

I also validated the existing `.build` case in this repo. With an existing SwiftPM `.build`, `TUIST_USE_SWIFTERPM=1 tuist install` succeeded, SwifterPM reused `.build` as its scratch directory, and it replaced only resolved package checkout/download entries with symlinks into its cache. I then ran `tuist generate tuist ProjectDescription --no-open` and built the `tuist` scheme in Xcode successfully against those symlinks.

pepicrft:
@fortmarek I went through each requested item against the current Tuist PR and the released SwifterPM `0.6.1` tag:

- Registry cache key and archive validation: fixed by `tuist/swifterpm#12`, which is included in `0.6.1`. Registry source and archive paths now include registry URL plus checksum, and cached registry archives are SHA-256 validated before reuse.
- `--replace-scm-with-registry`: fixed by `tuist/swifterpm#12` and wired through this PR. Tuist maps `--replace-scm-with-registry` and `--use-registry-identity-for-scm` into `SwifterPMResolutionRequest.scmToRegistryTransformation`, and SwifterPM applies that transformation in `PackageResolver.transformedDependencies`.
- mise and release assets: this PR no longer adds or bundles a `swifterpm` binary. Tuist links `SwifterPMCore` directly from `Package.swift`, and `0.6.1` also publishes the Linux asset, so the old mise binary asset mismatch no longer applies.
- Manifest cache location: fixed by `tuist/swifterpm#12`, included in `0.6.1`. The package-info cache is written under `scratchDir/swifterpm/package-info`, and raw manifest dumps use `.build/swifterpm/manifests/package.json` instead of `.swifterpm-manifest.json` next to `Package.swift`.
- Transitive local package dependencies: fixed by `tuist/swifterpm#12`, included in `0.6.1`. `ManifestFileSystemDependencyGraph.collect` now walks filesystem dependencies transitively and is used for resolution, package-info cache writing, binary artifact restore contexts, and workspace state.
- Version metadata: the only remaining real gap was SwifterPM's user-visible metadata still saying `0.5.1` or `0.3.0`. I opened https://github.com/tuist/swifterpm/pull/14 to update the CLI version, Bazel module metadata, plist version, Bazel tool default, README example, e2e example, and the guard test to `0.6.1`.

Validation I ran:

- In SwifterPM: `swift test --filter MainTests`
- In Tuist: `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistSupportTests/SwiftPackageManagerControllerTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`

The Tuist test passed 11 controller tests, including the opt-in SwifterPM request path, SCM registry transformation forwarding, conflict handling, and update path. While building that selected test target, Xcode compiled through the generated workspace using the existing `.build/registry/downloads` symlink paths backed by `~/.cache/swifterpm`, so the existing `.build` scenario is still covered locally.

mise/tasks/app/bundle-ios.sh
```diff
@@ -6,6 +6,22 @@ set -eo pipefail
...
+    until op read "${args[@]}"; do
+        if [ "$attempt" -ge "$max_attempts" ]; then
+            return 1
+        fi
+
+        echo "op read attempt $attempt failed; retrying..." >&2
+        sleep $((attempt * 5))
+        attempt=$((attempt + 1))
+    done
+}
```
fortmarek: why was this added? Do we really need this?

cli/Sources/TuistSupport/SwiftPackageManager/SwiftPackageManagerController.swift
```diff
@@ -205,4 +272,222 @@ public struct SwiftPackageManagerController: SwiftPackageManagerControlling {
...
+        var unsupportedArguments: [String] = []
+
+        var index = arguments.startIndex
+        while index < arguments.endIndex {
+            let argument = arguments[index]
+            if let option = ValueOption.allCases.first(where: { argument == $0.rawValue }) {
+                let nextIndex = arguments.index(after: index)
+                guard nextIndex < arguments.endIndex else {
+                    throw SwiftPackageManagerControllerError.missingSwifterPMArgumentValue(argument)
+                }
```
fortmarek: I assume we also have the SwiftPM argument parser also inside `swifterpm`, wonder if a better interface would be to send an array of strings to the SwifterPM library instead of having the parser in two places

.github/workflows/cli.yml
```diff
@@ -197,7 +197,7 @@ jobs:
   cli-unit-tests:
     name: Unit Tests
     runs-on: tuist-macos
-    timeout-minutes: 30
+    timeout-minutes: 60
```
fortmarek: unit tests certainly shouldn't take more than 30 minutes, let's revert this

.github/workflows/cli.yml
```diff
@@ -286,6 +286,8 @@ jobs:
       - name: Build tests
         id: shard
         run: tuist test --build-only --shard-total 2 --shard-granularity suite TuistAcceptanceTests
+        env:
+          TUIST_SHARD_REFERENCE: github-${{ github.run_id }}-cli-acceptance-tests
```
fortmarek: why is this needed (and it wasn't needed before)?

fortmarek (approved):

mise/tasks/app/bundle-ios.sh
pepicrft: Good catch. This was a CI flake workaround for transient 1Password reads, unrelated to the SwifterPM change. I removed it in c016f9c, so the script uses the direct `op read` calls again.

cli/Sources/TuistSupport/SwiftPackageManager/SwiftPackageManagerController.swift
pepicrft: I agree that would be the better long-term API shape. I checked the current SwifterPMCore release and `CLIParser` plus the parsed `CLI` model are internal, so Tuist cannot call that parser directly without a SwifterPM API change and another release. For this PR I kept Tuist on the typed library boundary and only map the small set of install passthrough flags we support into `SwifterPMResolutionRequest`. The follow-up I would make in SwifterPM is to expose a public request builder/parser so the CLI and Tuist share the same argument interpretation.

.github/workflows/cli.yml
pepicrft: Reverted in c016f9c. The failing run ended up being a self-hosted runner communication loss, not a reason to increase the unit-test timeout.

.github/workflows/cli.yml
pepicrft: Removed in c016f9c. The current GitHub CI integration already derives the shard reference from `GITHUB_RUN_ID` and `GITHUB_RUN_ATTEMPT`, so the explicit `TUIST_SHARD_REFERENCE` wiring was unnecessary.

cli/Sources/TuistSupport/SwiftPackageManager/SwiftPackageManagerController.swift
pepicrft: Addressed now that SwifterPM `0.7.0` is released. This branch pins `SwifterPMCore` to `0.7.0` and `SwiftPackageManagerController` now calls `SwifterPMCommandParser.parse(_:)` with the same argument array shape as the SwifterPM CLI, then dispatches the returned typed `SwifterPMResolutionRequest` for `resolve` and `update`.

That removes Tuist's duplicate `SwifterPMOptions` parser. Tuist only adds the default `--package-path` when the passthrough args do not already include one, preserves `printOutput == false` by forcing `quiet`, and lets SwifterPM own unsupported argument and conflict validation.

Validation:

- `swift package --replace-scm-with-registry resolve`
- `swift package --replace-scm-with-registry describe --type json`
- `git diff --check`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistSupportTests/SwiftPackageManagerControllerTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`

PR #11103, opened 2026-06-05
ci(server): cache server image build to cut deploy wait time

fortmarek:
## What

Every deploy currently waits on the server Docker image build, which takes ~20 min even when it should hit caches. This makes two changes:

1. **Shared registry layer cache on the push-to-main build.** `server-production-deployment.yml`'s `build` job (the one that gates every merge deploy) now reads/writes the same `ghcr.io/tuist/tuist-buildcache:server` ref that `server-deployment.yml` already used. Because this job runs on every push to main, it keeps the cache warm merge to merge.
2. **Persist the OG-image render across builds via `buildkit-cache-dance`.** Both deploy build jobs now round-trip the `/og-cache/marketing` and `/og-cache/docs` BuildKit cache mounts through `actions/cache`, so unchanged OG images are skipped instead of re-rendered.

## Why / root cause

Profiling the 20-min build ([run](https://github.com/tuist/tuist/actions/runs/26967719794/job/79574666929)) showed the cache buys almost nothing today, for two independent reasons:

- **The cache producer was too infrequent.** `server-production-deployment.yml` (push-to-main) and `server.yml`'s `docker_build` (per-PR CI) had **no `cache-to` at all**. Only the rarely-run `server-deployment.yml` wrote `buildcache:server`. By the time it runs, deps have drifted, so the import finds a manifest but the layer keys miss. In the profiled run only 15 base layers hit; `mix deps.compile` (216s), swift build (44s), and the OG stages all re-ran.
- **Registry cache fundamentally can't cache the OG render.** The OG pipeline is the single biggest cost: `og-images` `mix compile` (264s) + Chromium render (376s) ≈ 11 min, and it sits on the critical path because `mix release` waits on `COPY --from=og-images` ([Dockerfile:197](server/Dockerfile:197)). The Dockerfile already tries to skip unchanged renders via a `--mount=type=cache` keyed on `.cachekey` sidecars ([Dockerfile:152](server/Dockerfile:152)), but cache mounts live only in the BuildKit builder, which is empty on every ephemeral runner. `cache-to ... mode=max` exports *layers*, never cache-mount contents, so the render runs in full every time.

## Why this approach

- **`buildkit-cache-dance`** is the standard way to persist `--mount=type=cache` mounts across CI runs; it pairs with `actions/cache` and is orthogonal to the registry layer cache (we keep both). It requires no Dockerfile change and leaves the existing sidecar-skip logic intact.
- **Producer = the push-to-main build** rather than the per-PR `docker_build` job: `docker_build` runs on fork PRs where `packages: write` and registry login aren't available (writing cache from untrusted PRs is also a poisoning risk). The push-to-main build runs on trusted code on every merge, which is frequent enough to stay warm.

## Validation

Validated on this PR with a temporary build-only workflow (isolated `buildcache:perf` + `perf-og-cache-*`, no deploy), now **removed**. Full step-level breakdown is in [this comment](https://github.com/tuist/tuist/pull/11103#issuecomment-4629594604).

| | Cold | Warm (representative code-change deploy) |
|---|---|---|
| **Total build job** | **21m 02s** | **8m 11s** (~61% faster) |
| OG Chromium render | 384s | **16.6s** (cache-dance skip) |
| `mix deps.compile` | 277s | **cached** |
| swift NIF build | 174s | 35s |
| `builder` `mix compile` | 272s | 265s (re-runs, code changed) |
| export cache to registry | 173s | 79s |

The residual ~8 min is the two parallel `mix compile` passes plus the cache export. The real deploy build additionally busts the `runner-base` apt layer per commit (~58s), so expect warm deploys closer to ~9 min.

## Out of scope (follow-ups)

- Date-based `APT_CACHE_BUST` to cache the ~58s `runner-base` apt layer across same-day deploys.
- Adding the shared cache to `server.yml`'s `docker_build` to speed up PR CI (needs base-image-ref alignment and fork-PR-safe handling).
- Eliminating the duplicate `mix compile` between the `builder` and `og-images` stages (the biggest remaining lever).


fortmarek:
## Validation results

Measured via the temporary `server-docker-build-perf.yml` on this PR (build only, isolated `buildcache:perf` + `perf-og-cache-*`, `ubuntu-latest`, same build args as the deploy build).

- **Cold** (run [27002728035](https://github.com/tuist/tuist/actions/runs/27002728035)) — first build, empty caches.
- **Warm** (run [27003762016](https://github.com/tuist/tuist/actions/runs/27003762016)) — second build after a one-line `server/lib` change, i.e. a representative code-change deploy.

| | Cold | Warm |
|---|---|---|
| **Total build job** | **21m 02s** | **8m 11s** |

Step-level (the parts that dominate):

| Step | Cold | Warm | Why |
|------|------|------|-----|
| OG Chromium render | 384s | **16.6s** | cache-dance restored `/og-cache`; sidecars matched, render skipped |
| `mix deps.compile` | 277s | **cached** | registry layer cache hit (mix.lock unchanged) |
| swift NIF build | 174s | 35s | layer largely cached (Package.resolved unchanged) |
| `builder` `mix compile` | 272s | 265s | re-runs — code changed (unavoidable) |
| `og-images` `mix compile` | 266s | 263s | re-runs in parallel with the above |
| export cache to registry | 173s | 79s | most layers already present |

**~61% faster on a normal code-change deploy.** The headline win is the OG render collapsing from ~6.5 min to ~17s (#1) plus deps/toolchain going fully cached (#2). The residual ~8 min is the two parallel `mix compile` passes (~265s) and the cache export — addressed by the out-of-scope follow-ups (drop the duplicate compile / persist `_build`).

Note: the real deploy build (`buildcache:server`) additionally busts the `runner-base` apt layer per commit (~58s), so expect warm deploys closer to ~9 min. The date-based `APT_CACHE_BUST` follow-up would recover that.

PR #11105, opened 2026-06-05
ci(server): publish multi-arch release image

pepicrft:
Resolves N/A

Publish the public Tuist server Docker release image as a multi-arch manifest with `linux/amd64` and `linux/arm64` variants so Apple Silicon Docker Desktop can run the self-host compose stack natively. This only affects the release image pushed as `ghcr.io/tuist/tuist:<version>` and `ghcr.io/tuist/tuist:latest`; deployment `sha-*` images stay on the existing single-architecture build path.

The release workflow now builds amd64 and arm64 images in separate native jobs and publishes the public tags from a final manifest job after both digests exist. That avoids QEMU on the existing server release build path and keeps the wall-clock release cost close to the slower native image build rather than a serialized multi-arch build. The self-host compose assets let Docker select the host's native Linux architecture automatically by removing the `linux/amd64` platform pin, and the docs call out the Apple Silicon behavior plus the `DOCKER_DEFAULT_PLATFORM=linux/amd64` escape hatch for amd64 emulation.

### How to test locally

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/release.yml"); puts "ok .github/workflows/release.yml"'`
- `docker compose --env-file .env.example -f docker-compose.yml config --no-env-resolution --quiet` from `server/priv/static/server/self-host`
- `git diff --check`

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -60,9 +60,7 @@ jobs:
         run: swift test --replace-scm-with-registry
   gettext:
     name: Gettext
-    # Gettext extraction is memory-sensitive enough to make the self-hosted
-    # Linux runner lose communication under fleet load.
-    runs-on: ubuntu-latest
+    runs-on: tuist-linux
     timeout-minutes: 15
     if: github.event.pull_request.draft == false || github.event_name != 'pull_request'
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/server.yml
```diff
@@ -260,7 +260,9 @@ jobs:
         run: mise run --no-deps format --check
   esbuild:
     name: esbuild
-    runs-on: tuist-linux
```
fortmarek, line 263: let's keep using our runners, even if they are currently somewhat unstable

fortmarek (approved):

esnunes (approved):

PR #11110, opened 2026-06-05
feat(infra): support Vultr regional deployments

pepicrft:
Resolves N/A

Adds Vultr as a supported provider for Kura regional deployments so managed Kura servers can be provisioned in regions such as Australia and Brazil while preserving Hetzner as the default path.

- Adds a Kura controller `--load-balancer-provider` flag with Hetzner and Vultr annotation sets for public HTTPS and gRPC LoadBalancer Services.
- Wires the provider through the Tuist Helm chart, adds a Vultr platform values overlay, and updates regional deployment scripts to pass provider-specific values.
- Catalogs `au-southeast` and `br-south` as Vultr-backed managed Kura regions with kubeconfig, storage class, and scheduling configuration, while keeping them disabled until the clusters are reconciled and synced.
- Installs CAPVULTR v0.4.0 from the management-cluster apply workflow, renders Vultr CAPI templates from the Git-pinned `infra/k8s/clusters/vultr/images.yaml` catalog, and applies the Sydney and Sao Paulo regional Cluster resources on push to `main`.
- Adds `infra/vultr-capi-image` plus a manual `vultr-capi-image.yml` workflow to build the Vultr CAPI node snapshot from upstream Kubernetes image-builder. 1Password stores only the Vultr API token; the selected snapshot ID is committed in Git.
- Adds bootstrap support for Vultr workload clusters, including Vultr CCM/CSI setup and provider-specific platform values, so reconciled Vultr machines can become usable Kura regional nodes.

### How to test locally

- `mise exec go@1.25.0 -- go test ./...` from `infra/kura-controller`
- `bash -n infra/mise/tasks/k8s/apply-vultr-regionals.sh infra/vultr-capi-image/build.sh infra/mise/tasks/k8s/bootstrap-workload.sh infra/mise/tasks/k8s/deploy-kura-regionals.sh infra/mise/tasks/k8s/install-platform.sh`
- `VULTR_CAPI_SNAPSHOT_ID=snapshot-test VULTR_CONTROL_PLANE_PLAN_ID=vc2-4c-8gb VULTR_WORKER_PLAN_ID=vc2-4c-8gb ruby -ryaml -e 'Dir["infra/k8s/clusters/vultr/*.yaml.tmpl"].each do |path|; rendered = File.read(path).gsub("${VULTR_CAPI_SNAPSHOT_ID}", ENV.fetch("VULTR_CAPI_SNAPSHOT_ID")).gsub("${VULTR_CONTROL_PLANE_PLAN_ID}", ENV.fetch("VULTR_CONTROL_PLANE_PLAN_ID")).gsub("${VULTR_WORKER_PLAN_ID}", ENV.fetch("VULTR_WORKER_PLAN_ID")); docs = YAML.load_stream(rendered); raise "#{path}: empty" if docs.empty?; puts "#{path}: #{docs.compact.size} docs"; end; YAML.load_file("infra/k8s/clusters/vultr/images.yaml"); YAML.load_file(".github/workflows/vultr-capi-image.yml")'`
- `mise exec -- helm template tuist infra/helm/tuist -f infra/helm/tuist/values-managed-kura-region.yaml --set kuraController.image.tag=test --set kuraController.loadBalancer.provider=vultr`
- `mise exec -- helm dependency build infra/helm/platform`
- `mise exec -- helm template platform infra/helm/platform -f infra/helm/platform/values-vultr.yaml -f infra/helm/platform/values-kura-region.yaml --set external-dns.txtOwnerId=test`
- `git diff --check`
- `git diff --cached --check`
- `mise exec -- mix test test/tuist/kura/regions_test.exs test/tuist/kura/provisioner/kubernetes_controller_test.exs` from `server` failed during local test database setup with `ERROR 42P01 (undefined_table) relation "users" does not exist`.

### Follow-up before merge

- Run `gh workflow run vultr-capi-image.yml --ref feat/vultr-kura-provider`, copy the produced Vultr snapshot ID, and commit it to `infra/k8s/clusters/vultr/images.yaml` as `snapshotID`.


fortmarek:
I think this PR is worth pausing on at the architecture level before we wire the workflow details.

Right now the shape is “one dedicated Kubernetes cluster per new Vultr Kura region”. That is internally consistent with the existing `tuist-kura-us-east` / `tuist-kura-us-west` regional-cluster model, but I’m not sure the requirement actually calls for regional control-plane isolation.

If what we need is “place Kura nodes in specific Vultr regions”, then a full workload cluster per region feels like the heavy abstraction: every new region gets its own control plane, kubeconfig, bootstrap lifecycle, platform controllers, ESO, ExternalDNS ownership, monitoring, upgrade path, and incident surface. That may be the right tradeoff if we explicitly want independent regional control planes, but if the goal is just data-plane placement, I’d expect a lighter model: regional node pools / provider-specific capacity with strict scheduling, storage class, and LoadBalancer configuration for Kura.

This is also where the Scaleway comparison is useful: Scaleway adds provider-backed capacity inside an existing environment cluster; it does not make production manage a new sibling cluster per capacity pool. Vultr is different if we use CAPVULTR from `tuist-mgmt`, but the product need sounds closer to “regional Kura capacity” than “regional Kubernetes control planes”.

Could we document why a dedicated Kubernetes cluster per Vultr region is necessary? If the answer is not control-plane isolation, I’d strongly consider modeling this as regional Kura node placement instead of making every future Kura region imply another full cluster.

PR #11112, opened 2026-06-05
feat(cli): support SwiftPM prebuilt libraries

pepicrft:
Resolves N/A

Adds support for SwiftPM `prebuilts` entries in `workspace-state.json`, which SwiftPM now uses for precompiled SwiftSyntax macro support libraries. Macro and plugin dependency paths can now consume SwiftPM's prebuilt `MacroSupport` library, while regular runtime dependencies on SwiftSyntax continue to use source or cached framework targets.

- Decode SwiftPM prebuilt metadata and forward it into external package mapping.
- Replace eligible macro/plugin product dependencies with the prebuilt include, library search, and linker settings SwiftPM expects.
- Keep non-host dependencies on source targets so runtime SwiftSyntax products are not accidentally substituted by macro-only prebuilts.
- Add focused mapper and graph-loader tests for prebuilt forwarding, macro substitution, and runtime dependency preservation.

### How to test locally

- `swiftformat cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphLoader.swift cli/Sources/TuistLoader/Models/SwiftPackageManagerWorkspaceState.swift cli/Sources/TuistLoader/Models/SwiftPackageManagerPrebuilt.swift cli/Tests/TuistLoaderTests/Loaders/SwiftPackageManagerGraphLoaderTests.swift cli/Tests/TuistLoaderTests/SwiftPackageManager/PackageInfoMapperTests.swift`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme TuistUnitTests -only-testing TuistLoaderTests/PackageInfoMapperTests/map_whenMacroTargetDependsOnPrebuiltProduct_usesPrebuiltSettings -only-testing TuistLoaderTests/PackageInfoMapperTests/map_whenRegularTargetDependsOnPrebuiltProduct_keepsSourceDependency -only-testing TuistLoaderTests/SwiftPackageManagerGraphLoaderTests/load_whenWorkspaceStateContainsPrebuilts_passesPackagePrebuiltsToMapper CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `TUIST_EE=1 .build/arm64-apple-macosx/debug/tuist generate tuist ProjectDescription --no-open`
- `TUIST_EE=1 xcodebuild build -workspace Tuist.xcworkspace -scheme tuist -configuration Debug SYMROOT=/tmp/tuist-ee-products CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- In `~/Downloads/precompiled/MacroApp`: `/tmp/tuist-ee-products/Debug/tuist install`, `/tmp/tuist-ee-products/Debug/tuist cache warm --cache-profile all-possible`, `/tmp/tuist-ee-products/Debug/tuist generate --cache-profile all-possible --no-open`, and `xcodebuild build -workspace MacroApp.xcworkspace -scheme App -destination 'generic/platform=iOS Simulator' CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- In `~/Downloads/precompiled/RuntimeSwiftSyntax`: `/tmp/tuist-ee-products/Debug/tuist install`, `/tmp/tuist-ee-products/Debug/tuist cache warm --cache-profile all-possible`, `/tmp/tuist-ee-products/Debug/tuist generate --cache-profile only-external --no-open`, and `xcodebuild build -workspace RuntimeSwiftSyntax.xcworkspace -scheme RuntimeSwiftSyntax -destination 'generic/platform=macOS' CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `git diff --check`


fortmarek:
I found two behavioral issues in the mapper path:

1. `PackageInfoMapper.swift`: `prebuiltEligibleTargets` marks plugin targets as prebuilt-eligible (`case .macro, .plugin`), but `map(target:)` still filters plugin targets out before dependency mapping (`case .regular, .system, .macro, .test` is the only pass-through set). As a result the advertised plugin path never gets far enough to apply the prebuilt settings, while plugin targets can still influence the host-only eligibility heuristic. I think we should either map plugin targets here or remove `.plugin` from this new prebuilt path and coverage claim.

2. `PackageInfoMapper.swift`: `prebuiltDependency` calls `ProjectDescription.PlatformCondition.from(condition)` directly, while the normal `mapDependency` path catches conversion failures and skips the dependency. A prebuilt product dependency with an unsupported or config-only condition would now fail generation instead of being ignored. I think this should mirror the existing catch-and-return-nil behavior.

One broader testing note: I think this should also be covered end-to-end with an acceptance test. We originally disabled the macro acceptance tests because building SwiftSyntax/macros was too heavy, but this feature should make that scenario fast because the macro support library is expected to come from the SwiftPM prebuilt. An acceptance fixture that runs `install`/`generate`/`build` for a macro target with SwiftPM prebuilts would protect the real `workspace-state.json` decoding + graph loading + generated Xcode build settings path, which the current focused unit tests do not fully exercise.

pepicrft:
Thanks Marek. I addressed these in d428823417:

- Plugin targets no longer seed the prebuilt eligibility closure, since they are still filtered out before graph mapping.
- Prebuilt product dependencies now mirror the regular dependency path and skip unsupported or config-only conditions instead of failing generation.
- Added focused mapper coverage for both edge cases.
- Added an end-to-end acceptance fixture that runs install, generate, and build for a Project.swift app depending on a Package.swift macro package backed by SwiftSyntax prebuilts.

I also verified the focused unit tests and the new acceptance test locally. The current macOS CI failures do not show a test assertion or compile error. Both jobs were terminated mid-Xcode build while the step was still in progress, so I reran the failed jobs to validate with warmed caches.

cli/Tests/TuistAutomationAcceptanceTests/BuildAcceptanceTests.swift
```diff
@@ -391,6 +391,35 @@ struct BuildAcceptanceTestFrameworkWithSwiftMacroIntegratedWithStandardMethod {
     }
 }
 
+struct BuildAcceptanceTestSwiftPMPrebuiltMacro {
+    @Test(
+        .withFixture("generated_app_with_swiftpm_prebuilt_macro_dependency"),
+        .inTemporaryDirectory,
+        .timeLimit(.minutes(5))
```
fortmarek: I'd say this is too long if this is supposed to use a prebuilt?

cli/Tests/TuistAutomationAcceptanceTests/BuildAcceptanceTests.swift
```diff
@@ -391,6 +391,35 @@ struct BuildAcceptanceTestFrameworkWithSwiftMacroIntegratedWithStandardMethod {
...
+                "--platform",
+                "macos",
+                "--path",
+                fixtureDirectory.pathString,
+                "--derived-data-path",
+                derivedDataPath.pathString,
+                "--",
+                "-skipMacroValidation",
+            ]
+        )
```
fortmarek: might be worth checking also the generated project integrates the prebuilts as an xcframework (or in some other way ensure that the macro framework is not built in this scenario)

fortmarek (approved):

cli/Tests/TuistAutomationAcceptanceTests/BuildAcceptanceTests.swift
pepicrft: Fixed in 47dc6f220b. The acceptance test in `cli/Tests/TuistAutomationAcceptanceTests/BuildAcceptanceTests.swift` now only runs `install` and `generate`, and the time limit is down to 2 minutes. That keeps the test focused on validating the prebuilt integration path instead of paying for a full SwiftSyntax macro build.

cli/Tests/TuistAutomationAcceptanceTests/BuildAcceptanceTests.swift
pepicrft: Fixed in 47dc6f220b. The test now opens the generated `MacroDependency.xcodeproj`, asserts that the SwiftSyntax source targets are not generated, and checks the `MacroDependencyMacros` Debug settings for the prebuilt `MacroSupport` include path, library search path, and linker flag. That verifies the generated project is wired to the SwiftPM prebuilt instead of building the macro support framework from source.

PR #11116, opened 2026-06-05
fix(cli): HAR recording shutdown crash

pepicrft:
Resolves N/A

This fixes a crash pattern reported from `.ips` files where the CLI could continue HAR JSON encoding and persistence on detached background tasks while command execution was already completing.

The crash happened because HAR recording was detached and every entry persisted by JSON-encoding and writing `network.har` from the background task. On the affected GitLab runner, command completion could race those detached tasks, leaving `HARRecorder.persist()` active while `swift_errorInMain` was unwinding. This can look customer-specific because it depends on runner timing, macOS scheduling, network activity, and HAR being enabled, so most runs either finish the detached work earlier or exit before the race becomes visible.

The change keeps recording detached so the middleware does not block on HAR work. `HARRecorder` now owns both detached scheduling and scoped finishing: detached work appends entries to the recorder actor, and the active recorder is finished once at the session boundary to persist entries recorded so far. The recorder closes through a lock-protected state before awaiting actor isolation, so a high-concurrency burst cannot keep shutdown stuck behind a large queue of pending HAR messages. Commands that disable HAR recording temporarily shadow the task-local recorder with `nil`; commands that keep HAR enabled reuse the existing session recorder instead of creating a nested writer. Any detached recording that arrives after `finish()` is ignored, which can lose a late HAR entry but prevents JSON encoding and file writes from continuing during shutdown. File upload and download recording now use the same detached recorder API instead of depending on task-local state inside detached tasks.

### How to test locally

- `mise run cli:lint --fix`
- `tuist generate tuist TuistHAR TuistHARTests ProjectDescription --no-open`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistHARTests/HARRecordingMiddlewareTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistHARTests/HARRecorderTests -only-testing TuistHARTests/HARRecordingMiddlewareTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`

### DIFF
--- a/Tuist/ProjectDescriptionHelpers/Module.swift
+++ b/Tuist/ProjectDescriptionHelpers/Module.swift
@@ -1630,6 +1630,8 @@ public enum Module: String, CaseIterable {
                     .target(name: Module.testing.targetName),
                     .external(name: "FileSystem"),
                     .external(name: "FileSystemTesting"),
+                    .external(name: "HTTPTypes"),
+                    .external(name: "OpenAPIRuntime"),
                 ]
             case .tuistFixtureGenerator:
                 [
--- a/cli/Sources/TuistHAR/HARRecorder.swift
+++ b/cli/Sources/TuistHAR/HARRecorder.swift
@@ -1,5 +1,6 @@
 import Foundation
 import Path
+import Synchronization
 import TuistConstants
 import TuistLogging
 
@@ -22,27 +23,20 @@ public actor HARRecorder { // swiftlint:disable:this type_body_length
         "x-amz-credential",
     ]
 
-    private final class State: @unchecked Sendable {
-        private let lock = NSLock()
-        private var finished = false
+    private nonisolated let finishedState = Mutex(false)
 
-        var isFinished: Bool {
-            lock.lock()
-            defer { lock.unlock() }
-            return finished
-        }
+    private nonisolated var isFinished: Bool {
+        finishedState.withLock { $0 }
+    }
 
-        func markFinished() -> Bool {
-            lock.lock()
-            defer { lock.unlock() }
-
-            guard !finished else { return false }
-            finished = true
+    private nonisolated func markFinished() -> Bool {
+        finishedState.withLock { isFinished in
+            guard !isFinished else { return false }
+            isFinished = true
             return true
         }
     }
 
-    private nonisolated let state = State()
     private var log: HAR.Log
     private let filePath: AbsolutePath?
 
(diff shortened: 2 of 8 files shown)


fortmarek (changes requested):
Thanks for digging into this crash pattern. I think there are two things worth tightening before this lands:

- [P1] HAR is not persisted on command failures. `Dependencies.initDependencies` now relies on `HARRecorder.withCurrent(...).finish()` for the final write, but `TuistCommand.onError` still terminates with `_exit(exitCode)`. Because `_exit` does not unwind async scopes, the cleanup in `HARRecorder.withCurrent` never runs on the normal error path. Since `record` no longer persists each entry, failing commands can leave `network.har` missing or incomplete, which is especially painful for the crash/debugging case this PR targets. I’d finish the active recorder before calling `_exit`, or move the owning HAR scope inside `TuistCommand.main` so it completes before error presentation exits, and add coverage for a failing command path.

- I’d avoid `@unchecked Sendable` for `HARRecorder.State` in production code. The lock probably protects the boolean, but `@unchecked Sendable` bypasses the compiler around the exact shutdown race this PR is trying to make safer. Since `HARRecorder` is already an actor, I’d prefer actor-isolated state unless the “close before awaiting the actor queue” behavior is essential. If early nonisolated closing really is required, a real concurrency primitive such as `Synchronization.Mutex<Bool>` or an atomic would be a better fit than a custom unchecked reference type.

pepicrft:
Addressed both requested changes in `a6ab08c4b4`:

- `TuistCommand.onError` now awaits `HARRecorder.finishCurrent()` before every terminal exit path in the handler, including the `_exit(exitCode)` path. That flushes the active task-local recorder before `_exit` prevents async scope cleanup from running.
- `HARRecorder` now uses `Synchronization.Mutex<Bool>` for the nonisolated finished flag instead of the custom `@unchecked Sendable` state holder. The early nonisolated close remains, so detached recording tasks can be rejected quickly once shutdown starts.
- Added `HARRecorderTests.finishCurrent_persistsActiveRecorderBeforeTaskLocalScopeExits()` to cover explicit flushing of the active task-local recorder before the surrounding scope exits.

Validation:

- `tuist generate tuist TuistHAR TuistHARTests ProjectDescription --no-open`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistHARTests/HARRecorderTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`

fortmarek (approved):

PR #11123, opened 2026-06-05
feat(infra): tail actions/runner _diag logs to stdout so disconnects reach Loki

fortmarek:
## What

Add a small backgrounded tailer (`diag-tail.sh`) that copies the actions/runner's `_diag/{Runner,Worker}_*.log` files to stdout, line-prefixed with `[diag] `. Wired into `run-job.sh` next to `vitals.sh`; baked into the linux runner image.

## Why this and not the env var that was already there

`ACTIONS_RUNNER_PRINT_LOG_TO_STDOUT=1` was added in #11114 on community guidance ([orgs/community#69821](https://github.com/orgs/community/discussions/69821)) under the assumption that it mirrors the runner's `_diag` output to stdout. **Empirical check refutes that:** active-job runner pods land in Loki with **zero stdout during the job** — not even our own `RUNNER_VITALS` lines from `vitals.sh`. (Warm-standby pods in the same window stream `RUNNER_VITALS` every 3s and a freshly-created sibling pod was visible in Loki, so the shipping pipeline is healthy; the gap is the active-job pod's stdout itself.) So the env var does not surface `_diag` in our shape, and without `_diag` in Loki an exit-0 disconnect — runner agent exits cleanly while GitHub stops seeing its heartbeat — leaves no trace of why it stopped responding.

That's not hypothetical: it's the class of "lost communication" we just hit in production. The dead runner pod for that job terminated `exit=0 Completed` (recorded by the controller's `#11109` logging), yet GitHub flagged it as a runner disconnect. The smoking gun is in the runner's `_diag` files, which die with the pod.

Reading the on-disk diag files and echoing to stdout is the only mechanism that reliably reaches alloy-logs (which ships the pod's container stdout). And it gives us coverage for **both** disconnect modes we've now distinguished:

| Mode | Symptom | Where this helps |
|---|---|---|
| A: clean exit (today) | `exit=0 Completed` but GitHub says lost | Live `_diag` tail surfaces the agent's last network/HTTP/retry errors before the clean exit |
| B: silent microVM teardown (earlier exit-9 deaths) | `exit=9 Error` with no message | Lines flushed before death survive in `/var/log/pods/<pod>/runner/0.log` on the host even when the kata VM is gone; alloy-logs ships from there |

## How

- New `infra/linux-runner-image/diag-tail.sh`. Polls `$RUNNER_HOME/_diag/` every 2s, spawns one `tail -F -n +1 | sed -u "s|^|[diag] |"` per new `Runner_*.log` / `Worker_*.log` it sees. `sed -u` is line-buffered so lines reach stdout immediately instead of waiting for a 4 KiB block to fill — critical when the runner dies seconds after emitting its last error.
- `run-job.sh`: backgrounded next to `vitals.sh`, guarded so an older image without the script still boots the runner. Loki filtering for the agent's own output is one matcher: `|= "[diag]"`.
- Dockerfile: COPY + chmod, no other changes.

I'm intentionally **not** removing `ACTIONS_RUNNER_PRINT_LOG_TO_STDOUT=1` from `podtemplate.go` — it's free and may cover paths we haven't characterised — but the comment justifying it overstates its effect. A small follow-up cleanup of the comment makes sense once this lands.

## Validation

Local integration test: created a temp `_diag` dir, started the tailer, dropped a line into a `Worker_*.log`, later dropped a line into a `Runner_*.log`. Both surfaced on the tailer's stdout with the expected `[diag] ` prefix. `bash -n` clean on both scripts.

## Followups (not in this PR)

- The "active-job pods have no stdout at all" observation is its own anomaly — even our `RUNNER_VITALS` from `vitals.sh` vanishes mid-job. Worth a separate investigation into whether alloy-logs is actually shipping the active-job pod's container log file on the runner-tier nodes, or whether something in the kata-qemu → host log-forwarding path drops it for that pod state. This PR works around it by writing diag explicitly, but the underlying gap may also be hiding other signal.
- This PR doesn't yet open a Mac runner equivalent.

### DIFF
--- a/infra/linux-runner-image/Dockerfile
+++ b/infra/linux-runner-image/Dockerfile
@@ -200,7 +200,8 @@ USER root
 COPY dispatch-poll.sh /usr/local/bin/dispatch-poll.sh
 COPY run-job.sh /usr/local/bin/run-job.sh
 COPY vitals.sh /usr/local/bin/vitals.sh
-RUN chmod +x /usr/local/bin/dispatch-poll.sh /usr/local/bin/run-job.sh /usr/local/bin/vitals.sh
+COPY diag-tail.sh /usr/local/bin/diag-tail.sh
+RUN chmod +x /usr/local/bin/dispatch-poll.sh /usr/local/bin/run-job.sh /usr/local/bin/vitals.sh /usr/local/bin/diag-tail.sh
 
 USER runner
 
--- a/infra/linux-runner-image/diag-tail.sh
+++ b/infra/linux-runner-image/diag-tail.sh
@@ -1,0 +1,67 @@
+#!/usr/bin/env bash
+# Stream actions/runner diagnostic logs to stdout so they reach Loki via
+# the pod-log pipeline.
+#
+# The runner writes per-process diagnostic output (Listener handshake,
+# session lifecycle, Worker job execution, network/retry errors, HTTP
+# 5xx responses from the Actions service) to:
+#
+#   $RUNNER_HOME/_diag/Runner_*.log
+#   $RUNNER_HOME/_diag/Worker_*.log
+#
+# Those files live inside the container filesystem and die with the
+# Pod, so whenever a runner exits abnormally (silent microVM teardown,
+# or — as seen on the post-#11114 "lost communication" run — a clean
+# exit 0 while GitHub stops seeing the heartbeat) nothing survives in
+# Loki to explain it.
+#
+# `ACTIONS_RUNNER_PRINT_LOG_TO_STDOUT=1` is advertised in community
+# threads as the way to mirror this output to stdout (originally added
+# for actions-runner-controller), but verified empirically here that
+# active-job pods land in Loki with no stdout at all — the env var
+# does not surface _diag in our shape. Reading the files and copying
+# lines to stdout is the only mechanism that reliably reaches alloy.
+#
+# Each emitted line is prefixed with "[diag] " so Loki filtering is a
+# single `|= "[diag]"`. The tailer is backgrounded by run-job.sh before
+# `exec ./run.sh`, so it lives for the lifetime of the Pod. On a host-
+# side microVM teardown the tailer dies with the VM, but every line it
+# already flushed to stdout is preserved on the host's
+# /var/log/pods/<pod>/runner/0.log file that alloy-logs ships.
+
+set -u
+
+DIAG_DIR=${1:-/home/runner/actions-runner/_diag}
+
+# Create the directory pre-emptively so `tail -F` can start watching
+# before the runner writes its first log file. Failure is non-fatal:
+# the directory will exist by the time the runner starts, and a stale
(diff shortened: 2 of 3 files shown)


fortmarek:
Closing — the failure class this PR was built for turned out not to be the actual problem, and the PR's own unresolved caveat would have prevented it from delivering value anyway.

**What we found instead**

The runner-disconnect investigation that motivated this PR resolved in #11133. The actor was **`OrphanedStampedPodsWorker`** force-deleting runner pods from outside while their `runner` containers were healthy and executing customer jobs (Mode C — pod just disappears, no exit signal). The runner agent had nothing meaningful to write to `_diag` because it wasn't failing; it was being killed from the API.

**Why this PR wouldn't have helped find that**

A force-delete by the worker produces no runner-side diagnostics. The runner doesn't know it's about to die — `_diag` would just stop mid-line. The diagnostic signal that named the actor was the controller's #11109 termination log + the server's `runners: reaped orphaned stamped pod` line in Loki, both already in place.

**Unresolved blocker on this PR itself**

The PR description called out a structural gap: **active-job runner pods don't ship stdout to Loki at all**. Even our existing `RUNNER_VITALS` lines from `vitals.sh` disappear during the active-job window. Whatever is dropping the stdout for those pods would also drop the `[diag]` lines this PR emits — meaning the instrument quite likely wouldn't have reached Loki for the pods we cared about. We never closed that gap, and shipping more stdout into a known-dropped stream isn't useful.

**Cost we'd be paying for the speculation**

Every Linux runner pod gets a `while :; do sleep 2; done` polling loop that fans out `tail -F` subprocesses as `_diag` files appear. Tiny per pod, but multiplied across the warm pool it's ongoing churn for no confirmed benefit.

**What's kept (already merged, doing the real work)**

- #11109 — controller logs `containerStatuses[runner].state.terminated.{exitCode,reason,signal}` before reaping. Names how the runner ended, from the controller side.
- #11114 — `feature-node-logs` on alloy-logs DaemonSet, ships host journald (containerd / kubelet / kernel) to Loki. Covers Mode B (host-side teardown) if it ever materialises.
- #11133 — `OrphanedStampedPodsWorker` refuses to delete pods whose `runner` container has begun executing (Linux container guard) OR whose `RunnerSession` is open (shape-independent guard). Closes the silent-delete class that was producing the bulk of "lost communication" symptoms.

**If a future "lost communication" turns out to need this**

The implementation is small and tested (the integration test in the PR validates the file-tailer end-to-end). Reviving the branch is cheap. Leaving the branch in place rather than deleting so it stays one click away from being reopened.

**Worth a separate investigation regardless**

The active-job-pod-stdout-missing-from-Loki anomaly is real and not fixed by closing this. If alloy-logs is genuinely dropping the active-job pod's container stdout (which would explain why `vitals.sh` lines vanish mid-job), that's a structural log-shipping defect that hides more than `_diag` — every backgrounded instrument we run in the runner container is affected. Worth opening as its own ticket once we have time to chase it.

PR #11125, opened 2026-06-05
feat(infra): route Kura public traffic to one primary pod per region

fortmarek:
## What changed

Each managed Kura endpoint pins its public HTTPS and gRPC Services to a **single primary pod** (via the StatefulSet per-pod label `statefulset.kubernetes.io/pod-name`) instead of selecting the whole 3-replica StatefulSet. Selection is sticky with failover, driven by a new pod watch.

## Why

Each region's Kura endpoint is a 3-replica StatefulSet behind one Hetzner LoadBalancer, and the mesh replicates writes between pods **asynchronously**. Today the public LB fans every request across all three pods with no affinity, so a read can land on a pod that hasn't yet received a just-written blob. That makes read-after-write **inconsistent in steady state, not just during deploys**.

This is the root cause behind the flaky canary compilation-cache acceptance test (`generated_project_with_caching_enabled`), where a warm rebuild lands at a variable 84–91% hit rate instead of 100%: the App target's objects, uploaded last at the end of build 1, haven't propagated to all 3 pods by the time build 2's reads scatter across them. (The client-side test was made tolerant in [#11119](https://github.com/tuist/tuist/pull/11119); this PR fixes the cause.)

## The decision

Route all of a region's reads + writes to one primary pod → read-after-write is consistent in steady state, with no change to the write path (no read-repair, no synchronous replication — **pure async preserved**). Deploys stay **zero-downtime** because the primary role hands off to a warm, already-replicated standby. The deliberate tradeoff: brief inconsistency is allowed **only during deploy/failover windows** instead of all the time, in exchange for giving up per-region read scaling.

### Selection + failover
- The current primary is read back from the existing Service selector, so the choice is **sticky across reconciles and survives a controller restart** without a dedicated status field (the structural CRD would prune one, and a status field would couple to CRD/image deploy ordering).
- Keep the current primary while it's `Ready`; fail over to the **lowest-ordinal `Ready` pod** when it goes unready or is rolled; **never fail back** (each handoff costs a brief inconsistency window while async replication catches up on the new primary).
- A pod watch (filtered to Ready/terminating transitions) re-runs selection on readiness changes.
- The **headless Service keeps the broad selector** so peer replication still reaches every pod; the 3 volumes + async mesh are unchanged.

### Why not the obvious alternatives
- **`Service.sessionAffinity: ClientIP`** is a no-op here. The pod fan-out is at the Hetzner **tcp-passthrough LB** (client→node, `least_connections`), not at kube-proxy (node→pod is 1:1 under `externalTrafficPolicy: Local` with one pod per node), and the LB is unaware of the Service field.
- **Single-volume transfer** gives consistency but can't be zero-downtime — a Hetzner block volume attaches to one node, so the handoff has a detach/attach gap. "One writer per volume" still holds here; we transfer the **leader role**, not the disk.

## Tests / validation

- `choosePrimaryPod` unit tests: default-before-ready, lowest-ordinal pick, sticky-while-ready, failover-on-unready, keep-current-when-none-ready, numeric (not lexical) ordinal ordering.
- Reconcile integration test (fake client): public + gRPC Service selectors pin to `-0`; failover to `-1` when `-0` goes unready; no fail-back when `-0` recovers.
- `go build`/`go vet`/`go test ./...` all green.

> The actual LB routing + failover convergence against the Hetzner tcp-passthrough LB can only be confirmed on a live cluster — this is the head of the deploy cascade (canary), so it gets exercised there. The tests cover the selection/failover logic.


esnunes (approved):
LGTM

pepicrft (approved):

PR #11165, opened 2026-06-08
ci(kura): Bazel foundation + flip the compile gate to Bazel

esnunes:
## What

A **self-contained, minimal** landing of Bazel for Kura on `main`: just enough to build the crate and flip the `compile` CI gate from Cargo to Bazel. Unlike the full migration branch (`kura-bazel-phase1`), this carries **only the files the compile gate needs** — `MODULE.bazel` is trimmed to the foundation, so there are **no OCI deps or files** (the runtime image lands in a later PR).

When migrated to use a remote Kura node, the following files will be removed or adjusted:
- `ci-kura-cache.sh` will be completely removed
- `kura.yaml` workflow won't need to use disk cache, nor run a local kura instance

This PR is the 1 out of X.

## Files (13)

**Foundation**
- `kura/MODULE.bazel` — rules_rust + crate_universe + cross cc/platforms toolchains, and the mlua/`lua-src` vendored-C workaround. Trimmed: no rules_oci/pkg/distroless, no apt/geoip extensions.
- `kura/MODULE.bazel.lock` (regenerated for the trimmed module), `.bazelrc`, `.bazelversion`
- `kura/BUILD.bazel` (`rust_library`/`rust_binary`/`rust_test`), `kura/Cargo.Bazel.lock` (crate_universe pin)
- `kura/bazel/{platforms,toolchains/cc,third_party/lua}/BUILD.bazel`
- `kura/bazel/linux-dev.Dockerfile` (Debian Bookworm + cross GCC), `kura/bazel/ci-kura-cache.sh`
- `kura/.gitignore` — ignore `/bazel-*` convenience symlinks

**CI**
- `.github/workflows/kura.yml` — the `compile` job

## The `compile` job

- Runs on the host runner; builds the dev image in-step; runs **Kura as a normal `docker run` container** (`ci-kura-cache.sh`) so it can be **gracefully stopped** (SIGTERM → RocksDB flush) as the **last step before** the `actions/cache` save. The `actions/cache/restore`+`save` split is what lets the save happen *after* the flush (a `services:`/`container:` Kura can't guarantee that → corrupt snapshots).
- `bazel build` runs **inside the dev image** because the cc toolchain is **non-hermetic** (`compiler_files = ":empty"` → container gcc-12/glibc-2.36). Host Bazel would mis-compile against the wrong glibc and, since the compiler isn't a declared action input, **poison the cache** under matching keys.
- Builds **lib + binary for both Linux arches** (arm64 cross-compile breakage now caught in CI, not only at release) with **warnings-as-errors**; the **test target compiles on the runner's native (x86_64) arch only** — its test toolchain doesn't resolve when cross-built.
- Command is **inlined** (no new wrapper script); the OCI image is **not** built here.

## Coverage / cache

- Upgrade over the old `cargo check` (x86_64-only): both arches compiled + `-D warnings`.
- Caching is the **per-job snapshot** model (dedicated `kura-compile-data-*`, keyed on the lockfiles): warms future *compile* runs. Cross-job warming needs a shared persistent Kura (separate follow-up).

## Validation (CI, on this PR)

**Cold run — entire `kura.yml` green:** the Bazel `Compile` job *and* all unchanged cargo jobs (format, clippy, audit, tests, 4 e2e shards). Compile cold ≈ 45 min (empty Kura): dev image built, Kura started, both arches + native test compiled under `-D warnings`, Kura stopped cleanly (`exit 0`, `persist/fd error lines: 0`), data dir + repo cache saved.

**Warm re-run — ~1m51s**, snapshot round-trips:
- `Cache hit for: kura-compile-data-…` → restored.
- x86_64 lib+bin: `801 remote cache hit, 0 executed` (17 s, was ~22.7 min)
- arm64 lib+bin: `835 action + 303 remote cache hit, 0 executed` (7 s)
- native test: `915 action + 317 remote cache hit, 0 executed` (3 s)
- Kura `exit 0`, 0 errors; cache save skipped on hit.

This proves the design end-to-end: Kura is dogfooded as the remote cache, the graceful-stop snapshot captures a complete flushed RocksDB, and it restores as a warm cache (801 remote hits on the formerly-cold build, zero re-execution).

`actionlint` clean; no `bazel-*` symlinks leak into the commit.

## Before merging

- Whenever `Cargo.toml`/`Cargo.lock` change (incl. rebases), `Cargo.Bazel.lock` must be repinned in the same commit (`CARGO_BAZEL_REPIN=true bazel build //:kura --nobuild`) or Bazel fails at analysis with "lockfile is out of date for 'crates'".

## Follow-ups

- Runtime OCI image (rules_oci) as its own PR.
- Shared persistent Kura (warm + cross-job warming), then flip `tests` → `e2e` → release.


pepicrft (approved):
Validated the Kura Bazel compile migration against the live main target. The Bazel targets load, both Linux platform builds compile in the provided dev image, the test target compiles, the cache helper starts and stops cleanly, and the workflow passes actionlint.

.github/workflows/kura.yml
```diff
@@ -56,25 +56,93 @@ jobs:
   compile:
     name: Compile
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
```
fortmarek, line 59: is the kura compilation taking actually over 30 minutes? I would have expected it to be faster.

.github/workflows/kura.yml
```diff
@@ -56,25 +56,93 @@ jobs:
...
-      - name: Compile without warnings
-        run: env -u RUSTUP_TOOLCHAIN RUSTFLAGS="-D warnings" cargo check --locked --all-targets
+      - name: Start Kura as the Bazel remote cache
+        working-directory: kura
+        env:
+          KURA_CI_DATA_DIR: ${{ runner.temp }}/kura-data
+        run: bash bazel/ci-kura-cache.sh start
+
+      - name: Compile (Bazel, both Linux arches, warnings-as-errors)
+        run: |
```
fortmarek, line 104: I'd extract this into a mise script, so it can be easily run locally

.github/workflows/kura.yml
```diff
@@ -56,25 +56,93 @@ jobs:
   compile:
     name: Compile
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     if: github.event.pull_request.draft == false || github.event_name != 'pull_request'
+    # The compile gate is a Bazel build that dogfoods Kura as the remote cache. Kura runs as
+    # a normal `docker run` container (bazel/ci-kura-cache.sh) so it can be GRACEFULLY stopped
+    # (SIGTERM -> RocksDB flush) as the last step before its data dir is snapshotted by
```
fortmarek, line 63: I wonder if caching Kura with the `main` build of Kura makes this unnecessarily complicated.

I'd have probably aimed at:
- using Kura-based caching based on the latest release (however we end up releasing the Bazel integration)
- and then having a more scoped-down fixture where we do an actual e2e acceptance test by building kura off of `main` as part of that workflow

fortmarek (commented):
Looks good. I do wonder if we can simplify the workflow a bit by using a released version of Kura (or the Bazel integration) instead of trying to build Kura with the locally-built version of Kura. wdyt?

esnunes:
> Looks good. I do wonder if we can simplify the workflow a bit by using a released version of Kura (or the Bazel integration) instead of trying to build Kura with the locally-built version of Kura. wdyt?

The idea of the `compile` job is to ensure the code compiles without errors. The image OCI image uses pre-built binaries from cache to avoid compiling them again

.github/workflows/kura.yml
esnunes: With a warm cache the compilation is very fast (2 minutes or less), however, when cache is cold or a dependency changed and bazel has to recompile the C code (lua + librocksdb) then it may take up to 60 minutes in both platforms (arm and x86)

.github/workflows/kura.yml
esnunes: This is a temporary code that will be replaced with a simple bazel argument pointing to a remote kura node once authentication is implemented.

The idea of using a local Kura is to test/validate Kura with a real use-case. I'm fine to postpone the workflow changes till the authentication is implemented, it will remove the need of `ci-kura-cache.sh` and the steps related to running a local kura, WDYT?

esnunes:

PR #11166, opened 2026-06-09
ci(kura): flip the tests gate to Bazel

esnunes:
## What

Flips the production `tests` gate in `.github/workflows/kura.yml` from `cargo test --locked` to `bazel test //:kura_lib_test`, dogfooding Kura as the Bazel remote cache. This is **Stage A, part 2** of the Kura Bazel migration, stacked on #11165 (the compile-gate flip).

> **Stacked PR** — base is `kura-bazel-compile` (#11165), not `main`. Review/merge #11165 first; the diff here is just the `tests` job.

## Why this is a faithful replacement for `cargo test`

The migration plan flagged one blocker before flipping tests: confirm `bazel test` covers `cargo test`'s scope, since rules_rust skips doctests. Checked against the crate:

- **No doctests** — there are no `/// ```` doc-example fences anywhere in `src/`, so the rules_rust doctest gap doesn't apply.
- **No integration tests** — there is no `tests/` directory.
- **228 inline `#[test]`/`#[tokio::test]`** unit tests, all compiled into `kura_lib` and run by the `//:kura_lib_test` rust_test target (which recompiles the crate with `--test`).

So `//:kura_lib_test` is the complete test surface here.

## Cache design — reuse compile's warm graph

The test command deliberately uses the **same flags as the compile job's x86_64 leg** (`-c opt`, `--platforms=//bazel/platforms:linux_x86_64`, `-Dwarnings`). `extra_rustc_flags` is a global build setting, so matching it makes every third-party dependency's Bazel action key **identical** between the `compile` and `tests` jobs. A fresh `tests` run therefore reuses compile's warm dependency cache (the expensive `rocksdb`/`jemalloc`/`ring`/`mlua` graph) via the `restore-keys` fallback to `kura-compile-data-`; the only new work is recompiling `kura_lib` as a test crate and running the tests.

Tests run on the runner's **native (x86_64) arch only** — the Bazel test toolchain doesn't resolve when cross-built.

## Infra reuse

Identical Kura-as-remote-cache wiring to the compile job: Kura runs as a normal `docker run` container via `bazel/ci-kura-cache.sh` so it can be **gracefully stopped** (SIGTERM → RocksDB flush) before its data dir is snapshotted by `actions/cache` (restore/save split). Own cache namespace (`kura-tests-data-`) with a fallback to `kura-compile-data-` to warm on the first run.

## Validation

- `actionlint .github/workflows/kura.yml` passes.
- CI on this PR exercises the new `tests` job; first run warms deps from compile's cache, the second confirms the warm path (target parity with the compile job's already-validated cold→warm behavior).

### DIFF
--- a/.github/workflows/kura.yml
+++ b/.github/workflows/kura.yml
@@ -170,25 +170,87 @@ jobs:
   tests:
     name: Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
     if: github.event.pull_request.draft == false || github.event_name != 'pull_request'
+    # Tests run via `bazel test //:kura_lib_test` (the rust_test that recompiles kura_lib
+    # with --test so the crate's #[cfg(test)] unit tests run). Same Kura-as-remote-cache
+    # setup as the `compile` job: Kura runs as a normal `docker run` container so it can be
+    # gracefully stopped (SIGTERM -> RocksDB flush) before its data dir is snapshotted by
+    # actions/cache. Flags MATCH the compile job's x86_64 leg (-c opt, same platform,
+    # -Dwarnings) on purpose: that makes every third-party dep's action key identical, so a
+    # fresh tests run reuses compile's warm dep cache (restore-keys falls back to
+    # kura-compile-data-) and only the test-crate compile + test run are new work. There are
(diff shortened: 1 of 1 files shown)


pepicrft (approved):

esnunes:

PR #11167, opened 2026-06-09
ci(kura): build the OCI image with Bazel and run e2e against it

esnunes:
## What

Migrates the OCI image build and the end-to-end suite from the Docker/Buildx path to the **rules_oci** build proven on `kura-bazel-phase1`. This is **Stage A, part 3** of the Kura Bazel migration.

> **Stacked PR** — base is `kura-bazel-tests` (#11166), not `main`. Merge order is #11165 (compile) → #11166 (tests) → this. The diff here is the OCI build foundation + the `e2e-images`/`e2e` job flip.

## Why

The current `e2e-images` job builds `kura/Dockerfile` with `docker compose build` and `docker save`s three images; the `e2e` job loads them and retags them per suite. This PR replaces that with the same QEMU-free rules_oci build that compile/tests already dogfood through Kura.

## What changed

**OCI build into the Bazel graph** (verbatim from `kura-bazel-phase1`):
- `MODULE.bazel`: `rules_oci`/`rules_pkg`/`rules_distroless`, the digest-pinned `debian:bookworm-slim` base (both arches), the hermetic extra-packages layer (`tini`/`curl`/`libstdc++6`/`ca-certificates`, sha256-pinned apt lock), and the digest-pinned DB-IP geoip seed.
- `//bazel/oci`: an `oci_image` matching the production Dockerfile (entrypoint `tini -- kura`, `/opt/geoip/dbip-city-lite.mmdb`, port 4000) + a multi-arch `oci_image_index`. No Docker daemon / Buildx / QEMU — both arches build from one host via the cross toolchains.
- `transition.bzl`: `transitioned_image` makes the loadable tarball reuse the index's compiled binary instead of building the native arch twice (the OCI fragmentation fix).

**`e2e-images` job** — `bazel/ci-oci.sh` builds the multi-arch index + a native-arch docker-loadable tarball, dogfooding Kura as the remote cache (same container + graceful-stop + restore/save split as compile/tests). Uploads the single image tarball as the `kura-e2e-images` artifact.

**`e2e` job** — loads that one image and points every compose service (base + overlay files) at it via `KURA_IMAGE=kura-bazel:oci` + `KURA_E2E_SKIP_BUILD=1`, which removes the per-suite retag loop the Dockerfile path needed. Keeps `rust shellspec bazel buck2` in the mise install because `clients_spec` drives Bazel and Buck2 as REAPI clients against the cluster.

## Cache namespacing

The OCI build compiles the binary under a rules_oci platform transition (an `ST-<hash>` config), a **different** action keyspace than the compile/tests jobs' `--platforms` flag config — so seeding from compile's cache yields 0 hits. This job therefore keeps its own Kura cache namespace (`kura-oci-data-`), whose key also includes `bookworm_apt.lock.json`.

## Risk / coverage

Shadow CI on `kura-bazel-phase1` only ever ran the **`cluster`** shard against the Bazel image. This PR runs **all four gating shards** against it for the first time. The overlay compose files were checked to honor `KURA_IMAGE` (discovery's extra `kura-us-2` uses `${KURA_IMAGE:-}`; extension/mtls only patch existing services and inherit the base image), but the `clients`/`discovery-faults-handoff`/`extension-mtls` shards are exercised against the Bazel image here for the first time — CI is the validation.

The DB-IP geoip URL is pinned to `2026-06`; DB-IP keeps only recent months online, so it needs periodic bumping (re-pin: curl the new dump + `sha256sum`).

## Validation

- `actionlint .github/workflows/kura.yml` and `shellcheck bazel/ci-oci.sh` pass.
- The OCI files + `MODULE.bazel.lock` are byte-identical to `kura-bazel-phase1` (where the full OCI build + `cluster` e2e are green); `MODULE.bazel` differs only in its docstring.
- CI on this PR builds the image and runs all four e2e shards against it.

### DIFF
--- a/.github/workflows/kura.yml
+++ b/.github/workflows/kura.yml
@@ -281,21 +281,84 @@ jobs:
   e2e-images:
     name: E2E Images
     runs-on: ubuntu-latest
-    timeout-minutes: 60
+    timeout-minutes: 90
     if: github.event.pull_request.draft == false || github.event_name != 'pull_request'
-    env:
-      COMPOSE_PROJECT_NAME: kura
+    # Builds the runtime OCI image with rules_oci (//bazel/oci) — the multi-arch index plus
+    # a docker-loadable tarball for the runner's native arch — with no Docker daemon, Buildx,
+    # or QEMU, replacing the `docker compose build` of the bookworm Dockerfile. Dogfoods Kura
+    # as the Bazel remote cache, same as compile/tests. The OCI build compiles the binary
+    # under a rules_oci platform transition (ST-<hash> config), a DIFFERENT action keyspace
+    # than the compile/tests jobs' --platforms flag config, so this job keeps its OWN Kura
+    # cache namespace (kura-oci-data-) — seeding from compile's cache gives 0 hits. The native
+    # tarball is uploaded for the e2e shards to load and run against (KURA_IMAGE + skip-build),
+    # so every shard runs the real bookworm runtime (tini/curl/certs). Still no oci_push — the
+    # release image flip is a later stage.
     steps:
       - uses: actions/checkout@v4
 
-      - name: Show Docker tooling
-        run: docker compose version
+      - name: Restore Kura cache (data dir)
+        id: kura-cache
+        uses: actions/cache/restore@v4
+        with:
+          path: ${{ runner.temp }}/kura-data
+          key: kura-oci-data-${{ hashFiles('kura/MODULE.bazel.lock', 'kura/Cargo.Bazel.lock', 'kura/bazel/oci/bookworm_apt.lock.json') }}
+          restore-keys: |
+            kura-oci-data-
 
-      - name: Prebuild Docker images
-        run: docker compose build
+      - name: Cache Bazel repository downloads
+        uses: actions/cache@v4
+        with:
+          path: ${{ runner.temp }}/bazel-repo
+          key: bazel-repo-oci-${{ hashFiles('kura/MODULE.bazel.lock', 'kura/Cargo.Bazel.lock', 'kura/bazel/oci/bookworm_apt.lock.json') }}
+          restore-keys: |
+            bazel-repo-oci-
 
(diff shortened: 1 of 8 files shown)


pepicrft (approved):
I did a deep pass over the Bazel migration shape, the workflow wiring, the OCI image assembly, and the e2e compose path. I do not see any blocking correctness issues.

The approach looks clean for this stage of the migration: Bazel owns compile/test/image assembly, the e2e suite runs against the Bazel-built OCI tarball, and the compose image override path preserves local behavior while making CI consume the loaded image. The non-hermetic C/C++ toolchain is a real tradeoff, but it is contained in the Debian dev image and matches the runtime base, which is a reasonable choice here.

Validation I ran:

- `git diff --check main...HEAD`
- `docker compose config --quiet`
- `KURA_IMAGE=kura-bazel:oci docker compose config --images`
- `KURA_IMAGE=kura-bazel:oci docker compose -f docker-compose.yml -f test/e2e/docker-compose.discovery.yml config --images`
- `mise exec -- bazel query '//...'`\n- `mise exec -- bazel build --nobuild //:kura //:kura_lib //:kura_lib_test`\n- Built `kura-bazel-dev-review` from `kura/bazel/linux-dev.Dockerfile`\n- Inside that Debian dev image, ran `bazel build --nobuild //bazel/oci:index //bazel/oci:load_linux_<native> //:kura //:kura_lib //:kura_lib_test`\n\nTwo non-blocking cleanup notes for later:\n\n- `kura/bazel/linux-dev.Dockerfile` pins Bazel through `.bazelversion`, but downloads Bazelisk from `latest`. Pinning the Bazelisk release or checksum would make CI more reproducible.\n- The remote-cache setup is repeated across the workflow jobs. I would leave it inline while this is still settling, then extract the repeated rc setup and container invocation once the migration stabilizes.\n\nApproved.

esnunes:
closed in favor of https://github.com/tuist/tuist/pull/11530

PR #11169, opened 2026-06-09
fix(cli): Xcode 27 Swift package deployment floors

pepicrft:
Resolves none

Xcode 27 beta raises the supported deployment target range for package projects, which made Tuist-generated Swift package targets keep invalid historical floors such as iOS 12.0.

This updates Swift package deployment target defaults by Swift compiler era so Swift 6.x packages are clamped to deployment targets supported by newer Xcodes. Swift 5.9 keeps the existing iOS 12.0 behavior, while Swift 6.4, used by Xcode 27 beta, now generates package targets with iOS 15.0, macOS 12.0, watchOS 9.0, tvOS 15.0, and visionOS 1.0 floors.

The change also adds a regression test for Swift 6.4 package mapping and removes stray debug `dump` calls from the neighboring deployment target test.

### How to test locally

- `mise run cli:lint --fix`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistLoaderTests/PackageInfoMapperTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`

### DIFF
--- a/.github/workflows/cli.yml
+++ b/.github/workflows/cli.yml
@@ -64,9 +64,6 @@ concurrency:
 # fallback never crosses structural changes. See #10510 / e8fa267 for the incidents
 # (stale workspace state pointing at moved local SPM paths; Linux linker failures
 # from partial-artifact reuse) that motivate the narrower scope.
-#
-# macos-swifterpm-v2 invalidates caches that can contain registry downloads without
-# Package.swift files, which released Tuist versions cannot consume during graph loading.
 jobs:
   cli-lint:
     name: Lint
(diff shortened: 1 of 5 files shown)


fortmarek (approved):

.github/workflows/cli.yml
```diff
@@ -162,9 +165,9 @@ jobs:
         uses: actions/cache/restore@v4
         with:
           path: .build
-          key: ${{ runner.os }}-cli-${{ hashFiles('.xcode-version') }}-${{ hashFiles('Package.resolved', 'Package.swift') }}
+          key: ${{ runner.os }}-cli-macos-swifterpm-v2-${{ hashFiles('.xcode-version') }}-${{ hashFiles('Package.resolved', 'Package.swift') }}
           restore-keys: |
-            ${{ runner.os }}-cli-${{ hashFiles('.xcode-version') }}-${{ hashFiles('Package.swift') }}-
+            ${{ runner.os }}-cli-macos-swifterpm-v2-${{ hashFiles('.xcode-version') }}-${{ hashFiles('Package.swift') }}-
```
pepicrft: We should reoncile this on the swifterpm side, but until we do, let's invalidate the cache. I'll probably just introduce version without mentioning swifterpm in the hash

PR #11171, opened 2026-06-09
ci(github): cache swifterpm global sources

pepicrft:
Resolves N/A

This updates the SwiftPM-related CI caches now that swifterpm leaves project-local `.build` checkouts with links back to its global source cache. Instead of persisting `.build` for the Tuist CLI and App jobs, the workflows now restore and save `~/.cache/swifterpm`, which is swifterpm's XDG global source cache fallback when `XDG_CACHE_HOME` is unset.

The cache namespace is bumped to `v2` and renamed with `swifter` in the key so CI does not reuse archives created with the old `.build` behavior. Plain SwiftPM-only jobs, such as Linux SwiftPM builds and the server XCActivityLog NIF test, continue using their existing `.build` caches.

### How to test locally

- `ruby -e 'require "yaml"; ARGV.each { |path| YAML.load_file(path); puts "ok #{path}" }' .github/workflows/cli.yml .github/workflows/app.yml .github/workflows/server.yml`
- Inspected `tuist/swifterpm` 0.7.1 source and README to confirm the default global cache path is `$XDG_CACHE_HOME/swifterpm`, or `~/.cache/swifterpm` when `XDG_CACHE_HOME` is unset.


.github/actions/materialize-build-cache/action.yml
```diff
@@ -0,0 +1,46 @@
...
+        parent_path="$(dirname "$cache_path")"
+        cache_name="$(basename "$cache_path")"
+        materialized_path="$parent_path/.${cache_name}.materialized.$$"
+
+        rm -rf "$materialized_path"
+        mkdir -p "$materialized_path"
+        trap 'rm -rf "$materialized_path"' EXIT
+
+        cp -R -L -p "$cache_path"/. "$materialized_path"/
+        rm -rf "$cache_path"
```
pepicrft: This symlinking materialization is complexity I'll compress in Once scripts :)

PR #11177, opened 2026-06-09
fix(kura): fallback legacy project auth after inactive introspection

pepicrft:
> Fixes the Kura compatibility regression where project-scoped cache requests from older CLIs can be rejected after Kura introspection reports the token inactive.

Kura now falls back to the existing `/api/cache/access` authorization path for project-scoped requests when `/oauth2/introspect` returns `active: false`. This matches the legacy cache path that older CLIs already rely on, while keeping account-scoped Kura authorization tied to introspection and preserving 401s when the fallback endpoint also rejects the token.

This fixes the backward-compat acceptance failure where a CLI logs in before creating a throwaway project, receives a token whose embedded cache grants do not include that project, then gets a Kura 401 on the first artifact request.

### How to test locally

- `cd kura && mise exec -- cargo fmt --check`
- `cd kura && mise exec -- cargo test extension::tests::tuist_hook -- --nocapture`

Note: `mise` emitted a postinstall warning about `apple.swift-protobuf` while resolving the repo toolchain, but the Kura test command completed successfully.

### DIFF
--- a/kura/ops/helm/kura/hooks/tuist.lua
+++ b/kura/ops/helm/kura/hooks/tuist.lua
@@ -323,6 +323,11 @@ local function authenticate_via_introspection_endpoint(token, authorization, tar
       }
     end
 
+    -- Legacy CLI compatibility: older clients can hold a token whose
+    -- embedded grants predate a newly-created project. The server's
+    -- /api/cache/access path still authorizes those project-scoped
+    -- requests, so keep this fallback until those CLI versions are no
+    -- longer supported.
     if target.scope == "project" then
       return authenticate_via_cache_access_endpoint(authorization)
     end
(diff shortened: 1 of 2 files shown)


kura/ops/helm/kura/hooks/tuist.lua
```diff
@@ -321,6 +323,10 @@ local function authenticate_via_introspection_endpoint(token)
       }
     end
 
+    if target.scope == "project" then
```
fortmarek, line 331: I'd mention here that we do this for legacy reasons to support older CLI versions and that this should be possible to remove in the future.

fortmarek (approved):

kura/ops/helm/kura/hooks/tuist.lua
pepicrft: Added the note in `authenticate_via_introspection_endpoint` explaining that this branch is legacy CLI compatibility for project-scoped requests whose embedded grants predate project creation, and that it can go away once those CLI versions are no longer supported.

Validation: `cd kura && mise exec -- cargo fmt --check` and `cd kura && mise exec -- cargo test extension::tests::tuist_hook -- --nocapture`.

PR #11183, opened 2026-06-09
feat(server): add reliability automation monitor

fortmarek:
## What changed

- Added `reliability_rate` as an automation monitor type alongside flakiness rate and flaky run count.
- Wired reliability monitor evaluation through alert validation, worker dispatch, API schemas/controllers, and the Project Automations LiveView UI.
- Added ClickHouse success-count aggregate state for daily and rolling per-test-case evaluation, with backfills and companion materialized views.
- Updated data export documentation and gettext templates for the new automation UI strings.
- Added focused tests for alert validation, monitor evaluation, worker dispatch, API creation, and LiveView form defaults.

## Why it changed

Tests that fail nearly every run can have very low reliability while still not matching Tuist's flaky-test definition, because flakiness is based on flaky runs rather than a simple success-rate drop. This adds the automation Curtis asked for: quarantine/mute/skip tests when their reliability falls below a configured threshold.

## Root cause

The existing automation model could react to flakiness rate or flaky run count, but it did not have a monitor that measured successful runs over total runs. Very unreliable tests with no flaky runs were therefore invisible to automatic quarantine rules.

## Impact

Project maintainers can now create automations such as "reliability rate < 90% over 30 days" or over the last N runs. Existing flakiness and event-driven automations keep their current behavior.

## Validation

- `mix format`
- `mix gettext.extract`
- `git diff --check`
- Focused server tests: `mix test test/tuist/automations/alerts/alert_test.exs test/tuist/automations/monitors/flaky_tests_monitor_test.exs test/tuist/automations/workers/alert_evaluation_worker_test.exs test/tuist_web/controllers/api/automations/alerts_controller_test.exs test/tuist_web/live/project_automations_live_test.exs` (`107 tests, 0 failures`)


server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -277,6 +277,14 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
     FlakyTestsMonitor.evaluate_by_run_count(alert, test_case_ids)
   end
 
+  defp evaluate_monitor(%{monitor_type: "reliability_rate"} = alert, nil) do
```
pepicrft, line 280: This adds the worker dispatch for `reliability_rate`, but the surrounding scheduling/scoped-evaluation allowlists still only include `flakiness_rate` and `flaky_run_count`. `Tuist.Automations.@flaky_monitor_types` filters the ingestion-driven scoped jobs, and `AutomationScheduler.scheduled_alert?/1` only suppresses established rolling alerts for those two types. As a result, established rolling reliability alerts will keep getting full-project cadence evaluations instead of the scoped insert-driven path, which is the expensive path this rolling architecture is meant to avoid. I would include `reliability_rate` in that shared monitor set and add coverage in `AutomationsTest` and `AutomationSchedulerTest`.

server/priv/marketing/changelog/2026.06.09-test-reliability-automations.md
```diff
@@ -0,0 +1,9 @@
+---
+title: Test reliability automations
+category: "Product"
+pull_request: 11183
+---
+
+Some tests are not intermittently flaky; they are broken, or failing so often that they drown out useful signal. Automations can now monitor **Test reliability**, the same success-rate metric shown on the test case detail page, and trigger when it drops below your chosen threshold. Pair it with actions like changing the test's quarantine state, marking it as flaky, or notifying Slack, so broken or consistently failing tests can be handled automatically.
```
pepicrft: The copy says this is the same success-rate metric shown on the test case detail page, but the new automation aggregate is not branch-aware. The detail-page metric first calculates reliability on the project default branch and only falls back to all branches when there are no default-branch runs. These MVs group by project/date/test case only, and the monitor query has no branch predicate, so feature-branch failures can trigger an automation for a test that is healthy on the default branch, or passing feature-branch runs can hide default-branch failures. If the all-branch behavior is intentional, I would adjust the UI/changelog wording. If the advertised behavior is intended, the aggregate needs to carry enough branch information to match the existing metric.

pepicrft (approved):
Approved with the two comments above. I would address them before merging, but the overall direction looks good.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
fortmarek: Good catch. I added `reliability_rate` to the scoped monitor set in `Tuist.Automations` and to the rolling-alert skip guard in `AutomationScheduler`, so established rolling reliability alerts now use the same insert-driven scoped path as flakiness/runcount monitors instead of the full-project cadence path.

I also added coverage in `AutomationsTest` for ingestion-driven reliability evaluation jobs and in `AutomationSchedulerTest` for established rolling reliability alerts. The Server workflow is green on `c09953c2395763b3f11680b01199e0e158334ac4`.

server/priv/marketing/changelog/2026.06.09-test-reliability-automations.md
fortmarek: I kept the automation behavior as all-branch aggregation for this PR and adjusted the product copy to avoid implying that it matches the default-branch-first detail-page metric.

The automation summary now says `test reliability across branches`, the gettext template was refreshed, and the changelog now describes the monitor as operating across branches rather than calling it the same metric as the test case detail page. CI is green on `c09953c2395763b3f11680b01199e0e158334ac4`.

PR #11184, opened 2026-06-09
fix(kura): sign legacy module downloads

pepicrft:
## What changed

- Added a Tuist-managed Kura `response_headers` hook that signs successful `GET /api/cache/module/{id}` responses with `x-tuist-signature`.
- The signature payload is the module artifact hash, matching the existing Tuist server cache controller behavior.
- Added Kura extension tests for successful signing, missing signer configuration, and failed downloads.
- Added optional Helm support to project `KURA_EXTENSION_SIGNER_TUIST_LICENSE_SECRET` into Kura shared secrets when a `licenseSigningKey` field is configured.

## Why

Kura now serves module cache downloads directly for managed deployments. The Tuist server still signs these responses through its cache controller, but Kura did not emit the legacy `x-tuist-signature` header. This restores that server-side cache API contract on Kura without adding Tuist-specific Rust paths.

## Root cause

The Kura hook covered authentication and authorization only. Once clients were routed to Kura for `/api/cache/module/{id}`, successful module downloads bypassed the server plug that signs responses with the Tuist license metadata `signingKey`.

## Important compatibility note

I checked the `4.150.0` CLI used by the backward compatibility job. Its `DownloadModuleCacheService` returns the binary response body and the generated OpenAPI model for `GET /api/cache/module/{id}` does not expose an `x-tuist-signature` response header, so this PR should be treated as compatibility hardening rather than proof that the current `4.150.0` acceptance failure is fixed.

I also checked 1Password without printing secret values. The canary and production `kura-introspection-oauth-client` items do not currently include a signer field, and the old Kamal cache secrets do not carry a Tuist license key or signing key. The chart only mounts the signer when `kuraController.sharedSecrets.externalSecret.fields.licenseSigningKey` is set, so the missing 1Password field needs to be added before managed Kura nodes can emit the header in canary or production.

## Validation

- `cd kura && mise exec -- cargo fmt`
- `cd kura && mise exec -- cargo test tuist_hook`
- `cd kura && mise exec -- cargo clippy --all-targets -- -D warnings`
- `helm template tuist infra/helm/tuist -f infra/helm/tuist/values-managed-common.yaml -f infra/helm/tuist/values-managed-canary.yaml -f infra/helm/tuist/values-ci.yaml --set server.image.tag=test --set kuraController.image.tag=test --set kuraRuntime.image.tag=test --namespace tuist-canary`
- `helm template tuist infra/helm/tuist -f infra/helm/tuist/values-managed-common.yaml -f infra/helm/tuist/values-managed-canary.yaml -f infra/helm/tuist/values-ci.yaml --set server.image.tag=test --set kuraController.image.tag=test --set kuraRuntime.image.tag=test --set kuraController.sharedSecrets.externalSecret.fields.licenseSigningKey=KURA_EXTENSION_SIGNER_TUIST_LICENSE_SECRET --namespace tuist-canary`


pepicrft:
Closing this because the signature/license path is not the failing compatibility surface for CLI 4.150.0. The replacement fix is in #11186 and keeps the old CLI on the module-cache pull path while avoiding the unrelated old activity-log parser during setup.

PR #11185, opened 2026-06-09
build(kura): Bazel build + test foundation via rules_rs

esnunes:
## What

A self-contained Bazel **build + test foundation** for Kura: it builds the kura crate (lib, binary, and unit tests) under Bazel and makes Bazel the compile/test gate in CI. The third-party crate graph is resolved by **`rules_rs`**. Rebased cleanly on `main` and self-contained.

## Why

Kura needs a Bazel build so it can be cached and (later) imaged through the same Bazel/REAPI path as the rest of the stack.

The crate graph is resolved by **`rules_rs`** rather than the conventional `rules_rust` `crate_universe`, for **host-reproducibility.** `crate_universe` pins the resolved graph in a generated lockfile whose digest is host-dependent — a lock generated on macOS is rejected by Linux CI and vice versa, so there is no single committed lock both accept (Kura is developed on macOS and built/released on Linux). `rules_rs` resolves the graph directly from `Cargo.toml`/`Cargo.lock` against an explicit `platform_triples` set, so the same checked-in tree builds identically on macOS and Linux with nothing to regenerate. `rules_rs` is the crate-graph ruleset used by the `rules_rust`-recommended starter (`bazel-starters/rust`).

CI must also build untrusted fork PRs without exposing the remote-cache credentials, and a Tuist/cache outage must not block merging a Kura PR.

## Contents

**Build + test foundation** — enough for `bazel build //...` and `bazel test //...`:
- `MODULE.bazel` (+ `.lock`), `.bazelrc`, `.bazelversion` (Bazel 9.1.0), `BUILD.bazel`, `bazel/third_party/{lua,lz4}/`, plus `.gitignore` / `tuist.toml`.
- `rules_rs` 0.0.93 (which provisions a patched `rules_rust` fork plus its own rust rules and toolchains), `rules_cc` 0.2.19, `platforms` 1.1.0, BCR `lz4` 1.10.0.bcr.1; Rust 1.94.1, edition 2024. The autodetected host C/C++ toolchain compiles the `-sys` crates (rocksdb, aws-lc, jemalloc, lua) from source.
- Targets: `//:kura_lib`, `//:kura` (binary), `//:kura_lib_test`. Rust rules load from `@rules_rs//rs:rust_{binary,library,test}`; deps come from the `@crates` hub (`all_crate_deps` / `aliases`).

**CI** (`.github/workflows/kura.yml`):
- `bazel-compile` / `bazel-test` (+ `-fork` variants) — the compile/test gate, alongside the cargo `clippy` / `format` / `audit` jobs. Non-fork runs get `id-token: write`, authenticate via OIDC, and use the Tuist remote cache; fork runs get `contents: read` only and the local repo/disk cache.
- Each job invokes a mise task (`mise run compile` / `mise run test-unit -- --define=kura_deny_warnings=true`), so the build/test commands have a single definition shared with local dev.

**Developer experience** (`kura/mise/tasks/`, `AGENTS.md`, `README.md`):
- `mise run compile` — build all targets for the host.
- `mise run test-unit` — `bazel test //...`.
- The crate graph resolves from `Cargo.lock` on each build, so changing Rust deps just updates `Cargo.lock` as usual. `AGENTS.md` and `README.md` document Bazel as the primary build/test path, with cargo as the fallback.

## Key decisions

- **`rules_rs` for the crate graph** (see Why). The trade-off is a dependency on `rules_rs` (v0.0.x) and the patched `rules_rust` fork it provisions, accepted for host-reproducible, regeneration-free crate resolution.
- **rocksdb/lz4 — working around a fork build-script bug.** `librocksdb-sys` compiles RocksDB's C++ with the lz4 backend (`#include <lz4.h>`) and resolves the include dir from `$DEP_LZ4_INCLUDE`. The patched `rules_rust` fork that `rules_rs` provisions mis-propagates that path across crates — it leaves an unresolved `${out_dir}` token in the dependent's build-script env, so rocksdb fails with `'lz4.h' file not found`. Workaround: turn `lz4-sys`'s build script off, link the BCR `@lz4` cc_libraries (`lz4` + `lz4_hc` + `lz4_frame` — rocksdb needs `lz4hc.h` and the HC object code too) into it, and set `DEP_LZ4_INCLUDE` on `librocksdb-sys` ourselves to a materialized header directory (a small `include_dir` rule, since cc-rs's `config.include()` needs a directory and `$(execpath)` of a header is a file). Reported upstream as [hermeticbuild/rules_rs#163](https://github.com/hermeticbuild/rules_rs/issues/163); an upstream fix removes the workaround. Only `lz4` is enabled in rocksdb's feature set here.
- **Lua via a `cc_library`, not lua-src's build script.** `mlua-sys`'s build script resolves its source dir from a compile-time `CARGO_MANIFEST_DIR` that does not survive Bazel sandboxing, so its build script is disabled and Lua is compiled as our own `cc_library` from `lua-src`'s vendored sources, with the platform define selected per-OS (`LUA_USE_LINUX` / `LUA_USE_MACOSX`). `rules_rs`'s `extra_aliased_targets` re-exposes lua-src's source/header filegroups on the `@crates` hub for that `cc_library` to consume.
- **`-D warnings` only in CI.** A `kura_deny_warnings` config_setting gates `-Dwarnings` on the first-party crates; it is off by default (warning-tolerant local builds, like `cargo build`) and CI passes `--define=kura_deny_warnings=true`.
- **The remote cache is an accelerator, not a gate.** The remote cache *is* Kura, so a Kura/Tuist incident is exactly when CI matters most. `tuist auth login` / `tuist bazel setup` run with `continue-on-error` and `.bazelrc` `try-import`s the generated config, so when auth/setup fails there is no remote-cache config and the build runs cold against the local cache (a `::warning::` makes that visible) instead of blocking the merge. Fork PRs never touch the remote cache or the OIDC token. (Bazel has no graceful local fallback for a *gRPC* cache that errors mid-build — unlike HTTP caches — so `tuist bazel setup` probing the endpoint before writing the config is the upstream complement to this.)
- **Recursive `//...`.** Build/test patterns use `//...` rather than the non-recursive `//:all`, so a test added in a subpackage is picked up automatically instead of being silently skipped.
- **Native only, no cross-compilation.** Each release arch is built on a runner of that arch with the host toolchain; the runtime OCI image is a later change. macOS is a local-dev target only (not released); `.bazelrc` raises the macOS min-OS floor so rocksdb's C++17 aligned new/delete compiles.

## Developer / contributor impact

- Local loop: `mise install` (Rust + Bazel), then `mise run compile` / `mise run test-unit`. Cargo still works as a fallback.
- The Tuist remote cache (`tuist bazel setup`) is available only with access to the `tuist/kura` project on Tuist; without it, Bazel builds against the local cache.
- Fork PRs build and test against the local cache with `contents: read` only — no credentials.

## Validation

- `bazel build //...` and `bazel test //...` (`//:kura_lib_test` PASSED) pass on both `aarch64-apple-darwin` (local) and `tuist-linux` (the branch's `bazel-compile` / `bazel-test` CI jobs) — building rocksdb/aws-lc/jemalloc/ring/lua and the `kura` binary natively on each.
- The rocksdb/lz4 fix was root-caused and confirmed in a standalone minimal project (rocksdb + lz4 on a bare `rules_rs` config): it fails with `'lz4.h' file not found` without the fix and builds + runs (round-trips a value through a real RocksDB) with it — an A/B proof that the issue is the fork bug, not anything kura-specific.


.github/workflows/kura.yml
```diff
@@ -53,28 +53,125 @@ jobs:
...
+          key: bazel-repo-lockfile-${{ hashFiles('kura/MODULE.bazel.lock', 'kura/Cargo.Bazel.lock') }}
+          restore-keys: |
+            bazel-repo-lockfile-
+
+      - name: Check Cargo.Bazel.lock is up to date
+        run: |
+          echo "common --repository_cache=${{ runner.temp }}/bazel-repo" >> "$HOME/.bazelrc"
+          mise run bazel-repin check
+
+  # Split into non-fork (`bazel-compile`) and fork (`bazel-compile-fork`) like
```
pepicrft: LLMs tend to be very verbose when it comes to comments, and this makes me wonder if we need all these comments here or it's something can be inferred from the code itself.

.github/workflows/kura.yml
```diff
@@ -53,28 +53,125 @@ jobs:
...
+      # The Tuist remote cache is an accelerator, not a gate: an auth/setup outage must not block
+      # merging a Kura PR (the cache *is* Kura, so this is exactly when CI is needed most). Auth and
+      # setup are best-effort (continue-on-error); `.bazelrc` try-imports `.bazelrc.tuist`, so if they
+      # fail the build runs cold against the local cache. (A gRPC cache reachable at setup but erroring
+      # mid-build still fails — Bazel has no local fallback for a gRPC cache.)
+      - name: Authenticate with Tuist
+        run: tuist auth login
+        continue-on-error: true
+      - name: Tuist Bazel setup
+        run: tuist bazel setup && cat .bazelrc.tuist
```
pepicrft: Why do we need the cat here? Isn't there a risk of leaking sensitive information?

.github/workflows/kura.yml
```diff
@@ -53,28 +53,125 @@ jobs:
...
+        continue-on-error: true
```
pepicrft, line 87: Why do we want to continue making errors?

.github/workflows/kura.yml
```diff
@@ -53,28 +53,125 @@ jobs:
...
+      - name: Warn when the Tuist remote cache is unavailable
+        if: ${{ hashFiles('kura/.bazelrc.tuist') == '' }}
+        run: echo "::warning::Tuist remote cache unavailable (.bazelrc.tuist not written); building cold against the local cache."
```
pepicrft, line 90: I'd rather make this fail so we can an extra signal to spot if something is not going as expected.

.github/workflows/kura.yml
```diff
@@ -53,28 +53,125 @@ jobs:
...
+
+      - name: Compile without warnings
+        run: mise run bazel-compile -- --define=kura_deny_warnings=true
```
pepicrft: What if we name the tasks `mise run compile` and use Mise's usage annotations to declare arguments and then pass someting like `mise run compile --build-system bazel/cargo`

.github/workflows/kura.yml
```diff
@@ -53,28 +53,125 @@ jobs:
...
+
+  bazel-compile-fork:
+    name: Bazel Compile Fork
+    runs-on: tuist-linux
+    timeout-minutes: 60
+    # Fork PRs only: `contents: read`, never `id-token: write`. Skips `tuist auth login`
+    # and `tuist bazel setup` (the OIDC consumers) and builds against the local
+    # repository/disk caches alone.
```
pepicrft: Can't we achieve this with control flow at the step level instead of having a separate job for it?

kura/BUILD.bazel
```diff
@@ -0,0 +1,73 @@
+# Root package for the Kura Bazel workspace.
+#
+# Exposes the Cargo manifests so the crate_universe extension can read them as
+# labels, and defines the kura library + binary targets.
+
+load("@crates//:defs.bzl", "aliases", "all_crate_deps")
+load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_library", "rust_test")
+
+exports_files([
```
pepicrft: What is this for?

pepicrft (approved):
Two non-blocking findings from the review:

1. `.github/workflows/kura.yml` makes `tuist auth login` and `tuist bazel setup` best-effort. I validated the latest GitHub Actions run and `tuist bazel setup` failed with “None of the cache endpoints are reachable,” then the Bazel compile continued cold and passed. That matches the intent in the workflow comments, but if this pull request should validate the remote cache path too, we should make that a separate required signal or fail this step.
2. `.github/workflows/kura.yml` prints `.bazelrc.tuist` after setup. The generated file does not include the token itself, based on `BazelSetupCommandService`, but it exposes endpoint, account, project, and credential-helper path in logs. I would remove the `cat` since it does not add much diagnostic value.

I validated the lockfile guard, Bazel compile with warnings denied, Bazel unit tests, cargo clippy with warnings denied, and `git diff --check`. The latest GitHub Actions run is green, including Bazel compile, Bazel test, linting, audit, and end-to-end shards. Approving with the above as follow-ups.

.github/workflows/kura.yml
esnunes: It is not needed, I've been using it to know which cache server was selected. Regarding leaking data, this file contains simple configuration, no credentials are exposed. I will remove the cat, or change it to print only the selected cache server (maybe expose it as a job output for debugging purpose)

.github/workflows/kura.yml
esnunes: Those steps prepare the remote cache configuration, when they fail, the build falls back to local cache, without this flag the build would fail, preventing us from shipping kura fixes for example

.github/workflows/kura.yml
esnunes: I will do a cleanup, there are many comments, most of them are too verbose and some can be easily inferred

.github/workflows/kura.yml
esnunes: If it fails the workflow will fail and it will cascade to the github check, preventing us from merging fixes that might bring the cache system back, ignoring the failure makes bazel fallback to local cache. I will evaluate other alternatives outside of this PR, notifying us on slack could be a good alternative

.github/workflows/kura.yml
esnunes: Github Action permission are set at the workflow or job level, this is the pattern already used in the `gradle.yml`, `android.yml` and `cli.yml` workflows. I don't have access to the repository settings to better understand whether this is required or it is an outdated pattern. I will address this in parallel to this PR

.github/workflows/kura.yml
esnunes: I've added this to my backlog, a single entrypoint for both cargo and bazel sounds good

kura/BUILD.bazel
esnunes: it is not, it is used to tell bazel that those files are available for reference from other packages, they are indeed referred by the MODULE.bazel but it is within the same package, not requiring them to be exported, thanks for bringing this up!

PR #11210, opened 2026-06-10
feat(server): build/test duration metrics API + Grafana data source plugin

fortmarek:
## Purpose

Make build and test run durations queryable as time series with percentile metrics (p50/p90/p99 and average) over a chosen time range, and ship a Grafana data source plugin that consumes them. This came out of a request to measure how much CI pipelines are getting optimized over time and to feed those numbers into Grafana dashboards.

The percentile math already existed internally (it powers the dashboard); the gap was a public, queryable API and a client. This PR closes both.

## What changed

**Server: build/test duration metrics API**
- New `TuistWeb.API.MetricsController` exposing, per project:
  - `GET .../builds/metrics/duration` and `.../tests/metrics/duration` — time-bucketed `average` + `p50`/`p90`/`p99` series.
  - `.../builds/metrics/schemes`, `.../builds/metrics/configurations`, `.../tests/metrics/schemes` — to back filter dropdowns and template variables.
- New `Builds.Analytics.build_duration_percentiles_analytics/2` returning average + p50/p90/p99 in a single grouped ClickHouse query. The tests equivalent (`test_run_duration_analytics/2`) already returned this shape and is reused as-is.
- New `DurationMetrics` OpenApiSpex response schema: `{ dates, average{values,total}, p50, p90, p99, trend }`, durations in ms, dates as Unix seconds.

**New component: `grafana-datasource/`** — a Grafana data source plugin (Go backend + React), a thin client over the endpoints above:
- `QueryData` dispatches on `queryType` (`buildDuration`/`testDuration`) and renders a wide time-series frame (one field per selected series, unit ms). Default series is average + p50/p90/p99, matching the dashboard.
- `CallResource` + `metricFindQuery` proxy `GET /api/projects` and the `.../metrics/schemes`/`configurations` endpoints, so query-editor dropdowns and dashboard variables resolve without exposing the token.
- Filters: **Project** and **Environment** (CI/Local) are both exposed per-panel and as dashboard variables (via `applyTemplateVariables`), so one dashboard works across every project the account token can reach, mirroring the builds page filters. Scheme/configuration filters too.
- `ConfigEditor` stores the server URL in `jsonData` and the account token in `secureJsonData` (encrypted, never sent to the browser).
- Built on the standard `@grafana/create-plugin` harness. The `.config/` directory is vendored, auto-generated scaffolding (webpack/jest/eslint/tsconfig + the Docker dev-server) carrying a DO-NOT-EDIT banner and pinned via `.config/.cprc.json` — it is refreshed with `npx @grafana/create-plugin@latest update`, not hand-edited, so it is not worth reviewing line by line. What is actually ours: `pkg/` (Go backend), `src/` (React/TS), `plugin.json`, the root config-extension files, and the release/CI wiring. The Go backend lives in its own nested `go.work` (see SDK isolation below); the component is added to the root AGENTS.md repository map.

## Why these choices

- **Endpoint placement** — the metrics hang off the existing `/builds` and `/tests` resources behind a `/metrics` infix (GitHub `{entity}/stats` style) instead of a dedicated `/analytics` namespace, which is already overloaded by the ingest `POST /analytics` and the page-view AnalyticsPlug. The infix also avoids colliding with `/:build_id`.
- **Auth** — reuses the existing `AuthorizationPlug` (`:build`/`:test`), so account tokens with `project:builds:read` / `project:tests:read` work with no new authorization category. The plugin authenticates with an account token, which gives the account-wide project dropdown for free via `GET /api/projects`.
- **ClickHouse load** — `from`/`to` come from the consumer's time range; the server derives bucket granularity (hour/day/month) from the range, which bounds the point count (~60 buckets max). A max-range guard rejects ranges over 366 days. Reads go through the read replica (`Tuist.ClickHouseRepo`). A 60s response cache (`KeyValueStore.get_or_update`, keyed per project/range/filters) collapses the Grafana refresh storm to one query, and a per-account rate-limit plug (`MetricsRateLimitPlug`, 300 req/min, in-memory token bucket) caps abuse. `build_runs` is `ORDER BY (project_id, inserted_at, id) PARTITION BY toYYYYMM(inserted_at)`, so each query is an indexed range scan with partition pruning, not a full scan.
- **Backend plugin** (Go, like grafana/sentry-datasource) keeps the token server-side and enables Grafana alerting.
- OpenApiSpex only; the CLI client is intentionally not regenerated (these endpoints are not for the CLI).

## How to test locally

Server:
```
cd server
mix test test/tuist/builds/analytics_test.exs test/tuist_web/controllers/api/metrics_controller_test.exs
```
Then, with the dev server running, hit (account-token auth):
```
GET /api/projects/<account>/<project>/builds/metrics/duration?from=<unix>&to=<unix>&is_ci=true
GET /api/projects/<account>/<project>/tests/metrics/duration?from=<unix>&to=<unix>
```

Plugin:
```
cd grafana-datasource   # Node pinned to 24 via mise.toml
npm install
npm run typecheck && npm run lint && npm run build
go build ./... && go test ./...
```

## Validation run

- Server: 2 new analytics unit tests + 8 controller tests (user and account-token paths, `is_ci` filter, 400 range validation, 403 forbidden); full builds-analytics + controller suites green (80 tests). Compiles with `--warnings-as-errors`, credo clean.
- Plugin backend: `go build` / `go vet` / `go test` green, gofmt clean.
- Plugin frontend: `npm run typecheck`, `npm run lint` (0 errors), `npm run build` (production) all green against `@grafana/*` 12.4.2.
- **End to end against a running Grafana → plugin → local server**: datasource health "Connected to Tuist"; queries return real series (build p50 ~38s / p90 ~89s / p99 ~99s); the Environment filter changes results (CI ~41.6s vs Local ~35.0s); switching Project changes the series. CI coverage for the backend is mocked (`httptest`): `CallResource` path dispatch + `CheckHealth` scope validation in `datasource_test.go`, frame building in `frames_test.go`.

## Screenshot

The Grafana dashboard rendering build and test duration percentiles (average, p50, p90, p99) from a Tuist server, with the Project and Environment filters as dashboard variables:

Grafana dashboard with Project and Environment dropdowns showing Tuist build and test duration percentiles over the last 30 days

A marketing changelog entry ships with this change at `server/priv/marketing/changelog/2026.06.10-grafana-data-source.md`.

- Grafana catalog signing/publishing.
- Optional `build_runs`/`test_runs` daily-stats materialized views if raw scans get hot.

## Publishing the plugin

Prepared in this PR (catalog readiness, verified with `@grafana/plugin-validator`):
- `plugin.json` with the catalog 3-part id `tuist-metrics-datasource`, the Tuist logo, a dashboard screenshot, and description/keywords/links.
- A catalog README (what it does, requirements, configuration, and unsigned/private/catalog install paths) with absolute image/link URLs.
- Release wiring follows the repo convention (git-cliff + push-to-main, like every other component): `grafana-datasource` is registered in `mise/tasks/release/components.json` with its own `cliff.toml`, and `.github/workflows/grafana-datasource-release.yml` triggers on a push to `main` touching `grafana-datasource/**`, computes the next version with git-cliff, and — only when there are releasable commits — builds the frontend + multi-platform Go binaries, signs when `GRAFANA_ACCESS_POLICY_TOKEN` is set, validates, tags `grafana-datasource@x.y.z`, and publishes a GitHub release with the zip + SHA1. Checks run separately on PRs in `.github/workflows/grafana-datasource.yml` (typecheck/lint/test/build for the frontend, gofmt/vet/test for the backend), one job per check, mirroring the other components.

Post-merge / pre-submission (need a grafana.com account + maintainer action, so not doable in the PR):
1. Register/confirm the `tuist` org on grafana.com — the plugin id prefix must match it.
2. Create a `GRAFANA_ACCESS_POLICY_TOKEN` (plugins:write) and add it to repo secrets — do this before merge if you want the first release signed.
3. Merging this PR auto-releases `grafana-datasource@0.1.0` (push-to-main, via the workflow above): signed if the token from step 2 is set, otherwise an unsigned build. Later `feat:`/`fix:` commits touching the plugin bump it automatically.
4. Submit at grafana.com → My Plugins → Submit New Plugin (zip URL + SHA1 + public source URL). Grafana reviews (code + security + install testing), grants the Community signature, and lists it.

SDK / dependency isolation: the plugin lives in its **own** `go.work` (`use .`, go 1.26.3), separate from the repo-root infra workspace, so it tracks the **latest** `grafana-plugin-sdk-go` without forcing its k8s/transitive deps onto the infra controllers (sharing the workspace had broken their build via MVS — apimachinery `structured-merge-diff` v4-vs-v6). With the latest SDK, `@grafana/plugin-validator` no longer flags the "SDK older than 5 months" check or the grpc/otel/kin-openapi CVEs. Verified: the plugin builds and tests on go 1.26.3 with the latest SDK, and the infra modules build unchanged.

Using it before the catalog listing: self-hosted Grafana via the unsigned allowlist (`GF_PLUGINS_ALLOW_LOADING_UNSIGNED_PLUGINS=tuist-metrics-datasource`) or a private signature; Grafana Cloud needs the published catalog version.


server/lib/tuist_web/controllers/api/metrics_controller.ex
```diff
@@ -0,0 +1,264 @@
...
+    |> put_status(:bad_request)
+    |> json(%{message: message})
+  end
+
+  # Short-lived cache so a dashboard refreshing many identical panels (or several
+  # viewers on the same dashboard) collapses to a single ClickHouse query per
+  # project/range/filters within the window.
+  defp cached_duration_metrics(entity, project_id, opts, func) do
+    KeyValueStore.get_or_update(
+      [:metrics_duration, entity, project_id, :erlang.phash2(opts)],
```
pepicrft: The cache key hashes the exact `start_datetime`/`end_datetime`, but Grafana sends `q.TimeRange.To.Unix()` which advances every refresh. So sequential refreshes of a 'last 24h' panel never share a key, only concurrent ones do. Either bucket-align `from`/`to` to the chosen interval before hashing, or update the comment so it's clear this only collapses concurrent panels.

server/lib/tuist_web/controllers/api/metrics_controller.ex
```diff
@@ -0,0 +1,264 @@
...
+           properties: %{values: %Schema{type: :array, items: %Schema{type: :string}}}
+         }},
+      bad_request: {"The request was invalid", "application/json", Error},
+      forbidden: {"You don't have permission to access this resource", "application/json", Error}
+    }
+  )
+
+  def build_dimension_values(%{assigns: %{selected_project: project}} = conn, %{dimension: dimension}) do
+    case dimension do
+      "scheme" -> json(conn, %{values: Builds.project_build_schemes(project)})
```
pepicrft: Dimension endpoints are rate-limited but not cached. CheckHealth probes both `builds` and `tests` scheme every health cycle, and QueryEditor refetches on mount and project change. A 5 to 15 min TTL here would take real load off ClickHouse since scheme/configuration lists change slowly.

server/lib/tuist_web/controllers/api/metrics_controller.ex
```diff
@@ -0,0 +1,264 @@
...
+  # project/range/filters within the window.
+  defp cached_duration_metrics(entity, project_id, opts, func) do
+    KeyValueStore.get_or_update(
+      [:metrics_duration, entity, project_id, :erlang.phash2(opts)],
+      [ttl: to_timeout(minute: 1)],
+      func
+    )
+  end
+
+  defp validate_range(%{from: from, to: to}) do
```
pepicrft, line 265: No lower bound on `from`. Negative or extreme values let `DateTime.from_unix!` raise and surface as a 500. A `from < 0` guard returning a 400 closes that off cheaply.

grafana-datasource/pkg/plugin/datasource.go
```diff
@@ -0,0 +1,171 @@
...
+	if len(projects) == 0 {
+		return healthError("The token has no accessible projects. Grant it access to at least one project."), nil
+	}
+
+	// Listing projects only needs an account token, so probe a metric endpoint to
+	// confirm the read scopes — otherwise a token missing project:builds:read /
+	// project:tests:read reports healthy while every panel query fails with 403.
+	project := projects[0].FullName
+	_, buildErr := d.client.dimensionValues(ctx, entityBuilds, "scheme", project)
+	_, testErr := d.client.dimensionValues(ctx, entityTests, "scheme", project)
```
pepicrft: These two probes run sequentially, so worst case is 2x the client's 30s timeout (60s+) before CheckHealth returns. Easy to parallelize with errgroup so the cap is one timeout, not two.

server/lib/tuist_web/plugs/metrics_rate_limit_plug.ex
```diff
@@ -0,0 +1,22 @@
...
+  def init(opts), do: opts
+
+  def call(%Plug.Conn{} = conn, _opts) do
+    case RateLimit.Metrics.hit(conn) do
+      {:allow, _} ->
+        conn
+
+      {:deny, _} ->
+        conn
+        |> put_status(:too_many_requests)
```
pepicrft, line 20: The 429 has no `Retry-After` or `X-RateLimit-*` headers, so clients (including the Grafana plugin) can't back off intelligently. `InMemory.hit/4` already returns `{:deny, scale_ms}` you could translate into `Retry-After`.

grafana-datasource/src/components/QueryEditor.tsx
```diff
@@ -0,0 +1,153 @@
...
+            <Select
+              isClearable
+              options={configurations}
+              value={query.configuration}
+              placeholder="All configurations"
+              onChange={(v) => update({ configuration: v?.value })}
+            />
+          </InlineField>
+        )}
+      </InlineFieldRow>
```
pepicrft, line 162: `status`, `category`, and `tag` are wired through the TS types and the Go client but not exposed here, so they're only reachable via dashboard JSON or template variables. Either render selectors for them or drop them from the model. Half-plumbed is worse than either.

server/test/tuist_web/controllers/api/metrics_controller_test.exs
```diff
@@ -0,0 +1,230 @@
...
+          p50_values: [],
+          p90_values: [],
+          p99_values: []
+        }
+      end)
+
+      url = ~p"/api/projects/#{project.account.name}/#{project.name}/builds/metrics/duration?from=#{@from}&to=#{@to}"
+
+      assert conn |> Authentication.put_current_user(user) |> get(url) |> json_response(200)
+      assert conn |> Authentication.put_current_user(user) |> get(url) |> json_response(200)
```
pepicrft: The test name says 'caches identical duration queries so ClickHouse is hit once' but only asserts the analytics function runs once. One extra `assert body1 == body2` would make the test actually verify what the name promises.

grafana-datasource/pkg/plugin/client.go
```diff
@@ -0,0 +1,134 @@
...
+	if err := c.get(ctx, path, nil, &out); err != nil {
+		return nil, err
+	}
+	return out.Values, nil
+}
+
+func firstNonEmpty(values ...string) string {
+	for _, v := range values {
+		if strings.TrimSpace(v) != "" {
+			return v
```
pepicrft: `firstNonEmpty` returns the untrimmed `v`, so `'  https://tuist.dev  '` survives. Then `TrimRight(baseURL, '/')` at line 35 strips nothing (trailing char is space) and the URL parse fails with a cryptic error. Trim before returning.

grafana-datasource/src/components/ConfigEditor.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+export function ConfigEditor(props: Props) {
+  const { onOptionsChange, options } = props;
+  const { jsonData, secureJsonFields, secureJsonData } = options;
+
+  const onUrlChange = (event: ChangeEvent<HTMLInputElement>) => {
+    onOptionsChange({
+      ...options,
+      jsonData: { ...jsonData, url: event.target.value },
+    });
```
pepicrft, line 18: `client.go` falls back to `settings.URL`, but this editor only writes `jsonData.url`. The fallback is unreachable. Either drop it from `firstNonEmpty` or move to writing the standard URL field.

pepicrft (commented):
Did a deep pass over the server endpoints, the Grafana plugin (Go + TS), rate limiting, and tests. The shape is good and the test coverage hits the right beats. Inline comments flag 9 substantive items worth a look before merge.

If you only want to do a few, the top three are:

1. Bucket-align `from`/`to` before hashing in `cached_duration_metrics` (otherwise the cache only helps concurrent panels, not sequential refreshes).
2. Cache the dimension endpoints (CheckHealth + QueryEditor mount fire enough of these to matter).
3. Fix `firstNonEmpty` to actually trim its return, and parallelize CheckHealth's two probes.

Nothing here is a blocker; all are quality-of-life. No security concerns: SQL interpolation is safely whitelist-sourced (`INTERVAL` strings, percentile floats), and filter values go through Ecto bindings or typed ClickHouse params throughout.

pepicrft (approved):
Approving. The 9 inline findings in my prior review are quality-of-life, not blockers.

server/lib/tuist_web/controllers/api/metrics_controller.ex
fortmarek: Good catch — fixed in f7c63e6d8c. The key now floors `from`/`to` to the cache TTL grid (60s) before hashing, so sequential refreshes share a key within the TTL, not only concurrent ones. The query still runs against the real window; only the key is gridded.

server/lib/tuist_web/controllers/api/metrics_controller.ex
fortmarek: Done in f7c63e6d8c — scheme/configuration lists now get a 10-min cache (`cached_dimension_values`), so the CheckHealth probes and editor refetches stop hitting ClickHouse on every cycle.

server/lib/tuist_web/controllers/api/metrics_controller.ex
fortmarek: Fixed in f7c63e6d8c — `validate_range` now rejects negative `from`/`to` and out-of-range timestamps with a 400, using the non-raising `DateTime.from_unix/1` instead of the `!` variant.

grafana-datasource/pkg/plugin/datasource.go
fortmarek: Done in 7f04d306f1 — the builds and tests probes now run in parallel via a `WaitGroup`, so the worst case is one client timeout instead of two.

server/lib/tuist_web/plugs/metrics_rate_limit_plug.ex
fortmarek: Added in f7c63e6d8c — the 429 now sets `Retry-After` from the deny scale (`{:deny, scale_ms}` → seconds). Test asserts it.

grafana-datasource/src/components/QueryEditor.tsx
fortmarek: Addressed in 7f04d306f1 — went with exposing rather than dropping: `status` (Success/Failure) and `category` (Clean/Incremental) now have build-only selectors in the query editor and are interpolated in `applyTemplateVariables`. Note `tag` was never actually wired into the plugin (only the server endpoint accepts it), so there was nothing half-plumbed there.

server/test/tuist_web/controllers/api/metrics_controller_test.exs
fortmarek: Fixed in f7c63e6d8c — the test now asserts `body1 == body2` on top of the single analytics call, so it verifies the cache actually serves the same payload.

grafana-datasource/pkg/plugin/client.go
fortmarek: Fixed in 7f04d306f1 — `firstNonEmpty` trims before returning, so a padded URL no longer slips past `TrimRight`.

grafana-datasource/src/components/ConfigEditor.tsx
fortmarek: Fixed in 7f04d306f1 — dropped the unreachable `settings.URL` fallback; the client now resolves `jsonData.url` (or the default), matching what the config editor writes.

PR #11212, opened 2026-06-10
feat(server): reason-gate operator access to customer projects

fortmarek:
image
image

## Summary

Replaces blanket operator access to customer projects with a per-access, reason-gated, time-boxed grant. An operator who isn't a member of a customer account is redirected to `ops.tuist.dev` to justify access, then redirected back with a signed grant the customer `server/` verifies offline. **Read** access is self-serve (reason only); **admin** access ("sign in as admins") goes through the **same Slack JIT approval as kubectl writes**. A valid grant also **bypasses the customer's SSO enforcement**, which is what makes SSO-enforced orgs reachable.

Built on #10988, which has since merged into `main`, so this targets `main` directly.

## Why

Today operator access is a static blanket: `ops_access` returns true whenever `user.account.name in ops_user_handles()`, granting read of any project's dashboard, runs, bundles, billing, etc. There is no per-access justification, no time bound, and no audit of why an operator looked at a customer's data. Separately, `require_sso_authentication` redirects operators into the customer's own SSO, which they can't complete, so SSO-enforced orgs are inaccessible today.

This makes operator access a deliberate, reason-logged, time-boxed, auditable act, and reuses #10988's JIT machinery (Tailscale role as source of truth, Slack approval, fail-closed) for the admin tier.

## How it works

```
Operator -> tuist.dev/{account}/{project}        (not a member; org may enforce SSO)
  server: no grant -> redirect to ops.tuist.dev/grants/new?return_to=...&account=...
  ops (Pomerium / Google OIDC; X-Pomerium-Claim-Email):
     read  -> record grant, mint Ed25519 JWT, 302 back with ?operator_grant=<jwt>
     admin -> Slack approval card; second human approves; then 302 back with the jwt
  server: verify jwt offline (EdDSA-strict, iss/aud pinned, TTL capped),
          pin account_id, store in session, strip the param,
          skip SSO enforcement, authorize via the grant
```

### tuist-ops (the grant issuer)
- `lib/tuist_ops/project_access/` new feature module: `request.ex` + `grant.ex` (schemas), `approvals.ex` (read = self-serve grant inline; admin = pending until Slack approval), `policy.ex` (read self-serve / admin-approver gates, reusing `JIT.TailscaleClient`), `token.ex` (mints the Ed25519 grant), `slack_blocks.ex`.
- `grant_controller.ex` + router `/grants/*`: the reason form, identity from `X-Pomerium-Claim-Email`. The admin "pending approval" page is a plain HTML poll of `/grants/:id/status`, not LiveView, since tuist-ops has no asset pipeline.
- `slack_controller.ex`: `pa_approve` / `pa_deny` interactive actions.
- Migration `create_project_access_tables`, `jose` dep, new env config.

### server (offline verification + enforcement)
- `lib/tuist_web/operator_grant.ex`: `verify/1` (`JOSE.JWT.verify_strict(_, ["EdDSA"], _)` only, rejecting `none`/`HS256` confusion tokens; `exp`, max-TTL ceiling, `iss`/`aud` pinning), the `accept_operator_grant` handoff plug, `load_operator_grant` plug + on_mount, and the `redirect_to_ops_if_operator` gate.
- Authorization: split the shared `ops_access` check. The internal `/ops` panel keeps its static behavior via the new `internal_ops_access`; customer-data objects move to a grant-scoped `ops_access` / `ops_write_access` (admin tier required for writes), resolving each object to its account via `object_account_id/1`.
- SSO bypass in `require_sso_authentication` when the operator holds a valid grant for that account.
- `ops_user_handles` removed. Operator eligibility is now `Tuist.Accounts.tuist_operator?/1` (confirmed `@tuist.dev` email), and the redirect gate additionally requires a Google-authenticated session (the "google sso identity under the @tuist.dev org" signal), so password and test sessions stay on the normal path.

### infra / docs
- `PROJECT_ACCESS_SIGNING_KEY` added to the tuist-ops ESO.
- `infra/k8s/operator-project-access-audit.md`: the three-trail audit story plus the deployment runbook.

## Reasoning behind the non-obvious choices

- **Signed grant, not a server->ops call.** The server verifies offline with only the public key, so operator access and the customer-facing app stay on separate failure domains (the same decoupling #10988 is built on). Asymmetric, so even a compromised server can't mint grants. Revocation is short TTL + keypair rotation.
- **Google-auth gate on the redirect.** The eligibility heuristic is a routing decision, not the security boundary (the boundaries are Pomerium/Google-OIDC at ops and the offline verification). Requiring `auth_method == :google` is the faithful reading of "google sso identity" and keeps the many `@tuist.dev` test fixtures on the normal path.

## Security notes for the reviewer

- `X-Pomerium-Claim-Email` is the operator identity for `/grants/*` and is spoofable on a raw ingress. Those routes **must** be Pomerium-fronted (which #10988 deferred); they are deliberately **not** added to the unprotected public ingress here. Called out in the runbook.
- The `?operator_grant=` token is stripped via redirect before any page renders or any observability plug logs the query string.

## Deployment (manual, in the runbook)

1. Generate an Ed25519 keypair. Private PEM -> 1P `TUIST_OPS_BOT.project_access_signing_key`; public PEM -> server `TUIST_OPERATOR_GRANT_PUBLIC_KEY`.
2. Stand up Pomerium in front of `ops.tuist.dev/grants/*` (Google OIDC), carving out `/webhooks/slack/*`.

## Validation

- tuist-ops: 131 tests pass (new suites for `policy`, `approvals`, `token`); `mix format` + `mix credo` clean; compiles with `--warnings-as-errors`.
- server: full suite 4923/4924 pass; `mix format` + `mix credo` clean; edited tests compile with `--warnings-as-errors`. The single failing test (`TuistWeb.TestsLiveTest`, a `render_async` 100ms ClickHouse-latency flake) was confirmed pre-existing on the base commit by stashing all changes and re-running.
- New server tests cover `verify/1` (valid / `alg:none` / `HS256` / expired / over-TTL / wrong aud / wrong iss), the grant-based checks across every object shape, the redirect gate (operator vs non-operator vs grant-holder vs unauthenticated), the token handoff, and a full account-settings integration test exercising grant -> plug -> on_mount -> SSO bypass -> `ops_write_access`.


server/lib/tuist/authorization/checks.ex
```diff
@@ -194,22 +194,87 @@ defmodule Tuist.Authorization.Checks do
...
-    false
+  def internal_ops_access(_, _), do: false
+
+  @doc """
+  Operator READ access to a customer's data. Granted only by an
+  active, unexpired operator grant (minted at ops.tuist.dev, verified
+  offline, attached to `user.operator_grant`) that covers the object's
+  account. Fail-closed: no grant, wrong account, wrong tier, expired,
+  or unknown object shape → false.
+  """
```
pepicrft, line 215: **[CONFIRMED] API/CLI operator access regresses to fail-closed.**

`ops_access`/`ops_write_access` now require `user.operator_grant`, but the grant virtual field is populated only by `TuistWeb.OperatorGrant.load_operator_grant`, which is plugged into the `/:account_handle` and `/:account_handle/:project_handle` browser scopes (`router.ex` ~970, ~1017). Non-browser pipelines (`:authenticated_api`, `:mcp`, `:scim_api`) never load it, so `operator_grant` stays `nil` and `ops_access` falls into the catch-all clause at the bottom — denied.

Previously these paths matched `user.account.name in ops_user_handles()`. Concrete impact: an operator running the CLI against a customer account they aren't a member of (e.g. `tuist cache --project acme/ios-app` to repro a bug) now 403s with no signal that the auth model changed.

If the design intent is "operators must go through the browser grant flow for CLI too," that's fine — but it deserves an explicit deny message pointing them at ops.tuist.dev, not a silent 403 indistinguishable from a real permission denial.

tuist-ops/lib/tuist_ops_web/controllers/grant_controller.ex
```diff
@@ -0,0 +1,239 @@
...
+      true ->
+        grant_or_request(conn, subject, account, return_to, tier, reason, ttl_seconds)
+    end
+  end
+
+  def pending(conn, %{"id" => id}) do
+    subject = subject(conn)
+
+    case Approvals.get_request(id) do
+      %{requester_email: ^subject} = req when not is_nil(subject) ->
```
pepicrft: **[CONFIRMED] 403 vs 404 leaks request-ID existence to unauthenticated callers.**

```elixir
case Approvals.get_request(id) do
  %{requester_email: ^subject} = req when not is_nil(subject) -> render_pending(...)
  nil -> render_error(conn, 404, ...)
  _   -> render_error(conn, 403, ...)
end
```

The `:browser` pipeline in `router.ex:20` doesn't enforce verified identity (no `require_operator_identity` plug — `audit_controller.ex` got one in 01a93b657e for exactly this reason). When `subject == nil` (a tailnet-direct caller hitting tuist-ops without going through Pomerium), the first clause's `when not is_nil(subject)` guard fails and the wildcard `_` clause fires for any existing request → 403. Missing request → 404. Sequential int IDs (`get_request/1` parses `Integer.parse`) → enumeration.

The 403 page also surfaces the message that the request belongs to someone else, leaking `requester_email`/`account_handle`/`return_to` indirectly. Fix: gate the whole scope on the same plug `audit_controller` uses.

*(Also see comment on `pending/2`'s case-sensitive email match.)*

server/lib/tuist/environment.ex
```diff
@@ -307,16 +307,58 @@ defmodule Tuist.Environment do
...
+  def operator_grant_public_key(secrets \\ secrets()) do
+    get([:operator_grant, :public_key], secrets)
+  end
+
+  @doc """
+  The `aud` claim required on operator grant tokens. Pinned per
+  environment so a token minted for a different env can't be replayed.
+  Must match ops.tuist.dev's `OPERATOR_GRANT_AUDIENCE`.
+  """
+  def operator_grant_audience(secrets \\ secrets()) do
```
pepicrft, line 335: **[CONFIRMED] Self-host docs still reference the removed `TUIST_OPS_USER_HANDLES`, no docs for the new var.**

`grep TUIST_OPERATOR_EMAIL_DOMAIN server/priv/docs/` returns zero hits, while `server/priv/docs/en/guides/server/self-host/control-plane.md:153` and `:663` still document `TUIST_OPS_USER_HANDLES` as the way to grant ops access.

A self-hoster upgrading from `main`:
- Their existing `TUIST_OPS_USER_HANDLES=admin` is silently ignored.
- The new `tuist_operator?/1` defaults to `ends_with?(email, "@tuist.dev")`, which is always false for `admin@theircorp.com`.
- They lose `/ops/` access entirely with no migration path, and the SSO bypass branch (which also gates on `tuist_operator?`) becomes unreachable.

Needs either a doc update + a migration note, or the new var defaulted from an explicit "self-host" mode.

server/lib/tuist/accounts.ex
```diff
@@ -1269,6 +1269,30 @@ defmodule Tuist.Accounts do
```
pepicrft, line 1269: **[PLAUSIBLE] `get_account_by_handle/1` is case-insensitive (ilike); token handle can resolve to a different-cased account row.**

(Anchoring here for diff visibility — the bug is in the existing function around line 103, which uses `ilike(a.name, ^handle)`.)

In `operator_grant.ex` the grant claims map is stored under `claims.account_handle` (the verbatim token string), and the account is resolved separately via `Accounts.get_account_by_handle/1`. If two same-letter different-case accounts ever coexist (or one is created later), ilike returns an arbitrary row without ORDER BY — so the grant gets keyed by `"Acme"` while the resolved `account_id` belongs to a different organisation's `"acme"` row. Authorization check then passes against the wrong customer.

Safer: normalise the handle (force lowercase in both mint and lookup), or compare exact + case-insensitive index. Either way, don't trust an ilike to return a unique account.

server/lib/tuist_web/operator_grant.ex
```diff
@@ -0,0 +1,407 @@
...
+
+  defp do_accept(conn, token) do
+    with {:ok, claims} <- verify(token),
+         :ok <- check_subject_is_current_operator(conn, claims),
+         %{id: account_id} <- Accounts.get_account_by_handle(claims.account_handle) do
+      grants =
+        conn
+        |> get_session(@session_key)
+        |> normalize_grants()
+        |> Map.put(claims.account_handle, Map.put(claims, :account_id, account_id))
```
pepicrft: **[PLAUSIBLE] Session grants keyed by token's `claims.account_handle`, looked up by raw URL param — case drift loses the grant.**

Here the session map gets `Map.put(grants, claims.account_handle, ...)` using the verbatim string from the JWT. In `active_session_grant/2` (around L318) it's `Map.get(grants, account_handle)` against `conn.params["account_handle"]` — exact-match.

Scenario: operator typed `acme` in the reason form on ops.tuist.dev (lowercased); the customer link they were sent to is `/Acme/dashboard` (handle preserves case). Session stores `%{"acme" => claims}`; loader does `Map.get(grants, "Acme")` → `nil` → no grant attached → `redirect_to_ops_if_operator` bounces them back to ops to mint *another* grant. Each mint accumulates a different-cased key, so the loop persists.

Downcase both the storage key and the lookup key (and the params handle the loader uses).

server/lib/tuist_web/operator_grant.ex
```diff
@@ -0,0 +1,407 @@
...
+  True when the current user holds a valid (loaded, unexpired) operator
+  grant for the account on this request. Used by the SSO-enforcement
+  bypass — so it carries the same operator-identity binding as the
+  authorization checks: the session user must be a confirmed Tuist
+  operator and the one the grant was minted for.
+  """
+  def active_grant?(conn) do
+    case conn.assigns[:current_user] do
+      %User{operator_grant: %{exp: exp} = grant, email: email} = user when is_integer(exp) ->
+        exp > System.system_time(:second) and
```
pepicrft, line 310: **[PLAUSIBLE] `active_grant?/1` doesn't re-check `grant.account_handle` against URL params; SSO bypass trusts the loader's keying.**

The predicate validates `exp`, `tuist_operator?(user)`, and `email == sub`, but never compares `grant.account_handle` to `conn.params["account_handle"]`. The SSO-bypass call site (`authentication.ex` ~335) calls `active_grant?(conn)` as the only check — it trusts that `load_operator_grant` keyed by the correct URL handle.

That's fine today because the loader pulls the entry from a handle-keyed map. But the moment someone refactors `load_operator_grant` (e.g. "fall back to the first valid grant when the per-account lookup misses", or a LiveView lifecycle that carries `current_user.operator_grant` across `live_action` navigations between accounts), the SSO bypass becomes a silent cross-account bypass.

Defence-in-depth fix: have `active_grant?/1` take `account_handle` and assert `grant.account_handle == account_handle` itself. Same shape as the `sub == email` check that's already there.

tuist-ops/lib/tuist_ops_web/controllers/grant_controller.ex
pepicrft: **[PLAUSIBLE] Pending-page `requester_email: ^subject` strict-equality with no downcase; case drift 403s the legitimate operator.**

`subject(conn)` returns `TuistOps.Pomerium.verified_email(conn)`, which forwards the JWT `email` claim verbatim (no `String.downcase`). If the email is ever returned with different casing across requests (Google's response normalisation, an SSO provider that preserves the user's typed case, manual admin row in mixed case), the strict pin against `req.requester_email` fails → wildcard `_` clause → 403 "belongs to another operator".

Narrow fix: `String.downcase` both `req.requester_email` at insert time and `subject` here before comparing. Pair with a citext column / lowercase index.

*(See also the broader 403/404 enumeration comment on this same area.)*

tuist-ops/lib/tuist_ops/project_access/approvals.ex
```diff
@@ -0,0 +1,288 @@
...
+           attrs
+           |> request_attrs(ttl, approval_expires_at)
+           |> Map.put(:slack_channel_id, channel_id)
+           |> Request.create_changeset()
+           |> Repo.insert(),
+         {:ok, ts} <-
+           SlackClient.post_message(channel_id, SlackBlocks.pending(request),
+             fallback_text: @fallback_text
+           ),
+         {:ok, updated} <-
```
pepicrft, line 99: **[PLAUSIBLE] Slack post succeeds, `slack_message_ts` persist fails → orphaned approve card whose button still mints a grant.**

The `with` chain inserts the Request row (status `"pending"`), `SlackClient.post_message`s the approval card, then `Repo.update`s `slack_message_ts`. The Slack post is HTTP (can't be in a DB transaction) and the post + persist aren't wrapped in any compensating logic.

Failure mode: transient DB hiccup between Slack 200 and the ts persist. The `with` returns `{:error, ...}` to the controller (which surfaces a 500), but the request row is already persisted as `pending` and a live approve/deny card exists in the channel encoding the request id. An approver clicks Approve, `Approvals.approve/2` matches `status: "pending"` and `Policy.admin_approver_allowed?` — it does NOT require `slack_message_ts`. Grant minted, token signed, operator gets admin access, original card never updates (`notify_active` quietly no-ops on `nil` ts).

Fixes that work here: (a) make the Slack post-update reconcile in a job loop that backfills `slack_message_ts` and retries the original `Repo.update`; (b) gate `approve/2` on `slack_message_ts != nil`; (c) if ts persist fails, delete the request row before returning the error.

server/lib/tuist_web/operator_grant.ex
```diff
@@ -0,0 +1,407 @@
...
+      conn
+      |> put_session(@session_key, grants)
+      |> redirect(to: stripped_path(conn))
+      |> halt()
+    else
+      _ ->
+        # Invalid/unknown token, or a valid token presented by someone
+        # who is not the operator it was minted for: strip it and
+        # continue unauthenticated. The grant is never stored.
+        conn |> redirect(to: stripped_path(conn)) |> halt()
```
pepicrft, line 209: **[PLAUSIBLE] Grant-back URL strips the token before login, costing a wasted mint + double audit row.**

`accept_operator_grant` runs in the `:browser_app` pipeline before `require_authenticated_user`. When `current_user` is `nil`, `check_subject_is_current_operator` returns `{:error, :no_current_user}`, the plug strips `?operator_grant=` from the URL and 302s to the cleaned path (this clause around L209). The user then hits the login flow without the grant in tow.

After Google login they land on the stripped URL, re-trip `redirect_to_ops_if_operator`, and ops mints a *second* grant. Not infinite (the second pass succeeds), but:
- The just-approved grant was the one the operator's Slack approver signed off on — discarding it desynchronises the audit row from the actually-used grant.
- Two `jti`/audit entries per access click, with the first one orphaned.
- If grant minting has any side effects we want exactly-once (counters, rate-limits, expiry windows that start at mint), they're firing twice.

Safer: stash the grant in the session and re-attempt acceptance after login, instead of stripping the URL when there's no current user.

infra/helm/pomerium/templates/configmap.yaml
```diff
@@ -79,3 +79,44 @@ data:
...
+      # tuist-ops verifies (it does NOT trust the bare claim header), so identity
+      # is cryptographic rather than resting on Pomerium stripping a spoofed copy.
+      # Gated on `opsRoute.enabled`: off in the chart default (staging/canary),
+      # ON in values-production. Enabling it is the whole switch — the chart also
+      # adds {{ .Values.opsRoute.host }} to the Pomerium Ingress (TLS + rule) and
+      # the tuist-ops chart drops its own Ingress for the host, so cert-manager
+      # and external-dns converge the cutover on deploy. The only out-of-band
+      # step is the `signing_key` 1P field (already stored). The Slack webhook
+      # stays public (Slack signs its own requests and cannot do OIDC);
+      # /api/v1/policy is deliberately NOT a public route — it stays tailnet-only.
```
pepicrft, line 96: **[CONFIRMED] No Pomerium route for `ops.tuist.dev/` — bare domain returns Pomerium 404.**

The routes here cover `/grants`, `/audit`, `/webhooks/slack`. The tuist-ops router has `get "/", AuditController, :root` (`tuist_ops_web/router.ex:39`). An operator typing `https://ops.tuist.dev/` (the natural entry point you'd send people in Slack DMs or paste from memory) gets a Pomerium 404 instead of the audit dashboard.

Either add a fourth route with `prefix: "/"` last (Pomerium does longest-prefix-first so `/grants` and `/audit` still take priority), or drop the bare `"/"` from the ops router.

pepicrft (approved):
Heuristic review (high-effort, 7 finder angles → single-vote verify). Surfacing 10 findings inline; approving on @pepicrft's call. Worth a pass before merge:

**Confirmed**
- `ops_access` now requires `user.operator_grant`, which is only populated on the `:account`/`:project` browser pipelines — API/CLI/MCP paths silently lose operator fallback access.
- `/grants/:id/pending` differentiates 403 vs 404 with no `require_operator_identity` plug; a tailnet-direct caller (bypassing Pomerium) can enumerate request IDs.
- Self-host docs still document the removed `TUIST_OPS_USER_HANDLES` and don't mention `TUIST_OPERATOR_EMAIL_DOMAIN`.
- Pomerium routes for `ops.tuist.dev` don't cover `/` — bare-domain hit returns Pomerium 404.

**Plausible**
- `Accounts.get_account_by_handle/1` uses `ilike`, so a token-claimed handle could resolve to a different-cased account row.
- Session grants map keyed by token's `claims.account_handle` but looked up by raw URL `params["account_handle"]` — case drift = silent grant miss → redirect loop.
- `active_grant?/1` doesn't re-check `grant.account_handle` against URL params; SSO bypass trusts the loader keying with no defence-in-depth.
- Pending-page `requester_email` strict-equality vs Pomerium claim with no `downcase` — case drift 403s the legitimate operator.
- Slack post + `slack_message_ts` persist isn't atomic; transient DB error between them → orphaned approve card whose Approve button still mints a grant.
- `accept_operator_grant` runs before `require_authenticated_user`: unauthenticated grant-back URL strips the token before login, costing a second mint after auth.

Refuted in verify (not posting): per-render token re-mint (`exp` is bound to grant row), hardcoded `signing_salt` (real secret is `secret_key_base`), SameSite/Refresh loop (identity comes from Pomerium header, not session), anonymous-traffic DB queries (`google_authenticated?` short-circuits first), JOSE PEM parse per render (only fires on one-time redirect-back).

— Claude (high-effort review)

tuist-ops/lib/tuist_ops_web/controllers/grant_controller.ex
fortmarek: Fixed in 15fc5ea937: added a `require_operator_identity` plug to `GrantController`, so every action 401s without a verified Pomerium identity before the 403/404 distinction — the same plug `audit_controller` got in 01a93b657e. New test asserts the 401.

server/lib/tuist/environment.ex
fortmarek: Fixed in 15fc5ea937: replaced `TUIST_OPS_USER_HANDLES` with `TUIST_OPERATOR_EMAIL_DOMAIN` in the English `control-plane.md` (the env-var table at :153 and the `/ops` authorization note at :663). A self-hoster sets it to their own domain (e.g. `theircorp.com`).

server/lib/tuist_web/operator_grant.ex
fortmarek: Fixed in 15fc5ea937: session grants are now keyed by the case-folded handle (`grant_key/1`) on both store (`do_accept`) and lookup (`active_session_grant`), so a case difference between the stored claim and the URL param can't lose the grant.

server/lib/tuist_web/operator_grant.ex
fortmarek: Fixed in 15fc5ea937: `active_grant?/2` now takes the request's `account_handle` and asserts `grant.account_handle` matches it (case-insensitive), instead of trusting the loader's keying — same shape as the existing `sub == email` check.

tuist-ops/lib/tuist_ops_web/controllers/grant_controller.ex
fortmarek: Fixed in 15fc5ea937: `pending/2` now compares `requester_email` to the subject case-insensitively (the self-approval check already did), so case drift no longer 403s the legitimate operator.

tuist-ops/lib/tuist_ops/project_access/approvals.ex
fortmarek: Fixed in 15fc5ea937 via option (b): `approve/2` now refuses to mint for a request whose `slack_message_ts` is nil (half-created), so an orphaned approve card can't sign a grant. The requester re-requests instead.

infra/helm/pomerium/templates/configmap.yaml
fortmarek: Fixed in 15fc5ea937: added an OIDC-gated `/` catch-all Pomerium route. Pomerium matches longest-prefix-first, so `/grants`, `/audit`, and `/webhooks/slack` still take priority; `/` covers the bare domain (which tuist-ops redirects to `/audit`) and the static assets.

server/lib/tuist/authorization/checks.ex
fortmarek: Intended behaviour, not a silent permission regression to hide: operators reach customer data only through the reason-gated browser grant, and `ops_access` only ever matched `%User{}` — never `AuthenticatedAccount` tokens — so token-auth CLI/MCP/SCIM never had operator fallback to begin with. The case that does change is a `%User{}` session hitting a non-browser pipeline. Agreed the bare 403 is unfriendly, but a per-operator "go to ops.tuist.dev" hint means threading operator-awareness through the generic API authz / FallbackController, which couples a generic layer to the operator concept — so I'd rather do that deliberately as a follow-up than bolt it on here. Leaving this thread open for that call.

server/lib/tuist/accounts.ex
fortmarek: Mitigated rather than silently-wrong: `get_account_by_handle/1` resolves with `Repo.one`, which raises on multiple matches, so if `Acme` and `acme` ever coexisted the grant accept would fail closed (500) — it can't bind the wrong account_id. `accounts.name` is case-sensitive-unique app-wide and the function is shared with runner dispatch, so the real fix is case-insensitive-unique account names (a separate, app-wide change), not something to special-case in the grant path. The session key is now case-folded (see the P2 thread). Leaving open for the app-wide name-uniqueness decision.

server/lib/tuist_web/operator_grant.ex
fortmarek: Real edge, and worse than a wasted mint for admin tier (it discards the approved grant). The clean fix is stashing the verified claims server-side and re-applying after login — but doing that without reintroducing the token-in-URL leak on the public-project path (where `require_authenticated_user` doesn't fire) needs care, so I'm flagging it as a focused follow-up rather than shipping a half-fix in this batch. Leaving open.

PR #11213, opened 2026-06-10
fix(server): re-arm flaky-test alerts when the condition clears, regardless of recovery

fortmarek:

## What this fixes

Automation alerts fire their trigger actions only on a **transition** into the triggered set:

```elixir
newly_triggered = triggered_ids \ already_triggered   # already_triggered = latest event is "triggered"
```

A test case leaves `already_triggered` only when a `recovered` event is appended — and today that append is gated entirely behind `recovery_enabled`. So a **recovery-disabled** alert (for example the default per-project "Flaky test detection") latches in `triggered` the first time a test case matches and never acts on it again, even after:

- the condition clears and later returns, or
- a *different* alert undoes the side effect (e.g. "Auto-mute flaky tests" recovery clearing `is_flaky`).

This surfaced on a real test case that kept flaking but was never re-marked: the alert responsible for marking it flaky was permanently latched, so it silently stopped acting.

## What changed

Re-arming (the bookkeeping that lets the next rising edge fire) is now **decoupled** from the opt-in undo actions:

- `handle_recovery` always runs. When a triggered test case clears its condition and the recovery dwell elapses, the alert appends a `recovered` event so it can fire again on the next rising edge.
- Undo side effects stay **opt-in**: only `recovery_enabled` alerts pass their `recovery_actions` (unmute, remove label). A recovery-disabled alert re-arms but leaves its effect in place until a human clears it — a legitimate, intentional config, not a bug.

### Why the recovery scope was reworked (avoiding a ClickHouse regression)

The recovery candidate set previously used the monitor's `all` field. For recovery-disabled alerts that field was `[]` (cheap); populating it would have meant a per-tick `SELECT DISTINCT id FROM test_cases` for **every project's default flaky alert**, because `AutomationScheduler` enqueues full evaluations for `last_days` alerts every cadence. Given the canary pool / alert-contention history, that's not acceptable.

Instead, the recovery scope is now derived from the worker's `scoped_test_case_ids`:

- **Scoped tick** → only re-arm test cases in the evaluated chunk.
- **Full tick** → every active event was measured, so all are fair game.

The `all` field was a no-op filter anyway (active events are always real test cases), so it and `load_all_test_case_ids` are removed entirely. Net result: **cheaper than before**, including for recovery-enabled alerts, with no new per-tick query.

## Notes for reviewers

- Re-arming still respects the recovery **dwell** (hysteresis) to avoid flapping re-fires on notification alerts. For recovery-disabled alerts with no `recovery_config` it falls back to the existing 14-day default — moot in practice for the affected alerts (Flaky-detection's condition only clears after 30d clean; Auto-disable's re-fire is an idempotent no-op).
- This fixes the **latching** class of bug (alerts that silently stop acting). It does not, by itself, re-mark a test whose condition never went false — that case is two alerts contending over the same `is_flaky` field, which is a separate single-writer / desired-state change.

### How to test locally

\`\`\`bash
cd server
mix test test/tuist/automations/workers/alert_evaluation_worker_test.exs \
         test/tuist/automations/monitors/flaky_tests_monitor_test.exs
\`\`\`

Validation run in this branch:

- `alert_evaluation_worker_test.exs` — 23 pass (incl. two new tests: re-arms-but-withholds-actions when recovery is disabled, and respects-the-dwell)
- `flaky_tests_monitor_test.exs` — 18 pass
- Full `test/tuist/automations/` — 113 pass
- `mix format --check-formatted` clean, `mix credo` no issues

## Behavior change / migration note

Before this change, `recovery_enabled: false` effectively meant "fire the trigger actions once, then stay latched forever." With re-arming decoupled from recovery, a recovery-disabled alert now re-arms **as soon as its condition no longer holds** (no dwell — recovery means exactly that). Practical impact:

- For idempotent trigger actions (mute, mark-flaky, disable) the re-fire is a no-op.
- For a **`send_slack`** trigger on a flaky condition, this introduces a re-ping on the natural flap cycle (each time the test goes clean and then flaky again). Anyone relying on the old fire-once behavior can move to `recovery_enabled: true` with empty `recovery_actions` to keep the alert latched until they act.

### DIFF
--- a/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
+++ b/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
@@ -15,13 +15,6 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
 
   require Logger
 
-  # Re-arm dwell applied when recovery is disabled. There is no per-alert knob
-  # for this case (recovery_config is only validated when recovery is on), so a
-  # disabled alert always falls back to this fixed window. Kept explicit rather
-  # than relying on filter_recovered_candidates' default so the value is
-  # discoverable; revisit if a customer needs "never re-arm" or a faster one.
-  @disabled_recovery_rearm_window "14d"
-
   @impl Oban.Worker
   def perform(%Oban.Job{args: %{"alert_id" => alert_id} = args}) do
     case Automations.get_alert(alert_id) do
@@ -163,21 +156,19 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
       |> reject_unevaluated_this_tick(scoped_test_case_ids)
 
     # Re-arming (appending the "recovered" event so the next rising edge can
-    # fire again) is unconditional — an alert that never re-arms is latched in
-    # `triggered` forever and silently stops acting. The dwell window and the
-    # undo side effects, however, stay opt-in: `Alert.changeset` only validates
-    # `recovery_config`/`recovery_actions` when recovery is enabled, so for a
-    # disabled alert we ignore the persisted (possibly stale) config — falling
-    # back to the default re-arm dwell — and run no undo actions. The test case
-    # re-arms but keeps its effect until a human clears it.
-    {recovery_config, recovery_actions} =
+    # fire again) happens for every alert once its condition clears — without
+    # it, an alert latches in `triggered` forever and silently stops acting.
+    # When recovery is enabled the user's dwell and undo actions apply; when
+    # it's disabled we re-arm the moment the condition clears (no dwell, no
+    # undo) and leave any effect in place until a human clears it. The
+    # persisted recovery_config is intentionally ignored on the disabled path
+    # because `Alert.changeset` only validates it when recovery is on.
+    {recovered, recovery_actions} =
       if alert.recovery_enabled do
-        {alert.recovery_config || %{}, alert.recovery_actions}
+        {filter_recovered_candidates(alert, candidates, alert.recovery_config || %{}), alert.recovery_actions}
       else
-        {%{"window" => @disabled_recovery_rearm_window}, []}
+        {candidates, []}
       end
-
-    recovered = filter_recovered_candidates(alert, candidates, recovery_config)
 
     Enum.each(recovered, fn event ->
       entity = %{type: :test_case, id: event.test_case_id}
--- a/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
+++ b/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
@@ -131,8 +131,18 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
       end
     end)
 
-    handle_recovery(alert, triggered_ids, active_events, scoped_test_case_ids)
+    # Event-driven monitors (`test_updated`) have no triggered/recovered ledger
+    # — each event is a discrete one-shot (see
+    # `Automations.dispatch_test_case_event/2`). They never accrue `triggered`
+    # events through this worker, so recovery bookkeeping must not run for them,
+    # even if a monitor-type change left stale events behind.
+    if !event_driven_monitor?(alert) do
+      handle_recovery(alert, triggered_ids, active_events, scoped_test_case_ids)
+    end
   end
+
+  defp event_driven_monitor?(%{monitor_type: "test_updated"}), do: true
+  defp event_driven_monitor?(_alert), do: false
 
   defp active_alert_events(alert, nil), do: Automations.list_active_alert_events(alert.id)
 
@@ -141,22 +151,28 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
 
   defp handle_recovery(alert, currently_triggered_ids, active_events, scoped_test_case_ids) do
     currently_triggered_set = MapSet.new(currently_triggered_ids)
-    recovery_config = alert.recovery_config || %{}
 
     candidates =
       active_events
       |> Enum.reject(&MapSet.member?(currently_triggered_set, &1.test_case_id))
       |> reject_unevaluated_this_tick(scoped_test_case_ids)
 
-    recovered = filter_recovered_candidates(alert, candidates, recovery_config)
-
     # Re-arming (appending the "recovered" event so the next rising edge can
     # fire again) is unconditional — an alert that never re-arms is latched in
-    # `triggered` forever and silently stops acting. The undo side effects are
-    # opt-in: only `recovery_enabled` alerts reverse their trigger actions
-    # (unmute, remove label). A recovery-disabled alert still re-arms but
-    # leaves its effect in place until a human clears it.
-    recovery_actions = if alert.recovery_enabled, do: alert.recovery_actions, else: []
+    # `triggered` forever and silently stops acting. The dwell window and the
+    # undo side effects, however, stay opt-in: `Alert.changeset` only validates
+    # `recovery_config`/`recovery_actions` when recovery is enabled, so for a
+    # disabled alert we ignore the persisted (possibly stale) config — falling
+    # back to the default re-arm dwell — and run no undo actions. The test case
+    # re-arms but keeps its effect until a human clears it.
+    {recovery_config, recovery_actions} =
+      if alert.recovery_enabled do
+        {alert.recovery_config || %{}, alert.recovery_actions}
+      else
+        {%{}, []}
+      end
+
+    recovered = filter_recovered_candidates(alert, candidates, recovery_config)
 
     Enum.each(recovered, fn event ->
       entity = %{type: :test_case, id: event.test_case_id}
--- a/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
+++ b/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
@@ -15,6 +15,13 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
 
   require Logger
 
+  # Re-arm dwell applied when recovery is disabled. There is no per-alert knob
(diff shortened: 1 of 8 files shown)


server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -131,24 +131,46 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
...
     end)
 
-    if alert.recovery_enabled do
-      handle_recovery(alert, triggered_ids, active_events, all_ids)
+    # Event-driven monitors (`test_updated`) have no triggered/recovered ledger
+    # — each event is a discrete one-shot (see
+    # `Automations.dispatch_test_case_event/2`). They never accrue `triggered`
+    # events through this worker, so recovery bookkeeping must not run for them,
+    # even if a monitor-type change left stale events behind.
+    if !event_driven_monitor?(alert) do
```
pepicrft: worth a heads-up in the PR description: customers who set `recovery_enabled: false` to mean "fire once, stay latched" will now see trigger_actions re-fire after the 14d re-arm. for alerts with a `send_slack` trigger on a flaky condition, that's a new periodic ping. anyone on that pattern can move to `recovery_enabled: true` with empty recovery_actions, but the migration note seems worth calling out.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -289,12 +323,12 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
   # (see `Tuist.Automations.dispatch_test_case_event/2`), so the scheduled
   # evaluator has nothing to do for them.
   defp evaluate_monitor(%{monitor_type: "test_updated"}, _test_case_ids) do
-    %{triggered: [], all: []}
+    %{triggered: []}
   end
 
   defp evaluate_monitor(alert, _test_case_ids) do
```
pepicrft, line 325: small regression here. before this change, the catch-all returned `all: []` so candidates in handle_recovery ended up empty. now an alert with an unknown/legacy monitor_type still falls through to recovery with the full active_events set. cold path (changeset rejects bad types at write) but a real behavior change. could short-circuit at the top of `evaluate_and_execute` alongside the event-driven check.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -131,24 +131,46 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
...
-  defp handle_recovery(alert, currently_triggered_ids, active_events, all_ids) do
+  defp active_alert_events(alert, scoped_test_case_ids),
+    do: Automations.list_active_alert_events(alert.id, scoped_test_case_ids)
+
+  defp handle_recovery(alert, currently_triggered_ids, active_events, scoped_test_case_ids) do
     currently_triggered_set = MapSet.new(currently_triggered_ids)
-    all_ids_set = MapSet.new(all_ids)
-    recovery_config = alert.recovery_config || %{}
 
     candidates =
```
pepicrft, line 153: the old `load_all_test_case_ids` was a soft guard against active events for test_case_ids that no longer exist in clickhouse. with `recovery_enabled: true` and a last_days dwell, an orphan event now reaches execute_actions; if the action errors on the missing row, no recovered event is written and the worker re-tries the same failure every tick. cold path because test_cases is a ReplacingMergeTree without deletes, so probably fine in practice, but flagging in case data-surgery tooling ever changes that.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -131,24 +131,46 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
...
+
+    # Re-arming (appending the "recovered" event so the next rising edge can
+    # fire again) is unconditional — an alert that never re-arms is latched in
+    # `triggered` forever and silently stops acting. The dwell window and the
+    # undo side effects, however, stay opt-in: `Alert.changeset` only validates
+    # `recovery_config`/`recovery_actions` when recovery is enabled, so for a
+    # disabled alert we ignore the persisted (possibly stale) config — falling
+    # back to the default re-arm dwell — and run no undo actions. The test case
+    # re-arms but keeps its effect until a human clears it.
+    {recovery_config, recovery_actions} =
```
pepicrft: the 14d fallback dwell here is unreachable by the user once `recovery_enabled: false` is set (no UI knob, no per-alert override, no env var). a customer who wants "never re-arm" or "re-arm faster" has no path. could either document the implicit 14d in the product help or expose a `re_arm_dwell` field independent of recovery_enabled.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -131,24 +131,46 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
...
+    # — each event is a discrete one-shot (see
+    # `Automations.dispatch_test_case_event/2`). They never accrue `triggered`
+    # events through this worker, so recovery bookkeeping must not run for them,
+    # even if a monitor-type change left stale events behind.
+    if !event_driven_monitor?(alert) do
+      handle_recovery(alert, triggered_ids, active_events, scoped_test_case_ids)
     end
   end
 
+  defp event_driven_monitor?(%{monitor_type: "test_updated"}), do: true
```
pepicrft: third copy of this predicate. `AutomationScheduler.scheduled_alert?` (workers/automation_scheduler.ex:44) and `ProjectAutomationsLive.event_driven_monitor_type?` (project_automations_live.ex:552) both hardcode `"test_updated"`. if a fourth event-driven type ever lands, drift across the three is easy to miss. moving to `Alert.event_driven?/1` would centralize it.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -131,24 +131,46 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
       end
     end)
 
-    if alert.recovery_enabled do
-      handle_recovery(alert, triggered_ids, active_events, all_ids)
+    # Event-driven monitors (`test_updated`) have no triggered/recovered ledger
```
pepicrft: the comment isn't quite right. test_updated alerts do accrue triggered events in `automation_alert_events`, just via `dispatch_test_case_event -> create_alert_event` (automations.ex:371) rather than through this worker. the guard itself is correct; the wording could trip up a future maintainer writing a backfill or UI query on the assumption that the ledger is empty for event-driven alerts. could reword to something like "event-driven monitors write triggered events from dispatch_test_case_event, not this worker, so recovery dwell semantics don't apply".

pepicrft (approved):
lgtm overall. left a few notes on edge cases and a couple of cleanup items worth considering.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
fortmarek: Good call — added a **Behavior change / migration note** to the PR description spelling out that recovery_enabled:false alerts now re-arm after the dwell, and that send_slack-on-flaky users can switch to recovery_enabled:true with empty recovery_actions to keep the old fire-once behavior.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
fortmarek: Fixed. Recovery is now gated positively on `Alert.recovery_ledger?/1` (only the metric monitors), so an unknown/legacy monitor_type no longer falls through to recovery with the full active_events set — it skips it, same as test_updated. Added Alert predicate tests covering the unknown case.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
fortmarek: Leaving as-is for now: `test_cases` is a ReplacingMergeTree with no deletes, so orphan events don't arise in practice, and a failing recovery action is logged once per tick (not retried within a job), so it can't hot-loop. Noted the assumption here in case deletion tooling ever lands and changes it.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
fortmarek: Made the 14d explicit as a named constant (`@disabled_recovery_rearm_window`) with a comment, and documented the behavior in the PR description. Holding off on a user-facing `re_arm_dwell` field until a customer actually needs 'never re-arm' / 're-arm faster', to avoid a speculative knob.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
fortmarek: Centralized as `Alert.event_driven?/1` (plus `Alert.recovery_ledger?/1`), derived from the monitor-type taxonomy on the schema. `AutomationScheduler.scheduled_alert?` and `ProjectAutomationsLive.event_driven_monitor_type?` now delegate to it, so there's a single source of truth.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
fortmarek: Reworded — the comment now notes event-driven monitors *do* write triggered events, just via `dispatch_test_case_event/2` rather than this worker, so a future maintainer won't assume the ledger is empty for them.

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
fortmarek: Update: went with your suggestion and removed the 14d dwell entirely. Recovery-disabled alerts now re-arm the instant the condition clears (no dwell, no per-alert knob to reason about), so the hidden window this thread was about is gone. Updated the PR description's behavior note accordingly.

PR #11216, opened 2026-06-10
fix(cli): resolve suite shard enumeration destination

fortmarek:
## What changed

- Normalizes shard-plan destinations before suite-level test enumeration.
- Resolves platform-only simulator destinations such as `platform=iOS Simulator` into concrete simulator destinations with an `id`.
- Preserves already-concrete destinations that include `id` or `name`.
- Adds regression coverage for suite shard planning with a platform-only iOS simulator destination.

## Why

Suite granularity calls `xcodebuild test-without-building -enumerate-tests` after the build-for-testing phase. The build run succeeded, but enumeration failed because Tuist passed only `platform=iOS Simulator`, and Xcode could not match that broad destination from `.xctestproducts`.

## Root Cause

The normal build path resolves a concrete simulator destination before invoking Xcode, but shard planning discarded that shape and sent only a platform-level simulator destination into `XCTestEnumerator`. Xcode accepts enough information to build the test products, but `test-without-building -enumerate-tests -testProductsPath` needs a concrete simulator destination to enumerate suites reliably.

## Impact

Suite-level sharding now enumerates tests using the same kind of concrete simulator destination that the build/test path already resolves, avoiding error 70 for platform-only simulator destinations.

## Validation

- Ran `swiftformat` on the two touched files.
- Built the committed iOS fixture:
  - `xcodebuild -project examples/xcode/xcode_project_with_tests/App.xcodeproj -scheme App build-for-testing -destination "platform=iOS Simulator,id=99A1ADBF-B5DD-402F-B8DB-BDEBA7220E04" -derivedDataPath /private/tmp/tuist-suite-shard-e2e.ogeIBI/DerivedData -testProductsPath /private/tmp/tuist-suite-shard-e2e.ogeIBI/TestProducts.xctestproducts CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=`
- Reproduced the reported failure:
  - `xcodebuild test-without-building -enumerate-tests -testProductsPath /private/tmp/tuist-suite-shard-e2e.ogeIBI/TestProducts.xctestproducts -destination "platform=iOS Simulator"`
  - Result: exit 70 with `Unable to find a device matching the provided destination specifier: { platform:iOS Simulator }`.
- Verified the fixed destination shape:
  - Booted simulator `99A1ADBF-B5DD-402F-B8DB-BDEBA7220E04`.
  - `xcodebuild test-without-building -enumerate-tests -testProductsPath /private/tmp/tuist-suite-shard-e2e.ogeIBI/TestProducts.xctestproducts -destination "platform=iOS Simulator,id=99A1ADBF-B5DD-402F-B8DB-BDEBA7220E04"`
  - Result: exit 0 and full suite enumeration:
    - `AppFrameworkTests/AppFrameworkTests/example()`
    - `AppTests/AppTests/example()`
    - `AppTests/AppTests/parameterized(value:)`

### DIFF
--- a/cli/Sources/TuistKit/Services/TestService.swift
+++ b/cli/Sources/TuistKit/Services/TestService.swift
@@ -2045,7 +2045,7 @@ public struct TestService { // swiftlint:disable:this type_body_length
         case let .device(udid):
             return "\(platform.xcodebuildPlatformDestination),id=\(udid)"
         case .mac:
-            return try await simulatorController.macOSDestination()
+            return try await simulatorController.macOSDestination(catalyst: false)
         case .macCatalyst:
             return try await simulatorController.macOSDestination(catalyst: true)
         }
--- a/cli/Sources/TuistKit/Services/TestService.swift
+++ b/cli/Sources/TuistKit/Services/TestService.swift
@@ -1944,7 +1944,7 @@ public struct TestService { // swiftlint:disable:this type_body_length
             }
 
             let destinationVersion = xcodebuildDestinationParameter("OS", in: explicitDestination)?.version() ?? version
-            return try await concreteShardPlanDestination(
+            return await concreteShardPlanDestinationIfAvailable(
                 schemes: schemes,
                 platform: simulatorPlatform,
                 version: destinationVersion,
@@ -1955,40 +1955,50 @@ public struct TestService { // swiftlint:disable:this type_body_length
 
         if let platform {
             let buildPlatform = try XcodeGraph.Platform.from(commandLineValue: platform)
-            return try await concreteShardPlanDestination(
+            return await concreteShardPlanDestinationIfAvailable(
                 schemes: schemes,
                 platform: buildPlatform,
                 version: version,
                 deviceName: deviceName,
                 graphTraverser: graphTraverser
-            )
+            ) ?? buildPlatform.xcodebuildPlatformDestination
         }
 
-        for scheme in schemes {
-            guard let target = buildGraphInspector.testableTarget(
-                scheme: scheme,
-                testPlan: nil,
-                testTargets: [],
-                skipTestTargets: [],
-                graphTraverser: graphTraverser,
-                action: .build
-            ) else { continue }
+        if let inferredDestination = inferPlatformDestination(schemes: schemes, graphTraverser: graphTraverser) {
+            guard let inferredPlatform = xcodebuildPlatform(from: inferredDestination) else {
+                return inferredDestination
+            }
 
-            guard let resolvedPlatform = target.target.destinations.first?.platform,
-                  target.target.destinations.platforms.count == 1
-            else { continue }
+            return await concreteShardPlanDestinationIfAvailable(
+                schemes: schemes,
+                platform: inferredPlatform,
+                version: version,
+                deviceName: deviceName,
+                graphTraverser: graphTraverser
+            ) ?? inferredDestination
+        }
 
-            return try await xcodebuildDestination(
-                for: target,
-                scheme: scheme,
-                platform: resolvedPlatform,
+        return nil
+    }
+
+    private func concreteShardPlanDestinationIfAvailable(
+        schemes: [Scheme],
+        platform: XcodeGraph.Platform,
+        version: Version?,
+        deviceName: String?,
+        graphTraverser: GraphTraversing
+    ) async -> String? {
+        do {
+            return try await concreteShardPlanDestination(
+                schemes: schemes,
+                platform: platform,
                 version: version,
                 deviceName: deviceName,
                 graphTraverser: graphTraverser
             )
+        } catch {
+            return nil
         }
-
-        return nil
     }
 
     private func concreteShardPlanDestination(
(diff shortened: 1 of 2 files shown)


cli/Sources/TuistKit/Services/TestService.swift
```diff
@@ -1922,6 +1927,166 @@ public struct TestService { // swiftlint:disable:this type_body_length
...
+            }
+
+            let destinationVersion = xcodebuildDestinationParameter("OS", in: explicitDestination)?.version() ?? version
+            return try await concreteShardPlanDestination(
+                schemes: schemes,
+                platform: simulatorPlatform,
+                version: destinationVersion,
+                deviceName: deviceName,
+                graphTraverser: graphTraverser
+            ) ?? explicitDestination
```
pepicrft, line 1953: `??` only catches `nil`, not throws. If `findAvailableDevice` throws (no matching sim, pretty common in CI without the right runtime installed) the user's explicit `-destination` is lost and the whole shard-build step fails. Before this PR, the explicit destination was passed to the planner verbatim. Either wrap in `do/catch` and return `explicitDestination` from the catch, or make the inner returns non-throwing optionals.

Same shape on the `if let platform` branch above (line 1956 onward): there isn't even a fallback there, so a missing sim turns `--platform iOS` into a hard failure where it used to produce `platform=iOS Simulator`.

cli/Sources/TuistKit/Services/TestService.swift
```diff
@@ -1922,6 +1927,166 @@ public struct TestService { // swiftlint:disable:this type_body_length
         }
         return nil
     }
```
pepicrft, line 1929: `inferPlatformDestination` is now only used by the test at `TestServiceTests.swift:5530`. The loop body at line 1967 reimplements it. Either delete the function (and the test) or have `shardPlanDestination` delegate to it so the scheme-loop logic isn't sitting in three places.

cli/Sources/TuistKit/Services/TestService.swift
```diff
@@ -1922,6 +1927,166 @@ public struct TestService { // swiftlint:disable:this type_body_length
...
+    private func xcodebuildDestinationParameter(_ parameter: String, in destination: String) -> String? {
+        let expectedKey = parameter.lowercased()
+        for component in destination.components(separatedBy: ",") {
+            let pair = component.split(separator: "=", maxSplits: 1).map {
+                String($0).trimmingCharacters(in: .whitespaces)
+            }
+            guard pair.count == 2 else { continue }
+
+            let key = pair[0].lowercased()
+            if key == expectedKey || key.hasSuffix("/\(expectedKey)") {
```
pepicrft: Curious what `hasSuffix("/\(expectedKey)")` is guarding against. Xcodebuild destination keys aren't namespaced with slashes that I'm aware of, and no test hits this branch. If there's a real case worth covering, a comment + test would help; otherwise I'd drop it.

cli/Tests/TuistKitTests/Services/TestServiceTests.swift
```diff
@@ -4894,6 +4894,100 @@ final class TestServiceTests: TuistUnitTestCase {
             .called(1)
     }
 
+    func test_run_build_resolvesPlatformOnlySimulatorDestinationForSuiteShardPlanning() async throws {
```
pepicrft, line 4969: Coverage thought: the new resolver has four observable branches and only the platform-only sim one is exercised. Cases worth adding:

- explicit `-destination platform=iOS Simulator,id=...` passes through verbatim (skips `findAvailableDevice`)
- `platform=macOS` passes through verbatim
- `platform=iOS Simulator,OS=18.0` drives `findAvailableDevice` with version 18.0 (the only real way to confirm the OS parser works end to end)
- no `--platform` and no `-destination` hits the bottom scheme-loop branch at TestService.swift:1967

Also noting per `.blick/skills/tuist-swift-review/SKILL.md` rule 3 the new test should ideally be in Swift Testing, but the whole file is XCTest so I'm fine deferring the migration to a sweep rather than blocking on it.

pepicrft (approved):
Approving. Logic looks right and the bug was real (`platform=iOS` is not a valid xcodebuild destination). A few non-blocking things inline that I think are worth a follow-up before this gets too settled.

cli/Sources/TuistKit/Services/TestService.swift
fortmarek: Done in 52d2a71c1f8. I dropped the hasSuffix("/...") allowance and now require exact destination-key matches; there was no real namespaced-key case to preserve.

cli/Tests/TuistKitTests/Services/TestServiceTests.swift
fortmarek: Done in 52d2a71c1f8. I added coverage for concrete simulator pass-through, macOS pass-through, OS-version parsing into findAvailableDevice, inferred scheme-loop destinations, and the simulator/macOS fallback cases. I kept these in XCTest for consistency with the existing TestServiceTests file.

cli/Sources/TuistKit/Services/TestService.swift
fortmarek: Done in 52d2a71c1f8. I moved the concrete destination lookup behind a non-throwing helper, so explicit destinations and platform-derived destinations fall back to the previous value when simulator or macOS destination resolution fails. I also added tests for both fallback paths.

PR #11239, opened 2026-06-11
fix(cli): quote `-fmodule-map-file` paths to account for whitespace

waltflanagan:
The new generated module map files would break the `-fmodule-map-file` build setting if the target name contained spaces.  This led to a path with a space in it and the command line build invocation would split on that newline causing the build to fail. 

### How to test locally

Includes an update to `generated_app_with_spm_dependencies` to exhibit the bug in 4.198.2 and validate the fix correctly allows it to build.

### DIFF
--- a/cli/Tests/TuistDependenciesAcceptanceTests/DependenciesAcceptanceTests.swift
+++ b/cli/Tests/TuistDependenciesAcceptanceTests/DependenciesAcceptanceTests.swift
@@ -21,6 +21,7 @@ import TuistTesting
 @Suite(.serialized)
 struct DependenciesAcceptanceTests {
     @Test(
+        .disabled(),
         .inTemporaryDirectory,
         .withMockedEnvironment(inheritingVariables: ["PATH"]),
         .withMockedNoora,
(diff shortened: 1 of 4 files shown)


cli/Tests/TuistDependenciesAcceptanceTests/DependenciesAcceptanceTests.swift
```diff
@@ -21,7 +21,6 @@ import TuistTesting
 @Suite(.serialized)
 struct DependenciesAcceptanceTests {
     @Test(
-        .disabled(),
```
fortmarek, line 24: these tests take a long time. Let's make a new dedicated acceptance tests that would just build on macOS. We try not to boot iOS simulators and run tests on them since then we can't parallelize these efficiently.

fortmarek (approved):
The fix looks good! Before merging, I'd just create a new, more lightweight fixture to run the acceptance test again

cli/Tests/TuistDependenciesAcceptanceTests/DependenciesAcceptanceTests.swift
waltflanagan: Is the key thing not running tests and just building (even an iOS app)? or should i find a much smaller dependency than firebase that triggers the module map generation?

waltflanagan: I'll re-disable these and follow up with another PR with better tests so that this can go ahead and merge and be released since it's a big regression.

pepicrft (approved):

PR #11246, opened 2026-06-11
fix(kura): harden segment disk accounting at rotation and startup

esnunes:
Two disk-safety hardening fixes for the CAS segment ring, surfaced while stress-testing the margin assumptions behind #11222 (the disk-derived capacity work). Both matter more as that PR ships: larger rings mean far more rotations, so more chances to crash inside the rotation window and more traffic through the free-space guard.

## What changed

1. **The rotation free-space check now scales with the incoming artifact.** Previously it demanded a fixed `MAX_SEGMENT_BYTES * SEGMENT_FREE_SPACE_MARGIN` (1 GiB) no matter what was about to be written. The required bytes are now `max(MAX_SEGMENT_BYTES, incoming_size) * SEGMENT_FREE_SPACE_MARGIN`. Behavior is bit-for-bit identical for artifacts ≤ 512 MiB.
2. **Startup sweep for orphaned segment files.** `Store::sweep_orphaned_segments` runs once at startup (after `Store::open`, before any traffic): it lists `data_dir/segments/`, and every `.seg` file not referenced by the persisted ring state is routed through the existing `evict_segment` path. A failed sweep logs a warning and does not block startup.

## Root causes

**Oversized artifacts under-provisioned the guard.** Rotation triggers when `current_size + incoming > MAX_SEGMENT_BYTES`, and then the *whole* incoming artifact is appended into the fresh segment — segment files can legitimately exceed 512 MiB (e.g. a 2 GiB replication body, `MAX_REPLICATION_BODY_BYTES`). The rotation check is the **only** free-space checkpoint (appends between rotations are unchecked), so a 2 GiB append could pass a 1 GiB guard and hit raw ENOSPC mid-append — made worse by the fact that during `append_to_segment` the artifact's bytes transiently exist twice when the tmp dir shares the filesystem (the staged source is deleted only after the copy completes). The fix keeps the same 2× proportionality the margin has always encoded: one width for the new segment's guaranteed growth, one width of slack for unchecked co-writers.

**Crash-window orphans had no reclamation path.** Rotation is deliberately ordered for crash consistency: persist ring state *without* the evicted segment → append → commit metadata → unlink the evicted file. The safe direction of that ordering means a crash (or an error return) between the state save and the unlink strands the segment file — **and** the manifests and index entries of every artifact inside it — with nothing referencing them and no code path that will ever clean them up. Each such event leaks up to a segment's worth of disk permanently. The repo had no startup sweep of the segments directory at all.

## Why this shape

- The sweep reuses `evict_segment` rather than just unlinking files: that path already (atomically, with the tested batch logic) deletes the stranded manifests, the namespace and segment index entries, and the cached FD handle — so the crash-window case cleans up *metadata-consistently*, never leaving manifests pointing at deleted bytes.
- It is safe by construction: it runs under the `DataDirLock` writer lock (acquired before `Store::open` in `app.rs`), before any rotation can run, so it cannot race the legitimate "state references a segment whose file does not exist yet" window of a fresh rotation. The inverse direction (state entry without a file) is untouched, as is everything outside `segments/` (blob files, RocksDB).
- The required-bytes computation is extracted as a pure function so the policy is unit-testable.

## Rollout safety

Node-local only: no on-disk format, wire format, or replication protocol changes. Segment files are still reclaimed exclusively by unlink (the mmap SIGBUS invariant in `kura/AGENTS.md` is upheld). Mixed-version meshes are unaffected. For existing deployments the sweep is a no-op unless a node has accumulated orphans, in which case it frees disk on the next restart.

## Validation

- New unit tests:
  - `segment_rotation_requires_margin_for_oversized_artifacts` — fixed floor for small/equal sizes, scaled requirement for oversized ones.
  - `sweep_orphaned_segments_returns_zero_without_segments_dir` — first-boot no-op.
  - `sweep_orphaned_segments_removes_stray_files_and_keeps_live_segments` — stray file deleted, live segment and its artifact untouched and still readable.
  - `sweep_orphaned_segments_reclaims_crash_window_segment_and_metadata` — simulates the exact crash window (ring state saved without the segment, file still on disk) and asserts both the file and the stranded manifest are reclaimed.
- `cargo test --lib` — 238 passed, 0 failed; `cargo clippy --all-targets -- -D warnings` — clean; `cargo fmt` — no changes.

### DIFF
--- a/kura/src/app.rs
+++ b/kura/src/app.rs
@@ -122,6 +122,11 @@ async fn run_with_config(
         config.memory_hard_limit_bytes,
     );
     let store = Store::open(&config, io.clone(), memory.clone())?;
+    match store.sweep_orphaned_segments().await {
+        Ok(0) => {}
+        Ok(swept) => tracing::info!(swept, "removed orphaned segment files"),
+        Err(error) => tracing::warn!("failed to sweep orphaned segments: {error}"),
+    }
     let peer_client_factory = crate::peer_tls::PeerClientFactory::from_config(&config).await?;
     let client = peer_client_factory.build()?;
     let runtime = RuntimeState::new();
--- a/kura/src/store.rs
+++ b/kura/src/store.rs
@@ -1061,7 +1061,7 @@ impl Store {
         };
 
         if needs_new_segment {
-            let required_bytes = MAX_SEGMENT_BYTES.saturating_mul(SEGMENT_FREE_SPACE_MARGIN);
+            let required_bytes = segment_rotation_required_bytes(incoming_size);
             if let Some(available) = available_disk_bytes(&self.data_dir)
                 && available < required_bytes
             {
@@ -1184,6 +1184,58 @@ impl Store {
         }
 
         Ok(())
+    }
+
+    /// Removes segment files that the segment ring state no longer
+    /// references, along with any metadata still pointing at them.
+    ///
+    /// Rotation persists the ring state without the evicted segment before
+    /// the file is unlinked, so a crash (or an error) in that window strands
+    /// the file — and the manifests of the artifacts inside it — with no code
+    /// path left to reclaim them. Must run at startup, under the data-dir
+    /// writer lock and before any traffic, so it cannot race a rotation
+    /// creating a segment whose state entry is not yet visible.
+    pub async fn sweep_orphaned_segments(&self) -> Result<usize, String> {
+        let segments_dir = self.data_dir.join("segments");
+        let mut entries = match tokio::fs::read_dir(&segments_dir).await {
+            Ok(entries) => entries,
+            Err(error) if error.kind() == std::io::ErrorKind::NotFound => return Ok(0),
+            Err(error) => {
+                return Err(format!(
+                    "failed to list segments directory {}: {error}",
+                    segments_dir.display()
+                ));
+            }
+        };
+
+        let state = self.load_segment_state()?;
+        let mut swept = 0;
+        loop {
+            let entry = entries.next_entry().await.map_err(|error| {
+                format!(
+                    "failed to read segments directory {}: {error}",
+                    segments_dir.display()
+                )
+            })?;
+            let Some(entry) = entry else {
+                break;
+            };
+            let file_name = entry.file_name();
+            let Some(segment_id) = file_name
+                .to_str()
+                .and_then(|name| name.strip_suffix(".seg"))
+            else {
+                continue;
+            };
+            if state.generation_of(segment_id).is_some() {
+                continue;
+            }
+            tracing::warn!(segment_id, "removing orphaned segment");
+            self.evict_segment(segment_id).await?;
+            swept += 1;
+        }
+
+        Ok(swept)
     }
 
     fn segment_path(&self, segment_id: &str) -> PathBuf {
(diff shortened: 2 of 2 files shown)


fortmarek (approved):

pepicrft (approved):
Approving. Both fixes do exactly what's needed and the design is sound.

**Free-space check** — `segment_rotation_required_bytes(incoming_size) = max(MAX_SEGMENT_BYTES, incoming_size) * SEGMENT_FREE_SPACE_MARGIN` preserves bit-for-bit the current behavior for ≤ 512 MiB artifacts and scales correctly above that. The doc comment about the 2x margin (new segment growth + slack for unchecked co-writers like RocksDB flushes, the not-yet-unlinked evicted segment, and tmp staging coexistence) is accurate. The tmp coexistence point in particular is real: `append_to_segment` removes the source after `tokio::io::copy` finishes, so on a shared filesystem both copies are resident during the append.

**Sweep ordering** — verified the writer-lock claim. `DataDirLock::acquire` runs at `kura/src/app.rs:102`, `Store::open` at `:124`, `sweep_orphaned_segments` at `:125`, and the listener wiring comes later. So the sweep does run under the writer lock with no traffic in flight, which makes the "state may temporarily reference a segment whose file doesn't exist yet during a rotation" inverse case unreachable here.

**Reusing `evict_segment`** is the right call. The crash-window orphan strands not just the file but also the manifests and segment-index entries, and `evict_segment` cleans all three atomically through the existing batch. Idempotent if rerun on the same id (segment_artifacts prefix is empty, `saw_entries=false` skips the write, file removal is best-effort, `remove_segment` returns false).

**Test coverage** is on point — the crash-window test sets up the exact `state.remove_segment` + file-still-present condition and asserts the manifest is reclaimed. Nothing I'd add.

Minor things, not blocking:
- The sweep uses `tokio::fs::read_dir` directly while the rest of the store goes through `IoController` for metrics. Startup one-off so acceptable, but slightly inconsistent.
- No test for non-`.seg` files being preserved by the filter. The `strip_suffix(".seg")` short-circuit is obvious enough that it's fine.
- The symmetric "state references a segment whose file is gone" case is correctly left out of scope per the PR description.

Nice follow-up to #11222.

pepicrft:
One follow-up thought after sitting with this: it's worth being clear that this PR makes the existing leaky design *correct*, but doesn't restore the bounded-segment invariant the capacity model in #11222 leans on. `MAX_SEGMENT_BYTES` is still only a rotation trigger, not a per-write cap, so segments can legitimately be up to `MAX_MODULE_TOTAL_BYTES` / `MAX_REPLICATION_BODY_BYTES`. The ring is byte-unbounded and only count-bounded, and the new free-space formula is essentially a tax on that.

The *sweep* half I'd keep regardless of direction. Crash-window recovery is a separate concern (the rotation ordering here is already the safer of the two options, and a startup sweep is the right way to handle the inevitable leak). What's worth revisiting is the free-space scaling and, behind it, the segment-size assumption.

Three ways to actually bound segments, in rough order of surgery:

1. **Raise `MAX_SEGMENT_BYTES` to match `MAX_MODULE_TOTAL_BYTES`** (2 GiB). Rotation trigger == per-write cap == segment size. Simplest, no new code paths. Cost: coarser rotation, more tail-wasted bytes per segment, and the legacy 5-segment floor becomes 10 GiB minimum disk (today it's 2.5 GiB). Probably fine for the deployment shape but worth checking against the smallest tier.
2. **Reject `size > MAX_SEGMENT_BYTES` at the API boundary** and lower the advertised REAPI cap accordingly. Smallest patch surface, but a behavioral break for module uploads over 512 MiB.
3. **Dual pool: segment ring + standalone blobs with their own budget.** `ArtifactManifest.blob_path` and every read path already handle the blob branch, so the read side is free. The work is on the write side (route `size > MAX_SEGMENT_BYTES` to a blob) plus a blob retention policy (LRU, size budget tied to the same disk-percent ceiling as the segment ring). Without that retention the spillover just relocates the unboundedness, so it's the design choice that matters, not the wiring.

#1 is the cheapest "actually bounded" option; #3 is the most flexible. Either way, once segments are truly bounded the free-space check goes back to the original 1 GiB and `segment_rotation_required_bytes` can be deleted. Flagging for whoever picks up the followup — not blocking on this PR.

PR #11247, opened 2026-06-11
perf(kura): cache the parsed segment ring state in memory

esnunes:
Performance follow-up to #11222. That PR grows the CAS segment ring from a fixed 5 segments to a disk-derived count (~300 segments on a 200 Gi volume, up to 16,384 at the ceiling) — which turns a previously invisible inefficiency into a per-request tax that scales with ring size.

## The problem

Every serving-path generation check went through `load_segment_state`: a RocksDB point-get of the entire ring state blob plus a full `serde_json` parse (one `String` allocation per segment), followed by a linear scan in `generation_of`. Call sites:

- `prepare_artifact_for_serving` — **every serve** of a segment-backed artifact
- `maybe_refresh_manifest` — pre-check plus re-check under the refresh lock, so a request that triggers a refresh parsed the state up to **3×**
- `active_segment` — every artifact write
- `snapshot()` — the metrics loop

The state only changes on rotation (once per ~512 MiB ingested); between rotations every parse reconstructs an identical value. And there is no cross-process freshness to buy: the data-dir writer lock guarantees a single writer process, and every mutation already funnels through `save_segment_state` on the same `Store`.

Cost by ring size (~75–80 bytes of JSON per segment reference):

| Ring | State blob | Per-parse | Per-request impact |
| --- | --- | --- | --- |
| 5 segments (legacy) | ~400 B | ~1 µs | noise |
| ~300 segments (#11222 on a 200 Gi PVC) | ~24 KB | ~50–100 µs + 300 allocs | 1–3 parses per request; ~5–15% of a core at 1k req/s |
| 16,384 segments (`MAX_DESIRED_SEGMENTS`) | ~1.3 MB | 2–4 ms + 16k allocs | would dominate serve latency |

## What changed

The parsed state is now cached in memory as a `SegmentStateSnapshot` — the `SegmentState` plus a `HashMap<segment_id, SegmentGeneration>` index:

- `Store::open` seeds the snapshot with one parse at startup (`load_segment_state_from_db` is now only called there).
- `save_segment_state` keeps writing RocksDB exactly as before (durability unchanged) and then swaps the snapshot, so the cache cannot be forgotten by a mutation path — both mutators (`active_segment` rotation, `evict_segment`'s defensive removal) already funnel through it.
- `segment_generation` becomes an `Arc` clone plus a hash lookup: no RocksDB read, no parse, no linear scan.
- `active_segment` reads the snapshot and only clones the state when actually rotating, so the per-write cost drops to a pointer clone too.
- The generation index is rebuilt once per save — O(ring) once per ~512 MiB ingested instead of O(ring) per request.

`SegmentState::generation_of` (the linear scan) is removed — the snapshot index is its only remaining caller's replacement, and clippy's dead-code deny would flag it. Note for #11246: its orphan sweep calls `generation_of`; whichever PR merges second resolves the one-liner to `snapshot.generations.contains_key(...)`.

Deliberately **not** changed: the persisted format. The JSON blob shape is what a rolled-back binary reads (downgrade safety), and with a single parse per process lifetime plus a ~0.25% write amplification per rotation even at the maximum ring, there is no remaining reason to split it into per-segment keys.

## Concurrency notes

Mutations were already effectively serialized: rotation runs under `segment_write_lock`, and `evict_segment`'s save is a near-no-op in practice because `push_new` removes ring-evicted segments from the state before eviction runs. The cache swap happens inside `save_segment_state`, after the RocksDB write succeeds, so readers see either the previous or the new fully-built snapshot — never a partial one. Readers hold the mutex only long enough to clone an `Arc`.

## Validation

- New tests: `segment_generation_tracks_saved_state` (cache coherence through `save_segment_state` for all three generations + miss), `evicting_a_segment_updates_the_cached_generation` (the defensive eviction branch updates the cache), `segment_state_snapshot_survives_reopen` (a second `Store::open` on the same data dir seeds the snapshot from RocksDB).
- The pre-existing test that calls `save_segment_state` directly to fabricate an `old`-generation segment keeps passing unchanged, confirming the refresh path reads through the cache.
- `cargo test --lib` — 236 passed, 0 failed; `cargo clippy --all-targets -- -D warnings` — clean; `cargo fmt` — no changes.

### DIFF
--- a/kura/src/segment/state.rs
+++ b/kura/src/segment/state.rs
@@ -1,6 +1,6 @@
 use serde::{Deserialize, Serialize};
 
-use crate::segment::{generation::SegmentGeneration, reference::SegmentReference};
+use crate::segment::reference::SegmentReference;
 
 #[derive(Clone, Debug, Default, Serialize, Deserialize, PartialEq, Eq)]
 pub struct SegmentState {
@@ -12,31 +12,6 @@ pub struct SegmentState {
 impl SegmentState {
     pub fn active(&self) -> Option<&SegmentReference> {
         self.new.last()
-    }
-
-    pub fn generation_of(&self, segment_id: &str) -> Option<SegmentGeneration> {
-        if self
-            .old
-            .iter()
-            .any(|segment| segment.segment_id == segment_id)
-        {
-            return Some(SegmentGeneration::Old);
-        }
-        if self
-            .current
-            .iter()
-            .any(|segment| segment.segment_id == segment_id)
-        {
-            return Some(SegmentGeneration::Current);
-        }
-        if self
-            .new
-            .iter()
-            .any(|segment| segment.segment_id == segment_id)
-        {
-            return Some(SegmentGeneration::New);
-        }
-        None
     }
 
     pub fn push_new(
@@ -86,7 +61,7 @@ fn remove_from_segments(segments: &mut Vec<SegmentReference>, segment_id: &str) 
 #[cfg(test)]
 mod tests {
     use super::SegmentState;
-    use crate::segment::{generation::SegmentGeneration, reference::SegmentReference};
+    use crate::segment::reference::SegmentReference;
 
     #[test]
     fn push_new_rebalances_generations() {
@@ -113,22 +88,5 @@ mod tests {
         let evicted = state.push_new(SegmentReference::new("new-6".into(), 6), 1, 2, 2);
         assert_eq!(evicted.len(), 1);
         assert_eq!(evicted[0].segment_id, "new-1");
-    }
-
-    #[test]
-    fn generation_of_tracks_segment_location() {
-        let state = SegmentState {
-            old: vec![SegmentReference::new("old".into(), 1)],
-            current: vec![SegmentReference::new("current".into(), 2)],
-            new: vec![SegmentReference::new("new".into(), 3)],
-        };
-
-        assert_eq!(state.generation_of("old"), Some(SegmentGeneration::Old));
-        assert_eq!(
-            state.generation_of("current"),
-            Some(SegmentGeneration::Current)
-        );
-        assert_eq!(state.generation_of("new"), Some(SegmentGeneration::New));
-        assert_eq!(state.generation_of("missing"), None);
     }
 }
--- a/kura/src/store.rs
+++ b/kura/src/store.rs
@@ -73,6 +73,7 @@ pub struct Store {
     rocksdb_write_buffer_manager: WriteBufferManager,
     segment_write_lock: Mutex<()>,
     segment_refresh_lock: Mutex<()>,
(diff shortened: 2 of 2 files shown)


fortmarek (approved):

pepicrft (approved):
Approving. The design is exactly right for this access pattern: read-heavy, write-rare, single-writer process.

**Cache coherence is mechanically enforced.** I traced every mutation point in `store.rs`: `active_segment` at the rotation branch and `evict_segment`'s defensive remove_segment branch both end in `save_segment_state`, and that function is the *only* place `replace_segment_state_snapshot` is called. The snapshot is also seeded once in `Store::open` from `load_segment_state_from_db`. No other path mutates `SegmentState`, and the doc comment on `SegmentStateSnapshot` warns future maintainers about the invariant.

**Lock and Arc choices are appropriate.** `StdMutex<Arc<SegmentStateSnapshot>>` lets readers acquire the lock just long enough to clone the Arc; writers build the new snapshot outside the lock and only swap the pointer inside. No I/O ever happens under the lock, so `StdMutex` over `tokio::Mutex` is the right call. The cache update lands *after* the RocksDB `put_cf` succeeds, so a failed write leaves the cache consistent with disk.

**The right ordering within rotation.** In `active_segment`, the snapshot is cloned only on the rotation branch (`let mut state = snapshot.state.clone()`), so the common no-rotation path is now a pointer clone plus an `active()` peek. Good.

**Tests cover the load-bearing transitions.** `segment_generation_tracks_saved_state` exercises all three generations + miss through `save_segment_state`; `evicting_a_segment_updates_the_cached_generation` covers the defensive eviction branch; `segment_state_snapshot_survives_reopen` validates the startup seed by reopening on the same data dir.

**Merge-order with #11246.** The PR description already flags it: #11246's `sweep_orphaned_segments` calls the removed `state.generation_of(...)` and the renamed `load_segment_state`. Whichever merges second needs to rewrite that to `snapshot.generations.contains_key(...)`. Small fixup, just need to remember.

A couple of pre-existing things this PR doesn't touch (not blocking, just noting):

- `save_segment_state` uses a non-sync `put_cf`, so a crash between the write returning and the WAL flush could lose a rotation. Recoverable on next boot (eviction is idempotent), and the sweep from #11246 mops up. Pre-existing behavior, just preserved here.
- `evict_segment` reads the snapshot, mutates a clone, saves — so two concurrent `evict_segment` calls on disjoint live segments could lost-update. In practice the only concurrent callers are rotation (which holds `segment_write_lock`) and the startup sweep (which runs before traffic), so it isn't reachable today. Same race existed pre-cache with `load_segment_state`.

Memory footprint at the 16k ceiling is ~1–2 MB extra (the index plus the cached state) — negligible against the ~1.3 MB per-request parse cost it removes. Numbers in the PR description match the asymptotic argument.

Nice, focused change.

PR #11249, opened 2026-06-12
fix(kura): serialize segment ring state mutations

esnunes:
## Problem

`active_segment` (rotation, under `segment_write_lock`) and `evict_segment` (previously unlocked) both perform read-modify-write cycles on the segment ring state: clone the latest state, mutate the clone, save it. With no common lock, a stale clone can overwrite a concurrent save, either resurrecting an evicted segment in the ring or orphaning a freshly rotated one.

Today this race is **latent, not live**. Eviction and rotation overlap routinely on a busy node — eviction runs outside the write lock, and its RocksDB index scan plus 512 MB unlink take long enough for the next rotation to start — but during that overlap only the rotation ever writes state, because of an accidental invariant chain:

1. `push_new` removes the evicted segment from the ring before the producing rotation saves,
2. so by the time `evict_segment` runs, its conditional save (`if state.remove_segment(...)`) is always a no-op in the rotation-produced flow,
3. so every save that actually executes is a rotation save, serialized by `segment_write_lock`.

None of that is written down or enforced — review flagged this as a live P1 lost-update race precisely because a careful reader cannot tell the code is safe. And the chain breaks the moment anything evicts a segment that is still in the ring. With disk-sized rings (#11221, `kura-cas-capacity-from-disk`), pressure-eviction of ring members is the natural next caller, and at that point the interleavings become real:

- An eviction save landing inside rotation's clone-to-save window gets overwritten by the rotation's stale clone, resurrecting the evicted segment: file unlinked and manifests deleted, but the ring and capacity accounting count a phantom segment, pushing the next pressure pass to evict a live segment early — dangling REAPI action-cache references, the #11221 failure class back as an intermittent race.
- The mirror ordering orphans the freshly rotated segment from the ring: its artifacts still serve via manifests, but it never ages out (a permanent segment-sized disk leak), and `active()` reverts to the already-full previous segment, forcing an immediate spurious re-rotation.

## Fix

Every segment-state read-modify-write now funnels through a new `mutate_segment_state` helper guarded by a dedicated `segment_state_lock`: acquire the lock, copy the latest cached snapshot, apply the mutation, persist only when the state actually changed. Both rotation and eviction call it. This replaces a global, fragile correctness argument ("audit every caller and the ring semantics") with a local, enforced one ("all RMW goes through the helper; the helper holds the lock") — before the capacity work invalidates the old argument.

Design notes:

- A dedicated mutex was chosen over reusing `segment_write_lock` because that lock is held across whole segment appends; evictions are awaited on the persist hot path and would otherwise queue behind unrelated bulk writes. Lock ordering stays an acyclic hierarchy (`segment_refresh_lock` → `segment_write_lock` → `segment_state_lock`), and the state lock is a leaf with a synchronous-only critical section, so it cannot participate in a deadlock cycle.
- `active_segment`'s rotate decision still reads a snapshot taken before the state lock is acquired. That stays valid because eviction — the only other mutator — never removes the active segment (documented in a code comment). This is also a standing constraint on any future pressure-evictor: it must only target non-active generations.

## Validation

The two new multi-threaded tests simulate exactly the future caller that activates the race — they operate on ring members directly instead of rotation overflow (today's only production flow, where the race cannot fire):

- 16 concurrent `mutate_segment_state` pushes — every pushed segment survives in the final state.
- 16 concurrent `evict_segment` calls against a pre-seeded ring — every segment is gone from the final state, none resurrected by a stale save.

Both pass locally (`cargo test do_not_lose`).

Stacked on #11247; only the top commit is new here.

---

[Compound Engineering v2.60.0](https://github.com/EveryInc/compound-engineering-plugin)
🤖 Generated with Claude Fable 5 via [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kura/src/store.rs
+++ b/kura/src/store.rs
@@ -73,6 +73,7 @@ pub struct Store {
     rocksdb_write_buffer_manager: WriteBufferManager,
     segment_write_lock: Mutex<()>,
     segment_refresh_lock: Mutex<()>,
+    segment_state_lock: Mutex<()>,
     // Wrapped in `Arc` so readers clone the snapshot under a brief lock and then
     // use it without holding the mutex (unlike the sibling caches below, which
     // are read and mutated in place under their lock).
@@ -359,6 +360,7 @@ impl Store {
             rocksdb_write_buffer_manager,
             segment_write_lock: Mutex::new(()),
             segment_refresh_lock: Mutex::new(()),
+            segment_state_lock: Mutex::new(()),
             segment_state_cache: StdMutex::new(Arc::new(SegmentStateSnapshot::default())),
             segment_handles: Mutex::new(SegmentHandleCache::new(config.segment_handle_cache_size)),
             manifest_cache: StdMutex::new(ManifestCache::new(config.manifest_cache_max_bytes)),
@@ -1096,14 +1098,19 @@ impl Store {
                 ));
             }
             let segment = SegmentReference::new(Uuid::now_v7().to_string(), now_ms());
-            let mut state = snapshot.state.clone();
-            let evicted_segments = state.push_new(
-                segment.clone(),
-                self.segment_ring_limits.desired_old_segments,
-                self.segment_ring_limits.desired_current_segments,
-                self.segment_ring_limits.desired_new_segments,
-            );
-            self.save_segment_state(&state)?;
+            // The rotate decision above used a snapshot taken before the
+            // state lock; that stays valid because evictions, the only other
+            // mutator, never remove the active segment.
+            let evicted_segments = self
+                .mutate_segment_state(|state| {
+                    state.push_new(
+                        segment.clone(),
+                        self.segment_ring_limits.desired_old_segments,
+                        self.segment_ring_limits.desired_current_segments,
+                        self.segment_ring_limits.desired_new_segments,
+                    )
+                })
+                .await?;
             Ok((segment, evicted_segments))
         } else {
             Ok((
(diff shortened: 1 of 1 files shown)


fortmarek (approved):

pepicrft (approved):
Approving. This is exactly the fix for the latent race I flagged on #11247, and the design is the right shape.

**Lock hierarchy verified.** I traced the three async mutexes in `store.rs`: `segment_refresh_lock` (839) → `segment_write_lock` (999) → `segment_state_lock` (1150). The state lock is a strict leaf — the only thing held across it is the `save_segment_state` call, which only touches the `segment_state_cache` StdMutex (a strictly lower-level lock, sync, no I/O under it). No cycle possible.

**Funneling RMW through one helper is the right replacement for the fragile invariant chain.** Before this, correctness depended on the accidental property that `push_new` removes evicted segments before saving, which made the `if state.remove_segment(...)` save inside `evict_segment` a no-op in practice. The PR description correctly diagnoses that this is a code-review nightmare — sound today, but unenforced, and the moment a pressure-evictor touches ring members the interleavings turn live. Replacing a global argument with a local "all RMW goes through `mutate_segment_state`" rule is the move.

**Dedicated lock over reusing `segment_write_lock` is the right call.** `segment_write_lock` is held for the full `tokio::io::copy` of a segment append (potentially seconds for a 2 GiB body). Queueing evictions behind that would push eviction latency onto the persist hot path. The state lock's critical section is ~microseconds (clone + closure + json + put_cf + cache swap), so contention is negligible.

**The snapshot-then-lock pattern in `active_segment` is subtle but sound.** The rotate decision reads a snapshot before acquiring the state lock, so the lock guarantees the *push* operates on the latest state but the *decision* doesn't. That stays valid because evictions never remove the active segment — and the code comment at the call site plus the PR description make that a documented constraint for any future pressure-evictor. Worth noting it's documented, not enforced; a future contributor could write an evictor that violates it. Not blocking, just a note.

**`if state != snapshot.state` short-circuit.** Nice. Means a no-op `remove_segment` (id not in state) doesn't churn RocksDB or invalidate the cache. The deep PartialEq is O(ring) but in the same ballpark as one rebuild of the generations index — fine.

**On the tests.** The two concurrent-mutator tests demonstrate the API doesn't lose updates, but with such short critical sections they'll often pass even with the lock removed — they're regression tests more than race detectors. If you wanted determinism, a failpoint that holds threads inside `mutate_segment_state` between the snapshot read and the save would force the interleaving. Not blocking, the structural correctness argument is the load-bearing one.

**Bounded-design context.** This PR is orthogonal to the bounded-segment thread — neither tightens nor loosens the invariant — but it's prerequisite work for any future direction. Both the dual-pool retention path and a pressure-evictor policy that targets ring members (one of the natural next steps once segments are actually bounded) need this lock in place before they're safe to wire in. Glad to see it land before that work.

**Merge order.** Stacked on #11247 (base = `kura-segment-state-cache`), so #11247 needs to merge first. Combined with the #11246 fixup (sweep call sites) that's three PRs in a chain. Worth coordinating the merge order so each rebases cleanly.

Solid change.

PR #11253, opened 2026-06-12
feat(server, noora): update table design and improve UX

asmitbm:
A set of UX improvements to the Noora table component, applied across all dashboard tables:

* Floating header & scrollbar — on long tables, the header pins to the viewport top and a proxy scrollbar pins to the viewport bottom until the table's own edge scrolls into view. Scrollbar styling is now consistent across engines: webkit-styled native bar in Chrome/Safari, and a custom synced, draggable bar on Firefox/overlay-scrollbar platforms.

https://github.com/user-attachments/assets/ddd46f97-54aa-4078-98e4-77503fad67e8

* Sortable columns — sortable headers take a sort_order attr and render a morphing arrow indicator (spring-animated path morph between up/down). Column widths are "ratcheted" (pinned as ever-growing min-width, only once a table overflows) so sorting no longer shifts columns left/right.

https://github.com/user-attachments/assets/f9299bcb-89c5-4cb1-ae87-6fc0f42b1b61

* Expandable rows — the expand chevron crossfades/rotates, and expanded content reveals with a 200ms ease-out-cubic grid-template-rows: 0fr ⟷ 1fr transition.

https://github.com/user-attachments/assets/4910b266-bf72-4e92-9bed-076bf0168dbd

### DIFF
--- a/noora/js/Table.js
+++ b/noora/js/Table.js
@@ -112,9 +112,20 @@ export default {
       capture: true,
       passive: true,
     });
-    window.addEventListener("resize", this.onViewportChange, {
-      passive: true,
-    });
+
+    // A resize can change which ancestor scrolls — content that fit at load (so the window was
+    // the scroller) can start overflowing the content pane when the window shrinks, making the
+    // pane the scroller. Re-resolve the scroll parent on resize (kept off the scroll path, which
+    // must stay cheap, since scrolling never changes the scroller). Throttled to one frame.
+    this.onResize = () => {
+      if (this.resizeFrame) return;
+      this.resizeFrame = requestAnimationFrame(() => {
+        this.resizeFrame = null;
+        this.scrollParent = this.findScrollParent();
+        this.sync();
+      });
+    };
+    window.addEventListener("resize", this.onResize, { passive: true });
 
     // Header widths only depend on horizontal geometry, and syncing them measures every header
     // cell and writes to the body-level clone — a forced full-page reflow. Height-only resizes
@@ -150,19 +161,26 @@ export default {
     this.sync();
   },
 
-  // Nearest ancestor that scrolls vertically. The search starts above `.noora-table` so it skips
-  // the table's own horizontal scroll container (whose `overflow-y` computes to `auto` but never
-  // scrolls). Returns null when nothing between here and the document scrolls — i.e. the window
-  // is the scroller — in which case the viewport edges are used.
+  // Nearest ancestor that actually scrolls vertically. The search starts above `.noora-table` so
+  // it skips the table's own horizontal scroll container. Returns null when nothing between here
+  // and the document scrolls — i.e. the window is the scroller — in which case the viewport edges
+  // are used.
+  //
+  // Two false positives have to be filtered out, both stemming from the CSS rule that a non-
+  // `visible` value on one axis forces the other axis from `visible` to `auto`: the table's own
+  // `overflow-x: auto` container, and any ancestor with `overflow-x: hidden` (a card that only
+  // clips horizontally). Both report a computed `overflow-y: auto` while having no vertical scroll
+  // room. Requiring `scrollHeight > clientHeight` excludes them — otherwise the floating header
+  // would pin to a box that merely scrolls away with the table and would never detach.
   findScrollParent() {
     let el = this.el.parentElement;
     while (el && el !== document.body && el !== document.documentElement) {
       const overflowY = getComputedStyle(el).overflowY;
-      if (
+      const scrollable =
         overflowY === "auto" ||
         overflowY === "scroll" ||
-        overflowY === "overlay"
-      ) {
+        overflowY === "overlay";
+      if (scrollable && el.scrollHeight > el.clientHeight) {
         return el;
       }
       el = el.parentElement;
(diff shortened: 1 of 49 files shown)


noora/js/Table.js
```diff
@@ -0,0 +1,327 @@
...
+        this.thumbDrag = null;
+      };
+      this.overlayThumb.addEventListener("pointerdown", this.onThumbDown);
+      this.overlayThumb.addEventListener("pointermove", this.onThumbMove);
+      this.overlayThumb.addEventListener("pointerup", this.onThumbUp);
+    }
+
+    this.header = document.createElement("div");
+    this.header.className = "noora-table noora-table-floating-header";
+    this.header.setAttribute("aria-hidden", "true");
```
fortmarek, line 79: The floating header is hidden from the accessibility tree, but `refreshHeader()` clones the full `<thead>` later, including sortable header links/focusable controls and SVG IDs. When this fixed header is visible, those cloned controls can still receive pointer/focus interaction while assistive tech ignores them, and the document also gets duplicate IDs. Could we make the cloned header inert/non-interactive, or strip the interactive attributes/IDs when cloning, so the real header remains the only accessible control surface?

fortmarek (approved):
Looks and feels great 💯 great job!

pepicrft (commented):
Did a deeper pass with a real browser. Findings below are the ones I could reproduce.

noora/lib/noora/table.ex
```diff
@@ -145,87 +141,146 @@ defmodule Noora.Table do
...
-              <td
-                :for={{col, index} <- Enum.with_index(@col)}
-                data-selectable={
-                  !is_expandable &&
-                    (not is_nil(@row_navigate) or
-                       (not is_nil(@row_click) && not is_nil(@row_click.(row))))
-                }
+                data-expandable={is_expandable}
+                data-state={is_expandable && if(is_expanded, do: "expanded", else: "collapsed")}
               >
```
pepicrft, line 199: Tested this in a browser and three things flow from putting `phx-click` on the `<tr>`:

1. Tab never reaches the row (no `tabindex`/`role`/`aria-expanded`), so keyboard users can't expand at all.
2. Clicking a `<button>` inside an expandable row fires the button's own handler AND toggles the row, so anything actionable we add to an expandable cell later trips this.
3. Drag-selecting text in the row fires expand on mouseup, so trying to copy a cache key from an expanded row collapses what you opened.

All three go away if the disclosure trigger is a `<button aria-expanded>` inside cell 1 next to the chevron, with the row itself passive.

noora/lib/noora/table.ex
```diff
@@ -145,87 +141,146 @@ defmodule Noora.Table do
...
+                      />
                       {render_slot(col, row)}
-                    </.link>
+                    </div>
                   <% else %>
-                    {render_slot(col, row)}
+                    <%= if @row_navigate do %>
+                      <.link navigate={@row_navigate.(row)} data-part="link">
+                        {render_slot(col, row)}
+                      </.link>
```
pepicrft: `row_navigate` puts one `<a>` per cell, so a 5-column row ends up with 5 identical links. Tab order grows by N×rows and screen readers announce each cell as a duplicate link. A single row-wrapping `<a>` with `display: contents` (HTML5 transparent content allows it) would behave more like a real row link without the duplication.

noora/js/Table.js
```diff
@@ -0,0 +1,327 @@
...
+
+  refreshHeader() {
+    if (!this.thead) {
+      this.header.replaceChildren();
+      this.headerTable = null;
+      return;
+    }
+    const table = document.createElement("table");
+    table.style.tableLayout = "fixed";
+    table.appendChild(this.thead.cloneNode(true));
```
pepicrft: `cloneNode(true)` keeps the children's IDs. The sort indicators in the thead get `id={"#{@id}-sort-#{col_index}"}`, so every sortable column ends up with a duplicated ID in `document.body` once the floating header materializes. Two side effects: it violates the unique-id rule, and the cloned sort arrow won't animate because its `IconTransition` hook never re-initializes on the clone. Stripping IDs after the clone covers both:

```js
clone.querySelectorAll('[id]').forEach((el) => el.removeAttribute('id'))
```

noora/js/Table.js
```diff
@@ -0,0 +1,327 @@
...
+    const widths = Array.from(cells, (c) => c.getBoundingClientRect().width);
+    for (let i = 0; i < cells.length; i++) {
+      // Exact fractional widths, with jitter tolerance: rounding up would make the pinned sum
+      // exceed the container and manufacture an overflow on its own.
+      if (widths[i] > this.columnWidths[i] + 0.5) {
+        this.columnWidths[i] = widths[i];
+      }
+      if (this.columnWidths[i] > 0) {
+        cells[i].style.minWidth = `${this.columnWidths[i]}px`;
+      }
```
pepicrft, line 244: Verified by direct invocation that the pin only goes up: after widening one column to ~1600px and shrinking the content back to a single char, the inline `min-width` stays at ~1600px. I think it's intentional for stream-update jitter, but it surprised me on a sort/filter sequence where one batch widens a column and the next narrower one is stuck wide. A note in the leading comment saying the pin is one-way by design would help future-me.

pepicrft (approved):

PR #11254, opened 2026-06-12
feat(server): cap local kind Kura CAS capacity at 10Gi

esnunes:
## What changed

Sets `KURA_CAS_CAPACITY_BYTES` to 10Gi (`10737418240`) for the local kind dev Kura region (`local-controller`), via a new `cas_capacity_bytes` knob in the region's `provisioner_config` that the Kubernetes provisioner emits through the `KuraInstance` `extraEnv`. Also bumps the provisioner's manifest revision (`2026-06-12-local-cas-capacity-v1`) so the reconciler re-applies the updated manifest to existing instances.

## Why

#11222 made the Kura runtime derive the CAS segment-ring budget from the data volume's filesystem size (50% of capacity, capped at 80%) when `KURA_CAS_CAPACITY_BYTES` is unset. That works on managed regions, where the data dir is a real `hcloud-volumes` block device sized by the PVC. But on the local kind dev cluster the default storage class is the hostPath-backed local-path provisioner, which does **not** enforce the PVC's `10Gi` request — `statvfs` on the data dir reports the developer machine's whole disk, so the derived CAS budget would be half the host disk rather than the nominal 10Gi volume.

Capping it explicitly keeps local dev cache usage bounded to match the region's declared `storage_size`.

## Why this shape

There is no dedicated `KuraInstance` CRD field for CAS capacity, and adding one isn't warranted for a dev-only knob: the existing `extraEnv` passthrough (`extension_env/1`) is the established channel for non-secret per-region runtime configuration. Managed regions intentionally do not set the variable, so they keep the disk-derived budget — a test now pins that behavior.

The manifest revision bump is required for the change to reach already-provisioned instances: the reconciler only re-applies a manifest when the desired revision differs from the `tuist.dev/kura-manifest-revision` annotation on the live CR. For managed regions the re-apply is a no-op (their manifest is unchanged).

## Validation

- `mix test test/tuist/kura/regions_test.exs test/tuist/kura/provisioner/kubernetes_controller_test.exs` — 45 tests, 0 failures
- `mix credo` on the changed lib files — no issues
- New assertions: local-controller manifest carries `KURA_CAS_CAPACITY_BYTES=10737418240`; the managed eu-central manifest does not carry the variable
- Applied end-to-end on a live local kind cluster (`kura-dev-991`): ran `Tuist.Kura.reconcile_orphaned_deployments()` with this branch's code, confirmed the KuraInstance picked up the new revision annotation and env var, the controller propagated it into the StatefulSet, and the restarted pod runs with `KURA_CAS_CAPACITY_BYTES=10737418240` in its environment

### DIFF
--- a/server/lib/tuist/kura/provisioner/kubernetes_controller.ex
+++ b/server/lib/tuist/kura/provisioner/kubernetes_controller.ex
@@ -19,7 +19,7 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   alias Tuist.Kura.Server
 
   @namespace "kura"
-  @manifest_revision "2026-06-10-uuid-kura-introspection-v1"
+  @manifest_revision "2026-06-12-local-cas-capacity-v1"
   @manifest_revision_annotation "tuist.dev/kura-manifest-revision"
   @gateway_annotation "tuist.dev/kura-gateway"
   @gateway_controller_image "registry.k8s.io/ingress-nginx/controller:v1.11.3"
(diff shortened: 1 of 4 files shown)


esnunes:
superseded by 11879

PR #11258, opened 2026-06-12
chore(cli): bump swifterpm for subprocess-based resolution and Package.resolved stability

pepicrft:
## Summary

Updates the pinned [tuist/swifterpm](https://github.com/tuist/swifterpm) revision from `672a723` to `f75a054`, picking up three upstream commits:

- **[#35](https://github.com/tuist/swifterpm/pull/35) — Make SwifterPMCore resolve through SwiftPM subprocess.** Moves dependency resolution out of in-process linkage into a SwiftPM subprocess, which simplifies the integration surface and decouples our binary from internal SwiftPM symbols.
- **[#32](https://github.com/tuist/swifterpm/pull/32) — Fix binary artifact downloads and `Package.resolved` stability.** Addresses two long-standing issues: binary artifacts (e.g. Sparkle's xcframework) failing to download cleanly under some cache states, and unnecessary churn in `Package.resolved` between runs.
- `f75a054` — release-trigger commit so the above ship.

Only `Package.swift` (revision pin) and `Package.resolved` (lockfile refresh) change in this repo. No code changes were required — the new `SwifterPMCore` revision is API-compatible with our current callers in `TuistSupport` and `TuistLoader`.

## Why this matters

The subprocess-based resolution path (#35) is the bigger of the two. With the previous in-process linkage we paid a recurring tax every time SwiftPM's internals shifted; the subprocess boundary insulates us from that. Combined with the `Package.resolved` stability fix, this should mean fewer spurious diffs in `Package.resolved` on user machines.

## Validation

End-to-end against this repo with a locally built `tuist` linking the new revision:

- `tuist install` — clean run from an empty cache restored 82 registry packages + 3 SCM packages and cached `package-info` (~25s cold, ~2.6s warm). Restored entries in `.build/checkouts` and `.build/registry/downloads` are symlinks into `~/.cache/swifterpm/sources/...`, as expected.
- `tuist generate tuist ProjectDescription --no-open` — succeeded, exercising `SwiftPackageManagerGraphLoader` which reads the new package-info cache. Only the pre-existing duplicate-dependency warnings appear; no new diagnostics.
- `xcodebuild build -workspace Tuist.xcworkspace -scheme tuist` — `** BUILD SUCCEEDED **`.

Unit tests covering the integration:

- `TuistSupportTests/SwiftPackageManagerControllerTests` — 12/12 passing (including all `swifterpm_is_enabled` resolve/update paths).
- `TuistLoaderTests/SwiftPackageManagerGraphLoaderTests` — 10/10 passing (including `swifterPMPackageInfoCache` fresh/stale and registry/SCM mixed-dependency cases).

## Test plan

- [ ] CI green
- [ ] Spot-check a fresh `tuist install` on a clean cache after merging


Package.swift
```diff
@@ -1821,7 +1821,7 @@ let package = Package(
         .package(id: "stephencelis.SQLite_swift", from: "0.16.0"),
         .package(
             url: "https://github.com/tuist/swifterpm",
-            revision: "672a723f962cf243f1b6de65274c744c51e82acc"
+            revision: "f75a054d40c0ca0c414864e98d125f23eb22f4f5"
```
fortmarek: I believe we should be able to use actual released versions now?

fortmarek (approved):

Package.swift
pepicrft: Good catch. Switched to `exact: "0.8.9"`. The tag points to the same commit, so no resolution change.

PR #11260, opened 2026-06-12
docs(handbook): document engineer read access via Pomerium kubeconfig

fortmarek:
## What changed

Adds a self-contained **Engineer read access (Pomerium kubeconfig)** section to `infra/k8s/onboarding.md` with the ready-to-merge three-context kubeconfig (staging / canary / production), and cross-links it from the cluster-access section in `infra/AGENTS.md`.

## Why

The read path — Google Workspace identity → Pomerium gateway (`kube-<env>.tuist.dev`) → `view`-tier impersonation via the `kube-impersonator` sidecar — was already *described* in `infra/helm/pomerium/NOTES.md` and `infra/AGENTS.md`, but there was no copy-paste kubeconfig artifact anywhere. Onboarding a new engineer therefore meant "ask a teammate to hand you their stanza," which is exactly the tribal-knowledge step the repo convention calls out:

> When a new managed-cluster operational step becomes reproducible, document it in `k8s/onboarding.md`.

A client-side kubeconfig that's identical for every engineer is precisely that.

## Why a documented snippet rather than a checked-in kubeconfig file

The three contexts contain **no secrets** — just public hostnames plus a `pomerium-cli k8s exec-credential` exec plugin; all authentication happens at call time through Google OIDC and the cached Pomerium session. So a documented, ready-to-merge snippet is safe and is the right form. We deliberately do **not** check in anyone's real `~/.kube/config` (those mix in the break-glass 1Password admin contexts, CAPI/mgmt contexts, and personal `current-context` state).

The section also draws an explicit line against the cluster-admin kubeconfigs in the `tuist-k8s-<env>` 1Password vaults: those bypass Pomerium/impersonation entirely, are break-glass only, and are biometric-gated on purpose — not for routine reads, and never to be fetched by an agent.

## Impact

A new engineer (or anyone re-setting-up a laptop) can get `view` access to all three workload clusters by installing `pomerium-cli` (already pinned in the root `mise.toml`) and pasting one block — no grant, no provisioning, no asking around.

## Validation

Docs-only change. Verified the gateway hostnames against `infra/helm/pomerium/values-*.yaml` (note production is `kube-prod`, not `kube-production`), the `pomerium-cli` pin against the root `mise.toml`, and the exec-plugin invocation against `infra/helm/pomerium/NOTES.md`.

### DIFF
--- a/infra/AGENTS.md
+++ b/infra/AGENTS.md
@@ -77,7 +77,7 @@ This section is normative for agents (Claude, Codex, anything driving `kubectl` 
 
 The team's Google Workspace identity carries through to every workload cluster via Pomerium fronting `https://kube-<env>.tuist.dev`. From an agent running on a teammate's laptop, `kubectl --context tuist-k8s-<env>` works for any read operation: `get pods`, `logs`, `describe`, `get configmap`, `get events`. The `pomerium-cli` exec credential plugin handles the bearer; the teammate's existing browser-cached session is used silently. Each request then flows through the `kube-impersonator` sidecar in the Pomerium pod, which calls tuist-ops's `/api/v1/policy` over the tailnet; the policy returns `Impersonate-User: <email>` + `Impersonate-Group: tuist-eng` (or `tuist-admins` for Owner/Admin tailnet roles). RBAC binds those groups to `view`, which **excludes `Secret`s** — deliberate, so `MASTER_KEY`, `DATABASE_URL`, and ESO-synced secrets stay out of agent context.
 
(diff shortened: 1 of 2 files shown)


pepicrft (approved):

esnunes (approved):

fortmarek:
Folded into #11267 — both changes came from the same incident (the all-cluster kubectl outage), so combining them into one PR. The Pomerium read-access doc commit is cherry-picked there verbatim.

PR #11081, opened 2026-06-04
feat(registry): extract from cache into standalone service

pepicrft:
## What changed

This pull request extracts the Swift Package Registry read surface out of `cache/` into a standalone Phoenix service under `registry/`, while keeping the write and sync logic inside the Tuist server.

The current shape is:

- The `registry` pod is a stateless read frontend. It has no Postgres, no SQLite, no Ecto repo, no Oban journal, no local persistent volume, and no disk artifact cache.
- Swift registry writes and sync work live in the server under `Tuist.Registry.Swift.*`. A separate Deployment boots the existing server image with `TUIST_MODE=swift_registry_sync` and consumes only the `:swift_registry_sync` and `:swift_registry_release` queues.
- The read service serves canonical Swift Package Manager routes under `/swift/*` and preserves the legacy `/api/registry/swift/*` prefix that cache exposed before extraction.
- Legacy prefix responses carry `Deprecation: true` and `Sunset: Thu, 31 Dec 2026 23:59:59 GMT` headers. Generated release URLs and alternate-manifest `Link` URLs stay on the prefix the client used.
- Source archives return `303 See Other` redirects to presigned Tigris URLs. Manifests are served in-process so the registry response headers and alternate-manifest `Link` header stay on the registry origin.
- No live traffic moves on merge. `infra/registry-router/` still routes public registry traffic to the existing cache origins until a follow-up Cloudflare Worker change moves it intentionally.

## Why

The registry has a different operating shape from cache: Swift Package Manager content negotiation, package metadata reads, source archive redirects, and catalog sync from Swift Package Index. Keeping the read surface inside cache made registry changes a cache deploy and cache changes a registry risk.

The read service benefits from an independent ingress, runtime secrets, metrics, and rollout path. The writer side does not need a hard service boundary yet, so it follows the existing server `TUIST_MODE` pattern and reuses the server database, Oban queues, secrets, and observability wiring with a dedicated least-privilege `tuist_swift_registry_sync` role.

### Managed Helm deployment and nginx

The nginx requirement was removed instead of copied forward.

The controller no longer emits `X-Accel-Redirect`, and the chart does not need `/internal/local`, `/internal/remote`, `/storage`, `/data`, or a local hot cache. Archive requests now return `303 See Other` redirects to presigned object-storage URLs. Manifest routes stay in-process because Swift Package Manager depends on the registry response headers, especially alternate-manifest `Link` headers.

Relevant files:

[...]

PR #11263, opened 2026-06-12
ci(swift-registry): register deploy workflow files

pepicrft:
## Summary

Adds the three new `.github/workflows/swift-registry*.yml` files to `main` so they become dispatchable via `gh workflow run`. Without this, the staging deploy workflow defined in #11081 cannot be triggered against that PR's HEAD to produce pre-merge staging deployment evidence — GitHub only registers `workflow_dispatch` workflows that exist on the default branch.

## Why split out

#11081 (`refactor/extract-swift-registry`) is the substantive PR that adds the new service and chart. The fortmarek review on that PR asks for staging deployment evidence against its current SHA before merge. Triggering the staging deploy workflow requires it to be registered on main first.

## Net effect on main today

Zero. All three workflows are inert until #11081 lands:

- `swift-registry.yml` — `on: push / pull_request` with `paths: swift-registry/**, tuist_common/**`. Neither path exists on main, so the workflow registers but never fires.
- `swift-registry-deploy.yml` — `on: push: branches: [main], paths: [swift-registry/**, tuist_common/**]`. Same — paths don't exist, no auto-fire.
- `swift-registry-staging-deploy.yml` — `workflow_dispatch` only.

Once merged, `gh workflow run swift-registry-staging-deploy.yml --ref refactor/extract-swift-registry -f commit_sha=<sha>` becomes valid and we can deploy #11081's HEAD to `swift-registry-staging.tuist.dev` for smoke-testing.

## Test plan
- [ ] CI lint passes (workflows are syntactically valid)
- [ ] After merge, `gh api repos/tuist/tuist/actions/workflows` lists all three workflow IDs
- [ ] `gh workflow run swift-registry-staging-deploy.yml --ref refactor/extract-swift-registry -f commit_sha=170a8b1f1dea9bda6064b9805abeaec98f460d8e` accepts the dispatch

### DIFF
--- a/.github/workflows/swift-registry-deploy.yml
+++ b/.github/workflows/swift-registry-deploy.yml
@@ -1,0 +1,233 @@
+name: Swift Registry Deploy
+
+on:
+  workflow_dispatch:
+    inputs:
+      commit_sha:
+        description: Commit SHA to deploy (defaults to github.sha)
+        required: false
+      image_tag:
+        description: Pre-built image tag to deploy instead of building
+        required: false
+  push:
+    branches:
+      - main
+    paths:
+      - swift-registry/**
+      - tuist_common/**
+      - infra/helm/swift-registry/**
+      - .github/workflows/swift-registry-deploy.yml
+
+permissions:
+  contents: read
+  packages: write
+
+concurrency:
+  group: swift-registry-deploy-production
+  cancel-in-progress: false
+
+env:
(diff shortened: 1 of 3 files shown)


pepicrft:
Superseded — added a label-gated pull_request trigger to the staging deploy workflow in #11081 directly, so we don't need to land workflow files on main first. Applying the deploy-staging label on #11081 now fires the staging deploy from that PR's HEAD.

PR #11264, opened 2026-06-12
feat(noora, server): improve sidebar dropdowns, add overlay scrollbar…

asmitbm:
### Sidebar dropdowns 

* Clicking a group's chevron now only expands/collapses the group; clicking the label/icon navigates. Previously the whole row was a single link that did both, so it was impossible to peek into a group without leaving the current page. The trigger is now a real `<button>` with an aria-label, which also gives the toggle proper keyboard and screen-reader semantics.

* The dashboard layouts previously worked around the single-element design by disabling each group's collapsible on every page except its own `(disabled={@current_path != ...})` and only rendering the link conditionally. Those workarounds are removed, groups are always navigable and always toggleable. The Runners group had this logic inverted, which is also fixed by the removal.

* The chevron is now an animated icon (right → down, crossfade-rotate).

* Expanding/collapsing animates using zag.js's measured --height variable: open is a 250ms grow with fade-in; close is a faster (160ms) fade-dominant sweep, where items are fully faded by the halfway point so closing reads as a fade rather than a reversed slide. 

https://github.com/user-attachments/assets/86097bfd-889a-4c3f-b4b5-632849b2f209

### Overlay scrollbar for the sidebar (new NooraScrollArea hook)

* The sidebar's native scrollbar is replaced with a overlay: a 3px thumb that thickens to 6px on hover or drag, with drag-to-scroll and click-to-jump on the track. 

* A bottom gradient hints that there's more to scroll, and fades out when the bottom is reached so it never permanently obscures the last item.

https://github.com/user-attachments/assets/ae75ef7e-1376-4745-b87f-1f682ce62981

### Dashboard app-shell layout (server)

* The dashboard no longer scrolls the document. The shell is a fixed 100dvh column — headerbar on top, sidebar and content below — and the content pane scrolls in its own container, like the sidebar already does.

* The divider under the headerbar is now a border-bottom on the header itself rather than a separate element after it, so it moves (or rather, stays) as one unit with the header. Header vertical padding tightened to spacing-4.

Screenshot 2026-06-12 at 20 31 28

### Visual tweaks

The navigation loading bar is now Noora purple instead of topbar's default blue — `purple-500` in light mode, `purple-400` in dark. 
Screenshot 2026-06-12 at 20 30 44


fortmarek (commented):
https://github.com/user-attachments/assets/4db1f52d-5d8d-4fe6-bbe5-532e3ac21b21

The animation of the chevron should only run when it collapses/expands, right now, it runs every time.

noora/js/ScrollArea.js
```diff
@@ -0,0 +1,137 @@
+// Overlay scrollbar for a scrollable viewport. The native scrollbar is hidden
+// (so it never reserves a gutter or shifts content) and replaced by an
+// absolutely positioned track + thumb. The thumb is sized and moved from the
+// viewport's scroll metrics, and supports dragging and track-click paging.
+// The track only renders for fine pointers; touch scrolling is untouched.
+
```
fortmarek, line 6: I would suggest not to be replacing base browser components like a scrollbar with completely custom JS code. This breaks some basic behavior like scroll on space. Are you sure there's not an alternative to this?

fortmarek (changes requested):

asmitbm:
> The animation of the chevron should only run when it collapses/expands, right now, it runs every time.

This is fixed

noora/js/ScrollArea.js
asmitbm: I looked into the native-CSS alternative (`::-webkit-scrollbar` + `scrollbar-width/scrollbar-color`). The blockers there:

* Firefox doesn't support the `::-webkit-scrollbar` pseudo-elements at all, so the thin→thick-on-hover behavior just can't be done, it'd be a flat, inconsistent look across browsers.

* On OSes set to "always show scrollbars," the native bar reserves a gutter, which introduces the content-shift

So native doesn't actually give us the cross-browser-consistent. I've manually tested the custom one across keyboard (space/page/arrows/home-end), wheel, drag, track-click, and reduced-motion, and it behaves correctly. Scope is intentionally just the sidebar for now, if we roll it out more widely (dropdowns, etc.) I'm happy to revisit and harden it, or reconsider native at that point

fortmarek: > Firefox doesn't support the ::-webkit-scrollbar pseudo-elements at all, so the thin→thick-on-hover behavior just can't be done, it'd be a flat, inconsistent look across browsers.

I wouldn't prioritize browser consistency of a scrollbar, personally.

> I've manually tested the custom one across keyboard (space/page/arrows/home-end), wheel, drag, track-click, and reduced-motion, and it behaves correctly.

What about a11y?

I'm just still not fully convinced the minor wins are worth rolling out our own scrollbar.

asmitbm: > I wouldn't prioritize browser consistency of a scrollbar, personally.

It’s not just the scrollbar thickness, there’s also a slight layout shift happening in the sidebar. The scrollbar behavior was inspired by Vercel’s dashboard and works very similarly.

Since this scrollbar is scoped only to the sidebar and doesn’t affect other areas of the dashboard, it should be completely fine.

Screenshot 2026-06-15 at 17 25 32

> What about a11y?

As for a11y:

* Screen readers - unaffected. SR users never touch a scrollbar; they navigate by virtual cursor and the browser auto-scrolls the real container to follow. Because the viewport is a genuine scroll container, this works unchanged. The thumb/track are empty, non-focusable divs, so they add no tab stops or announcements.
* Touch - untouched. The track only renders under @media (pointer: fine), so touch users get native momentum scrolling.
* Reduced motion - respected. The thumb position is a direct transform (not animated), and the width transition is gated by the reduced-motion query.

fortmarek: @asmitbm have you checked Supabase? I don't think they use a completely custom scrollbar
image

Another thing that for example typically works with a scrollbar is clicking above/below a scrollbar scrolls up/down. These interactive components are just _so_ hard to get completely right and I personally dislike when web pages break default behavior like this. 

What are all the things you've tried to retain the default scrollbar while fixing basically the only issue worth fixing (the layout shift). I also wasn't able to reproduce the layout shift, but I'm not 100 % sure how this was supposed to be reproduced in the first place.

fortmarek:
The chevron doesn't seem to be vertically centered with the text

image

image

asmitbm:
> The chevron doesn't seem to be vertically centered with the text

Ahh missed this, fixed now

noora/js/ScrollArea.js
asmitbm: > @asmitbm have you checked Supabase? I don't think they use a completely custom scrollbar

Yupp, they have mostly changed just the color of their scrollbar to match their branding.

> Another thing that for example typically works with a scrollbar is clicking above/below a scrollbar scrolls up/down. 
 
This behaviour is preserved.

https://github.com/user-attachments/assets/8d72a089-f7c4-4add-b9a2-087bb96b1bb1

> These interactive components are just so hard to get completely right and I personally dislike when web pages break default behavior like this.

> What are all the things you've tried to retain the default scrollbar while fixing basically the only issue worth fixing (the layout shift). I also wasn't able to reproduce the layout shift, but I'm not 100 % sure how this was supposed to be reproduced in the first place.

I tried using the native scrollbar, but the interaction I was aiming for, transitioning from 3px in the default state to 6px on hover or drag, isn't really possible with the native implementation. That limitation was also contributing to the layout shift.

Another reason for this approach is visual consistency. A thicker native scrollbar would only appear occasionally when the sidebar content overflows, such as when all three dropdowns are expanded or when the Tests section is expanded. In most cases, the scrollbar won't be visible at all. When it does appear, having a thicker scrollbar alongside the expanded sidebar adds more visual noise than necessary, which is what I was trying to avoid.

I also looked into whether a native scrollbar could be overlaid using CSS, but that isn't possible without implementing a custom solution. As shown in the screenshot, native scrollbars reserve a gutter of around 20px, even when the scrollbar itself isn't visible.

With the custom scrollbar, I can overlay it on top of the content without reserving any extra space, allowing it to respect the sidebar width and maintain a cleaner layout.

Screenshot 2026-06-15 at 18 58 57

Screenshot 2026-06-15 at 19 02 02

A few other platforms which use custom scrollbar only for sidebar:

Cloudflare:
Screenshot 2026-06-15 at 19 03 16

Claude:
Screenshot 2026-06-15 at 19 03 38

Ona:
Screenshot 2026-06-15 at 19 57 57


Firecrawl:
Screenshot 2026-06-15 at 19 07 56

fortmarek (approved):
Let's roll this out with the custom scrollbar 👍

PR #11266, opened 2026-06-12
fix(infra,kura): restore Kura gRPC upload throughput (gateway + runtime HTTP/2 windows, activity-based timeout)

esnunes:
## What changed

Restores large-blob upload throughput to Kura over gRPC (REAPI `ByteStream.Write`). The regression has two stacked ceilings — the gateway nginx in front of Kura, and Kura's own gRPC server — so the fix has two layers, plus a CI/deploy change that made the second layer testable on staging.

**1. Gateway nginx HTTP/2 windows** — raised on every Kura gateway, in both render paths:
- Dedicated gateways (`kgw-<account-hash>-<region>`): the `KuraGateway` controller ConfigMap (`gatewayNginxConfigData`).
- Regional gateways (`kura-eu-central` / `kura-us-east` / `kura-us-west`): the platform chart's ingress-nginx `controller.config`.

```yaml
client-body-buffer-size: "4m"                 # WINDOW_UPDATE pacing while streaming upstream
http2-max-concurrent-streams: "32"            # bounds worst-case nginx memory per connection
http-snippet: "http2_body_preread_size 4m;"   # initial advertised request-body window
```

**2. Kura gRPC runtime** (`kura/src/reapi/mod.rs`) — so the Kura hop isn't the next bottleneck once nginx is fixed:
- Raise the tonic/hyper HTTP/2 windows from their 1 MiB defaults: `initial_stream_window_size = 4 MiB`, `initial_connection_window_size = 16 MiB`.
- Keep `max_connection_age` at 300s (so connections still recycle every 5 min and rebalance onto fresh pods after a rolling deploy) but raise `max_connection_age_grace` 300s → 900s. The age-boundary GOAWAY only stops *new* streams — an in-flight upload runs until the grace expires — so the 300s grace was the actual bug that cut large uploads at age+grace=600s. 900s comfortably outlasts the largest observed upload (~680s), and worst-case connection lifetime stays ~20 min (vs the 2h a 3600+3600 pair would let a slow-but-active stream hold a connection + temp-file slot).
- Add a **per-chunk stall timeout** on the upload path: each received chunk resets a 60s timer, so a connection with data actively flowing is never cut, but a stalled/vanished client is reclaimed promptly (returns `DEADLINE_EXCEEDED` and removes the temp file). This is the "kill only on inactivity, not on activity" behavior — tonic has no activity-based connection timeout, so it's enforced at the stream level where the upload actually happens.

**3. Build + deploy the Kura runtime image from Server Deployment** (`.github/workflows/server-deployment.yml`):
- The workflow already built the server and `kura-controller` images per commit, but the **kura runtime image** is otherwise only published by `release.yml` on push-to-main. So deploying a branch to staging shipped the chart (layer 1) while the kura **binary** stayed pinned to the latest released `kura@` tag — making layer 2 impossible to validate pre-release. (Confirmed in practice: a staging deploy of this branch left the pods on `ghcr.io/tuist/kura:0.10.1` and the 784MB blobs still died at exactly `300+300s`.)
- Now the build job also builds `ghcr.io/tuist/kura:sha-<SHA>` (from source, `kura/Dockerfile`, mirroring the controller step). The deploy job defaults `kuraRuntime.image.tag` to that per-commit image **on staging**, while **canary/production keep pinning the resolved `kura@` release tag** so they never run an unreleased binary. An explicit `kura_runtime_image_tag` input still overrides everywhere.

## Why

Large Bazel REAPI uploads to Kura were stuck in an infinite retry loop. The client's binary gRPC log showed every `ByteStream.Write` capped at **~310KB/s** = nginx's default 64KB HTTP/2 body window ÷ the 193ms RTT. At that rate the 784MB `librocksdb-sys` artifacts need ~42 min, but streams are force-closed at ~600s (Kura's `max_connection_age` + grace), so Bazel restarts from byte 0 forever: `Write` → 502 at 600s → `QueryWriteStatus` → `NOT_FOUND` → restart.

**Root cause** of the window cap: regression from #11192, which moved Kura gRPC from per-instance TCP-passthrough LBs (TLS terminated by tonic, 1MB windows ≈ 5.4MB/s/stream) behind ingress-nginx gateways. That PR's analysis covered aggregate NIC/CPU but not HTTP/2 per-stream flow control, and its smoke test exercised HTTP roundtrips, not large REAPI writes — so it shipped unnoticed. Both nginx knobs must move together: `http2_body_preread_size` sets the initial window, `client_body_buffer_size` (ingress default 8k) paces the `WINDOW_UPDATE`s once nginx relays the body.

The **kura-runtime layer was added after a first staging run showed the nginx fix alone plateaus**: per-stream throughput rose ~20x but capped at ~1.2MB effective window (Kura's own 1 MiB tonic window on the nginx→kura hop), and the 784MB blob still died at the hard 600s `max_connection_age`. Raising Kura's windows lifts the plateau; replacing the age-based kill with an inactivity timeout means active uploads finish regardless of size.

## Real-infra validation (staging)

Two staging phases, against `grpcs://grpc.tuist-eu-central-1-staging.kura.tuist.dev` (RTT ~209ms, client in Brazil, 300 Mbps uplink):

**Phase A — gateway fix only** (pods still on released `0.10.1`): per-stream upload rose from the ~0.3 MB/s baseline to ~6 MB/s solo, but plateaued at ~1.2MB effective window and the two 784MB blobs were severed at connection-age 600.9s (= `0.10.1`'s `300+300`), at 79% / 88%. This is what motivated layers 2 and 3.

**Phase B — full fix deployed** (pods on `ghcr.io/tuist/kura:sha-518f3597e34c`, the branch binary, via change #3): the regression is gone.

| Metric | baseline (pre-fix) | Phase A (nginx fix only) | **Phase B (nginx + runtime)** |
| --- | --- | --- | --- |
| 784MB blob outcome | ∞ retry loop | 502 at ~88%, died at 600s | ✅ **100% in 167–223s, single stream** |
| big-blob per-stream | ~0.31 MB/s | ~1.15 MB/s (before dying) | **3.5–4.7 MB/s** |
| both giants combined | — | ~2.2 MB/s (failed) | **7.0 MB/s** |
| aggregate upload phase | — | 4.0 MB/s | **5.9 MB/s** |

In Phase B **all 337 `Write`s succeeded with zero `QueryWriteStatus`** — both ~784MB RocksDB artifacts uploaded to 100% in a single stream (`784,550,880` in 167s, `783,930,496` in 223s), no connection-age teardown, no restart-from-zero.

**Warm-cache run** (everything already cached): 36s end-to-end, 803 `GetActionResult` hits + 98 `Read`s (90.7 MB, incl. the single 90.6 MB artifact) all OK, zero uploads, no auth errors — the happy path is fast and clean.

## Local e2e validation

`kura/test/e2e/grpc-upload-throughput/` — a toxiproxy + nginx + kura docker-compose harness that runs one kura backend behind baseline-vs-patched nginx configs (generated from the live chart values) under identical WAN latency, and asserts the patched path is ≥4x faster. Measured (16MB payload): at 192ms RTT, baseline 0.32 MB/s vs patched 12.24 MB/s (**38.7x**); baseline reproduces the production ~310 KB/s exactly. Runs as the `gateway-throughput` shard of the Kura e2e workflow. A controller unit test (`TestGatewayNginxConfigMatchesChart`) keeps the dedicated and regional render paths in sync.

Note: the e2e exercises a single stream against a bare fast kura, so it validates the nginx window in isolation and over-predicts; the staging runs are what surfaced the kura-side plateau (hence layer 2) and proved the end-to-end fix (hence layer 3).

## Verification

- `infra/kura-controller` `go test ./...` passes (incl. the chart-vs-controller equality assertion).
- `helm template` renders the three window keys into each regional gateway ConfigMap (anchored to one source).
- `kura/src/reapi/mod.rs` changes compile (`cargo check --lib`) and pass `cargo clippy`.
- `server-deployment.yml` lints clean (`actionlint`); staging deploy confirmed the pods rolled onto `kura:sha-518f3597e34c`.

## Deferred follow-up

Credential-token TTL vs. long-lived sessions. Now that uploads complete and builds run longer, this is the leading remaining symptom: in the full cold build, **1,089 of 2,886 `Read`s failed `UNAUTHENTICATED` in a single burst ~414s (~7 min) into the build** — a fixed-TTL token expiring mid-invocation, not a Kura fault. Bazel doesn't retry them, so those cache reads are lost (build still completes, with a degraded cache-hit rate where the burst lands). The warm-cache build (36s, under the TTL) never hits it. Fix is outside the upload path — longer-lived REAPI tokens or a Bazel credential helper that refreshes proactively / on `UNAUTHENTICATED` — and is tracked separately.


fortmarek:
Codex review findings:

- [P2] Staging re-deploys with `image_tag` can deploy a non-existent Kura runtime image. The build job is skipped when `inputs.image_tag` is set, but staging still defaults `kuraRuntime.image.tag` to `$IMAGE_TAG` in `.github/workflows/server-deployment.yml`. Since the new `ghcr.io/tuist/kura:sha-*` image is only built inside the skipped build job, a manual staging re-roll of a prebuilt server image can send Kura into `ImagePullBackOff` unless the caller also knows to pass `kura_runtime_image_tag`. I’d only use `$IMAGE_TAG` for staging when this run actually built the runtime image; otherwise fall back to the resolved `kura@` tag unless explicitly overridden.

- [P2] `ByteStream.Write` still leaks partial temp files on immediate stream errors. The timeout branch removes `temp_path`, but `Ok(result) => match result?` in `kura/src/reapi/mod.rs` returns early on transport/cancel errors without cleanup. That is exactly the “client vanished” class this change is trying to reclaim; those partial files count toward `ensure_tmp_dir_capacity`, so repeated cancelled uploads can exhaust the tmp budget. Handle `Err(status)` explicitly and remove the temp file before returning.

- [P3] The throughput harness docs/output are stale after the runtime-window change in this same PR. `kura/test/e2e/grpc-upload-throughput/client/main.go` and the harness README still describe `direct_kura` as tonic’s default ~1 MiB window, but Kura now advertises 4 MiB. Future readers will misinterpret the control path and measured ceilings.

Validation: I ran `go test ./...` in `infra/kura-controller` from a PR-head snapshot and it passed. I did not run the Rust compile/e2e throughput suite locally.

kura/src/reapi/mod.rs
```diff
@@ -753,7 +781,24 @@ impl ByteStream for ReapiService {
...
         let mut finished = false;
 
-        while let Some(chunk) = stream.message().await? {
+        loop {
+            // Bound only on inactivity: every received chunk resets the timer,
+            // so an upload that keeps sending is never cut, but a stalled client
+            // is reclaimed instead of holding the temp file open indefinitely.
+            let chunk =
+                match tokio::time::timeout(REAPI_WRITE_STALL_TIMEOUT, stream.message()).await {
+                    Ok(result) => match result? {
```
pepicrft: When `stream.message()` returns `Ok(Err(Status))` (transport drop, client RST mid-upload), the `result?` here propagates without calling `remove_file_if_exists(&temp_path)`. The `Err(_elapsed)` arm right below does clean up. Same pattern on the `write_all` `?` further down at line 857. Both are pre-existing, but the new 1200s connection lifetime widens the window the leaks live in. Worth a small helper that always cleans temp on early return.

kura/src/reapi/mod.rs
```diff
@@ -58,6 +58,32 @@ use crate::{
...
+
+// A REAPI connection is recycled after this age: the server sends GOAWAY (stop
+// opening new streams), then lets in-flight RPCs run for the grace period before
+// closing. GOAWAY does not sever an in-flight upload — the grace does — so the
+// age stays short (mesh rebalancing onto fresh pods after a rolling deploy), and
+// only the grace is sized to outlast the largest legitimate upload. The previous
+// 300s grace was the bug: it cut large uploads at age+grace=600s. Worst-case
+// connection lifetime is now age+grace=20min, down from the 2h a 3600+3600 pair
+// allowed a slow-but-active stream to hold. Stalls are still bounded by
+// REAPI_WRITE_STALL_TIMEOUT, so a connection with data flowing is never cut.
```
pepicrft: The comment says a connection with data flowing is never cut, but age + grace = 1200s is still an absolute cap. The patched throughput makes a typical large upload finish in well under a minute so this is fine in practice, but a TB-class artifact or a customer on a genuinely slow link can still hit it. Mostly want to either soften the comment to match reality, or expose the grace as an env knob so on-call can bump it without a release.

.github/workflows/server-deployment.yml
```diff
@@ -647,13 +670,22 @@ jobs:
...
+          #   2. staging tracks the per-commit sha-<SHA> image built in the
+          #      build job — like server + kuraController — so an in-flight
+          #      kura change is exercised on staging before it's released.
+          #   3. canary/production pin the released kura@ tag resolved from
+          #      git, so they never run an unreleased binary.
           kura_runtime_tag="${KURA_RUNTIME_IMAGE_TAG:-}"
           if [ -n "$kura_runtime_tag" ]; then
-            echo "Overriding resolved Kura runtime tag with input ${kura_runtime_tag}"
+            echo "Kura runtime: using explicit input tag ${kura_runtime_tag}"
+          elif [ "${{ inputs.environment }}" = "staging" ]; then
```
pepicrft: This breaks for a manual `workflow_dispatch` with `inputs.image_tag` set and no `kura_runtime_image_tag`. The build job is gated `if: inputs.image_tag == ''` (line 125) so it skips entirely, but this elif still pins `kuraRuntime.image.tag` to `$IMAGE_TAG`. Helm ends up referencing `tuist/kura:<image_tag>` that was never pushed, kura pods land in ImagePullBackOff, and the deploy rolls back. Either also resolve from the released `kura@` tag when the build was skipped, or only take this branch when the build actually ran.

.github/workflows/server-deployment.yml
```diff
@@ -198,6 +199,28 @@ jobs:
...
+      # Staging-only on purpose, in lockstep with the deploy job: only staging
+      # pins kuraRuntime.image.tag to this sha image; canary/production resolve
+      # the released `kura@` tag and never reference it, so building it there
+      # would be a wasted ~10-20min Rust compile + an unused push. The
+      # production cascade already skips this whole job (it passes image_tag),
+      # so this guard only matters for a manual canary/production dispatch.
+      # kura/Dockerfile builds from source (context = kura/); the staging
+      # node pool is amd64 (Hetzner ccx), so the runner-native build suffices.
+      - uses: docker/build-push-action@v6
+        name: Build and push Kura runtime image
```
pepicrft, line 219: Two things on this step. No `cache-from`/`cache-to`, so every staging deploy pays a full cold Rust + librocksdb-sys compile, while the server build right above uses registry buildcache. And the build fires unconditionally on every staging deploy, so a server-only PR still pays the 10-20min compile for a byte-identical kura image. Mirroring the server cache config (`buildcache:kura`) would cover the first; a paths-filter check on `kura/**` would skip the build when kura wasn't touched.

infra/kura-controller/controllers/gateway_nginx_config_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+// chartValuesPath is the platform Helm chart that renders the regional Kura
+// gateway ingress-nginx ConfigMaps, relative to this package.
+const chartValuesPath = "../../helm/platform/values.yaml"
+
+// windowKeys are the HTTP/2 upload-window settings that must be identical
+// between the dedicated-gateway ConfigMap (this controller) and the regional
+// gateway config (the platform chart). If they diverge, large gRPC uploads
+// regress on one path but not the other.
+var windowKeys = []string{
```
pepicrft: windowKeys covers only the 3 new HTTP/2 keys, but the controller and chart also share the 4 base keys (use-forwarded-headers, use-proxy-protocol, compute-full-forwarded-for, upstream-keepalive-timeout) plus a handful more. If someone bumps one of those in only one place, this still passes. Also the test samples us-west specifically and trusts the YAML anchor to cover eu-central / us-east; if a future PR de-anchors a region, drift on the other two is invisible. Worth iterating over every `kura-*-ingress-nginx` block and asserting all shared keys.

kura/test/e2e/grpc-upload-throughput/generate-confs.sh
```diff
@@ -0,0 +1,94 @@
...
+#
+#   patched.conf  = template + window directives derived from values.yaml
+#   baseline.conf = template with the window directives removed (nginx defaults)
+set -euo pipefail
+
+cd "$(dirname "$0")"
+
+# Repo-root chart that renders the regional Kura gateway nginx config.
+CHART_VALUES="${CHART_VALUES:-../../../../infra/helm/platform/values.yaml}"
+GATEWAY_KEY="${GATEWAY_KEY:-kura-us-west-ingress-nginx}"
```
pepicrft: Same drift fragility as the controller test: hardcoding us-west means if the anchor ever gets unwound for a per-region override, this harness silently keeps passing because it only reads us-west's config. The anchor sharing is a comment-level invariant, nothing enforces it.

.github/workflows/kura.yml
```diff
@@ -7,11 +7,17 @@ on:
     paths:
       - kura/**
       - .github/workflows/kura.yml
+      # The gateway nginx config the upload-throughput e2e validates lives in
+      # the platform chart, so chart window changes must exercise this workflow.
+      - infra/helm/platform/values.yaml
```
pepicrft: This triggers the full kura matrix (clippy, tests, audit, every e2e shard) on any `values.yaml` edit, including ones that touch metrics-server / observability sections with no kura impact. Could scope to just the gateway-throughput shard for this file, or further filter to PRs that actually touch the `kura-*-ingress-nginx` blocks.

kura/test/e2e/grpc-upload-throughput/client/main.go
```diff
@@ -0,0 +1,271 @@
...
+	)
+	if err != nil {
+		return 0, 0, err
+	}
+	defer conn.Close()
+	client := bs.NewByteStreamClient(conn)
+
+	var lastErr error
+	for attempt := 0; attempt < 60; attempt++ {
+		if err := uploadBlob(client, 4096, chunk, seed*100000+attempt); err == nil {
```
pepicrft, line 372: Warmups run serially per target with up to 60x1s retries, so worst case the harness sits ~180s before any measurement starts. The three targets are independent and could warm in parallel. Also each retry writes a unique blob name (`seed*100000+attempt`), so flaky warmups create up to 60 distinct write paths per target; a fixed warmup blob per target would let kura's content addressing dedup short-circuit.

kura/test/e2e/grpc-upload-throughput/generate-confs.sh
```diff
@@ -0,0 +1,94 @@
...
+YQ_IMAGE="${YQ_IMAGE:-mikefarah/yq:4}"
+
+if [ ! -f "$CHART_VALUES" ]; then
+  echo "generate-confs: chart values not found at $CHART_VALUES" >&2
+  exit 1
+fi
+
+# Read one config key from the gateway block. The three regional blocks share
+# one anchored node, so any of them yields the deployed values.
+cfg() {
```
pepicrft: `cfg()` shells out to `docker run mikefarah/yq:4` three times here. The first invocation also pulls from Docker Hub anonymously, which is rate-limit territory on CI. A single yq invocation emitting all three values via `@tsv` would drop 2 of the 3 container starts.

pepicrft (approved):
Did a deep review on this. Two real correctness things I'd flag (temp file leak on stream errors, and the ImagePullBackOff scenario on a manual staging dispatch with image_tag set), plus a few drift / CI efficiency notes inline. Approving so this isn't gated on them since the throughput win is the main thing, but worth tracking the first two as follow-ups.

.github/workflows/server-deployment.yml
esnunes: Regarding making the build conditional on kura changes:
- As you may deploy any branch to staging the path filter won't work, the condition would have to consider a diff between HEAD of the branch and the latest version of kura, which would make the workflow even more complex

.github/workflows/kura.yml
esnunes: I will address this request later, I've created this [issue](https://github.com/tuist/tuist/issues/11321) to cover it

kura/test/e2e/grpc-upload-throughput/generate-confs.sh
esnunes: I've added this request to the [issue](https://github.com/tuist/tuist/issues/11321) and will work on it later

kura/test/e2e/grpc-upload-throughput/client/main.go
esnunes: the warm up exists only to make sure the connection is establish and is responding, the data is ignored, the code expects one valid upload to kick off the real measurement

kura/test/e2e/grpc-upload-throughput/generate-confs.sh
esnunes: I've replaced this script with the go client

PR #11294, opened 2026-06-15
feat(server): self-hosted Kura nodes with per-account mesh CA enrollment

fortmarek:
> Lets a customer run their own Kura cache nodes, authorized by Tuist's control plane, and bridge them into the same mutually-authenticated mesh as their Tuist-managed nodes. Typically the nodes sit next to a customer's runners, so writes against the self-hosted nodes replicate into the shared managed cache, and a node warms from that cache when it joins. A customer never has to choose between owning the whole mesh and running a single node; spinning one up is a credential and a URL.

### Why

Customers want Kura cache nodes in their own infrastructure (locality, data residency). The control plane should own authentication, certificate issuance, and discovery so the customer supplies only a credential and a URL.

Two load-bearing constraints shaped the design:

- Kura's hot-path auth first tries to verify a Tuist Guardian JWT locally using a **symmetric** HS512 secret. That secret can mint tokens for any tenant, so it cannot be handed to a customer. Self-hosted nodes run **without** the verifier and authorize uncached requests via the control plane's introspection endpoint; making that introspection tenant-scoped is the security boundary.
- Kura's peer mTLS verifier trusts any client cert that chains to the configured CA, with no identity check. The **CA is the trust boundary**, so isolation comes from a **per-account CA**.

### What changed

**Control-plane authorization**
- Tenant-scoped credential: per-account `kura_self_hosted_clients` (Bcrypt at rest with the standard pepper, plaintext shown once, a suffix-masked hint thereafter), issued and revoked from the account's Cache page.
- Per-account introspection and usage constraint: both endpoints branch the unconstrained Tuist control-plane client from a tenant-scoped client, so a self-hosted node can only introspect its own tenant's tokens and report its own usage.
- Lease-based registration: self-hosted nodes self-register their client-facing URL via heartbeats, and that URL is the endpoint the CLI routes to (resolved under the `:kura` technology). A node ages out of rotation when heartbeats stop. There is no manually-managed endpoint list; the registered node is the endpoint.
- Cache-API authentication: the Tuist auth hook ships in the Kura image, so a self-hosted node enables it with `KURA_EXTENSION_*` to require a valid Tuist token on every cache request, authorized via tenant-scoped introspection (no symmetric verifier secret, fail-closed). Covered by introspection-only tests against the real hook; managed pods keep their ConfigMap-mounted hook, which shadows the bundled file.

**Enrollment (controller-managed per-account CA)**
- The Kura controller owns the per-account peer CA (the `kura-<handle>-peer-ca` secret). `Mesh.enroll_node` and `sign_node_certificate` read that secret and sign the node's CSR from it with an issuer-controlled SAN, so the node's leaf is trusted by the managed pods (same CA) and vice versa. The enroll endpoint returns `503 ca_unavailable` when the account has no managed mesh yet.
- Kura enroll-on-boot generates the keypair locally (the private key never leaves the node), enrolls, writes the cert material to the `KURA_INTERNAL_TLS_*` paths, and adopts the returned tenant and peers.

**Two-way bridge**
Managed-mesh discovery advertises each pod's **in-cluster** `*.svc.cluster.local` peer URL, which an off-cluster node cannot reach. The bridge closes that gap:
- Public peer plane (controller): new `KuraInstance` fields `meshPublicPeerHost`, `meshExternalPeers`, and `meshPublicPeerLoadBalancerAnnotations`. `reconcileAccountPublicPeerService` provisions an account-level L4 LoadBalancer on the peer port (TLS-passthrough, `externalTrafficPolicy: Local`, hcloud `location`/`node-selector` plus external-dns annotations); the peer-cert SAN covers the public host; the NetworkPolicy admits the peer port from `0.0.0.0/0` when public peering is on (the mutual-TLS client cert is the auth boundary).
- Gateway advertisement (Kura): managed pods set `KURA_PEER_GATEWAY_URL`, and `internal_status` returns the gateway URL to any request that arrived via the gateway host, read from the HTTP/2 `:authority` and falling back to the h1 `Host` header. A node skips a discovered peer whose `node_url` is its own gateway, so same-region managed pods do not hairpin through the public LoadBalancer.
- Server: the provisioner sets these fields per mesh region, and `Mesh.mesh_peers` seeds the gateway URL into the enrollment peer list (`Regions.peer_public_url`).

Net: a self-hosted node enrolls, receives the gateway as its peer, pushes its writes through it into the managed mesh (which re-replicates internally), and warms from that mesh at join, all without ever accepting an inbound connection.

### Replication model and scope

Replication is point-to-point push of each node's own client writes: on a write, a node enqueues and pushes the artifact to its configured peers, while artifacts it receives are applied locally and not forwarded (there is no gossip). Self-hosted nodes typically sit next to a customer's runners, which do most of the writing, so the directions that matter are covered:

- **Self-hosted to self-hosted (continuous).** Each node pushes its writes to its sibling peers, so a customer's own mesh stays consistent. The nodes are configured to peer with each other.
- **Self-hosted to Tuist-managed (continuous).** Each enrolled node pushes its writes through the gateway into the managed mesh, which re-replicates internally. Because the nodes are outbound-only, managed cannot pull from them; the data flows by the nodes pushing. Each write-ingesting node peers with the gateway (with no gossip, a write on a non-bridged node does not reach managed transitively).
- **Tuist-managed to self-hosted (join-time snapshot).** A node warms from the managed mesh when it joins. Continuously propagating new managed-side writes to an already-joined node is a deferred follow-up; it is not required for the runner-adjacent use case where the self-hosted nodes are the write origin.

### Security model

Isolation is the per-account CA described above: a node's leaf chains only to its account's CA, so it can join only that account's mesh and cannot present a cert any other account's nodes would accept. Properties that hold in both topologies:

- The node's private key is generated on the node and never leaves it (enrollment sends only a CSR).
- Self-hosted nodes never receive the symmetric Guardian verifier secret (which could mint tokens for any tenant); cache-API auth is tenant-scoped introspection, fail-closed.
- Nodes are outbound-only and accept no inbound connections.

What crosses into Tuist depends on the topology, and customers should pick deliberately:

- **Standalone** (own CA, own mesh, no managed region): artifact bytes never leave the customer's infrastructure. Tuist receives only the control-plane metadata it needs to function (registration heartbeats, usage events, token introspection). This is the zero-trust path.
- **Bridged** (Tuist-managed CA + node-to-managed push): by design, artifacts a node ingests are pushed into the managed mesh, so their bytes and manifests (keys, namespaces, sizes) reside in Tuist's infrastructure. The per-account CA is also Tuist-managed (the server reads it to sign CSRs). Because nodes are outbound-only, a compromise of that CA key cannot pull a node's local cache; its worst case is **integrity** (a malicious gateway serving poisoned artifacts to a node that pulls from it), not exfiltration. A customer who requires that the vendor be cryptographically unable to touch their cache should run standalone, or a future customer-held-CA option for bridged.

### End-to-end validation (staging)

A locally-run node was deployed against the real staging managed mesh. It enrolled, got an account-CA-signed leaf (issuer is the account's peer CA), joined, and reached `state: serving`.

- Warm-up pull (managed to self-hosted, at join): the node pulled the account's full mesh as its bootstrap snapshot; its data directory grew to ~6.8 GB.
- Push (self-hosted to managed, continuous): artifacts written through the node were pushed to the gateway and accepted by the managed mesh (`upsert_artifact result=ok`, with the replication counter incrementing across writes).

### Caveats and follow-ups

- **Continuous managed-to-node propagation is a deferred follow-up.** As described under Replication model and scope, a self-hosted node warms from the managed mesh at join (validated above) but does not yet continuously pull new managed-side writes. The node and node-to-managed directions, which carry the runner-adjacent use case, work via push and are the supported scope here.

### Docs

`server/priv/docs/en/guides/cache/self-host.md` covers Kura mesh self-hosting (superseding the Elixir cache instructions) for both topologies: bridged (managed mesh plus self-hosted nodes) and standalone (self-hosted only), including what enrollment auto-provisions (the peer TLS) versus what the customer provides.

### Validation

- Server: `mix test test/tuist/kura/ test/tuist/oauth/introspection_test.exs test/tuist_web/controllers/internal/kura_mesh_controller_test.exs test/tuist_web/controllers/internal/kura_usage_controller_test.exs test/tuist_web/live/cache_live_test.exs`.
- End-to-end on staging as described above (enroll to serving, two-way replication).


server/lib/tuist/kura/self_hosted_clients.ex
```diff
@@ -0,0 +1,97 @@
...
+      nil ->
+        Bcrypt.no_user_verify()
+        :error
+
+      %SelfHostedClient{} = client ->
+        if Bcrypt.verify_pass(
+             client_secret <> Environment.secret_key_password(),
+             client.encrypted_secret_hash
+           ) do
+          {:ok, client.account}
```
pepicrft, line 83: Can we move the self-hosted cache entitlement check into this authorization path, or into every caller before accepting the credential? Creation is gated by `Entitlements.allows?(account, :self_hosted_cache)`, but an existing credential still verifies here after the account loses that entitlement. Enrollment, usage ingestion, and token introspection all trust this function, so a downgraded account can keep running self-hosted nodes. The registered cache address path also only checks the Kura feature flag before exposing ready self-hosted addresses to command-line clients.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -36,10 +37,12 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
     with {:ok, hook_script} <- hook_script(inputs) do
       gateway = gateway_assignment(account, region)
 
+      external_peers = mesh_enabled?(region) && Mesh.self_hosted_peer_urls(account)
```
pepicrft: Can enrollment trigger a manifest refresh for active managed Kura servers, or can the reconciler compare `meshExternalPeers` as desired state? This value is only read while rendering a `KuraInstance`. When a user-hosted node enrolls later, the database gets the peer address, but already-converged managed pods keep their old `KURA_PEERS` until an unrelated rollout reapplies the manifest. That leaves the managed-to-user-hosted replication leg stale.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -346,6 +349,67 @@ func (r *KuraInstanceReconciler) reconcileAccountPeerService(ctx context.Context
...
+// reconcileInstancePublicPeerService exposes this region's peer plane to the
+// internet so a customer's self-hosted Kura node can dial in for replication.
+// It is a plain L4 LoadBalancer: the peer connection is end-to-end mTLS, so
+// nothing terminates TLS here. external-dns publishes the region's
+// MeshPublicPeerHost to the assigned address. One Service per instance (region)
+// so the host, cert, and selected pods stay aligned across a multi-region
+// account. Only created when the controller owns the account CA
+// (meshManagedPeerTLS) and a public host is requested.
+func (r *KuraInstanceReconciler) reconcileInstancePublicPeerService(ctx context.Context, instance *kurav1alpha1.KuraInstance) error {
+	if !meshManagedPeerTLS(instance) || instance.Spec.MeshPublicPeerHost == "" {
```
pepicrft, line 362: Can this branch delete the public peer Service when mesh is disabled or `meshPublicPeerHost` is cleared? Right now the function returns without touching the existing `*-peers-public` Service, so an internet-facing LoadBalancer and its hostname annotation can remain after the spec stops asking for it. The other public resources in this controller delete themselves when disabled, so this should follow the same pattern.

server/data-export.md
```diff
@@ -24,6 +24,8 @@ Sensitive authentication data (passwords, tokens) are excluded from exports.
 - Organization SSO configuration metadata, including the configured SSO provider, provider URL, and full OAuth2 endpoint URLs
 - Kura server records (`kura_servers` table): per-account Kura server configuration including region, image tag, public URL, status, and the observed-state projection (`observed_image_tag`, `last_observed_at`) recording which image the backing cluster reports running and when it was last observed
 - Kura deployment history (`kura_deployments` table): rollout attempts for the account's Kura servers including image tag, status, error messages, and start/finish timestamps
+- Self-hosted Kura credentials (`kura_self_hosted_clients` table): tenant-scoped credentials a customer uses to run self-hosted Kura nodes, including the public `client_id`, friendly name, last-used timestamp, and `secret_last_four` (the trailing four characters of the secret, kept as a masked-preview hint). The `encrypted_secret_hash` (Bcrypt hash of the client secret) is excluded from exports as an authentication secret, and the plaintext secret is never stored.
```
pepicrft, line 27: Can this mention the enrolled self-hosted peer addresses stored as `account_cache_endpoints.technology = :kura_self_hosted_peer`? This change introduces customer infrastructure address data in the existing table. The document currently covers credentials and registered client-facing endpoints, but not these peer addresses.

pepicrft (changes requested):
I am requesting changes because the entitlement and reconciliation issues can leak through after merge.

Validation run:

- `mix test test/tuist/kura/mesh_test.exs test/tuist/kura/registrations_test.exs test/tuist/kura/self_hosted_clients_test.exs test/tuist/oauth/introspection_test.exs test/tuist_web/controllers/internal/kura_mesh_controller_test.exs test/tuist_web/controllers/internal/kura_usage_controller_test.exs test/tuist_web/live/cache_live_test.exs`: 56 tests, 0 failures
- `mise exec -- cargo test enrollment`
- `mise exec -- cargo test registration`
- `mise exec -- cargo test skips_self_and_own_gateway_but_adopts_other_peers`

I could not run the Go controller tests locally because the Go executable is not available in this environment.

server/lib/tuist/kura/self_hosted_clients.ex
fortmarek: Done. `verify/2` now also requires `Entitlements.allows?(account, :self_hosted_cache)`, so a downgraded account stops authenticating for enrollment, usage ingestion, and introspection even with a live credential. `Accounts.registered_kura_endpoint_urls/1` gates the advertised node addresses on the same entitlement.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
fortmarek: Done. The desired manifest revision now folds in the account's enrolled self-hosted peers (`manifest_revision/2`), so enrolling or dropping a node bumps the revision and the reconciler reapplies the `KuraInstance` instead of leaving `meshExternalPeers` stale until an unrelated rollout.

infra/kura-controller/controllers/kurainstance_controller.go
fortmarek: Done. The disabled branch now deletes the `*-peers-public` Service (ignoring NotFound), matching the other public resources in the controller, so the internet-facing LoadBalancer no longer lingers after the spec stops asking for it.

server/data-export.md
fortmarek: Done. `data-export.md` now documents the enrolled self-hosted peer addresses stored as `account_cache_endpoints.technology = :kura_self_hosted_peer`.

pepicrft (approved):
I validated the follow-up changes and the previous findings are addressed.

The self-hosted credential path now enforces the self-hosted cache entitlement at verification time, and registered self-hosted endpoints are hidden for non-entitled accounts. The managed Kura manifest revision now folds in enrolled peer addresses, so peer changes trigger a reapply. The public peer Service is deleted when the public peer host is cleared, and the data export documentation now covers enrolled self-hosted peer addresses.

Validation run:

- `mix test test/tuist/kura/mesh_test.exs test/tuist/kura/registrations_test.exs test/tuist/kura/self_hosted_clients_test.exs test/tuist/oauth/introspection_test.exs test/tuist_web/controllers/internal/kura_mesh_controller_test.exs test/tuist_web/controllers/internal/kura_usage_controller_test.exs test/tuist_web/live/cache_live_test.exs test/tuist/kura/provisioner/kubernetes_controller_test.exs test/tuist/accounts_test.exs:3980 test/tuist/accounts_test.exs:3997`: 95 tests, 0 failures
- `mise x go@1.25.0 -- go test ./...`

Approved.

PR #11324, opened 2026-06-17
fix(infra): give runner nodes host-memory headroom and scrape their node-exporter

fortmarek:
## What

Two infrastructure changes that stop the recurring "self-hosted runner lost communication with the server" failures on the Linux CI runner fleet, and close the observability gap that hid the cause.

1. **Reserve host-memory headroom on the bare-metal runner nodes** so the kube-scheduler stops bin-packing Kata microVMs to ~100% of physical RAM.
2. **Scrape the Linux node-exporter** so the runner hosts actually report host memory / CPU / PSI to Grafana.

## Why — root cause

The annotation "lost communication" has bitten us before, and the prior remedy (route heavy server jobs to the larger `tuist-linux-large` 4 vCPU / 16 GB shape) addressed *guest-internal* OOM — a single VM's own compile exceeding 8 GB. This incident is a different, host-level failure mode that the larger shape does not fix (and arguably worsens).

Walking the telemetry for the failing `main` run:

- The failing `Test` and `Gettext` jobs (both `tuist-linux-large`, different runners) went silent within ~15 s of each other and were both scheduled onto the **same** bare-metal node. GitHub's "lost communication" fires ~10 min after the last heartbeat, so both runners stopped heartbeating at essentially the same instant — a host event, not independent per-VM OOM.
- That node stayed `Ready` the whole time (no reboot, no NotReady), yet its journal showed a QEMU process killed by **SIGBUS** (`ANOM_ABEND ... qemu-system-x86 ... sig=7`) and multiple Kata `sandbox stopped unexpectedly: failed to ping hypervisor process: guest failure: internal-error` in the same window. There was **no** kernel cgroup-OOM line — the classic signature of the host being unable to back the guests' memory.
- Quantitatively: summed pod memory **requests** on that node peaked at **248 GiB against 251 GiB physical (~99%)** seconds before the die-off, with **10 concurrent `4vcpu-16gb` VMs** plus the warm pool packed on. The kube-scheduler bin-packs by pod request, which equals the guest RAM only — the QEMU + virtiofsd + kata-shim overhead and the host daemons (kubelet, containerd, Cilium, Alloy) live **outside** that request, so the node was physically oversubscribed even though the scheduler thought it was at 99%.

Two root contributors:

- **The bare-metal runner nodes run with kubeadm defaults — no `system-reserved` / `kube-reserved` / meaningful eviction threshold — so allocatable == capacity.** Nothing stopped the scheduler from filling guest RAM to the last GiB.
- **The kata-qemu RuntimeClass `overhead.podFixed` was 100Mi**, an idle-VMM estimate that undercounts the busy-VMM footprint, so even the overhead the scheduler *did* account for was too small.

## What changed

**`infra/k8s/clusters/bare-metal.yaml`** — add `kubeletExtraArgs` to the shared bare-metal runner bootstrap:

```yaml
kubeletExtraArgs:
  - { name: system-reserved, value: "cpu=1,memory=2Gi" }
  - { name: kube-reserved,   value: "cpu=1,memory=2Gi" }
  - { name: eviction-hard,   value: "memory.available<10%,nodefs.available<10%,imagefs.available<15%,nodefs.inodesFree<5%" }
```

- The memory eviction threshold is a **percentage** on purpose: the bootstrap template is shared across staging / canary / production, whose boxes may differ in size, so a fixed GiB reservation would mis-size the smaller envs. 10% scales with the host and also gives a graceful evict-with-event backstop before the host hits the SIGBUS wall. `system`/`kube-reserved` are flat because daemon cost is roughly node-size-independent. The disk thresholds are kept at kubelet defaults so the `--eviction-hard` override doesn't drop disk-pressure eviction.
- Net effect on the 251 GiB production boxes: allocatable drops to **~222 GiB (~88%)** — the headroom that was missing.

**`infra/helm/tuist/templates/kata-qemu.yaml`** — raise `overhead.podFixed` from `100Mi` / `50m` to `512Mi` / `250m`, so the scheduler accounts for the real per-VM busy-VMM cost. Applies to newly-scheduled pods on the next chart deploy; no node roll needed.

**`infra/helm/k8s-monitoring/values.yaml`** — add an explicit Alloy scrape of the Linux node-exporter DaemonSet (`job="tuist-linux-node-exporter"`). node-exporter is deployed and running on the runner nodes, but the chart's built-in `hostMetrics` feature ships **no** `node_*` series to Grafana Cloud for any Linux k8s node — only the macOS fleet and the standalone cache hosts (`integrations/unix`) have them. So the runner hosts had zero host memory / CPU / PSI signal, which is exactly why this host-level die-off was invisible in metrics and only the node journal told the story. The new job mirrors the proven `tuist-macos-node-exporter` pattern the repo already uses where the built-in path didn't deliver; `instance`/`node` are set to the node name so dashboards survive a CAPI machine rename.

- **Just grow the fleet** (order more bare metal): real capacity lever and planned as a follow-up, but it doesn't stop a single node from being packed to 99% — the headroom fix is the actual guard.
- **Fixed GiB reservation**: unsafe in the shared template across differently-sized envs; the percentage eviction threshold scales automatically.
- **Fix the built-in `hostMetrics` feature instead of an explicit scrape**: the feature is enabled but ships nothing and couldn't be root-caused without rendering the upstream chart against the live cluster; the explicit scrape is the same call the authors already made for macOS and is the reliable path. `hostMetrics` is left enabled (documented) and the distinct job name avoids any future collision.

## Impact

- Runner nodes will schedule fewer concurrent microVMs (prod allocatable ~88% vs ~100%), trading a little peak density for stability. A 3rd bare-metal host is planned to recover the capacity.
- Host-level memory / CPU / PSI metrics for the runner fleet become queryable and alertable for the first time.

## Rollout notes

- **The `bare-metal.yaml` change requires a roll of the bare-metal MachineDeployment**, which reprovisions the Hetzner Robot hosts (rescue → installimage → kubeadm-join). On the current 2-node fleet this is disruptive; the plan is to order a 3rd `tuist-bm-production-*` host and bump `replicas` first, then roll, so capacity doesn't dip. The kata-overhead and node-exporter changes are non-disruptive.
- **Post-deploy validation for the node-exporter scrape** (couldn't be run against the live cluster from here): confirm `up{job="tuist-linux-node-exporter"}` and `node_memory_MemAvailable_bytes{job="tuist-linux-node-exporter", node=~"bm-tuist-runners-linux.*"}` return data after the k8s-monitoring deploy.

## Validation run here

- YAML parses for both manifests; `yq` confirms the `kubeletExtraArgs` structure.
- `helm template` renders kata-qemu with the new overhead.
- `helm dependency update` + `helm template` of k8s-monitoring (the CI render path) succeeds (3624 lines) and the `tuist_linux_node_exporter` discovery/relabel/scrape blocks are present in the rendered Alloy config.

### DIFF
--- a/infra/helm/k8s-monitoring/values.yaml
+++ b/infra/helm/k8s-monitoring/values.yaml
@@ -382,6 +382,59 @@ k8s-monitoring:
           scrape_timeout = "10s"
         }
 
+        // Linux k8s node-exporter. `hostMetrics` (the chart feature meant
+        // to scrape the node-exporter DaemonSet) is enabled above but ships
+        // no Linux node series to Grafana Cloud — only the macOS job and the
+        // standalone cache hosts' `integrations/unix` job carry node_*
+        // metrics — so the bare-metal runner Nodes have zero host
+        // memory / CPU / PSI signal. That blind spot is why a host-level
+        // microVM die-off (QEMU SIGBUS under RAM pressure) was invisible in
+        // metrics and only the Node journal showed it. Scrape the DaemonSet
+        // Pods directly, same explicit pattern as the macOS node-exporter
+        // above. The Pods are hostNetwork (:9100), so Pod IP == host IP;
+        // kubelet/cAdvisor already scrape these Nodes over the same path, so
+        // reachability is proven. instance = Node name so dashboards survive
+        // a CAPI machine rename. Distinct job name (not `integrations/unix`)
+        // so this never collides with the cache fleet or a future revived
+        // `hostMetrics`.
+        discovery.kubernetes "tuist_linux_node_exporter" {
+          role = "pod"
+          namespaces {
+            names = ["observability"]
+          }
+        }
+
+        discovery.relabel "tuist_linux_node_exporter" {
+          targets = discovery.kubernetes.tuist_linux_node_exporter.targets
+          rule {
+            source_labels = ["__meta_kubernetes_pod_name"]
+            regex = "k8s-monitoring-node-exporter-.*"
+            action = "keep"
+          }
+          rule {
+            source_labels = ["__meta_kubernetes_pod_host_ip"]
+            regex = "(.+)"
+            replacement = "$1:9100"
+            target_label = "__address__"
+          }
+          rule {
+            source_labels = ["__meta_kubernetes_pod_node_name"]
+            target_label = "instance"
+          }
+          rule {
+            source_labels = ["__meta_kubernetes_pod_node_name"]
+            target_label = "node"
+          }
+        }
+
+        prometheus.scrape "tuist_linux_node_exporter" {
+          targets = discovery.relabel.tuist_linux_node_exporter.output
+          forward_to = [prometheus.remote_write.grafana_cloud_metrics.receiver]
+          job_name = "tuist-linux-node-exporter"
+          scrape_interval = "30s"
+          scrape_timeout = "10s"
+        }
+
         // CloudNativePG instance metrics. Managed clusters don't install
         // prometheus-operator CRDs, and CNPG doesn't add prometheus.io/*
         // annotations to instance pods, so annotationAutodiscovery can't
--- a/infra/helm/k8s-monitoring/values.yaml
+++ b/infra/helm/k8s-monitoring/values.yaml
@@ -118,10 +118,30 @@ k8s-monitoring:
     enabled: true
     collector: alloy-singleton
 
-  # Node Exporter for node-level CPU / memory / disk / network metrics.
+  # Node-level CPU / memory / disk / network / PSI metrics from the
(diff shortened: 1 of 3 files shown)


pepicrft (approved):
Approving. I validated the rendered k8s-monitoring overlays, the Tuist chart RuntimeClass output with deploy-time image values stubbed, YAML parsing for the bare-metal CAPI substrate, and the mgmt apply path for immutable template changes.

One non-blocking thing I would tighten before merging: the custom Linux node-exporter scrape currently forwards raw node_exporter output straight to Grafana Cloud. I would either use the upstream hostMetrics.linuxHosts path with the extra PSI series included, or add an explicit relabel keep/drop block so we only ingest the runner host signals we actually need.

PR #11325, opened 2026-06-17
fix(cli): support SwiftPM static library artifact bundles

### LINKED ISSUE #11322: tuist generate fails on SE-0482 `staticLibrary` artifactbundle: treated as xcframework, "doesn't contain an Info.plist"
### Summary

`tuist generate` aborts when a SwiftPM dependency vendors a native backend as an **SE-0482 `staticLibrary` artifactbundle** (a `.binaryTarget` whose `info.json` declares `"type": "staticLibrary"`). Tuist routes the `.artifactbundle` through its **xcframework** code path and fails looking for an `Info.plist` that `staticLibrary` artifactbundles don't have (they use `info.json`).

Xcode's native SwiftPM integration handles the same package fine — only Tuist's package integration trips on it.

### Error

```
✖ Error
  The .xcframework at path /…/Tuist/.build/artifacts/swift-hf-api/HFAPIRust/HFAPIRust.artifactbundle/Info.plist doesn't contain an Info.plist. It's possible that the .xcframework was not generated properly or that got corrupted. Please, double check with the author of the framework.
```

Note the malformed path — `…/HFAPIRust.artifactbundle/Info.plist` — Tuist appends `Info.plist` to the bundle root, expecting an `.xcframework` layout.

(With a warm `.build` cache the error is masked because Tuist skips re-ingesting the already-extracted artifact; it reproduces on a clean/empty cache.)

### Reproduction

Real-world package: [`DePasqualeOrg/swift-hf-api`](https://github.com/DePasqualeOrg/swift-hf-api) `0.4.2`, whose Rust backend is shipped as a `staticLibrary` artifactbundle:

```swift
// swift-hf-api/Package.swift (excerpt)
.binaryTarget(
    name: "HFAPIRust",
    url: "https://github.com/DePasqualeOrg/swift-hf-api/releases/download/hfapi-rust-0.4.2/HFAPIRust-0.4.2.artifactbundle.zip",
    checksum: "d1975558aee9fb32cb7242cb2f573427b0081aa3ee5374b0976a91b60a428e04"
)
```

The artifactbundle `info.json` (no `Info.plist`):

```json
{
  "schemaVersion": "1.0",
  "artifacts": {
    "HFAPIRust": {
      "type": "staticLibrary",
      "version": "0.4.2",
      "variants": [
        { "path": "apple-macos/libhf_api_rust.a",          "supportedTriples": ["arm64-apple-macosx", "x86_64-apple-macosx"], "staticLibraryMetadata": { "headerPaths": ["include"], "moduleMapPath": "include/module.modulemap" } },
        { "path": "apple-ios-device/libhf_api_rust.a",      "supportedTriples": ["arm64-apple-ios"], "staticLibraryMetadata": { … } },
        { "path": "apple-ios-simulator/libhf_api_rust.a",   "supportedTriples": ["arm64-apple-ios-simulator", "x86_64-apple-ios-simulator"], "staticLibraryMetadata": { … } },
        { "path": "linux-x86_64/libhf_api_rust.a",          "supportedTriples": ["x86_64-unknown-linux-gnu"], "staticLibraryMetadata": { … } },
        { "path": "linux-aarch64/libhf_api_rust.a",         "supportedTriples": ["aarch64-unknown-linux-gnu"], "staticLibraryMetadata": { … } }
      ]
    }
  }
}
```

Steps:

1. In `Tuist/Package.swift` add the dependency:
   ```swift
   .package(url: "https://github.com/DePasqualeOrg/swift-hf-api", .upToNextMajor(from: "0.4.2"))
   ```
2. Consume a product from a target, e.g. `.external(name: "HFAPI")`.
3. `tuist install`
4. `tuist generate` (on a clean cache, e.g. after removing `Tuist/.build`).

### Expected

Tuist ingests SE-0482 `staticLibrary` artifactbundles — parse `info.json`, pick the slice whose `supportedTriples` match the build, and link the `.a` with its `moduleMapPath` / `headerPaths` — the same way Xcode's native SwiftPM integration does.

### Actual

Generation aborts, treating the `staticLibrary` artifactbundle as a (broken) `.xcframework`.

### Workaround

Keep the package **Xcode-native**: declare it in `Project.swift`'s `packages:` and consume via `.package(product:)` instead of Tuist's `.external` integration.

### Environment

- Tuist **4.200.5**
- macOS (Apple Silicon), Xcode 26.x
- SwiftPM artifactbundle schema `1.0`, artifact `type: staticLibrary` (SE-0482)

pepicrft:
## Summary
- Teach `PackageInfoMapper` to recognize SE-0482 `staticLibrary` `.artifactbundle` artifacts instead of treating every binary artifact path as an `.xcframework`.
- Materialize a lightweight XCFramework-compatible wrapper for Apple static-library variants under SwiftPM scratch derived state: `tuist-derived/XCFrameworks/<package>/<target>-<source-fingerprint>.xcframework`.
- Symlink static archives, headers, and module maps from the derived wrapper back to the immutable artifact bundle contents, and emit the XCFramework plist keys Xcode expects.

Fixes #11322.

## Why
`swift-hf-api` ships its Rust backend as an SE-0482 `staticLibrary` artifact bundle. SwiftPM resolves that artifact correctly, but Tuist previously passed the resolved `.artifactbundle` path into the existing XCFramework path, which made generation fail before the project could be produced.

This matters for packages that adopt static library artifact bundles instead of wrapping their Apple binaries as XCFrameworks. They should be usable from Tuist projects without package authors having to republish binaries in a Tuist-specific layout.

## Root Cause
Tuist's SwiftPM mapping code assumed binary artifact paths resolved from the workspace state were XCFrameworks. That was true for the older binary target shape, but SE-0482 adds artifact bundles whose `info.json` can describe static libraries for multiple triples.

When one of those artifact bundle paths reached `XCFrameworkMetadataProvider`, Tuist looked for `Info.plist` inside the `.artifactbundle` and failed with the reported error because the artifact bundle uses `info.json` instead.

## Solution
The mapper now detects static-library artifact bundles by reading `info.json`, filters their variants down to Apple triples, maps those triples to XCFramework library metadata, and returns a generated `.xcframework` path to the rest of the graph.

The generated wrapper is intentionally only a compatibility layer for Xcode and the existing Tuist graph model. Xcode still expects a physical `.xcframework` root with `LibraryIdentifier` directories and relative `LibraryPath` entries, so modeling this purely as arbitrary external paths would bypass the XCFramework contract.

The wrapper is treated as Tuist derived state rather than mutating SwiftPM-owned package or artifact folders. Production mapping passes the SwiftPM scratch-derived location into the mapper, and test-only fallback infers that same location for artifacts under `.../artifacts/...`. The wrapper name includes a source fingerprint derived from the artifact bundle path and `info.json` contents, so artifact updates for the same package and target do not reuse stale generated wrappers. To avoid duplicating potentially large static archives, the wrapper symlinks the archive, headers, and module map back to the artifact bundle contents.

I also updated `XCFrameworkInfoPlist` encoding to include the plist fields produced by `xcodebuild -create-xcframework`. Without those top-level keys, a minimal plist can be decoded by Tuist but rejected by Xcode's `ProcessXCFramework` step.

## User Impact
Projects depending on packages like `swift-hf-api` can run `tuist generate` successfully when the package contains Apple-compatible static library artifact bundle variants.

Existing XCFramework binary targets keep using the previous flow. Non-Apple artifact variants are ignored because they cannot be represented in an Xcode project.

## Validation
- `swift build --replace-scm-with-registry --target TuistLoader`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistLoaderTests/PackageInfoMapperTests -only-testing TuistLoaderTests/SwiftPackageManagerGraphLoaderTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `xcodebuild -scheme Probe -destination 'generic/platform=macOS' -derivedDataPath DerivedData ARCHS=arm64 ONLY_ACTIVE_ARCH=YES build` on a throwaway Swift package whose XCFramework wrapper symlinked its library and headers to an artifact-bundle-like layout
- `mise run cli:lint --fix`
- `git diff --check`

### DIFF
--- a/cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift
+++ b/cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift
@@ -4,6 +4,7 @@ import Mockable
 import Path
 import ProjectDescription
 import TSCUtility
+import TuistConstants
 import TuistCore
 import TuistLogging
 import TuistRootDirectoryLocator
@@ -84,12 +85,13 @@ public enum PackageType {
     case external(
         origin: ExternalOrigin = .remote,
         artifactPaths: [String: AbsolutePath],
-        packagePrebuilts: [String: [String: SwiftPackageManagerPrebuilt]] = [:]
+        packagePrebuilts: [String: [String: SwiftPackageManagerPrebuilt]] = [:],
+        derivedXCFrameworksPath: AbsolutePath? = nil
     )
 
     fileprivate var includesTestTargets: Bool {
         switch self {
-        case .local, .external(origin: .local, artifactPaths: _, packagePrebuilts: _):
+        case .local, .external(origin: .local, artifactPaths: _, packagePrebuilts: _, derivedXCFrameworksPath: _):
             return true
         case .external:
             return false
@@ -97,14 +99,14 @@ public enum PackageType {
     }
 
     fileprivate var isLocalExternal: Bool {
-        if case .external(origin: .local, artifactPaths: _, packagePrebuilts: _) = self {
+        if case .external(origin: .local, artifactPaths: _, packagePrebuilts: _, derivedXCFrameworksPath: _) = self {
             return true
         }
         return false
     }
 
     fileprivate var isRemoteExternal: Bool {
-        if case .external(origin: .remote, artifactPaths: _, packagePrebuilts: _) = self {
+        if case .external(origin: .remote, artifactPaths: _, packagePrebuilts: _, derivedXCFrameworksPath: _) = self {
             return true
         }
         return false
@@ -114,7 +116,7 @@ public enum PackageType {
         switch self {
         case .local:
             return [:]
-        case let .external(origin: _, artifactPaths: _, packagePrebuilts: packagePrebuilts):
+        case let .external(origin: _, artifactPaths: _, packagePrebuilts: packagePrebuilts, derivedXCFrameworksPath: _):
             return packagePrebuilts
         }
     }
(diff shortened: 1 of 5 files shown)


cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift
```diff
@@ -1051,6 +1081,194 @@ public struct PackageInfoMapper: PackageInfoMapping {
...
+        }
+
+        guard let derivedXCFrameworksPath = derivedXCFrameworksPath
+            ?? Self.derivedXCFrameworksPath(containingArtifact: artifactBundlePath)
+        else { return nil }
+        let xcframeworkPath = derivedXCFrameworksPath
+            .appending(component: Self.sanitize(targetName: packageName))
+            .appending(component: "\(Self.sanitize(targetName: targetName)).xcframework")
+        let infoPlistPath = xcframeworkPath.appending(component: "Info.plist")
+        if try await fileSystem.exists(infoPlistPath) {
```
fortmarek, line 1129: This early return can leave stale generated wrappers behind. The derived path only includes the package and target name (`tuist-derived/XCFrameworks/<package>/<target>.xcframework`), so after a package/artifact update with the same package and target, we reuse the existing `Info.plist` and symlink tree even if the archive path, headers, module map, supported triples, or artifact contents changed. That can make `tuist generate` keep linking stale or broken files until the user manually deletes derived state. Could we regenerate the wrapper each time for artifact bundles, or persist/check a small source fingerprint that covers the artifact bundle path plus `info.json` contents before reusing it?

fortmarek (commented):
One issue found around stale generated XCFramework wrappers.

fortmarek (approved):
otherwise, looks good 👍

cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift
pepicrft: Fixed in 50eedca. The generated wrapper path now includes a source fingerprint derived from the artifact bundle path and info.json contents, so a package or artifact update with the same package and target name gets a new derived XCFramework instead of reusing the previous one. I also extended PackageInfoMapperTests to mutate info.json and assert that the resolved dependency points at the new wrapper path.

PR #11327, opened 2026-06-17
feat(kura): enable runner-cache Kura nodes in production and canary

fortmarek:
## What changed

Extends the runner-cache Kura node setup from [#10982](https://github.com/tuist/tuist/pull/10982) (which enabled it for staging only) to **production** and **canary**.

- `server/lib/tuist/kura/regions.ex` — adds two private region specs, `hetzner-production-runners` (`cluster_id: "production"`) and `hetzner-canary-runners` (`cluster_id: "canary"`), mirroring the existing `hetzner-staging-runners`. Both pin to their cluster's `kura` node pool, `hcloud-volumes`, single replica, `ClusterIP`-only (no public host/ingress/certificate).
- `infra/helm/tuist/values-managed-production.yaml` — adds `hetzner-production-runners` to `TUIST_KURA_AVAILABLE_REGIONS`.
- `infra/helm/tuist/values-managed-canary.yaml` — adds `hetzner-canary-runners` to `TUIST_KURA_AVAILABLE_REGIONS`.
- `server/test/tuist/kura/regions_test.exs` — extends coverage to the new regions, plus a guard that each Hetzner runner-cache region pins a node pool that actually exists in its cluster manifest.

## Why

Accounts with runner profiles get an in-cluster Kura build cache co-located with their runner fleet, so build-cache traffic stays on the cluster's internal Service DNS instead of round-tripping the public ingress. Staging proved the topology end-to-end (#10982); this brings the same benefit to the environments that actually serve customer runners.

## Why this approach

Each environment's server reconciles `KuraInstance`s into **its own** workload cluster (the provisioner uses the in-cluster Kubernetes API; `cluster_id` is only an instance-name/audit field). Production and canary clusters already carry the `kura` and `runners-linux` node pools, and Linux fleets pass `fleet_on_cluster_network?`, so the staging topology (Kura cache node + Linux runners in the same Hetzner cluster, reached over `*.svc.cluster.local`) replicates cleanly with no infra changes.

The existing `scw-fr-par-runners` Scaleway region is deliberately **not** the lever here: its `kura-scw-fr-par` pool lives in a different cluster than the Hetzner prod runners, and the dispatch-time cluster-locality check it depends on (the PR's Phase 3) has not landed, so it would never schedule in the Hetzner prod cluster. This is documented in the region catalog comment.

## Rollout / reviewer notes

- **No manual CRD step.** The `spec.private` field is already committed to `infra/helm/tuist/crds/kura.tuist.dev_kurainstances.yaml` (by #10982). The deploy pipeline's `kubectl apply CRDs` step in `server-deployment.yml` re-applies `crds/` on every run, before the helm upgrade, for each leg of the production cascade (canary + production). So the field reaches both clusters automatically on their next deploy. (The "one-time manual apply" mentioned in #10982's description was the author seeding staging out-of-band for live testing, not a pipeline requirement.)
- **Provisioning scope.** Once enabled, the RunnerCache reconciler provisions a single-replica Kura node on the `kura` pool for every account that has runner profiles **and** the `:runners` FunWithFlags gate explicitly enabled. Worth a glance at how many accounts that is in prod/canary before merging.

## Validation

- `mix test test/tuist/kura/regions_test.exs` — 25 tests, 0 failures
- `mix test test/tuist/kura/provisioner/kubernetes_controller_test.exs` — 23 tests, 0 failures
- `mix format --check-formatted` clean on the changed files

### DIFF
--- a/infra/helm/tuist/values-managed-canary.yaml
+++ b/infra/helm/tuist/values-managed-canary.yaml
@@ -68,8 +68,13 @@ server:
       value: "30"
     - name: TUIST_OTEL_EXPORTER_OTLP_ENDPOINT
       value: http://k8s-monitoring-alloy-receiver.observability.svc.cluster.local:4317
+    # `hetzner-canary-runners` is the private (runner-cache) region:
+    # customers can't select it, the RunnerCache reconciler provisions
+    # nodes on this cluster's `kura` pool for accounts with runner
+    # profiles, and runner dispatch routes their build-cache traffic to
+    # it in-cluster.
     - name: TUIST_KURA_AVAILABLE_REGIONS
-      value: eu-central
+      value: eu-central,hetzner-canary-runners
 
 objectStorage:
   external:
--- a/infra/helm/tuist/values-managed-production.yaml
+++ b/infra/helm/tuist/values-managed-production.yaml
@@ -31,9 +31,13 @@ server:
       value: http://k8s-monitoring-alloy-receiver.observability.svc.cluster.local:4317
     # Exposes the managed Kura regions in account settings. Regions now
     # reconcile inside this workload cluster; per-region Kura node pools
-    # decide placement.
+    # decide placement. `hetzner-production-runners` is the private
+    # (runner-cache) region: customers can't select it, the RunnerCache
+    # reconciler provisions nodes on this cluster's `kura` pool for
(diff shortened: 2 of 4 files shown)


fortmarek:
Closing as we're moving both Linux runners and Kura nodes to Scaleway

PR #11330, opened 2026-06-17
DO NOT MERGE — red-before proof for #11329 (fix reverted, e2e test kept)

fortmarek:
Temporary verification PR. The suite-granularity Swift Testing fix from #11329 is **reverted** here, but the new acceptance test (`TestAcceptanceTestShardSuiteGranularityWithSwiftTesting`) and its fixture are kept. The acceptance run should **fail** that test — proving it catches the zero-tests regression before the fix. Will be closed once the red result is captured.


fortmarek:
Closing — not valid red-before evidence. The shard-1 failure was a Tigris `bundle.zip` download timeout (the separate pre-existing flake), not the e2e test; and a Swift Testing test gets silently dropped by the buggy outer suite-granularity sharding rather than failing. The definitive red-before/green-after proof is the local mechanism A/B on #11329.

PR #11338, opened 2026-06-17
feat: CNPG per-query latency observability (pg_stat_statements metrics + dashboard panels)

fortmarek:
> Draft. Stacked on #11336 (base is `fix/server-oban-jobs-bloat`) so it inherits the dashboard fixes; retarget to `main` once that merges.

## Why

During the `oban_jobs` bloat incident, the one signal we never had was **per-query latency**. The CNPG cluster already preloads `pg_stat_statements` (the values comment even says "for per-query latency observability"), but the extension was never created and nothing exported it, so `SELECT … FROM pg_stat_statements` just errored. We had to infer slow queries from `pg_stat_user_tables` scan counts. This wires that observability up.

## What

Behind `postgresql.cnpg.queryStats.enabled` (on in managed envs, off in base/preview where the library isn't preloaded):

1. **`infra/cnpg/pg-stat-statements.sql`** — re-runnable `CREATE EXTENSION pg_stat_statements`, run once against the **`postgres`** database (the view is instance-global; the exporter reads it from the maintenance DB). Documented in `infra/cnpg/README.md`.
2. **Custom CNPG monitoring query** (ConfigMap → `monitoring.customQueriesConfigMap`) that buckets `pg_stat_statements` read/write/other for the app database and exports `cnpg_tuist_query_stats_{calls,total_exec_time_ms,rows}`. 3 series, low cardinality. Metrics scrape via the existing Alloy autodiscovery (no PodMonitor needed).
3. **Two dashboard panels**: `Query Latency (read/write)` (mean ms = `rate(total_exec_time_ms)/rate(calls)`) and `Query Throughput (read/write)` (ops), both `op`-split and scoped to the primary instance (same primary-scoping as the rate-panel fix in #11336).

## Rollout / ordering

- `shared_preload_libraries` already includes `pg_stat_statements`, so **no Postgres restart** is needed — adding the custom query is a live exporter-config reload.
- The extension must exist before the exporter query returns data. Until `pg-stat-statements.sql` is run once per cluster, the exporter query errors and the metric is simply absent (panels show "No data"); not harmful. Run it per the `infra/cnpg/README.md` runbook (it's a DDL write, so it's a deliberate operator step, not something the pipeline does).

## Validation

- `helm template` against `values-managed-production.yaml` renders the ConfigMap and the `customQueriesConfigMap` reference with matching names.
- Parsed the rendered ConfigMap's embedded `queries` block as YAML to confirm CNPG can consume it.
- Dashboard JSON validated (round-trip stable; panels parse; non-overlapping gridPos).
- Note: the live metric could not be exercised end-to-end yet because creating the extension is a prod DDL write (out of scope for an automated run) and the chart isn't deployed.

## Follow-up captured from the incident

The dashboard's `Database Size` stat and `Connections` count still sum across the 3 instances (same over-aggregation class as the gauge/rate fixes in #11336); worth the same primary-scoping in a later pass.

### DIFF
--- a/infra/cnpg/README.md
+++ b/infra/cnpg/README.md
@@ -8,6 +8,7 @@ The chart-rendered `Cluster` CR ([`infra/helm/tuist/templates/postgresql-cnpg.ya
 
 - [`tuist-processor-grants.sql`](./tuist-processor-grants.sql) — `oban_jobs` / `oban_peers` write grants for the `tuist_processor` role. The role itself is created declaratively by CNPG via `managed.roles[]`; this file adds the per-table privileges the worker needs.
 - [`tuist-ops-ro-grants.sql`](./tuist-ops-ro-grants.sql) — `CONNECT` on the application database for the `tuist_ops_ro` role, plus an explicit `REVOKE` of write privileges on `public` (defense-in-depth against a future grant-by-default change in Postgres widening `pg_read_all_data`). The role is for ad-hoc operator psql access; the `/ops/db` LiveView uses Tuist.Repo and enforces read-only at the app layer (see `Tuist.Ops.Database.execute/2`).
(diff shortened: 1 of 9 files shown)


fortmarek:
Folding these commits into #11336 instead of stacking, per review.

PR #11344, opened 2026-06-18
feat(cli): resolve Bazel remote cache to the Kura gRPC endpoint

esnunes:
## What changed

`tuist bazel setup` generates `.bazelrc.tuist` with a `--remote_cache` flag pointing at a Kura cache node. `CacheURLStore` returns the node's **HTTP(S)** URL, but Kura serves the Bazel REAPI over **gRPC** on a separate `grpc.`-prefixed host. The generated config therefore pointed Bazel at a host that doesn't serve gRPC.

This change resolves the remote cache to the correct gRPC endpoint entirely on the client, in `BazelSetupCommandService`:

- **gRPC conversion of the resolved endpoint** — the scheme becomes `grpc` (plaintext, e.g. local development deployments) or `grpcs` (TLS), and the host is prefixed with `grpc.`. For example, `https://acme-eu-central-1.kura.tuist.dev` becomes `grpcs://grpc.acme-eu-central-1.kura.tuist.dev`. The `http→grpc` / `https→grpcs` scheme conversion already existed; the **`grpc.` host prefix is the new piece**.
- **`TUIST_CACHE_GRPC_ENDPOINT` override** — when set, it short-circuits resolution and is used verbatim as `--remote_cache` (no scheme conversion, no `grpc.` prefix). This mirrors the existing `TUIST_CACHE_ENDPOINT` HTTP override and is useful for local/self-hosted gRPC endpoints.

## Why this approach

The endpoint-to-gRPC mapping is deterministic, so it can live next to where `.bazelrc.tuist` is written rather than being threaded through the server's `/api/cache/endpoints` response. Keeping it client-side means **no server change and no API/OpenAPI contract change**, and the override env var gives an escape hatch without round-tripping through the server.

## Notable detail: credential helper host must track the cache host

`--credential_helper` is now registered for the **same `grpc.`-prefixed host** as `--remote_cache`. Bazel matches the credential helper against the host it actually connects to, so the two must stay in sync — otherwise the prefixed gRPC host would have no registered helper and authentication would fail. Both the resolved-endpoint path and the override path return a single `(remoteCache, credentialHelperHost)` pair to guarantee this.

## User impact

`tuist bazel setup` now produces a `.bazelrc.tuist` whose `--remote_cache` actually reaches Kura's gRPC REAPI endpoint, so Bazel remote caching works against Kura nodes. Users can override the endpoint with `TUIST_CACHE_GRPC_ENDPOINT`.

## Validation

- `swift build --replace-scm-with-registry --target TuistBazelCommand` compiles cleanly.
- `swiftformat --lint` clean on the changed files.
- Updated the existing `BazelSetupCommandServiceTests` assertions for the `grpc.` prefix on both `--remote_cache` and `--credential_helper`, and added tests for the `TUIST_CACHE_GRPC_ENDPOINT` verbatim short-circuit and its invalid-URL error path. Note: this test target runs via the Xcode workspace on macOS CI (it is not an SPM target), so the suite was not executed locally — the production target's compilation and the format check were verified locally.

### DIFF
--- a/cli/Sources/TuistBazelCommand/BazelSetupCommandService.swift
+++ b/cli/Sources/TuistBazelCommand/BazelSetupCommandService.swift
@@ -55,26 +55,18 @@ public struct BazelSetupCommandService: BazelSetupCommandServicing {
             throw BazelSetupCommandServiceError.notAuthenticated
         }
 
-        let cacheURL = try await cacheURLStore.getCacheURL(for: serverURL, accountHandle: accountHandle)
-        guard let host = cacheURL.host else {
-            throw BazelSetupCommandServiceError.invalidCacheEndpoint(cacheURL.absoluteString)
-        }
-        let endpoint = if let port = cacheURL.port {
-            "\(host):\(port)"
-        } else {
-            host
-        }
-        // Plaintext cache endpoints (e.g. local development deployments) expose
-        // plaintext gRPC, while TLS endpoints terminate TLS for gRPC too.
-        let remoteCacheScheme = cacheURL.scheme == "http" ? "grpc" : "grpcs"
+        let (remoteCache, credentialHelperHost) = try await resolveRemoteCache(
+            serverURL: serverURL,
+            accountHandle: accountHandle
+        )
 
         let credentialHelperPath = try await createCredentialHelperScriptIfNeeded()
 
         let bazelrcPath = directoryPath.appending(component: ".bazelrc.tuist")
         let bazelrcContent = """
-        build --remote_cache=\(remoteCacheScheme)://\(endpoint)
+        build --remote_cache=\(remoteCache)
         build --remote_header=x-tuist-account-handle=\(accountHandle)
-        build --credential_helper=\(host)=\(credentialHelperPath.pathString)
+        build --credential_helper=\(credentialHelperHost)=\(credentialHelperPath.pathString)
         build --remote_instance_name=\(projectHandle)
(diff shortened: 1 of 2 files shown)


cli/Sources/TuistBazelCommand/BazelSetupCommandService.swift
```diff
@@ -84,12 +76,50 @@ public struct BazelSetupCommandService: BazelSetupCommandServicing {
...
+            throw BazelSetupCommandServiceError.invalidCacheEndpoint(cacheURL.absoluteString)
+        }
+        let grpcHost = "grpc.\(host)"
+        let endpoint = if let port = cacheURL.port {
+            "\(grpcHost):\(port)"
+        } else {
+            grpcHost
+        }
+        let scheme = cacheURL.scheme == "http" ? "grpc" : "grpcs"
+        return ("\(scheme)://\(endpoint)", grpcHost)
```
pepicrft, line 120: I'm not sure I'd go the convention way here, 'cause it creates an implicit dependency on the URL schema that can break. What if we make the server respond the right URL based on the client indicating whether they want to hit the HTTP vs the gRPC API?

esnunes: I will do a further investigation on this

esnunes:
Closed in favor of https://github.com/tuist/tuist/pull/11356

PR #11348, opened 2026-06-18
feat: add Slack-requested preview environments

pepicrft:
Resolves N/A

Adds Slack-requested preview environments that can be created and deleted through tuist-ops, deployed by GitHub Actions into the preview Kubernetes cluster, and cleaned up by TTL or idle-time sweeps.

### What changed

- Adds `/preview create` and `/preview delete` handling in tuist-ops, persists requests in `preview_requests`, renders Slack Block Kit cards, and dispatches GitHub Actions with the requested PR or commit.
- Adds the on-demand preview deployment workflow and extends preview sweeping so expired or idle previews uninstall Helm, delete the shared Kura instance, and remove the preview namespace.
- Adds preview Helm and cluster wiring for embedded ClickHouse, embedded object storage, wildcard preview TLS, preview node placement, and local Kind validation with dedicated preview and Kura workers.
- Adds shared preview Kura support: the server can advertise configured Kura endpoints for every account, and the Kura hook keeps tenant matching by default while allowing preview-scoped shared tenants only when explicitly enabled.
- Fixes private Kura instances so the controller skips public Ingress and cert-manager Certificates, and fixes controller image builds for the requested platform.

### How it works

- An operator requests a preview from Slack. tuist-ops validates the request, stores it, posts a Slack status card, and dispatches the preview workflow with the slug, PR or commit, TTL, requester, and optional Kura replica count.
- The workflow resolves the target ref, builds the server image, and installs the Tuist chart with the preview overlays. The preview release includes in-cluster Postgres, ClickHouse, and MinIO object storage so it does not point at ClickHouse Cloud or Tigris.
- When Kura is enabled, the workflow creates one preview-scoped `KuraInstance` in the `kura` namespace with a public `kura-<slug>.preview.tuist.dev` host. The server receives that URL through `TUIST_KURA_ENDPOINTS`.
- Server cache endpoint discovery returns the configured Kura endpoint for any account when the client requests Kura. This avoids creating per-account `kura_servers` rows for previews.
- The Kura Lua hook still authorizes every account and project request through Tuist grants. The only preview-specific relaxation is that `KURA_EXTENSION_TUIST_ALLOW_SHARED_TENANTS=1` lets one preview mesh serve multiple account handles inside the same preview instance.
- The sweep workflow reads namespace labels and annotations, deletes previews whose TTL expired or whose last-active timestamp is too old, and removes the matching shared Kura instance.

### Why

Previews are isolated at the instance level, so a single Kura mesh per preview is simpler than provisioning per-account Kura nodes inside that preview. It gives every account in the preview the same cache endpoint while preserving account and project authorization in the Kura hook. Embedding ClickHouse and object storage makes previews closer to staging, canary, and production behavior without sharing managed production-like data services.

### User and developer impact

- Operators can request, inspect, and delete preview environments from Slack.
- Preview environments carry their own analytics and object-storage dependencies.
- Kura in previews is instance-scoped and shared across accounts, not represented as account-specific managed Kura servers.
- Expiration and idle cleanup happen from GitHub Actions using Kubernetes labels and annotations.

### How to test locally

- `cd tuist-ops && mix format && mix test test/tuist_ops_web/controllers/slack_controller_test.exs test/jit/changesets_test.exs`
- `ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts "OK #{f}" }' .github/workflows/preview-ondemand-deploy.yml .github/workflows/preview-sweep.yml infra/helm/tuist/values-preview-kura.yaml infra/k8s/kind-preview.yaml`
- `bash -n infra/mise/tasks/helm/preview-up.sh && bash -n infra/mise/tasks/helm/preview-down.sh`
- `helm lint infra/helm/tuist -f infra/helm/tuist/values-preview.yaml -f infra/helm/tuist/values-preview-kura.yaml --set server.image.tag=latest --set server.ingress.enabled=true --set server.ingress.host=demo.preview.tuist.dev --set server.appUrl=https://demo.preview.tuist.dev --set kuraController.image.tag=latest --set kuraRuntime.image.tag=latest`
- `helm template ondemand-demo infra/helm/tuist -f infra/helm/tuist/values-preview.yaml -f infra/helm/tuist/values-preview-kura.yaml --set server.image.tag=latest --set server.ingress.enabled=true --set server.ingress.host=demo.preview.tuist.dev --set server.appUrl=https://demo.preview.tuist.dev --set kuraController.image.tag=latest --set kuraRuntime.image.tag=latest >/tmp/tuist-preview-ondemand-rendered.yaml`
- `helm dependency build infra/helm/tuist`
- `helm template preview-test infra/helm/tuist -f infra/helm/tuist/values-preview.yaml -f infra/helm/tuist/values-preview-kura.yaml --set server.image.tag=test --set server.ingress.enabled=true --set server.ingress.host=preview-test.preview.tuist.dev --set server.appUrl=https://preview-test.preview.tuist.dev --set kuraController.image.tag=test --set kuraRuntime.image.tag=test --set "server.kuraEndpointUrls[0]=https://kura-preview-test.preview.tuist.dev" >/tmp/tuist-preview-kura-render.yaml`
- `ruby -e 'content = File.read(ARGV.fetch(0)); %w[TUIST_KURA_ENDPOINTS TUIST_KURA_RUNTIME_IMAGE_TAG letsencrypt-cloudflare preview-test-tuist-clickhouse preview-test-tuist-object-storage].each { |needle| abort("missing #{needle}") unless content.include?(needle) }; abort("unexpected TUIST_KURA_AVAILABLE_REGIONS") if content.include?("TUIST_KURA_AVAILABLE_REGIONS"); puts "render assertions passed"' /tmp/tuist-preview-kura-render.yaml`
- `cd infra/kura-controller && mise exec go -- go test ./controllers`
- `cd kura && mise exec -- cargo fmt --check`
- `cd kura && mise exec -- cargo test tuist_hook`
- `cd server && MIX_ENV=test mise exec -- mix ecto.reset`
- `cd server && mise exec -- mix test test/tuist/environment_test.exs:290 test/tuist/environment_test.exs:297 test/tuist/accounts_test.exs:4220 test/tuist/accounts_test.exs:4233`
- `git diff --check`
- `git diff --cached --check`
- `rg -n "^(<<<<<<<|=======$|>>>>>>> )" .`
- Local Kind validation created a five-node `tuist-preview` cluster. The full server rollout reached dependency scheduling for ClickHouse, object storage, Postgres, cache, and server on the preview worker, but server readiness needs a valid `TUIST_LICENSE_KEY` or `OP_SERVICE_ACCOUNT_TOKEN`, which was not available locally. A controller-only Kura validation created a private two-replica `KuraInstance`; both runtime pods reached Ready on separate `role=kura` Kind workers.


.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,381 @@
...
+        run: |
+          kubectl label nodes -l '!node-role.kubernetes.io/control-plane,node.cluster.x-k8s.io/pool!=kura' role=preview --overwrite
+          kubectl taint nodes -l '!node-role.kubernetes.io/control-plane,node.cluster.x-k8s.io/pool!=kura' role=preview:NoSchedule --overwrite
+      - name: helm upgrade --install
+        run: |
+          args=(
+            "$RELEASE" "$HELM_CHART_PATH"
+            --namespace "$NAMESPACE" --create-namespace \
+            -f "$HELM_CHART_PATH/values-preview.yaml" \
+            -f "$HELM_CHART_PATH/values-preview-kura.yaml" \
```
fortmarek: This makes each on-demand preview install the Kura controller chart into the fixed `kura` namespace. The chart renders shared objects such as `Namespace/kura` and `Secret/kura-shared-secrets` from the release, so the second `ondemand-*` preview will collide on Helm ownership, and uninstalling one preview can also delete shared controller resources used by another. I think the controller/shared secrets need to be installed once at the platform level, or made preview-scoped, before multiple Slack previews can coexist.

.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,381 @@
...
+      - name: Load kubeconfig
+        run: |
+          mkdir -p "$(dirname "$KUBECONFIG")"
+          echo "${{ secrets.KUBECONFIG }}" | base64 -d > "$KUBECONFIG"
+          chmod 600 "$KUBECONFIG"
+      - name: helm uninstall + delete namespace
+        run: |
+          set -euo pipefail
+          if helm status "$RELEASE" --namespace "$NAMESPACE" >/dev/null 2>&1; then
+            helm uninstall "$RELEASE" --namespace "$NAMESPACE" --wait --timeout 5m || true
```
fortmarek: This uninstalls the Helm release before deleting the `KuraInstance` below. Because the same release owns the Kura controller and `KuraInstance` has a finalizer, deleting the CR after the controller is gone can leave it stuck terminating. Please delete and wait for `${RELEASE}-kura` before uninstalling the release, or keep the controller independent from the preview release.

.github/workflows/preview-deploy.yml
```diff
@@ -318,7 +319,21 @@ jobs:
           echo "${{ secrets.KUBECONFIG }}" | base64 -d > "$KUBECONFIG"
           chmod 600 "$KUBECONFIG"
 
+      - name: Temporarily remove preview taint for platform hooks
+        run: |
+          kubectl taint nodes -l '!node-role.kubernetes.io/control-plane' role=preview:NoSchedule- || true
```
fortmarek: Now that `cluster-preview.yaml` adds a dedicated `node.cluster.x-k8s.io/pool=kura` MachineDeployment, this selector also matches Kura workers. The later re-taint step puts `role=preview:NoSchedule` on every non-control-plane node, while controller-managed Kura pods only have a node selector and no matching toleration, so a regular PR preview deploy can make the Kura pool unschedulable for Slack preview Kura pods. The on-demand workflow already excludes `pool=kura`; this workflow should do the same.

fortmarek (commented):
I found a few blocking/runtime-risk issues in the preview/Kura workflow wiring.

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Fixed in d9c0f9e — I went with the preview-scoped option. The on-demand workflow now passes `--set kuraController.namespace=$NAMESPACE` to helm upgrade, so each preview owns its own kura controller Deployment, `kura-shared-secrets` Secret, Namespace, and ClusterRole bindings (the chart already uses the release name in `tuist.componentName`, so cluster-scoped names don't collide between previews). The KuraInstance is created in the per-preview namespace too. `helm template` for a slug `demo` renders the namespace as `preview-ondemand-demo` and the cluster-scoped resources as `ondemand-demo-tuist-kura-controller-*`, so two on-demand previews can coexist without Helm ownership fights, and uninstalling one no longer touches another's controller.

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Fixed in d9c0f9e. Both the on-demand delete job and the sweep workflow now delete `${RELEASE}-kura` with `--wait=true --timeout 5m` *before* `helm uninstall`. Combined with moving the controller into the per-preview namespace, the controller is still up when the CR's finalizer needs to drain the public Ingress + cert-manager Certificate, so the CR isn't left stuck terminating and the LE leaf cert isn't orphaned.

.github/workflows/preview-deploy.yml
pepicrft: Fixed in 5eb6c6c. Both the temporary-untaint step and the re-apply step in `preview-deploy.yml` now use `'!node-role.kubernetes.io/control-plane,node.cluster.x-k8s.io/pool!=kura'`, matching what the on-demand workflow already does, so the kura pool stays schedulable for Slack-preview Kura pods.

.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,445 @@
...
+            storageSize: 10Gi
+            extraEnv:
+              - name: KURA_CONTROL_PLANE_URL
+                value: http://${RELEASE}-tuist-server.${NAMESPACE}.svc.cluster.local:80
+              - name: KURA_EXTENSION_HTTP_CLIENT_TUIST_BASE_URL
+                value: http://${RELEASE}-tuist-server.${NAMESPACE}.svc.cluster.local:80
+              - name: KURA_EXTENSION_TUIST_ALLOW_SHARED_TENANTS
+                value: "1"
+            nodeSelector:
+              node.cluster.x-k8s.io/pool: kura
```
fortmarek: **[P2] Don't make normal previews depend on dedicated Kura capacity**

This makes every Kura-enabled Slack preview schedule its runtime onto the dedicated `kura` pool, and the workflow defaults `kura_instance_replicas` to `1`, so the normal on-demand path now depends on Kura-specific capacity. That seems misaligned with where Kura is heading: if we are moving Kura from elastic cloud VMs toward bare metal, this preview model bakes in the short-lived assumption that we can cheaply spin up or consume isolated Kura worker capacity per preview. For most previews, we probably only need to validate Tuist Server <-> Kura integration for the PR/SHA under test; the Kura pod itself can run colocated with the existing preview workloads on the preview node pool. I'd make the default Kura preview profile run a colocated Kura runtime in the preview namespace using the same preview scheduling/tolerations, and reserve dedicated Kura/bare-metal placement for explicit infra/performance tests. That should reduce cost, capacity pressure, and cleanup risk while keeping the preview mechanism compatible with the bare-metal direction.

.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,445 @@
+name: On-Demand Preview Deploy
```
fortmarek: **[P2] Reconcile Slack previews with the existing PR preview lifecycle**

This adds a second full preview lifecycle beside `.github/workflows/preview-deploy.yml`: both workflows build an image, install the same Helm chart, label preview namespaces for TTL sweeping, smoke test, and tear down releases, but they now differ in important semantics. PR previews auto-update on `synchronize`, show GitHub Deployments/comments, and are torn down by label removal/PR close; Slack previews are slug-based upserts with TTL/audit metadata and optional Kura. I'd avoid letting these become two separate preview systems. A better shape would be one shared deploy/delete implementation, exposed through thin trigger wrappers: PR label/sync/close passes `source=github-pr`, `slug=pr-<number>`, `ref=<head sha>`, `profile=app`; Slack passes `source=slack`, `slug=<request slug>`, `ref=<pr|sha|default>`, `profile=app-with-kura` or `kura_replicas=N`. That keeps update semantics, labels, cleanup order, docs, and future Kura placement changes in one place while preserving the different trigger surfaces.

fortmarek (commented):
Two follow-up architecture questions from the preview lifecycle review.

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Fixed in 699462e + f6a2e79. The KuraInstance CRD now carries a `tolerations` field that the controller passes through to the runtime pod template, so a KuraInstance can colocate on the existing preview workers (which already carry `role=preview:NoSchedule`) instead of needing dedicated Kura-class capacity.

In the workflow, the on-demand KuraInstance spec is now:
```yaml
nodeSelector:
  role: preview
tolerations:
  - key: role
    operator: Equal
    value: preview
    effect: NoSchedule
```

I also dropped the dedicated `kura` MachineDeployment from `cluster-preview.yaml` and the two `role=kura` workers from `kind-preview.yaml` since the colocated path is now the default — the standard on-demand preview lifecycle no longer depends on Kura-specific capacity. The stale `pool!=kura` exclusions in the preview taint/label selectors are gone too. Managed environments (`values-tuist-*.yaml`) still pin to the dedicated regional Kura pools — only the preview path changes.

An explicit infra/perf-test preview profile can re-introduce a dedicated pool later when it ships, with `tolerations` already part of the CRD contract.

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Agreed on the shape — one shared deploy/delete implementation with thin per-trigger wrappers is the right end state, especially for keeping update semantics, labels, cleanup order, and future Kura placement changes in one place.

I'd rather do that refactor in a focused follow-up PR rather than expand the scope of this one. This PR already touches enough surfaces (Slack flow, Kura runtime placement, Helm chart wiring, controller CRD), and folding the PR-preview workflow into a shared reusable workflow is a meaningful restructuring of `preview-deploy.yml` semantics (`workflow_call`, parameterized profiles, GitHub Deployments wiring) that I'd like to land behind its own review.

The plan I have in mind for the follow-up:
- `.github/workflows/preview-deploy-impl.yml` exposed via `workflow_call` taking `source`, `slug`, `ref`, `profile` (`app` / `app-with-kura`), `ttl_hours`, requester metadata
- `preview-deploy.yml` becomes a thin PR-label/sync/close → impl wrapper (GitHub Deployments + comments stay there)
- `preview-ondemand-deploy.yml` becomes a thin Slack → impl wrapper (audit metadata stays there)
- Single sweep + single delete implementation behind both

Let me open a tracking issue so the follow-up doesn't slip — does that work for you?

.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,456 @@
...
+      # `state=creating` marker is cleared.
+      - name: Mark namespace as creating
+        env:
+          REQUESTER_EMAIL: ${{ inputs.requester_email }}
+          REQUESTER_SLACK_ID: ${{ inputs.requester_slack_id }}
+          REQUEST_REASON: ${{ inputs.reason }}
+        run: |
+          set -euo pipefail
+          NOW=$(date +%s)
+          kubectl get namespace "$NAMESPACE" >/dev/null 2>&1 || kubectl create namespace "$NAMESPACE"
```
fortmarek: **[P1] Avoid pre-creating a namespace Helm still renders**

This pre-creates `preview-ondemand-<slug>` before Helm runs, but the Kura overlay still renders a `Namespace` object for the same namespace through `templates/kura-controller.yaml`. Helm usually refuses to install over an existing namespace that lacks Helm ownership metadata (`app.kubernetes.io/managed-by`, `meta.helm.sh/release-*`), so a first Slack preview create can fail before deploying anything. We should either stop rendering the `Namespace` when `kuraController.namespace == .Release.Namespace`, or add the Helm ownership metadata here before `helm upgrade --install` tries to adopt it.

tuist-ops/priv/repo/migrations/20260617120000_create_preview_requests.exs
```diff
@@ -0,0 +1,44 @@
...
+    create index(:preview_requests, [:expires_at])
+
+    # Stop two concurrent `/preview create <slug>` calls from both racing
+    # past insert and dispatching workflows for the same namespace; the
+    # second one fails closed at the DB. Scoped to in-flight statuses so
+    # a previous failed/terminal row does not block a fresh request.
+    create unique_index(
+             :preview_requests,
+             [:slug],
+             where: "status IN ('requested', 'provisioning', 'deleting')",
```
fortmarek: **[P1] Active-slug index blocks delete and same-slug updates**

Create requests transition to `provisioning`, but nothing in this PR ever transitions that row to a terminal status because the GitHub Actions workflow has no completion callback into tuist-ops. Since the partial unique index treats `provisioning` as active, the first `/preview create demo ...` leaves an active row behind and later `/preview delete demo`, the Delete button, or a same-slug create-as-update all try to insert another active row and hit this constraint. That makes the advertised delete/update lifecycle fail after the first successful dispatch. Consider a single mutable row per slug, workflow callbacks that mark create/delete rows terminal, or a narrower lock that only protects concurrent creates without blocking delete/update actions.

.github/workflows/preview-sweep.yml
```diff
@@ -45,28 +45,56 @@ jobs:
...
 
-          while IFS='=' read -r NS EXP; do
+          while IFS=$'\t' read -r NS RELEASE EXP STATE; do
             [ -z "$NS" ] && continue
+            if [ -z "$RELEASE" ]; then
+              # Strip the `preview-` prefix to get the helm release name.
+              RELEASE="${NS#preview-}"
+            fi
+
+            if [ "$STATE" = "creating" ]; then
```
fortmarek, line 87: **[P2] Add a backstop for namespaces stuck in creating**

Skipping `state=creating` avoids racing an in-flight install, but there is no age or failure backstop. If a create run is cancelled or fails after this label is written, especially after some resources are created but before `state=ready` and `expires-at` are set, the hourly sweeper will skip that namespace forever. Please add a cutoff using the `created-at` annotation, write a temporary expiry before Helm starts, or add a failure/cancel cleanup path so partially-created previews cannot leak indefinitely.

infra/k8s/onboarding.md
```diff
@@ -231,6 +231,8 @@ The [`infra/helm/k8s-monitoring/`](../helm/k8s-monitoring/) chart forwards Kuber
 
 Preview environments live on the `tuist-preview` workload cluster, which runs Postgres / ClickHouse / MinIO embedded alongside the server. Each preview is its own Helm release in its own namespace, with auto-deletion driven by a TTL label and an hourly sweep workflow.
 
+Slack-requested previews use the same cluster and the same embedded dependency shape, but layer `infra/helm/tuist/values-preview-kura.yaml` after `values-preview.yaml` to enable the Kura controller. App pods land on the tainted preview worker pool; controller-managed Kura pods land on the autoscaled `kura` worker pool. Requests enter through `/preview` in Slack, are audited in `tuist-ops`, and dispatch `.github/workflows/preview-ondemand-deploy.yml`; TTL cleanup is handled by `.github/workflows/preview-sweep.yml`.
```
fortmarek: **[P3] Update docs after moving Kura previews onto the preview pool**

This still says controller-managed Kura pods land on an autoscaled `kura` worker pool, but the latest workflow creates the `KuraInstance` with `nodeSelector: role: preview` and the preview toleration, and `cluster-preview.yaml` now describes only the shared preview worker pool. `infra/AGENTS.md` has the same stale statement. Please update both so operators and agents do not go looking for a Kura pool that this PR intentionally removed.

fortmarek (commented):
Re-review of the latest preview/Kura changes.

.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,456 @@
...
+          install_args: "helm kubectl jq"
+          cache: "false"
+      - name: Load kubeconfig
+        run: |
+          mkdir -p "$(dirname "$KUBECONFIG")"
+          echo "${{ secrets.KUBECONFIG }}" | base64 -d > "$KUBECONFIG"
+          chmod 600 "$KUBECONFIG"
+      - name: Temporarily remove preview taint for platform hooks
+        run: |
+          kubectl taint nodes -l '!node-role.kubernetes.io/control-plane' role=preview:NoSchedule- || true
```
fortmarek: **[P2] Configure platform scheduling instead of opening preview workers**

This temporarily removes the `role=preview:NoSchedule` taint so the platform chart's ingress-nginx/cert-manager admission hooks can schedule, because those platform pods/jobs do not have the preview toleration. That works as a bootstrap workaround, but it also creates a window where any non-preview workload can land on preview workers, and platform pods created during the window may later fail to reschedule after the taint is re-applied if they still lack tolerations. The same pattern exists in `preview-deploy.yml`. Can we instead set the preview overlay's platform chart tolerations/node selectors explicitly, or keep a small untainted system worker pool for platform infrastructure, so preview deploys do not have to mutate node taints around Helm?

fortmarek (commented):
One additional scheduling concern from the platform taint workaround.

fortmarek (commented):
One additional privilege/lifecycle concern around platform reconciliation from preview deploys.

.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,456 @@
...
+      - name: Reconcile platform chart
+        run: |
+          set -euo pipefail
+          CLOUDFLARE_API_TOKEN="$(kubectl -n platform get secret cloudflare-api-token \
+            -o jsonpath='{.data.api-token}' | base64 -d)"
+          echo "::add-mask::$CLOUDFLARE_API_TOKEN"
+          export CLOUDFLARE_API_TOKEN
+          mise -C infra run k8s:install-platform "$KUBECONFIG" tuist-preview
```
fortmarek: **[P2] Keep platform reconciliation out of preview deploys**

This makes every Slack preview deploy read the Cloudflare API token and run the full platform chart installer, which can mutate cert-manager, ingress-nginx, external-dns, CRDs/admission hooks, the wildcard cert, and load balancer configuration. That is much broader privilege and blast radius than an ephemeral app preview should need. I'd split this into a dedicated platform workflow, triggered by changes under `infra/helm/platform/**` / `install-platform.sh`, which owns Cloudflare token access and reconciles ingress/TLS/DNS. The preview workflows should only assert readiness (`helm status platform`, ingress-nginx rollout, `ClusterIssuer` ready, wildcard TLS Secret present) and then deploy the app. If we still want self-healing, `install-platform.sh` should at least only require `CLOUDFLARE_API_TOKEN` when the `platform/cloudflare-api-token` Secret is missing, so normal preview deploys don't need to read or propagate the DNS token.

infra/k8s/onboarding.md
fortmarek: **[P3] Refresh the Slack preview placement description**

This paragraph is no longer up to date after the latest Kura placement change. It still says controller-managed Kura pods land on an autoscaled `kura` worker pool, but the workflow now creates the preview `KuraInstance` with `nodeSelector: role: preview` plus the preview toleration, and `cluster-preview.yaml` describes a single preview worker pool. This should say that app pods and preview Kura runtime pods land on the tainted preview worker pool, with the Kura runtime created as a preview-scoped `KuraInstance` in the preview namespace.

fortmarek (commented):
One more docs follow-up for the updated Kura preview placement.

infra/mise/tasks/helm/preview-up.sh
```diff
@@ -4,6 +4,10 @@
 #USAGE flag "--release <name>" help="Helm release name" default="pr-demo"
 #USAGE flag "--remote" help="Use pre-built images from ghcr.io instead of building locally"
 #USAGE flag "--version <version>" help="Image version tag when using --remote" default="latest"
+#USAGE flag "--with-kura" help="Also enable the Kura controller and create a shared private two-node KuraInstance"
```
fortmarek: **[P2] Make Kura part of the default preview shape**

Now that Kura preview runtimes colocate on the normal preview worker instead of using a dedicated Kura capacity profile, keeping `--with-kura` makes the default local preview diverge from Slack previews. Pulling existing Kura images should be cheap relative to the rest of this task, and if the change touches Kura or the controller, building/loading those images is genuine validation work rather than optional overhead. I'd invert this: include Kura by default so `helm:preview-up` reproduces the Slack preview shape, and only add a `--without-kura` / `--profile app-only` escape hatch if we really need a faster app-only local loop. The Kura placement check can then be unconditional for the default path.

fortmarek (commented):
One more follow-up on the local preview profile after colocating Kura.

.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,456 @@
...
+            replicas: ${KURA_REPLICAS}
+            publicHost: ${KURA_HOST}
+            ingressClassName: nginx
+            storageSize: 10Gi
+            extraEnv:
+              - name: KURA_CONTROL_PLANE_URL
+                value: http://${RELEASE}-tuist-server.${NAMESPACE}.svc.cluster.local:80
+              - name: KURA_EXTENSION_HTTP_CLIENT_TUIST_BASE_URL
+                value: http://${RELEASE}-tuist-server.${NAMESPACE}.svc.cluster.local:80
+              - name: KURA_EXTENSION_TUIST_ALLOW_SHARED_TENANTS
```
fortmarek: **[P2] Reconsider preview-only tenant multiplexing**

This enables a single preview KuraInstance (`tenantID: preview-${SLUG}`) to serve arbitrary requested account tenants by relaxing the hook's `requested tenant == server tenant` guard. That made sense when each tenant implied expensive Kura infrastructure such as dedicated VMs/LBs, but after the latest changes a preview KuraInstance is just pods/PVCs behind shared ingress on the preview pool. Keeping tenant multiplexing now adds preview-only authorization behavior (`KURA_EXTENSION_TUIST_ALLOW_SHARED_TENANTS` plus `TUIST_KURA_SHARED_MESH`) and makes previews less production-shaped. Can we instead use the normal one-tenant-per-KuraInstance model in previews, reusing the shared ingress/LB and bounding cost with TTL/small storage/quota, so we do not need this cross-tenant bypass path?

fortmarek (commented):
One more design follow-up on preview tenant multiplexing.

tuist-ops/lib/tuist_ops/previews/request.ex
```diff
@@ -0,0 +1,90 @@
...
+  why, which workflow was dispatched, and the TTL metadata the sweeper uses.
+  """
+
+  use Ecto.Schema
+  import Ecto.Changeset
+
+  @actions ~w(create delete)
+  @statuses ~w(requested provisioning deleting failed)
+
+  schema "preview_requests" do
```
fortmarek: **[P2] Model previews as current state instead of request rows**

I think `preview_requests` may be more data model than we need here, and it is already causing lifecycle friction because create/delete/update are modeled as separate active rows for the same slug. For a first version, a single `previews` table as the source of truth should be enough: `slug`, `status`, `host`, `namespace`, `release`, `ref_kind/ref_value` (and maybe resolved SHA), creator/updater Slack/email fields, reason, `expires_at`, `deleted_at`, `failed_at`, failure reason, and workflow metadata. Then `/preview create <slug>` can upsert/update that row, `/preview delete <slug>` can transition it to deleting/deleted, and the UI can show currently active previews directly. We can add separate request/event history later if we need full audit trails, but the current state table covers who created what and avoids the partial unique-index problem.

fortmarek (commented):
One more data-model suggestion for preview state.

tuist-ops/slack-app-manifest.yaml
```diff
@@ -15,7 +16,7 @@
 
 display_information:
   name: Tuist JIT
```
fortmarek: **[P3] Rename the Slack app now that it is not just JIT**

The manifest still names the app and bot `Tuist JIT`, but this PR expands it to own `/preview` as well, so the Slack surface is now broader than just just-in-time access. Pedro also named the destination channel `eng-ops`, so the manifest/comments/default config should reflect that operational scope instead of the old JIT-only framing. Consider renaming the app/bot to something like `Tuist Ops`, updating the manifest instructions that point at `Tuist JIT`, and making sure `SLACK_PREVIEWS_CHANNEL_ID` is documented/configured for `#eng-ops` rather than silently falling back to `#tailscale-jit-approvals`.

fortmarek (commented):
One more naming/config follow-up for the Slack app now that it handles previews too.

tuist-ops/slack-app-manifest.yaml
fortmarek: Don't forget that this also need to be re-pasted in the Slack app admin page

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Outdated by the platform-level Kura controller refactor in 972ca0e. The per-preview release no longer enables `kuraController`, so the chart no longer renders a `Namespace` object for the preview namespace at all (the only `Namespace` template was inside `kura-controller.yaml`, gated on `kuraController.enabled`). The preview namespace is created by `kubectl create namespace` before `helm upgrade`, helm respects it, and no chart template tries to adopt it.

tuist-ops/priv/repo/migrations/20260617120000_create_preview_requests.exs
pepicrft: Fixed in 2bf06c0. The migration was edited in place (this PR's migration hasn't shipped yet) to narrow the index to `action='create' AND status IN ('requested', 'provisioning')`, renamed to `preview_requests_active_create_slug_index`. Concurrent `/preview create demo` calls still fail closed at the DB; subsequent `/preview delete demo`, the Delete button, and `/preview create demo` as an update are no longer blocked by the orphaned `provisioning` row from the first dispatch. The unique_constraint on the changeset follows the rename. Tests pass.

The deeper concern about rows staying at `provisioning` forever is real and worth addressing — the cleanest fix is the GH Actions completion callback you mentioned, which I'll do in a follow-up so this PR ships unblocked.

.github/workflows/preview-sweep.yml
pepicrft: Fixed in 2bf06c0. The sweep now treats `state=creating` as in-flight only when `tuist.dev/created-at` is within `STUCK_CREATE_AGE_SECONDS` (default 1800s = 30 min). Past that, the namespace is treated as a botched create — same teardown path as an expired preview — so cancelled or failed creates can't leak indefinitely. Tunable per-run via `STUCK_CREATE_AGE_SECONDS` if we ever need to relax it.

infra/k8s/onboarding.md
pepicrft: Fixed in 2bf06c0. Both `infra/k8s/onboarding.md` §8 and `infra/AGENTS.md` now reflect the new shape: app pods and the preview's Kura runtime pods both land on the `role=preview` worker pool, the Kura controller is installed once cluster-wide in the `kura` namespace by `mise -C infra run k8s:install-kura-platform`, and each preview's `KuraInstance` is created in that `kura` namespace (cross-namespace Service DNS for the server↔runtime hop). The dedicated `kura` MachineDeployment is gone from `cluster-preview.yaml`.

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Fixed in 2bf06c0. Went with the "configure platform scheduling" option since `cluster-preview.yaml` only has the one preview worker pool. Added `role=preview:NoSchedule` tolerations to every platform subchart in `infra/helm/platform/values-tuist-preview.yaml` — cert-manager (controller + webhook + cainjector + startupapicheck), ingress-nginx (controller + admission webhook patch Job + defaultBackend), external-dns, external-secrets (controller + webhook + certController), metrics-server. The deploy workflow's untaint / re-apply dance is gone from `preview-deploy-impl.yml`; only the idempotent `role=preview` re-label step survives (CAPI doesn't surface node labels via topology variables, so fresh MachineDeployment nodes still come up unlabeled).

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Fixed in 2bf06c0. Went with the smaller fix you proposed at the bottom of the comment: `install-platform.sh` now skips reading and writing the Cloudflare DNS-edit token when the `platform/cloudflare-api-token` Secret is already materialised on the cluster. Normal preview deploys no longer propagate the credential — only the bootstrap that created the Secret in the first place needs it (from `$CLOUDFLARE_API_TOKEN` in CI, or `op read` for local invocations).

The bigger refactor (dedicated platform workflow triggered by `infra/helm/platform/**` changes, preview workflows asserting readiness only) is the right end state but a meaningful re-shape of CI on top of this PR. Happy to open a follow-up if you'd like that tracked explicitly.

infra/k8s/onboarding.md
pepicrft: Fixed in 2bf06c0. The §8 paragraph in `onboarding.md` now reads: "App pods and the preview's Kura runtime pods both land on the tainted preview worker pool — Kura previews colocate on `role=preview` with a matching toleration, not a dedicated Kura node pool. The Kura controller is installed cluster-wide in the `kura` namespace once per cluster by `mise -C infra run k8s:install-kura-platform`, and each preview's `KuraInstance` is created in that `kura` namespace." Matches the implementation.

infra/mise/tasks/helm/preview-up.sh
pepicrft: Fixed in 2bf06c0. `preview-up.sh` now runs the with-Kura path by default and exposes `--without-kura` as the escape hatch for fast app-only iteration. The Kura placement check that asserts pods land on `role=preview` workers runs unconditionally for the default path now.

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Agreed on the direction — now that the runtime is just pods/PVCs on the shared preview pool, the one-tenant-per-KuraInstance model is what previews should look like long-term. Keeping `ALLOW_SHARED_TENANTS` + the server's `TUIST_KURA_SHARED_MESH` short-circuit is preview-only authorization behavior that doesn't match production, and I'd rather not normalize that.

The simplest path to remove the cross-tenant bypass without breaking previews:

1. Drop `KURA_EXTENSION_TUIST_ALLOW_SHARED_TENANTS` from the KuraInstance's `extraEnv`. The Lua hook then strictly enforces tenant matching.
2. Drop `TUIST_KURA_SHARED_MESH` + the `kura_shared_mesh?` flag + the `:kura` shortcut in `accounts.ex`. The server uses normal per-account routing.
3. Seed a `kura_servers` row in the preview DB at deploy time so the preview's test account routes to this KuraInstance via the production path. The KuraInstance's `tenantID` matches the seeded account's tenant.

The third step is where the work lives — it requires either a deploy-time mix task that the workflow shells into the server pod to run, or a chart-rendered Helm hook Job that does the insert. Neither exists today.

I'd rather peel this out of this PR and do it as a focused follow-up: it touches server code, the chart, and the deploy workflow, and the design (which account handle to seed, whether to add the `kura_servers` row via mix task vs. Helm hook, how the local kind validation maps onto it) deserves its own review. The current preview is still useful for testing server↔Kura integration; the cleanup is a quality improvement, not a correctness blocker.

Want me to open the tracking issue?

tuist-ops/lib/tuist_ops/previews/request.ex
pepicrft: Fixed in f2629c3. Restructured as a state table along the lines you sketched:

- Table renamed `preview_requests` → `previews`, `action` column dropped (state-table doesn't model action events), single `unique_index(:previews, [:slug])` since slug is the natural key.
- Schema renamed `TuistOps.Previews.Request` → `TuistOps.Previews.Preview` (file renamed in place). Statuses are `creating | active | deleting | deleted | failed`; `active` and `deleted` are reserved for the future GH Actions completion callback (the cleanup you flagged in the active-slug thread).
- `Previews.create/1` upserts by slug: inserts when missing, replaces when the existing row is `deleted` / `failed`, and rolls back with `:already_exists` when the row is still in flight (`creating | active | deleting`). `Previews.delete/1` looks up by slug and transitions to `deleting`. Failures flip to `failed` and update the Slack card.
- `:already_exists` and `:not_found` get human-readable messages in the slash response.

Since the migration hasn't shipped, the table is renamed in place — no data migration step needed. Full tuist-ops suite is green.

The single-mutable-row-per-slug shape means the next iteration (workflow callbacks, separate event history if we end up wanting full audit trails) drops cleanly on top: callbacks transition the existing row to `active` / `deleted`, and a history table can stay additive.

tuist-ops/slack-app-manifest.yaml
pepicrft: Fixed in d77fa1d. Manifest now names the app and bot "Tuist Ops" with description "Slack-driven operations for the Tuist team (JIT elevation, previews, ...)" — the operational umbrella you described. The doc comment at the top of `slack-app-manifest.yaml` reflects the wider surface. Re-applying the manifest in Slack preserves the bot user, install state, and existing tokens.

Also switched the default fallback channel from `#tailscale-jit-approvals` to `#eng-ops` in `slack_controller.ex` so it matches where the team has converged. `SLACK_APPROVALS_CHANNEL_ID` / `SLACK_PREVIEWS_CHANNEL_ID` env vars still override per environment.

`chat:write.public` scope comment updated too.

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Done — 6f70aff. Went with the strict-tenant-matching shape:

**Authorization (Lua hook):**
- Dropped `KURA_EXTENSION_TUIST_ALLOW_SHARED_TENANTS` from the KuraInstance's `extraEnv` in `preview-deploy-impl.yml`. The hook now enforces `requested tenant == server tenant` in previews the same way it does in production.

**Server routing:**
- Dropped `TUIST_KURA_SHARED_MESH` from `values-preview-kura.yaml`.
- Dropped `Environment.kura_shared_mesh?/0` from the server.
- Dropped the `:kura` shortcut in `Accounts.get_cache_endpoints_for_handle/2`. The function now goes through the normal per-account `kura_servers` lookup in every environment. The preview-only branch in `accounts.ex` is gone; the tests collapsed to the production shape (full server suite stays green).

**Tradeoff (documented):**
A fresh preview no longer routes a signed-up account to Kura automatically — no `kura_servers` row exists yet. Testers exercising the server↔Kura path do a one-time seed against the preview's server pod. I added a copy-pasteable snippet to `infra/k8s/onboarding.md` §8 that:

1. Picks an account by handle (defaults to `tuistrocks`),
2. Inserts a `kura_servers` row pointing at `https://kura-<slug>.preview.tuist.dev` with `technology: :kura`,
3. Enables the `kura_cache` feature flag for that account.

Auto-seeding on deploy is a future-improvement option — a chart-rendered Helm hook Job, a seed in the deploy workflow, or a `TUIST_KURA_PREVIEW_AUTO_PROVISION` server-side flag that creates the row on first sign-up. Worth a focused issue if you'd rather make `/preview create` give a fully Kura-ready preview out of the box; this PR ships the architectural cleanup you asked for and leaves the seed mechanism for that follow-up so we're not blending two changes here.

.github/workflows/preview-deploy-impl.yml
```diff
@@ -0,0 +1,433 @@
...
+          metadata:
+            name: ${KURA_INSTANCE}
+            namespace: kura
+            labels:
+              tuist.dev/preview-source: ${SOURCE}
+              tuist.dev/slug: ${SLUG}
+              tuist.dev/shared: "true"
+          spec:
+            accountHandle: preview-${SLUG}
+            tenantID: preview-${SLUG}
```
fortmarek: [P1] This KuraInstance sets `tenantID` to `preview-${SLUG}`, but the documented setup later points real accounts, for example `tuistrocks`, at `https://kura-<slug>.preview.tuist.dev`. The hook still rejects requests where the request tenant/account differs from `KURA_TENANT_ID` unless shared tenants are enabled, and this manifest does not enable that flag. As written, a seeded real account would route to this Kura endpoint and then get a 403. We either need one KuraInstance per account, to keep shared-tenants enabled for this shared preview runtime, or to route preview requests with a tenant that matches `preview-${SLUG}`.

infra/mise/tasks/k8s/install-kura-platform.sh
```diff
@@ -0,0 +1,68 @@
...
+if [ -z "$CLIENT_ID" ]; then
+  CLIENT_ID="kura-platform-$(openssl rand -hex 4)"
+fi
+if [ -z "$CLIENT_SECRET" ]; then
+  CLIENT_SECRET="$(openssl rand -hex 32)"
+fi
+echo "::add-mask::$CLIENT_SECRET"
+
+log "Rendering Kura controller manifests for namespace=$KURA_NAMESPACE tag=$KURA_CONTROLLER_IMAGE_TAG"
+helm template kura-platform "$CHART_PATH" \
```
fortmarek, line 58: [P1] The controller rendered here has no preview toleration because `kuraController.tolerations` is never set. Since preview workers carry `role=preview:NoSchedule` and the deploy workflow removed the temporary untaint step, this controller pod can remain Pending on the preview cluster; then the rollout wait below times out before any KuraInstance exists. Please pass the preview toleration for this install, or keep the controller on an untainted system pool.

tuist-ops/lib/tuist_ops/previews.ex
```diff
@@ -0,0 +1,218 @@
...
+        {:error, :not_found}
+
+      %Preview{status: "deleted"} ->
+        {:error, :not_found}
+
+      %Preview{} = preview ->
+        provision(
+          preview,
+          %{
+            status: "deleting",
```
fortmarek: [P1] This only records the row as `deleting` before dispatching the workflow; there is no callback or follow-up code in this PR that marks the preview `deleted` after GitHub Actions succeeds. Because `Preview.active_status?/1` treats `deleting` as active, `/preview create <same-slug>` will be rejected forever after a successful delete dispatch. We need a completion path, for example webhook/poll/sweeper, or a different delete-state model so slugs become reusable.

.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,180 @@
...
+      requester_slack_id:
+        description: Requesting Slack user id
+        required: false
+      reason:
+        description: Request reason
+        required: false
+      kura_controller_image_tag:
+        description: ghcr.io/tuist/kura-controller tag (cluster-wide controller install)
+        required: false
+        default: latest
```
fortmarek: [P2] These Kura image inputs default to mutable `latest`, and the shared preview workflow only builds/pushes the server image. A Slack preview for a PR that changes `infra/kura-controller` or `kura/` will therefore deploy the previous/latest Kura controller/runtime instead of the PR bits. Since the preview is meant to validate Kura changes too, please build/push PR-scoped Kura images when those paths change or require immutable tags derived from the PR commit.

infra/k8s/onboarding.md
```diff
@@ -231,6 +231,24 @@ The [`infra/helm/k8s-monitoring/`](../helm/k8s-monitoring/) chart forwards Kuber
...
 
+Slack-requested previews use the same cluster and the same embedded dependency shape, but layer `infra/helm/tuist/values-preview-kura.yaml` after `values-preview.yaml` to enable a shared Kura runtime. App pods and the preview's Kura runtime pods both land on the tainted preview worker pool — Kura previews colocate on `role=preview` with a matching toleration, not a dedicated Kura node pool. The Kura controller is installed cluster-wide in the `kura` namespace once per cluster by `mise -C infra run k8s:install-kura-platform`, and each preview's `KuraInstance` is created in that `kura` namespace; the preview server reaches the runtime over cross-namespace Service DNS. Requests enter through `/preview` in Slack, are audited in `tuist-ops`, and dispatch `.github/workflows/preview-ondemand-deploy.yml` (which delegates the lifecycle to `preview-deploy-impl.yml`); TTL cleanup is handled by `.github/workflows/preview-sweep.yml`.
```
fortmarek: [P3] This new preview section is now the source of truth for the Slack preview flow, but the later 8.3 runbook still says `preview-deploy` / `preview-sweep` scale the MachineDeployment and asks operators to provision `KUBECONFIG_MGMT`. The new workflows no longer reference that secret or scale the MachineDeployment, so future bootstrap work will set up unused credentials and expect scale-to-zero behavior that will not happen. Please update or remove that stale scaler section as part of this docs change.

fortmarek (commented):
How was the Slack preview flow validated end to end? From the inline findings, it looks like the Kura controller may not schedule on the tainted preview pool, the shared Kura instance has a synthetic tenant that the strict Lua hook will reject for real account handles, and deleted preview rows never transition to a reusable terminal state. Can you share the exact validation run, including creating a Slack preview with Kura, routing a real/test account through it, performing a cache read/write, deleting it, and recreating the same slug?

.github/workflows/preview-ondemand-deploy.yml
```diff
@@ -0,0 +1,180 @@
...
+            echo "expires_at=$EXPIRES_AT"
+            echo "ttl_hours=$TTL_HOURS"
+            echo "profile=$PROFILE"
+            echo "kura_instance_replicas=$KURA_REPLICAS"
+          } >> "$GITHUB_OUTPUT"
+
+  call-impl:
+    name: Lifecycle
+    needs: resolve
+    uses: ./.github/workflows/preview-deploy-impl.yml
```
fortmarek: [P3] This wrapper plus `preview-deploy.yml` plus `preview-deploy-impl.yml` leaves one preview lifecycle spread across three workflows. I understand the intent is to share the deploy/delete implementation between PR-triggered and Slack-triggered previews, but this indirection makes the update/delete/Kura semantics much harder to review and operate. Unless there is a concrete GitHub Actions limitation that requires the reusable-workflow split, can we collapse the Slack/manual/PR entrypoints into `preview-deploy.yml` with a resolver job and conditional PR decoration, or move the shared lifecycle into mise/scripts and keep at most one deploy workflow plus `preview-sweep.yml`?

tuist-ops/priv/repo/migrations/20260617120000_create_previews.exs
```diff
@@ -0,0 +1,42 @@
+defmodule TuistOps.Repo.Migrations.CreatePreviews do
+  use Ecto.Migration
+
+  def change do
+    # One row per preview slug, mutated as state evolves. Modeling create /
```
fortmarek: [P3] Please remove this implementation-history comment from the migration. Migrations become permanent project history, and this refers to an earlier table shape plus review context that never shipped as part of the final PR. The schema itself is clear enough here; if we want to document the lifecycle model, a short current-state comment in the context/schema module would age better.

infra/mise/tasks/helm/preview-up.sh
```diff
@@ -1,9 +1,13 @@
 #!/usr/bin/env bash
-#MISE description="Spin up a multi-node kind cluster and deploy a preview release pinned to role=preview node"
+#MISE description="Spin up a multi-node kind cluster and deploy a preview release (with Kura by default) onto the role=preview worker"
 #USAGE flag "--cluster <name>" help="Kind cluster name" default="tuist-preview"
 #USAGE flag "--release <name>" help="Helm release name" default="pr-demo"
 #USAGE flag "--remote" help="Use pre-built images from ghcr.io instead of building locally"
 #USAGE flag "--version <version>" help="Image version tag when using --remote" default="latest"
+#USAGE flag "--without-kura" help="Skip the cluster-wide Kura controller install and the shared KuraInstance. Use only for fast app-only local loops; the default matches the Slack preview shape."
```
fortmarek: [P2] I think this `--without-kura` mode should go away completely. The preview chart already has to pull/start several images for the server, cache, embedded Postgres, ClickHouse, MinIO, and migration jobs, so Kura should not be special just because it adds another image. If the default path is slow, we should optimize the Kura setup path or use prebuilt images, not keep an app-only preview mode that diverges from Slack/GitHub previews and can miss Kura regressions.

infra/mise/tasks/helm/preview-up.sh
```diff
@@ -126,35 +173,140 @@ if helm status "$RELEASE_NAME" >/dev/null 2>&1; then
...
+kind: KuraInstance
+metadata:
+  name: ${KURA_INSTANCE_NAME}
+  namespace: kura
+  labels:
+    tuist.dev/preview-source: slack
+    app.kubernetes.io/instance: ${RELEASE_NAME}
+    tuist.dev/shared: "true"
+spec:
+  accountHandle: preview-${RELEASE_NAME}
```
fortmarek: [P2] Please remove the synthetic shared preview account/tenant entirely. If Kura is part of the normal preview shape, the preview Kura pod should exercise the same account routing semantics as production instead of using `preview-${RELEASE_NAME}` plus the shared-tenant escape hatch below. Keeping this special shared account means the local preview can pass while real account/project cache access still fails, and it preserves the tenant-multiplexing model we are trying to retire.

tuist-ops/priv/repo/migrations/20260617120000_create_previews.exs
```diff
@@ -0,0 +1,42 @@
...
+      add :slug, :string, null: false
+      add :status, :string, null: false, default: "creating"
+      add :requester_email, :string, null: false
+      add :requester_slack_id, :string, null: false
+      add :ref_kind, :string
+      add :ref_value, :string
+      add :reason, :text, null: false
+      add :ttl_seconds, :integer
+      add :host, :string
+      add :namespace, :string
```
fortmarek: [P3] Please trim the preview table to the fields this PR actually consumes, or wire the consumers in this PR. `namespace` and `release` are deterministic from the slug and are never read after insert, `workflow_id`/`workflow_ref` are stored but not used for links, polling, or reconciliation, and `deleted_at` is only cleared but never set because there is no successful delete completion path. `expires_at` is only displayed/indexed while cleanup still happens from Kubernetes labels. Keeping these columns makes the schema look like it supports lifecycle/UI/reconciliation behavior that is not implemented yet.

.github/workflows/preview-deploy-impl.yml
pepicrft: Fixed in e58810f. The KuraInstance manifest now uses `accountHandle: preview` / `tenantID: preview`, and the deploy workflow runs a new `mise -C infra run preview:seed-account` step after the smoke test that:

1. Creates a user `preview@preview.tuist.dev` (handle `preview`, default password `preview-temp-password`) via `Tuist.Accounts.create_user/2` if it doesn't exist;
2. Inserts a `kura_servers` row for that account pointing at `https://kura-<slug>.preview.tuist.dev` with `technology: :kura`;
3. Enables `FunWithFlags.kura_cache` for the account.

The Lua hook's strict tenant check now passes because the seeded account routes to a KuraInstance whose `tenantID` matches its handle. The step is idempotent — re-runs reuse the existing user, swallow the cache-endpoint uniqueness error, and re-enable the flag. `infra/mise/tasks/preview/seed-account.sh` is shared between the deploy workflow and `preview-up.sh`, so local kind matches the managed cluster's shape exactly.

Onboarding §8 was updated to drop the manual seed snippet — the preview is Kura-ready out of the box; testers log in as `preview@preview.tuist.dev` with the default password.

infra/mise/tasks/k8s/install-kura-platform.sh
pepicrft: Fixed in e58810f. `install-kura-platform.sh` now passes `--set kuraController.tolerations[0]={role,Equal,preview,NoSchedule}` through to the `helm template --show-only` render so the controller pod schedules onto the role=preview workers. The matching platform-chart tolerations from the previous round (`values-tuist-preview.yaml`) cover everything else, so we don't need the untaint workaround anywhere on the preview cluster.

tuist-ops/lib/tuist_ops/previews.ex
pepicrft: Fixed in e7e9dae. `Previews.delete/1` now treats a successful workflow dispatch as terminal: the row transitions straight to `deleted` (with `deleted_at` set) and `Preview.active_status?/1` drops `deleting` from its in-flight set. `/preview create <same-slug>` immediately upserts the row back to `creating` and dispatches the deploy workflow.

The trade-off (and how I've sequenced the cleanup): we don't try to roll the state back if the delete workflow itself fails in CI. The sweep workflow's TTL is the backstop for orphaned k8s objects, and a follow-up create over the same slug calls `helm upgrade --install` against whatever half-deleted namespace remains. A real completion callback (webhook from the delete job → tuist-ops endpoint) is the natural next iteration and can land additively without re-shaping the state model.

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Fixed in 7a460df. The new consolidated `preview-deploy.yml` has a `paths-changed` job that diffs the deploying ref against `origin/main` (or the PR base when present) and flags whether `infra/kura-controller/**` or `kura/**` was touched. Conditional `build-kura-controller` and `build-kura-runtime` jobs run when their respective paths changed, producing immutable `sha-<short>` tags pushed to ghcr. The `resolve` job propagates the same tags down to the KuraInstance manifest and the platform-controller install, so a Kura-touching preview validates the PR's bits instead of `:latest`.

Slack-dispatched runs against a SHA without a PR (no base ref available) default to building both Kura images so the preview always validates the ref it was asked for. PR/SHA flows that don't touch Kura paths skip the builds and reuse the latest controller image.

infra/k8s/onboarding.md
pepicrft: Fixed in e58810f. §8.3 is gone (the new workflows don't scale the MachineDeployment or use `KUBECONFIG_MGMT`). §8.4 became §8.3 (first preview). Future operators won't set up a credential or expect a behavior the workflow doesn't have.

.github/workflows/preview-ondemand-deploy.yml
pepicrft: Fixed in 7a460df. Collapsed `preview-deploy.yml` + `preview-deploy-impl.yml` + `preview-ondemand-deploy.yml` into a single `preview-deploy.yml` with:

- `resolve` job normalising every trigger (pull_request label/sync/unlabel/close + workflow_dispatch by PR / SHA / slug) into one shape
- `paths-changed` job for Kura-aware builds
- `build-server` (always for deploys) + conditional `build-kura-controller` + conditional `build-kura-runtime`
- one `deploy` job that handles both `app` and `app-with-kura` profiles
- one `delete` job
- two PR-only finalize jobs (GitHub Deployments + PR comments)

tuist-ops dispatches the new file (`Environment.preview_workflow_id/0` default updated to `preview-deploy.yml`); the dispatch action value changed from `create` → `deploy` to match what the workflow expects. The internal `Previews.create/1` API name is unchanged. All 161 tuist-ops tests pass; `preview-sweep.yml` is the only other workflow.

Going further (Marek's second option — moving the deploy/delete bodies into mise scripts) is still a clean follow-up on top of this if you'd rather review more of the lifecycle as bash than YAML.

tuist-ops/priv/repo/migrations/20260617120000_create_previews.exs
pepicrft: Fixed in e7e9dae. The migration is back to just the schema + indexes — no commentary about earlier iterations, no `Marek's review` reference. Future archaeology can use `git log` on the migration.

infra/mise/tasks/helm/preview-up.sh
pepicrft: Fixed in e58810f. `--without-kura` is gone from `preview-up.sh`. Kura is part of the default local preview shape — same as Slack previews — and the script always installs the cluster-wide controller, copies the introspection Secret, creates the `KuraInstance`, and runs the placement check. No app-only mode.

infra/mise/tasks/helm/preview-up.sh
pepicrft: Fixed in e58810f. The KuraInstance in `preview-up.sh` now uses `accountHandle: preview` / `tenantID: preview` (matching what the managed cluster does) and the `KURA_EXTENSION_TUIST_ALLOW_SHARED_TENANTS=1` env is gone. The same `preview:seed-account` mise task runs at the end of the local install: creates the `preview` account, wires the kura_servers row, enables the feature flag. Local kind exercises the same routing as the managed cluster — no synthetic shared tenant or escape hatch.

tuist-ops/priv/repo/migrations/20260617120000_create_previews.exs
pepicrft: Fixed in e7e9dae. Dropped `namespace`, `release`, `workflow_id`, `workflow_ref`, and `expires_at` from both the migration and the `TuistOps.Previews.Preview` schema (and from the `:expires_at` index that came with it). Kept `host`, `slug`, `status`, requester fields, `ref_kind`/`ref_value`, `reason`, `ttl_seconds`, `deleted_at`, `failed_at`, `failure_reason`, and the Slack channel/message id — every one of those has a live consumer in this PR. `Previews.create/1` derives `namespace`/`release` inline from the slug when handing off to the GitHub Actions client.

Full tuist-ops suite is green (161 tests).

pepicrft:
Honest answer: I haven't run a Slack preview end-to-end against the real preview cluster, and I should have flagged that earlier.

The chicken-and-egg is the immediate blocker — `gh workflow run` requires the workflow file to exist on `main` first (GitHub looks up the workflow by filename in the default branch's registry, regardless of `--ref`). So I can't dispatch `preview-deploy.yml` against this branch until either (a) the workflow file lands on `main` ahead of the rest of the PR, or (b) the PR merges. Same wall hits the Slack `/preview` path because tuist-ops uses the same dispatch API. The local kind validation (`mise run helm:preview-up`) needs an `OP_SERVICE_ACCOUNT_TOKEN` or `TUIST_LICENSE_KEY` that I don't have access to.

What's actually been validated so far is static:

- `helm template` renders for both profiles + `helm lint` is clean
- `helm template --show-only templates/kura-controller.yaml` renders the kura-platform install path
- The chart helper resolves `KURA_CONTROL_PLANE_CLIENT_ID/SECRET` correctly for managed (`kuraController.enabled=true`) and preview (`server.kuraIntrospection.useSharedSecret=true`) layouts
- `actionlint` is clean across the consolidated workflow modulo the self-hosted runner labels
- `kura-controller` Go build + `go test ./controllers` passes, including the flip-to-private Certificate cleanup test
- 161 tuist-ops tests pass, including the delete-then-recreate-same-slug path through `Previews.create` / `Previews.delete`
- The server `Accounts.get_cache_endpoints_for_handle/2` tests pass for the new (no shared-mesh) path

What I have **not** validated behaviorally:

- That the `preview-deploy.yml` resolver maps the Slack dispatch shape correctly under real conditions
- That the seed `kubectl exec` command actually runs to completion against the server pod and the `Tuist.Accounts` / `FunWithFlags` API I called (I inferred the signatures from grepping the server)
- That a cache read/write actually flows through the preview Kura with strict tenant matching
- That delete + recreate for the same slug works end-to-end through the consolidated workflow

To actually verify, I'd need one of:

1. **Cherry the consolidated workflow + chart bits to `main` as a separate small PR.** ~5 additive files, no-op until called. Then I can `gh workflow run preview-deploy.yml --ref chore/i-d-like-to-design` with a slug and walk through the full create → cache read/write → delete → recreate flow. This is the path I'd prefer.
2. **Hand me an `OP_SERVICE_ACCOUNT_TOKEN` or `TUIST_LICENSE_KEY`** (or run `mise run helm:preview-up` yourself) so the local kind validation can run.
3. **Merge first, validate after** — which I know inverts the order you'd want.

Option 1 is what I'd recommend; want me to prep that cherry-PR?

tuist-ops/lib/tuist_ops/previews/slack_blocks.ex
```diff
@@ -0,0 +1,102 @@
...
+          """
+        }
+      },
+      %{
+        type: "context",
+        elements: [
+          %{
+            type: "mrkdwn",
+            text:
+              "GitHub Actions is provisioning it. It expires <!date^#{DateTime.to_unix(preview.expires_at)}^{date_short_pretty} at {time}|soon>."
```
fortmarek: [P1] This still references `preview.expires_at`, but the `Preview` schema/migration no longer defines `expires_at` after the schema trim. `Previews.create/1` calls `SlackBlocks.provisioning/1` before dispatching GitHub Actions, so `/preview create` will raise here and never create the preview. Please either restore and set `expires_at`, or derive the displayed expiry from `ttl_seconds` without reading a missing field.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -550,12 +550,14 @@ var grpcREAPIPathPrefixes = []string{
...
-	// gRPC co-hosts on PublicHost (it is the rule host below), so an empty
+	// Private instances expose no public endpoint, so neither the public
+	// nor the gRPC Ingress is reconciled. For non-private instances, gRPC
+	// co-hosts on PublicHost (it is the rule host below), so an empty
 	// PublicHost has nowhere to route — delete rather than emit an Ingress
 	// with an empty host, even when GRPCPublicHost is set. The provisioner
 	// always emits grpcPublicHost and publicHost together (or neither), so in
 	// practice this only guards hand-written, gRPC-only CRs.
-	if instance.Spec.GRPCPublicHost == "" || instance.Spec.PublicHost == "" {
+	if instance.Spec.Private || instance.Spec.GRPCPublicHost == "" || instance.Spec.PublicHost == "" {
```
fortmarek: [P2] This gate still suppresses the gRPC Ingress unless `grpcPublicHost` is set, but the preview workflow creates a `KuraInstance` with `publicHost` only. The surrounding comments say gRPC co-hosts on `publicHost` and `grpcPublicHost` is legacy, so previews can pass the HTTP `/ready` smoke test while the REAPI/ByteStream gRPC routes are missing. Please drop the `GRPCPublicHost` requirement here and in `grpcPublicURL`, or keep setting `grpcPublicHost` until the controller no longer gates on it.

kura/ops/helm/kura/hooks/tuist.lua
```diff
@@ -19,8 +19,11 @@
 --
 --   2. authorize — resolve the request target from `ctx.tenant_id` /
 --      `ctx.namespace_id`, require the requested tenant to match
---      `ctx.server_tenant_id`, then check the target against first-
---      class account/project cache grants for the requested action.
+--      `ctx.server_tenant_id` by default, then check the target against
+--      first-class account/project cache grants for the requested action.
+--      Preview environments may set
```
fortmarek: [P2] Please remove the shared-tenant escape hatch completely. The current preview workflow no longer sets `KURA_EXTENSION_TUIST_ALLOW_SHARED_TENANTS` and instead seeds a single `preview` account, so keeping this bypass and its Rust test preserves the tenant-multiplexing model we wanted to retire and lets future deployments disable strict tenant routing again. The hook should always enforce request tenant == server tenant now.

infra/k8s/onboarding.md
```diff
@@ -231,6 +231,10 @@ The [`infra/helm/k8s-monitoring/`](../helm/k8s-monitoring/) chart forwards Kuber
...
```
fortmarek: [P3] This section is stale relative to the current workflow shape. The PR no longer has `preview-ondemand-deploy.yml` or `preview-deploy-impl.yml`; Slack, manual, and PR previews now all enter through `.github/workflows/preview-deploy.yml`. Please update this paragraph so operators do not follow deleted workflow names while bootstrapping or debugging previews.

fortmarek (approved):

tuist-ops/lib/tuist_ops/previews/slack_blocks.ex
pepicrft: Fixed in `a806836d6c`. `TuistOps.Previews.SlackBlocks.provisioning/1` now derives the Slack expiry from `ttl_seconds` and the preview row timestamp instead of reading `preview.expires_at`. I added `tuist-ops/test/previews/slack_blocks_test.exs` to cover the rendered Slack date.

infra/kura-controller/controllers/kurainstance_controller.go
pepicrft: Fixed in `6efd93b15e`. `reconcileGRPCIngress/2` and `grpcPublicURL/1` now gate on `spec.publicHost` only, while still skipping private or hostless instances. The controller test for missing `grpcPublicHost` now asserts the gRPC ingress and status web address are created from `publicHost`. I also updated the custom resource definition printer column and controller guidance so `grpcPublicHost` stays documented as ignored.

kura/ops/helm/kura/hooks/tuist.lua
pepicrft: Fixed in `e3a7773325`. The Tuist hook no longer reads `KURA_EXTENSION_TUIST_ALLOW_SHARED_TENANTS`, and the mismatch branch always returns 403 when the request tenant differs from the server tenant. I removed the Rust helper and test that allowed cross-tenant access with that environment variable. The existing strict mismatch test still covers the denial path, and the namespace-only Bazel gRPC test still passes.

infra/k8s/onboarding.md
pepicrft: Fixed in `8a20f28524`. `infra/k8s/onboarding.md` now points Slack, manual, and pull request previews at `.github/workflows/preview-deploy.yml`, and the adjacent preview docs describe the seeded `tuist` organization plus test-user sign-in button. I also updated the remaining stale workflow comments in `cluster-preview.yaml`, `values-preview-kura.yaml`, and `install-kura-platform.sh`.

PR #11354, opened 2026-06-18
test(cli): cover SwifterPM materialization passthrough

pepicrft:
> Covers configuring SwifterPM cached directory materialization through Tuist's existing install passthrough arguments instead of introducing a dedicated environment variable.

SwifterPM already exposes `--cached-directory-materialization` with `automatic`, `copy`, and `symlink`. Tuist also already has an install passthrough path, both from `Config.project.generatedProject.installOptions.passthroughSwiftPackageManagerArguments` and from the `TUIST_INSTALL_PASSTHROUGH_ARGUMENTS` environment variable used by `tuist install`. Reusing that path keeps the CLI surface smaller and avoids adding a second environment variable for one SwifterPM option.

- Verifies install config passthrough forwards `--cached-directory-materialization symlink` into the Swift Package Manager controller.
- Verifies the SwifterPM-backed controller parses `--cached-directory-materialization symlink` from its existing arguments array.
- Verifies the default Tuist request leaves the materialization option unset, which SwifterPM resolves as `automatic` internally.
- Keeps SwifterPM gated by the existing truthy `TUIST_USE_SWIFTERPM` opt-in, so SwiftPM fallback behavior is unchanged.

This means users can configure the strategy with the existing env path, for example `TUIST_INSTALL_PASSTHROUGH_ARGUMENTS=--cached-directory-materialization,symlink`, or through install options in `Tuist.swift`.

### How to test locally

- `swiftformat cli/Sources/TuistConstants/Constants.swift cli/Sources/TuistSupport/SwiftPackageManager/SwiftPackageManagerController.swift cli/Tests/TuistSupportTests/SwiftPackageManager/SwiftPackageManagerControllerTests.swift cli/Tests/TuistKitTests/Services/InstallServiceTests.swift`
- `TUIST_USE_SWIFTERPM=0 .build/debug/tuist generate tuist TuistSupport TuistSupportTests TuistKit TuistKitTests ProjectDescription --no-open`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistSupportTests/SwiftPackageManagerControllerTests -only-testing TuistKitTests/InstallServiceTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`

### DIFF
--- a/cli/Tests/TuistKitTests/Services/InstallServiceTests.swift
+++ b/cli/Tests/TuistKitTests/Services/InstallServiceTests.swift
@@ -339,7 +339,10 @@ final class InstallServiceTests: TuistUnitTestCase {
             .willReturn(
                 Tuist.test(project: .generated(.test(
                     installOptions: .test(
-                        passthroughSwiftPackageManagerArguments: ["--replace-scm-with-registry"]
+                        passthroughSwiftPackageManagerArguments: [
+                            "--cached-directory-materialization",
+                            "symlink",
+                        ]
                     )
                 )))
             )
@@ -370,7 +373,14 @@ final class InstallServiceTests: TuistUnitTestCase {
             .resolve(at: .any, arguments: .any, printOutput: .any)
             .called(1)
         verify(swiftPackageManagerController)
-            .resolve(at: .any, arguments: .value(["--replace-scm-with-registry"]), printOutput: .any)
+            .resolve(
+                at: .any,
+                arguments: .value([
+                    "--cached-directory-materialization",
+                    "symlink",
+                ]),
+                printOutput: .any
+            )
             .called(1)
     }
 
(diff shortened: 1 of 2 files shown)


cli/Sources/TuistConstants/Constants.swift
```diff
@@ -68,6 +68,7 @@ public enum Constants {
         public static let detailedLog = "TUIST_CONFIG_DETAILED_LOG"
         public static let osLog = "TUIST_CONFIG_OS_LOG"
         public static let useSwifterPM = "TUIST_USE_SWIFTERPM"
+        public static let swifterPMCachedDirectoryMaterialization = "TUIST_SWIFTERPM_CACHED_DIRECTORY_MATERIALIZATION"
```
fortmarek: instead of a dedicated environment variable, should we consider using the same passthrough install arguments option that we have already?

fortmarek (approved):

PR #11356, opened 2026-06-18
feat(infra,server): serve Kura HTTP cache + gRPC (Bazel REAPI) from a single host

esnunes:
## What & why

Serve the Kura **HTTP cache API** and **gRPC (Bazel REAPI)** from a **single hostname** — `https://<host>` for the REST cache and `grpcs://<host>` for the Remote Execution API — eliminating today's separate `grpc.<host>` hostname. This is "Option E": one host, path-based routing at the nginx ingress.

The backend was already a single `Service` exposing both `http:4000` and `grpc:50051`; the *only* reason two hostnames existed is that ingress-nginx's `backend-protocol` annotation is per-Ingress-object. This PR keeps two Ingress objects but co-hosts them on one host, split by **path** (gRPC service prefixes → `grpc_pass`; everything else → `proxy_pass`).

Implements the plan in `kura-single-endpoint-grpc-http.md`.

## Phase 0 — validation spike (the hard gate)

The risky premise is whether ingress-nginx honors `backend-protocol` **per-location** when two Ingresses share one host. This is not the documented per-Ingress model, so it was demonstrated, not assumed.

Spun up an isolated kind cluster with **ingress-nginx v1.11.3** (the same image the repo's gateway controller pins), a single Service exposing `http`+`grpc` ports (mirroring Kura's shape), and two Ingresses on one host `spike.local`:
- Ingress A: `/` (Prefix) → `http`, `backend-protocol: HTTP`, TLS for the host.
- Ingress B: REAPI/ByteStream path prefixes → `grpc`, `backend-protocol: GRPC`, **no TLS of its own**.

**Results — PASS:**

| Check | Result |
|---|---|
| `curl https://H/up` + arbitrary REST path | ✅ 200 via `proxy_pass` (HTTP backend) |
| gRPC reflection (bidi-streaming) over `grpcs://H` | ✅ service list returned |
| gRPC unary (`DummyUnary`) | ✅ echo round-trip |
| gRPC server-streaming (`DummyServerStream`, the ByteStream read/write analog) | ✅ 10 messages, trailers intact |
| Single TLS cert serves both | ✅ Ingress B declared **no** TLS; relied on Ingress A's cert for the shared host |
| Per-location `backend-protocol` on merged host | ✅ one `server` block; `grpc_pass` for gRPC prefixes, `proxy_pass` for `/` |
| nginx host/annotation merge-conflict warnings | ✅ none |

Generated `nginx.conf` for the merged host (ground truth):

```nginx
## start server spike.local            # ← ONE server block
server { server_name spike.local ;
  location ~* "^/build\.bazel\.remote\.execution\.v2\." { ... grpc_pass grpc://upstream_balancer; }
  location ~* "^/google\.bytestream\."                  { ... grpc_pass grpc://upstream_balancer; }
  location ~* "^/"                                       { ... proxy_pass http://upstream_balancer; }
}
```

### Critical finding that shaped the implementation

`pathType: Prefix`/`ImplementationSpecific` for the gRPC prefixes is **wrong**. ingress-nginx renders `/build.bazel.remote.execution.v2.` as `location /build.bazel.remote.execution.v2./` plus an exact `= /build.bazel.remote.execution.v2.`. A real method path — `/build.bazel.remote.execution.v2.Capabilities/GetCapabilities` — matches **neither** (the char after the package prefix is `C`, not `/`) and falls through to `/` (HTTP). The fix, verified in the spike, is **`nginx.ingress.kubernetes.io/use-regex: "true"`** with anchored, regex-escaped paths (`^/build\.bazel\.remote\.execution\.v2\.`), which render as `location ~* "..."` and match correctly. This is the exact pitfall the plan flagged.

### Phase 1 — Operator (`infra/kura-controller`, Go)
- `reconcileGRPCIngress`: co-hosts on `spec.publicHost` (not `grpcPublicHost`) with the REAPI/ByteStream regex path prefixes (`use-regex`), `backend-protocol: GRPC`, `grpc` backend port; **declares no TLS** (the public Ingress's cert covers the shared host). Using `PublicHost` for the rule host means existing CRs that still carry a legacy `grpc.<host>` converge onto the single host the moment the controller rolls out.
- `reconcileGRPCCertificate`: now a **deleter** — retires the dedicated `<name>-grpc-tls` Certificate *and* the Secret cert-manager leaves behind. The public certificate covers the host.
- `grpcPublicURL` status derives from `PublicHost` → `grpcs://<public host>`.
- Tests updated; `go build/vet`, `gofmt`, and `go test ./...` (incl. `TestGatewayNginxConfigMatchesChart`) pass.

### Phase 2 — Server provisioner (Elixir, `server/`)
- `regions.ex`: the managed-region gRPC host template now **equals** the public host template (env-suffix handling unchanged). One change drives both the CR's `grpcPublicHost == publicHost` and `grpc_public_url` → `grpcs://<public host>`.
- `kubernetes_controller.ex`: `@manifest_revision` bumped so the reconciler re-applies the manifest to all existing instances.
- Tests updated.
- **gRPC readiness check: intentionally skipped.** The existing HTTP `/up` probe (`kura.ex`) now covers the whole endpoint — gRPC shares the same host, pod, TLS, and nginx server block, so a 200 on `/up` already proves the server block is live.

### Phase 3 — CLI (Swift, `cli/`)
- **No changes needed.** PR #11344 (which added the `grpc.` host prefix + `TUIST_CACHE_GRPC_ENDPOINT` override) is not in this branch. `BazelSetupCommandService` already scheme-swaps on the **same** host (`https://<host>` → `grpcs://<host>`) and registers the credential helper for that same host; its tests already assert this. Confirmed there is no `grpc.`-prefix logic anywhere in `cli/Sources`.

### Intent layer
- `infra/kura-controller/AGENTS.md` updated to describe the single-host model.

## Phase 4 — Rollout & backward compatibility

The migration is **infra-only** and there are **no stranded clients**:
- The CLI already derives the gRPC endpoint by scheme-swapping the HTTP cache host; it never reads the server's `grpc_public_url`.
- `Tuist.Kura.Provisioner.grpc_public_url/2` has **zero callers** in the codebase — the legacy `grpc.<host>` value is only ever materialized as an Ingress/DNS/cert by the operator.

Sequence:
1. **Ship infra (Phases 1–2).** The single `<name>-grpc` Ingress object is updated **in place** (same object name) from host `grpc.<host>`/path `/` to host `<host>`/REAPI path prefixes. external-dns prunes the now-orphaned `grpc.<host>` A record once that host leaves the Ingress; the retired gRPC cert/secret are cleaned up by the controller. Because gRPC now lands on a host the CLI was already targeting, this is a pure improvement with no regression window.
2. CLI requires no change (Phase 3).
3. No separate "remove legacy" step is required, since nothing kept serving `grpc.<host>`.

**Enterprise `KuraGateway`:** automatically covered. The gateway controller creates no Ingress objects — it only runs a dedicated ingress-nginx for a dedicated `IngressClass`; the co-hosting change lives on the `KuraInstance` ingresses, which inherit that class.

## Validation run

- `infra/kura-controller`: `gofmt -l` clean, `go vet ./...`, `go build ./...`, `go test ./...` all pass (in `golang:1.25`).
- Phase 0 end-to-end spike on kind + ingress-nginx v1.11.3 (HTTP + gRPC unary/streaming/reflection, single cert, nginx.conf inspected) — PASS. Throwaway cluster torn down.
- `server/`: changed `.ex`/`.exs` files parse cleanly (`Code.string_to_quoted!`). **Could not run `mix test`/`credo`/`format` locally** — this worktree isn't bootstrapped (`deps/`/`_build/` absent) and the 1Password-provisioned `priv/secrets/test.key` is not present, which the app boot in `test_helper.exs` requires. Relying on CI for the server suite.
- The gateway's `http2_body_preread`/buffer tuning is unchanged (it lives in the gateway ConfigMap/Helm values, applied controller-wide regardless of host/path), so large gRPC uploads keep the same tuning on the co-hosted location.


infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -790,44 +824,42 @@ func (r *KuraInstanceReconciler) reconcilePublicCertificate(ctx context.Context,
...
+// certificate is provisioned. cert-manager does not garbage-collect the
+// issued Secret when its Certificate is deleted, so the Secret is removed
+// explicitly. Get-before-Delete keeps a converged reconcile read-only
+// (the Gets are served from the controller-runtime cache) instead of
+// issuing DELETE writes on every loop forever.
+//
+// TODO(kura single-host migration): once every environment has converged
+// (no <name>-grpc-tls Certificate or Secret remains anywhere), drop the
+// call from Reconcile and delete this function.
+func (r *KuraInstanceReconciler) retireLegacyGRPCCertificate(ctx context.Context, instance *kurav1alpha1.KuraInstance) error {
```
pepicrft, line 845: This now runs every reconcile regardless of `--grpc-cluster-issuer`. On a cluster without the cert-manager CRD installed, `r.Get` on the unstructured `Certificate` returns `meta.NoKindMatchError`, which `client.IgnoreNotFound` doesn't catch (it only handles `apierrors.IsNotFound`). Reconcile then fails forever before reaching the StatefulSet.

Pre-PR `reconcileGRPCCertificate` had the same shape but was only reached when `GRPCClusterIssuer`/`GRPCPublicHost` was empty, so managed Tuist (both always set) never hit it. With the unconditional retire path, any on-prem operator who enables `kuraController.enabled: true` without cert-manager hits this on every reconcile.

A cheap fix is to skip the retire path when `r.GRPCClusterIssuer == ""`, on the grounds that without an issuer we never minted the legacy cert in the first place. Worth also confirming the controller's RESTMapper behavior here against a real cluster without the cert-manager CRD before we ship.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -515,9 +514,43 @@ func (r *KuraInstanceReconciler) reconcilePublicIngress(ctx context.Context, ins
...
+// anchor itself when use-regex is set, so it is not — and must not be —
+// included here.
+//
+// Note: only the REAPI and ByteStream packages are routed. Kura registers no
+// gRPC health or reflection service, so those paths intentionally fall to the
+// HTTP backend; add `/grpc\.` here if Kura ever serves them.
+var grpcREAPIPathPrefixes = []string{
+	`/build\.bazel\.remote\.execution\.v2\.`,
+	`/google\.bytestream\.`,
+}
```
pepicrft, line 540: Migration concern for on-prem: this rewrites the gRPC Ingress in place from `grpcPublicHost` to `publicHost`. Managed is fine (external-dns auto-prunes the old `grpc.<host>` A record, CLI fetches fresh URLs every cache resolution). For a self-host customer with their own DNS for `grpc.<host>` and Bazel WORKSPACE files pinned to `grpcs://grpc.<host>:443`, the moment they pull the new chart the Ingress no longer claims that host, ingress-nginx returns 404, and Bazel breaks until WORKSPACE configs are updated.

Options: keep both hosts on the Ingress for one release as a grace window, or document a clear migration step in the self-host guide (retire `grpc.<host>` DNS + update WORKSPACE configs) before this rolls out. Current state has no migration affordance.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -515,9 +514,43 @@ func (r *KuraInstanceReconciler) reconcilePublicIngress(ctx context.Context, ins
...
+// so this object keeps backend-protocol: GRPC (grpc_pass/h2c) while the
+// public Ingress keeps backend-protocol: HTTP (proxy_pass) on `/`. The
+// public Ingress terminates TLS for the host, so this object declares no TLS
+// of its own. Host equals PublicHost (not GRPCPublicHost) so existing CRs
+// that still carry a legacy grpc.<host> in grpcPublicHost converge onto the
+// single host as soon as this controller rolls out.
 func (r *KuraInstanceReconciler) reconcileGRPCIngress(ctx context.Context, instance *kurav1alpha1.KuraInstance) error {
 	ingress := &networkingv1.Ingress{ObjectMeta: metav1.ObjectMeta{Name: grpcServiceName(instance), Namespace: instance.Namespace}}
-	if instance.Spec.GRPCPublicHost == "" {
+	if instance.Spec.GRPCPublicHost == "" || instance.Spec.PublicHost == "" {
```
pepicrft, line 558: The new gate deletes the gRPC Ingress when `PublicHost == ""` even if `GRPCPublicHost` is explicitly set. Our provisioner always sets both so we don't hit it, but a custom CR with gRPC-only exposure silently loses the Ingress (and the `Status.GRPCPublicURL` clears with no event). Intentional? If yes, a one-liner comment on the gate would save the next reader a trip through the history.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -1393,7 +1425,12 @@ func publicIngressAnnotations() map[string]string {
 }
 
 func grpcIngressAnnotations() map[string]string {
```
pepicrft, line 1432: Worth being explicit somewhere (chart README, AGENTS.md, or the CR doc): path-split routing with `use-regex` + per-location `backend-protocol` is ingress-nginx specific. Pre-existing constraint but single-host hardens the dependency, since with split-host an alternative controller could at least serve the gRPC host with its own config. After this PR, Traefik/Contour/HAProxy/Cilium clusters silently route gRPC to the HTTP backend via `proxy_pass`.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -515,9 +514,43 @@ func (r *KuraInstanceReconciler) reconcilePublicIngress(ctx context.Context, ins
...
+//
+// The values must start with `/` (the Kubernetes Ingress API rejects a path
+// that does not begin with a slash); ingress-nginx adds the leading `^`
+// anchor itself when use-regex is set, so it is not — and must not be —
+// included here.
+//
+// Note: only the REAPI and ByteStream packages are routed. Kura registers no
+// gRPC health or reflection service, so those paths intentionally fall to the
+// HTTP backend; add `/grpc\.` here if Kura ever serves them.
+var grpcREAPIPathPrefixes = []string{
```
pepicrft, line 537: The comment already calls this out, just flagging: if any Bazel client or sidecar (Envoy/Linkerd) probes `/grpc.health.v1.Health/Check` or reflection paths over the established h2 connection, it falls through to the HTTP `/` location and gets `proxy_pass` over HTTP/1.1, corrupting framing on an h2 stream. Probably not exercised by stock Bazel, but worth confirming none of the Bazel client configs we ship enable health checking.

infra/kura-controller/controllers/kurainstance_controller_test.go
```diff
@@ -1137,6 +1166,146 @@ func TestKuraInstanceReconcileSkipsGRPCWhenHostUnset(t *testing.T) {
...
+	updated := &kurav1alpha1.KuraInstance{}
+	if err := reconciler.Get(ctx, types.NamespacedName{Name: instance.Name, Namespace: instance.Namespace}, updated); err != nil {
+		t.Fatal(err)
+	}
+	if got := updated.Status.GRPCPublicURL; got != "" {
+		t.Fatalf("expected empty gRPC public URL when the public host is unset, got %q", got)
+	}
+}
+
+func TestKuraInstanceReconcileConvertsLegacyGRPCIngressToSingleHost(t *testing.T) {
```
pepicrft, line 1218: Nit: this test seeds only the legacy Ingress, not a legacy Certificate or Secret. The realistic mid-migration shape (old Ingress + old Cert/Secret both present) is only exercised in `TestKuraInstanceReconcileExposesGRPCWhenHostSet`. A regression that swallows the retire path on a converged cluster would still keep this test green. Trivial to seed `legacyGRPCCert` + `legacyGRPCSecret` here too.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -19,7 +19,7 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   alias Tuist.Kura.Server
 
   @namespace "kura"
-  @manifest_revision "2026-06-16-node-port-and-env-scoped-public-host-v1"
+  @manifest_revision "2026-06-18-node-port-and-single-host-grpc-v1"
```
pepicrft, line 22: Bumping the revision triggers an immediate re-reconcile of every CR on first server start after upgrade, which is what we want for managed. For any on-prem enterprise that has forked `regions.ex` with their own templates, the Ingress rewrite happens in the first reconcile loop after upgrade with no staged rollout flag. Fine for us, just want to make sure we're explicit about the blast radius.

pepicrft (commented):
Did a deeper pass with the on-prem story in mind. Two findings I'd want to address before this lands, the rest are smaller.

**On-prem follow-ups (not necessarily blocking, but worth landing alongside or right after):**

- `infra/helm/tuist/values.yaml:27` still describes `tlsClusterIssuer` as covering "the per-instance public HTTPS and gRPC TLS Certificates" (plural). After this PR there's one Cert per instance; the comment will mislead operator inventory tooling (kyverno/falco/GitOps) into expecting `<name>-grpc-tls` and possibly fighting our delete.
- Nothing in `server/priv/docs/en/guides/server/self-host/` mentions the single-host model, the cert-manager requirement on the controller path, the ingress-nginx requirement, or DNS migration steps for operators with their own `grpc.<host>` records. Worth a short page or section before this lands.
- Standalone Kura chart at `kura/ops/helm/kura/values.yaml:63` and `kura/README.md:442` still document the split-host pattern (`kura-grpc.local`/`kura-grpc.example.com`). Out of scope for this PR, just flagging so the public Kura deployment story doesn't drift from what we ship.

Inline comments below.

infra/kura-controller/controllers/kurainstance_controller.go
esnunes: At the moment, and prior to this PR, the grpc url isn't exposed to users, tuist cli also doesn't point to the hostname with the `grpc` prefix, making the grpc endpoint reachable but _invisible_, therefore not necessarily requiring a migration path

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
esnunes: it is a valid point, but the blast radius in this case is zero for the reason described here: https://github.com/tuist/tuist/pull/11356#discussion_r3440813028

infra/kura-controller/controllers/kurainstance_controller.go
esnunes: The existing code already expects the cert-manager to be installed, the `reconcilePublicCertificate` method called before this one suggests that. I've created [an issue](https://github.com/tuist/tuist/issues/11388) for addressing this problem in another PR

pepicrft (approved):

infra/kura-controller/controllers/kurainstance_controller.go
esnunes: Intentional, gRPC co-hosts on PublicHost, so an empty one has nowhere to route. Added an inline comment. The GRPCPublicURL will be deprecated since it will have no real use, I've created [this issue](https://github.com/tuist/tuist/issues/11390) to work on it later

infra/kura-controller/controllers/kurainstance_controller.go
esnunes: Confirmed, Kura registers only the four REAPI services, no grpc.health.v1.Health or reflection, and the .bazelrc we generate doesn't enable health checking, so stock Bazel won't hit it. Tracked the gRPC health service + the /grpc\. ingress routing in #11391

esnunes:
> Did a deeper pass with the on-prem story in mind. Two findings I'd want to address before this lands, the rest are smaller.
> 
> **On-prem follow-ups (not necessarily blocking, but worth landing alongside or right after):**
> 
> * `infra/helm/tuist/values.yaml:27` still describes `tlsClusterIssuer` as covering "the per-instance public HTTPS and gRPC TLS Certificates" (plural). After this PR there's one Cert per instance; the comment will mislead operator inventory tooling (kyverno/falco/GitOps) into expecting `<name>-grpc-tls` and possibly fighting our delete.
> * Nothing in `server/priv/docs/en/guides/server/self-host/` mentions the single-host model, the cert-manager requirement on the controller path, the ingress-nginx requirement, or DNS migration steps for operators with their own `grpc.<host>` records. Worth a short page or section before this lands.
> * Standalone Kura chart at `kura/ops/helm/kura/values.yaml:63` and `kura/README.md:442` still document the split-host pattern (`kura-grpc.local`/`kura-grpc.example.com`). Out of scope for this PR, just flagging so the public Kura deployment story doesn't drift from what we ship.
> 
> Inline comments below.

I've added a comment in the values.yaml and created the following issues to tackle the other two items:
- https://github.com/tuist/tuist/issues/11394
- https://github.com/tuist/tuist/issues/11393

PR #11357, opened 2026-06-18
fix(swifterpm): match SwiftPM submodule checkout behavior

pepicrft:
Resolves user-reported SwifterPM restore failures with invalid submodule gitlinks and tracked symlinks that point into missing local-only directories.

This aligns SwifterPM's Git fallback checkout behavior with SwiftPM:

- Stops running `git submodule update --init --recursive` after a fallback checkout. SwiftPM resolves and builds a package with a malformed submodule gitlink without initializing submodules, leaving the gitlink path as an empty directory.
- Keeps broken symlinks such as `.ruby_version -> DevOnly/ruby` in successful checkouts, matching SwiftPM's working copy behavior.
- Adds retry cleanup hardening so broken symlinks are unlinked correctly if a temporary checkout needs to be reset before trying another source-control URL candidate.
- Wraps source restore failures with package identity, location, and revision so future restore errors identify the dependency that failed.

The root cause was that SwifterPM's fallback path did more than SwiftPM does. It checked out the requested revision and then recursively initialized submodules. For repositories that use gitlinks as local-development-only placeholders, that extra command fails with `fatal: No url found for submodule path ... in .gitmodules`. When a tracked symlink pointed into that missing local-only directory, the retry cleanup could leave the broken symlink behind, causing the next checkout attempt to fail with a misleading untracked-file overwrite error.

The chosen fix is to remove recursive submodule initialization rather than special-case `.gitmodules`, `.ruby_version`, or a specific filename. That matches SwiftPM semantics and avoids making SwifterPM responsible for materializing submodule contents that SwiftPM itself does not restore.

For users, packages with malformed or intentionally local-only submodule gitlinks restore the same way they do under SwiftPM, and genuine restore failures now include the package context needed for debugging.

### How to test locally

- `swift test --replace-scm-with-registry --filter FileSystemSupportTests --filter RestoreTests`
- `mise run test:unit`
- Manual SwiftPM comparison using a temporary package with a bad gitlink and `.ruby_version -> DevOnly/ruby`.
- Manual SwifterPM reproduction against the same repository shape, confirming restore now exits 0 and preserves the broken symlink in the checkout.

### DIFF
--- a/swifterpm/Sources/swifterpm/FileSystemSupport.swift
+++ b/swifterpm/Sources/swifterpm/FileSystemSupport.swift
@@ -57,6 +57,10 @@ extension FileSystem {
 
     /// Remove the item at `url`. No-op if absent.
     func removePath(_ url: URL) async throws {
+        if isSymlink(url) {
+            try unlinkPath(url)
+            return
+        }
         try await remove(url.absolutePath)
     }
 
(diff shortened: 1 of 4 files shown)


swifterpm/Sources/swifterpm/FileSystemSupport.swift
fortmarek, line 1: out-of-scope here, but this whole file smells like the model generated some impromptu helpers instead of using our existing `tuist/FileSystem` library. Would consider doing a follow-up, I don't think this file should exist

fortmarek (approved):

swifterpm/Sources/swifterpm/FileSystemSupport.swift
pepicrft: Agreed this is broader than this PR. I kept this change scoped to matching SwiftPM checkout behavior and only touched these helpers so retry cleanup can unlink broken symlinks. I’ll leave replacing this file with tighter `tuist/FileSystem` abstractions for a follow-up.

PR #11358, opened 2026-06-18
Clarify SPM build tool plugin integration

pepicrft:
Resolves N/A

Document why SPM build tool plugins need to use Xcode's default SwiftPM integration instead of Tuist's XcodeProj-based external dependency mapping.

- Explains that build tool plugins need SwiftPM and Xcode to evaluate plugin code, discover commands, model declared inputs and outputs, and account for generated files.
- Points users to `Project.packages` plus `.package(product:type: .plugin)` as the supported alternative.
- Calls out the implications of that alternative and removes the raw `Package.swift` example so the Tuist API stays central.

### How to test locally

- `git diff --check`
- Not run: full docs build, Markdown-only documentation change.

### DIFF
--- a/server/priv/docs/en/guides/features/projects/dependencies.md
+++ b/server/priv/docs/en/guides/features/projects/dependencies.md
@@ -242,6 +242,9 @@ The tradeoff is that Xcode owns that plugin integration:
 - Tuist cannot give the plugin the same level of control, validation, caching, and selective-testing behavior that it can provide for statically mapped package products.
 - CI and local builds must allow Xcode's package resolution for those plugin packages.
 
+Because of these implications, we generally don't recommend using SPM build tool plugins in Tuist projects.
+When possible, prefer a build phase script declared in the target, or run the tool from a script outside the generated Xcode project before invoking Tuist or Xcode.
+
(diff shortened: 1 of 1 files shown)


server/priv/docs/en/guides/features/projects/dependencies.md
```diff
@@ -219,37 +219,31 @@ let target = .target(name: "MyTarget", dependencies: [
...
-        ),
-    ]
-)
-```
+The tradeoff is that Xcode owns that plugin integration:
+
+- Xcode resolves and builds the plugin package through its SwiftPM integration.
+- The package appears in the generated project as a Swift Package reference instead of a Tuist-generated external target.
+- Tuist cannot give the plugin the same level of control, validation, caching, and selective-testing behavior that it can provide for statically mapped package products.
+- CI and local builds must allow Xcode's package resolution for those plugin packages.
```
fortmarek, line 244: I'd probably mention that because of this, we generally don't recommend using build tool plugins and instead rely on build phase scripts and/or scripts completely outside of the Xcode project ...

fortmarek (approved):

PR #11377, opened 2026-06-19
fix(cache): deduplicate registry manifest variants

fortmarek:
## Summary

This fixes a SwiftPM registry-resolution failure that could block fresh worktrees before the CLI workspace could even be generated.

The change makes the cache registry treat Swift package manifest variants by their canonical Swift tools version instead of only by their filename. When a package has both a default `Package.swift` manifest and an alternate `Package@swift-X.Y.swift` manifest that resolve to the same Swift tools version, the registry now keeps the default manifest and stops advertising or persisting the redundant alternate.

## Motivation

New Codex worktrees were repeatedly getting stuck in this sequence:

1. A focused CLI build or Xcode test needs the generated workspace.
2. `tuist generate tuist ProjectDescription --no-open` fails because external dependencies have not been installed yet.
3. `tuist install` delegates to SwiftPM registry resolution with `--replace-scm-with-registry`.
4. SwiftPM aborts while resolving registry metadata with errors like:

```text
duplicate key found: 'Package@swift-5.5.0.swift'
duplicate key found: 'Package@swift-6.0.0.swift'
```

I reproduced the failure locally while trying to build the CLI. One concrete case was `pointfreeco.swift-snapshot-testing`, where SwiftPM failed with `duplicate key found: 'Package@swift-6.0.0.swift'`. A similar failure showed up earlier for `swiftlang.swift-docc-symbolkit` with `Package@swift-5.5.0.swift`.

This was especially frustrating because the error happens before the CLI code is compiled or tested. It makes agents report that they cannot run focused tests because the worktree cannot get past dependency resolution.

## Root Cause

The registry was returning alternate manifest links based on the manifest filenames in metadata or S3:

```text
Package.swift
Package@swift-6.0.swift
```

That is valid only if the alternate manifest represents a distinct Swift tools version. Some packages, however, publish an alternate manifest whose declared tools version is equivalent to the default manifest after SwiftPM canonicalization.

For `pointfreeco/swift-snapshot-testing`, the default manifest and `Package@swift-6.0.swift` both declare Swift tools version `6.0`. Newer SwiftPM/Xcode canonicalizes that to a three-component key such as `6.0.0`, so the two manifests collide internally as:

```text
Package@swift-6.0.0.swift
```

The archive itself does not need to contain two literal `Package@swift-6.0.0.swift` files. The duplicate is created by SwiftPM while normalizing equivalent manifest versions from the registry response.

## Approach

I added `Cache.Registry.ManifestVariants` as the single place that understands package manifest variants:

- parses alternate filenames such as `Package@swift-5.9.swift`
- parses `// swift-tools-version:` declarations
- canonicalizes tools versions to `major.minor.patch`
- filters alternates whose canonical tools version matches the default `Package.swift`
- filters duplicate alternates that canonicalize to the same tools version

That logic is applied in three places:

- `RegistryController` uses it before generating `Link` headers for alternate manifests. This fixes the active registry response path that SwiftPM consumes.
- `ReleaseWorker` uses it before persisting release manifest metadata. This prevents newly indexed packages from storing redundant manifest variants.
- `AlternateManifests` uses it in the S3 fallback path. When metadata is missing or stale, fallback discovery now reads the root manifest alongside alternates so it can make the same canonical-version decision.

Unknown or unparsable tools versions are still preserved. The filter only drops an alternate when it can prove that the canonical tools version was already represented by the default manifest or by an earlier alternate.

## Alternatives Considered

I considered treating this as a CLI/worktree issue and relying on `tuist install --force-resolved-versions`, which did unblock this one local checkout. That is only a workaround: fresh worktrees without a usable resolved file still hit the registry response and fail before generation or testing.

I also considered suppressing all alternate manifests. That would be too broad because alternate manifests are still useful for packages that genuinely need different manifests across Swift tools versions.

Exact filename de-duplication was not enough either. The observed failures are caused by SwiftPM canonicalization, so `6`, `6.0`, and `6.0.0` need to be compared as the same tools version.

## Impact

After this cache service change is deployed, fresh worktrees should be able to run `tuist install`, generate the CLI workspace, and run focused CLI builds/tests without hitting this duplicate-manifest SwiftPM failure.

The change is intentionally cache-side rather than CLI-side because the registry is the source of the invalid manifest variant set. Fixing the response also helps every SwiftPM client that resolves through the Tuist registry, not just Codex worktrees.

## Validation

I validated the cache change with focused tests:

```sh
mix test test/cache/registry/alternate_manifests_test.exs test/cache/registry/release_worker_test.exs test/cache_web/controllers/registry_controller_test.exs
```

Result:

```text
89 tests, 0 failures
```

I also formatted the touched cache files:

```sh
mix format lib/cache/registry/manifest_variants.ex lib/cache/registry/alternate_manifests.ex lib/cache/registry/release_worker.ex lib/cache_web/controllers/registry_controller.ex test/cache/registry/alternate_manifests_test.exs test/cache/registry/release_worker_test.exs test/cache_web/controllers/registry_controller_test.exs
```

Finally, I verified the original developer workflow after unblocking this local checkout with the resolved-file workaround that is needed until the live cache service is deployed:

```sh
tuist generate tuist ProjectDescription --no-open
xcodebuild build -workspace Tuist.xcworkspace -scheme tuist CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""
```

Result:

```text
** BUILD SUCCEEDED ** [422.510 sec]
```

### DIFF
--- a/cache/lib/cache/registry/alternate_manifests.ex
+++ b/cache/lib/cache/registry/alternate_manifests.ex
@@ -40,15 +40,25 @@ defmodule Cache.Registry.AlternateManifests do
 
     case Cachex.get(cache_name, key) do
       {:ok, nil} ->
-        manifests = discover(scope, name, version)
-        Cachex.put(cache_name, key, manifests, ttl: @ttl)
-        manifests
+        discover_and_maybe_cache(cache_name, key, scope, name, version)
 
       {:ok, manifests} ->
         manifests
 
       _ ->
-        discover(scope, name, version)
+        {_status, manifests} = discover(scope, name, version)
+        manifests
+    end
+  end
+
+  defp discover_and_maybe_cache(cache_name, key, scope, name, version) do
+    case discover(scope, name, version) do
+      {:ok, manifests} ->
+        Cachex.put(cache_name, key, manifests, ttl: @ttl)
+        manifests
+
+      {:error, manifests} ->
+        manifests
     end
   end
 
@@ -72,7 +82,7 @@ defmodule Cache.Registry.AlternateManifests do
             inspect(error)
         )
 
-        []
+        {:error, []}
     end
   end
 
@@ -80,28 +90,49 @@ defmodule Cache.Registry.AlternateManifests do
     alternate_objects =
       Enum.filter(objects, fn {_key, filename} -> ManifestVariants.alternate_manifest?(filename) end)
 
-    if alternate_objects == [] do
-      []
-    else
-      descriptors_with_alternates(objects, alternate_objects, bucket, scope, name, version)
-    end
+    if alternate_objects == [],
+      do: {:ok, []},
+      else: descriptors_with_alternates(objects, alternate_objects, bucket, scope, name, version)
   end
 
   defp descriptors_with_alternates(objects, alternate_objects, bucket, scope, name, version) do
-    default_descriptor =
-      objects
-      |> Enum.find(fn {_key, filename} -> filename == "Package.swift" end)
-      |> case do
-        nil -> []
-        {key, filename} -> descriptor_for(bucket, key, filename, scope, name, version)
-      end
+    default_object = Enum.find(objects, fn {_key, filename} -> filename == "Package.swift" end)
 
-    alternates =
-      Enum.flat_map(alternate_objects, fn {key, filename} ->
-        descriptor_for(bucket, key, filename, scope, name, version)
-      end)
+    case default_descriptor_for(default_object, bucket, scope, name, version) do
+      {:ok, default_descriptor} ->
+        alternates =
+          Enum.flat_map(alternate_objects, fn {key, filename} ->
+            descriptor_for(bucket, key, filename, scope, name, version)
+          end)
 
-    ManifestVariants.linkable_alternates(default_descriptor ++ alternates)
+        {:ok, ManifestVariants.linkable_alternates(default_descriptor ++ alternates)}
+
+      {:error, _reason} ->
+        {:error, []}
+    end
+  end
+
+  defp default_descriptor_for(nil, _bucket, _scope, _name, _version), do: {:ok, []}
+
+  defp default_descriptor_for({key, filename}, bucket, scope, name, version) do
+    case fetch_content(bucket, key) do
+      {:ok, content} ->
+        {:ok,
+         [
+           %{
+             "swift_version" => filename_swift_version(filename),
+             "swift_tools_version" => ManifestVariants.swift_tools_version(content)
+           }
+         ]}
+
+      {:error, reason} ->
+        Logger.warning(
+          "Failed to fetch default manifest #{key} for #{scope}/#{name}@#{version}: " <>
+            inspect(reason)
+        )
+
+        {:error, reason}
+    end
   end
 
   defp descriptor_for(bucket, key, filename, scope, name, version) do
--- a/cache/lib/cache/registry/alternate_manifests.ex
+++ b/cache/lib/cache/registry/alternate_manifests.ex
@@ -19,14 +19,12 @@ defmodule Cache.Registry.AlternateManifests do
 
   alias Cache.Config
   alias Cache.Registry.KeyNormalizer
+  alias Cache.Registry.ManifestVariants
 
   require Logger
 
   @cache_name :registry_alternate_manifests_cache
   @ttl to_timeout(minute: 10)
-  @alternate_manifest_regex ~r/\APackage@swift-(\d+)(?:\.(\d+))?(?:\.(\d+))?\.swift\z/
-  @swift_tools_version_regex ~r/^\/\/ swift-tools-version:\s?(\d+)(?:\.(\d+))?(?:\.(\d+))?/
-
   def child_spec(_opts) do
     %{
       id: __MODULE__,
@@ -65,10 +63,8 @@ defmodule Cache.Registry.AlternateManifests do
       |> ExAws.S3.list_objects_v2(prefix: prefix)
       |> ExAws.stream!()
       |> Stream.map(fn %{key: key} -> {key, Path.basename(key)} end)
-      |> Stream.filter(fn {_key, filename} -> alternate_manifest?(filename) end)
-      |> Enum.flat_map(fn {key, filename} ->
-        descriptor_for(bucket, key, filename, scope, name, version)
-      end)
+      |> Enum.to_list()
+      |> descriptors(bucket, scope, name, version)
     rescue
       error ->
         Logger.warning(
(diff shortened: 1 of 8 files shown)


cache/lib/cache/registry/manifest_variants.ex
```diff
@@ -0,0 +1,70 @@
...
+      end
+    end)
+    |> elem(1)
+    |> Enum.reverse()
+  end
+
+  defp canonical_tools_version(nil), do: nil
+
+  defp canonical_tools_version(version) when is_binary(version) do
+    case Regex.run(~r/\A(\d+)(?:\.(\d+))?(?:\.(\d+))?\z/, version) do
```
pepicrft, line 63: Canonical normalization is untested, and it is the core of this fix.

I verified this concretely: I temporarily made `canonical_tools_version/1` return its input unchanged (no padding to `major.minor.patch`), ran the three touched suites, and all 89 tests still passed. Every fixture feeds string-identical tools versions (default `"6.0"` vs alternate `"6.0"`), so the collision is caught by exact MapSet equality and the canonicalization branches for 1-, 2-, and 3-component versions are never required to make a test pass.

That means a regression here would go unnoticed. Suggest adding `cache/test/cache/registry/manifest_variants_test.exs` with cases like:
- default tools-version `"6.0"`, alternate `Package@swift-6.0.0.swift` with `"6.0.0"` -> alternate dropped
- default `"6"`, alternate with `"6.0.0"` -> alternate dropped
- `filename_swift_version("Package@swift-6.swift")` -> `"6"`

The existing integration tests are good for the wiring; they just do not lock in the equivalence semantics.

cache/lib/cache/registry/manifest_variants.ex
```diff
@@ -0,0 +1,70 @@
...
+    default_tools_versions =
+      manifests
+      |> Enum.filter(&is_nil(&1["swift_version"]))
+      |> Enum.map(&canonical_tools_version(&1["swift_tools_version"]))
+      |> Enum.reject(&is_nil/1)
+      |> MapSet.new()
+
+    manifests
+    |> Enum.reject(&is_nil(&1["swift_version"]))
+    |> Enum.reduce({default_tools_versions, []}, fn manifest, {seen_tools_versions, alternates} ->
```
pepicrft, line 42: Two branches in this reduce are untested:

1. Inter-alternate collision: the accumulating `seen_tools_versions` MapSet dedupes alternates against each other, not just against the default. But no test seeds two alternates that canonicalize to the same version (e.g. `Package@swift-6.swift` and `Package@swift-6.0.swift` both declaring `// swift-tools-version:6.0`). Every existing test only collides an alternate with a default.

2. The `is_nil(canonical_tools_version) -> keep` branch (preserve unparsable tools versions) and the `_ -> version` fallback in `canonical_tools_version/1` for malformed input like `"6.0.0-beta"`. This is a deliberate safe-by-default choice (never drop what you cannot prove is a duplicate); worth pinning with a test so a future change to "drop unparseable" is caught.

Also worth noting: first-seen alternate wins on collision, ordered by the S3 listing / metadata order. No test asserts which one survives, so a flip to last-wins would be silent.

cache/lib/cache/registry/alternate_manifests.ex
```diff
@@ -80,15 +76,41 @@ defmodule Cache.Registry.AlternateManifests do
...
+        nil -> []
+        {key, filename} -> descriptor_for(bucket, key, filename, scope, name, version)
+      end
+
+    alternates =
+      Enum.flat_map(alternate_objects, fn {key, filename} ->
+        descriptor_for(bucket, key, filename, scope, name, version)
+      end)
+
+    ManifestVariants.linkable_alternates(default_descriptor ++ alternates)
```
pepicrft: Minor reliability note on the S3 fallback path, not a blocker.

If the default `Package.swift` is present in the S3 listing but its GET fails transiently, `descriptor_for` returns `[]` and `default_descriptor` becomes `[]`, indistinguishable from "Package.swift absent". `linkable_alternates([] ++ alternates)` then runs with an empty `default_tools_versions` set, so an alternate that collides with the default is not filtered out and gets re-advertised via the Link header. The result is then cached for the 10-minute Cachex TTL (`list/4` line 44 puts whatever `discover` returned).

This is strictly better than pre-PR behavior (the fallback never deduped against the default at all), and inter-alternate dedup still works because the seen set accumulates across alternates. So it does not reintroduce the SwiftPM `duplicate key found` error. Just flagging that a transient S3 hiccup on `Package.swift` suppresses the default-vs-alternate dedup for the full TTL window.

If it matters, the fix is to distinguish "absent from listing" from "GET failed" and skip caching (or use a shorter TTL) in the degraded case. Low priority given this is a compatibility fallback for old releases missing the `manifests` field.

pepicrft (commented):
Did a deep review and validated the findings by running the suite and mutation-testing the core logic.

Verification: `mix test` on the three touched suites passes (89 tests, 0 failures). `mix format` and `mix credo` are clean. The dedup algorithm itself is correct, no correctness bugs found across three independent review passes.

The main thing worth addressing before merge: I temporarily replaced `canonical_tools_version/1` with a no-op (return the input string unchanged) and the full suite still passed. That means the PR's central mechanism, canonical tools-version equivalence (`6` == `6.0` == `6.0.0`), is not actually exercised by any test. Every fixture uses string-identical versions, so exact-match comparison passes regardless of canonicalization.

Details inline. No correctness blockers; the suggestions below are about pinning the behavior that is currently load-bearing but unverified.

pepicrft (approved):
Approving. The dedup algorithm is correct and I verified it across three independent review passes, plus mutation-tested the core logic (no-op'ing `canonical_tools_version/1` confirmed the canonicalization is currently untested, but the algorithm itself is sound). `mix test` passes 89/0, format and credo are clean.

The untested canonicalization is a follow-up gap, not a merge blocker: the behavior is correct, and the integration tests do cover the wiring through all three call sites (controller, ReleaseWorker, S3 fallback). A focused `manifest_variants_test.exs` would be a good follow-up to lock in the `6` == `6.0` == `6.0.0` equivalence and the preserve-unparseable branch.

cache/lib/cache/registry/manifest_variants.ex
fortmarek: Addressed in `7c28093356b`.

I added `cache/test/cache/registry/manifest_variants_test.exs` to exercise the canonical equivalence directly instead of relying only on integration wiring. The new tests cover:

- default tools version `6.0` with alternate tools version `6.0.0` being dropped
- default tools version `6` with alternate tools version `6.0.0` being dropped
- parsing `Package@swift-6.swift` as Swift version `6`

I re-ran the focused cache suite after the change: `94 tests, 0 failures`.

cache/lib/cache/registry/manifest_variants.ex
fortmarek: Addressed in `7c28093356b`.

The new `ManifestVariantsTest` now pins the reduce branches you called out:

- inter-alternate collision: `Package@swift-6.swift`, `Package@swift-6.0.swift`, and `Package@swift-6.0.0.swift` canonicalize to the same tools version, and the test asserts the first alternate survives
- missing tools version is preserved
- malformed tools version such as `6.0.0-beta` is preserved

That keeps the safe-by-default behavior explicit: we only drop an alternate when we can prove its canonical tools version has already been represented. I also re-ran `mix credo`, which found no issues.

cache/lib/cache/registry/alternate_manifests.ex
fortmarek: Addressed in `7c28093356b`.

I changed the fallback discovery path to distinguish a listed-but-unfetchable `Package.swift` from an absent root manifest. If the root manifest is listed but its GET fails, fallback discovery now returns no alternate links and skips caching that degraded result, so a transient S3 failure cannot populate Cachex with a response that missed default-vs-alternate deduplication.

I added a regression test covering that behavior by calling `AlternateManifests.list/4` twice and asserting S3 is hit twice, proving the failed fallback result was not cached.

PR #11399, opened 2026-06-19
fix(cli): allow scheme code coverage targets from external projects

### LINKED ISSUE #11570: Scheme code coverage targets from external/SPM projects fail project linting
### What happened?

Xcode supports gathering code coverage for targets from anywhere in the workspace (Test action → *"Gather coverage for some targets"*), including Swift Package Manager packages.

Tuist, however, rejects these configurations during project linting:

- Using a target name directly:

  > The target 'DesignSystem' specified in Scheme code coverage targets list isn't defined in the project.

- Using an explicit `TargetReference(projectPath:target:)` pointing to another project:

  > The target 'DesignSystem' specified in scheme 'Scheme' is not defined in the project named 'X'. Consider using a workspace scheme instead to reference a target in another project.

As a result, `tuist generate` fails even though the configuration is perfectly valid in Xcode.

**Expected behavior:** `codeCoverageTargets` in a project scheme should be allowed to reference targets from other projects in the workspace (including SPM-generated projects), matching what Xcode already supports. Invalid references should still be caught by the workspace-level graph validation.

### How do we reproduce it?

1. Create a workspace with two projects (or one project plus an SPM dependency that generates its own project, e.g. `DesignSystem`).
2. In a project scheme's Test action, set `codeCoverageTargets` to a `TargetReference(projectPath:target:)` pointing at a target defined in another project.
3. Run `tuist generate`.
4. See the linting error.

### Error log

```sh
The target 'DesignSystem' specified in scheme 'Scheme' is not defined in the project named 'X'. Consider using a workspace scheme instead to reference a target in another project.
```

### macOS version

26.5.2 (25F84)

### Tuist version

latest

### Xcode version

26.6 (17F113)

dogo:
Resolves https://github.com/tuist/tuist/issues/11570

## Summary

Allow `codeCoverageTargets` in project schemes to reference targets from other projects in the workspace, including SPM-generated projects.

## Why

Xcode already supports this. In a scheme's Test action, you can choose *"Gather coverage for some targets"* and select targets from anywhere in the workspace, including Swift Package Manager packages.

Tuist, however, was stricter and rejected these configurations during project linting:

* Using a target name directly:

  > The target 'DesignSystem' specified in Scheme code coverage targets list isn't defined in the project.

* Using an explicit `TargetReference(projectPath:target:)` pointing to another project:

  > The target 'DesignSystem' specified in scheme 'Scheme' is not defined in the project named 'X'. Consider using a workspace scheme instead to reference a target in another project.

As a result, `tuist generate` failed even though the configuration was perfectly valid in Xcode.

## Root cause

This happened because three different pieces of logic interacted in an unexpected way:

1. `TestAction+ManifestMapper.swift` resolves code coverage targets without an explicit `projectPath` to the local project.
2. `SchemeLinter.lintCodeCoverageTargets` only validates targets that exist in the current project.
3. `SchemeLinter.projectSchemeCantReferenceRemoteTargets` treats `codeCoverageTargets` as regular target dependencies and rejects references to targets outside the project.

At the same time, `GraphLinter.lintSchemesUnknownTargets` already validates scheme target references against the entire workspace graph and reports unresolved targets when needed.

## Change

This PR makes two focused changes in `SchemeLinter.swift`:

1. `lintCodeCoverageTargets` now only validates targets that belong to the current project. References to targets in other projects are left to the graph linter.
2. `projectSchemeCantReferenceRemoteTargets` no longer considers `codeCoverageTargets` when checking for cross-project dependencies, since code coverage targets are coverage configuration rather than actual build dependencies.

This keeps the existing validation for local targets (including typo detection) while allowing valid cross-project coverage configurations that Xcode already supports.

## User impact

Project schemes can now include code coverage targets from any project in the workspace, including SPM-generated projects, by using an explicit `TargetReference`.

Existing behavior remains unchanged for local targets, and invalid references are still caught by the workspace-level graph validation.

### DIFF
--- a/cli/Sources/TuistAutomation/XcodeBuild/XcodeBuildController.swift
+++ b/cli/Sources/TuistAutomation/XcodeBuild/XcodeBuildController.swift
@@ -22,7 +22,7 @@ public struct XcodeBuildController: XcodeBuildControlling {
     private let formatter: Formatting
     private let simulatorController: SimulatorController
     private let commandRunner: CommandRunning
-    private static let showBuildSettingsTimeout: Duration = .seconds(20)
+    private static let showBuildSettingsTimeout: Duration = .seconds(60)
 
     public init() {
         self.init(
--- a/cli/Sources/TuistGenerator/Linter/SchemeLinter.swift
+++ b/cli/Sources/TuistGenerator/Linter/SchemeLinter.swift
@@ -23,7 +23,11 @@ struct SchemeLinter: SchemeLinting {
             schemes: project.schemes,
             settings: project.settings
         ))
-        issues.append(contentsOf: lintCodeCoverageTargets(schemes: project.schemes, targets: Array(project.targets.values)))
+        issues.append(contentsOf: lintCodeCoverageTargets(
+            schemes: project.schemes,
+            targets: Array(project.targets.values),
+            project: project
+        ))
         issues.append(contentsOf: lintExpandVariableTarget(schemes: project.schemes, targets: Array(project.targets.values)))
         issues.append(contentsOf: projectSchemeCantReferenceRemoteTargets(schemes: project.schemes, project: project))
         return issues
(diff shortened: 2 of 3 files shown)


dogo:
Hi @fortmarek and @pepicrft 👋
I had a quick question about the PR review process.
This PR has been open for a few days without any feedback, so I was wondering if there's anything else I should do on my side (assign reviewers, add labels, etc.), or if it's just waiting in the review queue.

Just want to make sure I'm following the expected contribution process correctly. 
Thanks! 🙏

fortmarek (approved):
Looks good thanks! And sorry for the long review 🙇

PR #11406, opened 2026-06-22
fix(server): make table row navigation work in Safari

asmitbm:
Whole-row navigation relied on a `position: absolute` `::after` overlay on the first cell's link, stretched across a `position: relative` <tr>. Safari never establishes a containing block on a <tr>, so every row's overlay resolved against the <table> and covered the whole grid — the topmost row's overlay swallowed every click and hover, so clicking any row always opened the same one and hover states stuck.

Drive the whole-row click from the NooraTable hook instead: a single real anchor stays in the first cell (one tab stop, announced once) and clicks on the rest of the row are forwarded to it. The hit area is row-wide in every engine, Cmd/Ctrl-click still opens a new tab, and row text is now selectable.

Screenshot 2026-06-22 at 10 34 09

### DIFF
--- a/noora/js/Table.js
+++ b/noora/js/Table.js
@@ -149,6 +149,40 @@ export default {
     const table = this.container.querySelector("table");
     if (table) this.resizeObserver.observe(table);
 
+    // Whole-row navigation. Each navigable row carries a single real anchor
+    // (`[data-part="row-link"]`) in its first cell for keyboard and assistive tech; here we widen
+    // the pointer hit area to the whole row by forwarding a click on any passive part of the row
+    // to that anchor (LiveView's document-level listener picks up the synthetic click and performs
+    // the live navigation). This is done in JS rather than with a stretched CSS overlay because an
+    // absolutely-positioned overlay needs the `<tr>` as its containing block, and Safari/WebKit
+    // never makes a `<tr>` one — there the overlay escaped to cover the entire table and hijacked
+    // every click and hover.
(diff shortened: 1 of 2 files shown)


noora/js/Table.js
```diff
@@ -149,6 +149,42 @@ export default {
...
 
+    // Whole-row navigation. Each navigable row carries a single real anchor
+    // (`[data-part="row-link"]`) in its first cell for keyboard and assistive tech; here we widen
+    // the pointer hit area to the whole row by forwarding a click on any passive part of the row
+    // to that anchor (LiveView's document-level listener picks up the synthetic click and performs
+    // the live navigation). This is done in JS rather than with a stretched CSS overlay because an
+    // absolutely-positioned overlay needs the `<tr>` as its containing block, and Safari/WebKit
+    // never makes a `<tr>` one — there the overlay escaped to cover the entire table and hijacked
+    // every click and hover.
+    this.onRowClick = (e) => {
```
fortmarek, line 160: let's really try to minimize the amount of JS we use over trying to use browser capabilities. 

This breaks for example showing the URL this would navigate to in browsers like Firefox, but also Safari page previews.

fortmarek (commented):
Let's fix this in a follow-up, but the rows seem to be not vertically aligned well:

Image

noora/js/Table.js
fortmarek: Before:
image

After:
image

PR #11410, opened 2026-06-22
fix(server): fix status bar link on hover

asmitbm:
Fixed the alignment between table header and cell

Screenshot 2026-06-22 at 12 32 08

Fix link previews and page preview (in Safari)
Screenshot 2026-06-22 at 12 53 29
Screenshot 2026-06-22 at 12 58 46

## What the hook does, and whether a browser capability can replace it 

|  JS responsibility  | Native replacement?  |  Verdict | Image |
|---|---|---|---|
| Whole-row navigation | Real `<a href>` | ✅ Done — already removed | Screenshot 2026-06-22 at 12 47 54 |
| Floating header (fixed clone of thead) | `position: sticky` — but the thead sits inside `[data-part="scroll-container"]`'s `overflow-x: auto`, which becomes the sticky scrollport. That box is auto-height (the page scrolls vertically), so a sticky header has no vertical travel and just scrolls away.| ❌ Load-bearing in the current scroll model| Screenshot 2026-06-22 at 12 48 26 |
| Floating scrollbar proxy (webkit, pinned to viewport bottom) | None — browsers can't relocate a scrollbar to the viewport edge | ❌ Load-bearing | Screenshot 2026-06-22 at 12 49 16|
| Custom overlay scrollbar (Firefox / overlay-scrollbar platforms) | Partial — `scrollbar-width/color` already does the in-place bar; the floating version has no native equivalent | ❌ Load-bearing | Screenshot 2026-06-22 at 12 49 53 |
| Column-width pinning (grow-only anti-jump across LiveView swaps) | None — no CSS remembers the widest column across DOM patches| ❌ Load-bearing | Screenshot 2026-06-22 at 12 50 36 when you sort the columns, the column width stays consistent |
| `findScrollParent` | No native "which ancestor scrolls" API | ❌ Load-bearing |

The row navigation was the only piece with a clean native swap. Everything else backs behavior the browser has no native equivalent for given the current layout (page scrolls, header floats at the navbar, table can be any height).

### DIFF
--- a/noora/css/table.css
+++ b/noora/css/table.css
@@ -173,6 +173,12 @@
           }
         }
 
+        /* Navigable/clickable cells are the containing block for the stretched row-link overlay
+         * (`<td>` works in every engine; `<tr>` does not, which is why the overlay is per-cell). */
+        & td[data-selectable] {
+          position: relative;
+        }
+
         &:hover {
           & td[data-selectable] {
             background: var(--noora-surface-background-secondary);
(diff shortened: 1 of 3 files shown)


fortmarek (approved):
Thanks!

PR #11414, opened 2026-06-22
feat(kura): add zstd compression to REAPI ByteStream

esnunes:
## What

Adds negotiated **zstd wire compression** to Kura's REAPI gRPC surface (the Bazel/Buck2 Remote Execution API v2) for **ByteStream Read and Write**:

- **Read** — when a client requests a `compressed-blobs/zstd/{hash}/{size}` resource, the stored blob is streamed through a zstd encoder on the fly.
- **Write** — compressed uploads are decoded on the fly into the upload temp file.
- **GetCapabilities** — advertises zstd in `supported_compressors` when enabled.
- **Batch RPCs** — intentionally *not* offered compression: `BatchUpdateBlobs` rejects compressed items, `BatchReadBlobs` returns identity, and `supported_batch_update_compressors` stays empty.
- Gated by a new `KURA_REAPI_COMPRESSION_ENABLED` env/Helm flag (default `false`).

## Why

For WAN cache meshes, CAS blob transfer bytes dominate cost. Bazel's `--experimental_remote_cache_compression` sends and receives zstd-compressed blobs, but Kura previously advertised no compressors and rejected compressed uploads, so every blob crossed the wire uncompressed.

## Design decisions

- **No storage/format change.** Blobs are always stored *uncompressed* and keyed by their *uncompressed* digest. Compression is purely on the wire, per request, so it never touches the append-only on-disk segment format or the replication wire format (preserves the mmap-serving invariant).
- **Rollout-safe under a mixed fleet.** The flag gates *only* GetCapabilities advertisement; the ByteStream accept/serve paths handle compression **unconditionally**. So a half-rolled or mixed-flag fleet can never advertise support on one node and reject it on another. Two-phase rollout: deploy with the flag off, then flip the Helm value to `true` once every pod runs the new code.
- **Decompression-bomb safety.** The Write path feeds the streaming zstd decoder into a `BoundedBuffer` capped at the declared uncompressed size, so a malicious payload fails *during* decode with bounded memory (≈ limit + one zstd block) instead of materializing the full output first. (An earlier iteration checked size only *after* a chunk's full decoded output was already in memory — caught and fixed during review.)
- **Why ByteStream-only.** The Batch RPCs are small-blob-oriented and are not streaming — they hold whole blobs in memory. Supporting compressed batch updates meant an unbounded in-memory decompression path (a single item could allocate up to the declared size, with no per-request budget). Rather than add that machinery, compression is scoped to the streaming ByteStream paths, which stay at a few MB of RAM regardless of blob size (decompressed bytes land on the temp file, not the heap).

## Impact

With the flag enabled, Bazel/Buck2 clients negotiate zstd and transfer large CAS blobs compressed over ByteStream, cutting WAN bytes. Uncompressed clients are unaffected. No data migration; the on-disk and replication formats are unchanged.

## Validation

- `cargo test --lib` → **262 passing**, including: compressed write → compressed *and* plain read round-trip (proves storage stays uncompressed); single- and multi-chunk decompression-bomb rejection; post-decode digest mismatch; truncated frame; `read_limit` rejection on compressed reads; capability-gating; and a `BoundedBuffer` boundary unit test (`produced == limit` succeeds, `+1` fails).
- `cargo clippy --all-targets -- -D warnings` → clean.
- e2e shellspec adds a cross-region compressed Bazel build (`--experimental_remote_cache_compression`).
- Reviewed across three `/code-review` passes; the bomb-guard insufficiency and a dead-code vestige from the batch revert were both surfaced and fixed.

## Commits

1. `feat(kura): add zstd compression to REAPI ByteStream read/write` — implementation, config flag, deps, unit tests.
2. `feat(kura): wire REAPI zstd rollout via Helm, compose, e2e, and docs` — Helm value + StatefulSet env, docker-compose e2e enablement, e2e spec, README + architecture docs.

## Follow-ups (deferred, non-blocking)

- First-chunk compressed `write_offset` is validated against the compressed-byte counter; the spec defines the *initial* offset in uncompressed space. No practical impact (Bazel uploads start at offset 0).
- `QueryWriteStatus.committed_size` for a completed compressed resource returns the uncompressed size rather than `-1`.


esnunes:
superseded by https://github.com/tuist/tuist/pull/13104

PR #11427, opened 2026-06-22
fix(infra): drop the xcresult-processor nvram push workarounds (real fix was tart-kubelet GC)

fortmarek:
## TL;DR

The repeated `Error: The file "nvram.bin" doesn't exist` on the xcresult-processor push was **never a workflow bug**. The builder-fleet hosts run `tart-kubelet` *and* the GitHub Actions runner; `tart-kubelet`'s orphan-VM garbage collector was reaping the in-flight Packer build VM mid-`tart push` (every local VM looks orphaned on a builder node). Its own flag help says exactly this:

> The GC deletes every local Tart VM not backed by a Pod scheduled to this Node. On builder-fleet Nodes … the GC reaps the in-flight build VM mid-`tart push` (it fails at the NVRAM layer with `nvram.bin doesn't exist`). Set this on builder Nodes.

The fix is operational — `--disable-vm-gc` on the builders' `tart-kubelet` (already rendered by `infra/macos-host-bootstrap` per #11072, but the running hosts had a stale plist: binaries stream independently of bootstrap rolls since #11065, so they got the newer binary without the re-rendered plist). Applied to both builders; a verification build then pushed green **on attempt 1**.

## What this PR does

Two earlier PRs (#11417, #11419) chased the wrong layer, adding `TART_NO_AUTO_PRUNE` and a clone-the-base "restore nvram.bin" step to the push. With the GC reaping the *whole* bundle mid-push, the restore's `cp` failed (`No such file or directory` on the bundle dir) and the base-clone added ~10 min of disk pressure. None of it helped.

This strips both push steps back to a plain 3× retry — identical to `macos-xcode-image.yml`, which publishes the same-size image via the same path. Applied to the production release job (`server-production-deployment.yml`) and the dispatch workflow (`xcresult-processor-image.yml`).

## Verification

Dispatch build on this branch: [run 27983133232](https://github.com/tuist/tuist/actions/runs/27983133232) — `Pushed ghcr.io/tuist/tuist-xcresult-processor:e1554ce6 on attempt 1`.

## Follow-up (not in this PR)

The live plist edit on the two builders is operational, not durable. A full `macos-host-bootstrap` roll (not just a binary stream) is needed so re-rolled hosts keep `--disable-vm-gc` — worth checking why binary streams don't re-render the plist.

### DIFF
--- a/.github/workflows/server-production-deployment.yml
+++ b/.github/workflows/server-production-deployment.yml
@@ -983,45 +983,21 @@ jobs:
           TART_REGISTRY_HOSTNAME: ghcr.io
           TART_REGISTRY_USERNAME: tuist-bot
           TART_REGISTRY_PASSWORD: ${{ secrets.GITHUB_TOKEN }}
-          # tart's auto-prune fires opportunistically during `tart push`
-          # and races the upload of the source bundle (cirruslabs/tart#860).
-          # reclaim-tart-disk already trimmed the cache at the start of the
-          # job, so nothing needs pruning here.
-          TART_NO_AUTO_PRUNE: "1"
-          XCODE_VERSION: "26.5"
-        # Two failure modes have repeatedly broken this push, both
-        # surfacing as `Error: The file "nvram.bin" doesn't exist.`:
-        #   1. packer-plugin-tart v1.20.0 occasionally bakes a bundle
-        #      with no nvram.bin at all.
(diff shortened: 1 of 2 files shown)


fortmarek:
Folded into #11428, which now carries both the workflow cleanup and the tart-kubelet GC durability fix in one PR.

Issue #11430, opened 2026-06-23
kura: mmap residency tests fail on virtio-fs (Kata) runners — mincore reports sub-page files non-resident


esnunes:
## Summary

Three of Kura's `mmap` unit tests fail when run under `bazel test` on the `tuist-linux` runners but pass on `ubuntu-latest`. The root cause is **not** the Bazel sandbox, the compiler, memory pressure, or virtio-fs in general. It is a specific virtio-fs behavior:

> On **virtio-fs**, `mincore(2)` reports a file's final **partial page** (a page that extends past EOF because the file is **smaller than one page**) as **not resident**, even immediately after `write` + `fsync`. Pages that are fully backed by file data report resident normally.

`tuist-linux` runners are Kata Containers QEMU/Firecracker microVMs whose container filesystem is virtio-fs (`infra/runners-controller`, `infra/linux-runner-image`). `ubuntu-latest` is a normal VM with `ext4`.

Kura's `map_file_region` (`kura/src/mmap.rs`) gates mmap serving on `mincore` residency (`mapping_is_resident`). The three failing tests map **sub-page** files, so on virtio-fs their single mapped page (which extends past EOF) reports non-resident → `map_file_region` returns `Ok(None)` → the residency assertions panic.

## Affected tests

| Test | Location | File mapped |
|---|---|---|
| `mmap::tests::maps_unaligned_file_regions` | `kura/src/mmap.rs` | 16 bytes (`b"0123456789abcdef"`), region offset 3 len 8 |
| `store::tests::mmap_artifact_bytes_is_opportunistic_under_memory_pressure` | `kura/src/store.rs` | `b"hello"` (5 bytes) |
| `store::tests::mmap_artifact_bytes_maps_non_zero_segment_offsets` | `kura/src/store.rs` | `b"second-artifact-payload"` (23 bytes) |

All three map files smaller than a 4 KiB page.

```
---- mmap::tests::maps_unaligned_file_regions stdout ----
panicked at src/mmap.rs: freshly written region should be page-cache resident
---- store::tests::mmap_artifact_bytes_is_opportunistic_under_memory_pressure stdout ----
panicked at src/store.rs: normal memory pressure should permit mmap serving
test result: FAILED. 257 passed; 3 failed
```

## Reproduction

A standalone Rust app using Kura's exact residency code (`memmap2 0.9.10` + `libc::mincore`, `mapping_is_resident` copied verbatim) reproduces it deterministically.

`Cargo.toml`:
```toml
[...]

PR #11431, opened 2026-06-23
fix(kura): exempt the final partial page from the mmap residency gate

esnunes:
## What

`map_file_region` gates mmap serving on `mincore` residency (`mapping_is_resident`). It required **every** page queried by `mincore` to be resident — including the file's **final partial page**, the page that extends past EOF when the file size isn't a multiple of the page size.

This change exempts that final partial page: only pages **fully backed by file data** must be resident. A small pure helper, `fully_backed_pages_resident`, makes the rule unit-testable.

## Why

On **virtio-fs**, `mincore` reports the partial past-EOF page as **non-resident even when its bytes are cached**. So the gate needlessly returned `Ok(None)` and fell back to the streaming reader for any **sub-page (or unaligned-tail) file**. On a native filesystem (ext4) the partial page reports resident, so this was invisible — until Kura's Bazel tests ran on the virtio-fs-backed **Kata microVM `tuist-linux`** runners, where three mmap tests (mapping 16-byte / 5-byte / 23-byte files) panicked on `freshly written region should be page-cache resident`.

Full investigation and reproduction: #11430.

## Impact

- **Not a correctness change.** The reader fallback was always correct; this only restores the mmap fast path for sub-page files where `mincore` under-reports residency.
- The only behavioral change: the exempted partial page could fault once as a **single small read** if it were genuinely cold. Every fully-backed page still gates strictly, so large artifacts are essentially unaffected (only the ≤1-page tail is exempt).
- Makes the residency gate correct on virtio-fs-backed data dirs, not just native filesystems.

## Validation

- `cargo test --lib mmap::tests` — all pass, including the existing `maps_unaligned_file_regions` (the 16-byte sub-page case) and two new deterministic tests for `fully_backed_pages_resident` (the virtio-fs partial-page case, plus the fully-backed-must-be-resident case).
- `cargo clippy --lib -- -D warnings` and `cargo fmt --check` clean.

Refs #11430

### DIFF
--- a/kura/src/metrics.rs
+++ b/kura/src/metrics.rs
@@ -122,6 +122,7 @@ pub struct Metrics {
     initial_discovery_completed: Gauge,
     writer_lock_owned: Gauge,
     writer_lock_acquire_failures: Counter,
+    mmap_partial_page_exemptions: Counter,
 }
 
 #[derive(Default)]
@@ -268,6 +269,7 @@ impl Metrics {
         let initial_discovery_completed = Gauge::default();
         let writer_lock_owned = Gauge::default();
         let writer_lock_acquire_failures = Counter::default();
+        let mmap_partial_page_exemptions = Counter::default();
         let process_start_time_seconds = Gauge::<i64>::default();
         process_start_time_seconds.set(
             SystemTime::now()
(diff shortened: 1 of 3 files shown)


fortmarek (approved):

pepicrft (approved):
Reviewed the memory-map residency change and validated the focused tests, full Kura library test suite, and Clippy locally. The remaining store comment wording drift is non-blocking.

PR #11439, opened 2026-06-23
feat(infra): drift macOS host config on a fleet-wide config hash

fortmarek:
## What

The Scaleway Apple Silicon CAPI provider's drift loop re-pushes host config to already-Ready Mac minis only when the **tart-kubelet binary SHA** changes. This PR broadens the trigger to a fleet-wide **host-config hash** so a change to *anything the operator pushes* propagates on the next reconcile:

- New `bootstrap.HostConfigHash(cfg)` — a sha256 over every rendered install script (firewall + vmnat, PN interface, launchd job + plist, tailscale, node_exporter, tart-kubelet install) plus the SHA of each embedded binary (tart-kubelet, tailscale, node_exporter).
- The manager computes it once at startup (like the binary SHA) from a `Config` carrying only operator-image + fleet-config inputs, with **every per-host field zeroed** (NodeName, IP, kubeconfig, VLAN, auth key, `DisableVMGC`, ...), so the fingerprint is identical across the fleet.
- New `Status.HostConfigHash` on the CR (Go type **and** the helm-chart CRD schema — see below); the reconciler stamps it on a successful re-push.
- The drift gate moves from `binaryDrift` (tart-kubelet SHA only) to `configDrift` (the host-config hash). `TartKubeletBinarySHA` is still computed and stamped for observability.
- The install functions' inlined `script :=` strings are extracted into pure `render*(cfg) string` helpers shared by the installer and the hash — **what gets pushed to hosts is byte-for-byte unchanged**.

## Why

Last week's macOS runner cache outage needed three host-config fixes (a NAT vlan-detection fix, a pfctl direct-anchor-load fix, an EM pn0 hardening). Each landed in the operator image, but the drift loop ignored them: it only re-pushes on a tart-kubelet *binary* change, so a firewall/script-only or fleet-config-only fix never reached the already-provisioned hosts. The fix had to be hand-applied over SSH to all 9 hosts and the CRs force-patched to re-trigger the loop. This closes that gap: a script tweak, a fleet CIDR/tag/accept-routes change, or a re-baked tailscale/node_exporter binary now rolls to existing hosts automatically.

### Why a fleet-wide canonical hash (over per-host or operator-binary)

- **Per-host hash** would have to exclude volatile fields (kubeconfig token, auth key) by hand or it re-pushes on every token rotation. Zeroing *all* per-host fields and computing one fleet-wide hash sidesteps that entirely — there are no per-host inputs left to churn.
- **Hashing the operator binary** would fire on every unrelated provider code change (controller bugfixes, base-image bumps), re-pushing for no host-config reason. Hashing the rendered scripts + pushed binaries is the right granularity.

`TartTarball` is deliberately excluded from the binary set: Tart is bootstrap-only (the hypervisor can't be swapped under running VMs, so `UpdateTartKubelet` never re-installs it); a Tart bump rolls via Machine replacement, not config drift.

## CRD registration (important)

The `ScalewayAppleSiliconMachine` CRD status schema is **structural** and enumerates its properties with no `x-kubernetes-preserve-unknown-fields`, so the API server **prunes** any status field not in the schema. The new `hostConfigHash` is therefore added to the checked-in CRD (`infra/helm/tuist/crds/...scalewayapplesiliconmachines.yaml`, auto-applied per-env by the deploy pipeline). Without it, `Status.HostConfigHash` would be silently dropped on write → read back empty every reconcile → `configDrift` stays true → the loop re-pushes the host config **on every reconcile**. (The helm CRD is a hand-maintained artifact — it carries chart labels beyond raw controller-gen output — so the field is added surgically rather than by a lossy raw regen.)

## Migration

Existing Machines have an empty `Status.HostConfigHash`, so the first reconcile after this operator image rolls out drifts once and re-pushes — the intended one-time migration. The re-push is zero-downtime: running Tart VMs survive `UpdateTartKubelet` (`nohup`-detached, re-bound by the kubelet's startup state recovery). Terminal-failed CRs stay excluded until `Status.FailureReason` is cleared.

## Validation

- Both Go modules (`macos-host-bootstrap`, `cluster-api-provider-scaleway-applesilicon`): `go build ./...`, `go vet ./...`, `gofmt -l .` (clean), `go test ./...` all pass.
- New `bootstrap` tests: hash is stable for the same config, **independent of per-host fields** (incl. `VMCachePNVLAN`, `DisableVMGC`), **changes** on a fleet-config field (CIDR/tags/accept-routes) and on any embedded-binary change.
- New controller test `TestHostConfigDrift` covers match / mismatch / the empty-machine-hash migration case / empty-operator-hash (never drifts).
- CRD YAML re-parsed: `status.properties.hostConfigHash` present as a string alongside `tartKubeletBinarySHA`.
- Verified the `render*` extraction is byte-faithful via `git diff` (only `script :=` assignments and `RunCommand*` wrappers moved; no heredoc body changed).

### DIFF
--- a/infra/cluster-api-provider-scaleway-applesilicon/AGENTS.md
+++ b/infra/cluster-api-provider-scaleway-applesilicon/AGENTS.md
@@ -67,6 +67,18 @@ redoing finished ones. Failures requeue with backoff; only terminal
 errors (Scaleway 400s, validation failures) set
 `Status.FailureReason`.
 
+Beyond first bootstrap, a drift loop re-pushes host config to already-Ready
+hosts when the operator's `bootstrap.HostConfigHash` differs from the
+Machine's `Status.HostConfigHash`. That hash is a fleet-wide fingerprint over
+everything the operator pushes — the rendered install scripts plus the
+embedded binaries (tart-kubelet, tailscale, node_exporter) — computed once at
+startup from operator-image + fleet-config inputs with every per-host field
+zeroed. So shipping a new operator image with a changed script, fleet CIDR/tag,
+or re-baked binary rolls to existing hosts on the next reconcile, not only on a
+tart-kubelet binary change. The re-push is zero-downtime (running Tart VMs
+survive `UpdateTartKubelet`). Terminal-failed CRs are excluded until
+`Status.FailureReason` is cleared.
+
 Two auxiliary controllers run alongside it:
 
 - **OrphanReclaimer** (`controllers/orphan_reclaimer.go`) — a
--- a/infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
+++ b/infra/cluster-api-provider-scaleway-applesilicon/api/v1alpha1/scalewayapplesiliconmachine_types.go
@@ -205,6 +205,16 @@ type ScalewayAppleSiliconMachineStatus struct {
 	// +optional
 	TartKubeletBinarySHA string `json:"tartKubeletBinarySHA,omitempty"`
 
+	// HostConfigHash is the fleet-wide canonical hash of every host
+	// config the operator pushes — the rendered install scripts plus the
+	// embedded binaries (bootstrap.HostConfigHash). Drift between this
+	// and the operator's own computed hash re-pushes the host config on
+	// the next reconcile, so a change to ANY pushed config (a script
+	// tweak, a fleet CIDR, or a re-baked binary) rolls to existing hosts
+	// instead of only a tart-kubelet binary change.
+	// +optional
+	HostConfigHash string `json:"hostConfigHash,omitempty"`
+
 	// TartKubeletUpdateAttempts counts consecutive failures of the
 	// drift-loop's UpdateTartKubelet call. Reset to zero on success.
 	// Once it crosses the operator's max-attempts threshold the CR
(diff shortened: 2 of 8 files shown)


pepicrft (approved):
Approving. I validated the rollout trigger shape against the Kubernetes ecosystem, and content hashes are the normal way to make configuration changes observable.

Examples we can lean on when the inputs live in manifests:

- Kubernetes Deployments use a pod template hash: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#pod-template-hash
- Helm documents checksum annotations with `sha256sum` for config-driven rollouts: https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments
- Kustomize appends content hashes to generated ConfigMaps and Secrets: https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/#generatoroptions
- `kubectl create configmap --append-hash` supports the same idea for generated ConfigMaps: https://kubernetes.io/docs/reference/kubectl/generated/kubectl_create/kubectl_create_configmap/#options

For this host drift loop, the inputs are rendered inside the controller rather than by Helm or Kustomize, so a deterministic Go digest is the right fit. The important thing is to keep the hash over canonical desired host state, not raw per-host or secret values. If a rendered branch should trigger drift, such as Tailscale being enabled or builder hosts needing virtual machine garbage collection disabled, I would model that as an explicit non-secret input to the hash.

PR #11455, opened 2026-06-23
feat(server): reap runner jobs stranded in the queued phase

fortmarek:
## What changed

Adds `StaleQueuedJobsWorker`, a recovery worker that resolves `runner_jobs` rows stranded in the `queued` phase. It runs on the hosted-only crontab every 5 minutes.

A `workflow_job` recorded in the ClickHouse `runner_jobs` table advances through `queued → claimed → running → completed`, one INSERT per transition. Recovery previously existed only for the later states:

- `StaleClaimsWorker` reaps Postgres `claimed` rows after 5 min.
- `OrphanedRunnersWorker` reaps CH `running` rows by cross-checking GitHub.
- `WebhookRedeliveryWorker` re-requests failed webhook deliveries, but only within a 15-minute window.

Nothing reaped the `queued` state.

## Why

A row stranded in `queued` is invisible to every existing recovery path: it has no Postgres claim (so `StaleClaimsWorker` can't see it) and never reaches `running` (so `OrphanedRunnersWorker` can't see it). The only thing that moves a queued row to a terminal state is the `workflow_job.completed` webhook. When that webhook never fires (GitHub kept the job queued on its side because no self-hosted runner ever accepted it) or was lost past the 15-minute redelivery window, the row sits at `queued` on the dashboard indefinitely.

This surfaced as jobs showing `Queued` for 28 days in the Runners dashboard, with GitHub itself also reporting the same jobs as queued (never picked up). The root cause is that the lifecycle design implicitly trusted GitHub to always deliver a terminal `completed` event; when that assumption breaks, the row is orphaned forever.

## How the fix works

`StaleQueuedJobsWorker` lists queued rows older than 1h (`Jobs.list_stale_queued/1`) and, per row, cross-checks GitHub the same way `OrphanedRunnersWorker` does (`GET /repos/{owner}/{repo}/actions/jobs/{id}`):

| GitHub status | Action |
|---|---|
| `completed` | Reconcile — mark completed with GitHub's real conclusion (we missed the webhook) |
| `404` (pruned) | Complete — it cannot be live |
| `in_progress` | Leave it — a runner accepted it; it will fire `completed` within GitHub's per-job limit |
| `queued` | Leave it — still legitimately pending, unless past the hard backstop |

**Hard backstop at 24h:** any row queued longer than 24h that GitHub still reports `queued`, cannot be addressed (empty repository on legacy pre-profiles rows), or cannot be verified (GitHub API down) is force-completed with conclusion `"stale"`. Past that age nothing will ever move it, so the guarantee that a job cannot stay stuck in `queued` holds even when GitHub never resolves it. `in_progress` is the only state never force-completed (it is provably live). The worker defensively frees any leaked Postgres claim (PG-first) before recording the CH terminal state, matching the webhook completion path.

### Why a GitHub cross-check rather than a pure time-based reaper

Cross-checking yields the correct conclusion for the recoverable case (a missed `completed` webhook shows `success`/`failure` rather than a synthetic `stale`), and it mirrors the existing `OrphanedRunnersWorker` philosophy that GitHub's status is the source of truth. The time-based backstop is retained purely as the airtight guarantee for the unrecoverable case.

## Impact

No job can remain in the `queued` phase indefinitely. Genuinely-pending jobs (account at cap, pool scaling) are left untouched until 1h, and even then only acted on when GitHub agrees they are terminal or the 24h backstop fires. Steady-state GitHub API cost is near nil because the candidate set is normally empty.

## How to test locally

```
cd server
mix test test/tuist/runners/workers/stale_queued_jobs_worker_test.exs \
         test/tuist/runners/jobs_test.exs \
         test/tuist/oban/runtime_config_test.exs
```

Validation run: 64 tests pass (worker branch coverage including the 24h backstop reap, the lookup-failure backstop, and the no-repository path; `list_stale_queued/1`; and crontab membership), credo clean, format clean.


server/lib/tuist/runners/jobs.ex
```diff
@@ -1030,6 +1030,42 @@ defmodule Tuist.Runners.Jobs do
...
+  covers `queued`.
+
+  Returns the fields the worker needs to address GitHub's Actions
+  jobs API (`repository`) and to apply the hard backstop
+  (`enqueued_at`).
+  """
+  def list_stale_queued(%DateTime{} = threshold) do
+    ClickHouseRepo.all(
+      from(j in Job,
+        where: j.enqueued_at < ^threshold,
```
pepicrft: This is the only part I would keep an eye on operationally. Even if the candidate count is near zero, this still has to aggregate historical `runner_jobs` rows older than the threshold before `argMax` can decide which jobs are currently queued. If this shows up in ClickHouse, I would bound the lookback or move current runner-job state into a cheaper read model.

server/lib/tuist/runners/workers/stale_queued_jobs_worker.ex
```diff
@@ -0,0 +1,203 @@
...
+  alias Tuist.Accounts
+  alias Tuist.GitHub.Client, as: GitHubClient
+  alias Tuist.Runners.Claims
+  alias Tuist.Runners.Jobs
+  alias Tuist.Runners.Telemetry
+  alias Tuist.VCS
+
+  require Logger
+
+  @verify_after_seconds 3_600
```
pepicrft, line 91: The threshold split feels right to me. One hour keeps normal queueing away from the GitHub lookup path, and the 24-hour backstop lines up with GitHub's self-hosted runner queue limit.

server/lib/tuist/runners/workers/stale_queued_jobs_worker.ex
```diff
@@ -0,0 +1,203 @@
...
+
+  defp reap_if_past_backstop(_candidate, _reap_threshold), do: false
+
+  # Free the PG cap slot first (defensive — a queued row normally has
+  # no claim, but the StaleClaimsWorker CH-first crash window can leave
+  # CH=`queued` / PG=`claimed`), then record the terminal state in CH
+  # for customer visibility. Both are idempotent. A DB error crashes
+  # the tick; the cron retries in 5 min and the row stays queued
+  # meanwhile, which is harmless.
+  defp complete(workflow_job_id, conclusion, kind) do
```
pepicrft, line 206: Keeping this in the same PostgreSQL-first order as the webhook completion path is the right tradeoff. Capacity is freed immediately, while the ClickHouse transition remains an idempotent visibility update.

pepicrft (approved):
Approved. I left a few non-blocking notes on the pieces where the recovery contract and operational shape matter most. The overall approach looks right for the job: GitHub stays the source of truth, and the backstop prevents queued rows from lingering forever.

server/lib/tuist/runners/jobs.ex
fortmarek: Good call. Added an `enqueued_after` floor (7d, set by the worker) so the query prunes to a week of `enqueued_at` partitions instead of aggregating the full `runner_jobs` history. The floor sits well beyond the 24h backstop, so a stuck job is always reaped while still inside the window (the worker runs every 5 min, so it would have to be down ~6 consecutive days for a job to escape). Done in f93516834a1.

server/lib/tuist/runners/workers/stale_queued_jobs_worker.ex
fortmarek: Kept the split as-is. The 1h verify threshold keeps normal queueing off the GitHub lookup path, and the 24h backstop lines up with GitHub's self-hosted runner queue limit, as you noted. The follow-up commit (f93516834a1) also adds a 7d `enqueued_at` floor purely to bound the ClickHouse scan; it sits far beyond the 24h backstop so it never interferes with the reaping window.

server/lib/tuist/runners/workers/stale_queued_jobs_worker.ex
fortmarek: Agreed, kept the PostgreSQL-first ordering to match the webhook completion path (`Tuist.Runners.Dispatch.mark_completed`): the cap slot is freed immediately, and the ClickHouse transition stays an idempotent visibility update. No change.

PR #11483, opened 2026-06-24
ci(cli): auto-cut an RC on every push to a release branch

fortmarek:
## Motivation

Today, getting `X.Y.0-rc.(N+1)` out after a backport merges requires a human to manually dispatch the **CLI Release Candidate** workflow with `branch=releases/X.Y.x`. That means teams wait on a manual step before they can retest a fix on an RC.

## Change

Add a `push` trigger on `releases/*.x` to `cli-rc.yml`, so **every commit that lands on a release branch** (i.e. each merged backport PR) auto-publishes the next RC. Mode 3 in the workflow header. The existing manual modes are unchanged:

1. **Cut a new line** — manual dispatch, empty `branch` (rc.1 from main + create the release branch).
2. **Force an iteration** — manual dispatch, `branch=releases/X.Y.x` (still available, e.g. to re-publish a failed run).
3. **Auto-iterate on push** *(new)* — a commit on `releases/X.Y.x` publishes `rc.(N+1)`.

### Avoiding a duplicate cut on line creation

The new-line flow publishes `rc.1` and *then* pushes `releases/X.Y.x` at that same commit — which would itself fire the new `push` trigger. A `gate` step skips the run when the branch HEAD is already the commit the latest RC points at (the creation push, or any re-run with no new commits), so no duplicate RC is cut. `create-branch` is now scoped to manual new-line cuts (`github.event_name == 'workflow_dispatch' && inputs.branch == ''`) so it never runs on push.

### No trigger loop

Publishing creates **tags** + a GitHub prerelease, never a branch push, and the `branches:` filter ignores tag pushes — so an auto-cut RC can't re-trigger the workflow. `cli-build-publish.yml` contains no `git push` to a branch.

## Notes / follow-ups

- **Activation on the current line:** GitHub runs the workflow file *as it exists on the pushed branch*, so this must be cherry-picked onto active `releases/*.x` branches (e.g. `releases/4.201.x`) — same backport flow — to take effect there. New lines cut from `main` inherit it automatically. Happy to open that backport once this merges.
- **Scope:** this fires on *every* commit to a release branch, per the goal of "an RC per commit". If RC noise becomes a concern, a `paths:` filter (e.g. `cli/**`, `Package.resolved`, the submodule pointer, build scripts) could be added later.
- Validated with `actionlint` (only the pre-existing `tuist-linux` self-hosted-runner-label warning).

### DIFF
--- a/.github/workflows/cli-rc.yml
+++ b/.github/workflows/cli-rc.yml
@@ -2,23 +2,34 @@ name: CLI Release Candidate
 
 # Publish a release candidate (X.Y.0-rc.N) for the upcoming stable minor.
 #
-# Two modes, selected by the `branch` input:
+# Three modes:
 #
-#   1. Cut a new line (branch empty): publishes X.Y.0-rc.1 for the current canary
-#      minor and, only after that publish succeeds, creates the protected
-#      releases/<major>.<minor>.x branch at the built commit. The moment that RC
-#      tag exists, main's canary advances to the next minor, so main keeps moving
-#      while the line soaks. The branch is created last so a failed publish leaves
-#      no half-cut line: re-running with an empty branch starts cleanly.
+#   1. Cut a new line (manual dispatch, branch empty): publishes X.Y.0-rc.1 for
+#      the current canary minor and, only after that publish succeeds, creates
+#      the protected releases/<major>.<minor>.x branch at the built commit. The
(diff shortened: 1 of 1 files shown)


fortmarek:
Closing for now to keep things simple — may revisit later. Leaving the branch `ci/auto-cut-rc-on-release-branch-push` in place so the work isn't lost.

PR #11486, opened 2026-06-24
fix(cli): support nib resources in SwiftPM packages

DanilShkarupin:
## What

Adds support for `.nib` resources in Swift Packages, on par with `.xib`:

1. `PackageInfoMapper` now recognizes `.nib` in `defaultSpmResourceFileExtensions`, so packages without explicit `resources:` declarations get `.nib` files picked up by the auto-generated globs.
2. `AbsolutePath.isOpaqueDirectory` now includes `.nib`, so the compiled `.nib` wrapper is bundled as a single reference.

## Why

A compiled `.nib` is usually a directory wrapper (`wrapper.nib`) containing `keyedobjects.nib` and a runtime-versioned variant. It must be copied whole — the way SwiftPM and Xcode treat it. Without the opaque-directory handling, resource globbing descends into the wrapper and reassembles it as a group of its inner files, producing a malformed bundle that crashes at runtime when the nib is loaded.

## Impact

Swift Packages shipping `.nib` resources now have them bundled correctly as wrapper references and load without crashing.

## Validation

- `xcodebuild build … -scheme tuist` succeeds.
- `xcodebuild test … -only-testing TuistLoaderTests/PackageInfoMapperTests -only-testing TuistSupportTests/AbsolutePathExtrasTests` passes.
- Confirmed on a real project that the previously-crashing `.nib` is now bundled and loads correctly.

Tests updated: `PackageInfoMapperTests` and `AbsolutePath+ExtrasTests`.

## Scope note

Targets the SwiftPM path only. `.xib` also appears in `Target.validResourceExtensions` (buildable/synchronized folders in first-party projects); `.nib` parity there is intentionally left out to keep this PR focused.

### DIFF
--- a/cli/Sources/TuistSupport/Extensions/AbsolutePath+Extras.swift
+++ b/cli/Sources/TuistSupport/Extensions/AbsolutePath+Extras.swift
@@ -85,6 +85,7 @@ extension AbsolutePath {
             "mlmodelc",
             "xcmappingmodel",
             "icon",
+            "nib",
         ]
         .contains(self.extension ?? "")
     }
--- a/cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift
+++ b/cli/Sources/TuistLoader/SwiftPackageManager/PackageInfoMapper.swift
@@ -1888,6 +1888,7 @@ extension ProjectDescription.ResourceFileElements {
     /// Check https://developer.apple.com/documentation/swift_packages/bundling_resources_with_a_swift_package
     private static let defaultSpmResourceFileExtensions = Set([
         "xib",
+        "nib",
         "storyboard",
         "xcdatamodeld",
         "xcmappingmodel",
--- a/cli/Sources/XcodeGraph/Sources/XcodeGraph/Models/Target.swift
+++ b/cli/Sources/XcodeGraph/Sources/XcodeGraph/Models/Target.swift
@@ -28,7 +28,7 @@ public struct Target: Equatable, Hashable, Comparable, Codable, Sendable {
     ]
     public static let validHeaderExtensions: [String] = ["h", "hpp", "hh", "hxx"]
     public static let validFolderExtensions: [String] = [
-        "framework", "bundle", "app", "xcassets", "appiconset", "scnassets",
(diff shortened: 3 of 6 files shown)


cli/Sources/TuistSupport/Extensions/AbsolutePath+Extras.swift
```diff
@@ -85,6 +85,7 @@ extension AbsolutePath {
             "mlmodelc",
             "xcmappingmodel",
             "icon",
+            "nib",
```
fortmarek, line 88: This makes `PackageInfoMapper` emit `.glob(View.nib)` for explicit SwiftPM `.process("View.nib")` / `.copy("View.nib")` wrapper directories, but that glob is later unfolded through `XcodeGraph.Target.isResource`. That filter only accepts package directories, non-directories, or extensions in `Target.validFolderExtensions`; `.nib` is not in that folder allowlist, and `UTType(filenameExtension: "nib").conforms(to: .package)` returns `false` on macOS. So an explicit `.nib` directory wrapper still gets dropped before generation. The new test currently creates a plain `file.nib`, which does not cover the real wrapper case. Could we allow `.nib` directory resources in the downstream filter, likely via the folder allowlist or an opaque-directory-aware branch, and add a test with `View.nib/keyedobjects.nib`?

fortmarek (commented):
Found one issue to address before this fully fixes the SwiftPM `.nib` wrapper case.

cli/Sources/TuistSupport/Extensions/AbsolutePath+Extras.swift
DanilShkarupin: Hey, I tested this in Playground. You are correct that `UTType(filenameExtension: "nib").conforms(to: .package)` returns `false` on macOS. But the method used in code actually returns `true`. Probably that's why my implementation worked for me when I tested it on example app.

I think it still makes sense to introduce the change to make behaviour explicit, but let me know what you think

Screenshot 2026-06-25 at 12 59 49

DanilShkarupin: @fortmarek pushed the change, let me know your thoughts

PR #11491, opened 2026-06-25
fix(server): cancel unprocessable build jobs instead of retrying

fortmarek:
## Context

Sentry [TUIST-10H](https://tuist.sentry.io/issues/130210055/) fired with:

```
Oban.PerformError: Tuist.Builds.Workers.ProcessBuildWorker failed with
{:error, "The line lications/Xcode.app/C doesn't seem like a valid SLF line"}
```

The fragment `lications/Xcode.app/C` is a truncated slice of `/Applications/Xcode.app/C…` — the uploaded `.xcactivitylog` archive was corrupt/truncated, so the Swift `XCLogParser` lexer (via `Tuist.Processor.XCActivityLogNIF`) can't parse it.

## Problem

The download had already succeeded, so the same bytes get re-downloaded and re-parsed on every retry and fail identically. `ProcessBuildWorker` returned `{:error, reason}`, which Oban treats as a `:failure`:

- it retries **up to 5×**, all guaranteed to fail the same way, and
- each attempt emits `[:oban, :job, :exception]`, which the Sentry logger handler captures — so a single bad upload floods error tracking.

## Fix

Distinguish a **permanent** parse failure from a **transient** one (download / project lookup):

- `process_build/4` tags a post-download parse error as `{:error, {:unprocessable, reason}}`.
- `perform/1` handles that by marking the build `failed_processing` **once** and returning `{:cancel, reason}`. Per Oban's executor, cancelled jobs don't retry and emit `[:oban, :job, :stop]` (not an exception), so they don't alert.
- Download/project-lookup errors keep their existing retry-until-final-attempt behaviour.

A truly truncated archive has no recoverable data, so handling it at the worker layer (mark failed, stop, don't alert) is the right place — making the upstream Swift parser tolerant wouldn't change the outcome.

## Tests

TDD: added a test replaying the exact Sentry error on attempt 1/5 — it must mark the build failed immediately and return `{:cancel, reason}`. Updated the VCS-comment-on-failed-processing test, whose intent is unchanged, to the new `{:cancel, _}` semantics. All worker tests pass; `mix format`, `mix credo`, and `mix compile --warnings-as-errors` are clean.

### DIFF
--- a/server/lib/tuist/builds/workers/process_build_worker.ex
+++ b/server/lib/tuist/builds/workers/process_build_worker.ex
@@ -59,6 +59,17 @@ defmodule Tuist.Builds.Workers.ProcessBuildWorker do
           _ -> :ok
         end
 
+      {:error, {:unprocessable, reason}} ->
+        # The archive can't be parsed (e.g. a corrupt/truncated xcactivitylog).
+        # Re-downloading the same bytes and re-parsing always fails the same
+        # way, so cancel the job instead of retrying: this marks the build
+        # failed once, skips the remaining attempts, and keeps the permanent
+        # failure out of error tracking (cancelled jobs don't alert).
+        Logger.warning("Build #{build_id} archive is unprocessable, not retrying: #{inspect(reason)}")
+        mark_failed_build_processing(build_id, project_id, account_id, build_metadata)
+
+        {:cancel, reason}
+
       {:error, reason} ->
         if attempt >= max_attempts do
           Logger.error("Build processing failed permanently for build #{build_id}: #{inspect(reason)}")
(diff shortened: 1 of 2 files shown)


fortmarek:
Closing — the root cause is **not** a corrupt archive or an XCLogParser bug, so cancelling/marking `failed_processing` on parse errors is the wrong fix (it would turn a self-healing transient into permanent data loss).

Verified against the actual artifact (Notion build `8fce77ae`): the xcactivitylog is valid and parses cleanly with XCLogParser 0.2.48 on macOS, stock Linux, and a faithful reconstruction of the prod runtime image. The build was processed **successfully** (`status=success`, `duration=4658`) — a row was inserted at 07:27:22.091, ~1s *before* the 07:27:23 error.

Two jobs ran concurrently for the same build (the get_or_create_build duplicate-enqueue race). They share a download temp path keyed only on `build_id` (`/tmp/build_<build_id>.zip`); one job's download/`File.rm` clobbered the other mid-read, corrupting the *local* copy → spurious `invalidLine`. Real fix (unique temp path per job) coming in a separate PR.

PR #8962, opened 2025-12-19
refactor(cli): rename fixtures to examples and simplify fixture handling

pepicrft:
In preparation for adding the first pieces to support Android builds (Gradle) I'm doing some house-keeping work in the repo:

- Since `fixtures` are no longer tied to `cli`, I hoisted them from `cli/Fixtures` to `examples`.
- I renamed them to `examples` since it's a more discoverable term when looking at the repository on GitHub. We also refer to them as examples in our docs, so I think it makes more sense to use that term.
- I moved the existing examples under `examples/xcode` and created `examples/gradle` where I plan to start adding the Gradle examples.
- I renamed all the Xcode fixtures to have the `generated_...` prefix to differenciate them from those that start with `xcode_project`
- I removed the enum that we have to match a case to a string, since it brings no value.

PR #11513, opened 2026-06-26
fix(cli): allow UI test targets on visionOS

### LINKED ISSUE #11512: visionOS UI test (.uiTests) targets are rejected at generation
### What happened?

tuist generate fails for a .uiTests target with visionOS destinations, even though Apple supports XCUITest on visionOS. Two model gaps cause it:
- GraphLinter.validLinks has no (.visionOS, .uiTests) entry, so lintDependency can't find a valid link and errors out.
- Product.forPlatform(.visionOS) returns .unitTests but never .uiTests.

Expected: visionOS UI test targets generate like iOS/tvOS/macOS do.

### How do we reproduce it?

1. In Project.swift, declare a visionOS app and a visionOS UI test target:
.target(name: "App", destinations: [.appleVision], product: .app, bundleId: "dev.tuist.App"),
.target(name: "AppUITests", destinations: [.appleVision], product: .uiTests,
        bundleId: "dev.tuist.AppUITests", dependencies: [.target(name: "App")]),
2. Run tuist generate.
3. Generation fails with the error below.

### Error log

Target AppUITests has platform 'visionOS' and product 'uiTests' which is an invalid or not yet supported combination.

### macOS version

 N/A — platform-model issue, not OS-specific

### Tuist version

 4.200.5

### Xcode version

N/A — not Xcode-specific

kikeenrique:
Resolves <https://github.com/tuist/tuist/issues/11512>

### Summary

`tuist generate` rejects `.uiTests` targets on visionOS. Two entries are missing
from Tuist's platform model — one in `GraphLinter.validLinks`, one in
`Product.forPlatform` — and this adds both, with unit tests.

While adding a UI test target to the `generated_visionos_app` example to prove
it, I found that example hasn't compiled for visionOS since #8962 (`UIScreen` is
unavailable on the platform). So the example is fixed too: migrated to the UIKit
scene-based life cycle, with a UI test that now runs green on a visionOS 26.5
Simulator. That run is the evidence for this PR; everything below is detail.

### Why

Xcode builds and runs XCUITest targets against visionOS-SDK apps in practice —
verified on a visionOS 26.5 Simulator with Xcode 26.6 (see *How to test locally*).
Tuist, however, refuses to generate them:

```
Target AppUITests has platform 'visionOS' and product 'ui tests' which is an
invalid or not yet supported combination.
```

Apple's documentation is ambiguous here. The XCUIAutomation framework page still
carries a note saying UI testing *"isn't available to apps you build using the
visionOS SDK"*, while [`XCUIApplication`'s availability
metadata](https://developer.apple.com/documentation/xcuiautomation/xcuiapplication)
lists visionOS with no unavailable or beta flag. The note appears to predate
current toolchains.

Either way, it seems the graph linter's platform table could use an update here:
it already permits `(.watchOS, .uiTests)` and `(.visionOS, .unitTests)`, and
since Xcode accepts the visionOS UI test configuration, generating it and letting
Xcode decide would be consistent with the other platforms.

### Root cause

- `GraphLinter.validLinks` had no `(.visionOS, .uiTests)` key, so `lintDependency`
  fell through its `validLinks` lookup and raised the error above.
- `Product.forPlatform(.visionOS)` returned `.unitTests` but never `.uiTests`.

### Changes

- Add a `(.visionOS, .uiTests)` entry to `GraphLinter.validLinks`, mirroring the
  existing visionOS `.unitTests` entry.
- Append `.uiTests` for `.visionOS` in `Product.forPlatform`.
- Unit tests: `GraphLinterTests.lint_when_visionOS_UITests_depends_on_app` and
  `ProductTests.test_forPlatform_when_visionOS`.
- Add an `AppUITests` target to the `generated_visionos_app` example, and fix the
  example so that target is actually exercisable.

> Note: `Product.swift` lives in the embedded `tuist/XcodeGraph` package, so that
> half ideally also lands upstream to avoid drift on the next sync.

### Fixing the example

The `AppUITests` target alone proved nothing: `generated_visionos_app` did not
compile for its own platform. `AppDelegate.swift` used
`UIWindow(frame: UIScreen.main.bounds)`, and `UIScreen` is unavailable in
visionOS, so the `App` target failed to build and anything depending on it failed
before a test could run. That looks like the reason this example's acceptance test
has its `BuildCommand` commented out behind a `TODO: Fix`.

Rather than swapping `UIScreen.main.bounds` for a bare `UIWindow()`, the example
moves to the UIKit scene-based life cycle — [Apple's migration
guide](https://developer.apple.com/documentation/uikit/transitioning-to-the-uikit-scene-based-life-cycle)
notes that as of visionOS 26 UIKit logs *"UIScene lifecycle will soon be
required"* for apps that haven't migrated, so the minimal patch would have left
the example on a deprecated path. `AppDelegate` now vends a `UISceneConfiguration`
and a new `SceneDelegate` builds the window with `UIWindow(windowScene:)`, which
needs no `UIScreen`.

`Info.plist` gains `UIApplicationSceneManifest` and drops iOS-template leftovers
that were wrong for visionOS: `LSRequiresIPhoneOS`, `UIRequiredDeviceCapabilities`
of `armv7` (visionOS is arm64-only), and the iPhone/iPad orientation arrays.

`AppUITest` replaces the empty `testExample()` stub with a test that launches the
app, waits for the root view by accessibility identifier, and asserts the app is
running in the foreground (an empty test passes without proving anything).

### How to test locally

Steps 1, 2 and 4 exercise Tuist itself on the host; step 3 runs the generated
project on a visionOS Simulator.

**1. Reproduce the bug** with a released Tuist, against this branch's example:

```sh
tuist generate --path examples/xcode/generated_visionos_app --no-open
# ✖ Target AppUITests has platform 'visionOS' and product 'ui tests' which is an
#   invalid or not yet supported combination.
```

**2. Build this branch and regenerate** — generation now succeeds and
`AppUITests` is present in the generated project:

```sh
tuist generate tuist ProjectDescription --no-open
xcodebuild build -workspace Tuist.xcworkspace -scheme tuist \

TUIST_FRAMEWORK_SEARCH_PATHS="<DerivedData>/Build/Products/Debug" \
  "<DerivedData>/Build/Products/Debug/tuist" \
  generate --path examples/xcode/generated_visionos_app --no-open
```

**3. Run the example's tests on a visionOS Simulator** — this is the part that
demonstrates XCUITest driving a visionOS-SDK app:

```sh
xcodebuild test \
  -workspace examples/xcode/generated_visionos_app/App.xcworkspace \
  -scheme App \
  -destination 'platform=visionOS Simulator,name=Apple Vision Pro' \
# ** TEST SUCCEEDED **
# AppUITest.testAppLaunchesAndPresentsRootView passed (4.3s)
```

**4. Run the added unit tests:**

```sh
tuist generate TuistGeneratorTests XcodeGraphTests --no-open
  -only-testing:TuistGeneratorTests/GraphLinterTests \
  -only-testing:XcodeGraphTests/ProductTests \
# ** TEST SUCCEEDED ** — 72 tests, 0 failures
```

Because a passing UI test could in principle be vacuous, I also ran a negative
control: a second test asserting an accessibility identifier that doesn't exist.
It failed while the real one passed, confirming the queries genuinely inspect the
app's accessibility hierarchy rather than short-circuiting. That control was
deliberately not committed, since a permanently-failing test would break CI.


kikeenrique:
Since the last update I've pushed a follow-up commit that makes the
`generated_visionos_app` example actually build, so the UI test target added here
is now exercised rather than inert. Full details in the updated description; the
short version is that the example hasn't compiled for visionOS since #8962
(`UIScreen` is unavailable there), which also looks like the reason that
example's acceptance test still has `BuildCommand` commented out behind a
`TODO: Fix`:

https://github.com/tuist/tuist/blob/8bb413dfeda22edd402a64b5ba4a4d43d2a2902b/cli/Tests/TuistGeneratorAcceptanceTests/GenerateAcceptanceTests.swift#L927-L931

That build failure is what this PR resolves — `xcodebuild build -scheme App`
now succeeds against visionOS 26.5.

I've left the line commented and the test `.disabled()` rather than enabling it
here: 62 of the 86 tests in that file are disabled, so that reads like a
deliberate CI-cost policy rather than neglect, and adding a visionOS build to CI
feels like a maintainer call. Happy to uncomment it in this PR if you'd prefer,
or to leave it for a follow-up.

pepicrft (approved):

PR #11519, opened 2026-06-26
feat(server): internal Atlas read-only database API

fortmarek:
## What

Adds read-only database endpoints under `/api/internal/atlas/db`, for Atlas' MCP to consume:

- `POST /api/internal/atlas/db/query` — run a single read-only SQL statement
- `GET  /api/internal/atlas/db/tables` — list app-owned tables (size + estimated rows)
- `GET  /api/internal/atlas/db/tables/:schema/:name` — describe a table's columns

## How

Backed by `Tuist.Ops.Database` — the **same read-only engine behind `/ops/db`**: a SELECT/WITH/EXPLAIN/SHOW grammar gate, a `BEGIN READ ONLY` transaction, and a clamped `statement_timeout`. Reachable only by the Atlas service account through the existing `InternalAtlasAuthPlug` (workload identity / pinned JWKS) — the same `:atlas_internal_api` pipeline as the existing `/api/internal/atlas/accounts/:handle/usage` endpoint.

Adds `Tuist.Ops.Database.to_json_map/1` for JSON-safe API responses (preserves column order + truncation metadata).

## Companion

Atlas adds the executive-gated `*_tuist_postgres*` MCP tools that call these endpoints in tuist/atlas#344.

## Tests

Controller test added, mirroring the existing `AtlasUsageController` test. I couldn't get ExUnit to finish locally (the fresh worktree's cold CLDR compile wedged), so **relying on CI** to validate.

### DIFF
--- a/infra/helm/tuist/values-managed-production.yaml
+++ b/infra/helm/tuist/values-managed-production.yaml
@@ -71,6 +71,10 @@ server:
       value: tuist-server
     - name: TUIST_ATLAS_TOKEN_MAX_TTL_SECONDS
       value: "3600"
+    # Least-privilege role the internal Atlas SQL runner drops to via SET ROLE
+    # (requires the GRANT in infra/cnpg/tuist-ops-ro-grants.sql).
+    - name: TUIST_ATLAS_DB_READONLY_ROLE
+      value: tuist_ops_ro
     - name: TUIST_ATLAS_TOKEN_JWKS
       value: >-
         {"keys":[{"use":"sig","kty":"RSA","kid":"RlVpaMmFrqqIHY46_HT42inm4mfvqBKfwLVJiPCcl7g","alg":"RS256","n":"urnlzTgGSYHPHyc-sjXI7lzdx3ChfMpqUMpwuj1vgBMywpvdjFEKZ-5xWBBfS_GIna4q--YBau4zo21xFw0Jd-Ye8l4daXozFTwB1oOv-xOoIL7YITR8dtXsXzqeGIKx1rt90jE2vgvYQDBnkIf9ohgDIApbLICYGSAko5MfdypYhiO52KOxum8nB5dwIaKhr2-DiLQtFQbl__c6Uk2dFpW0_5u4d-uO6yspp5iVo54sc7XIlFNxPBlx2uVQZe8tvTH9Ipc_sovDjMfoKH7g0fG14q6xQU5CkQD1pNS__6J3xFvFYQEp3wXu4i_Rin65EI9GvNKtbZkH-sQ5e8jzLw","e":"AQAB"}]}
--- a/infra/helm/tuist/values-managed-production.yaml
+++ b/infra/helm/tuist/values-managed-production.yaml
@@ -71,10 +71,16 @@ server:
       value: tuist-server
     - name: TUIST_ATLAS_TOKEN_MAX_TTL_SECONDS
       value: "3600"
-    # Least-privilege role the internal Atlas SQL runner drops to via SET ROLE
-    # (requires the GRANT in infra/cnpg/tuist-ops-ro-grants.sql).
-    - name: TUIST_ATLAS_DB_READONLY_ROLE
-      value: tuist_ops_ro
+    # The internal Atlas SQL runner can drop to the least-privilege tuist_ops_ro
+    # role via `SET LOCAL ROLE` by setting:
+    #   - name: TUIST_ATLAS_DB_READONLY_ROLE
+    #     value: tuist_ops_ro
+    # Left unset on purpose: that switch requires `GRANT tuist_ops_ro TO
+    # tuist_web` from infra/cnpg/tuist-ops-ro-grants.sql, which is a manual CNPG
+    # bootstrap/restore step — NOT applied by the Helm deploy. Enabling the env
+    # var before the grant is (re-)applied would make every Atlas DB query raise
+    # on `SET LOCAL ROLE`. Until then the runner stays on the web role behind the
+    # read-only transaction; flip this on in a follow-up once the grant has run.
     - name: TUIST_ATLAS_TOKEN_JWKS
       value: >-
         {"keys":[{"use":"sig","kty":"RSA","kid":"RlVpaMmFrqqIHY46_HT42inm4mfvqBKfwLVJiPCcl7g","alg":"RS256","n":"urnlzTgGSYHPHyc-sjXI7lzdx3ChfMpqUMpwuj1vgBMywpvdjFEKZ-5xWBBfS_GIna4q--YBau4zo21xFw0Jd-Ye8l4daXozFTwB1oOv-xOoIL7YITR8dtXsXzqeGIKx1rt90jE2vgvYQDBnkIf9ohgDIApbLICYGSAko5MfdypYhiO52KOxum8nB5dwIaKhr2-DiLQtFQbl__c6Uk2dFpW0_5u4d-uO6yspp5iVo54sc7XIlFNxPBlx2uVQZe8tvTH9Ipc_sovDjMfoKH7g0fG14q6xQU5CkQD1pNS__6J3xFvFYQEp3wXu4i_Rin65EI9GvNKtbZkH-sQ5e8jzLw","e":"AQAB"}]}
--- a/server/lib/tuist/ops/database.ex
+++ b/server/lib/tuist/ops/database.ex
@@ -570,29 +570,36 @@ defmodule Tuist.Ops.Database do
   defp validate_grammar(_), do: {:error, "Query must be a string"}
 
   defp run_read_only(sql, limit, role) do
-    Repo.transaction(fn ->
-      {:ok, _} = Repo.query("SET TRANSACTION READ ONLY")
-      {:ok, _} = Repo.query("SET LOCAL statement_timeout = #{@statement_timeout_ms}")
-      # When a `role` is given (the internal Atlas runner passes `tuist_ops_ro`),
-      # drop to it for the duration of the transaction so the statement executes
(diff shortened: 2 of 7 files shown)


server/lib/tuist_web/controllers/internal/atlas_database_controller.ex
```diff
@@ -0,0 +1,52 @@
...
+  Run a read-only SQL statement. Body: `{"query": "...", "limit": n}`.
+  """
+  def query(conn, %{"query" => sql} = params) when is_binary(sql) do
+    opts =
+      case params do
+        %{"limit" => limit} when is_integer(limit) and limit > 0 -> [limit: limit]
+        _ -> []
+      end
+
+    case Database.execute(sql, opts) do
```
pepicrft, line 31: Blocking: this hands the Atlas service account arbitrary database query execution through `Tuist.Repo`, which uses the web runtime database role. The existing Atlas surface only returns a curated usage field, and the repository already has a `tuist_ops_ro` role for broad inspection.

Please keep Atlas on curated read models, or run this through a separate least-privilege database role with explicit table and column allowlists before exposing a general query runner. The current guard is not enough for arbitrary input: I validated in PostgreSQL that `BEGIN READ ONLY` blocks `CREATE TABLE`, but `SELECT pg_advisory_lock(...)` inside the same read-only transaction succeeds and leaves a session advisory lock after commit until explicitly released. On a pooled connection that can poison the connection or create denial of service.

server/lib/tuist_web/controllers/internal/atlas_database_controller.ex
```diff
@@ -0,0 +1,52 @@
...
+
+  alias Tuist.Ops.Database
+
+  @doc """
+  Run a read-only SQL statement. Body: `{"query": "...", "limit": n}`.
+  """
+  def query(conn, %{"query" => sql} = params) when is_binary(sql) do
+    opts =
+      case params do
+        %{"limit" => limit} when is_integer(limit) and limit > 0 -> [limit: limit]
```
pepicrft: This accepts any positive `limit`, which bypasses the `@max_rows 200` cap in `Tuist.Ops.Database`. A caller can ask the cursor path to fetch `limit + 1` rows and force the web process to build a very large response.

Please clamp the requested value to the server-side maximum, and add a test that an oversized limit still returns at most the configured maximum.

server/lib/tuist_web/controllers/internal/atlas_database_controller.ex
```diff
@@ -0,0 +1,52 @@
...
+    conn |> put_status(:bad_request) |> json(%{error: "missing query"})
+  end
+
+  @doc "List app-owned tables with size + estimated-row stats."
+  def tables(conn, _params) do
+    json(conn, %{tables: Database.list_table_overviews()})
+  end
+
+  @doc "Describe a single table's columns. 404 when the table is not visible."
+  def describe(conn, %{"schema" => schema, "name" => name}) do
```
pepicrft, line 50: The describe endpoint does not apply the same app-owned schema filter used by `list_table_overviews/0`. I validated locally that `information_schema.tables` exposes `pg_catalog` tables, so a caller can describe catalog relations even though the table list hides them.

Please reuse a shared visibility predicate for both listing and describing tables, or return not found for schemas outside the intended application-owned set.

pepicrft (changes requested):
Found security issues that should be tightened before this merges. The inline comments include the validation behind the read-only transaction concern and the row-limit bypass.

server/lib/tuist_web/controllers/internal/atlas_database_controller.ex
fortmarek: Clamped the requested limit to the server-side `@max_rows` (200) inside `Tuist.Ops.Database.execute/2`, so an oversized value can't make the cursor fetch more. Added a controller test that `limit: 100_000` returns at most 200 rows with `truncated: true`.

server/lib/tuist_web/controllers/internal/atlas_database_controller.ex
fortmarek: `describe` now goes through a new `Database.app_table_exists?/2` that reuses the same app-owned schema filter as `list_table_overviews/0` (excludes `pg_catalog` / `information_schema` / `pg_*` / `cnpg_*` / timescaledb), so catalog relations 404. Added a test that `pg_catalog/pg_class` returns `table_not_found`.

server/lib/tuist_web/controllers/internal/atlas_database_controller.ex
fortmarek: Moved the runner to the least-privilege option: it now `SET LOCAL ROLE tuist_ops_ro` for the operator statement, so writes are blocked by the role's lack of grants, not only by `SET TRANSACTION READ ONLY`. The membership grant is added in `infra/cnpg/tuist-ops-ro-grants.sql` with `INHERIT FALSE` (tuist_web can `SET ROLE` but doesn't passively gain `pg_read_all_data`), enabled in prod via `TUIST_ATLAS_DB_READONLY_ROLE`.

For the advisory-lock leak you demonstrated, the read-only transaction now runs `pg_advisory_unlock_all()` before the connection returns to the pool, so a `SELECT pg_advisory_lock(...)` can't ride a pooled connection.

I kept this as a general read-only runner (same posture as the ClickHouse access Atlas already exposes) rather than table/column allowlists — leaving this thread open in case you'd still want allowlist scoping on top.

server/test/tuist_web/controllers/internal/atlas_database_controller_test.exs
```diff
@@ -0,0 +1,129 @@
+defmodule TuistWeb.Internal.AtlasDatabaseControllerTest do
+  use TuistTestSupport.Cases.ConnCase, async: false
```
pepicrft: Can we adjust the implementation or the test so that we can make this one `async: true`?

infra/helm/tuist/values-managed-production.yaml
```diff
@@ -71,6 +71,10 @@ server:
       value: tuist-server
     - name: TUIST_ATLAS_TOKEN_MAX_TTL_SECONDS
       value: "3600"
+    # Least-privilege role the internal Atlas SQL runner drops to via SET ROLE
+    # (requires the GRANT in infra/cnpg/tuist-ops-ro-grants.sql).
+    - name: TUIST_ATLAS_DB_READONLY_ROLE
+      value: tuist_ops_ro
```
pepicrft: Can we make the rollout order explicit here, or keep this unset until the grant has been applied? This value makes production call `SET LOCAL ROLE "tuist_ops_ro"` on every Atlas database query, but the membership that permits that switch is only added in `infra/cnpg/tuist-ops-ro-grants.sql`. That file is a manual CloudNativePG bootstrap or restore step and will not be applied by the Helm deploy on the existing production cluster. If the deploy lands before that grant is rerun, the endpoint will raise on the first query instead of returning the read model.

server/lib/tuist/ops/database.ex
```diff
@@ -534,19 +569,45 @@ defmodule Tuist.Ops.Database do
...
-      else
-        build_result(Repo.query(sql), limit)
-      end
+      # `SET TRANSACTION READ ONLY` blocks writes but still permits side-effecting
+      # functions like `SELECT pg_advisory_lock(...)`, whose session-level lock
+      # survives commit and would ride a pooled connection into the next caller
+      # (or collide with Oban's advisory locks). Release any such locks before
+      # the connection returns to the pool.
+      {:ok, _} = Repo.query("SELECT pg_advisory_unlock_all()")
+      result
```
pepicrft, line 603: Can we make the advisory-lock cleanup run even when the query errors? As written, `pg_advisory_unlock_all()` only runs after `fetch_via_cursor/2` or `build_result/2` returns normally. A query such as a materialized common table expression that takes `pg_advisory_lock(...)` and then fails will call `Repo.rollback/1` before line 595, leaving the session-level lock on the pooled connection. The transaction rollback does not release session-level advisory locks, so the next checkout can inherit the lock or collide with Oban. This needs cleanup outside the transaction on the same checked-out connection, or the runner should reject these advisory-lock functions.

pepicrft (approved):
Approving after the addressed changes. I left two follow-up comments on rollout and cleanup edge cases.

server/test/tuist_web/controllers/internal/atlas_database_controller_test.exs
fortmarek: Switched the test to `async: true` — it stubs via Mimic (`set_mimic_from_context` runs in private mode under async) and only issues read-only queries with no shared fixtures, so it's isolation-safe. Consistent with the other `async: true` + Mimic tests in the suite.

infra/helm/tuist/values-managed-production.yaml
fortmarek: Good catch — left `TUIST_ATLAS_DB_READONLY_ROLE` unset in production and documented the two-step rollout inline: the runner stays on the web role (behind the read-only transaction) until `GRANT tuist_ops_ro TO tuist_web` from the CNPG grants file has been re-applied, then the env var is flipped on in a follow-up. So a deploy can't land `SET LOCAL ROLE` before the membership exists.

server/lib/tuist/ops/database.ex
fortmarek: Fixed — `run_read_only/3` now wraps the transaction in `Repo.checkout/2` and runs `pg_advisory_unlock_all()` after it on the same pinned connection, so the cleanup happens on both the commit and the rollback path (including a lock-taking CTE that then fails).

pepicrft (approved):

PR #11522, opened 2026-06-26
fix(cli): fail SwifterPM --force-resolved-versions on out-of-date Package.resolved

fortmarek:

## What changed

SwifterPM's `--force-resolved-versions` (a.k.a. `--disable-automatic-resolution` / `--only-use-versions-from-resolved-file`) now fails when `Package.resolved` is out of date with `Package.swift`, instead of silently restoring the stale pins.

Two layers, both calling a shared satisfiability check:

- **`validateResolvedFileSatisfiesManifest`** — fast, fail-fast check of the root manifest's **direct** dependencies in `resolveOrLoad` (runs before the restore, needs no checkouts).
- **`validateResolvedGraphSatisfiesManifests`** — post-restore walk of the **whole pinned graph**, recursing into each pinned package's manifest (dumped from its materialized checkout / registry download) to validate **transitive** constraints. The error names the requiring package.

Wired into both resolution paths: `SwifterPM.runResolution` (used by tuist) and `CLIRunner.runResolutionCommand` (the standalone `swifterpm` binary).

## Why

A customer running a SwifterPM-enabled `build_and_ut` shadow CI job (alongside the default SwiftPM one) hit a silent failure: a direct dependency was declared `exact: "1.16.0-ALPHA"` in `Package.swift` while the committed `Package.resolved` still pinned `1.15.1-ALPHA`. With `--force-resolved-versions`, SwifterPM restored the stale `1.15.1-ALPHA` and exited 0, so the build compiled against the wrong version.

### Root cause

The `readOnly` branch of `PackageResolver.resolveOrLoad` read `Package.resolved` verbatim and returned it with **no validation against the manifest's requirements** — it failed open. Pure SwiftPM fails closed in the same situation:

```
error: an out-of-date resolved file was detected at .../Package.resolved,
which is not allowed when automatic dependency resolution is disabled
```

So this was a genuine SwiftPM-parity gap, and it also swallowed exactly the drift signal the flag is meant to surface.

### Why this solution

I verified against the SwiftPM source (`swiftlang/swift-package-manager`, `ResolverPrecomputationProvider`): SwiftPM runs the resolver with each package's *only* available version fixed to its pin (`LocalPackageContainer.versionsAscending` returns `[pinnedVersion]`), so the out-of-date error fires when a pin **cannot satisfy a constraint** — and notably **not** merely because a newer version exists within a still-satisfied range. This PR reproduces that satisfiability semantics rather than, say, an `originHash` comparison (which can't be shared across tools) or a `git diff` heuristic.

### Deliberate, documented divergences (to stay false-positive-free)

A wrong "out-of-date" error would break a working lockfile, so the check is conservative:

- Version constraints are strict (`VersionRange.contains`); branch/revision are checked leniently.
- Pins that can't be matched by identity (e.g. an SCM→registry identity remap) or carry an unparseable version are skipped.
- A pin that still satisfies a loose range is **not** flagged just because a newer version exists — matching SwiftPM. That class of drift still wants a `swift package resolve` + `git diff --exit-code Package.resolved` guard.
- Missing/superfluous packages are not flagged (SwiftPM does), again to avoid identity-remap false positives.

## User impact

CI using `--force-resolved-versions` now fails fast with an actionable message when `Package.swift` and `Package.resolved` drift, instead of silently building stale versions:

```
an out-of-date Package.resolved was detected at .../Package.resolved, which is not
allowed when --force-resolved-versions is set: 'swift-collections' is pinned to 1.0.0
but 'swift-async-algorithms' requires 1.0.4..<2.0.0. Update the resolved file
(e.g. run `tuist install` or `swift package resolve`) and commit it.
```

In-sync lockfiles are unaffected.

### How to test locally

End-to-end against a real graph (`swift-async-algorithms` → `swift-collections`):

1. Build the CLI: `swift build --replace-scm-with-registry --product swifterpm` in `swifterpm/`.
2. In a package whose `Package.swift` declares a versioned dependency, seed a lockfile: `swifterpm resolve --write --package-path <pkg>`.
3. Bump the dependency's `exact:`/range in `Package.swift` past the pin (or hand-edit `Package.resolved` to pin a transitive dep out of range), then run `swifterpm resolve --force-resolved-versions --package-path <pkg>` → exits 1 with the out-of-date error.
4. Restore the lockfile to a satisfying state → exits 0.

Automated: `swift test --replace-scm-with-registry --filter ResolveTests` (adds 4 regression tests: direct fail / direct in-sync / transitive fail / transitive in-sync).

### DIFF
--- a/swifterpm/Sources/swifterpm/Resolve.swift
+++ b/swifterpm/Sources/swifterpm/Resolve.swift
@@ -213,11 +213,10 @@ enum PackageResolver {
                     "Package.resolved is required when forcing resolved versions, but no file exists at \(resolvedFileURL.path)"
                 )
             }
-            let resolved = try await ResolvedFile.read(packageDir: packageDir)
-            try await validateResolvedFileSatisfiesManifest(
-                resolved, packageDir: packageDir, disableSandbox: disableSandbox
-            )
-            return resolved
+            // The pins are trusted as-is here; whether they're still in sync with
+            // the manifest is verified by `assertResolvedFileUpToDate`, which the
+            // resolution flow runs once the checkouts are materialized.
+            return try await ResolvedFile.read(packageDir: packageDir)
         }
         // `--skip-update` is an explicit "trust the on-disk pins" signal:
         // read the file as-is even when it predates the `originHash` field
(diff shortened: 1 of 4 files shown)


swifterpm/Sources/swifterpm/Resolve.swift
```diff
@@ -257,6 +261,227 @@ enum PackageResolver {
...
+    /// drift, so it's skipped instead of risking a false positive that would
+    /// break a working lockfile.
+    private static func validateConstraints(
+        _ dependencies: [ManifestDependency],
+        pinsByIdentity: [String: ResolvedPin],
+        packageDir: URL,
+        requiredBy: String
+    ) throws {
+        for dependency in dependencies {
+            guard let pin = pinsByIdentity[dependency.identity.lowercased()] else { continue }
```
pepicrft: Could we avoid growing this into a partial resolver check? Swift Package Manager already catches missing pins when automatic resolution is disabled. I verified a local package where `Package.swift` adds a new dependency `b` while `Package.resolved` only contains `a`: `swift package --force-resolved-versions resolve` fails, but this branch returns success because this guard skips unmatched dependencies. I think SwifterPM should either delegate this out-of-date validation to Swift Package Manager or explicitly fail unmatched manifest dependencies only when the identity can be mapped with the same data Swift Package Manager uses.

swifterpm/Sources/swifterpm/Resolve.swift
```diff
@@ -257,6 +261,227 @@ enum PackageResolver {
...
+            // A version requirement against a branch/revision pin is genuine
+            // drift (the manifest no longer asks for a versioned dependency).
+            guard let versionString = pin.state.version else { return false }
+            // An unparseable pin version is not proof of drift; don't fail on it.
+            guard let version = try? SemVer(versionString) else { return true }
+            return range.contains(version)
+        case let .revision(revision):
+            guard let pinned = pin.state.revision, !pinned.isEmpty, !revision.isEmpty else { return true }
+            return pinned.hasPrefix(revision) || revision.hasPrefix(pinned)
+        case let .branch(branch):
```
pepicrft: Same concern here: this is duplicating Swift Package Manager precomputation and the fallback currently accepts a version pin for a branch requirement. I verified by changing a dependency from `exact: "1.0.0"` to `branch: "main"`: Swift Package Manager fails with an out-of-date resolved file, while this branch exits successfully because `pin.state.branch` is nil. I would keep this path from trying to interpret branch and revision drift unless it is delegating to Swift Package Manager's resolver state.

pepicrft (approved):
Approving the direction. The inline comments are about keeping this from becoming a partial copy of Swift Package Manager's out-of-date resolved-file validation.

swifterpm/Sources/swifterpm/Resolve.swift
fortmarek: Good point. Instead of extending the hand-rolled check I replaced it with delegation to Swift Package Manager. After the restore, the force-resolved path runs `swift package resolve --force-resolved-versions`, which reuses the workspace state we just materialized so it stays a fast precomputation rather than a fresh resolve. That picks up added or otherwise unmatched dependencies natively. Verified that adding a dependency that is not present in `Package.resolved` now fails.

swifterpm/Sources/swifterpm/Resolve.swift
fortmarek: Agreed. I dropped the hand-rolled satisfiability altogether, including the lenient branch and revision handling, and now delegate to Swift Package Manager via `swift package resolve --force-resolved-versions`. The `exact:` to `branch:` change now fails the way Swift Package Manager intends, and there is no duplicated precomputation left to maintain.

PR #11532, opened 2026-06-26
fix(cli): make Copy Module Map script phase idempotent

JanC:
The `Copy Module Map` post-build script phase emitted by `ModuleMapMapper` used a bare `cp`, which isn't idempotent. This PR switches it to `install -m 0644`.

**Why a bare `cp` breaks:** SwiftPM checks out package sources read-only (mode `444`), including `module.modulemap`. `cp` (no `-f`) both refuses to overwrite a read-only destination and copies the `444` source mode onto the destination. That's fine for a single build: the destination doesn't pre-exist and the script runs once. It breaks under `tuist cache warm`, which drives several `*-Cache-*` schemes into one shared `SYMROOT`/derivedData with divergent build settings (e.g. `Binaries-Cache` gets `DEBUG_INFORMATION_FORMAT=dwarf STRIP_INSTALLED_PRODUCT=YES SWIFT_SERIALIZE_DEBUGGING_OPTIONS=NO COMPILER_INDEX_STORE_ENABLE=NO`, `Macros-Cache` does not). The mismatch leaves a C target such as `_SwiftSyntaxCShims` dirty for the next scheme, which re-runs the phase against a now read-only destination:

```
cp: <SYMROOT>/Debug/_SwiftSyntaxCShims.framework/Modules/module.modulemap: Permission denied
```

`set -eu` turns that into a hard abort and `xcodebuild` exits 65:

```
PhaseScriptExecution Copy\ Module\ Map .../_SwiftSyntaxCShims.build/Script-<hash>.sh
  (in target '_SwiftSyntaxCShims' from project 'swift-syntax')
** BUILD FAILED **
```

Declared `inputPaths`/`outputPaths` + `basedOnDependencyAnalysis` normally let Xcode skip the second run, but the shared-`SYMROOT` multi-scheme flow re-dirties the target and re-runs the phase anyway. The phase is new (introduced in #11135, shipped in `4.201.0-canary.18`) and was never idempotent.

**Why `install -m 0644`:** `cp -f` and `rm -f dst && cp` fix the overwrite but leave the destination at `444`, so every run depends on the unlink step. `install -m 0644` recreates the destination via a temp file + rename (overwriting regardless of an existing read-only target) and normalizes the mode, so the `444` source mode no longer propagates. One idempotent command with the right mode for a module map.

### How to test locally

Isolated shell repro (source at mode `444`, copy run twice into a shared destination):

```sh
printf 'module X {}\n' > src.modulemap; chmod 444 src.modulemap
cp src.modulemap dst && cp src.modulemap dst                              # 2nd cp -> "Permission denied", exit 1
install -m 0644 src.modulemap dst2 && install -m 0644 src.modulemap dst2  # both runs pass; dst2 is 0644
```

End to end: run `tuist cache warm` on a project that depends on a SwiftPM package with a custom/clang module map (e.g. swift-syntax's `_SwiftSyntaxCShims`) for a universal macOS build (`ONLY_ACTIVE_ARCH=NO`). Before this change the warm fails with the `cp ... Permission denied` / exit 65 above; after it, the `Copy Module Map` phase succeeds on re-run.

Unit coverage: `ModuleMapMapperTests` asserts the emitted script; the assertion was updated to expect the `install -m 0644` form.


KaiOelfke:
I'm also affected. I guess this should be fixed before 4.201.0 can be called stable.

JanC:
> I guess this should be fixed before 4.201.0 can be called stable.


that would be great if it could go the next RC of 4.201.0

JanC:
closing because this issue was fixed in https://github.com/tuist/tuist/pull/11646

PR #11535, opened 2026-06-27
ci(kura): build all release binaries with Bazel (cross-compiling macOS x86_64)

esnunes:
## What changed

`release-kura-binaries` in `server-production-deployment.yml` now builds **every** kura release binary with **Bazel**, reusing the Tuist remote cache like the rest of Kura CI. The cargo path is gone entirely — including `x86_64-apple-darwin`, which now **cross-compiles from the arm64 `tuist-macos` runner**.

| Target | Runner | Build |
|---|---|---|
| `x86_64-unknown-linux-gnu` | `tuist-linux` | Bazel (native) |
| `aarch64-unknown-linux-gnu` | `ubuntu-24.04-arm` | Bazel (native) |
| `aarch64-apple-darwin` | `tuist-macos` (arm64) | Bazel (native) |
| `x86_64-apple-darwin` | `tuist-macos` (arm64) | Bazel (**cross**, `--config=macos-x86_64-cross`) |

This is the follow-up #11533 deferred: *"The `release-kura-docker` job (still cargo `--release`, multi-arch) is migrated in a separate PR."*

## Why

`release-kura-binaries` was the last cargo step of the kura release path. It ran `cargo build --release` with no cache, **cold-compiling the `-sys` crates (rocksdb, aws-lc) from source on every release** with zero reuse of the Tuist remote cache that `bazel-compile`/`test`/`clippy` and the e2e image (#11530) already share — and the `-sys` C/C++ is exactly the expensive part that cache exists to amortize.

## How

- **Bazel builds the binary on every leg** via `mise run --skip-tools compile -- --binary-only`, with `id-token: write` + best-effort `tuist bazel setup` (continue-on-error; `.bazelrc` try-imports `.bazelrc.tuist`, so a cache miss falls back to a cold local build instead of blocking the release).
- **macOS x86_64 cross-compiles, no `MODULE.bazel` change.** `apple_support` (already transitive via rules_rs) supplies a cross `darwin_x86_64` cc toolchain for the `-sys` crates, and rules_rs already carries the `x86_64-apple-darwin` rust toolchain (it's in `platform_triples`). A `//bazel/platforms:macos_x86_64` platform + a `build:macos-x86_64-cross` `.bazelrc` config (`--platforms=…`) are all that's needed.
- **The release builds `//:kura`, not `//...`.** The wildcard pulls in the test target, which can't cross-compile (no test toolchain for a non-host target platform). To keep the shared `mise` task as the entrypoint, `compile.sh` gains a `--binary-only` usage flag (`//:kura` instead of `//...`) while still forwarding arbitrary bazel flags — the cross `--config` and `kura.yml`'s existing `--define` — through `"$@"`. (The `[bazel_args] var=#true` arg only *declares* that passthrough; its captured value is read from `"$@"` because usage's own capture var duplicates `--key=value` tokens and `double_dash` isn't supported in this mise version.)
- **Docker assembly unchanged.** `release-kura-docker` consumes the same `kura-binary-*-artifacts` (same tar layout) via `Dockerfile.release` — untouched. The job id stayed `release-kura-binaries`, so all downstream `needs`/`result` gates are intact.

### Why cross-compile rather than drop macOS x86_64

This supersedes the open question on the earlier revision (which left x86_64 on cargo). The latest macOS — Tahoe (26) — still runs on Intel, so the target has a live audience this cycle; cross-compiling keeps it **and** removes the cargo special-case, leaving the job uniform. apple_support makes it essentially free, so there's no longer a reason to special-case or drop it.

### `--compilation_mode=opt`

Already on `main` via #11530 / #11533 (the binary must match the optimized binary that ships, and one mode keeps every Bazel job on a single warm cache keyspace), so this PR doesn't re-add it. The `kura/.bazelrc` macOS comment is updated to note both release binaries (arm64 native + x86_64 cross) now build there.

## Validation

- Verified locally on an arm64 Mac through the real task: `mise run --skip-tools compile -- --binary-only` → `Mach-O … arm64`; `… -- --binary-only --config=macos-x86_64-cross` → `Mach-O … x86_64` (full rocksdb/aws-lc `-sys` stack compiles clean under the cross toolchain).
- `actionlint` clean apart from the expected `tuist-linux`/`tuist-macos` self-hosted-label notes shared by every existing Bazel job in `kura.yml`.

### DIFF
--- a/.github/workflows/server-production-deployment.yml
+++ b/.github/workflows/server-production-deployment.yml
@@ -373,24 +373,39 @@ jobs:
     if: needs.check-releases.outputs.kura-should-release == 'true'
     runs-on: ${{ matrix.runner }}
     timeout-minutes: 60
+    # id-token: OIDC for `tuist bazel setup`. Job permissions replace (not merge) the workflow's,
+    # so contents:read must be restated for checkout.
+    permissions:
+      contents: read
+      id-token: write
     strategy:
       fail-fast: false
       matrix:
+        # macOS x86_64 is the only non-native target (cross from the arm64 runner), and the Bazel
+        # setup has no cross C/C++ toolchain for the -sys crates (kura/MODULE.bazel), so it uses cargo.
         include:
-          - runner: ubuntu-latest
+          - runner: tuist-linux
             target: x86_64-unknown-linux-gnu
+            build: bazel
+            mise-tools: "rust bazel tuist"
           - runner: ubuntu-24.04-arm
             target: aarch64-unknown-linux-gnu
-          - runner: tuist-macos
-            target: x86_64-apple-darwin
+            build: bazel
+            mise-tools: "rust bazel tuist"
           - runner: tuist-macos
             target: aarch64-apple-darwin
+            build: bazel
+            mise-tools: "rust bazel tuist"
+          - runner: tuist-macos
+            target: x86_64-apple-darwin
+            build: cargo
+            mise-tools: "rust"
     steps:
       - uses: actions/checkout@v4
       - uses: jdx/mise-action@v4.0.1
         with:
           version: ${{ env.MISE_VERSION }}
-          install_args: "rust"
+          install_args: ${{ matrix.mise-tools }}
           cache: "false"
           working_directory: kura
           github_token: ${{ env.MISE_GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


esnunes:
### 🧭 Open question: should we keep building `x86_64-apple-darwin`?

This PR leaves macOS x86_64 on cargo **only** because Bazel here can't cross-compile it from the arm64 runner. It's the single reason a cargo branch still exists in `release-kura-binaries`. Worth deciding whether that target still earns its keep:

**Case for dropping it**
- No new Intel Mac has shipped since the **27″ iMac, Aug 2020**; Intel Macs went end-of-sale by mid-2023.
- **macOS 26 "Tahoe" (Sept 2025) is the last macOS to support Intel** — macOS 27 (~fall 2026) is slated Apple-Silicon-only. The x86_64 audience starts its terminal decline next cycle.
- Dropping it makes **every** remaining target native-per-runner, so the whole job collapses to a uniform Bazel build with **no cargo branch** — simpler and fully on the shared remote cache.

**Case for keeping it (for now)**
- The latest macOS *does* still run on x86_64, so there's a live, supported audience **this** cycle (2019–2020 Intel models on Tahoe).
- It's cheap to keep — one cargo leg, no extra toolchain work.

pepicrft (approved):

PR #11536, opened 2026-06-27
ci(server): build the staging Kura runtime image from a Bazel-compiled binary

esnunes:
## What changed

The `Server Deployment` workflow (`server-deployment.yml`) built the ad-hoc **staging Kura runtime** image (`ghcr.io/tuist/kura:sha-<sha>`) — the Rust mesh node, not the controller — via an in-Docker `cargo build --release`. This PR moves it onto the same **Bazel + Tuist remote-cache** path the rest of Kura CI already uses, baking the prebuilt binary into `Dockerfile.release`.

## Why

`server-deployment.yml` builds a per-commit `sha-<sha>` Kura runtime image so a feature branch can hit **staging** before it's merged and `release-kura-docker` cuts a `kura@` release. That build was the last cargo-based step of Kura CI:

- It used `kura/Dockerfile`, which runs `cargo build --release` **inside** the Docker build, **cold-compiling the `-sys` crates (rocksdb, aws-lc) from source on every staging deploy** with zero reuse of the Tuist remote cache that `bazel-compile`/`test`/`clippy` already share (only a per-branch registry layer cache, invalidated whenever `kura/` changes — i.e. exactly the deploys you care about).

The Kura **controller** image and the heavy **server** image build are unaffected.

## How

Mirrors [#11533](https://github.com/tuist/tuist/pull/11533), which did the identical migration for the standalone `kura-runtime-image.yml` dispatch workflow:

- **Extract a dedicated `build-kura-runtime` job.** It compiles the binary with Bazel (`mise run --skip-tools compile`; opt via `kura/.bazelrc`), reusing the Tuist remote cache, then bakes it into `Dockerfile.release` — the same prebuilt-binary assembly the release pipeline uses (no in-image compilation; `ubuntu:24.04` base matches the `tuist-linux` glibc).
- **Runs on `tuist-linux` with `id-token: write`** for the persistent local Bazel cache + the OIDC token `tuist bazel setup` consumes (best-effort: `tuist auth login && tuist bazel setup` is `continue-on-error`, and `.bazelrc` try-imports `.bazelrc.tuist`, so a cache miss builds cold locally instead of blocking the deploy). The `build` job keeps building the server + controller images on its own runner.
- **Same gating, now explicit.** The old step ran only on `environment == 'staging' && kura_runtime_image_tag == ''` inside a `build` job already guarded by `image_tag == ''`. The new standalone job spells out all three (`staging` + `image_tag == ''` + `kura_runtime_image_tag == ''`) and pushes the identical `sha-<SHA>` tag the deploy pins.
- **Wired into `deploy`** via `needs` + a `success || skipped` clause in its `if` (the job is skipped on canary/production and on staging re-rolls of a prebuilt image, where the deploy resolves the released `kura@` tag instead).
- **`--compilation_mode=opt` in `kura/.bazelrc`** so the staged binary is the optimized one rather than fastbuild, and every Bazel job stays on one warm cache keyspace. Side effect: `bazel test` then runs with `debug_assertions` off (like `cargo test --release`) — verified safe for kura in #11530.

## Coordination

The `kura/.bazelrc --compilation_mode=opt` line is the **same line #11533 and #11530 add**. It's included here so this change is correct standalone (otherwise `mise run compile` would stage an unoptimized fastbuild binary). Whichever of these PRs merges last hits a trivial one-line add/add conflict — resolve by keeping a single copy.

## User / developer impact

Staging deploys that touch `kura/` no longer pay a full cold Rust + librocksdb-sys compile inside Docker; they reuse the shared Bazel remote cache. No change to canary/production (they pin the released `kura@` tag and skip this job) or to the server/controller images.

## Validation

- `actionlint` clean apart from the expected `tuist-linux` self-hosted-label notes — shared by every existing Bazel job in `kura.yml` and the 4 pre-existing `tuist-linux` jobs in this workflow.
- Could not run the Bazel-on-Linux build locally (local Bazel is macOS/arm64; the image needs the Linux binary CI produces) — but the binary-build recipe and the `Dockerfile.release` assembly are both the proven paths from #11533 and the release pipeline.

### DIFF
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -120,6 +120,7 @@ permissions:
   packages: write
 
 env:
+  MISE_VERSION: "2026.5.15"
   REGISTRY: ghcr.io
   IMAGE_NAME: tuist/tuist
   KURA_CONTROLLER_IMAGE_NAME: tuist/kura-controller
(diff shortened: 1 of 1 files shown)


.github/workflows/server-deployment.yml
```diff
@@ -223,40 +224,97 @@ jobs:
...
+        continue-on-error: true
+      - name: Warn when the Tuist remote cache is unavailable
+        if: ${{ hashFiles('kura/.bazelrc.tuist') == '' }}
+        run: echo "::warning::Tuist remote cache unavailable (.bazelrc.tuist not written); building cold against the local cache."
+      - name: Build the Kura binary with Bazel
+        run: |
+          mise run --skip-tools compile
+          mkdir -p bin/amd64
+          cp bazel-bin/kura bin/amd64/kura
+          chmod +x bin/amd64/kura
```
pepicrft, line 295: Why do we need to copy it to another directory?

pepicrft (approved):

.github/workflows/server-deployment.yml
esnunes: we don't, the dockerfile.release expects it in a folder with another path pattern. I will adjust it, good catch!

fortmarek (approved):

.github/workflows/server-deployment.yml
esnunes: There is a problem with it, bazel-bin is a symbolic link and it is not part of docker context when building the image. I will leave it like this and address this later, once all workflows are migrated

PR #11547, opened 2026-06-28
fix(cli): isolate host-less unit test derived data in default test runs

JanC:
### Context

> [!NOTE]
> This is AI  PR - review with care :) 
> I did review it myself 

Follow-up to #11225 and the community report: https://community.tuist.dev/t/tuist-test-runs-host-less-module-unit-tests-in-the-app-launch-context-xctest-crashes-at-bootstrap/

#11225 made bare `tuist test` split a mixed workspace scheme so host-less unit tests run through their own generated project scheme instead of the app-launch aggregate. The split fixed the *scheme*, but not the *build-products directory*.

I tested `4.201.0-rc.3` against a minimal repro (app links point-free `swift-sharing` as a dynamic framework, plus a Foundation-only framework module with a host-less `.unitTests` bundle). `AppTests` runs via the App scheme, then the host-less bundle runs via the `Feature` scheme, and it still crashes with the same `exit 65`:

```
xctest encountered an error (Early unexpected exit, operation never finished bootstrapping ...
The test runner crashed while preparing to run tests: xctest at <external symbol>)
** TEST FAILED **
```

### Root cause

For bare `tuist test`, `derivedDataPath` is `nil`, so every split scheme resolves to the same per-workspace derived data. The hosted scheme runs first and populates `Build/Products/Debug-iphonesimulator` with the app's frameworks (`Sharing.framework` and its closure). The host-less `Feature` scheme then runs into that same directory, the host-less process `dlopen`s those frameworks at test discovery, and hits the Apple XCTest class-realization segfault.

Running the exact same `Feature` scheme into a clean, isolated derived data passes, which isolates the products directory as the variable:

```bash
xcodebuild test -scheme Feature -workspace Sample.xcworkspace \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro,OS=26.4' \
  -derivedDataPath /tmp/clean-dd
# ** TEST SUCCEEDED **
```

### Change

When `tuist test` splits a mixed workspace scheme, run each split host-less scheme against its own derived data so its products directory does not contain the app's frameworks. Only host-less schemes are isolated; the hosted scheme keeps the shared derived data and its incremental builds. An explicit `--derived-data-path` is honored by nesting under it.

### Trade-offs

- The host-less module subtree builds fresh into the isolated directory. That is the module subtree, not the whole app, and correctness beats reuse here.
- Binary cache / selective-testing hashing is unaffected; only build reuse for the isolated scheme changes.

### DIFF
--- a/cli/Sources/TuistKit/Services/TestService.swift
+++ b/cli/Sources/TuistKit/Services/TestService.swift
@@ -1194,6 +1194,11 @@ public struct TestService { // swiftlint:disable:this type_body_length
             uploadCacheStorage = cacheStorage
         }
 
+        let passthroughDerivedDataPath = try? await xcodeBuildAgumentParser
+            .parse(passthroughXcodeBuildArguments)
+            .derivedDataPath
+        let hostlessDerivedDataBasePath = derivedDataPath ?? passthroughDerivedDataPath
+
         let passthroughSkippedTargetNames = passthroughSkippedTestTargetNames(passthroughXcodeBuildArguments)
         let testSchemeRuns = schemes.compactMap { testScheme -> (scheme: Scheme, testTargets: [TestIdentifier])? in
             let testSchemeTargetNames = Set(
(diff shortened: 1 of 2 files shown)


JanC:
hi @fortmarek 
I rebased this PR to resolve the conflcits. Any chance to get this reviewed and merged?

PR #11553, opened 2026-06-29
fix(cli): restore SwifterPM package submodules

fortmarek:
## What changed

- Initialize Git submodules when SwifterPM falls back to a shallow Git checkout.
- Treat cached source checkouts with missing submodule contents as stale instead of accepting them just because `Package.swift` exists.
- Add regression coverage for packages whose target sources live inside a required submodule, modeled after `libwebp-Xcode`.

## Why

A reported SwifterPM install generated a checkout for `SDWebImage/libwebp-Xcode` `1.3.1`, but Tuist graph loading later failed with invalid source globs for `libwebp/src/**` and `libwebp/sharpyuv/**`.

The upstream package declares those source folders in `Package.swift`, while the actual folders live inside the `libwebp` Git submodule. SwifterPM correctly rejected GitHub source archives for repositories with `.gitmodules` and fell back to Git, but the fallback checkout did not initialize submodules. The cache then considered that incomplete checkout valid because the manifest file existed, so retries after `tuist clean` could keep reusing the bad cached source.

## Impact

SwifterPM restores source-control packages with required submodules fully enough for Tuist to map package target sources. Existing incomplete cache entries are refreshed automatically the next time the package is restored.

## Validation

All validation was run from the clean PR worktree based on `origin/main`:

1. `git diff --check`
   - Confirmed the final patch has no whitespace errors.
2. `swift test --replace-scm-with-registry --filter RestoreTests`
   - Built SwifterPM from the PR branch.
   - Ran the focused `RestoreTests` suite.
   - Passed 13 tests, including the new coverage for initializing required Git submodules and refreshing stale cached sources with missing submodule contents.

### DIFF
--- a/swifterpm/Sources/swifterpm/Restore.swift
+++ b/swifterpm/Sources/swifterpm/Restore.swift
@@ -626,14 +626,13 @@ enum WorkspaceRestorer {
 
     static func ensureSource(cache: Cache, pin: ResolvedPin) async throws -> URL {
         let destination = try cache.sourcePath(pin: pin)
-        let manifest = destination.appendingPathComponent("Package.swift")
-        if try await fileSystem.exists(manifest.absolutePath) {
+        if try await cachedSourceIsUsable(destination) {
             return destination
         }
 
         let lock = try await cache.lock(namespace: "sources", key: destination.path)
         _ = lock
-        if try await fileSystem.exists(manifest.absolutePath) {
+        if try await cachedSourceIsUsable(destination) {
             return destination
         }
         if try await fileSystem.exists(destination.absolutePath) {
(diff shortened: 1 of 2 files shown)


pepicrft (approved):
I did a deeper pass over the restore changes and I am approving this. The main behavior is covered by the new tests and the package unit target passes. I found two follow-up items that are worth tracking, but I do not think they should block this fix:

1. `swifterpm/Sources/swifterpm/Restore.swift:678` only validates submodules declared by the top-level `.gitmodules` when deciding whether a cached source tree is usable. A cache that has a first-level submodule but is missing a nested submodule is accepted as usable. I validated this by seeding a non-recursive cache: restore exited successfully, but `Sub/NestedLib/src/nested.c` was missing. Removing that cached source and restoring again populated the nested file.
2. `swifterpm/Sources/swifterpm/Restore.swift:883` swallows `.gitmodules` parsing failures and treats them as no submodules. Swift Package Manager fails for the same malformed repository with `fatal: bad config line`, while SwifterPM restore exited successfully. This only affects malformed repositories, so I see it as a lower-risk compatibility gap.

Validation:
- `git diff --check main...HEAD`
- `swift test --replace-scm-with-registry --filter RestoreTests`
- `mise run test:unit`

PR #11554, opened 2026-06-29
fix(infra): release the xcresult-processor image on baked-file changes, not commit scope

fortmarek:
## What changed

The `xcresult-processor-image` release gate now triggers on **file changes** to its baked paths, not on conventional-commit **scope**. Implemented via a new per-component `release_on_path_change` flag in `mise/tasks/release/components.json` (set only for `xcresult-processor-image`) + corresponding logic in `mise/tasks/release/check.sh`. All other components are unchanged.

## Why — root cause of a prod incident

The xcresult-processor runs on a macOS Tart VM whose image **bakes a server release** (`server/lib/tuist/**`, `server/native/xcresult_nif/**`, `server/mix.exs`, `server/mix.lock`). The release gate ran `git cliff --bumped-version` over those include-paths, but cliff's `commit_parsers` `skip` `chore`/`ci`/other-scoped commits — so a `chore`-scoped change to baked code yields **no version bump and no rebuild**.

That is how **#11460** (`chore(server): source secrets from 1Password and retire Namespace`) drifted the image:
- It removed the `Namespace` code **and** dropped `accounts.namespace_tenant_id` (migration).
- The server rolled forward to the new schema; the xcresult-processor kept running the **old baked code** (`0.31.5`, built before #11460) and queried the dropped column.
- Result: `Tuist.Tests.Workers.ProcessXcresultWorker` jobs erroring/discarding (`Postgrex 42703 undefined_column`, Sentry TUIST-11S) — broken xcresult test-result processing in prod.

For an image that bakes another component's code, **scope is the wrong signal** — any change to the baked files can drift it from the live schema. The right signal is "did the baked files change."

## How

When `release_on_path_change: true`, `check.sh` decides `should-release` from `git diff --name-only <last-tag>..HEAD` over the component's `include_paths` (scope-agnostic). Version comes from cliff's bump when it produces one, else a forced patch bump (so a chore/refactor-only change to baked code still ships). Default behavior (scope-gated) is unchanged for every other component.

## Tradeoff (intentional)

The macOS image will now rebuild whenever `server/lib/tuist/**` (etc.) changes — **more often than the old scope-gated cadence**. That's the point: it keeps the xcresult-processor in lockstep with the server release it bakes rather than silently drifting. The build is async on the dedicated `vm-image-builder` Mac mini.

## Impact

Merging this makes the next production cascade's `release-xcresult-processor-image` job fire (baked paths changed since `0.31.5` via #11460), cutting `0.31.6` from current `main` (Namespace code gone), rewriting the chart tag, and deploying it — which permanently resolves TUIST-11S. (Interim mitigation already in place: `accounts.namespace_tenant_id` was manually re-added so the old image keeps working; once `0.31.6` is live the column can be dropped for good.)

## Validation

- `mise run release:check xcresult-processor-image` → `release? true`, next `0.31.6` (cliff found no feat/fix in range, patch-bump fallback applied)
- `cli` / `server` / `kura` (scope-gated) → unchanged, no crash
- `bash -n` clean

### DIFF
--- a/mise/tasks/release/check.sh
+++ b/mise/tasks/release/check.sh
@@ -22,12 +22,31 @@ emit_output() {
   fi
 }
 
+# Increment the patch of an X.Y.Z version. Used by path-triggered components
+# when the baked files changed but cliff found no feat/fix to bump on.
+patch_bump() {
+  local v=$1 a b c
+  IFS='.' read -r a b c <<< "$v"
+  printf '%s.%s.%s' "${a:-0}" "${b:-0}" "$(( ${c:-0} + 1 ))"
+}
+
+# Strip a trailing glob (/**/*, /**, /*) off an include-path so it can be used
+# as a plain git pathspec for diffing.
+pathspec_of() {
+  local p=$1
+  p="${p%/\*\*/\*}"
+  p="${p%/\*\*}"
+  p="${p%/\*}"
+  printf '%s' "$p"
+}
+
 check_component() {
-  local component=$1 tag_prefix=$2 tag_regex=$3 cliff_config=$4 initial=$5
-  shift 5
+  local component=$1 tag_prefix=$2 tag_regex=$3 cliff_config=$4 initial=$5 path_triggered=$6
+  shift 6
+  local include_paths_local=("$@")
   local include_args=()
   local p
-  for p in "$@"; do include_args+=(--include-path "$p"); done
+  for p in "${include_paths_local[@]}"; do include_args+=(--include-path "$p"); done
 
   local latest next_num latest_num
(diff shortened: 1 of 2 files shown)


fortmarek:
Going a different way — superseded by #11557.

Making the gate path-based would rebuild the xcresult-processor image on ~every `server/lib/tuist/**` change, and that Packer build runs on a single bare-metal Mac mini and is expensive. We'd rather keep the gate conservative (scope-gated) and trigger a rebuild manually when a schema/runtime-affecting server change lands under a skipped scope. #11557 does that for the post-#11460 drift and documents the runbook in the image's AGENTS.md.

PR #11555, opened 2026-06-29
perf(cli): consolidate Swift framework search paths

pepicrft:
## What changed

Tuist now consolidates Swift framework search paths for targets that depend on many cached or external precompiled frameworks.

The mapper creates a generated per-target search directory under `Derived/FrameworkSearchPaths/Swift/<target>`, adds symbolic links for uniquely named `.framework` and `.xcframework` artifacts, and passes that single directory through `OTHER_SWIFT_FLAGS`.

It also adds side-effect support for symbolic links and cleans stale generated framework links when the graph changes.

## Why

A reproduction showed that large projects can emit hundreds of Swift `-F` entries when cached frameworks are restored into separate artifact directories. Swift import resolution walks those search paths while resolving imports, so every additional generated framework directory adds repeated filesystem work during clean builds.

## Root cause

`FrameworkSearchPathsGraphMapper` already consolidated the framework paths used by Clang and the linker into response files, but Swift still received every precompiled framework directory inline. That avoided an Xcode 26 issue where the ClangImporter and integrated SwiftDriver mishandle an `@file` token in Swift flags, but it left Swift with the full generated search-path list.

## Approach

This keeps the response-file behavior for Clang and linking unchanged, then gives Swift a separate generated lookup root.

Only artifacts with unique basenames are linked into the generated Swift search directory. Conflicting framework names stay inline so lookup behavior remains deterministic, and unsupported precompiled artifact paths also stay inline.

## Impact

Targets with many precompiled frameworks pass far fewer Swift framework search paths, which reduces the filesystem work during import resolution. Smaller targets below the existing consolidation threshold keep the previous `FRAMEWORK_SEARCH_PATHS` behavior.

## Validation

- `mise run cli:lint --fix`
- `tuist generate tuist TuistGenerator TuistGeneratorTests ProjectDescription --no-open`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistGeneratorTests/FrameworkSearchPathsGraphMapperTests -only-testing TuistGeneratorTests/SideEffectDescriptorExecutorTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`

### DIFF
--- a/cli/Sources/TuistCore/Descriptors/SideEffectDescriptor.swift
+++ b/cli/Sources/TuistCore/Descriptors/SideEffectDescriptor.swift
@@ -17,6 +17,9 @@ public enum SideEffectDescriptor: Equatable, CustomStringConvertible {
     /// Create / remove a directory
     case directory(DirectoryDescriptor)
 
+    /// Create / remove a symbolic link
+    case symbolicLink(SymbolicLinkDescriptor)
+
     /// Perform a command
     case command(CommandDescriptor)
 
@@ -35,6 +38,8 @@ public enum SideEffectDescriptor: Equatable, CustomStringConvertible {
             return fileDescriptor.description
         case let .directory(directoryDescriptor):
             return directoryDescriptor.description
+        case let .symbolicLink(symbolicLinkDescriptor):
+            return symbolicLinkDescriptor.description
         case let .command(commandDescriptor):
             return commandDescriptor.description
         case let .testPlan(descriptor):
--- a/cli/Sources/TuistCore/Descriptors/SymbolicLinkDescriptor.swift
+++ b/cli/Sources/TuistCore/Descriptors/SymbolicLinkDescriptor.swift
@@ -1,0 +1,47 @@
+import Path
+
+/// Symbolic Link Descriptor
+///
+/// Describes a symbolic link operation that needs to take place as part of generating a project
+/// or workspace.
+///
+/// - seealso: `SideEffectDescriptor`
+public struct SymbolicLinkDescriptor: Equatable, CustomStringConvertible {
+    public enum State {
+        case present
+        case absent
+    }
+
+    /// Path to the symbolic link.
+    public var path: AbsolutePath
+
+    /// Path the symbolic link points to.
+    public var destination: AbsolutePath
+
+    /// The desired state of the symbolic link.
+    public var state: State
+
+    /// Creates a SymbolicLinkDescriptor.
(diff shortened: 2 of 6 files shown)


fortmarek (commented):
I took another pass on the latest head (`68ac36c`). The dangling symlink replacement case looks covered now, but I still have one correctness concern before I would trust this in real generated projects.

**[P1] Cleanup can traverse generated symlink directories and delete files inside the real artifact**

The Swift search path cleanup is registered against the broad `Derived/FrameworkSearchPaths` directory with recursive patterns for `**/*.framework` and `**/*.xcframework`:

https://github.com/tuist/tuist/blob/68ac36c6db564755c744010fb558f5852fec6e46/cli/Sources/TuistGenerator/Mappers/FrameworkSearchPathsGraphMapper.swift#L160-L167

Those generated entries are symlinks to framework/xcframework directories. The `FileSystem.glob` implementation follows symlinked directories while walking, so an active link such as `Swift/App/Foo.xcframework -> /cache/Foo.xcframework` can also enumerate nested matches like `Swift/App/Foo.xcframework/ios-arm64/Foo.framework`. Only the top-level link is in `activeFilesByDirectory`, so the cleanup executor will treat that nested framework path as stale. Because that path resolves through the symlink, `removeExistingEntry` can end up deleting the nested framework directory from the real cached artifact rather than just deleting a generated link:

https://github.com/tuist/tuist/blob/68ac36c6db564755c744010fb558f5852fec6e46/cli/Sources/TuistGenerator/Descriptors/SideEffectDescriptorExecutor.swift#L110-L130

I think the cleanup here needs to be symlink-node aware: enumerate/remove only the generated links themselves, without descending into symlink targets. A focused regression test would be to create a real `Module.xcframework/ios-arm64/Module.framework`, symlink `Derived/FrameworkSearchPaths/Swift/App/Module.xcframework` to it, run the framework-link cleanup with the top-level link marked active, and assert the nested real `Module.framework` still exists.

Related confidence note: the current validations cover mapper output, basic symlink creation, dangling link replacement, and stale broken-link cleanup. I would still want a real clean build with 20+ cached xcframeworks, plus an incremental rebuild where an artifact changes but keeps the same basename, before calling the symlink search-path approach production-safe.

fortmarek (approved):

PR #11563, opened 2026-06-29
docs(infra): restore drill for the Barman Cloud Plugin recovery path

fortmarek:
Follow-up to the approving review on #11544 (pepicrft): before flipping `…backup.plugin.enabled` in any env, the restore-validation drill needs to recover through the plugin, not in-tree `barmanObjectStore`.

Updates `infra/cnpg/README.md` so the drill documents the plugin recovery shape: `bootstrap.recovery` + `externalClusters[].plugin` referencing the chart-rendered `ObjectStore` by `barmanObjectName`, with `serverName` pinned to the cluster name (the archive prefix). Keeps the in-tree shape too, since the flag is per-cluster during the transition.

Docs-only (restore is an ad-hoc operator action, not chart-rendered), so no deploy impact. This is the prerequisite the review flagged for the cutover; the plugin install itself (#11544) is unaffected.


fortmarek:
Folded into #11564 — the restore-drill docs now ship in the same PR as the cutover, so the recovery docs land exactly when the flip happens.

PR #11586, opened 2026-06-30
feat(server): module invalidation analytics on Module Cache

fortmarek:
## What changed

Adds per-module cache analytics to Module Cache so teams can prioritise modularisation work: which modules miss most often, why they miss, and how much of the project each one rebuilds when it changes.

**Modules page** (`/module-cache/modules`, a sidebar entry under Module Cache)

- Analytics card whose widgets each drive their own chart: **Modules** (how many the project has, from the latest commit on its default branch), **Hits**, and **Misses** with a dropdown over All, Changed, Upstream and Cold.
- Table of Module, Misses, Cache hit rate and Dependents. Cursor paginated 25 at a time, sorted from the header with a direction toggle, searchable, and scoped by environment and date range.

**Module detail page** (`/module-cache/modules/:module`)

- The same four analytics: Hits, Misses with the reason dropdown, Cache hit rate, and Dependents, each with its own chart.
- **Builds card**: every build the module took part in, newest first, with the scheme, whether it was a local hit, a remote hit or a miss, why it missed, the branch, the commit sha and when it ran. Rows link to the run. Filterable by branch, commit sha prefix and reason, sortable by time, cursor paginated.

**Module Cache dashboard**: a Modules card with the five most-missed modules and summary widgets, sharing one table component with the modules page so the two cannot drift apart.

## Why count-based, not build-time

The first instinct was "wasted build time", but summing per-target `binary_build_duration` ignores Xcode's parallelism: independent targets compile concurrently, and a target only adds wall clock if it sits on the critical path, so the ranking would be unreliable. Miss frequency and dependent count are exact and a truer modularisation signal, since "causes many downstream rebuilds" is the cause of long builds, measured directly.

## How it works

**Why a module missed.** A ClickHouse window query (`lagInFrame`) diffs each missed build against that module's previous build on the same branch, so `main` is never compared against a feature branch. Changed means its own content subhashes differ; upstream means only `dependencies_hash` or `external_hash` moved; cold covers a miss with no comparable prior build, which is first-seen or evicted. Cold is a first-class column rather than an unlabelled remainder, because on a busy repo it is the majority: most builds are the first for that module on their branch.

**Dependents.** Reverse reachability (BFS over inverted edges) across the module dependency graph. This adds a `dependencies Array(String)` column to `xcode_targets` carrying graph edges, which the CLI now sends per target. The graph is stored per build, so the detail chart rebuilds it per day and a refactor that adds or removes dependents shows as a step.

**Scheme on the Builds table.** Command events that produce an activity log carry the build run they belong to, which is where the scheme lives, so the query joins `build_runs` through `build_run_id`. Roughly half the events carrying cacheable targets link this way (every `tuist test` run does); `generate` and `cache` produce no activity log and read "None".

**Two different pagination strategies, deliberately.** The modules table pages in memory: Dependents is computed in Elixir by walking the graph, so a SQL `ORDER BY` cannot see it, and the query returns one row per module, which is bounded by the project rather than growing. Its fetch is guarded to re-run only when the environment or date range changes, so sorting, searching and paging never go back to the database. A module's build history grows without limit, so that one is cursor paginated in ClickHouse with `(ran_at, id)` cursors, stable across builds sharing a timestamp.

## Vocabulary

Cache outcomes use the terms the dashboard already uses: hits, misses, hit rate, and Local, Remote and Missed badges matching the Module Cache tab on the build run page. Invalidation language is reserved for cause and propagation, where it is not a synonym for a miss: a miss is changed, upstream or cold, and a module invalidates downstream. "Changed" and "upstream" follow established build-system usage, where a node is changed when its own input moved and a miss propagates downstream.

## Developer impact

- New ClickHouse migration `20260626120000_add_dependencies_to_xcode_targets`.
- The `xcode_targets` changeset reads the edge list from the build payload and is empty for older clients. The OpenAPI spec and generated Swift client were regenerated accordingly.
- `server/data-export.md` updated for the new column.
- Seeds now produce data that reflects production: command events carry a branch and a commit sha, cache outcomes vary per build rather than being fixed per target, and `test` events link to a build run the way they do in production. Previously a module was either always a hit or always a miss, and no cache-carrying event had a scheme.

## Bugs found while building this

- Rows in the shared table carried no `:id`, so every row landed on the same DOM id and LiveView could not tell them apart while patching. This affected the dashboard card too.
- Sorting had no tiebreaker, so modules with equal miss counts came back in whatever order ClickHouse produced. Cursors address rows by position and need a total order.
- Changing the sort kept the pagination cursor, which points at a position in the previous order, so it silently landed the reader mid-list.
- The reason series partitioned its window by branch alone, which was correct only while the query was always filtered to one module. Once it covered the project it compared one module against whichever sorted next, so almost nothing could be attributed upstream.
- The analytics widgets first summed the table, which lists only modules that missed at least once, so hits and hit rate omitted every perfectly cached module.

## Validation

- Verified against a seeded dev database rather than by reading the code: every analytics function returns the intended result, and the queries run in ~250ms on ~80k command events.
- 115 tests covering classification (changed, upstream, cold, branch isolation), dependents, the project-wide series, module counting across commits, build history filters and paging in both directions, widget selection driven through actual clicks, table pagination, and sorting.
- `mix credo` clean across the repo.

## Follow-ups (deliberately out of scope)

**Expose the new analytics over MCP and REST.** The graph edges reach the server through the public API, but nothing reads them back: no endpoint or MCP tool returns `xcode_targets.dependencies`, and none of the analytics functions have a surface outside the LiveViews. So an agent can see which targets missed in one run but not which module churns most, why, or what depends on it, which is exactly what it would want when advising on modularisation. Worth noting for whoever picks this up: the REST endpoint and the MCP target tool already expose a field named `dependencies`, and it is `dependencies_hash`, a subhash string rather than the edge array, so adding the real one needs disambiguating.

This PR is scoped to the module (binary) cache. The same three ideas, miss frequency, why it missed, and downstream impact, extend to our other caches, but each needs data we do not collect yet:

**Gradle** is the cheaper of the two. Module attribution is already free: `gradle_tasks.task_path` (for example `:feature:login:compileKotlin`) yields the module by prefix, and `cache_key`, `outcome` and `remote_cache_miss` give the same key-diff and hit/miss signal, so the frequency ranking ports almost as-is. What is missing is input-fingerprint decomposition (one opaque `cache_key`, with no equivalent of our subhashes) and task and project graph edges. The Tooling API can supply both.

**Xcode compilation cache** has the highest ceiling and the largest gap. The natural graph is module interface to dependent compilations, which answers "whose header or module interface change recompiles the most files", a classic Swift and ObjC build-time driver. But `cacheable_tasks` has no target or module column at all (`description` is only free text such as "Compiling NetworkManager.m"), so there is nothing to group by yet; per-task target attribution is the prerequisite. It also carries known analytics gotchas (not scannable by `inserted_at`, and run-level counters undercount locally replayed keys) that a port would have to handle.


server/lib/tuist/builds/analytics.ex
```diff
@@ -2252,6 +2252,907 @@ defmodule Tuist.Builds.Analytics do
...
+        []
+    end
+  end
+
+  # Returns each module's most recent dependency edges (in the window/filters) as a
+  # `%{module_name => [dependency names]}` map. Uses `argMax(dependencies, ran_at)`
+  # per module so every module gets its latest known edges independently — a build
+  # that doesn't include some module doesn't drop that module's edges. Empty when no
+  # build carries edges yet, which keeps blast radius unknown (nil).
+  defp latest_graph_dependencies(opts) do
```
pepicrft, line 2383: We group by target name alone here. Two projects with the same-named target (say `Support`) get merged: miss counts sum, the `PARTITION BY name, product, branch` window in `module_invalidations` and `module_miss_reasons_timeseries` compares one project's build against the other's (so a cross-project hash difference reads as `changed`), and `argMax` in `latest_graph_dependencies` overwrites one project's edges with the other's. Should key by name + project.

server/lib/tuist/builds/analytics.ex
```diff
@@ -2252,6 +2252,907 @@ defmodule Tuist.Builds.Analytics do
...
+    {filter_sql, filter_params} = module_invalidation_filters(opts)
+
+    params =
+      Map.merge(%{project_id: project_id, start: start_datetime, end: end_datetime}, filter_params)
+
+    query = """
+    SELECT
+      toDate(e.ran_at) AS day,
+      xt.name AS name,
+      argMax(xt.dependencies, e.ran_at) AS deps
```
pepicrft: `argMax(xt.dependencies, e.ran_at)` returns `[]` on days where the newest event that day came from a pre-graph CLI. The `edges_for_day` fallback only carries the previous day forward when every module is empty, so on a mixed-CLI day this module's edges silently drop while its neighbours' don't, and the dependents chart bounces to 0.

server/lib/tuist_web/live/module_cache_live.ex
```diff
@@ -56,6 +58,7 @@ defmodule TuistWeb.ModuleCacheLive do
       |> assign(:current_params, params)
       |> assign_analytics(params)
       |> assign_recent_runs(params)
```
pepicrft, line 60: `assign_recent_runs` doesn't take `analytics_branch` into account. The widgets and the Modules card requery scoped to the selected branch, but the Recent Runs list keeps returning every branch, so we show filtered and unfiltered results together on the same page.

server/lib/tuist_web/live/module_cache_module_live.ex
```diff
@@ -0,0 +1,327 @@
...
+           miss_reasons_series: miss_reasons_series
+         }}
+      end
+    )
+  end
+
+  # When a module has invalidations its row exists; otherwise synthesize a
+  # zeroed row from the time series so the page still renders (e.g. a module
+  # that only ever reused from cache in the window).
+  defp build_module(nil, name, timeseries) do
```
pepicrft: The nil branch in `build_module` hardcodes `blast_radius: nil`. A module that was 100% cached in the window (or lives past the top 1000 we fetch) falls through here, and its Dependents widget reads 0 on the detail page even though `latest_graph_dependencies` knows the real number.

server/lib/tuist_web/live/module_cache_module_live.ex
```diff
@@ -0,0 +1,327 @@
...
+               end_datetime: end_datetime
+             )
+         }}
+      end)
+
+    assign_async(
+      socket,
+      [:module, :timeseries, :dependents_series, :miss_reasons_series],
+      fn ->
+        all_modules = opts |> Keyword.put(:limit, 1000) |> Analytics.module_invalidations()
```
pepicrft: We fetch the top 1000 missed modules and then look up the target by name. On a big project the module can sit past that cutoff (it's a bounded top-N by miss count, not the specific module we care about). `index[name]` is nil, `build_module(nil, ...)` fires, and the detail page loses product, miss reasons, and blast_radius for a module that actually has data. Better to fetch that one module by name directly rather than search the top-N.

server/lib/tuist_web/live/modules_live.ex
```diff
@@ -0,0 +1,313 @@
...
+    opts = [project_id: project.id, start_datetime: start_datetime, end_datetime: end_datetime]
+
+    case env do
+      "ci" -> Keyword.put(opts, :is_ci, true)
+      "local" -> Keyword.put(opts, :is_ci, false)
+      _ -> opts
+    end
+  end
+
+  defp sort_modules(modules, "blast_radius", order), do: sort_by_value(modules, &(&1.blast_radius || -1), order)
```
pepicrft: Ascending sort by Dependents coerces unknown `blast_radius` to `-1`, so any module whose CLI hasn't reported edges yet lands ahead of real 0-dependent leaves. If someone sorts asc to find leaves, they can't tell 'no dependents' from 'no data yet'. Either surface nils explicitly or push them to the end regardless of direction.

server/lib/tuist/builds/analytics.ex
```diff
@@ -2252,6 +2252,907 @@ defmodule Tuist.Builds.Analytics do
...
+    ClickHouseRepo.all(
+      from(e in Event,
+        where:
+          e.project_id == ^project_id and
+            e.ran_at >= ^start_datetime and
+            e.ran_at <= ^end_datetime and
+            e.cacheable_targets_count > 0 and
+            not is_nil(e.git_branch) and e.git_branch != "",
+        distinct: true,
+        order_by: [asc: e.git_branch],
```
pepicrft: Capped at 100 alphabetically-first branches with no truncation hint. Long-lived projects with more than 100 active branches in the window silently lose the tail (anything from later in the alphabet) from the branch dropdown.

server/lib/tuist/builds/analytics.ex
```diff
@@ -2252,6 +2252,907 @@ defmodule Tuist.Builds.Analytics do
...
+    )
+    WHERE commit = (
+      SELECT argMax(#{String.replace(commit_key, "%{alias}", "e2")}, e2.ran_at)
+      FROM xcode_targets AS xt2
+      INNER JOIN command_events AS e2 ON xt2.command_event_id = e2.id
+      WHERE e2.project_id = {project_id:Int64}
+        AND e2.ran_at >= {start:DateTime64(6)}
+        AND e2.ran_at <= {end:DateTime64(6)}
+        AND e2.git_branch = {branch:String}
+        AND xt2.binary_cache_hash IS NOT NULL#{String.replace(filter_sql, "e.", "e2.")}
```
pepicrft: Building the inner subquery via `String.replace(filter_sql, "e.", "e2.")` is fragile. Any future filter that carries the two chars `e.` outside an alias position (a bound value like `foo e.bar`, an `endsWith(..., 'e.foo')` clause, a new table alias starting with `e`) gets silently rewritten. Safer to build both clauses from a common builder that takes the alias as a parameter.

server/lib/tuist_web/live/module_cache_live.ex
```diff
@@ -281,6 +290,53 @@ defmodule TuistWeb.ModuleCacheLive do
     end)
   end
 
+  # One row per module, so this bounds the project's module count rather than a
+  # growing event stream.
+  @summary_module_limit 5_000
+
+  defp assign_module_invalidations(%{assigns: %{selected_project: project}} = socket, _params) do
```
pepicrft, line 304: We rerun `module_invalidations` (up to 5000 rows), `cache_branches`, and `summarize_invalidations` on every URL change here. `modules_live.ex` guards on `if opts == socket.assigns[:modules_opts]`; this one doesn't, so toggling widget selection or hit-rate type on the dashboard pays the full ClickHouse cost.

server/lib/tuist/builds/analytics.ex
```diff
@@ -2252,6 +2252,907 @@ defmodule Tuist.Builds.Analytics do
...
+      toDate(e.ran_at) AS day,
+      countIf(xt.binary_cache_hit = 'miss') AS invalidations,
+      countIf(xt.binary_cache_hit != 'miss') AS reuses
+    FROM xcode_targets AS xt
+    INNER JOIN command_events AS e ON xt.command_event_id = e.id
+    WHERE e.project_id = {project_id:Int64}
+      AND e.ran_at >= {start:DateTime64(6)}
+      AND e.ran_at <= {end:DateTime64(6)}
+      AND xt.binary_cache_hash IS NOT NULL
+      AND xt.binary_cache_hash IS NOT NULL#{filter_sql}#{name_sql}
```
pepicrft, line 2968: `AND xt.binary_cache_hash IS NOT NULL` appears on two consecutive lines. Harmless but reads like a merge slip.

pepicrft (approved):
Nice work. Comments inline; nothing blocking, but a few worth folding in before we merge or as a follow-up.

server/lib/tuist/builds/analytics.ex
fortmarek: Fixed: the day's edges now come from `argMaxIf(xt.dependencies, e.ran_at, notEmpty(xt.dependencies))`, so a build from a pre-graph CLI no longer blanks that module while its neighbours keep theirs.

server/lib/tuist_web/live/module_cache_live.ex
fortmarek: Fixed: `assign_recent_runs` now applies the selected branch, so the list is scoped the same way as the card and widgets above it.

server/lib/tuist_web/live/module_cache_module_live.ex
fortmarek: Fixed: the synthesized row calls `module_dependents_count/1`, which reads the count from the same graph and returns nil only when no build in the window carries edges.

server/lib/tuist_web/live/module_cache_module_live.ex
fortmarek: Fixed: the page now asks for its own module by name rather than searching a bounded top-N. `latest_graph_dependencies` is never name-scoped, so the dependent count still sees the whole graph.

server/lib/tuist_web/live/modules_live.ex
fortmarek: Fixed: unknown counts are split out and appended after the sorted known ones, so they land at the end in both directions instead of ranking as -1.

server/lib/tuist/builds/analytics.ex
fortmarek: Fixed: it now takes the busiest 100 branches by run count and sorts those for the dropdown, so the cut falls on branches nobody filters by rather than on the tail of the alphabet.

server/lib/tuist/builds/analytics.ex
fortmarek: Fixed: `module_invalidation_filters/2` takes the alias as a parameter and both clauses are built from it, so nothing rewrites the generated SQL.

server/lib/tuist_web/live/module_cache_live.ex
fortmarek: Fixed: guarded on the analytics opts the same way `modules_live.ex` is, so widget and chart-type selections no longer requery.

server/lib/tuist/builds/analytics.ex
fortmarek: Fixed, removed.

server/lib/tuist/builds/analytics.ex
fortmarek: Not changed, with a measurement behind it. Across 1,996,825 (build, target name) pairs in 3 hours of production, no build contains two targets sharing a name (max rows per name = 1), so the merge does not happen on real data today.

It is a real hazard in the Xcode model though, and a full fix needs the CLI: `CommandEventFactory` sends edges as `.map(\`.target.name)`, bare names with no project, so keying rows by name + project would leave the graph unable to resolve which `Support` an edge points at. Adding the project to the grouping and the window alone would also mean joining `xcode_projects` into three hot queries for a case that occurs zero times in 2M samples, while the graph stayed ambiguous.

Tracking it as a follow-up: qualify the edge names in the CLI first, then key rows and the graph by project.

Issue #11596, opened 2026-07-01
App Clip can't have bundle as dependencies


WalterWong-Wise:
### What happened?

Assigning bundle as dependency to AppClip target throws error.

```
Target `AppClipTarget` has platforms 'iOS' and product 'appClip' and depends on target `AppClipBundle` of type 'bundle' and platforms 'iOS' which is an invalid or not yet supported combination.
```

### How do we reproduce it?

1. Create a project with two targets, one of type `appClip`, another of type bundle
2. Add the bundle target as a dependency of the app clip target
3. Run tuist generate

### Error log

```
Target `AppClipTarget` has platforms 'iOS' and product 'appClip' and depends on target `AppClipBundle` of type 'bundle' and platforms 'iOS' which is an invalid or not yet supported combination.
```

### macOS version

26.5.1

### Tuist version

4.174.2

### Xcode version

26.5

WalterWong-Wise:
@dosu I have created a PR to add this capability. Requires review

PR #6415, opened 2024-06-17
Allow iOS app extensions to depend on bundles

DevilDimon:
Resolves https://github.com/tuist/tuist/issues/6406

### Short description 📝

Previously, Tuist didn't allow iOS app extensions to depend on bundle targets. This is, however, a combination that is possible with Xcode and is natively supported in Xcodegen, for example.

After reporting the issue I received confirmation this was a bug. This PR fixes it by adding the combination to the graph linter. For testing, the PR extends the `app_with_extensions` fixture & acceptance tests by adding a dummy bundle that is then made a dependency of one of the extensions. I also added a unit test for the graph linter.

### How to test the changes locally 🧐

1. Generate the `app_with_extensions` Tuist project. The generate command should be successful
2. Open the resulting workspace and build any target that depends on the Bundle target (for example, App)
3. Open the derived data and navigate to the build products of the build, go inside the package contents
4. Bundle.bundle should be there

### Contributor checklist ✅

- [x] The code has been linted using run `mise run lint-fix`
- [x] The change is tested via unit testing or acceptance testing, or both
- [x] The title of the PR is formulated in a way that is usable as a changelog entry
- [ ] In case the PR introduces changes that affect users, the documentation has been updated

### Reviewer checklist ✅

- [ ] The code architecture and patterns are consistent with the rest of the codebase
- [ ] Reviewer has checked that, if needed, the PR includes the label `changelog:added`, `changelog:fixed`, or `changelog:changed`, and the title is usable as a changelog entry

PR #11597, opened 2026-07-01
[FEAT] Added bundle support to AppClip

WalterWong-Wise:
Resolves 
-https://github.com/tuist/tuist/issues/11596

### Description
When declaring a .bundle target and making an .appClip depend on it, Tuist would reject the graph with a lint error — even though App Clips fully support resources and can embed bundles.

The root cause was a one-line omission in GraphLinter.swift. PR #6415 added .bundle as a valid dependency for .appExtension targets but the same entry was never added for .appClip.

The rest of the pipeline (resource embedding, Bundle.module accessor generation, canEmbedBundles) already handles .appClip correctly — only the linter was blocking it.

### How to test locally
1. Generate `appClip` target with `bundle` as dependency
2. Run `generate` command, it will be successful.
3. Open workspace and run
4. Navigate to the build products of the build, go inside the package contents

### DIFF
--- a/cli/Sources/TuistGenerator/Linter/GraphLinter.swift
+++ b/cli/Sources/TuistGenerator/Linter/GraphLinter.swift
@@ -548,6 +548,7 @@ public struct GraphLinter: GraphLinting {
             LintableTarget(platform: .iOS, product: .framework),
             LintableTarget(platform: .iOS, product: .staticFramework),
             LintableTarget(platform: .iOS, product: .appExtension),
+            LintableTarget(platform: .iOS, product: .bundle),
             LintableTarget(platform: .macOS, product: .macro),
         ],
         LintableTarget(platform: .iOS, product: .extensionKitExtension): [
--- a/cli/Tests/TuistGeneratorAcceptanceTests/GenerateAcceptanceTests.swift
+++ b/cli/Tests/TuistGeneratorAcceptanceTests/GenerateAcceptanceTests.swift
@@ -909,6 +909,11 @@ struct GenerateAcceptanceTestiOSAppWithAppClip {
             destination: "Debug-iphonesimulator",
             extension: "AppClip1Widgets"
         )
+        try await XCTAssertProductWithDestinationContainsResource(
+            "AppClip1.app",
+            destination: "Debug-iphonesimulator",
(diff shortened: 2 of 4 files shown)


WalterWong-Wise:
Hello there @fortmarek, may I request a review on this PR? Thank you, sir 🙏🏻

WalterWong-Wise:
Sorry for tagging you again. @fortmarek just wondering what is the process to get this reviewed? Thank you !

pepicrft:
Thanks @WalterWong-Wise 🙏🏼. I'm reviewing it

pepicrft (approved):
Reviewed the PR and it looks good. The one-line fix in `GraphLinter.validLinks` is exactly the right approach — it mirrors what PR #6415 did for `.appExtension` → `.bundle`, and the rest of the pipeline (`resourceBundleDependencies`, `supportsResources`, `generateResourceBundle`) already handles AppClips correctly.

**What I verified:**
- `Target.supportsResources` already returns `true` for `.appClip`, so `GraphTraverser.resourceBundleDependencies` will discover bundle deps and embed them via the resources build phase.
- The `StaticProductsGraphLinter` `(.app, .appClip)` case already stops traversal, so there's no risk of double-link warnings when an app embeds an AppClip that depends on a bundle.
- The unit test (`lint_appClip_canDependOnBundle`) is placed right after the analogous `lint_appExtension_canDependOnBundle` and uses the same `Target.empty` pattern. Good.
- The example project changes are identical in shape to the existing `generated_ios_app_with_extensions` Bundle target — same bundleId convention, same empty `dummy.jpg` resource, same missing `infoPlist: .default` (which is intentional for bundles; `InfoPlistContentProvider` handles it).

**One non-blocking observation:** The acceptance test assertions are added to `GenerateAcceptanceTestiOSAppWithAppClip.ios_app_with_appclip`, which is marked `@Test(.disabled())` (disabled in #10128 during CI consolidation). So these assertions document expected behavior but won't run in CI. The unit test is what actually guards the linting rule. Not a blocker since the acceptance test was already disabled before this PR.

pepicrft:
Closing in favor of #12115 which includes the same changes plus a re-enabled acceptance test, pushed from a non-fork branch so CI runs fully.

PR #11603, opened 2026-07-01
feat(server): enable the Kura cache surface by default on self-hosted, and fix self-hosting docs

fortmarek:
## Purpose

Fixes the self-hosting guides where they were inaccurate/out-of-date, and removes the setup friction that surfaced while helping a customer stand up a standalone node: on a self-hosted server the Cache page was hidden behind a feature flag they couldn't easily reach.

## Behavior change: Kura on by default for self-hosted

On a self-hosted Tuist server, cache routing is driven by the operator's global `TUIST_CACHE_ENDPOINTS`, not by the per-account `:kura` rollout flag. Gating the Cache page (plus the sidebar entry, Usage page, and settings section) behind `:kura` was therefore pure friction: an operator had to reach `/ops/flags` to flip a flag before they could generate the control-plane credential, even though the deployment already grants the self-hosted-cache entitlement unconditionally.

- `FeatureFlags.kura_enabled?/1` now returns true on any non-hosted deployment (`not tuist_hosted?()`), mirroring `Billing.Entitlements` where the deployment's license is the entitlement. The hosted server stays gated behind the `:kura` toggle.
- The inline duplicates in `cache_live` and `account_settings_live` now delegate to that helper so the sidebar, LiveView guards, and settings page can't drift.
- The managed cache-server provisioning UI (inert on self-hosted: no managed regions, no Tuist-managed Kubernetes) is hidden when there are no servers and no available regions, so a self-hoster sees only credential generation and the registered-nodes list.

## Docs corrections

Tracing the CLI endpoint resolution (`Accounts.get_cache_endpoints_for_handle/2`) showed it branches on `tuist_hosted?`:
- **Hosted tuist.dev:** per-account resolution routes the CLI to self-hosted nodes from their registration heartbeats.
- **Self-hosted server:** that branch is never reached; it returns the global `TUIST_CACHE_ENDPOINTS`.

The docs instead told self-hosters to set `TUIST_KURA_ENDPOINTS` / `server.kuraEndpointUrls`, which has no consumer (`Environment.kura_endpoints/2` is never called), so that guidance was a no-op. Corrected across all three guides, plus:
- Registration is clarified as dashboard-only on a self-hosted server (routing is `TUIST_CACHE_ENDPOINTS`).
- Peer mTLS (`KURA_INTERNAL_TLS_*`) is documented as multi-node-only; a single node uses an `http://` `KURA_NODE_URL` with peer TLS off.
- The Cache page is documented as available by default on a self-hosted server.

## Scope / follow-ups

The dead `TUIST_KURA_ENDPOINTS` / `Environment.kura_endpoints/2` code and the `kuraEndpointUrls` Helm value (whose comment still says "point this at Kura instances") are left for a separate cleanup.

## Validation

- `mix compile --warnings-as-errors` passes (confirms the removed `Environment` aliases and the delegations).
- New tests in `cache_live_test`: self-hosted default-on, and the managed-section hide when no regions are available. `disable_cache`/`disable_kura` helpers now stub `tuist_hosted? -> true` (the only case the surface stays hidden). Full suite runs in CI.

### DIFF
--- a/server/lib/tuist_web/live/cache_live.html.heex
+++ b/server/lib/tuist_web/live/cache_live.html.heex
@@ -11,7 +11,7 @@
 
   <%= if @cache_enabled do %>
     <.cache_servers_section
-      :if={@servers != [] or @available_regions != []}
+      :if={@managed_cache_visible?}
       servers={@servers}
       available_regions={@available_regions}
       add_cache_server_form={@add_cache_server_form}
--- a/server/lib/tuist_web/live/cache_live.html.heex
+++ b/server/lib/tuist_web/live/cache_live.html.heex
@@ -11,6 +11,7 @@
 
   <%= if @cache_enabled do %>
     <.cache_servers_section
+      :if={@servers != [] or @available_regions != []}
       servers={@servers}
       available_regions={@available_regions}
       add_cache_server_form={@add_cache_server_form}
--- a/server/priv/docs/en/guides/cache/self-host.md
+++ b/server/priv/docs/en/guides/cache/self-host.md
@@ -53,13 +53,20 @@ The key difference in configuration is that the **bridged** topology uses **enro
 
 ## Create a control-plane client {#control-plane-client}
 
-Every node authenticates to Tuist with an account-scoped control-plane client.
+A node uses an account-scoped control-plane client to authenticate cache requests (token introspection), report to the dashboard, and deliver usage. It is **not** how clients are routed to your nodes (see [How clients reach your nodes](#routing)), so a single node on a trusted network can run without one.
 
-1. Open your account's **Cache** page.
+1. Open your account's **Cache** page. If you don't see it, enable the `kura` feature flag for the account. On a self-hosted Tuist server the Enterprise entitlement is granted automatically, so the flag is the only prerequisite.
 2. Choose **Generate credential**.
 3. Copy the `client_id` and the one-time `secret`.
 
 The server derives the account from this credential, so the node never asserts its own tenant. Rotate or revoke it from the same page.
+
+## How clients reach your nodes {#routing}
+
+How the Tuist CLI is pointed at your nodes depends on which server your nodes report to:
+
+- **Hosted Tuist server (`tuist.dev`).** The server routes clients to your nodes automatically from their registration heartbeats. Set `KURA_REGISTRATION_URL` and `KURA_ADVERTISED_HTTP_URL` on each node (below), and the advertised URL is handed to the CLI once the node is ready.
+- **Self-hosted Tuist server.** You tell the server which cache endpoints to advertise by setting `TUIST_CACHE_ENDPOINTS` (Helm `server.cacheEndpointUrl`) to your node's client-facing URL, comma-separated for multiple nodes. Registration heartbeats still populate the **Cache** page, but they do not drive routing on a self-hosted server.
 
 ## Bridged setup {#bridged-setup}
 
@@ -131,7 +138,10 @@ services:
       - "4000:4000"
       - "7443:7443"
     environment:
-      # Register with Tuist (dashboard, CLI routing, usage, introspection)
+      # Report to Tuist: dashboard visibility, usage, and token introspection (cache auth).
+      # On a self-hosted server these do not drive routing (see the Routing section), so the
+      # control-plane client and KURA_REGISTRATION_URL/KURA_ADVERTISED_HTTP_URL are optional
+      # there. KURA_TENANT_ID is always required.
       KURA_TENANT_ID: "<account-handle>"
       KURA_CONTROL_PLANE_URL: "https://tuist.dev"
       KURA_REGISTRATION_URL: "https://tuist.dev/_internal/kura/mesh/registrations"
@@ -143,7 +153,9 @@ services:
       KURA_NODE_URL: "https://kura-1.acme.internal:7443"
       KURA_PEERS: "https://kura-2.acme.internal:7443,https://kura-3.acme.internal:7443"
 
-      # YOU provide these certs: your CA and a leaf per node
+      # Peer mTLS secures node-to-node traffic, so it only applies with more than one node.
+      # A single node has no peers: omit these three and set KURA_NODE_URL to http://...:7443
+      # (the peer URL must be http when peer TLS is off).
       KURA_INTERNAL_TLS_CA_CERT_PATH: "/tls/ca.pem"
       KURA_INTERNAL_TLS_CERT_PATH: "/tls/tls.crt"
       KURA_INTERNAL_TLS_KEY_PATH: "/tls/tls.key"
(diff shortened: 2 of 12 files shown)


esnunes (commented):
Solid, well-scoped change. The env matrix checks out — hosted stays byte-identical (gated by `:kura`), self-hosted is on by default as intended — the inline checks are correctly consolidated into `FeatureFlags.kura_enabled?/1`, and every consumer (both LiveViews, usage_live, and the sidebar) routes through it, so nav and page stay in sync.

---

_Two notes below land on lines this PR doesn't change, so they're here rather than inline:_

**`server/test/tuist_web/live/cache_live_test.exs` (`enable_cache/1`, ~L291) and `usage_live_test.exs` (`enable_kura/1`) — test coverage.** `enable_cache/1` (and `enable_kura/1` in usage_live_test.exs) enables the surface by stubbing `Environment.tuist_hosted? -> false`, so `kura_enabled?` short-circuits on `not tuist_hosted?()` and the `FunWithFlags.enabled?(:kura, ...)` disjunct is never exercised. That means the one branch this flag actually controls in production — hosted tuist.dev with the flag on — has no positive test: if it regressed (renamed flag, wrong actor), the hosted Cache/Usage surface would silently never render and the suite would stay green. Could we add a case that stubs `tuist_hosted? -> true` with the flag on and asserts the surface appears?

**`server/lib/tuist_web/live/cache_live.ex:590` — possibly-dead `<:empty_state>`.** With the new `:if={@servers != [] or @available_regions != []}` guard at the call site, `cache_servers_section` may never mount with an empty table, so its `<:empty_state>` (lines 590-595, "No cache servers available") and those two dgettext strings may be dead.

server/priv/docs/en/guides/features/cache/self-hosting.md
```diff
@@ -32,15 +32,14 @@ helm upgrade --install kura oci://ghcr.io/tuist/charts/kura \
...
 
 ```yaml
 server:
-  kuraEndpointUrls:
-    - http://kura.kura.svc.cluster.local:4000
+  cacheEndpointUrl: "http://kura.kura.svc.cluster.local:4000"
 ```
 
-This renders `TUIST_KURA_ENDPOINTS` in the server pod. When clients request the Kura cache technology, the server returns those endpoints.
+This renders `TUIST_CACHE_ENDPOINTS` in the server pod. On a self-hosted server the CLI is routed to whatever `TUIST_CACHE_ENDPOINTS` lists, so point it at your Kura service. For multiple nodes, use a comma-separated list.
```
esnunes, line 42: The chart renders `TUIST_CACHE_ENDPOINTS` only inside `{{- if .Values.cache.enabled }}` (server-deployment.yaml:282), so `server.cacheEndpointUrl` is a no-op when the bundled cache is disabled. The default is `cache.enabled: true` so the happy path works — but this guide's scenario is "run Kura as its own release", and a self-hoster who sets `cache.enabled: false` will set `cacheEndpointUrl` per these docs and get no `TUIST_CACHE_ENDPOINTS` (routing silently does nothing). Worth a one-line caveat that `server.cacheEndpointUrl` requires `cache.enabled: true` — or rendering the var unconditionally. Separately, the old `server.kuraEndpointUrls` key still ships in the chart with a comment claiming it advertises endpoints, but `TUIST_KURA_ENDPOINTS` now has no reader; a follow-up cleanup would keep the chart and these rewritten docs in sync.

server/lib/tuist_web/live/cache_live.html.heex
```diff
@@ -11,6 +11,7 @@
 
   <%= if @cache_enabled do %>
     <.cache_servers_section
+      :if={@servers != [] or @available_regions != []}
```
esnunes: Now that `kura_enabled?` returns true on self-hosted (where there are no managed regions), every place that renders the managed section has to carry the extra `(@servers != [] or @available_regions != [])` clause — it's duplicated at cache_live.html.heex:14 and account_settings_live.html.heex:15 & :181, expressed three slightly different ways. A fourth render site (or a copied desktop/mobile block) that forgets the clause will resurface an empty "Cache servers" card on every self-hosted deployment. Consider computing one derived assign (e.g. `managed_cache_visible? = kura_enabled? and (servers != [] or regions != [])`) in mount/load_*_state and gating on that, so the rule lives in one place.

server/priv/docs/en/guides/cache/self-host.md
```diff
@@ -53,14 +53,21 @@ The key difference in configuration is that the **bridged** topology uses **enro
 
 ## Create a control-plane client {#control-plane-client}
 
-Every node authenticates to Tuist with an account-scoped control-plane client.
+A node uses an account-scoped control-plane client to authenticate cache requests (token introspection), report to the dashboard, and deliver usage. It is **not** how clients are routed to your nodes (see [How clients reach your nodes](#routing)), so a single node on a trusted network can run without one.
 
-1. Open your account's **Cache** page.
+1. Open your account's **Cache** page. On a self-hosted Tuist server it is available by default (the deployment's license is the entitlement). On the hosted `tuist.dev` server it requires an Enterprise plan and the `kura` feature flag.
```
esnunes: "On the hosted tuist.dev server it requires an Enterprise plan and the `kura` feature flag." — page visibility (`kura_enabled?`) on hosted checks only the `:kura` flag; there's no plan check in that guard. The Enterprise entitlement gates the self-hosted-cache routing/section (`Entitlements.allows?(:self_hosted_cache)`), not whether the Cache page renders. Tightening this to "the `kura` feature flag (Enterprise is required for self-hosted cache routing)" would match the code.

esnunes (approved):

server/priv/docs/en/guides/features/cache/self-hosting.md
fortmarek: Fixed: the chart now renders `TUIST_CACHE_ENDPOINTS` whenever `server.cacheEndpointUrl` is set (`{{- if or .Values.cache.enabled .Values.server.cacheEndpointUrl }}`), so the documented `cacheEndpointUrl` guidance works even with `cache.enabled: false`. Left the dead `kuraEndpointUrls`/`TUIST_KURA_ENDPOINTS` removal as the follow-up you noted.

server/lib/tuist_web/live/cache_live.html.heex
fortmarek: Consolidated into a single derived assign per LiveView (`managed_cache_visible?` / `managed_kura_visible?`), computed in `load_servers_state`/`load_kura_state` and gated at every render site, so the rule lives in one place instead of being repeated three ways.

server/priv/docs/en/guides/cache/self-host.md
fortmarek: Tightened to: the hosted page requires the `kura` flag, and generating a self-hosted credential there additionally requires an Enterprise plan. Matches the code where `kura_enabled?` checks only the flag while `Entitlements.allows?(:self_hosted_cache)` gates the self-hosted section.

PR #11607, opened 2026-07-01
fix(server, infra): grant processor access to automation alerts

pepicrft:
Fixes TUIST-WZ: https://tuist.sentry.io/issues/TUIST-WZ

The production Sentry issue comes from the managed `tuist_processor` [PostgreSQL](https://www.postgresql.org/) role missing read access to `automation_alerts`. Marek's recent pull request [#11602](https://github.com/tuist/tuist/pull/11602) moved processor-role grants into `Tuist.Release.migrate/0`, so this keeps that mechanism and adds `automation_alerts` to the automatic grant list. It also updates the [CloudNativePG](https://cloudnative-pg.io/) fallback runbook so bootstrap and restore flows stay in sync with the migration-time grant surface.

The follow-up review fixes make the migration-time grant application transactional, generate table grants from a single audited list, and add a focused test that keeps the release helper and fallback runbook aligned.

### How to test locally

- `mix deps.get && mix format lib/tuist/release.ex`
- `mix test test/tuist/release_test.exs`
- `git diff --check`


server/lib/tuist/release.ex
```diff
@@ -204,7 +204,8 @@ defmodule Tuist.Release do
           "#{quoted_schema}.oban_jobs, #{quoted_schema}.oban_peers TO #{role}",
         "GRANT USAGE, SELECT ON SEQUENCE #{quoted_schema}.oban_jobs_id_seq TO #{role}",
         "GRANT SELECT ON TABLE " <>
-          "#{quoted_schema}.accounts, #{quoted_schema}.projects, #{quoted_schema}.webhook_endpoints TO #{role}"
+          "#{quoted_schema}.accounts, #{quoted_schema}.projects, #{quoted_schema}.automation_alerts, " <>
+          "#{quoted_schema}.webhook_endpoints TO #{role}"
       ],
       &SQL.query!(repo, &1, [])
```
esnunes: `do_grant_processor_role` runs the `REVOKE ALL` and the re-GRANTs as separate autocommit statements (`Enum.each` + `SQL.query!`, no wrapping transaction), so every managed migrate opens a brief window where `tuist_processor` has no table privileges at all — including `oban_jobs`, which live processors poll continuously during deploys. A query landing in that gap raises the same transient `42501` class this PR is eliminating. The SQL fallback wraps the identical REVOKE + re-GRANT in `BEGIN`/`COMMIT` (`infra/cnpg/tuist-processor-grants.sql:33-63`); wrapping this statement list in `repo.transaction/1` would close the window on the Elixir side too. Pre-existing from #11602, but since this PR is in the function anyway it seems worth folding in.

server/lib/tuist/release.ex
```diff
@@ -204,7 +204,8 @@ defmodule Tuist.Release do
           "#{quoted_schema}.oban_jobs, #{quoted_schema}.oban_peers TO #{role}",
         "GRANT USAGE, SELECT ON SEQUENCE #{quoted_schema}.oban_jobs_id_seq TO #{role}",
         "GRANT SELECT ON TABLE " <>
-          "#{quoted_schema}.accounts, #{quoted_schema}.projects, #{quoted_schema}.webhook_endpoints TO #{role}"
+          "#{quoted_schema}.accounts, #{quoted_schema}.projects, #{quoted_schema}.automation_alerts, " <>
```
esnunes: This is the second enumeration-drift 42501 in two days (#11602 added `webhook_endpoints`, this adds `automation_alerts` — a read that had existed since #10816), and both times the only detector was a production Sentry issue: nothing in `server/test` references `do_grant_processor_role` or exercises the processor workers under the restricted role. Since `CREATE ROLE`, `GRANT`, and `SET LOCAL ROLE` are all transactional in Postgres, a sandbox test could create a scratch role, apply this grant list, `SET LOCAL ROLE`, and drive the two workers' Postgres surface — or, more cheaply, a static test pinning the grant list against an audited allowlist of tables the ingestion path reads would force the next new `Repo` call to update the list at review time instead of after a Sentry alert. Worth adding alongside the second hotfix for the same failure mode?

infra/cnpg/tuist-processor-grants.sql
```diff
@@ -38,11 +39,11 @@ GRANT USAGE ON SCHEMA :"tuist_schema" TO tuist_processor;
...
--- Read-only lookups the worker performs. accounts.get_account_by_id/1
--- resolves the account for S3 scoping; Project |> Repo.get/2 reads the
--- project to decide whether to broadcast PubSub. Neither row is written
--- to by the worker — hence no INSERT/UPDATE.
-GRANT SELECT ON :"tuist_schema".accounts, :"tuist_schema".projects TO tuist_processor;
+-- Read-only lookups the workers perform. The build and xcresult workers
+-- resolve the project account for storage scoping; the xcresult worker also
+-- reads automation alerts when it schedules scoped flaky-test evaluations.
+-- These rows are not written by the processors, hence no INSERT/UPDATE.
+GRANT SELECT ON :"tuist_schema".accounts, :"tuist_schema".projects, :"tuist_schema".automation_alerts TO tuist_processor;
```
esnunes: The table-grant list appears twice in this file: the commented block (lines 39/46/53) is wiped by the `REVOKE ALL` at line 58 and re-granted byte-identically at 59-61 — which is why this PR has to edit lines 46 and 60 with the same change. Whoever adds the next table only to the commented block (the natural landing spot, since it carries the docs) gets a fallback that silently lacks the grant, discovered at the worst time — bootstrap/restore. `do_grant_processor_role` already uses the safer shape (REVOKE first, single grant list); mirroring that here would keep the deny-by-default intent while halving the copies to sync. Relatedly, the README's prose enumeration of the tables (line 10) is a fourth copy — it could defer to this file as the canonical list.

server/lib/tuist/release.ex
```diff
@@ -204,7 +204,8 @@ defmodule Tuist.Release do
...
+          "#{quoted_schema}.webhook_endpoints TO #{role}"
```
esnunes: Tiny one: the read-table GRANT is now three concat segments with the list comma sitting at a string seam (`"…automation_alerts, " <> "…webhook_endpoints…"`). Building it from a table list — `Enum.map_join(~w(accounts projects automation_alerts webhook_endpoints), ", ", &"#{quoted_schema}.#{&1}")` — keeps the SQL identical, makes the next addition a one-word change, and removes the seam where a comma can silently go missing (a malformed GRANT fails the migrate Job and blocks the deploy).

esnunes (approved):
The fix itself checks out: I traced both processor code paths' full Postgres surface (build worker + xcresult worker through `Tests.create_test`), and with `automation_alerts` added the enumerated grant set is complete — and SELECT-only is right, since writes happen only in `AlertEvaluationWorker` on the `:default` queue, which processor pods never consume. The SQL fallback and the Elixir applier are consistent.

The notes are hardening/accuracy follow-ups rather than blockers. The two worth acting on: the Elixir applier's REVOKE→GRANT sequence isn't transactional (unlike the SQL fallback), so every migrate opens a brief zero-privilege window for live processors; and this is the second enumeration-drift 42501 in two days with no test pinning the grant list to the processors' actual Postgres surface — a regression test (scratch role + SET LOCAL ROLE, or a static allowlist pin) would move the next miss from Sentry to CI.

server/lib/tuist/release.ex
pepicrft: Wrapped the processor grant execution in `repo.transaction/1` inside `Tuist.Release.do_grant_processor_role/2`, so the revoke and re-grants now commit atomically. The fallback grant file still uses `BEGIN`/`COMMIT`.

server/lib/tuist/release.ex
pepicrft: Added `server/test/tuist/release_test.exs` to pin `Tuist.Release.processor_role_grant_statements/3` and verify the fallback grant file contains the audited read grant exactly once. Validated with `mix test test/tuist/release_test.exs`.

infra/cnpg/tuist-processor-grants.sql
pepicrft: Mirrored the safer shape in `infra/cnpg/tuist-processor-grants.sql`: revoke first, then grant each table surface once. I also removed the table enumeration from the README prose so the grant file and release helper remain the canonical list.

server/lib/tuist/release.ex
pepicrft: Changed the release helper to build grants from `@processor_write_tables` and `@processor_read_tables` through `qualify_tables/2`, which removes the manual comma seam and makes the next table addition a one-word list change. The new test pins the generated statement.

fortmarek:
I already have a PR for this at https://github.com/tuist/tuist/pull/11604

PR #11616, opened 2026-07-02
fix(cli): align cache warm hashes with generation

fortmarek:
## What changed

- Keep the cache-warm preload graph normalized with `FrameworkSearchPathsGraphMapper`.
- Preserve the same normalized hashing graph before `TargetsToCacheBinariesGraphMapper` runs in generate/build flows that were not already preserving it.
- Add unit coverage for the mapper ordering so cache replacement receives stable hashes in warm, generate, and build paths.
- Add an acceptance test that warms a framework wrapping precompiled dependencies, runs `generate --cache-profile all-possible`, and verifies the warmed binary is actually reused and linked.

## Why

A cache-warmed target could be stored under a different key than the one requested later by `tuist generate --cache-profile all-possible`.

The concrete failure mode was a dynamic framework wrapping external binary/framework dependencies. `tuist cache warm <target>` hashed a graph where generated framework search paths had already been injected into `targetSettings`, while the unfocused `generate --cache-profile all-possible` path computed cache keys from a graph that had not preserved that same normalized state before replacing targets with cache binaries. The target's sources, resources, dependencies, project settings, and other hash components matched, but `targetSettings` diverged, so generate missed the warmed artifact.

The first implementation tried to remove framework-search-path normalization from cache warm. That aligned the `all-possible` reproduction, but it broke existing default/focused cache reuse paths that depend on warm hashing the normalized graph. The final fix keeps warm behavior intact and instead makes generate/build preserve the same normalized hashing graph before binary replacement.

## Regression source

This regression was introduced by [#11054](https://github.com/tuist/tuist/pull/11054) (`de9b946c217`), which moved framework search path setup into `FrameworkSearchPathsGraphMapper`.

That change made generated framework search paths part of the mapped graph earlier in some flows. Cache warm included those settings in its target-settings hash, while unfocused generate/build cache replacement could request binaries using a hash computed before that same normalization had been preserved.

## Approach

The mapper factory now inserts `CacheHashingGraphMapper` before `TargetsToCacheBinariesGraphMapper` for binary-cache generate/build flows that do not already preserve the hashing graph through focused targets. The hashing mapper runs the same normalization mappers used by warm, including framework search paths, so both warm and consume paths derive cache keys from the same graph.

This avoids reintroducing the prior default/focused cache regression because `FrameworkSearchPathsGraphMapper` remains in the cache-warm preload chain. Build-time framework search paths are still generated where the produced Xcode project needs them.

## Impact

Targets warmed with `tuist cache warm <target>` should now be reusable by `tuist generate --cache-profile all-possible` and by build/generate cache replacement paths when their logical inputs are unchanged. The change is limited to cache hashing graph preservation and mapper ordering; it does not change artifact production or upload logic.

## Validation

- Reproduced the mismatch with a local fixture using `4.201.0-rc.5`: warm and generate produced different hashes for the wrapper target.
- Verified the locally built CLI reuses the warmed wrapper binary: generate reports `Using cache binaries for the following targets: Library`, and the generated project links `Library.xcframework`.
- Ran targeted cache acceptance coverage:
  - `generated_macos_tool_with_cached_nested_header_xcframework()`
  - `generated_ios_app_with_cached_xctest_support()`
  - `generate_reuses_warmed_framework_wrapping_precompiled_dependencies()`
- Ran focused mapper unit tests:
  - `TuistKitTests/CacheGraphMapperFactoryTests`
- Ran lint:
  - `mise run cli:lint --fix`
- Confirmed CI is green on the PR after rerunning a transient acceptance shard that timed out downloading the prebuilt test bundle.

### DIFF
--- a/cli/Sources/TuistKit/Mappers/Factories/GraphMapperFactory.swift
+++ b/cli/Sources/TuistKit/Mappers/Factories/GraphMapperFactory.swift
@@ -216,6 +216,15 @@ public struct GraphMapperFactory: GraphMapperFactorying {
             var mappers = self.default(config: config)
 
             if !ignoreBinaryCache {
+                mappers.append(
+                    CacheHashingGraphMapper(
+                        normalizationMappers: hashingGraphNormalizationMappers(
+                            config: config,
+                            includeWorkspaceScheme: true
+                        )
+                    )
+                )
+
                 let focusTargetsGraphMapper = TargetsToCacheBinariesGraphMapper(
                     config: config,
                     decider: CacheProfileTargetReplacementDecider(
@@ -263,6 +272,17 @@ public struct GraphMapperFactory: GraphMapperFactorying {
             )
 
             if cacheProfile != .none {
+                if !shouldPreserveHashingGraph {
+                    mappers.append(
+                        CacheHashingGraphMapper(
+                            normalizationMappers: hashingGraphNormalizationMappers(
+                                config: config,
+                                includeWorkspaceScheme: true
+                            )
+                        )
+                    )
+                }
+
                 let focusTargetsGraphMapper = TargetsToCacheBinariesGraphMapper(
                     config: config,
                     decider: CacheProfileTargetReplacementDecider(
@@ -354,6 +374,7 @@ public struct GraphMapperFactory: GraphMapperFactorying {
                 includedTargets: targetsToBinaryCache
             )
 
+            mappers.append(FrameworkSearchPathsGraphMapper())
             mappers.append(ForeignBuildSideEffectGraphMapper())
 
             mappers.append(
--- a/cli/Sources/TuistKit/Mappers/Factories/GraphMapperFactory.swift
+++ b/cli/Sources/TuistKit/Mappers/Factories/GraphMapperFactory.swift
@@ -354,7 +354,6 @@ public struct GraphMapperFactory: GraphMapperFactorying {
                 includedTargets: targetsToBinaryCache
             )
 
-            mappers.append(FrameworkSearchPathsGraphMapper())
             mappers.append(ForeignBuildSideEffectGraphMapper())
 
             mappers.append(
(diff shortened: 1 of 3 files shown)


esnunes (commented):
Post-merge review notes — adding these for the record.

The fix is correct and a real improvement: `build` and unfocused `generation` now snapshot the same normalized graph that `tuist cache` warms, so warmed binaries stop being silently rebuilt. `CacheHashingGraphMapper` is a nil-guarded pass-through, added exactly once across every branch — verified by an independent correctness pass. No blocker. Three follow-up notes:

---

**Explicit-dependency hashing asymmetry — `GeneratorFactory.swift:425,458` vs `:311,392` (pre-existing; worth a follow-up)**

This aligns the FrameworkSearchPaths/scheme normalization between warm and consume — nice. One gap it doesn't close: the warm pipelines strip `ExplicitDependencyGraphMapper` (`GeneratorFactory.swift:425,458`) but `generation`/`build` don't (`:311,392`). It runs before `CacheHashingGraphMapper` and rewrites `target.settings`, which feeds the content hash (`TargetContentHasher.swift:139`). So for projects with `enforceExplicitDependencies == true`, `generate`/`build` recompute a different hash than `tuist cache` stored, and warmed binaries are silently rebuilt for any target that has dependencies. Note the input hasn't changed between warm and consume — it's the same project at the same commit — so this is a false cache miss, not legitimate invalidation.

It predates this PR, but it's on the same alignment surface. The fix isn't to filter the mapper out of consume (the generated project must keep explicit deps) — it's to take the consume-side hash snapshot *before* `ExplicitDependencyGraphMapper` runs, so both sides hash the pre-explicit-deps graph while the mapper stays in the generated output.

---

**Document the split insertion point — `GraphMapperFactory.swift:275`**

The generation hashing mapper is added in one of two places depending on `shouldPreserveHashingGraph` — here when false, and inside `default(preserveGraphBeforeFocus:)` (line ~451) when true. Reading `generation()` alone that's easy to miss. A one-liner on this gate would help, e.g. `// When preserveGraphBeforeFocus is true, default() already appended the hashing mapper before focusing; otherwise add it here.` so the paired invariant is discoverable without cross-reading `default()`.

---

**Nit: one boolean, two names — `GraphMapperFactory.swift:267`**

`shouldPreserveHashingGraph` is computed then passed straight through as `preserveGraphBeforeFocus:` — same concept, two names. Aligning them (or a short note that they're the same value) removes a small mental hop.

PR #11620, opened 2026-07-02
feat(kura): Kura vs legacy CAS A/B benchmark (in-network runner)

fortmarek:
## What

A manually-dispatched benchmark comparing the **Kura cache** (bare-metal, local-NVMe content-addressed store) against the **legacy Kamal cache** (S3 write-through) over the real customer HTTP plane, run from a **Tuist Linux runner**.

- `kura/benchmarks/cas_ab.sh` — the benchmark.
- `.github/workflows/kura-cache-benchmark.yml` — `workflow_dispatch` job on `runs-on: tuist-linux`.

## Why run it on a runner

A laptop over the public internet can't fairly benchmark these two backends: they sit in different datacenters at different distances from the client, and a single-stream transfer never leaves the per-connection latency regime (a local run showed single-stream downloads ~10–17 MB/s while 10 parallel streams already hit 250–375 MB/s — the client link/latency was the ceiling, not the backend). An **in-cluster runner has low RTT and a fat pipe**, so the numbers reflect the backends rather than the client.

## What it measures

Over the shared `/api/cache/cas/{id}` contract both backends implement (`POST` stores by opaque id with **no server-side hash check**, `GET` reads back; fresh random bytes per `PUT` avoid dedup so writes are real):

- **Hit latency** — small-object `GET` p50/p90/p99, warm (reused connection) and cold (fresh TLS per request).
- **Single-stream throughput** — `POST`/`GET` MB/s medians across blob sizes (≤ 24 MB; the CAS cap is 25 MB).
- **Aggregate throughput** — many parallel streams, to find the ceiling a single latency-bound stream hides.

Endpoints + auth come from `tuist cache config` (the `kura` client feature flag selects the Kura endpoint, its absence selects legacy), so nothing is hardcoded and the token is masked in CI. Results are written to the workflow **job summary**.

## Early local signal (to be confirmed on the runner)

Upload was ~3× faster on Kura (local NVMe vs legacy's synchronous S3 write-through); downloads were comparable but client-link-bound; small-object latency was a touch lower on legacy, consistent with the Hetzner-vs-Scaleway distance from the client rather than a serving difference. The runner run removes that client confound.

## Safety

It writes content-addressed junk blobs into the resolved account's **production** cache on both backends (modest volume, reclaimed by normal retention), so it's an operator-run benchmark, not part of the automated suite.

The branch-scoped `push` trigger is **temporary** — it exists only to kick the first run from the feature branch (`workflow_dispatch` is available once the file is on the default branch) and is removed before merge.

### DIFF
--- a/.github/workflows/kura-cache-benchmark.yml
+++ b/.github/workflows/kura-cache-benchmark.yml
@@ -1,0 +1,63 @@
+name: Kura Cache Benchmark
+
+# Manually-dispatched CAS A/B benchmark: Kura (bare-metal, local-NVMe) vs the
+# legacy Kamal cache (S3 write-through), run from a Tuist Linux runner so the
+# client sits inside our network (low RTT, fat pipe) — the numbers a laptop over
+# the public internet can't produce. See kura/benchmarks/cas_ab.sh.
+
+on:
+  workflow_dispatch:
+    inputs:
+      sizes_mb:
+        description: "Throughput blob sizes in MB, space-separated (each <= 24; CAS cap is 25MB)"
+        default: "1 8 24"
+      iters:
+        description: "Iterations per size for the throughput medians"
+        default: "6"
+      parallel:
+        description: "Concurrent streams for the aggregate-throughput probe"
+        default: "16"
+      latency_kb:
+        description: "Object size (KB) for the latency sweep"
+        default: "256"
+  # TEMPORARY: kicks the first run from the feature branch, since workflow_dispatch
+  # only becomes available once this file lands on the default branch. Removed
+  # before merge — workflow_dispatch is the intended trigger.
+  push:
+    branches: [claude/kura-cas-benchmark]
+    paths:
+      - "kura/benchmarks/**"
+      - ".github/workflows/kura-cache-benchmark.yml"
+
+permissions:
+  contents: read
+
+env:
+  MISE_VERSION: "2026.5.15"
+  MISE_GITHUB_TOKEN: ${{ secrets.TUIST_APP_GITHUB_TOKEN || github.token }}
+  MISE_GITHUB_ATTESTATIONS: 0
+
+concurrency:
+  group: kura-cache-benchmark-${{ github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  benchmark:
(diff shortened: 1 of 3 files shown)


fortmarek:
## First runner result — [run 28601722851](https://github.com/tuist/tuist/actions/runs/28601722851) (ubuntu-latest → eu-central, both backends healthy)

| metric | Kura | Legacy |
|---|---:|---:|
| Hit latency warm p50 (256KB) | **93 ms** | 98 ms |
| Hit latency cold p50 | 295 ms | 304 ms |
| Upload 8 / 24 MB, single-stream | **6.0 / 11.5 MB/s** | 0.6 / 0.6 MB/s |
| Download 8 / 24 MB, single-stream | 5.4 / 12.0 MB/s | 4.9 / 11.7 MB/s |
| Aggregate upload (16×8 MB) | **28.1 MB/s** | 9.7 MB/s |
| Aggregate download (16×8 MB) | **76.1 MB/s** | 64.1 MB/s |

**Takeaways**
- **Latency parity** (Kura slightly ahead: 93 vs 98 ms warm). An earlier laptop run had suggested Kura was *slower* on small objects — that was purely network distance (the laptop is closer to the Hetzner legacy box than the Scaleway Kura box), not Kura's serving. From an equidistant vantage point the difference disappears.
- **Writes are Kura's big win** — 10–20× faster single-stream upload and ~3× parallel, because Kura writes to local NVMe while the legacy cache does a synchronous S3 write-through.
- **Reads are comparable-to-better** (76 vs 64 MB/s aggregate).

**Caveats**
- `ubuntu-latest` is ~90 ms from the EU box, so absolute throughput is latency-bound and understates the box (a local eu client saw 250–375 MB/s aggregate). The **A/B between backends from the same runner** is the meaningful signal; for absolute numbers run from a host near the region.
- In-cluster `tuist-linux` runners can't reach the box's public IP (cluster egress to it is blocked), so this runs on `ubuntu-latest`.
- One earlier run hit `503` on Kura — a transient primary-pod roll, healthy before and after.

fortmarek:
[kura-cache-benchmark-report.pdf](https://github.com/user-attachments/files/29605670/kura-cache-benchmark-report.pdf)

PR #11628, opened 2026-07-02
feat(cli): update Muted/Skipped Test Case Listing to Include All Pages

curtisyying:
Resolves issue of not being able to fetch all quarantined tests when the number of either muted or skipped tests exceeds 500 (1 page at pagesize 500). Allows running of `tuist test` without needing to pass `--skip-quarantine` for cases where there are many quarantined tests.

### How to test locally
`swift test --filter TestCaseListServiceTests --disable-sandbox`


fortmarek (approved):
Thanks!

PR #11643, opened 2026-07-03
feat(cli): add generated script file list paths

fortmarek:
## What changed

Adds `TargetScript.FileListPath.generated(_:)` so target scripts can ask Tuist to create an empty `.xcfilelist` placeholder during project generation while still passing the file-list path to Xcode as a normal script input or output file list.

The manifest-facing API is an enum with associated values:

```swift
inputFileListPaths: [
    .generated("SourceryInputs.xcfilelist"),
]
```

String literals and existing file-list path helpers remain available for normal paths, and the previous `[Path]` overloads are kept as deprecated disfavored overloads to preserve source compatibility for callers that already pass `[Path]` variables.

## Why

Xcode requires `.xcfilelist` files to exist before a build starts, even when an earlier script phase is responsible for refreshing their contents. This makes generated file-list workflows awkward: projects either have to commit placeholder files that are immediately rewritten, or wrap `tuist generate` in extra setup logic.

The new API lets Tuist own the placeholder creation step while keeping the actual file-list contents under the build script's control.

## Approach

The generated marker is consumed while mapping `ProjectDescription` into `XcodeGraph`. In the graph, file lists are represented as an enum:

```swift
TargetScript.FileListPath.path(String)
TargetScript.FileListPath.generated(AbsolutePath)
```

That intentionally follows the existing generated-sources model: generated source files are flattened into `Target.sources` as regular `SourceFile` values whose path is already the generated file location. Similarly, generated script file lists are regular script file-list entries whose generated case carries the file Tuist must create, rather than storing an optional placeholder path alongside every entry or maintaining a separate `generatedFileListPaths` side-channel.

During project mapping, `GenerateTargetScriptFileListProjectMapper` scans the normal script file-list collections for `.generated` entries, de-duplicates them, creates parent directories, and writes empty placeholders. Xcode project generation derives source-root-relative PBX file-list strings from generated absolute paths, while normal `.path(String)` entries continue to pass through as before. Script hashing and PBX import mapping continue to consume the Xcode-facing path string.

Generated input file lists are hashed by their stable source-root-relative path string rather than by placeholder file contents, so graph hashes do not change between a clean checkout, `tuist generate`, and later builds that refresh the file-list contents.

Generated file-list placeholders are constrained in the manifest mapper before path resolution: they must use `.relativeToManifest`, must not start with `/`, and must not contain `..` path segments. That keeps placeholder creation inside the manifest-owned path space while normal file-list paths keep their existing resolution behavior.

## Impact

Projects can now declare generated input or output file lists without committing placeholder `.xcfilelist` files. Existing callers that pass string literals continue to work. Existing callers that pass `[Path]` variables continue to compile through deprecated overloads and can migrate to `[TargetScript.FileListPath]` over time.

## Validation

- `git diff --check`
- `xcodebuild -quiet build -workspace /Users/marekfort/.codex/worktrees/879d/tuist/Tuist.xcworkspace -scheme tuist CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `tuist generate tuist TuistGenerator TuistGeneratorTests TuistLoader TuistLoaderTests XcodeGraph XcodeGraphTests XcodeGraphMapper XcodeGraphMapperTests TuistHasher TuistHasherTests ProjectDescription --no-open`
- `xcodebuild -quiet test -workspace /Users/marekfort/.codex/worktrees/879d/tuist/Tuist.xcworkspace -scheme Tuist-Workspace -only-testing:TuistLoaderTests/TargetScriptManifestMapperTests -only-testing:TuistGeneratorTests/GenerateTargetScriptFileListProjectMapperTests -only-testing:TuistGeneratorTests/BuildPhaseGeneratorTests/generateTarget_actions -only-testing:XcodeGraphTests/TargetScriptTests -only-testing:XcodeGraphMapperTests/PBXScriptsBuildPhaseMapperTests -only-testing:TuistHasherTests/TargetScriptsContentHasherTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `xcodebuild -quiet test -workspace /Users/marekfort/.codex/worktrees/879d/tuist/Tuist.xcworkspace -scheme Tuist-Workspace -only-testing:TuistGeneratorTests/GenerateTargetScriptFileListProjectMapperTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `xcodebuild -quiet test -workspace /Users/marekfort/.codex/worktrees/879d/tuist/Tuist.xcworkspace -scheme Tuist-Workspace -only-testing:XcodeGraphTests/TargetScriptTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `xcodebuild -quiet test -workspace /Users/marekfort/.codex/worktrees/879d/tuist/Tuist.xcworkspace -scheme Tuist-Workspace -destination 'platform=macOS,arch=arm64' -parallel-testing-enabled NO -only-testing:TuistHasherTests/TargetScriptsContentHasherTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`

### DIFF
--- a/cli/Sources/TuistHasher/TargetScriptsContentHasher.swift
+++ b/cli/Sources/TuistHasher/TargetScriptsContentHasher.swift
@@ -33,8 +33,24 @@ public struct TargetScriptsContentHasher: TargetScriptsContentHashing {
             var pathsToHash: [AbsolutePath] = []
             script.path.map { pathsToHash.append($0) }
 
+            let inputFileListPaths = script.inputFileListPaths.compactMap { fileListPath -> String? in
+                switch fileListPath {
+                case let .path(path):
+                    return path
+                case .generated:
+                    return nil
+                }
+            }
+            let generatedInputFileListPathStrings = script.inputFileListPaths.compactMap { fileListPath -> String? in
+                switch fileListPath {
+                case .path:
+                    return nil
+                case let .generated(path):
+                    return path.relative(to: sourceRootPath).pathString
+                }
+            }
             var dynamicPaths = resolvePathStrings(
-                script.inputPaths + script.inputFileListPaths.map(\.path),
+                script.inputPaths + inputFileListPaths,
                 sourceRootPath: sourceRootPath
             )
             .sorted()
@@ -52,13 +68,14 @@ public struct TargetScriptsContentHasher: TargetScriptsContentHashing {
                     pathsToHash.append(path)
                 }
             }
-            stringsToHash.append(contentsOf: try await pathsToHash.concurrentMap {
+            let inputPathHashes = try await pathsToHash.concurrentMap {
                 do {
                     return try await contentHasher.hash(path: $0)
                 } catch ContentHashingError.fileHashingFailed {
                     return $0.relative(to: sourceRootPath).pathString
                 }
-            }.sorted())
+            }
+            stringsToHash.append(contentsOf: (generatedInputFileListPathStrings + inputPathHashes).sorted())
             stringsToHash.append(
                 contentsOf: resolvePathStrings(
                     script.outputPaths + script.outputFileListPaths.map(\.path),
--- a/cli/Sources/TuistHasher/TargetScriptsContentHasher.swift
+++ b/cli/Sources/TuistHasher/TargetScriptsContentHasher.swift
@@ -33,8 +33,11 @@ public struct TargetScriptsContentHasher: TargetScriptsContentHashing {
             var pathsToHash: [AbsolutePath] = []
             script.path.map { pathsToHash.append($0) }
 
-            var dynamicPaths = resolvePathStrings(script.inputPaths + script.inputFileListPaths, sourceRootPath: sourceRootPath)
-                .sorted()
+            var dynamicPaths = resolvePathStrings(
+                script.inputPaths + script.inputFileListPaths.map(\.path),
+                sourceRootPath: sourceRootPath
+            )
+            .sorted()
             if let dependencyFile = script.dependencyFile {
                 dynamicPaths += [dependencyFile]
             }
@@ -57,8 +60,11 @@ public struct TargetScriptsContentHasher: TargetScriptsContentHashing {
                 }
             }.sorted())
             stringsToHash.append(
-                contentsOf: resolvePathStrings(script.outputPaths + script.outputFileListPaths, sourceRootPath: sourceRootPath)
-                    .map { $0.relative(to: sourceRootPath).pathString }
+                contentsOf: resolvePathStrings(
+                    script.outputPaths + script.outputFileListPaths.map(\.path),
+                    sourceRootPath: sourceRootPath
+                )
+                .map { $0.relative(to: sourceRootPath).pathString }
             )
 
             if let embeddedScript = script.embeddedScript {
--- a/cli/Tests/ProjectDescriptionTests/TargetScriptTests.swift
+++ b/cli/Tests/ProjectDescriptionTests/TargetScriptTests.swift
@@ -1,13 +1,38 @@
 import Foundation
 import TuistSupportTesting
 import XCTest
-
 @testable import ProjectDescription
 
 final class TargetScriptTests: XCTestCase {
     func test_toJSON_whenPath() {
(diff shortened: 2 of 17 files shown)


gmazzo (approved):
LGTM

cli/Sources/TuistHasher/TargetScriptsContentHasher.swift
```diff
@@ -33,8 +33,11 @@ public struct TargetScriptsContentHasher: TargetScriptsContentHashing {
             var pathsToHash: [AbsolutePath] = []
             script.path.map { pathsToHash.append($0) }
 
-            var dynamicPaths = resolvePathStrings(script.inputPaths + script.inputFileListPaths, sourceRootPath: sourceRootPath)
-                .sorted()
+            var dynamicPaths = resolvePathStrings(
+                script.inputPaths + script.inputFileListPaths.map(\.path),
```
pepicrft: Generated file lists are created later as project side effects, so on a clean checkout this path is missing and falls through to the relative path string below. After generation, the placeholder exists and the same target can hash the empty file contents instead. Could we treat `.generated` input file lists as stable path strings for hashing, or create these placeholders before graph content hashes are computed?

cli/Tests/ProjectDescriptionTests/TargetScriptTests.swift
```diff
@@ -10,6 +9,32 @@ final class TargetScriptTests: XCTestCase {
         XCTAssertCodable(subject)
     }
 
+    func test_toJSON_whenGeneratedInputFileListPath() {
```
pepicrft, line 12: The local Swift review guidance asks touched tests to move to Swift Testing when they are edited. Could this suite be converted to `import Testing`, `@Test`, `#expect`, and `#require` while adding these cases?

pepicrft (approved):
Approved. Left two inline comments.

cli/Tests/ProjectDescriptionTests/TargetScriptTests.swift
fortmarek: > The local Swift review guidance asks touched tests to move to Swift Testing when they are edited. Could this suite be converted to import Testing, @Test, #expect, and #require while adding these cases?

Let's instead spin up an agent to convert these in one go.

cli/Sources/TuistHasher/TargetScriptsContentHasher.swift
fortmarek: Good catch, agreed. I fixed this in `14efab0fb77` by treating `.generated` input file lists as stable source-root-relative path strings in the target script content hash. Normal `.path(...)` file lists keep the existing content-hash behavior.

I went with path-string hashing instead of creating placeholders before graph hashing because the placeholder contents can legitimately change later when the build phase refreshes the file list. That keeps the graph hash stable across a clean checkout, after `tuist generate`, and after later builds.

I also added a regression test in `TargetScriptsContentHasherTests` that stubs the generated file-list content hash and verifies it is not requested. Validated with:

`xcodebuild -quiet test -workspace /Users/marekfort/.codex/worktrees/879d/tuist/Tuist.xcworkspace -scheme Tuist-Workspace -destination 'platform=macOS,arch=arm64' -parallel-testing-enabled NO -only-testing:TuistHasherTests/TargetScriptsContentHasherTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`

PR #11645, opened 2026-07-03
feat(noora): add button transform

asmitbm:
Adds a subtle press feedback effect to .noora-button and .noora-neutral-button

### DIFF
--- a/noora/css/button.css
+++ b/noora/css/button.css
@@ -5,7 +5,7 @@
   justify-content: center;
   align-items: center;
   box-sizing: border-box;
-  -webkit-appearance: button;
+  --webkit-appearance: button;
   border: 0;
   border-radius: var(--noora-radius-3);
   padding: 0;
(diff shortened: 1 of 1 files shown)


noora/css/button.css
```diff
@@ -5,7 +5,7 @@
   justify-content: center;
   align-items: center;
   box-sizing: border-box;
-  -webkit-appearance: button;
+  --webkit-appearance: button;
```
fortmarek: **[P2] Restore the WebKit appearance declaration**

This now declares a CSS custom property instead of applying the vendor-prefixed appearance rule. Because `--webkit-appearance` is ignored as a property declaration, the shared button reset no longer applies to `.noora-button`, `.noora-link-button`, or `.noora-neutral-button`, which can reintroduce native button styling differences in WebKit browsers. This should stay `-webkit-appearance: button;`.

fortmarek (approved):

PR #11652, opened 2026-07-03
feat(infra): adopt Flux GitOps + Grafana Cloud health monitoring for the mgmt cluster

fortmarek:
Implements [hive/specs/72](https://hive.tuist.dev/specs/72).

> The hive / once / atlas / tuist workload `Cluster` CRs were applied by hand (or push-on-merge) with the break-glass management kubeconfig, and nothing reconciled or watched them. While downsizing the clusters, three failures surfaced: **manifest drift** (a declared worker pool never applied; orphaned VMs no `Cluster` tracked), **degraded control planes** unnoticed for weeks (`EtcdClusterHealthy=False`), and **untracked infrastructure**. Root cause: routine changes flow through a person running `kubectl apply` with the root-of-trust kubeconfig.

The spec has two pillars. **Pillar 2 (Grafana Cloud health alerting) and the per-topology control-plane MachineHealthCheck already landed on `main` in #12081** (management-cluster monitoring via `values-management.yaml` + `alerts.md`, installed by `mgmt-cluster-apply.yml`; the shared ClusterClass now remediates the control plane at an absolute `1` with `nodeStartupTimeoutSeconds: 0`). **This PR is the other half — Pillar 1 (Flux reconciliation) plus the paired reconciliation checks** — and complements what #12081 shipped rather than duplicating it.

### Pillar 1 — Flux reconciliation
- Moved the reconciled Cluster CRs to `infra/k8s/clusters/workloads/<cluster>/` (one subdir each): `staging` / `canary` / `production` (git-renamed from the flat `cluster-*.yaml`) plus the migrated **hive / once / atlas** tenants (copied in from `tuist/{hive,once,atlas}`, all reuse `org-tuist` + the `hetzner` Secret).
- `infra/flux/mgmt/cluster-*.yaml` — one Flux `Kustomization` per cluster: `prune: false` (a git deletion never tears down a live cluster), `force: false` (halts on an SSA conflict with CAPI's topology controller rather than stomping it), `healthCheckExprs` gating on the `Cluster` readiness condition.
- ESO on the mgmt cluster (`infra/k8s/mgmt/external-secrets.yaml` + `flux-git-externalsecret.yaml`) as the Flux bootstrap prerequisite (Decision 8): the `onepassword` `ClusterSecretStore` for the `tuist-k8s-mgmt` vault + Flux's git (GitHub App) credential.
- `infra/flux/mgmt/flux-system/` — Flux's own manifests (`gotk-components` / `gotk-sync` / `kustomization`), generated with `flux install --export` at **2.9.5** using bootstrap's default component set. Committed here rather than left to `flux bootstrap`, which would otherwise supply this half by pushing straight to `main` outside review; bootstrap is idempotent and adopts them. Flux is pinned to 2.9.5 in CI too (`fluxcd/flux2/action`), so CI and cluster agree — `flux check --pre` flagged the previous 2.4.0 as five minors stale.
- `infra/k8s/mgmt/flux-diff-rbac.yaml` (least-privilege SA; dry-run write on `clusters` only, read on the related v1beta2/v1beta1 CAPI kinds) + `.github/workflows/flux-diff.yml` posting the server-side dry-run on `workloads/**` PRs, same-repo branches only, authenticating from a **CI-only 1Password vault** (see below). The job is a green no-op when its credential is absent, so it never blocks PRs.

`mgmt-cluster-apply.yml` is unchanged by this PR: moving the Cluster CRs out of the flat `cluster-*.yaml` glob already stops it applying them (it keeps ClusterClass + bare-metal templates, preview, pentest, and the mgmt-side workloads — none of which Flux owns).

### Reconciliation checks (both directions)
- `infra/k8s/mgmt/reconciliation-checks.yaml` — a CronJob (+ scrape-target Pushgateway) reporting **orphan Hetzner servers** (no owning CAPI `Machine`) and **stale Clusters** (live but removed from git — the never-prune blind spot) as metrics.
- Their alert rules are folded into `infra/helm/k8s-monitoring/alerts.md` (main's canonical alert doc) alongside main's control-plane/etcd/egress alerts.

### Docs
`infra/AGENTS.md`, `infra/k8s/clusters/README.md`, `infra/k8s/onboarding.md`, and a new `infra/flux/mgmt/README.md` document the Flux-vs-`mgmt-cluster-apply.yml` ownership split, the never-prune destroy flow, and the break-glass recovery path.

## Why these choices

- **Flux over Argo CD** (spec's comparison): lighter/stateless on a just-downsized cluster, less exposed to the owner-reference pruning footgun, scoped never to prune `Cluster` objects.
- **Reconciliation and health are separate pillars** — the "degraded for weeks" failure was a monitoring gap (Pillar 2, already on `main`), not something a GitOps dashboard fixes.
- **Footguns handled explicitly:** preview + pentest carved out of Flux; autoscaled `md-processor` `replicas` omitted so Flux doesn't fight cluster-autoscaler; caph infra CRDs are v1beta1 (not v1beta2) in the flux-diff RBAC; per-cluster `Kustomization`s never prune `Cluster` objects.

## Security fix this PR carries (spec Decision 3 gap)

Decision 3 pinned down the `flux-diff` job's ServiceAccount, verbs, network path and fork restrictions — but not its **1Password scope**, and that omission made the rest decorative.

The job was specified to authenticate with a token scoped to `tuist-k8s-mgmt`. That vault also holds `kubeconfig: tuist-mgmt` (break-glass **cluster-admin** for every workload cluster), the Hetzner Robot credentials, the bare-metal SSH key and the Flux App key. Because the job triggers on `pull_request` — and GitHub runs the workflow file *from the PR branch* — anyone able to push a branch could have edited `flux-diff.yml` to `op document get "kubeconfig: tuist-mgmt"` and obtained cluster-admin. The narrow RBAC is worthless if the same job can fetch the admin credential. (`mgmt-cluster-apply.yml` safely holds the wider scope only because it triggers on push to `main`, where no PR supplies workflow content.)

Fixed by moving the flux-diff kubeconfig into a dedicated **`tuist-k8s-mgmt-ci`** vault holding nothing else, and scoping the CI service account to it. Verified: that token's `op vault list` returns only `tuist-k8s-mgmt-ci`. Both files document why the scoping is load-bearing, and why the environment deliberately has no protection rules (required reviewers would defeat automatic per-PR diffing; a branch policy can't match arbitrary PR branch names).

## Verification

Static:
- `kubectl kustomize infra/k8s/clusters/workloads/<cluster>` builds for all six clusters; manifests parse; docs prettier-clean.

Against the **live management cluster**:
- `flux check --pre` — `Kubernetes 1.36.0 >= 1.28.0-0`, prerequisites passed.
- ESO installed; `clustersecretstore/onepassword` reports `Valid`.
- All three `FLUX_GITHUB_APP` references resolve, proven end to end by a throwaway `ExternalSecret` that synced the PEM (then removed).
- `flux-diff` RBAC applied and asserted: **can** dry-run apply all six workload Clusters; **cannot** delete Clusters, write ClusterClass/MachineDeployments/KCP, read Secrets, reach another namespace, or escalate.
- The pre-enable gate, below.

## Roll-out status

Everything executable before merge is done. `infra/flux/mgmt/README.md` carries the full runbook.

| Step | State |
|---|---|
| ESO on the mgmt cluster + `onepassword` ClusterSecretStore | ✅ done |
| Flux GitHub App + `FLUX_GITHUB_APP` in 1Password | ✅ done |
| `flux-diff` SA/RBAC, kubeconfig, CI vault, Environment secret | ✅ done |
| Pre-enable no-op gate | ✅ done |
| **`flux bootstrap`** | ⏸ **after this merges** |
| `reconciliation-checks.yaml` + its Grafana rules | ⏸ after bootstrap |

**Bootstrap is deliberately post-merge.** Flux reconciles `--branch=main`, so the per-cluster Kustomizations and `workloads/` CRs only exist for it once this lands; bootstrapping earlier installs a reconciler with nothing to reconcile. After merge, run the pinned `flux bootstrap` from the README — it adopts the committed `flux-system/`, creates the six Kustomizations, and reconciles.

### Follow-up PRs (prepared, deliberately not merged)

The tenant repos still carry their own `Cluster` CR. Retiring them is part of the spec's migration, but must happen **after** this lands and Flux is verified reconciling — otherwise we delete the only remaining record if this stalls. Both are open as drafts:

- tuist/once#268 — retire `infra/k8s/cluster-production.yaml`
- tuist/atlas#450 — same, plus removes an `onboarding.md` step that hand-applies the CR (actively harmful once Flux owns it)

`tuist/hive` needs nothing — it no longer has a Cluster CR.

Each is replaced by a pointer rather than deleted, so the path still answers "where is my cluster defined?". Worth noting *why* this matters: those copies have never been applied to the management cluster and have silently drifted from live — neither declares `disableControlPlaneKubeletLocalMode` or `statefulRaidLevel`, which the live clusters carry. Leaving them would be a second source of truth that looks authoritative and does nothing.

The `reconciliation-checks` CronJob waits for bootstrap on purpose: its stale-cluster half derives "what's in git" from the live Flux `Kustomization` objects, so enabling it earlier would flag every cluster as stale.

Expect the flux-diff check to report *"Kustomization not found"* until bootstrap — those objects don't exist yet. It won't block the PR.

**Pre-enable gate — RUN, and it found real drift.** `kubectl diff` of each
`workloads/<cluster>` against the live mgmt cluster initially showed **four of six
clusters would not reconcile as a no-op**. The tenant manifests had been verified against
the `tuist/{hive,once,atlas}` repos, but those have themselves drifted from live because
nothing ever applied them to the management cluster. Fixed in `8518cb46` by syncing the
tenant CRs to live (declaring the full 21-variable set CAPI materializes, and dropping an
`md-0` pool label atlas had but live did not).

Current gate status: **staging / canary / hive / once / atlas are clean no-ops.**
`production` retains one intentional diff — `md-processor` losing `replicas: 2`, the
designed handoff of that pool to cluster-autoscaler. That is the one change Flux will make
on its first reconcile; confirm the live MachineDeployment is not rescaled when the field
disappears.

Retiring the `tuist/{hive,once,atlas}` `cluster-production.yaml` copies to a pointer is coordinated follow-up in those repos (no double-apply risk today — nothing applies them).


infra/flux/mgmt/flux-system/gotk-sync.yaml
```diff
@@ -0,0 +1,30 @@
...
+kind: GitRepository
+metadata:
+  name: flux-system
+  namespace: flux-system
+spec:
+  interval: 1m0s
+  # Without `provider: github`, source-controller treats the secret as generic
+  # basic-auth and silently ignores the GitHub App fields ESO syncs into it —
+  # rotating the App would then not rotate what Flux actually uses.
+  provider: github
```
pepicrft, line 12: `flux bootstrap github --token-auth` on 2.9.5 rewrites this file from CLI flags every time it runs, and there's no `--provider` flag on that version to keep `spec.provider: github`. So the moment someone re-bootstraps (or the docs say to), this hand-added line gets stripped. Once the short-lived bootstrap PAT expires, source-controller falls back to reading the `flux-system` Secret as basic auth, ignores the ESO-synced GitHub App fields, and reconciliation dies silently. Worth adding a kustomize patch that re-asserts `provider: github` after bootstrap so re-running it can't regress it.

infra/k8s/mgmt/reconciliation-checks.yaml
```diff
@@ -0,0 +1,223 @@
...
+                      key: hcloud
+                - name: PUSHGW
+                  value: "http://reconciliation-pushgateway.observability.svc.cluster.local:9091"
+              command:
+                - sh
+                - -c
+                - |
+                  set -euo pipefail
+
+                  # (a) Orphan Hetzner Cloud servers: every server id in the
```
pepicrft: The mgmt cluster's own `tuist-mgmt` VM is a Hetzner Cloud server with no owning CAPI Machine (mgmt doesn't self-manage), so `/v1/servers` returns it every time and `owned` never contains its id. `orphans >= 1` on every run and the `> 0` alert (pending 15m) fires forever within an hour of enabling this. Either exclude by hcloud server id, or by a label we set on the mgmt server itself.

.github/workflows/flux-diff.yml
```diff
@@ -0,0 +1,179 @@
...
+          set -uo pipefail
+          base="${{ github.event.pull_request.base.sha }}"
+          # Cluster subdirs with changes in this PR.
+          changed=$(git diff --name-only "$base"...HEAD -- infra/k8s/clusters/workloads \
+            | sed -n 's#^infra/k8s/clusters/workloads/\([^/]*\)/.*#\1#p' | sort -u)
+          out="$(mktemp)"
+          failed="$(mktemp)"
+          if [ -z "$changed" ]; then
+            echo "_No workload Cluster manifests changed._" > "$out"
+          else
```
pepicrft, line 121: The workflow triggers on changes to `infra/flux/mgmt/cluster-*.yaml`, but this diff step only inspects `infra/k8s/clusters/workloads`. A PR that only edits a Flux Kustomization CR (say, flipping `force: true` or dropping the health gate) posts `_No workload Cluster manifests changed._` and reviewers approve believing nothing meaningful shipped, while the reconciliation behavior actually changed.

.github/workflows/flux-diff.yml
```diff
@@ -0,0 +1,179 @@
...
+              step_out="$(mktemp)"
+              flux diff kustomization "cluster-$c" --path="./$dir" >"$step_out" 2>&1 || true
+              cat "$step_out" >> "$out"
+              if grep -qiE 'forbidden|is not authorized|no matches for kind|failed to|unable to|connection refused|unauthorized' "$step_out"; then
+                echo "$c" >> "$failed"
+              fi
+              echo '```' >> "$out"
+              echo >> "$out"
+            done
+          fi
```
pepicrft, line 146: The failure-signature grep misses the common `flux diff` error paths I'd expect to see in practice: `kustomization not found`, `context deadline exceeded`, `the server could not find the requested resource`. `|| true` swallows the exit code, so a genuine failure renders as a clean-looking code fence under the cluster heading with no `::error::` banner, and a reviewer skimming multiple diffs will merge on what looks like an empty diff.

infra/k8s/mgmt/reconciliation-checks.yaml
```diff
@@ -0,0 +1,223 @@
...
+                    resp=$(curl -sf -H "Authorization: Bearer $HCLOUD_TOKEN" \
+                      "https://api.hetzner.cloud/v1/servers?per_page=50&page=$page")
+                    server_ids="$server_ids $(echo "$resp" | jq -r '.servers[].id')"
+                    next=$(echo "$resp" | jq -r '.meta.pagination.next_page // empty')
+                    [ -z "$next" ] && break
+                    page=$next
+                  done
+                  owned=$(kubectl get machines.cluster.x-k8s.io -A -o json \
+                    | jq -r '.items[].spec.providerID // empty' \
+                    | sed -n 's#^hcloud://##p')
```
pepicrft, line 207: The exclusion list here is hardcoded to `tuist-preview|tuist-pentest`, but `mgmt-cluster-apply.yml`'s glob already picks up any new `cluster-*.yaml` without workflow edits. The moment we add another non-Flux Cluster (a new preview variant, a shared cluster), it flags as stale permanently, on-call learns to ignore the alert, and real stale events get missed. The comment says `keep this in step` but nothing enforces it.

pepicrft (commented):
Deep review of the Flux GitOps + Grafana Cloud monitoring rollout. Ten findings across correctness, operability, and defense-in-depth. Ranked most severe first in the inline comments below.

infra/k8s/mgmt/reconciliation-checks.yaml
```diff
@@ -0,0 +1,223 @@
...
+                  orphans=0
+                  for id in $server_ids; do
+                    echo "$owned" | grep -qx "$id" || orphans=$((orphans + 1))
+                  done
+
```
pepicrft, line 224: The alert only exports a scalar count, so on-call sees `stale_clusters=2` but has to break-glass into the mgmt cluster to find out which two, which defeats the point of pushing to Grafana Cloud. Emitting one series per resource (e.g. `capi_reconciliation_stale_cluster{name="..."} 1`) lets the alert annotation name them directly.

infra/k8s/mgmt/reconciliation-checks.yaml
```diff
@@ -0,0 +1,223 @@
...
+        prometheus.io/port: "9091"
+    spec:
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
+      containers:
+        - name: pushgateway
+          image: prom/pushgateway:v1.10.0
+          args:
```
pepicrft: Persistence is on an `emptyDir`, so any pod restart wipes the store. Combined with the `absent_over_time(...[1h])` threshold on `capi_reconciliation_last_success_timestamp_seconds`, a Pushgateway bounce plus one missed CronJob run will page on healthy reconciliation. Either back this with a PVC or drop persistence and tolerate the first empty scrape after a bounce in the alert.

infra/k8s/mgmt/reconciliation-checks.yaml
```diff
@@ -0,0 +1,223 @@
...
+          securityContext:
+            runAsNonRoot: true
+            runAsUser: 65532
+            seccompProfile:
+              type: RuntimeDefault
+          containers:
+            - name: check
+              # kubectl + jq + curl.
+              image: alpine/k8s:1.31.2
+              securityContext:
```
pepicrft, line 167: `alpine/k8s:1.31.2` ships kubectl 1.31 but the mgmt cluster is on 1.36 per the flux README, which is outside the supported N±1 skew. Today the CAPI kinds we `get` here happen to still parse fine, but if a newer CAPI kind changes shape, the jq pipelines below silently emit nothing and both checks report `orphans=0` / `stale=0`, which is exactly the failure mode we're trying to catch. Worth pinning to a kubectl within N±1 of the mgmt cluster.

infra/k8s/mgmt/flux-diff-rbac.yaml
```diff
@@ -0,0 +1,180 @@
...
+# pushable branch read the break-glass cluster-admin kubeconfig sitting in that
+# vault and bypass every restriction above. Scoping the CI service account to a
+# vault containing a single least-privilege credential is what makes the RBAC
+# here meaningful. Never exposed to fork PRs. Owned + rotated by the mgmt owner
+# (Decision 8).
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: flux-diff
```
pepicrft, line 32: The Role grants `create/update/patch` on `clusters` here, and the paired VAP `matchConstraints.resourceRules` only lists `resources: ["clusters"]`. If we later widen this Role to a subresource (e.g. `clusters/status` to make `flux diff` behave better against a status-populated cluster), non-dry-run writes from the flux-diff identity would slip past the admission guard. Adding `subresources: ["*"]` to the VAP would future-proof it.

infra/k8s/mgmt/flux-git-externalsecret.yaml
```diff
@@ -0,0 +1,47 @@
...
+  refreshInterval: 1h
+  secretStoreRef:
+    kind: ClusterSecretStore
+    name: onepassword
+  target:
+    name: flux-system
+    creationPolicy: Merge
+    deletionPolicy: Retain
+    template:
+      data:
```
pepicrft, line 36: `creationPolicy: Merge` needs the target `flux-system/flux-system` Secret to already exist. On a break-glass reseed (someone deletes the Secret and re-applies from git without re-bootstrapping), ESO sits in `SecretSyncedError`, GitRepository fails auth, and Flux stops reconciling with no obvious signal unless you're watching ES status. Either scope this to `Owner` on just the App fields, or document the reseed order in the README so the recovery path is covered.

pepicrft (approved):

PR #11655, opened 2026-07-03
fix(cli): memoize the CAS daemon's auth token to avoid a per-op keychain read

fortmarek:
## What changed

The CAS daemon resolved a fresh server auth token on every CAS operation, which meant a credentials-store read (keychain/file) and potential token refresh per compilation-cache key query. `ServerAuthenticationController` now memoizes the resolved token in memory and reuses the client across operations, refreshing only when the cached token expires.

## Why

Xcode compilation-cache builds issue thousands of CAS key queries per build. With per-op auth resolution, each query paid a credentials-store round trip, which dominated wall-clock time on cached builds: a mastodon fixture build against a warm same-host Kura cache ran for over an hour and a half against 4.203.0-canary.5 (which lacks this fix), with the build log showing key queries trickling through one at a time. The same reproduction with this branch's CLI resolves the token once.

## Root cause

`ServerAuthenticationController.authenticationToken(serverURL:)` hit the credentials store on every call, and the CAS daemon calls it on every operation. Nothing cached the result between operations within the daemon's lifetime.

## Validation

- Reproduced the per-op auth cost locally with a socket-attached CAS daemon and confirmed the memoized controller resolves credentials once per daemon lifetime (yesterday's local reproduction).
- Unit-level behavior covered by the existing ServerAuthenticationController tests; the memoization is transparent to callers.
- A staging mastodon xcode-cache benchmark against host-Kura 0.11.0 with this branch's CLI is running as the end-to-end validation.


fortmarek:
End-to-end validation results (staging runners-fleet Mac, golden VM, host Kura 0.11.0 with the 256 MiB CAS cap, clean DerivedData per timed run):

**The two fixes here took the mastodon warm cached build from 40+ minutes per build (canary.5, per-op auth) to 55-57 s, and eliminated all transport drops:**

| mastodon (iOS sim) | mean | vs no-cache (87.8 s) |
| --- | --- | --- |
| warm remote CAS | 55-58 s | 0.64 |
| floor: local CAS kept, remote detached | 46.8 s | 0.53 |
| first warm run (one-time re-upload) | 148 s | - |

| tuist CLI scheme (macOS) | mean | vs no-cache (274 s) |
| --- | --- | --- |
| warm remote CAS | 386-411 s | 1.4-1.5 (net-negative) |
| floor: local CAS kept | 120.3 s | 0.44 |

- Connection-drop validation: the cold CAS warm-up of the CLI fixture (a larger save storm than the one that produced 313 `saveCASArtifact`/`putKeyValue` failures before this fix) completed with **zero** "network connection was lost" lines, on two separate runs.
- A follow-up experiment with a Release-built daemon and `TUIST_HTTP_MAXIMUM_CONNECTIONS_PER_HOST=64` (forwarded via the new env passthrough) changed neither number, so the remaining warm-fetch overhead (~9 s on mastodon, ~4x floor on the CLI) is not client CPU or connection parallelism - it is the per-compilation-unit blocking round-trip shape of the CAS remote protocol. Projects with very high unit counts (like the CLI itself) stay net-negative on remote-CAS-alone regardless of transport, matching the runners-benchmark README's standing note; the win there is the combined module+Xcode strategy.

fortmarek:
Correction to the earlier diagnosis of the CLI-project net-negative result, after instrumenting with `COMPILATION_CACHE_ENABLE_DIAGNOSTIC_REMARKS=YES` and correlating daemon analytics + Kura request logs:

**The cache hits nearly 100%** - a warm CLI build shows 4,447 `cache key query hit` vs 2 misses. The keys are stable across runs (the daemon's `keyvalue_metadata` table does not grow across repeated builds). The net-negative wall time is NOT misses and NOT the remote protocol shape.

**The overhead is the daemon data-path cost per operation.** A fully-warm build that made zero requests to Kura (everything served from the daemon's local stores) still took 446 s vs the 120 s local-CAS floor - so ~330 s is spent between the compiler and the daemon across ~19k operations (~17 ms effective per op), independent of the network entirely. On the mastodon project (~2.2k ops) the same overhead is only ~9 s, which is why it stays clearly net-positive.

Suspects on the daemon hot path, in likely order of impact:
1. Value/artifact payloads decoded from JSON + base64 per operation (hundreds of MB per warm build through `getCacheValue` JSON entries).
2. Per-request unbounded `Task {}` spawning for analytics + CAS-node parsing contending one SQLite file at ~19k requests/build.
3. No in-memory fast path for repeat lookups; every hit goes through the on-disk value store.

This is a separate daemon-performance workstream from the fixes in this PR (which stand on their own: 40 min -> 56 s on mastodon, zero transport drops).

fortmarek:
Final root-cause analysis for the CLI-project net-negative (corrects my previous comment - the "zero-network 446s" build there ran against a stopped Kura and is invalid).

**Replay works.** Task-class timing on a fully-warm remote build vs no-cache:

| task class | warm remote | no-cache |
| --- | --- | --- |
| SwiftCompile (623) | 91 s | 535 s |
| CompileC (1047) | 28 s | 160 s |
| SwiftEmitModule (194) | 15 s | 90 s |
| SwiftDriver (194) | 162 s | 86 s |
| ScanDependencies (800) | 36 s | ~0 |
| **total task time** | **~360 s** | **~890 s** |
| **wall** | **445 s** | **260 s** |

The cache eliminates ~60% of the work, but effective parallelism collapses from 3.4x to **0.8x**: with compilation gone there is nothing to hide each target's fixed serial latency behind (SwiftDriver planning ~840 ms/target + key query + output materialization ~1 s/target of daemon round-trips), and the CLI's 194-module dependency chain serializes those latencies along the critical path. mastodon's shallow graph is why the same setup is clearly net-positive there (56 s vs 88 s no-cache, 47 s floor).

Falsified along the way (each with a measured non-result): cache misses (4,447 hits / 2 misses), key instability, daemon CPU (Release build: no change), `TUIST_HTTP_MAXIMUM_CONNECTIONS_PER_HOST=64` (no change), artifact prefetching in the daemon (landed here - it parallelizes loads within a target and reduces Kura round-trips, mastodon unregressed, but the chain latency dominates on the CLI), and `COMPILATION_CACHE_ENABLE_DETACHED_KEY_QUERIES=YES` (no change).

Practical conclusions:
1. For deep module graphs, remote-CAS-alone is structurally latency-bound on the dependency chain; the recommended configuration is the **combined module+xcode strategy**, where the binary cache flattens the graph and only the app target's units hit the CAS.
2. The follow-up workstream should trace per-target latency at the llbuild level (dynamic task wait times are invisible to -showBuildTimingSummary) before attempting further daemon work.

fortmarek:
Final attribution, closing the investigation. The artifact prefetcher in this PR was verified working (62% of CAS loads served sub-1ms, daemon load time 186s -> 54s on a warm CLI build) - **and wall time did not move** (7m18s). Together with a control build using the plugin CAS locally with no remote attached (2m40s, same dynamic-task machinery), this pins the remote-attach premium conclusively:

| configuration | wall |
| --- | --- |
| local CAS full replay | 120 s |
| plugin CAS local-only, no remote | 160 s |
| remote via daemon, ops at 17 ms mean | 445 s |
| remote via daemon, ops at ~0 ms (prefetched) | 438 s |

~280 s is spent **client-side in the toolchain** - the CAS plugin's gRPC client plus llbuild's dynamic-task choreography cost ~15-20 ms per operation across ~16k operations whenever a remote service path is configured, independent of server latency. This is not addressable from the daemon side of the protocol; it warrants an upstream swift-build/llbuild report (we have a clean reproduction: any deep-module-graph project, warm remote cache, clean DerivedData per build).

Everything in this PR stands on its own merits: memoization (40 min -> 56 s builds), transport fixes (zero drops), prefetcher (daemon op time -3.4x, fewer Kura round-trips). Product guidance for deep graphs remains the combined module+xcode strategy; shallow/normal graphs (mastodon-class) are net-positive today at 0.64x no-cache, on par with Namespace-native ratios.

fortmarek:
Closing the investigation into the deep-graph (CLI-project) warm-remote slowness with the final state of evidence.

**Validated and working**: mastodon-class projects run warm remote CAS at 56-58 s vs 47 s local floor vs 88 s no-cache with zero errors - the daemon, the transport fixes in this PR, and the on-host Kura node are all performing correctly.

**The deep-graph case (tuist CLI scheme, 194 chained modules) remains net-negative (325-440 s vs 160 s plugin-local vs 274 s no-cache), and the cause is now isolated to Xcode's CAS plugin internals.** Eight hypotheses were tested and falsified with measurements, including three via a custom-built swift-build service (engine-callback blocking, dynamic-task choreography - both surgically removed with no wall-time change) and one via a daemon-side gRPC config fix in this PR (server ping policing killing the plugin's connections - a real misconfiguration worth keeping fixed, but not the bottleneck). A mid-stall system spindump shows every participant idle - xcodebuild, the build service, the daemon, zero compiler processes - while libToolchainCASPlugin's embedded gRPC client event loops sit unscheduled for 30-50 s at a time. The build proceeds in bursts separated by these dead windows, which points at timer-driven request scheduling or an internal stall inside the closed-source plugin, out of reach from both the daemon side and the swift-build side.

Recommended follow-ups: file an Apple Feedback with this reproduction (any deep-module-graph project, warm remote cache, clean DerivedData; spindump attached), and keep the combined module+xcode strategy as the guidance for deep graphs - the binary cache flattens the dependency chain, which sidesteps the plugin's serial behavior entirely.

Everything in this PR stands independently: 40 min -> 56 s (auth memoization), zero transport drops (resource timeout + retries), reduced Kura round-trips (prefetcher), correct gRPC keepalive posture (ping policy).

PR #11658, opened 2026-07-04
fix(server): route Linux runner builds to the account's in-cluster Kura node

esnunes:
## What changed

`Tuist.Kura.runner_cache_endpoint_url/2` gains a second tier: when no **private** runner-cache node serves the fleet's platform, dispatch now hands out the account's active **public-region** Kura server addressed by its in-cluster Service DNS form (`http://kura-<account>-<cluster_id>.kura.svc.cluster.local:4000`).

- Public region specs may now declare `runner_platforms` to opt into this fallback; `eu-central` declares `[:linux]`. Undeclared public regions (us-east, us-west, ca-east) keep serving no runner fleet.
- New `Provisioner.internal_url/2` callback renders the in-cluster form from the region's `private_url_template`, which managed (public) regions now also carry. `public_url/3` behavior is unchanged (it resolves public templates first).
- The dispatch controller contract is unchanged — the claim's `cache_endpoint_url` is still gated by `Catalog.fleet_on_cluster_network?/1` and the platform, and the runner image already exports it as `TUIST_CACHE_ENDPOINT`.

## Why / root cause

Every trusted kura.yml job on the `tuist-linux` pool fails `tuist bazel setup` (masked by `continue-on-error`, so Bazel builds run cold — e.g. [this run](https://github.com/tuist/tuist/actions/runs/28684359838/job/85138411554)). The failure chain, confirmed from inside a runner pod with a debug workflow ([run 28710663346](https://github.com/tuist/tuist/actions/runs/28710663346)):

1. Prod advertises no Linux private cache region (`scw-fr-par-runners` is macOS-only, co-located with the Mac minis on the Scaleway Private Network), so dispatch stages no `TUIST_CACHE_ENDPOINT` for Linux fleets.
2. The CLI falls back to `/api/cache/endpoints`, which returns only the public `https://tuist-eu-central-1.kura.tuist.dev`.
3. That host resolves to a **cluster node's own IP** (the dedibox-fleet node hosting the Kura ingress). Cilium classifies traffic to node IPs as `remote-node` identity, which the runners' `0.0.0.0/0` egress `ipBlock` rule never matches — the connection is silently dropped, the REAPI `GetCapabilities` probe times out after 10s (`Swift.CancellationError`), and `.bazelrc.tuist` is never written.

The public ingress is therefore unusable from runner pods by construction; the fix routes them over the path that already works and is already allowed by the runners-namespace egress rule (runner pods → kura pods on 4000).

## Why this solution

- **In-runner probes** showed `http://kura-tuist-eu-central-1.kura.svc.cluster.local:4000` answering unauthenticated h2c `GetCapabilities` with 200 in ~57ms, while the public ingress and the Scaleway PN NodePort both time out from the Linux fleet. The same run's macOS job confirmed the PN NodePort path works end-to-end there (setup succeeds, ~2ms probes) — so the fix deliberately leaves macOS untouched: no public region declares `:macos`, and a stale private node still fails macOS over to the public cache (reachable from Tart VMs), never to a cluster-DNS URL the VMs can't resolve.
- Alternatives considered: loosening runner egress to reach the public ingress (reintroduces the hairpin through the public dataplane the private path exists to avoid, and widens the egress policy to cluster nodes), or provisioning a dedicated Linux private region in prod (new infrastructure for traffic the account's existing in-cluster node can already serve). Auth is identical in both tiers — the same Guardian JWT verified by the same `tuist.lua` hook.

## Impact

Linux runner jobs in production (and staging, which also exposes `eu-central` with linux cluster-networked) get a working Bazel remote cache again for accounts with an eu-central Kura server. Accounts without one keep today's behavior (no cache). macOS routing is byte-for-byte unchanged.

## Validation

- `mix test test/tuist/kura_test.exs test/tuist/kura/regions_test.exs test/tuist/kura/provisioner/kubernetes_controller_test.exs test/tuist_web/controllers/runners_controller_test.exs test/tuist/kura/runner_cache_test.exs` — 151 tests, 0 failures. New coverage: the fallback URL for `:linux`, private-node precedence, the macOS never-falls-back guarantees (both no-declaration and stale-heartbeat cases), inactive/absent server cases, and `internal_url/3` rendering.
- `mix credo` and `mix format` clean on the changed files.
- The branch also carries `.github/workflows/kura-cache-debug.yml` (Linux + macOS diagnostic jobs, credential-redacting log dumps). It's temporary tooling: after this deploys, dispatching it should show `TUIST_CACHE_ENDPOINT=http://kura-tuist-eu-central-1.kura.svc.cluster.local:4000` on the Linux job and `tuist bazel setup` succeeding; delete the workflow once verified.

### DIFF
--- a/.github/workflows/kura-cache-debug.yml
+++ b/.github/workflows/kura-cache-debug.yml
@@ -155,3 +155,99 @@ jobs:
           else
             echo ".bazelrc.tuist was not written"
           fi
+
+  # Same diagnostics from the macOS fleet: dispatch stages the Scaleway PN
+  # NodePort URL (http://<node PN address>:30815) for macOS, so this job shows
+  # what the CLI does with it after the single-listener cutover.
+  bazel-cache-debug-macos:
+    name: Bazel Remote Cache Debug (macOS)
+    runs-on: tuist-macos
+    timeout-minutes: 30
+    permissions:
+      contents: read
+      id-token: write
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: jdx/mise-action@v4.0.1
+        with:
+          version: ${{ env.MISE_VERSION }}
+          install_args: "tuist"
+          working_directory: kura
+
+      - name: Inspect runner cache environment
+        run: |
+          if [ -n "${TUIST_CACHE_ENDPOINT:-}" ]; then
+            echo "TUIST_CACHE_ENDPOINT=${TUIST_CACHE_ENDPOINT}"
+          else
+            echo "TUIST_CACHE_ENDPOINT is UNSET (dispatch did not stage a runner-local cache endpoint)"
+          fi
+          echo "--- TUIST_* variable names visible to job steps ---"
+          env | grep -oE '^TUIST_[A-Z0-9_]+' | sort
+
+      - name: Probe candidate cache endpoints (unauthenticated)
+        run: |
+          probe() {
+            desc="$1"; url="$2"
+            echo "=== ${desc}: ${url} ==="
+            host=$(echo "$url" | sed -E 's#^[a-z]+://##; s#[:/].*$##')
+            dig +short +time=3 +tries=1 "$host" 2>/dev/null || echo "DNS: no resolution for ${host}"
+            curl -sS -o /dev/null -w 'HTTP/1.1 GET / -> %{http_code} (%{time_total}s)\n' \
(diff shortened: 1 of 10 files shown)


esnunes:
Closing in favor of an egress-policy approach: hardcoding eu-central as the Linux-serving region doesn't generalize (accounts with Kura in other regions get nothing, and it pins routing to one region). Investigating allowing runner pods to reach the public Kura endpoints directly instead.

PR #11660, opened 2026-07-04
feat(cli): add cas-plugin, an Xcode CAS plugin with Tuist-remote read/write-through

fortmarek:
## What

Ships Tuist's Xcode compilation-cache integration end to end:

- `cas-plugin/`: a Rust `cdylib` (`libtuist_cas_plugin.dylib`) implementing the LLVM CAS plugin ABI (`llcas_*` v0.1) that Xcode's compilation caching loads via `COMPILATION_CACHE_PLUGIN_PATH`. It wraps Xcode's bundled `libToolchainCASPlugin.dylib` for local storage, hashing, and digests, and adds Tuist-remote (kura) read-through on cache misses and write-through on cache writes.
- `tuist-cas-proxy`: a per-machine daemon built from the same crate that owns the remote (REAPI) connection, the resolved-key map, the known-local set, and all publications, and opens the same on-disk local CAS the compilers use.
- Full CLI productization: `tuist setup cache` installs the proxy as a launchd agent, `tuist generate` wires the build settings into generated projects automatically, the dylib and proxy ship in CLI releases, and per-build cache analytics reach the server through the existing pipeline.
- The whole new path is gated on the `kura` client feature flag. The legacy per-project Swift CAS daemon (`tuist cache start`) is **kept**, not removed: accounts without the flag stay on the daemon they use today (see [Backwards compatibility](#backwards-compatibility-gated-on-the-kura-flag)).

It also lands three kura serving fixes that benefit every REAPI consumer, Bazel clients included.

## Why

Xcode's own remote-caching mode (`COMPILATION_CACHE_REMOTE_SERVICE_PATH`, the gRPC socket the old daemon served) is net-negative on deep module graphs. On a 194-module fixture (the tuist CLI scheme itself, 8-core VM, cache endpoint on the VM host at ~2ms latency), a fully warm remote cache built in 325-445s against a 274s no-cache baseline, despite 4447/4449 cache hits. An isolation matrix plus a custom-built swift-build service ruled out our daemon, our transport, the build system's engine callbacks, and its dynamic-task choreography: mid-stall spindumps show every process idle while the closed plugin's embedded gRPC client threads sit unscheduled for 30-50s at a time. That stall is not fixable from outside the closed plugin.

Pointing Xcode at our own plugin (`COMPILATION_CACHE_PLUGIN_PATH`) instead of its remote-service socket sidesteps the entire path: the build system runs in its fast plugin-local mode (no remote service configured), and all remote traffic happens inside code we control.

## Backwards compatibility: gated on the `kura` flag

The plugin + proxy talk to kura over REAPI, but the kura rollout is still in progress, so this cannot be the only path yet. Remote Xcode Cache is not a pilot feature: **72 external accounts** pulled ~4.2 TB of cache artifacts in the last 30 days (several at TB scale — Notion, Whatnot, Withings, Cabify, and more), all on the legacy per-project daemon. Deleting it would silently regress every one of them not yet on kura.

So the new path is gated on the `kura` client feature flag (`TUIST_FEATURE_FLAG_KURA`):

- **Flag set** → `tuist setup cache` installs the machine-wide proxy and `tuist generate` bakes the CAS plugin settings (`COMPILATION_CACHE_PLUGIN_PATH` + `tuist-instance`).
- **Flag unset** (most accounts today) → `tuist setup cache` installs the legacy per-project `cache-start` daemon and `tuist generate` bakes `COMPILATION_CACHE_REMOTE_SERVICE_PATH` (the daemon socket) — unchanged from today.

`setup` / `teardown` boot out whichever agent the other path installed, so only one runs. Migration is per-account: flip the flag once kura reaches that account. The setup docs on this PR stay on the legacy daemon; the new plugin/proxy guide is a separate PR (#11698) that lands at cutover.

## Why Rust, not Swift

**The dylib** implements the `llcas_*` C ABI and is `dlopen`ed in-process by every compiler frontend (thousands of short-lived `swift-frontend`/`clang` processes per build), where it wraps Apple's `libToolchainCASPlugin.dylib`, itself a C `llcas` plugin. A Swift dylib injected into arbitrary compiler processes would pull the Swift runtime into each one and risk a runtime/ABI clash with the toolchain the frontend already belongs to. A Rust `cdylib` has no runtime to inject, loads cheaply, and speaks C at both boundaries (`extern "C"` + `libloading`, `catch_unwind` at the ABI, caller-frees-through-`string_dispose` ownership) with no GC or runtime-init cost on the compiler's hot path.

**The proxy** speaks REAPI gRPC over HTTP/2 to kura, and kura is already Rust: reusing the same `tonic`/`tokio`/`prost` transport gives one REAPI client implementation, with TLS (ring / system roots), h2c, keepalive and bulk-window tuning shared with the server side, and no dependency on grpc-swift for the new path (grpc-swift/SwiftProtobuf stay in the CLI for the retained legacy daemon and `TuistREAPI`'s Bazel-setup probe). It reuses the CLI's Swift auth rather than reimplementing it by shelling out to a hidden `tuist auth token`, so the language boundary is a clean subprocess, not an FFI.

What makes it decisive either way is that **both halves are one crate**: the plugin and the proxy share the wire protocol (`proxy_proto`, one `PROTOCOL_VERSION`), the REAPI client (`reapi`), the analytics writer, and the CAS-node byte parsing. A Swift proxy would have to reimplement or FFI-bridge all of that across a language line; keeping both in Rust means no type or serialization drift between the two sides. The integration also began life as a Rust `llcas` benchmarking shim over Apple's plugin, so productizing it in place avoided a rewrite.

## How it works (shipped architecture)

- The plugin is loaded in-process by every compiler frontend. On a local miss it does one blocking unix-socket round trip to the cache proxy; a put is a write-ahead record plus a fire-and-forget notify.
- The cache proxy is one process per machine (one launchd agent, one socket at `~/.local/state/tuist/cas-proxy.sock`, alongside the analytics DB and where the old daemon's socket lived), multiplexing every project on the machine by REAPI `instance_name` (`account/project`) over kura's single shared gRPC endpoint. It materializes fetched value graphs into the shared local CAS before replying, so the compiler's demand loads are local hits. A `PROTOCOL_VERSION` byte on the plugin/proxy wire protocol rejects a stale proxy across a CLI upgrade (the plugin degrades to a local miss rather than misparsing).
- Instance routing reaches every frontend, including a plain Xcode ⌘B build that carries no CLI environment: the mapper bakes `-cas-plugin-option tuist-instance=<fullHandle>` into `OTHER_SWIFT_FLAGS`, and the proxy persists a `cas_path -> instance` registry so a ⌘B build still routes after a restart. Swift only: Xcode routes clang/ObjC compilation caching to its builtin CAS, never an external plugin, so C/ObjC compiles are locally cached but not remote. The legacy `REMOTE_SERVICE_PATH` did not remote-cache clang either, so this is not a regression.
- Transport is the Bazel Remote Execution API (REAPI) over one multiplexed HTTP/2 channel with keepalive, TLS (system trust roots) for public `https` kura and plaintext h2c for private-network endpoints. A cache hit is one `GetActionResult` plus one `BatchReadBlobs` for the locally-missing set; publication is `FindMissingBlobs` (server-side cross-process upload dedup), `BatchUpdateBlobs`, then `UpdateActionResult` last, which makes the entry-after-graph ordering structural.
- Auth is a shell-out: the proxy fetches and refreshes its bearer by running a hidden `tuist auth token --url <server>` command (over `ServerAuthenticationController`), so it reuses the CLI's full auth (keychain, refresh, OIDC) without reimplementing it and never writes a token to disk. It re-fetches only when the cached JWT is within a small window of its `exp` (checked each maintenance tick, inside the CLI's own 30s refresh threshold so the fetch mints a fresh token), so a long-lived proxy re-auths about once per token lifetime rather than on a fixed cadence. CI seeds `TUIST_CAS_TOKEN` directly.

**Proxy lifetime.** The proxy is a persistent per-machine daemon rather than one started on demand. Most of its value is within a single build (de-multiplexing the ~1-2k short-lived frontends onto one connection, one auth, and one local-CAS handle), which only needs build-duration life; cross-build persistence adds only modest amortization (one TLS handshake, one `tuist auth token` fetch, warm resolved/known-local maps). The on-demand alternative and why it was deferred are in Follow-ups.

## Results (same fixtures and hardware throughout)

| Scenario | no cache | remote-service path (old daemon) | this plugin + proxy | local-replay floor |
|---|---|---|---|---|
| tuist CLI scheme, warm remote, clean DerivedData | 274s | 325-445s (net-negative) | **113.6s** (0.41x) | 105.5s (**~8s over**) |
| Mastodon app, warm remote, clean DerivedData | 87.8s | 55-58s | **50.4s** | 50.1s (**at floor**) |
| tuist CLI scheme, cold populate | 274s | - | **271-293s (≈ no-cache)**, publication drains asynchronously after the build | — |

The local-replay floor (everything already in the local CAS) bounds what any remote transport can achieve; the last column shows how close each warm build lands to it. The shallow Mastodon graph runs at its floor directly; the deep CLI graph lands ~8s over after the kura serving fixes below (cumulative demand-fetch fell 168s to 6.4s per warm build, a 26x reduction). Warm builds perform zero uploads and ~2ms average cache GETs, with zero remote misses and zero transport errors across all validation phases.

By runner-to-kura latency, warm CLI-fixture builds against the improved kura:

| Topology | RTT | Warm (median) | vs no-cache |
|---|---|---|---|
| On-host bridge (8-core VM) | ~1ms | 113.6s | 0.41x |
| Private-network off-host (kura on a nearby fleet node) | 0.89ms | 118s | 0.42x |
| WAN (dev machine to Scaleway Paris) | 66ms | 118s | 0.76x |

Sub-ms topologies sit ~8s over the local floor; at 66ms WAN the plugin is still a clear win but ~65s over floor, because fetching the ~2.5GB namespace costs real round-trip time. That motivates the kura-managed CAS seed for ephemeral WAN machines (a follow-up), and confirms relocating kura off the runner Mac onto a nearby private-network node is nearly free.

## Design rules bought by regressions

Each rule below was a measured fix, kept because removing it regressed:

1. **Upload only on `actioncache_put`, never on `store_object`.** Mirroring every stored object made a warm build issue ~21k redundant uploads (20+ minutes of cumulative POST), saturating the endpoint and inflating GET latency ~2ms to ~49ms. The fix walks the value graph from the `actioncache_put` root in background workers and uploads only what replay needs.
2. **Prefetch ref graphs on entry hits.** Demand-driven fetching resolves one node per round trip; deep chains serialize them (the CLI fixture had 2.3x the keys of the app but 12.5x the remote overhead). A background pool walks the graph breadth-first the moment an entry hits.
3. **Durable publication.** Compiler frontends exit without disposing the CAS (874 of 877 putters in one measured build), so queued uploads died with them. Now put writes a write-ahead record; later plugin instances and the proxy sweep them. Combined with per-process put dedup and a stale-keepalive retry, misses across three consecutive warm builds fell 868 to 47 to 6 to 2 before the transport move eliminated the class entirely.
4. **Panics must not cross the C ABI or skip worker bookkeeping.** Both entry points run under `catch_unwind`; the worker loop decrements in-flight even on a panic so a drain cannot wedge.

Apple's plugin omits several optional `llcas` symbols (`store_from_filepath`, `export_data_to_filepath`, on-disk size/prune), so the loader has a required/optional split with fallbacks; all strings returned to the client are allocated by this crate since the client frees through our `string_dispose`.

## kura serving fixes (benefit every REAPI consumer)

Hunting the last ~50s of warm-over-floor with resolve-phase counters, a read microbench, a bandwidth probe, and a CPU-triggered kura sample exonerated local-CAS ingestion, client batching, the VM bridge (2.34GB/s), and kura's single-request path (15k blobs/s). The governor was a lock convoy: kura's artifact existence cache was touched on every read through one global mutex, capping concurrent serving near 1k blobs/s at 39% CPU. Three fixes landed: `BatchReadBlobs` reads blobs 16-wide instead of sequentially, the existence cache is 32-way sharded with O(1) LRU/TTL per shard, and the client uses bulk-sized HTTP/2 windows with concurrent chunk fetches.

<details>
<summary>Read-path microbench (used to pin the convoy, not committed)</summary>

This throwaway diagnostic pinned the existence-cache lock convoy and validated the sharding + 16-wide-batch fixes. It is not left in the tree; kept here so it can be re-run. Drop it back into `cas-plugin/src/bin/bench-read.rs` with a matching `[[bin]]` entry, set `TUIST_CAS_REMOTE_GRPC_URL` (and a token via `TUIST_CAS_TOKEN`), and `cargo run --release --bin bench-read`. It seeds 480 synthetic ~16KB blobs, then times the per-RPC atom (sequential single-blob batches), one whole-set batch, and 16 concurrent 30-blob batches (the proxy's warm-build shape).

```rust
//! Read-path microbench against a kura REAPI endpoint. Self-seeds synthetic
//! blobs sized like real CAS nodes, then measures: sequential single-blob
//! batches (per-RPC atom), one big batch (server-side batch behavior), and
//! 16 concurrent 30-blob batches (the proxy's warm-build workload shape).

use std::time::Instant;

use tuist_cas_plugin::reapi::{blob_digest, Remote, RemoteConfig};
use tuist_cas_plugin::token::TokenProvider;

fn main() {
    let Some(config) = RemoteConfig::from_env() else {
        eprintln!("TUIST_CAS_REMOTE_GRPC_URL required");
        std::process::exit(2);
    };
    let remote = Remote::new(config, TokenProvider::from_env());

    // Seed 480 blobs of ~16KB with distinct content.
    let mut blobs = Vec::new();
    let mut uploads = Vec::new();
    for index in 0u32..480 {
        let mut content = vec![0u8; 16 * 1024];
        content[..4].copy_from_slice(&index.to_be_bytes());
        for (offset, byte) in content.iter_mut().enumerate().skip(4) {
            *byte = ((offset as u32).wrapping_mul(2654435761).wrapping_add(index) >> 16) as u8;
        }
        let digest = blob_digest(&content);
        blobs.push(digest.clone());
        uploads.push((digest, content));
    }
    let started = Instant::now();
    remote.batch_update(uploads).expect("seed");
    println!("seeded 480x16KB in {:?}", started.elapsed());

    // Per-RPC atom: sequential single-blob batches.
    let started = Instant::now();
    for digest in &blobs[..100] {
        remote.batch_read(std::slice::from_ref(digest)).expect("read");
    }
    let elapsed = started.elapsed();
    println!("sequential single-blob: n=100 total={elapsed:?} per_rpc={:?}", elapsed / 100);

    // Whole set in one batch.
    let started = Instant::now();
    let contents = remote.batch_read(&blobs).expect("batch");
    println!(
        "one batch: n={} bytes={} elapsed={:?}",
        blobs.len(),
        contents.values().map(|value| value.len()).sum::<usize>(),
        started.elapsed()
    );

    // Workload shape: 16 concurrent 30-blob batches over the same channel.
    let chunks: Vec<Vec<_>> = blobs.chunks(30).map(|chunk| chunk.to_vec()).collect();
    let started = Instant::now();
    let mut per_batch = Vec::new();
    std::thread::scope(|scope| {
        let handles: Vec<_> = chunks
            .iter()
            .map(|chunk| {
                scope.spawn(|| {
                    let batch_started = Instant::now();
                    remote.batch_read(chunk).expect("chunk");
                    batch_started.elapsed()
                })
            })
            .collect();
        for handle in handles {
            per_batch.push(handle.join().expect("join"));
        }
    });
    per_batch.sort();
    println!(
        "concurrent 30-blob x{}: wall={:?} batch p50={:?} p95={:?}",
        chunks.len(),
        started.elapsed(),
        per_batch[per_batch.len() / 2],
        per_batch[per_batch.len() * 95 / 100]
    );
}
```

</details>

## CLI productization

- **Setup / teardown (kura-gated).** With the `kura` flag, `tuist setup cache` installs the machine-wide `tuist cache-proxy` launchd agent; without it, the per-project `cache-start` daemon agent. Each path boots out the other's agent so only one runs, and `teardown` removes both.
- **Generated projects (kura-gated).** With `enableCaching: true` in `Tuist.swift`, `tuist generate` bakes `COMPILATION_CACHE_ENABLE_CACHING` always, then on the kura path the CAS plugin settings (`ENABLE_PLUGIN/PLUGIN_PATH` + the `tuist-instance` compiler flag) or on the legacy path `COMPILATION_CACHE_REMOTE_SERVICE_PATH` (the daemon socket). The plugin path resolves to the dylib next to the running `tuist`, so it matches the version that set up the proxy. Non-generated projects set the settings once (docs cover the install locations).
- **Distribution.** `bundle.sh` ships the universal, signed, notarized dylib and proxy binary next to `tuist`; `ResourceLocator` resolves both relative to the executable and honors `TUIST_CAS_PLUGIN_PATH` / `TUIST_CAS_PROXY_PATH` overrides.
- **Analytics parity.** The proxy writes `cas_analytics.db` (rusqlite, WAL, background writer) at the exact `stateDirectory` path the build-report upload ships, matching the Swift `CASAnalyticsDatabase` schema (created_at as SQLite.swift's ISO8601 TEXT; `nodes`/`cas_outputs`/`keyvalue_metadata` encodings identical to what the server reads back, parsed from the closed-CAS-format byte pattern). `tuist cache-proxy` runs the Swift `migrate()` first so the proxy only inserts into the canonical schema. The upload and server (xcactivitylog NIF) side are unchanged.
- **`upload: false`** honored via a single `TUIST_CAS_UPLOAD` flag; when set the plugin serves read hits and publishes nothing.
- **Daemon retained (gated).** The legacy `cache-start` daemon, its gRPC handlers (`CASService`, `KeyValueService`, helpers), the `.proto` files + generated stubs, their tests, and the grpc-swift/SwiftProtobuf/zstd dependencies stay in `TuistCAS` to serve the non-kura path. (An earlier revision of this branch deleted them; that was reverted once the 72-account usage surfaced.)
- **CI.** The canary acceptance test builds the plugin and proxy from source (rust version resolved from `cas-plugin/mise.toml`, matching the kura precedent) and exercises `generated_project_with_caching_enabled` against canary; the release job builds the universal artifacts.

## Robustness (from review)

- `reapi.rs` uses `Endpoint::connect_lazy` so a transient connection failure during a Kura restart is never cached as a permanent error; the channel reconnects per request.
- The publisher drops an item from its dedup set once processed, so a failed publication's write-ahead record can be re-swept and retried (previously it was dropped until the proxy restarted).
- A remote miss no longer negative-caches over a value a concurrent local publish just stored.

## Validation

- Five-phase benchmarks on both fixtures against a fresh cache namespace (so the write path fed its own read path): cold local passthrough, local replay (parity with the plugin-free floor), cold remote populate, and two warm remote runs. Per-process hit/miss/latency counters confirmed the traffic shape at every step (warm: posts n=0, gets ~1.9ms, misses 0).
- `cas-plugin` `cargo test` green (transport, wire protocol, and analytics-encoding tests, including a regression that records through the proxy path into a Swift-created canonical schema).
- Project regenerates and `xcodebuild -scheme tuist` builds green from a clean slate.
- End-to-end locally against canary: `tuist setup cache` installs the proxy and resolves the endpoint; `tuist generate` bakes the settings and routes the instance; the plugin loads in `xcodebuild` and spools publications to the proxy.

## Follow-ups

- Migrate the ~72 daemon accounts to kura (flip `TUIST_FEATURE_FLAG_KURA` per account as kura reaches them), then make the plugin/proxy the default and retire the daemon. Land the new-architecture docs PR (#11698) at that cutover.
- Confirm the analytics dashboard join in staging. The proxy's node byte-scan is unit-tested and structurally identical to the daemon's, but the closed-CAS byte-pattern survival and the `nodeID`-to-build-log join can only be verified against a build with a live cache; a laptop cannot reach kura's REAPI gRPC through the public edge.
- Wait for uploads to drain before collecting the build report. Uploads are fire-and-forget: the plugin writes a durable write-ahead spool record and notifies the proxy, which uploads and writes the "write"-side analytics only once the transfer completes. The build-report collection (`UploadBuildRunService`) copies `cas_analytics.db` with no barrier waiting for that drain. Warm builds are unaffected (reads are recorded synchronously during the build and warm builds upload nothing), but on a cold/populate build publication drains after the build, so its upload analytics can be incomplete in the report; on an ephemeral CI runner torn down right after, the still-queued spool records and their uploads are lost entirely (no later build on that disk to sweep them). Bazel avoids this by owning the build and treating end-of-invocation as a barrier: it awaits outstanding cache uploads and the BEP/metrics upload before the command returns (`--bes_upload_mode=wait_for_upload_complete` by default), so the runner cannot be torn down mid-upload and its metrics reflect what landed. We could retrofit the same idea: at build end, `tuist` signals the proxy to flush its publish queue and sweep the spool, waits (bounded), then collects the report. Scope it to the populate path (`TUIST_CAS_UPLOAD` on, or an explicit `--wait`) so consume-cache jobs stay fast. Deferred; the write ordering is already durable (blobs before `UpdateActionResult`), so nothing is half-committed remotely, only under-reported for that build.
- kura-managed materialized-CAS seed (clonefile on dispatch, epoch-invalidated by kura) as the zero-transfer option for ephemeral WAN / runner fleets.
- Confirm whether developer laptops reach the REAPI gRPC service through the public cache endpoint, or only private-network clients do; the plugin's remote path depends on it.
- Socket under an `XDG_STATE_HOME` override. The proxy socket is `HOME`-anchored (`~/.local/state/tuist/cas-proxy.sock`) and does not follow `XDG_STATE_HOME`, because the plugin resolves it in-process where an Xcode ⌘B build carries no CLI environment and cannot see the override. The analytics DB does follow it (Swift resolves that path and passes it via env), so under an override the two split into different dirs. The daemon kept them together by baking the socket path into build settings; restoring that (bake `-cas-plugin-option socket=` like `tuist-instance`, plus set it in the launchd plist) would only cover generated projects, and `XDG_STATE_HOME` is rare on macOS, so it is deferred.
- On-demand proxy (alternative to the persistent daemon). A lazily-started, idle-terminating proxy would drop the `tuist setup cache` / `teardown cache` step and the version skew between a launchd-pinned proxy and an upgraded `tuist` (the agent hard-codes an absolute `tuist` path, so a `mise`/`brew` bump needs a re-run of setup to stay matched). Starting it from the plugin implies a thundering-herd race across hundreds of frontends and, more fundamentally, may be blocked by Xcode's compile sandbox; the cleaner form starts it once from a scheme pre-build action (outside the sandbox, on both ⌘B and `xcodebuild -scheme`) and self-terminates when idle. Deferred for v1; the sandbox/spawn question gates whether it is viable.


pepicrft (approved):
Went through this in depth, plugin, proxy, REAPI, and the CLI gating. The architecture holds up: I traced the failure paths and remote/proxy/socket/version errors all degrade to a local miss (and a HIT re-stores through the content-addressed store, so a mislabeled blob can never surface as the wrong object), the publish order genuinely gates UpdateActionResult behind all blobs, the circuit breaker fails open, and the analytics schema is test-pinned against the Swift one. The proxy-vs-daemon flag is consistent in the normal flow since both generate and setup read it from the same mise env.

Approving. Left three review threads with the non-blocking items:

- C/ObjC routing when a custom DerivedData path is shared across projects (question for you, since the instance is only baked into OTHER_SWIFT_FLAGS). Depends on whether we recommend one DerivedData path per project or a shared one.
- Sync llcas entry points are not catch_unwind-wrapped while the async ones are, which contradicts the "panics never cross extern C" invariant.
- The prefetch inflight decrement is manual rather than a Drop guard, so a poisoned-lock panic in the post-process cleanup would wedge drain_stop.

None of these block the merge. The routing one is the only one that needs a call from us on the intended DerivedData layout.

fortmarek:
Thanks for the thorough review, Pedro. All three addressed.

**1. C/ObjC routing with a shared DerivedData path.** There's no routing question here, and I want to correct something I'd assumed earlier: clang/ObjC never loads the CAS plugin at all. We spiked this on Xcode 26.5: `COMPILATION_CACHE_PLUGIN_PATH` feeds `swift-frontend` only, clang gets the builtin CAS (`-fcas-backend`) and never `-fcas-plugin-path`, and forcing it through `OTHER_CFLAGS` is ignored. So C/ObjC is cached locally but never remote, both under this plugin and under the legacy `COMPILATION_CACHE_REMOTE_SERVICE_PATH` (which didn't feed clang's service path either, so no regression). Since only Swift is ever remotely cached and Swift always carries `tuist-instance`, the DerivedData layout (shared or per-project) can't misroute anything. I also fixed a misleading comment in the mapper that implied clang falls back to the proxy registry.

**2. Sync entry points not `catch_unwind`-wrapped.** Fixed in 800c72163df. The synchronous `load` and `actioncache get`/`put` now go through the same guard as the async paths: load and get map a panic to `LLCAS_LOOKUP_RESULT_ERROR`, and the put paths swallow it (the local put has already succeeded and the remote publish is best-effort). A panic can no longer cross the `extern "C"` boundary and abort the compiler.

**3. Manual inflight decrement vs Drop guard.** Fixed in the same commit. The decrement (and its drain wakeup) is now an `InflightGuard` `Drop`, with poison recovery on the queue lock, so a panic in the post-process cleanup runs the decrement during unwind instead of leaving `drain_stop` waiting on a counter that never returns to zero.

cli/Sources/TuistKit/Services/Setup/SetupCacheCommandService.swift
```diff
@@ -146,24 +111,137 @@ struct SetupCacheCommandService {
...
+        if let accountHandle {
+            programArguments.append(contentsOf: ["--account", accountHandle])
+        }
+
+        var environmentVariables: [String: String] = [:]
+        // The proxy fetches and refreshes its bearer itself by shelling out to
+        // `tuist auth token`. On CI, where the credential is an environment token
+        // rather than a keychain session, seed it directly.
+        if let token = Environment.current.tuistVariables[Constants.EnvironmentVariables.token] {
+            environmentVariables["TUIST_CAS_TOKEN"] = token
```
esnunes, line 171: One thing I noticed on the CI path: `installProxy` seeds the agent environment with `TUIST_CAS_TOKEN` only, but the `tuist cache-proxy` wrapper gates on `ServerAuthenticationController.authenticationToken`, whose environment lookup reads only `TUIST_TOKEN`/`TUIST_CONFIG_TOKEN` (ServerAuthenticationController.swift:695) before falling back to the keychain — neither exists under launchd on CI. With `KeepAlive.SuccessfulExit = false` the wrapper's clean exit means the Rust proxy never starts, while `tuist setup cache` itself reported success (its own process env had the token). Forwarding `TUIST_TOKEN` alongside `TUIST_CAS_TOKEN` — or teaching the wrapper's guard to accept `TUIST_CAS_TOKEN` — would close the gap.

cli/Sources/TuistKit/Services/Setup/SetupCacheCommandService.swift
```diff
@@ -62,61 +62,26 @@ struct SetupCacheCommandService {
...
-        // feature flags above are not enough on their own: getCacheEndpoints
-        // returns the public, CLI-facing endpoint regardless of the kura flag.
-        if let cacheEndpoint = Environment.current.variables["TUIST_CACHE_ENDPOINT"] {
-            environmentVariables["TUIST_CACHE_ENDPOINT"] = cacheEndpoint
+        // The `kura` client feature flag selects the machine-wide CAS proxy +
+        // plugin. Without it, accounts stay on the legacy per-project cache daemon
+        // they rely on today, until they are migrated to kura.
+        let kuraEnabled = ClientFeatureFlags.contains("kura")
+        if kuraEnabled {
+            try await installProxy(fullHandle: fullHandle, serverURL: serverURL)
```
esnunes, line 76: It looks like `xcodeCache(upload: false)` doesn't cross over on the kura path: the legacy branch passes `--no-upload`, but `installProxy` doesn't propagate it, and I couldn't find anything setting `TUIST_CAS_UPLOAD` outside cas-plugin — so `env_bool("TUIST_CAS_UPLOAD", true)` (lib.rs:489) always defaults to uploads-on (the comment at lib.rs:214 documents the `xcodeCache(upload:)` intent). Accounts relying on read-only dev machines would silently start publishing once flagged onto kura. Since the proxy is per-machine and the setting per-project, the generated build settings (like `tuist-instance` via `-cas-plugin-option`) might be the right carrier rather than the agent env — but any of the channels would do.

cas-plugin/src/bin/tuist-cas-proxy.rs
```diff
@@ -0,0 +1,96 @@
...
+    let socket_path = std::env::var("TUIST_CAS_PROXY_SOCKET")
+        .ok()
+        .filter(|socket| !socket.is_empty())
+        .unwrap_or_else(tuist_cas_plugin::default_proxy_socket);
+    let Ok(grpc_url) = std::env::var("TUIST_CAS_REMOTE_GRPC_URL") else {
+        eprintln!("TUIST_CAS_REMOTE_GRPC_URL is required");
+        std::process::exit(2);
+    };
+    let tokens = TokenProvider::from_env();
+    let upstream_plugin = std::env::var("TUIST_CAS_UPSTREAM_PLUGIN").unwrap_or_else(|_| {
```
esnunes, line 42: The launchd-spawned proxy resolves Apple's plugin as `TUIST_CAS_UPSTREAM_PLUGIN` → `$DEVELOPER_DIR/…` → hardcoded `/Applications/Xcode.app`, and the env written by setup/`CacheProxyCommandService` includes neither of the first two — I couldn't find an `xcode-select`/`xcrun` fallback in the crate. So on machines where Xcode lives at a versioned path (most CI images, plenty of dev machines), every resolve fails `Upstream::load` and silently degrades to a local miss, visible only via `TUIST_CAS_LOG`. Resolving `DEVELOPER_DIR` in the Swift launcher (which can ask the system) and passing `TUIST_CAS_UPSTREAM_PLUGIN` explicitly — or adding an `xcode-select -p` fallback in Rust — seems worth doing before rollout.

cas-plugin/src/proxy.rs
```diff
@@ -0,0 +1,787 @@
...
+            Some(manifest) if !manifest.is_empty() => manifest,
+            _ => {
+                // A local compiler may have published (and locally stored) this
+                // key between our get_action miss and now; the publisher pool
+                // runs outside this resolve's single-flight guard. Never clobber
+                // that Some with a negative entry, or the long-lived proxy would
+                // answer misses for a key it can actually serve. Prefer the
+                // freshly-published value if present.
+                {
+                    let mut resolved = state.resolved.lock().unwrap();
```
esnunes, line 286: `resolve_uncached` negative-caches a remote miss as `None` in `resolved`, and the only thing that flips it is a *local* publish — there's no TTL, and `enforce_cache_bounds` only wholesale-clears past 1M entries. On a long-lived launchd proxy that means keys published later by another machine (overnight CI being the typical writer) stay cached misses until the daemon restarts. The incomplete-graph path already avoids negative-caching for exactly this kind of reason, so the machinery is close: a modest TTL on the `None` entries (positives are content-addressed and safe to keep forever) would let warm results land.

cli/Sources/TuistKit/Services/Setup/SetupCacheCommandService.swift
```diff
@@ -146,24 +111,137 @@ struct SetupCacheCommandService {
...
+        for (key, value) in ClientFeatureFlags.environmentVariables() {
+            environmentVariables[key] = value
+        }
+        if let cacheEndpoint = Environment.current.variables["TUIST_CACHE_ENDPOINT"] {
+            environmentVariables["TUIST_CACHE_ENDPOINT"] = cacheEndpoint
+        }
+
+        // Boot out the machine-wide proxy so the two do not both run.
+        let proxyLabel = Environment.current.casProxyLaunchAgentLabel()
+        try? await launchAgentService.teardownLaunchAgent(
```
esnunes, line 241: A machine-topology question: with the proxy being one machine-wide agent, `installLegacyDaemon` boots out `tuist.cas-proxy` and `installProxy` boots out the current project's legacy agent — so a machine with one kura-account project and one legacy-account project loses whichever transport ran setup earlier. And with two kura accounts, both setups write the same `tuist.cas-proxy` plist while the launcher resolves one account-scoped endpoint (proxy.rs:86 notes all instances share the machine's endpoint + token), so the last `--account` wins. Is one-account-per-machine an accepted constraint for the rollout? If so that's fair — but a warning when setup replaces an agent configured for a *different* account would make the failure visible instead of silent.

cli/Sources/TuistGenerator/Mappers/XcodeCacheSettingsProjectMapper.swift
```diff
@@ -32,12 +44,35 @@ public struct XcodeCacheSettingsProjectMapper: ProjectMapping {
...
-            baseSettings["COMPILATION_CACHE_ENABLE_PLUGIN"] = "YES"
-            baseSettings["COMPILATION_CACHE_REMOTE_SERVICE_PATH"] = .string(
-                Environment.current.cacheSocketPathString(for: fullHandle)
-            )
+            if kuraEnabled {
+                // kura path: route Xcode's compilation caching through the Tuist
+                // CAS plugin, which owns remote (kura) read/write-through via the
+                // per-machine proxy. Needs the bundled dylib; when it is absent the
+                // build stays on local-only caching rather than a broken plugin path.
+                if let casPluginPath = try await resolvedCASPluginPath() {
```
esnunes, line 52: When kura is enabled but the dylib isn't found, the mapper silently emits local-only settings — `resolvedCASPluginPath()` has no else branch, `casPluginCandidates()` swallows a malformed `TUIST_CAS_PLUGIN_PATH` via `try?`, and `setup cache` printed "enabled 🎉" without a dylib check (the `map_whenPluginMissing_addsLocalCacheSettingsOnly` test codifies the silence). Degrading rather than failing seems right; degrading *silently* means the first symptom is a cold cache. A one-line warning naming the missing dylib when kura is on and a fullHandle is set would save that support ticket.

cli/Sources/TuistKit/Services/Teardown/TeardownCacheCommandService.swift
```diff
@@ -37,23 +37,29 @@ struct TeardownCacheCommandService {
...
+            plistFileName: "\(proxyLabel).plist"
         )
 
-        let socketPath = Environment.current.cacheSocketPath(for: fullHandle)
-        if try await fileSystem.exists(socketPath) {
-            try await fileSystem.remove(socketPath)
+        // Best-effort teardown of the per-project cache daemon (the non-kura path,
+        // and any leftover from before the machine-wide proxy).
+        let resolvedPath = try await Environment.current.pathRelativeToWorkingDirectory(path)
+        if let config = try? await configLoader.loadConfig(path: resolvedPath),
```
esnunes, line 39: Teardown now wraps the config load and legacy-agent removal in `try?` and prints "torn down 🧹" unconditionally — a non-kura user whose config fails to load keeps a running per-project agent and still sees success (previously this threw `missingFullHandle`, which is now a dead error case). Best-effort feels right for teardown; a warning when the legacy agent couldn't be identified — plus removing the dead enum case — would keep the success message honest. Definitely not a blocker.

cas-plugin/src/reapi.rs
```diff
@@ -0,0 +1,533 @@
...
+                    }
+                }
+                Ok(all)
+            })?;
+            let mut contents = std::collections::HashMap::new();
+            for entry in outcomes {
+                let ok = entry.status.as_ref().map(|s| s.code == 0).unwrap_or(false);
+                if ok {
+                    if let Some(digest) = entry.digest {
+                        contents.insert(digest.hash, entry.data.to_vec());
```
esnunes: Two easy cleanups in reapi.rs: (1) `contents.insert(digest.hash, entry.data.to_vec())` deep-copies every fetched blob even though the loop consumes `entry` by value — `entry.data` moves for free, and batches run up to 32MB while the requesting compiler blocks on the resolve; (2) the `for attempt in 0..ATTEMPTS` retry scaffold is copy-pasted five times with three structural variants already drifting apart (one arm ends in `unreachable!()`), so a small `with_retry(op, f)` helper would keep the policy in one place. Neither is blocking.

cli/Tests/TuistKitTests/Services/Setup/SetupCacheCommandServiceTests.swift
```diff
@@ -55,12 +55,17 @@ struct SetupCacheCommandServiceTests {
...
+        given(launchAgentService)
+            .teardownLaunchAgent(label: .any, plistFileName: .any)
+            .willReturn()
     }
 
     @Test(.inTemporaryDirectory, .withMockedEnvironment(), .withMockedLogger()) func setupCache_withTuistProject() async throws {
         // Given
         let environment = try #require(Environment.mocked)
         environment.currentExecutablePathStub = AbsolutePath("/usr/local/bin/tuist")
+        environment.variables["TUIST_FEATURE_FLAG_KURA"] = "1"
```
esnunes, line 68: With the test update setting `TUIST_FEATURE_FLAG_KURA` in every setup test, the whole suite now exercises the proxy branch — which leaves `installLegacyDaemon` (the path every un-flagged account runs) without unit coverage: nothing pins the per-project label, the `cache-start` args, the `TUIST_TOKEN` env seeding, or `--no-upload`. The mapper's legacy branch got a test earlier in this PR; given the backwards-compat promise rests on the service branch staying byte-for-byte, one test without the flag pinning those assertions seems cheap insurance.

esnunes (approved):
This is a seriously impressive piece of engineering — the measurement-driven pivot from Xcode's remote-service mode to owning the plugin surface, the write-ahead spool for compilers that exit without disposing, and the consistently degrade-to-miss failure semantics all read as carefully thought through. A dedicated security pass over the trust boundaries (socket permissions, bearer containment, digest re-verification on fetched content, kura authz) came back clean.

The inline notes cluster into four themes:

1. **The launchd clean-environment boundary** — the CI token never reaches the wrapper's auth guard, and Apple's plugin can't be resolved on non-default Xcode installs. Same root cause, two symptoms; worth one deliberate "what env does the agent actually see" pass.
2. **The config→proxy contract** — the `xcodeCache(upload: false)` opt-out doesn't cross over on the kura path.
3. **Silent success/degradation** — a missing plugin dylib degrades generation to local-only with no warning, teardown reports success even when it couldn't identify the legacy agent, and the proxy's negative cache never expires, so keys published later by other machines stay misses until the daemon restarts.
4. **Rollout guardrails** — the single machine-wide agent vs multiple accounts/legacy projects question, and the legacy setup branch now having no unit coverage. A couple of easy `reapi.rs` cleanups are noted inline as well.

A few smaller things I didn't leave inline notes on: `TUIST_CAS_POOL` is documented in cas-plugin/AGENTS.md but nothing reads it (the plugin uses `TUIST_CAS_PREFETCH` and the proxy hardcodes 8); the proxy's thread-per-connection serving is fine at current load, though the Prefetcher pool in the same crate would make a cheap accept pool if it ever shows up in profiles; and cli/AGENTS.md says to avoid new code in TuistKit — `CacheProxyCommand` follows the existing `Commands/Cache/` precedent so I read it as accepted, just noting it.

Also for transparency, two scary-looking candidates were investigated and refuted: the `OTHER_SWIFT_FLAGS` string→array conversion is safe (Xcode re-tokenizes array elements on whitespace, and `SettingsHelper.merge` has used the same pattern for years), and kura's `trim_to` floor-division can't flush the cache with any real caller's targets.

fortmarek:
Thanks for the exceptionally thorough pass, @esnunes. Worked through all of it. Every inline note is addressed across three commits (5afc1d6, d393ba9, fdf179d), grouped below by your four themes.

**1. The launchd clean-environment boundary.**
- CI token vs. the wrapper's auth guard: `installProxy` now forwards `TUIST_TOKEN` alongside `TUIST_CAS_TOKEN`, so `authenticationToken` finds it under launchd instead of falling through to an empty keychain and exiting clean. (d393ba9)
- Apple's plugin on non-default Xcode: added an `xcode-select -p` fallback in `upstream_path()`, after `TUIST_CAS_UPSTREAM_PLUGIN`/`DEVELOPER_DIR`, so a versioned Xcode install resolves instead of silently degrading every resolve to a local miss. (5afc1d6)

**2. The config to proxy contract.**
- `xcodeCache(upload: false)` now crosses to the kura path. You were right that the build settings are the correct carrier since the proxy is per-machine and the setting per-project: the mapper bakes `-cas-plugin-option tuist-upload=false`, and the plugin reads it (`resolve_upload`: option, then `TUIST_CAS_UPLOAD`, then default-on). (5afc1d6 plugin, d393ba9 mapper, fdf179d test)

**3. Silent success/degradation.**
- Missing dylib: the mapper now warns (naming the dylib) when kura is on and a fullHandle is set but no dylib is found, rather than letting a cold cache be the first symptom. (d393ba9)
- Teardown honesty: warns when it can't resolve a fullHandle to identify the legacy agent; the now-dead `missingFullHandle` enum is removed. (d393ba9)
- Negative-cache staleness: `resolved` misses now carry a timestamp and re-resolve after a 60s TTL, so a key another machine publishes later stops being served as a cached miss without a proxy restart. Positives are kept forever (content-addressed). (5afc1d6)

**4. Rollout guardrails.**
- One-account-per-machine: accepted as a rollout constraint for now, so the single machine-wide agent is left as is.
- Legacy-branch coverage: added `setupCache_withoutKuraFlag_installsLegacyDaemon`, pinning the per-project label, `cache-start` args (incl. `--no-upload`), and `TUIST_TOKEN` seeding, since the setup suite moving entirely to the kura flag had left `installLegacyDaemon` uncovered. (fdf179d)

**reapi.rs cleanups.** `entry.data` now moves instead of `.to_vec()`-copying every fetched blob, and the five retry loops fold into `retry_call`/`retry_call_async` (one async variant because `batch_read` runs inside a spawned task). (5afc1d6)

**The smaller notes.** Dropped the dead `TUIST_CAS_POOL` from AGENTS.md and documented `tuist-upload`. Thanks for chasing down the `OTHER_SWIFT_FLAGS` array conversion and `trim_to` and confirming both are safe.

PR #11678, opened 2026-07-06
fix(server): make OG image generation resilient

fortmarek:
## What changed

- Added a shared `Tuist.Mix.OgImageRenderer` for build-time docs and marketing OG image generation.
- Centralized Carta/BrowseChrome startup, cache checks, render timeouts, and task-exit warning handling.
- Added a Chrome preflight and libvips fallback so server image builds can still produce OG images when Chromium DevTools startup is unavailable or stalls.
- Made the libvips fallback explicitly write JPEGs when the target path has no normal extension, covering docs root slugs such as `/en` that render to `en/.jpg`.
- Reused the shared renderer from both `mix docs.gen.og_images` and `mix marketing.gen.og_images`.
- Added a shared CLDR locale prefetch helper used by both the server test job and the Docker build.
- Cache the prefetched CLDR locale JSON directory across CI runs.
- Prefetch CLDR locale JSON in the Docker deps stage before the main builder and `og-images` stages compile.

## Why

The server image release jobs were failing in the `og-images` Docker stage while running `mix docs.gen.og_images`. The original logs showed repeated BrowseChrome pool initialization failures with `:devtools_timeout`, followed by the docs generator's `Task.async_stream` timing out after 30 seconds. After adding the fallback, the next CI run showed the fallback path being exercised but crashing on the docs home page image.

## Root cause

The docs and marketing OG image tasks assumed BrowseChrome would always start successfully and that every Carta render would complete within the async stream's 30-second timeout. In the failing Docker build, Chromium never became reachable over DevTools quickly enough, so worker initialization kept timing out and the stream killed the docs render task.

The follow-up Docker failure happened because the docs root slug `/en` maps to `en/.jpg`. Carta writes raw JPEG bytes and did not care about that hidden filename shape, but the libvips fallback infers the output type from `Path.extname/1`; for `.jpg` as a basename, that extension is empty, so `Image.write!/3` could not infer an image type.

The server `Test` job also failed once before tests ran because `ex_cldr` downloaded locale JSON from unauthenticated `raw.githubusercontent.com` during the full-locale compile and received HTTP 429 for `pl` and `pt`. After caching that path in the workflow, the Docker build still failed in `RUN mix compile --warnings-as-errors` for the same reason: the builder stage had no prefetched CLDR files, so it tried to download `pl`, `tr`, `yue-Hant`, and `zh-Hans` from `raw.githubusercontent.com` and hit 429s.

## Approach

The new renderer keeps Carta as the primary path, but treats Chrome availability as something to verify before the bulk render. If preflight or an individual render fails, it logs the problem and falls back to the existing libvips-based OpenGraph renderer. The fallback uses a separate cache key prefix, so a degraded Chrome build does not poison the normal Carta cache. The shared renderer also caps default concurrency at 4, supports `TUIST_OG_IMAGE_POOL_SIZE`, and extends per-render task timeouts to avoid killing slow-but-healthy renders too aggressively.

The libvips writer now adds `suffix: ".jpg"` only when the target path has no extension, preserving the existing filenames while making output type detection explicit.

For the full-locale compiles, the workflow restores a narrow Actions cache for `_build/cldr/locales` and a shared shell helper prefetches any missing CLDR locale JSON files required by the full Gettext locale set through the GitHub API with retries. That keeps `ex_cldr` from falling back to rate-limited `raw.githubusercontent.com` downloads during compile while avoiding the broader Mix build cache that the test job intentionally disables. The Dockerfile copies and runs the same helper in the shared `deps-builder` stage, so both the main release builder and the parallel `og-images` stage inherit the prefetched CLDR data before they compile.

## Impact

Server release Docker builds should no longer fail solely because headless Chromium cannot initialize DevTools during build-time OG generation. The normal high-fidelity Carta path remains the default when Chrome is healthy, while the fallback keeps the build producing usable OG images when Chrome is not.

The full-locale test compile and the Docker release image compile should also be less sensitive to GitHub raw rate limits on fresh CI runners.

## Validation

- `mix format lib/tuist/marketing/open_graph.ex`
- `mix format --check-formatted lib/tuist/marketing/open_graph.ex`
- CLDR prefetch script exercised locally against the GitHub API for all required locales.
- Confirmed the CI token-authenticated prefetch variant failed with HTTP 403, then switched to the unauthenticated GitHub API endpoint with retries.
- Confirmed the latest Docker failure moved to CLDR locale downloads in the builder stage, then moved the prefetch into the Docker deps stage.
- `bash -n server/mise/tasks/cldr/prefetch-locale-data.sh`
- Fresh temp-dir CLDR prefetch produced all 12 required locale JSON files.
- `MIX_ENV=prod mix compile`
- Hidden-filename fallback smoke test generated `/private/tmp/tuist-og-hidden/.jpg` through `Tuist.Marketing.OpenGraph.generate_og_image/2`.
- `MIX_ENV=test TUIST_DEV_ALL_LOCALES=1 mix compile --warnings-as-errors`
- `git diff --check`

I could not run the full Docker build locally because the Docker daemon is not available in this environment.

### DIFF
--- a/.github/workflows/server.yml
+++ b/.github/workflows/server.yml
@@ -243,6 +243,15 @@ jobs:
           mix-cache-key: mix-test-v2
           mix-cache-restore-key: mix-test-v2-
           restore-mix-cache: "false"
+      - name: Restore CLDR locale cache
+        uses: actions/cache@v4
+        with:
+          path: server/_build/cldr/locales
+          key: cldr-locales-v1-${{ hashFiles('server/mix.lock') }}
+          restore-keys: |
+            cldr-locales-v1-
+      - name: Prefetch CLDR locale data
+        run: mise/tasks/cldr/prefetch-locale-data.sh
       - name: Compile
         run: mix compile --warnings-as-errors
       - name: Start ClickHouse
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -63,7 +63,7 @@ WORKDIR /app
 # ========================================
 FROM ${BUILDER_IMAGE} AS deps-builder
 
-RUN apt-get update -y && apt-get upgrade -y && apt-get install -y build-essential git \
+RUN apt-get update -y && apt-get upgrade -y && apt-get install -y build-essential ca-certificates curl git \
   && apt-get clean && rm -f /var/lib/apt/lists/*_*
 
 WORKDIR /app
@@ -89,6 +89,7 @@ ENV HEX_HTTP_TIMEOUT=120
 ENV ELIXIR_ERL_OPTIONS="+S 2"
 
 COPY server/mix.exs server/mix.lock ./
+COPY server/mise/tasks/cldr mise/tasks/cldr
 RUN mkdir config
 
 # Copy path dependencies (only mix.exs/mix.lock for dep resolution)
@@ -121,6 +122,7 @@ COPY server/config/config.exs config/
 COPY server/config/prod.exs server/config/can.exs server/config/stag.exs config/
 
 RUN mix deps.compile
+RUN mise/tasks/cldr/prefetch-locale-data.sh
 
 # ========================================
 # Stage: OG image generator
--- a/server/lib/mix/og_image_renderer.ex
+++ b/server/lib/mix/og_image_renderer.ex
@@ -1,0 +1,165 @@
+defmodule Tuist.Mix.OgImageRenderer do
+  @moduledoc """
+  Build-time OG image renderer shared by the docs and marketing Mix tasks.
+  """
+  alias Tuist.Marketing.OgImageCache
+  alias Tuist.Marketing.OpenGraph
+
+  @max_pool_size 4
+  @preflight_timeout 45_000
+  @render_timeout 120_000
+  @stream_timeout @render_timeout + 10_000
+
+  def pool_size do
+    case System.get_env("TUIST_OG_IMAGE_POOL_SIZE") do
+      value when is_binary(value) ->
+        case Integer.parse(value) do
+          {pool_size, ""} when pool_size > 0 -> pool_size
+          _ -> default_pool_size()
+        end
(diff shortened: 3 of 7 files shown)


pepicrft (approved):

esnunes:
**CLDR prefetch fetches the GitHub API unauthenticated** — `server/mise/tasks/cldr/prefetch-locale-data.sh:42`

`prefetch-locale-data.sh` (new in this PR) fetches all 12 locales from `api.github.com` with no `Authorization` header, so it's bound by the REST API's ~60 req/hr unauthenticated limit per egress IP.

This isn't fragility from nothing — it replaces `ex_cldr`'s implicit `raw.githubusercontent.com` compile-time downloads (the 429 being fixed). But the two sit in different rate-limit buckets, and the REST `contents` ceiling (60/hr, shared with any other REST call from that IP) is *stricter* than raw's, so on the shared-egress runners a release cascade + concurrent PR cache-misses can breach it → 403 → `curl --retry-all-errors` retries the 403 → `set -e` hard-fails the build.

`GITHUB_TOKEN` is already in the step's env in `server.yml` (just unused); the earlier "token variant 403'd" reads like an *empty*-token auth failure in the Docker build (authenticated is 5000/hr), not a rate limit.

**Fix — authenticate on both paths:**
- **CI:** send `Authorization: Bearer $GITHUB_TOKEN` from the token already in the step env.
- **Docker:** use a BuildKit secret so the token never lands in an image layer — `RUN --mount=type=secret,id=github_token,env=GITHUB_TOKEN mise/tasks/cldr/prefetch-locale-data.sh`, wired via the `secrets:` input on `docker/build-push-action@v6` (or `--secret` on `docker buildx build`). Both build paths here already run under BuildKit, so it's a plumbing change, not an architecture one.

Have the script send the header only when a token is present, so local/dev builds still work unauthenticated.

PR #11679, opened 2026-07-06
perf(server): reduce ClickHouse CPU wait

fortmarek:
## What changed

- Add an id-ordered `proj_by_id` projection to `command_events`, split into metadata and materialization migrations.
- Rewrite the flaky-test rolling-window query so it deduplicates bounded recent-run arrays in-place instead of expanding every run through `ARRAY JOIN`, grouping, sorting, and `LIMIT BY`.
- Split ClickHouse read/write `max_threads` config helpers and fix the duplicate `IngestRepo` `pool_size` key.
- Move ClickHouse repo pool sizes to non-secret Helm chart values and render them into the server, migration, processor, and xcresult-processor workloads.

## Why

The slow-query alert was firing for two ClickHouse queries, but the broader CPU-wait investigation showed those queries were symptoms of a strained ClickHouse instance rather than the whole cause.

Using Atlas/Grafana against production `system.query_log` and `system.metric_log`, the exact alert window (`2026-07-06 10:53-11:23 UTC`) showed the largest CPU-wait contributor was `CommandEvents.get_command_event_by_id/1`:

- 16,008 executions in 30 minutes
- ~350 GiB read
- ~12.3B rows read
- ~2,417s CPU time
- ~1,113s CPU wait

That query filters `command_events` by id only, but the table is ordered by `(project_id, name, ran_at)`. Production already has an `idx_id` bloom-filter skip index, but `EXPLAIN indexes = 1` still showed it reading 291 of 7,925 granules for a single id lookup. The projection gives ClickHouse a physical layout that matches the point lookup.

The flaky rolling query also accumulated meaningful CPU wait in the same alert window. Keeping the recent-run work inside arrays avoids multiplying each active test case into hundreds of intermediate rows.

I intentionally dropped the proposed `test_case_runs` hydration partition hint from this PR after quantifying it. That query had S3/I/O tail latency, but it was not a meaningful contributor to the CPU-wait problem and added more complexity than it justified here.

## Impact

Run detail lookups by command event id should stop repeatedly scanning wide `command_events` granules under high refresh traffic. Flaky monitor evaluations should spend less CPU on rolling windows.

The new ClickHouse config helpers keep existing `max_threads` defaults: read and write `max_threads` both fall back to `clickhouse.max_threads` unless dedicated runtime config is provided. The `pool_size` cleanup now preserves the managed ClickHouse repo pool sizes instead of quietly reducing the write buffer pool. Production and staging set both `clickhouse.poolSize` and `clickhouse.bufferPoolSize` to `15`; canary sets both to `30`. These are chart-owned operational knobs, not values sourced from the runtime secret bundle, and the chart renders them into the server, migration, processor, and xcresult-processor pods.

The `proj_by_id` projection stores a full id-ordered copy of `command_events` (`SELECT * ORDER BY id`). Production `command_events` is currently ~26.8M rows / ~5.96 GB compressed, so materialization will add storage in that same order of magnitude and each insert will pay projection-maintenance cost. Rollout should watch ClickHouse disk usage and insert latency after `MATERIALIZE PROJECTION` finishes.

## Validation

- Reproduced the Grafana alert query through Atlas. Note: the alert rule computes `p99`, while the Slack description says `p90`.
- Queried production ClickHouse metrics for the alert window and identified the dominant CPU-wait hash.
- Ran `EXPLAIN indexes = 1` for the hot `command_events WHERE id = ?` query; confirmed the existing bloom filter still read 291 granules.
- Queried production ClickHouse via Atlas; `command_events` is currently ~26.8M rows / ~5.96 GB compressed before adding the projection.
- Rendered the managed Helm chart for production, canary, and staging with server/migration/processor/xcresult-processor templates selected; confirmed the ClickHouse pool env vars render into all four workloads with production/staging `15` and canary `30`.
- Validated the rewritten rolling-window ClickHouse expression against production data:
  - tiny project: old/new result counts matched with no diff
  - 200-case sample from a large project: old/new result counts matched with no diff
- Ran `git diff --check` / `git diff --cached --check`.
- Parsed touched Elixir files with `Code.string_to_quoted!`.

Could not run `mix format --check-formatted` or Mix tests because `mix deps.get --check-locked` reports the existing lock drift `hpax 1.0.3 => 1.0.4`. I left `mix.lock` untouched because that dependency update is unrelated to this ClickHouse fix.

### DIFF
--- a/server/config/runtime.exs
+++ b/server/config/runtime.exs
@@ -264,7 +264,7 @@ if Enum.member?([:prod, :stag, :can], env) do
     queue_interval: Tuist.Environment.clickhouse_queue_interval(secrets),
     settings: [
       readonly: 1,
-      max_threads: Tuist.Environment.clickhouse_max_threads(secrets),
+      max_threads: Tuist.Environment.clickhouse_read_max_threads(secrets),
       # Per-query memory ceiling so one heavy read fails on its own with a
       # `(for query)` error (retryable) rather than driving the process to its
       # `(total)` server ceiling and killing unrelated queries. Read path only:
@@ -283,14 +283,13 @@ if Enum.member?([:prod, :stag, :can], env) do
 
   config :tuist, Tuist.IngestRepo,
     url: Tuist.Environment.clickhouse_url(secrets),
-    pool_size: Tuist.Environment.clickhouse_pool_size(secrets),
+    pool_size: Tuist.Environment.clickhouse_buffer_pool_size(secrets),
     queue_target: Tuist.Environment.clickhouse_queue_target(secrets),
     queue_interval: Tuist.Environment.clickhouse_queue_interval(secrets),
     flush_interval_ms: Tuist.Environment.clickhouse_flush_interval_ms(secrets),
     max_buffer_size: Tuist.Environment.clickhouse_max_buffer_size(secrets),
-    pool_size: Tuist.Environment.clickhouse_buffer_pool_size(secrets),
     settings: [
-      max_threads: Tuist.Environment.clickhouse_max_threads(secrets)
+      max_threads: Tuist.Environment.clickhouse_write_max_threads(secrets)
     ],
     transport_opts: [
       keepalive: true,
--- a/server/priv/ingest_repo/migrations/20260706121835_add_id_projection_to_command_events.exs
+++ b/server/priv/ingest_repo/migrations/20260706121835_add_id_projection_to_command_events.exs
@@ -1,0 +1,39 @@
+defmodule Tuist.IngestRepo.Migrations.AddIdProjectionToCommandEvents do
+  @moduledoc """
+  Adds an id-ordered projection to `command_events`.
+
+  Production slow-query telemetry showed `get_command_event_by_id/1` issuing
+  `WHERE id = ?` lookups against `command_events` thousands of times in a
+  30-minute window. The table is ordered by `(project_id, name, ran_at)`, so an
+  id-only predicate cannot use the primary key and was reading ~1.17M rows per
+  lookup even with the existing `idx_id` bloom-filter skip index. An id-ordered
+  projection gives ClickHouse a physical layout where those lookups can read in
+  id order instead of scanning many wide granules.
+
+  `ADD PROJECTION` is metadata-only. Materialization happens in the follow-up
+  migration so the metadata change can propagate before the slower part rewrite
+  runs.
+  """
+
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  def up do
+    # excellent_migrations:safety-assured-for-next-line raw_sql_executed
+    execute """
+    ALTER TABLE command_events
+    ADD PROJECTION IF NOT EXISTS proj_by_id (
+      SELECT *
+      ORDER BY id
+    )
+    SETTINGS alter_sync = 2
+    """
+  end
+
+  def down do
+    # excellent_migrations:safety-assured-for-next-line raw_sql_executed
+    execute "ALTER TABLE command_events DROP PROJECTION IF EXISTS proj_by_id"
+  end
+end
--- a/infra/helm/tuist/templates/_helpers.tpl
+++ b/infra/helm/tuist/templates/_helpers.tpl
@@ -529,6 +529,22 @@ own way).
 {{- end -}}
 
 {{/*
+ClickHouse repo pool sizes are non-secret operational knobs. Render them from
+chart values so the server, migration, processor, and xcresult-processor pods
+stay aligned without relying on the runtime secret bundle.
+*/}}
+{{- define "tuist.clickhousePoolEnv" -}}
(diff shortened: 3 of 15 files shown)


server/config/runtime.exs
```diff
@@ -283,14 +283,13 @@ if Enum.member?([:prod, :stag, :can], env) do
 
   config :tuist, Tuist.IngestRepo,
     url: Tuist.Environment.clickhouse_url(secrets),
-    pool_size: Tuist.Environment.clickhouse_pool_size(secrets),
+    pool_size: Tuist.Environment.clickhouse_buffer_pool_size(secrets),
```
esnunes, line 286: One thing I wanted to flag — I think this might be a bit more than a duplicate-key cleanup, so it could be worth a line in the PR body.

I couldn't find `TUIST_CLICKHOUSE_POOL_SIZE` or `TUIST_CLICKHOUSE_BUFFER_POOL_SIZE` set anywhere in `infra/` (and they don't seem to be in the non-encrypted config) — if they aren't coming from the encrypted secrets either, then `clickhouse_pool_size` would resolve to `database_pool_size` and `clickhouse_buffer_pool_size` to its default of 5. From the history, the duplicate `pool_size` key also looks to have been inert since #9941 (`Keyword.get` takes the first occurrence), so `IngestRepo` has effectively been running at `database_pool_size` since then.

If that's right, the net effect on the ingest write pool would be around prod 15 → 5, canary 30 → 5, staging 15 → 5 (going by the `TUIST_DATABASE_POOL_SIZE` values in the managed values files). That's probably the value you're after anyway — a small batched-insert buffer pool fits the contention-reduction goal — but since it'd be a fairly quiet 3–6× drop, might be worth a quick double-check against the secrets and half an eye on ingest flush/backpressure after rollout. Definitely not a blocker.

server/priv/ingest_repo/migrations/20260706121835_add_id_projection_to_command_events.exs
```diff
@@ -0,0 +1,39 @@
...
+  use Ecto.Migration
+
+  @disable_ddl_transaction true
+  @disable_migration_lock true
+
+  def up do
+    # excellent_migrations:safety-assured-for-next-line raw_sql_executed
+    execute """
+    ALTER TABLE command_events
+    ADD PROJECTION IF NOT EXISTS proj_by_id (
```
esnunes, line 27: Just a heads-up, nothing to change here — since `proj_by_id` is `SELECT * ORDER BY id`, it effectively keeps a full id-ordered second copy of `command_events` (the highest-volume table, I think) and maintains it on every insert. That's the natural shape for a by-id row fetch and there's precedent with `test_case_runs`, so this is probably totally fine — might just be worth confirming the extra storage + insert-time cost feels acceptable and glancing at disk usage / ingest latency once the MATERIALIZE mutation finishes in the background.

esnunes (approved):
Really nice, well-scoped CPU-wait fix — thanks for the detailed write-up, it made this easy to follow. I walked through the flaky rolling-window rewrite and it looks equivalent to the old ARRAY JOIN + GROUP BY + LIMIT BY shape to me: both seem to normalize the two aggregate encodings to (run_key_micros, flag), dedup by run_key keeping max(flag), take the latest `size` distinct runs, and compute the same rate — the array form just skips the row fan-out. The existing rolling tests across bucket boundaries (2/5/75/247/751) give me some extra confidence there. The projection migrations look right too, following the same split add/materialize pattern as test_case_runs and leaving off ON CLUSTER like the other ingest migrations.

Left a couple of notes inline — nothing blocking, mostly one thing on `pool_size` that might be worth a mention in the PR body.

server/config/runtime.exs
fortmarek: Good catch, and I corrected this rather than accepting the quiet pool-size drop.

The PR now preserves the managed ClickHouse repo pool sizes and moves them to chart-owned config: `clickhouse.poolSize` / `clickhouse.bufferPoolSize` are set to `15` in production/staging and `30` in canary. The Helm templates render those env vars into the server, migration, processor, and xcresult-processor pods, so the value does not depend on the runtime secret bundle.

I also changed the environment helpers so ClickHouse pool sizes are read from env/defaults instead of ClickHouse pool entries in secrets.

server/priv/ingest_repo/migrations/20260706121835_add_id_projection_to_command_events.exs
fortmarek: Yep, agreed. I queried production ClickHouse through Atlas and added this to the PR body: `command_events` is currently ~26.8M rows / ~5.96 GB compressed, so `proj_by_id` should add storage in that same order of magnitude and each insert will pay projection-maintenance cost.

Given the hot by-id query was doing 16,008 executions in the 30-minute alert window and reading ~350 GiB, that tradeoff still feels worth it, but I added the rollout note to watch ClickHouse disk usage and insert latency after the materialization finishes.

PR #11682, opened 2026-07-06
feat(server): add runner interactive access dashboard

fortmarek:
## What Changed

This implements the macOS VNC slice of Hive spec 70 across tart-kubelet, Tuist Server, and the runner dashboard:

- Tart-backed macOS runner Pods launch with host-owned Tart VNC enabled, while the node-side relay stays closed unless tart-kubelet sees either the legacy host-local request file or the server-owned session annotation.
- Tart VNC endpoint/password parsing is redacted from VM logs and kept inside tart-kubelet state; workflow code, Pod annotations, runner dispatch, the browser, and any future CLI surface do not receive those credentials.
- Added `runner_interactive_sessions`, `Tuist.Runners.InteractiveSession`, and `Tuist.Runners.InteractiveSessions` for token-scoped, time-boxed interactive access grants.
- Added the `:runners_interactive` feature flag, project-write style authorization through account membership, hashed session tokens, pod-stop cleanup, and a scheduled TTL expiry worker.
- Added an Interactive tab to the runner job dashboard with Start VNC control, browser-rendered noVNC access, fullscreen support, and VNC viewport sizing based on the remote framebuffer.
- Updated gettext templates and `server/data-export.md` for the new stored access metadata.
- Refreshed the locked `hpax` entry because Mix upgraded it while resolving the current server dependency graph and would not run tasks against the stale lock entry.

## Why

Spec 70 introduces interactive runner access as a brokered debugging surface: VNC for live macOS Tart desktops first, shell later, with Tuist Server staying the authorization/lifecycle control plane and a dedicated gateway owning long-lived byte transport.

The important security boundary is credential custody. The browser-facing dashboard can request a server session, but it must not receive Tart VNC passwords, relay credentials, node endpoint metadata, or any direct node access path. This PR keeps that shape: the dashboard creates control-plane session state, while the actual VNC relay remains host-owned and closed by default.

## Approach

The implementation is split into two layers:

- **Host relay layer:** tart-kubelet enables Tart's experimental VNC path for runner VMs, parses the generated URL, redacts the password, and binds a source-CIDR-filtered relay only while either a legacy host-local request file exists or the server-owned session annotation is present.
- **Server/dashboard layer:** the dashboard records a requested VNC session for a running macOS job, persists only the token hash and lifecycle metadata, proxies browser WebSocket traffic to the runner relay through noVNC, closes open sessions on pod exit, and expires stale grants through Oban.

This PR does not add a Tuist CLI command for interactive access. The browser path is the first shipping client because it can keep VNC transport behind Tuist Server's authenticated WebSocket boundary while avoiding direct node credentials or local VNC viewer setup. A CLI command can be added later as a thin session opener or gateway client without changing the host credential-custody model.

This is feature work rather than a bugfix, so there was no existing buggy behavior being corrected.

## Impact

Account members with runner access can see an Interactive tab on running macOS runner jobs and start a VNC session from the dashboard. Existing runner dispatch, claim, execution, and billing-session lifecycle remain unchanged.

## Validation

- `mix gettext.extract`
- `mix test test/tuist/runners/interactive_sessions_test.exs test/tuist/oban/runtime_config_test.exs test/tuist/authorization_test.exs test/tuist_web/controllers/runner_pods_controller_test.exs test/tuist_web/live/runner_job_live_test.exs` (165 tests, 0 failures)
- `go test ./...` from `infra/tart-kubelet`
- `git diff --check`

## Screenshot

Runner interactive access screenshot

Marketing changelog asset: [`server/priv/static/marketing/images/changelog/2026.07.08-runner-interactive-access.jpg`](https://github.com/tuist/tuist/blob/a6cfab3089b1ead7eafd5ab548b0a4bc4a3fe889/server/priv/static/marketing/images/changelog/2026.07.08-runner-interactive-access.jpg)


server/assets/app/js/RunnerVNCClient.js
```diff
@@ -0,0 +1,207 @@
...
+
+    this.el.dataset.connection = "connecting";
+    this.rfb = new RFB(this.el, websocketURL(path));
+    this.rfb.background = "transparent";
+    this.rfb.scaleViewport = true;
+    this.rfb.resizeSession = false;
+    this.rfb.clipViewport = false;
+    this.rfb.dragViewport = false;
+    this.rfb.focusOnClick = true;
+    this.rfb.viewOnly = false;
```
pepicrft, line 117: This grants write-capable desktop access to the same users who only have `:runners_read`. The client sets Virtual Network Computing to interactive mode with `viewOnly = false`, while the server gates session requests through the read permission. A runner desktop can expose repository contents, job credentials, and deployment context, so this needs a distinct interactive-control authorization check, or regular reader access should be forced into view-only mode.

server/lib/tuist_web/controllers/runner_interactive_vnc_controller.ex
```diff
@@ -0,0 +1,28 @@
+defmodule TuistWeb.RunnerInteractiveVNCController do
+  use TuistWeb, :controller
+
+  import Plug.Conn
+
+  alias Tuist.Runners.InteractiveSessions
+  alias TuistWeb.RunnerVNCWebSock
+
+  def connect(conn, %{"token" => token}) do
+    with {:ok, session} <- InteractiveSessions.validate_token(token),
```
pepicrft: Token validation is not enough here. This route is scoped by account handle, but the upgrade only calls `validate_token/1`, whose query checks token hash, expiry, and closed state. It does not verify that the session belongs to the route account, that the current user can access that account, or that the token belongs to the requester. A copied token can therefore be used by any authenticated user. Please bind the lookup to the selected account and authorization result, and ideally to `requested_by_user_id` unless sharing sessions is intentional.

server/lib/tuist_web/live/runner_job_live.ex
```diff
@@ -468,4 +585,58 @@ defmodule TuistWeb.RunnerJobLive do
...
+    vnc_dev_placeholder? = Environment.dev?() and vnc_requestable?
+
+    vnc_session =
+      selected_account.id
+      |> InteractiveSessions.current_for_job(job.workflow_job_id, :vnc)
+      |> with_vnc_session_token(vnc_session_token)
+
+    vnc_websocket_path =
+      if not vnc_dev_placeholder? and vnc_session_ready?(vnc_session) and is_binary(vnc_session_token) do
+        "/#{selected_account.name}/runners/interactive/vnc/#{vnc_session_token}"
```
pepicrft: Putting the bearer token in the request path makes the desktop access secret show up in browser history, request telemetry, and structured request path metadata. This scope also runs the analytics plug, so the raw token is emitted as a page-view path. Given that the WebSocket upgrade currently accepts any valid token from any authenticated user, this is a direct session hijack path. Please move the secret out of tracked paths or redact this route before analytics and logging, then consume or bind the token on upgrade.

infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
```diff
@@ -987,12 +1000,27 @@ func (r *ScalewayAppleSiliconMachineReconciler) reconcileTailscaleEgressService(
 		svc.Spec.Ports = []corev1.ServicePort{
 			{Name: "node-exporter", Port: 9100, Protocol: corev1.ProtocolTCP},
 			{Name: "tart-kubelet", Port: 8080, Protocol: corev1.ProtocolTCP},
+			{Name: "vnc-relay", Port: DashboardVNCRelayPort, Protocol: corev1.ProtocolTCP},
```
pepicrft, line 1003: Exposing this service port makes the no-auth relay reachable through the normal cluster and tailnet path. The Tart-side Virtual Network Computing endpoint is authenticated, but `NewVNCForwarder` strips that authentication and presents Remote Framebuffer no-auth to clients. If any workload or tailnet identity can reach this service while a session is open, it can bypass the server token and audit path. Please restrict this port to only the server path or use a dedicated narrow allowlist for the interactive relay.

server/assets/app/js/RunnerVNCClient.js
fortmarek: I disagree on this one. I think runners read means being able to interact with them. Sure, you can read source code, but I think it's very unlikely members of the Tuist org wouldn't be able to read the source code anyways.

fortmarek: Kept this under `:runners_read` intentionally after discussing the permission model. For this feature, runner read means being able to inspect and debug the scoped job VM; runner write is reserved for mutating runner/fleet configuration. To tighten the access path without changing that model, 6510b35 binds VNC tokens to the account and requesting user, removes the token from the URL, and adds relay-local token authentication so an opened desktop session cannot be reused outside the authorized dashboard session.

server/lib/tuist_web/controllers/runner_interactive_vnc_controller.ex
fortmarek: Addressed in 6510b35. The WebSocket controller now resolves the `account_handle`, requires a current user, authorizes `:runners_read` for that account, and validates the token with the same account and user via `InteractiveSessions.validate_token/3`. That validation checks token hash, account id, requester id, expiry, and closed state before allowing the upgrade. I also added controller/context coverage for the valid upgrade path plus cross-user/account rejection.

server/lib/tuist_web/live/runner_job_live.ex
fortmarek: Addressed in 6510b35. The WebSocket route is now stable (`/:account_handle/runners/interactive/vnc`) and the browser token is no longer part of the request path. The LiveView renders the token separately, and the noVNC client sends it through `wsProtocols`, which the controller reads from `sec-websocket-protocol`. The LiveView test now asserts we do not render the old tokenized `/runners/interactive/vnc/<token>` path.

infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
fortmarek: Addressed in 6510b35 by authenticating the dashboard relay path in `tart-kubelet`, rather than relying on the service port being hard to reach. Dashboard pod-annotation requests now require both `tuist.dev/vnc-session-id` and `tuist.dev/vnc-relay-token-hash`; the server writes only the hash and clears it when the session closes. The forwarder now requires the server bridge to send a `tuist-vnc-token <token>` preface, hashes it, compares it with the annotation hash before dialing Tart, and rejects unauthenticated clients. The legacy host-control-file operator path is unchanged. Added podagent tests for the relay auth behavior.

pepicrft (approved):
Thanks for tightening this. I still think it may be worth introducing a dedicated authorization action such as `:runners_interact` or `:runners_debug`, even if it grants to the same roles for now, so the active-control boundary is explicit. I am not too strong on that given the current product model, and the token, account, requester, and relay hardening addresses the concrete security issues I was worried about.

fortmarek:
> I still think it may be worth introducing a dedicated authorization action such as :runners_interact or :runners_debug, even if it grants to the same roles for now, so the active-control boundary is explicit.

I don't think this distinction brings much as I would initially anyways give that permission to all members regardless, so I'll skip this for now.

PR #11683, opened 2026-07-06
fix(server): authenticate CLDR locale prefetch against the GitHub API

esnunes:
## What changed

The CLDR locale prefetch helper added in #11678 (`server/mise/tasks/cldr/prefetch-locale-data.sh`) now authenticates against the GitHub REST API when a token is available, on both the CI and Docker build paths:

- **Script** — sends `Authorization: Bearer $GITHUB_TOKEN` (falling back to `GH_TOKEN`) only when a token is present in the environment; otherwise it stays unauthenticated so local/dev builds keep working offline. The header is applied through an empty-array-safe expansion (`${auth_header[@]+"${auth_header[@]}"}`) so it survives `set -u` on Bash 3.2 (macOS).
- **CI** — the `Test` job's prefetch step already inherits the workflow-level `GITHUB_TOKEN`, so no workflow change was needed there.
- **Docker** — the `RUN` in `server/Dockerfile` mounts the token as a BuildKit secret (`--mount=type=secret,id=github_token,env=GITHUB_TOKEN`) so it never lands in an image layer. The `github_token` secret is wired into every workflow that builds `server/Dockerfile`: the CI `docker_build` (`--secret` on the buildx CLI), the prod amd64/arm64 release builds, the main-push server image, the single-env `server-deployment`, and `preview-deploy`.

## Why

`prefetch-locale-data.sh` fetched all 12 locales from `api.github.com` with **no** `Authorization` header, so it was bound by the REST API's ~60 req/hr unauthenticated ceiling per egress IP. That's stricter than — and in a different rate-limit bucket from — the `raw.githubusercontent.com` compile-time downloads it replaced (the 429s #11678 was fixing).

On the shared-egress runners, a release cascade plus concurrent PR cache-misses can breach 60/hr, and the failure mode is a hard build break: GitHub 403 → `curl --retry-all-errors` dutifully retries the 403 → `set -e` fails the step. So the prefetch, as merged, swapped a probabilistic `raw` 429 for a probabilistic REST 403 — it narrowed the fragility but didn't remove it.

`GITHUB_TOKEN` was already present in the CI step's env (just unused), and the earlier "token variant 403'd" during development reads like an *empty*-token auth failure inside the Docker build rather than a rate limit — authenticated requests get 5000/hr, which comfortably covers a release cascade. So the fix is to authenticate, not to drop auth.

## Why this shape

- The BuildKit secret keeps the token out of image layers (vs. a `--build-arg`, which would bake it into build history). Both build paths already run under BuildKit, so this is plumbing, not an architecture change.
- The secret mount is **optional**: builds that don't pass `github_token` (and local runs with no token) still succeed unauthenticated, exactly as today — so this can't regress any existing path.

## Impact

Server release Docker builds and the full-locale CI test compile become resilient to GitHub REST rate limits during release cascades and busy CI windows, instead of hard-failing on a 403. No behavior change when no token is present.

## Validation

- Ran `prefetch-locale-data.sh` end-to-end against the **real** GitHub API both unauthenticated and with a token — all 12 locale JSON files fetched and parsed as valid JSON both ways.
- Verified the empty-array expansion under `set -euo pipefail` on Bash 3.2 (macOS system bash) with and without a token, so the local/dev path doesn't break on the `unbound variable` trap.
- `bash -n` on the script.

### DIFF
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -342,6 +342,8 @@ jobs:
           target: runner
           push: true
           tags: ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ steps.tag.outputs.image_tag }}
+          secrets: |
+            github_token=${{ secrets.GITHUB_TOKEN }}
           build-args: |
             TUIST_HOSTED=1
             MIX_ENV=prod
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -207,6 +207,8 @@ jobs:
           push: true
           tags: |
             ${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ steps.tag.outputs.image_tag }}
+          secrets: |
+            github_token=${{ secrets.GITHUB_TOKEN }}
           # Shared `buildcache:server` ref, kept warm by the push-to-main
           # build in server-production-deployment.yml. `mode=max` exports
           # cache for every stage (deps.compile, builder, og-images).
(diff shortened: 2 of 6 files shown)


fortmarek (approved):
Thanks!

PR #11688, opened 2026-07-07
fix(server): handle concurrent SCIM user creation

fortmarek:
## What changed

- Handles `{:error, :email_taken}` from `Accounts.create_user/2` while provisioning a SCIM user.
- Re-fetches the user after that conflict and treats it as an existing user when the account was created concurrently.
- Adds a SCIM-specific Phoenix error renderer for the `application/scim+json` media type.
- Tags `/scim/v2` requests early in the endpoint so fallback error rendering uses a SCIM Error resource.
- Adds regression coverage for the provisioning race and for SCIM-shaped fallback error responses.

## Why

Okta SCIM provisioning can race with SSO account creation when SSO and SCIM are configured as separate Okta apps. In that flow, the SCIM request can check for a user, find none, then lose the create race because the SSO path created the user first.

The reported failure surfaced to Okta as `Bad Gateway` plus a non-JSON body that started with plain text. Okta then failed again while trying to parse the remote error response as JSON. Even when the root race is fixed, SCIM endpoints should return parseable SCIM error resources for fallback failures so the identity provider receives something it understands.

## Root cause

`SCIM.provisionable_user/1` expected `Accounts.create_user/2` failures to be changesets. `Accounts.create_user/2` normalizes unique email conflicts to `{:error, :email_taken}`, so the SCIM code passed an atom into `email_taken?/1`, which only accepts an `Ecto.Changeset`. That could raise before the SCIM controller had a chance to return a structured SCIM response. A retry then worked because the user already existed.

Phoenix also only had `html` and `json` fallback error renderers configured globally. Known SCIM controller errors already used `send_scim_error/4`, but exceptions, unmatched SCIM routes, and parser-level failures could bypass those helpers and fall back to a non-SCIM response shape.

## Approach

The provisioning fix keeps the existing SCIM behavior but makes the create path race-tolerant:

- If user creation succeeds, continue with the newly created user.
- If user creation returns `:email_taken`, re-read by email.
- If the re-read finds a user, attach that existing user to the organization.
- If the re-read still cannot find a user, preserve the existing `:email_taken` error path.
- If another non-changeset error appears, return it instead of crashing through the changeset-only helper.

The response-shape fix adds `TuistWeb.SCIM.ErrorJSON`, configured for `scim+json`, and a narrow endpoint plug that marks `/scim/v2` requests with the SCIM format before later fallback error rendering. That keeps ordinary web/API error rendering unchanged while ensuring SCIM fallbacks produce bodies like:

```json
{
  "schemas": ["urn:ietf:params:scim:api:messages:2.0:Error"],
  "status": "500",
  "detail": "Internal Server Error"
}
```

Malformed JSON requests that negotiate `application/scim+json` are returned as `400` with `scimType: invalidSyntax`.

## Alternatives considered

- Keep the separate Okta SSO and SCIM apps: still a valid operational setup, especially because current Okta custom OIDC integrations do not expose SCIM provisioning in the same app. This PR makes that setup reliable by handling the expected race and returning SCIM-shaped fallback errors.
- Move to a single Okta app through SAML + SCIM: possible if Tuist adds SAML SSO support. Okta custom SAML apps can enable SCIM provisioning in the same integration, but that is a larger product surface area than this bug fix.
- Submit a combined Okta Integration Network app: a better long-term distribution path for a polished OIDC/SAML + SCIM app, but it requires Okta review and does not remove the need for server-side race handling and SCIM error responses.
- Use a database upsert or transaction around lookup/create/attach: heavier than necessary here and would require touching account creation semantics shared outside SCIM. Re-reading on `:email_taken` is narrower and matches the existing `provisionable_user/1` control flow.
- Catch the exception only at the controller boundary: would hide this specific crash but still fail the first provisioning attempt, and it would not cover parser/no-route fallback errors.

## Impact

SCIM user creation becomes idempotent across the common SSO/provisioning race. First-attempt provisioning should attach the account that was just created by the SSO path instead of returning a transient gateway-style failure.

When a SCIM fallback error does occur, the response is now parseable as a SCIM Error resource with `application/scim+json`, so Okta should report the actual status/detail instead of failing to parse a plain-text upstream error.

## Validation

- `git diff --check HEAD~1..HEAD` for the SCIM error-rendering commit.
- `mix run --no-start -e 'Mix.Tasks.Test.run(["test/tuist_web/controllers/scim/error_json_test.exs", "test/tuist_web/controllers/scim/error_rendering_test.exs"])'` passed with `4 tests, 0 failures`.
- Earlier targeted SCIM controller/context validation passed with `44 tests, 0 failures` after the race fix.
- A later full targeted SCIM rerun could not complete in this worktree because ClickHouse is unavailable locally (`Mint.TransportError connection refused` / `Tuist.IngestRepo` queue timeout).

### DIFF
--- a/server/lib/tuist/scim.ex
+++ b/server/lib/tuist/scim.ex
@@ -291,7 +291,9 @@ defmodule Tuist.SCIM do
     |> Oban.insert!()
   end
 
-  defp provisionable_user(email) do
+  defp provisionable_user(email, opts \\ []) do
+    retry_after_email_taken = Keyword.get(opts, :retry_after_email_taken, true)
+
     case Accounts.get_user_by_email(email) do
       {:ok, %User{} = user} ->
         {:ok, user, :existing}
@@ -301,8 +303,11 @@ defmodule Tuist.SCIM do
           {:ok, user} ->
             {:ok, user, :created}
 
+          {:error, :email_taken} when retry_after_email_taken ->
+            provisionable_user(email, retry_after_email_taken: false)
+
           {:error, :email_taken} ->
-            provisionable_user_created_concurrently(email)
+            {:error, :email_taken}
 
           {:error, %Ecto.Changeset{} = changeset} ->
             if email_taken?(changeset), do: {:error, :email_taken}, else: {:error, changeset}
@@ -310,13 +315,6 @@ defmodule Tuist.SCIM do
           {:error, reason} ->
             {:error, reason}
         end
-    end
-  end
-
-  defp provisionable_user_created_concurrently(email) do
-    case Accounts.get_user_by_email(email) do
-      {:ok, %User{} = user} -> {:ok, user, :existing}
-      {:error, :not_found} -> {:error, :email_taken}
     end
   end
 
--- a/server/lib/tuist/scim.ex
+++ b/server/lib/tuist/scim.ex
@@ -301,9 +301,22 @@ defmodule Tuist.SCIM do
           {:ok, user} ->
             {:ok, user, :created}
 
-          {:error, changeset} ->
+          {:error, :email_taken} ->
+            provisionable_user_created_concurrently(email)
+
+          {:error, %Ecto.Changeset{} = changeset} ->
             if email_taken?(changeset), do: {:error, :email_taken}, else: {:error, changeset}
+
+          {:error, reason} ->
+            {:error, reason}
         end
+    end
+  end
+
+  defp provisionable_user_created_concurrently(email) do
+    case Accounts.get_user_by_email(email) do
+      {:ok, %User{} = user} -> {:ok, user, :existing}
+      {:error, :not_found} -> {:error, :email_taken}
     end
   end
 
--- a/server/lib/tuist_web/controllers/scim/error_json.ex
+++ b/server/lib/tuist_web/controllers/scim/error_json.ex
@@ -1,0 +1,31 @@
+defmodule TuistWeb.SCIM.ErrorJSON do
+  @moduledoc """
+  Renders Phoenix error responses as SCIM 2.0 Error resources.
+  """
+
+  alias Plug.Parsers
+  alias Tuist.SCIM.Resource
+
+  def render(template, assigns) do
+    status = Map.get(assigns, :status) || status_from_template(template)
+    {detail, scim_type} = error_detail(status, Map.get(assigns, :reason))
+
+    status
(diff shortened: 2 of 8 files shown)


server/lib/tuist_web/controllers/scim/error_json.ex
```diff
@@ -0,0 +1,31 @@
...
+    |> Resource.render_error(detail, scim_type)
+    |> JSON.encode!()
+  end
+
+  defp status_from_template(template) do
+    template
+    |> String.split(".")
+    |> hd()
+    |> String.to_integer()
+  rescue
```
pepicrft: This will mask errors. Do we need to be defensive?

server/lib/tuist/scim.ex
```diff
@@ -301,12 +301,25 @@ defmodule Tuist.SCIM do
...
         end
     end
   end
 
+  defp provisionable_user_created_concurrently(email) do
+    case Accounts.get_user_by_email(email) do
+      {:ok, %User{} = user} -> {:ok, user, :existing}
+      {:error, :not_found} -> {:error, :email_taken}
+    end
+  end
```
pepicrft: mmm decorating `get_user_by_email` to map the response format is a bit smelly, specially with a function name `created_concurrently` that leaks context of the issue you are trying to fix. I'd nudge the coding agent to take a different approach here.

pepicrft (approved):

server/lib/tuist_web/controllers/scim/error_json.ex
fortmarek: Addressed in `a3285c35b68`: I removed the defensive template parsing/rescue. `TuistWeb.SCIM.ErrorJSON.render/2` now only matches when Phoenix provides an integer `:status`, so a bad/missing status fails instead of silently being converted to a 500. The SCIM error response tests still cover the expected 400/500 payloads.

server/lib/tuist/scim.ex
fortmarek: Addressed in `a3285c35b68`: I removed `provisionable_user_created_concurrently/1` and folded the conflict handling back into `provisionable_user/2` as a neutral one-shot retry. When `create_user/2` returns `:email_taken`, it re-enters the normal lookup path once; if the user still is not found, it returns `:email_taken`. I also renamed the regression test to describe the behavior rather than the incident context.

PR #11695, opened 2026-07-07
ci(infra): add K3s Helm smoke check

pepicrft:
Resolves no linked issue

## What changed

This adds a [K3s](https://k3s.io/) smoke check for the Tuist Helm chart so chart changes are exercised against a real single-node K3s control plane before they merge.

The check runs in the Helm workflow and uses the infra-scoped mise tools from `infra/mise.toml`: [Helm](https://helm.sh/), [kubectl](https://kubernetes.io/docs/reference/kubectl/), and [k3d](https://k3d.io/). The local task lives at `infra/mise/tasks/helm/k3s-smoke.sh`, uses mise `#USAGE` annotations for `--render-only`, creates a disposable k3d cluster, lints and renders the chart with K3s smoke values, runs a server-side dry run, installs the chart, and waits for PostgreSQL, ClickHouse, and MinIO to become ready.

## Why

A self-hosted deployment can run on K3s, but we did not have automation that proved the chart still installs on that Kubernetes distribution. This gives us a repeatable check for the single-node shape without relying on a long-lived cluster.

## Approach

The smoke profile keeps the server Deployment rendered but scales it to zero because booting the production image requires a Tuist license. That keeps the check focused on chart and dependency compatibility without adding license material to automation.

Practically, the smoke test verifies that Helm can lint and render the chart, the K3s control plane accepts the manifests through server-side dry run, Helm can install them into a fresh cluster, persistent volume claims bind with K3s `local-path` storage, the embedded PostgreSQL, ClickHouse, and MinIO StatefulSets become ready, and the ClickHouse and MinIO initialization jobs complete.

It deliberately does not prove that the Tuist server image boots, migrations run, cache behavior works, license handling works, requests succeed end to end, or an Azure air-gapped network shape is correct.

## Impact

Chart changes now get a dedicated K3s compatibility signal in the Helm workflow. Developers can run the same check locally from the infra mise configuration.

### How to test locally

- `mise -C infra install helm kubectl k3d`
- `mise -C infra run helm:k3s-smoke --help`
- `mise -C infra tasks validate helm:k3s-smoke`
- `mise -C infra run helm:k3s-smoke --render-only`
- `mise -C infra run helm:k3s-smoke`
- `/Users/pepicrft/.local/share/mise/installs/actionlint/1.7.7/actionlint .github/workflows/helm.yml`

The full local run used K3s `v1.35.5+k3s1`, confirmed PostgreSQL, ClickHouse, and MinIO reached `Running`, confirmed both initialization jobs completed, confirmed all persistent volume claims bound with K3s `local-path` storage, and removed the disposable cluster afterward.


.github/workflows/helm.yml
```diff
@@ -183,3 +187,22 @@ jobs:
...
             exit 1
           fi
+
+  k3s:
+    name: K3s Smoke Install
+    runs-on: ubuntu-latest
+    timeout-minutes: 25
+    needs:
+      - lint
+      - render
```
fortmarek: why do we wait on this? This worsens the wall-clock time

fortmarek:
Review findings:

- [P2] `.github/workflows/helm.yml`: the new K3s job runs `infra/mise/tasks/helm/k3s-smoke.sh`, but the workflow `push`/`pull_request` path filters still ignore `infra/mise/tasks/**`, `infra/mise.toml`, and `infra/mise.lock`. A PR that changes only the smoke task or its tool pins would not run the smoke install it just changed. Please include those paths in the Helm workflow triggers so the check is self-validating.

- [P3] `infra/mise/tasks/helm/k3s-smoke.sh`: if the script fails before it writes the k3d kubeconfig, the diagnostic trap can run `kubectl get/describe` against the ambient current context. That is read-only, but surprising, especially for anyone with a real K3s context selected. Consider gating diagnostics until the task has selected the intended context, or only running them after `KUBECONFIG` is set / `TUIST_HELM_K3S_USE_CURRENT_CONTEXT=1` is explicitly opted into.

fortmarek (approved):

.github/workflows/helm.yml
pepicrft: Removed the `needs` dependency from `K3s Smoke Install`, so it starts independently from Helm lint and render. The smoke task still does its own lint, render, server-side dry run, and install, so it does not need to wait on those jobs.

.github/workflows/helm.yml
```diff
@@ -183,3 +187,22 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: jdx/mise-action@v4.0.1
+        with:
+          version: ${{ env.MISE_VERSION }}
+          install: "false"
+      - name: Install smoke test tools
+        run: mise -C infra install helm kubectl k3d
+      - name: Smoke test Tuist chart on K3s
+        run: mise -C infra run helm:k3s-smoke
```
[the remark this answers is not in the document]
pepicrft: Added `infra/mise.toml`, `infra/mise.lock`, and `infra/mise/tasks/helm/**` to the Helm workflow path filters for both push and pull_request. Changes to the smoke task or its infra tool pins now trigger the workflow.

PR #11697, opened 2026-07-07
feat(server): add Azure Blob storage support

fortmarek:
## What changed

Adds Azure Blob Storage as a first-class server artifact storage backend alongside the existing S3-compatible path.

- Added a new `Tuist.Storage.AzureBlob` adapter that uses Azure Blob REST APIs with Shared Key authentication, service SAS URLs, direct put/get/delete/list operations, and multipart-compatible block blob uploads.
- Routed `Tuist.Storage` through the configured provider while preserving account-level custom S3 storage behavior.
- Moved xcresult processor file upload through the storage abstraction instead of raw S3 multipart calls.
- Added Helm values and environment wiring for `server.storage.provider=azure_blob` and `server.azureBlob.*` across server, processor, xcresult processor, and migration workloads.
- Added CLI request handling for Azure direct Put Blob uploads by setting `x-ms-blob-type: BlockBlob` only for Azure Blob SAS upload URLs.
- Updated self-hosting and Helm docs, including the Azure-only chart shape where embedded MinIO can be skipped when S3-only optional workloads are disabled.

## Why it changed

Some self-hosted enterprise deployments run in Azure environments where Azure Blob Storage is the native object storage service, but Azure Blob is not S3-compatible. Previously those deployments needed an additional S3-compatible layer such as MinIO even for server-owned artifacts. This change lets the server store build archives, previews, test attachments, runner log archives, and related server artifacts directly in Azure Blob Storage.

## Approach

The implementation keeps the existing S3 path intact and adds provider selection at the storage boundary. S3 remains the default provider, and custom per-account S3 settings continue to force the S3 adapter so existing account-specific storage behavior does not change.

Azure multipart behavior is implemented with Azure block blobs: `multipart_start` creates an internal upload id, part URLs target `Put Block` with deterministic block ids, and completion commits the sorted block list. Direct server-side calls use Shared Key signed REST requests; client upload and download URLs use service SAS tokens.

The chart keeps top-level `objectStorage` S3-compatible because optional workloads such as cache and registry mirror still use S3 APIs. Azure settings live under `server.azureBlob`, and the docs describe how to set `objectStorage.mode=external` with empty external values when deploying an Azure-only server without embedded MinIO.

## Impact

Self-hosted deployments can now use Azure Blob Storage directly for server-owned artifacts by setting:

```yaml
server:
  storage:
    provider: azure_blob
  azureBlob:
    accountName: ...
    accountKey: ...
    containerName: ...
```

Existing S3-compatible deployments continue to use the default `server.storage.provider: s3` path.

## Validation

- `git diff --check`
- Swift focused test: `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistServerTests/AzureBlobUploadRequestTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- Azure adapter ExUnit smoke via `mix run --no-deps-check --no-start`, 7 tests passing
- Server test compilation via `MIX_ENV=test mix run --no-deps-check --no-start -e 'IO.puts(:ok)'`
- Helm renders:
  - default chart
  - Azure Blob chart
  - Azure Blob chart with processor and xcresult processor enabled
  - Azure-only chart with `objectStorage.mode=external`
- Local Azurite e2e smoke covering container creation, direct upload, object existence, string download, SAS download, SAS upload, multipart block upload, file upload/download, list, and delete.

## Notes

I did not update `server/mix.lock`. Normal `mix deps.get --check-locked` / standard Mix task entrypoints in this worktree currently report unrelated Hex resolution drift for `hpax`, `phoenix`, and `websock_adapter`, so validation used `--no-deps-check` where needed to avoid broadening this PR into dependency updates.


pepicrft (approved):
Thanks for adding the Azure Blob Storage path. I validated the request shape against Microsoft's Azure Blob Storage documentation: direct Put Blob uploads need `x-ms-blob-type`, multipart uses Put Block plus Put Block List, and listing uses `marker` plus `NextMarker`. The Helm chart also rendered cleanly in both default and Azure modes.

Approving to keep this moving, but I would still address these before treating Azure as production-ready:

1. Azure result attachment uploads currently do not match the shared storage return contract. `Tuist.Processor.XCResultProcessor.upload_attachment/5` only treats `{:ok, _}` as success, while the Azure `upload_file` path returns `:ok` after a successful block commit. That will make successful attachment uploads look like failures. I would normalize `Tuist.Storage.upload_file/4` so both providers return the same shape and cover it through `Tuist.Storage`, not just `Tuist.Storage.AzureBlob`.

2. Azure project cleanup can hide deletion failures. `Tuist.Storage.delete_all_objects/2` discards the provider result and returns `:ok` after telemetry. If the Azure list or delete fails, cleanup still appears successful.

3. Cache artifact retention is still wired to the [Amazon Simple Storage Service](https://aws.amazon.com/s3/) bucket names and the default provider path. Cache uploads now go through `Tuist.Storage.multipart_*`, so for normal accounts they can land in Azure, but the retention workers still inspect `TUIST_CACHE_S3_BUCKET_NAME` and do not pass `storage_provider: :azure_blob`.

4. Azure streaming should be checked for back pressure. The current implementation pushes chunks from a spawned task into the caller mailbox as fast as the download produces them. Large preview or build archive downloads can therefore accumulate in process memory instead of behaving like the existing [Amazon Simple Storage Service](https://aws.amazon.com/s3/)-compatible stream.

References I used while validating the storage calls:
- Put Blob: https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob
- Put Block: https://learn.microsoft.com/en-us/rest/api/storageservices/put-block
- Put Block List: https://learn.microsoft.com/en-us/rest/api/storageservices/put-block-list
- List Blobs: https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs

PR #11699, opened 2026-07-07
fix(server): unblock deploys from regional ingress webhooks

fortmarek:
## What changed

- Disabled admission webhooks on the regional Kura ingress-nginx chart aliases.
- Kept the main ingress-nginx admission webhook enabled so Ingress validation still runs through the primary controller.
- Updated `k8s:install-platform` to only require the main ingress admission secret when deciding whether platform hooks can be skipped.
- Hardened Dedibox adoption so the controller only adopts tagged servers whose install state is complete, skipping bare or actively reinstalling boxes.
- Changed Dedibox release deletion from a fire-and-forget reinstall to an observed lifecycle: the Machine finalizer stays until the release reinstall has been observed in progress and then complete.
- Added a bootstrap-side install-state gate so a claimed Dedibox server is not SSH-bootstrapped while Scaleway still reports its OS install as pending/running.

## Why

The server production deployment failed during canary at `helm upgrade --install` while patching the server Ingress. The Kubernetes API server attempted to call the regional Kura ingress-nginx admission webhook service and failed before Helm could finish the upgrade or rollback.

Failed run: [Server Production Deployment / canary deploy](https://github.com/tuist/tuist/actions/runs/28867435022/job/85635968337#step:11:246).

Regional Kura ingress-nginx controllers are still part of the data plane where regional fleets are present. The problem was the admission webhook surface, not the controller pods themselves: ingress-nginx admission webhooks are cluster-wide and are not scoped to only the IngressClass served by that controller.

The failing regional webhook had no endpoints because the canary Dedibox-backed regional pool temporarily had zero Ready nodes. That exposed a second issue: the Dedibox release path could hand a physical box back to the pool immediately after starting a reinstall, allowing a replacement Machine to re-adopt the same box while Scaleway was still reinstalling it.

## Root cause

In canary, `platform-kura-eu-central-ingress-nginx-controller-admission` existed as a ClusterIP Service with no endpoints because the corresponding regional DaemonSet had zero desired pods. Since that admission webhook was registered globally, unrelated Ingress updates such as the Tuist server Ingress were blocked on an empty regional admission Service.

The regional pool reached that empty state because the Dedibox Machine lifecycle treated release reinstall as asynchronous cleanup after the Machine was gone. The replacement controller could see the same tagged server as available before the reinstall had completed, then repeatedly try to bootstrap while the physical host was still transitioning through reinstall states.

## Impact

Server deploys no longer depend on every regional Kura ingress fleet having live admission endpoints. Regional Kura ingress-nginx dataplane pods continue to render and run normally; only their redundant global validating webhook registrations are removed.

Dedibox-backed Kura regions now avoid re-adopting boxes mid-reinstall and wait for a clean OS before bootstrap. This should prevent the canary Dedibox pool from collapsing to zero Ready nodes during a normal release/re-adoption cycle.

## Validation

- `bash -n infra/mise/tasks/k8s/install-platform.sh`
- `helm template` for canary, staging, and production platform overlays
- `helm lint` for canary, staging, and production platform overlays
- Confirmed rendered platform manifests keep the main ingress-nginx admission resources and omit regional Kura admission resources
- Read-only live checks confirmed the failing canary regional admission Service had no endpoints, while regional Kura ingress-nginx pods are still used where fleets are present
- `/Users/marekfort/.local/share/mise/installs/go/1.26.4/bin/go test ./internal/dedibox ./controllers/linux`
- `/Users/marekfort/.local/share/mise/installs/go/1.26.4/bin/go test ./...`
- `git diff --check`

### DIFF
--- a/infra/cluster-api-provider-tuist/api/v1alpha1/dediboxmachine_types.go
+++ b/infra/cluster-api-provider-tuist/api/v1alpha1/dediboxmachine_types.go
@@ -8,8 +8,8 @@ import (
 
 // DediboxMachineSpec is the desired state of one Scaleway Dedibox dedicated
 // server. Structurally it mirrors the OVH kind — a customer-facing public box
-// that adopts a pre-ordered server, installs on claim, self-joins over its
-// public IP, and (monthly contract) is not terminated on delete. It is driven
+// that adopts a pre-ordered, installed server, self-joins over its public IP,
+// and is reinstalled back to a clean, claimable state on delete. It is driven
 // through the Scaleway Dedibox API with a default-project IAM key (every Dedibox
 // in the org shares the default project — a Dedibox cannot be assigned to
 // another one — so the project can't isolate environments; a per-fleet tag
--- a/infra/cluster-api-provider-tuist/controllers/linux/dediboxmachine_controller.go
+++ b/infra/cluster-api-provider-tuist/controllers/linux/dediboxmachine_controller.go
@@ -48,6 +48,11 @@ const (
 	// dediboxInstanceType is the node.cluster.x-k8s.io/instance-type label value
 	// the self-join stamps.
 	dediboxInstanceType = "dedibox"
+
+	dediboxReleaseReinstallStartedAnnotation  = "dedibox.cluster.x-k8s.io/release-reinstall-started"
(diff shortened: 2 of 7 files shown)


pepicrft (approved):
Looks good to me.

PR #11700, opened 2026-07-07
fix(server): handle utf16 slf strings in build logs

fortmarek:
## What changed

- Added a local SLF lexer fallback for the `xcactivitylog` NIF parser.
- Added a loader that decodes gunzipped log bytes as UTF-8 data instead of C string data, preserving embedded NUL bytes.
- Wired the fallback so normal logs still use `XCLogParser.ActivityParser.parseActivityLogInURL`, and only parser failures retry through the local tokenization path before handing tokens back to `XCLogParser`.
- Added regression tests for UTF-16 string lengths and embedded NUL content in SLF strings.

## Why

Build processing can fail on valid `.xcactivitylog` files with errors like:

```text
The line Sources/Bundle+Locali doesn't seem like a valid SLF line
```

The upstream lexer scans SLF string payloads by UTF-16 offsets, but then rejects the scanned string when `String.count` does not match the encoded length. SLF string lengths are UTF-16 code units, while `String.count` counts grapheme clusters. A valid path or diagnostic containing multi-code-unit characters can therefore desynchronize tokenization and make the worker discard the build.

The C-string loader path can also truncate valid content at embedded NUL bytes. Decoding the decompressed bytes directly avoids that truncation.

## Impact

Valid Xcode activity logs with UTF-16-length string payloads or embedded NUL bytes should process instead of being discarded by `ProcessBuildWorker`. Existing successful logs continue to take the upstream parser path first.

## Validation

- `swift test --replace-scm-with-registry --cache-path "$tmp_cache" --manifest-cache local --filter XCActivityLogParserTests/slfLexer`
- `swift test --replace-scm-with-registry --disable-automatic-resolution --filter XCActivityLogParserTests/cleanBuild_parsesSuccessfully`
- `git diff --check`

### DIFF
--- a/server/native/xcactivitylog_nif/Package.resolved
+++ b/server/native/xcactivitylog_nif/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "eaf0120aada03126bb2d5c22eea625b8ba0565517b1041850b9ed2c70ce15868",
+  "originHash" : "a91f206daaf42a4ffc9f5e9fc32a810ad6c53ec1c96b93b780b6db3e814a40ec",
   "pins" : [
     {
       "identity" : "1024jp.gzipswift",
--- a/server/native/xcactivitylog_nif/Package.swift
+++ b/server/native/xcactivitylog_nif/Package.swift
@@ -20,6 +20,7 @@ let package = Package(
         ),
     ],
     dependencies: [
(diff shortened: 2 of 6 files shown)


fortmarek:
Closing this draft because the fix belongs upstream in XCLogParser. Replaced by https://github.com/MobileNativeFoundation/XCLogParser/pull/254.

PR #11704, opened 2026-07-07
feat(kura): record REAPI (gRPC) serving in kura_usage_events

### LINKED ISSUE #11703: Kura: REAPI (gRPC) serving is not recorded in kura_usage_events
## Summary

Kura's REAPI (gRPC) serving path is **not** recorded into the `kura_usage_events` ClickHouse table. As a result, `artifact_kind = 'reapi'` never appears in usage data, and all Bazel/REAPI cache bandwidth is invisible to usage/billing analytics.

## Root cause

Rows in `kura_usage_events` are only produced by `record_public_download` / `record_public_upload` (`kura/src/usage.rs`), and those are called from exactly two places, both on the **HTTP** serving path:

- `record_usage_event` — `kura/src/http.rs:2143`
- `record_usage` — `kura/src/accelerated_file_serving.rs:698` (download only)

The REAPI service (`kura/src/reapi/mod.rs`) is a separate gRPC stack (`ByteStream`, `ContentAddressableStorage`, `ActionCache`, `Capabilities`). It uses `ArtifactProducer::Reapi` throughout, but **only for Prometheus metrics** (`record_artifact_read`, `record_artifact_write`, `record_public_request_latency`). It has zero calls to `record_public_download` / `record_public_upload` / `record_usage`, and never routes through `accelerated_file_serving`.

So `artifact_kind_for_usage(Reapi) => "reapi"` (`kura/src/http.rs:2182`, `kura/src/accelerated_file_serving.rs:707`) is defined but never reached — no usage rollup is ever created with `producer = Reapi`.

## Impact

- REAPI/gRPC cache traffic (Bazel, other REAPI clients) contributes **0 bytes / 0 requests** to `kura_usage_events`.
- Any Grafana panel, usage report, or billing aggregation built on this table under-counts total egress/ingress by the entire REAPI volume, with no indication that data is missing.

## Proposed work

Bring the REAPI path to parity with the HTTP path by emitting usage rollups from the bandwidth-bearing gRPC handlers, tagged `traffic_plane = "public"`, `protocol = "grpc"`, `artifact_kind = "reapi"`:

- [ ] `ByteStream::read` (download) — `kura/src/reapi/mod.rs` (~L812/883) → `record_public_download`
- [ ] `ByteStream::write` (upload) — `kura/src/reapi/mod.rs` (~L247/412) → `record_public_upload`
- [ ] `ContentAddressableStorage::batch_read_blobs` (download) — `kura/src/reapi/mod.rs:730`
- [ ] `ContentAddressableStorage::batch_update_blobs` (upload) — `kura/src/reapi/mod.rs:677`
- [ ] Only record on the success paths (bytes actually transferred), mirroring how the HTTP handlers gate on the `"ok"` metric arm.

## Also: audit the accelerated serving path

The accelerated (zero-copy) serving path already records **download** usage (`kura/src/accelerated_file_serving.rs:698`, success arm only). As part of this work:

- [ ] Confirm every producer served through `accelerated_file_serving` reaches `record_usage` (no early-return before the OK arm that would drop a served download).
- [ ] If/when large REAPI reads are ever routed through accelerated serving, ensure they carry `producer = Reapi` so they're attributed as `reapi` rather than silently dropped.

## Verification

- After instrumenting, `SELECT DISTINCT artifact_kind, protocol FROM kura_usage_events` should include `('reapi', 'grpc')`.
- A Bazel build against a Kura REAPI endpoint should produce non-zero `bytes` / `request_count` rollups for the exercised account/project.

## Notes

- `protocol = "grpc"` is a new value for that column (today it's always `"http"`); downstream dashboards/queries that assume `http` should be reviewed.
- Decide explicitly whether REAPI should count toward the same usage/billing surface as HTTP cache traffic, or be reported separately — the schema already distinguishes them via `protocol` / `artifact_kind`.

_Discovered while querying `kura_usage_events` for `artifact_kind = 'reapi'` and finding it always empty._

esnunes:
Fixes #11703

## What changed

Kura's REAPI (gRPC) serving path is now recorded into the `kura_usage_events` rollups, on parity with the HTTP path. Six bandwidth-bearing gRPC handlers emit usage on their success arms:

| Handler | Direction | Bytes |
|---|---|---|
| `ByteStream::write` | upload | `manifest.size` |
| `ByteStream::read` | download | `bytes_to_read` |
| `CAS::batch_update_blobs` | upload | per successful blob |
| `CAS::batch_read_blobs` | download | per successful blob |
| `ActionCache::update_action_result` | upload | encoded action result |
| `ActionCache::get_action_result` | download | stored action result + inlined stdout/stderr/output files |

Rollups are tagged `traffic_plane="public"`, `protocol="grpc"`, `artifact_kind="reapi"`.

## Why

Rows in `kura_usage_events` were only produced by `record_public_download` / `record_public_upload`, both called exclusively from the **HTTP** serving path. The REAPI gRPC stack (`ByteStream`, `ContentAddressableStorage`, `ActionCache`, `Capabilities`) used `ArtifactProducer::Reapi` only for Prometheus metrics and never called into the usage recorder. As a result `artifact_kind_for_usage(Reapi) => "reapi"` was defined but **unreachable** — every byte of Bazel/REAPI cache traffic counted as 0 bytes / 0 requests, so any usage or billing aggregation on this table silently under-counted total egress/ingress by the entire REAPI volume.

## Root cause & solution choice

The gap was purely that the gRPC handlers had no call path to `Usage`. The fix adds `record_public_grpc_download` / `record_public_grpc_upload` to `Usage` (delegating to the same private `record`, so bucketing/outbox/delivery are unchanged) and calls them from the gRPC handlers via two small `ReapiService` helpers.

- **Success-arm gating**: usage is recorded only where the metric layer already records `"ok"`, so failed/aborted transfers don't inflate counts — matching the HTTP handlers.
- **Tenant attribution**: taken from the `x-kura-tenant-id` / `x-tuist-account-handle` metadata header — the gRPC analog of the HTTP `tenant_id` query param, and the same headers the extension authorizes against — falling back to the node's configured tenant so bytes are never silently dropped rather than dropping unattributed traffic.
- **Namespace**: derived from the REAPI `instance_name` / `resource_name`, exactly as the store keys blobs, so the control plane's `"{tenant}/{namespace}"` project resolution works for Bazel's `--remote_instance_name=<projectHandle>` + `x-tuist-account-handle=<accountHandle>` setup.
- **Parity refinements** (from review): a re-uploaded / already-present blob is not billed (the same rule as the HTTP path's `artifact_exists` short-circuit); a batch RPC books one request rather than one per blob (consistent with ByteStream/HTTP); usage is booked after the response is fully built so a late failure can't leave a booked-but-failed rollup; and authorization and billing now resolve the tenant header through one helper, so a duplicated header can't bill a different account than the one it authenticated as.
- **Under-lock presence, not a hot-path pre-check** (from review): the "was it already present" signal that gates upload billing comes from the store's persist itself — `persist_artifact_from_{path,bytes}_and_enqueue` return a `PersistedArtifact { manifest, already_present }`, with presence evaluated inside the per-artifact write lock next to the existence check the store already does there. This removes the extra `artifact_exists` lookup per uploaded blob the earlier revision ran on the gRPC hot path, and fixes the duplicate-accounting race: two concurrent uploads of the same missing blob now resolve to exactly one billed writer (the version-based `Applied`/`IgnoredStale` outcome can't stand in for this — a re-upload takes a newer version and still applies over a present blob). The HTTP put path keeps its pre-check (it short-circuits before reading the request body) but now also bills from the under-lock signal, closing the same race there.
- **ActionCache metering** (from review): `GetActionResult` egress (stored action result plus every blob it inlines) and `UpdateActionResult` ingress (the encoded action result) are now billed too, so the "same usage surface" claim holds for the whole REAPI serving surface. Action results are mutable entries whose content changes across updates, so `UpdateActionResult` bills each applied update with no CAS-style "already present" dedupe — the same rule as the HTTP key-value path.

`protocol = "grpc"` is a **new value** for that column (previously always `"http"`). REAPI is routed into the *same* usage surface as HTTP but stays distinguishable by `protocol` / `artifact_kind`, so downstream can split or combine.

### Accelerated serving path — audited, no change needed
REAPI is gRPC and never routes through the accelerated (zero-copy) HTTP serving path. That path already records usage only on the OK arm (no early return drops a served download) and already resolves `artifact_kind_for_usage(Reapi) => "reapi"` if a REAPI read were ever routed through it.

## Downstream consumers — audited

I traced every consumer of `kura_usage_events` on the server. **Nothing filters `protocol = 'http'` or whitelists `artifact_kind`, so the new grpc/reapi rows fold into the existing aggregations automatically — no downstream change is required.**

| Consumer | Effect on grpc/reapi rows |
|---|---|
| Ingest — `KuraUsageController` → `Kura.Usage.create_events` | Accepted: no `protocol`/`artifact_kind` validation; only auth + per-account tenant-scoping are enforced. |
| ClickHouse schema | `protocol` / `artifact_kind` are plain `String` — no enum or CHECK — so `"grpc"` / `"reapi"` insert cleanly. |
| Aggregations — `Kura.Usage` (`totals`, `per_region`, `traffic_time_series_by_region`, `project_ids_with_usage`) | Group by `account_id` / `direction` / `region` / `project_id` / `event_id` only — never by `protocol` / `artifact_kind` — so grpc/reapi bytes add straight into the sums. |
| Dashboard — `usage_live` (the only caller) | No per-protocol or per-kind widget, so nothing renders `"reapi"` as an unknown category. |
| Dedup — `event_id` | Composed with `…:protocol:artifact_kind`, so grpc/reapi rows get distinct ids from http/xcode; no ReplacingMergeTree collision. |

Ruled out: **billing** has no reference to `kura_usage_events`; there are **no Grafana dashboards** on this table (`grafana-datasource` is a thin client over the build/test-duration API, unrelated); `organization_usage` / `atlas_usage_controller` are a separate Atlas usage surface.

**The remaining gap is attribution, not exclusion.** A REAPI row whose tenant/namespace doesn't resolve to a Tuist account/project still lands — it just lands at `account_id = 0` / `project_id = 0` (unattributable) rather than being dropped. In practice Bazel sends `x-tuist-account-handle` + `--remote_instance_name`, so it resolves.

## Impact
- REAPI/gRPC cache traffic now contributes real `bytes` / `request_count` rollups, attributed to the exercised account/project.
- The customer usage dashboard (`usage_live`) now includes REAPI traffic in its egress / ingress / request totals — the intended "same surface" outcome, verified in the audit above (no `protocol = 'http'` filter silently drops it).

## Validation
All 288 `kura` lib tests pass via Bazel; clippy clean (rules_rust aspect, warnings as errors). Key tests:
- `cas_batch_transfers_record_grpc_usage_events` — drives `batch_update_blobs` + `batch_read_blobs`; asserts grpc/reapi rollups, that a batch RPC of N blobs books **one** request (not N), and that a stale re-upload isn't billed again.
- `bytestream_transfers_record_grpc_usage_events` — end-to-end over a real gRPC server; drives `ByteStream::write` (twice) + `read` and asserts the re-write isn't double-billed.
- `action_cache_transfers_record_grpc_usage_events` — drives `update_action_result` + `get_action_result` (with stdout inlining); asserts the upload rollup books the encoded action result, the download rollup books stored-result + inlined-stdout bytes, and each RPC counts one request.
- `persist_reports_already_present_across_re_uploads` — store-level: a re-upload reports `already_present` even though it applies with a newer version.
- `concurrent_persists_of_same_missing_key_report_one_not_present` — store-level: N concurrent persists of the same missing key (forced to overlap via the durable-append failpoint) report exactly one "newly stored", the property that kills the double-billing race.
- `tenant_id_from_metadata_takes_first_value_of_a_repeated_header` — authorization and billing resolve the tenant identically on a duplicated header.
- `usage_tenant_id_prefers_metadata_header_and_falls_back_to_node_tenant`.

The tests assert the issue's verification target: `('reapi', 'grpc')` now appears in the rollups' `(artifact_kind, protocol)`.

## Open decision from the issue
This routes REAPI into the same `kura_usage_events` surface as HTTP, distinguished by `protocol` / `artifact_kind`. That decision is realized end-to-end: because the dashboard aggregations don't filter by protocol, REAPI now shows in the account's usage totals. If billing should ever treat the two separately, the `protocol` column makes that a query-level change with no schema work.

## Usage dashboard

Dashboard showing the ingress/egress and # requests of a kura bazel build using kura as remote cache.
Screenshot 2026-07-08 at 01 36 56


kura/src/reapi/mod.rs
```diff
@@ -1109,14 +1210,23 @@ async fn maybe_read_cas_bytes(
...
 async fn persist_cas_blob(
     state: &SharedState,
     namespace_id: &str,
     digest: &reapi::Digest,
     bytes: &[u8],
-) -> Result<(), String> {
+) -> Result<bool, String> {
     validate_digest_bytes(digest, bytes)?;
     let key = blob_key(&digest_key(digest).map_err(|error| error.message().to_owned())?);
+    let already_present = state
```
fortmarek: [P2] Could we avoid doing this `artifact_exists` check on the request hot path? `BatchUpdateBlobs` now runs an extra manifest/storage existence lookup per uploaded blob, and `ByteStream::write` has the same pre-persist check. The store already serializes and re-checks this artifact under its write lock during persist; if two concurrent uploads race on a missing blob, both pre-checks can return `false`, the loser becomes `IgnoredStale` inside the store, but the wrapper still records it as newly stored. Exposing the store's `Applied` vs `IgnoredStale` outcome and recording usage only for `Applied` would remove the redundant hot-path lookup and fix the duplicate accounting race.

kura/README.md
```diff
@@ -417,6 +417,8 @@ When `KURA_CONTROL_PLANE_URL`, `KURA_CONTROL_PLANE_CLIENT_ID`, and `KURA_CONTROL
 POST {KURA_CONTROL_PLANE_URL}/_internal/kura/usage
 ```
 
+Both surfaces are metered: the HTTP cache path records rollups with `protocol = "http"`, and the REAPI (gRPC) path — `ByteStream` read/write and CAS `BatchReadBlobs`/`BatchUpdateBlobs` — records them with `protocol = "grpc"` and `artifact_kind = "reapi"`, so Bazel and other REAPI clients count toward the same usage surface.
```
fortmarek: [P2] This says REAPI clients count toward the same usage surface, but the ActionCache REAPI methods are still unmetered. `GetActionResult` can return the serialized action result plus inline stdout/stderr/output files, and `UpdateActionResult` uploads an encoded action result; neither path calls the new usage helpers. As written, some REAPI bytes still stay out of `kura_usage_events`, so either those methods should be metered too or the docs/PR description should narrow the claim to ByteStream and CAS batch traffic.

fortmarek (commented):
Two findings from review. The first one is the hot-path/performance concern we discussed, and it also affects duplicate accounting under concurrent uploads.

fortmarek (approved):
Looks good, would just double check tracking the usage doesn't impact the performance of the download/upload hot path.

kura/src/reapi/mod.rs
esnunes: Thanks Marek, both halves of this were spot on! Fixed in b782d9ef16, with one twist on the suggested mechanism: the store's `Applied`/`IgnoredStale` outcome can't stand in for "already present" — a re-upload takes a newer `now_ms()` version and still lands as `Applied`, so billing on `Applied` would re-bill every retry. Instead, the persist now reports the presence it *already* evaluates under the per-artifact write lock: `persist_artifact_from_{path,bytes}_and_enqueue` return `PersistedArtifact { manifest, already_present }`.

With that, `ByteStream::write` and `BatchUpdateBlobs` drop their pre-checks entirely — no extra lookup on the hot path (the store was already doing that read to decide staleness) — and billing on `!already_present` means two concurrent uploads of the same missing blob resolve to exactly one billed writer. The HTTP put path keeps its pre-check (it short-circuits before reading the request body) but now bills from the under-lock signal too, closing the same race there.

Both properties are pinned by new store tests: `persist_reports_already_present_across_re_uploads` (a re-upload reports present even though it applies) and `concurrent_persists_of_same_missing_key_report_one_not_present` (4 overlapping persists via the durable-append failpoint → exactly one "newly stored").

On the hot-path note from your approval: with the pre-check gone, the metering itself adds no I/O — recording is a bounded in-memory counter bump (hash + two saturating adds under a short-lived mutex).

kura/README.md
esnunes: Good catch! I went with metering them rather than narrowing the claim, so the "same usage surface" statement now holds for the whole REAPI serving surface (5ff6591809): `GetActionResult` books the stored action result plus any stdout/stderr/output-file bytes it inlines as egress, and `UpdateActionResult` books the encoded action result as ingress — both on the success arm after the response is fully built, matching the other handlers.

One deliberate difference from the CAS paths: action results are mutable entries whose content changes across updates, so `UpdateActionResult` bills every applied update with no "already present" dedupe — the same rule as the HTTP key-value path. The README paragraph now names the ActionCache methods explicitly.

Covered by `action_cache_transfers_record_grpc_usage_events`, which drives both methods (with stdout inlining) and asserts the exact byte counts and one request per RPC.

pepicrft (approved):
Reviewed the latest diff and validated the original findings are addressed. The Kura unit tests, clippy, and format check pass locally.

PR #11714, opened 2026-07-08
fix(server): support member management account tokens

fortmarek:
## What changed

This change wires the existing `account:members:read` and `account:members:write` scopes into organization membership authorization.

- `account:members:read` and `account:members:write` can now read organization membership and invitation data for the token's own account.
- `account:members:write` can now create/cancel invitations and update/remove organization members for the token's own account.
- Organization and invitation API controllers now authorize with the full authenticated subject instead of only `current_user`, so account-token requests are evaluated by the policy layer.
- Account-token authentication now carries the token creator as `issued_by`, letting invitation creation keep the existing invariant that invitations are attributed to a real user inviter.

## Why it changed

Headless automation needs to manage organization access without a browser-backed user session. The scopes for member management already existed on account tokens, but the relevant policies did not consume them and some controllers only passed `current_user` into authorization. That made scoped account tokens behave as unauthenticated/unauthorized for membership endpoints.

## Root cause

The buggy behavior was that account tokens minted with `account:members:read` or `account:members:write` still received `403` from endpoints such as `GET /api/organizations/:organization_name` and `POST /api/organizations/:organization_name/invitations`. The policy rules for organization, invitation, and member actions only allowed user-session admins, while token-authenticated requests were stored as `current_subject` rather than `current_user`.

## Approach

The fix keeps the existing user-admin checks intact and adds account-token checks that require both the right member-management scope and an account match. Write actions require `account:members:write`; read actions accept either `account:members:read` or `account:members:write`. Invitation creation still requires a real user attribution by resolving the user behind the token creator.

## Impact

Self-hosted and server-side integrations can now drive the same organization invite, cancel-invite, member lookup, role update, and removal flows with scoped account tokens. Tokens remain bounded to their owning account and read-only member tokens cannot perform write actions.

## Validation

Ran:

```sh
mix format lib/tuist/authentication.ex lib/tuist/authorization.ex lib/tuist_web/controllers/api/invitations_controller.ex lib/tuist_web/controllers/api/organizations_controller.ex test/tuist/authentication_test.exs test/tuist/authorization_test.exs test/tuist_web/controllers/api/invitations_controller_test.exs test/tuist_web/controllers/api/organizations_controller_test.exs
mix test test/tuist/authentication_test.exs test/tuist/authorization_test.exs test/tuist_web/controllers/api/invitations_controller_test.exs test/tuist_web/controllers/api/organizations_controller_test.exs
git diff --check
```

The focused test run completed with `196 tests, 0 failures`.

### DIFF
--- a/server/lib/tuist/authentication.ex
+++ b/server/lib/tuist/authentication.ex
@@ -48,7 +48,7 @@ defmodule Tuist.Authentication do
     project_token = Projects.get_project_by_full_token(token)
 
     if is_nil(project_token) do
-      case Accounts.account_token(token, preload: [:account, created_by_account: :user]) do
+      case Accounts.account_token(token, preload: [:account, :account_token_projects]) do
         {:ok, account_token} ->
           %AuthenticatedAccount{
             account: account_token.account,
@@ -56,8 +56,7 @@ defmodule Tuist.Authentication do
             all_projects: account_token.all_projects,
             project_ids: Enum.map(account_token.account_token_projects, & &1.project_id),
             token_id: account_token.id,
-            created_by_account_id: account_token.created_by_account_id,
-            issued_by: issued_by_user(account_token)
+            created_by_account_id: account_token.created_by_account_id
           }
 
         _ ->
@@ -67,9 +66,6 @@ defmodule Tuist.Authentication do
       project_token
     end
   end
-
-  defp issued_by_user(%{created_by_account: %{user: %User{} = user}}), do: user
-  defp issued_by_user(_account_token), do: nil
 
   @doc """
     Refreshes a given token, updating the account handle in the `preferred_username` claim.
--- a/server/lib/tuist/authentication.ex
+++ b/server/lib/tuist/authentication.ex
@@ -48,7 +48,7 @@ defmodule Tuist.Authentication do
     project_token = Projects.get_project_by_full_token(token)
 
     if is_nil(project_token) do
-      case Accounts.account_token(token, preload: [:account, :account_token_projects]) do
+      case Accounts.account_token(token, preload: [:account, created_by_account: :user]) do
         {:ok, account_token} ->
           %AuthenticatedAccount{
             account: account_token.account,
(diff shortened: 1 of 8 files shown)


pepicrft (changes requested):
I found a blocking security issue in the account-token authentication change.

`server/lib/tuist/authentication.ex:60` now sets `issued_by` on stored account tokens. Existing project authorization treats `issued_by` as user-delegated cross-account authority, so an all-projects token for account A can access projects in account B if the token creator also belongs to account B. I reproduced this locally with a token owned by account A and a target project owned by account B: `Checks.scopes_permit(subject, project_b, "project:cache:read")` returned `true`.

That lets an all-projects account token escape its owning account anywhere project-scoped account tokens are accepted. Please keep stored account tokens scoped to `subject.account`, and use a separate field for invitation attribution if needed.

fortmarek:
Thanks @pepicrft, good catch. Addressed in `ad608252625`.

What changed:
- Stored account-token authentication no longer preloads `created_by_account: :user` or sets `issued_by`.
- The authenticated account token subject now keeps only `created_by_account_id`, so project authorization remains scoped to `subject.account` instead of inheriting the token creator's cross-account memberships through `issued_by`.
- Invitation creation still has attribution, but it resolves the inviter locally from `created_by_account_id` inside `InvitationsController`. If the creator account does not map to a user, the flow fails closed with the existing forbidden response.

I also added a regression test for the exact escape case: an all-projects token owned by account A, created by a user who is admin in account B, must not pass `Checks.scopes_permit(subject, project_b, "project:cache:read")`.

Validation run:
- `mix format --check-formatted lib/tuist/authentication.ex lib/tuist_web/controllers/api/invitations_controller.ex test/tuist/authentication_test.exs`
- `mix test test/tuist/authentication_test.exs test/tuist_web/controllers/api/invitations_controller_test.exs` -> 33 tests, 0 failures
- `git diff --check`

pepicrft (approved):

PR #11718, opened 2026-07-08
feat(kura): inline every action output blob on GetActionResult wildcard

fortmarek:
## What changed

`GetActionResult` now accepts `"*"` in the standard REAPI `inline_output_files` hint, meaning "inline the contents of every output file the response budget affords". Inlined bytes ride the existing `OutputFile.contents` field. Two behaviors distinguish the wildcard from explicitly listed paths:

- **Best-effort**: a blob the `MaterializationBudget` cannot afford is skipped (later, smaller files may still fit) instead of failing the RPC. Explicitly listed paths keep today's hard `resource_exhausted` error.
- Everything else (authz target, metrics, budget accounting, `maybe_read_cas_bytes`) is the existing inlining machinery.

## Why

The Xcode CAS plugin's action results are closure manifests: one `OutputFile` per value-graph node, whose `path` is a digest the client cannot know before the response. Standard path-matched inlining is therefore unusable for it, and every remote cache hit costs **two WAN round-trips**: `GetActionResult` for the manifest, then `BatchReadBlobs` for the blobs. The compiler issues these lookups synchronously, so on a local dev machine building against a remote region (~28ms RTT) the second trip is pure serialized latency across thousands of resolves per warm build.

With the wildcard, the value-graph blobs ride the `GetActionResult` response — one round-trip per hit.

## Rollout safety

Additive and negotiated, per the mixed-version rules:

- **Old client + new server**: no `"*"` is ever sent; behavior unchanged.
- **New client + old server**: `"*"` matches no literal path, nothing is inlined, and the client falls back to `BatchReadBlobs` — today's behavior exactly.
- Partial inlining (budget exhaustion) degrades to the same fallback per-blob.

No wire-format, storage, or replication changes.

## Validation

- Two new unit tests: wildcard inlines every output file; wildcard degrades to partial (large-first ordering proves a rejected file doesn't stop later ones) under a constrained memory config. Existing explicit-path budget-rejection test unchanged and passing.
- `mise run clippy` and `mise run format -- --check` pass.
- End-to-end against a local kura built from this branch, with the companion CLI change and a +28ms-RTT latency relay, building the tuist CLI scheme (2,638 cacheable tasks, fully populated cache, fresh DerivedData, identical conditions for both runs):

| client | wall | hits | proxy op time |
|---|---|---|---|
| current (get + batch-read, 2 RTT) | 160s | 100% | action 157s + fetch 161s = 318s |
| inline (this + companion PR, 1 RTT) | **138s** | 100% | action 175s + fetch **0** = 175s |

The local-CAS floor for this fixture is ~143s: with inlining, the +28ms round-trip cost disappears into the compiler's own concurrency instead of serializing. `inlined=145,846 / blobs=0` in the proxy stats confirms every blob rode the action response.

Caveat for production expectations: in this simulation the old path's fetch leg is ~30ms/resolve, while against production eu-central it measured ~316ms/resolve (real transfer + server time). Inlining removes the per-resolve round-trip serialization; the byte transfer itself still flows (now overlapped with the action response). Production numbers require this to ship in a kura release and roll out to the fleet first.

## Rollout order

Land + release + deploy this before the companion CLI PR pays off; the CLI change is safe to land in any order (it falls back to batch-reads against servers without the extension).

### DIFF
--- a/kura/src/reapi/mod.rs
+++ b/kura/src/reapi/mod.rs
@@ -545,26 +545,57 @@ impl ActionCache for ReapiService {
             action_result.stderr_raw = bytes;
         }
         if !request.get_ref().inline_output_files.is_empty() {
+            // `"*"` is a Kura extension to the REAPI `inline_output_files`
+            // hint: inline the contents of every output file the response
+            // budget affords. It exists for clients (the Xcode CAS plugin)
+            // whose output-file paths are digests unknown before this
+            // response, collapsing the action lookup + blob fetch into one
+            // round-trip. Best-effort by design: a file the budget cannot
+            // afford stays un-inlined and the client falls back to
+            // BatchReadBlobs for it, so mixed client/server versions
+            // interoperate unchanged (an old server matches no literal `"*"`
+            // path and inlines nothing).
+            let inline_all = request
+                .get_ref()
+                .inline_output_files
+                .iter()
+                .any(|path| path == "*");
             for output_file in &mut action_result.output_files {
-                if !request
-                    .get_ref()
-                    .inline_output_files
-                    .iter()
-                    .any(|path| path == &output_file.path)
+                if !inline_all
+                    && !request
+                        .get_ref()
+                        .inline_output_files
+                        .iter()
+                        .any(|path| path == &output_file.path)
                 {
                     continue;
                 }
-                if output_file.contents.is_empty()
-                    && let Some(digest) = &output_file.digest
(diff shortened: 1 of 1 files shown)


fortmarek:
Folded into #11719 (single PR covering both the kura wildcard extension and the cas-plugin client, plus the release-path fix). The multi-scope squash title works with the changelog scope parsing, and release gating is path-based, so one merge cuts both releases.

PR #11720, opened 2026-07-08
fix(cli): CAS proxy — verify on-disk presence for cached hits + reclaim idle caches

fortmarek:
This PR hardens the machine-wide `tuist-cas-proxy`'s in-memory action-cache caches against a wiped on-disk CAS, and bounds the memory the long-lived proxy holds. Three related changes, sharing the `PathState::invalidate()` helper.

---

### What / why

The proxy cached resolved Hits in memory (`PathState.resolved`, `Resolution::Hit(value_digest)`) keyed by `cas_path` and returned them **without** checking the value graph's blobs still exist on disk. After a user wiped DerivedData (`xcodebuild clean` or deleting the DD folder) while the long-lived proxy kept running, the next build got stale "hit" answers for objects no longer on disk, and the compiler failed with `CAS error: missing object '<digest>'` at `SwiftExplicitDependencyGeneratePcm`. Reproduced deterministically: build → wipe DD → build against the same proxy → failure; a fresh proxy fixed it, isolating the staleness to the proxy's in-memory maps.

The non-proxy path in `lib.rs` (`actioncache_get_impl`) was already correct because it does an authoritative `llcas_cas_load_object` on every get; the proxy path short-circuited on the in-memory `resolved` map first.

- Extracted `Proxy::load_present()` — an authoritative load that **bypasses** the `known_local` shard cache. `is_local` memoizes a positive result on top of it (unchanged).
- Added an FFI-free `fast_path()` gate (+ `FastPath` enum): serve a Hit only when `load_present` confirms presence; on absence, `invalidate()` and re-resolve.
- Added `PathState::invalidate()` (clears `resolved` + `known_local`).

The presence load runs outside the single-flight lock and after releasing the `resolved` lock, so it never serializes other keys — one root-object load per action key.

### What / why

Change 1 alone was **incomplete** (caught in review): it only guards the cached-Hit serve path. If the first request after a wipe is an **uncached or changed** action key — or a concurrent resolve races ahead — it reaches `resolve_uncached`, whose `is_local` filter short-circuits on the in-memory `known_local` shards with no disk check. Those shards still hold the previous build's ~1.9M digests, so nodes shared with the new key's graph are treated as local, skipped from the fetch set, and never re-materialized. A `Resolution::Hit` is then cached and returned for a graph whose blobs are missing on disk — the original failure, still reachable for changed keys and parallel builds.

Detect the wipe at the **head of every resolve**, before any manifest filtering:
- Stamp each `PathState` with its CAS directory identity (`CasGeneration { ino, birth_nanos }` via `std::fs::metadata`).
- `check_generation()` compares it each resolve; on change (directory deleted and recreated) it invalidates `resolved` + `known_local` + `publish_cache` before the fast path or `resolve_uncached` can trust them.
- The generation lock is held **across** the invalidation, so a concurrent resolve can't observe the new generation as unchanged and filter against `known_local` mid-clear.

One `stat` per resolve (per action key, not per node) — negligible. This covers every entry path (uncached, stale-miss, parallel), not just re-requested Hits. Change 1's per-Hit presence check is kept as a complementary guard for same-generation partial loss (e.g. pruning). Residual, documented: a wipe that empties the CAS *in place* without changing the directory's identity is not caught by the stamp; the per-Hit check still covers re-requested keys in that case.

### What / why

Reviewing the fix surfaced a memory question: the proxy holds its per-path caches (`resolved`, 32 `known_local` shards, `publish_cache`) for the life of the process. `enforce_cache_bounds` only releases memory when a single build overruns a size cap — never for a project that stops being built — and `PathState` is `Box::leak`ed, so a path is never dropped. A long-lived launchd proxy therefore accumulates a few hundred MB per distinct DerivedData path it has served (dominated by `known_local`'s ~1.9M-digest warm working set), released only by a restart.

Added idle reclamation to the existing 10s maintenance loop:
- Per-path `last_used` (`AtomicU64` millis since a `Proxy.epoch`), bumped once per resolve/publish (per action key, not per node).
- `Proxy::reclaim_idle()` clears a path's caches when idle past `IDLE_RECLAIM` (30 min) **or** its on-disk CAS directory is gone (a deleted project/worktree).
- Decision factored into a pure `should_reclaim(idle, cas_dir_gone)`.

Low-risk tier: reclaims the ~99% of per-path memory in the digest maps while retaining the small `PathState` shell (llcas handle + empty maps, ~KB), so a later build re-warms from the remote. Freeing the shell + handle would require `Box::leak` → `Arc` + `Drop`, deferred until FD/handle accumulation across many paths is observed. `IDLE_RECLAIM` is a hardcoded constant like `NEGATIVE_TTL`, not a new env var.

---

## Impact

- Fixes spurious `CAS error: missing object` build failures after `xcodebuild clean` / DerivedData deletion when a long-lived proxy (`tuist setup cache` launchd agent) is running — for re-requested, changed, and uncached keys, and parallel builds.
- Bounds the proxy's steady-state RAM: projects that stop being built, and deleted DerivedData paths, no longer pin their caches until a restart.
- Warm-hit fast path adds one `stat` + one authoritative root-object load + one relaxed atomic store per action key.

## Validation

- `cargo test --manifest-path cas-plugin/Cargo.toml` — 24 lib tests + 1 integration test pass, including 10 regression tests in `proxy::tests`: wiped-hit, present-hit, fresh-miss, stale-miss, absent-key; the three `should_reclaim` policy cases; `generation_changed` policy; and a filesystem test that a recreated directory reads as a new generation (the DD-wipe reproduction at the FS layer). The correctness decisions are factored into FFI-free `fast_path()`, `should_reclaim()`, and `generation_changed()` so they are unit-testable without a live llcas handle.
- `cargo clippy` on `proxy.rs` and `bin/tuist-cas-proxy.rs` is clean. (Two pre-existing `not_unsafe_ptr_arg_deref` errors in `lib.rs:360-363` under clippy 1.94.0 are unrelated.)

---

## Update — review hardening (in-place prune + in-flight race)

Two follow-on gaps raised in review, both closed (test count now 24 → 25 lib):

### In-place prune (`llcas_cas_prune_ondisk_data`)

Xcode calls prune on the shared on-disk CAS to manage size; it removes objects **without recreating the directory**, so the change-2 generation stamp is blind to it and `known_local` keeps marking pruned blobs as local. Added an `OP_INVALIDATE` proxy op: the plugin's prune entry point clears its own `known_local` and tells the proxy to drop that path's `resolved` / `known_local` / `publish_cache`, so a later resolve re-fetches the pruned blobs. (Generation-scoped `known_local` entries alone wouldn't help, since prune doesn't change the directory identity.)

### In-flight resolve race

A resolve that observed the pre-wipe store could finish *after* another thread advanced the generation and cleared the maps, reinserting stale `known_local` / `resolved` entries for the replaced store. Added a monotonic `gen_counter` bumped by every invalidation (wipe or prune): a resolve snapshots it after its wipe check and commits a `known_local` or `resolved` write only if the counter is unchanged, checked **under the same map lock the write takes**. `invalidate()` bumps the counter **before** clearing, so a racing insert either sees the new counter and is skipped or is cleared by the invalidation. A generation change mid-resolve drops the value and returns a miss, re-materialized next resolve.

Residual: a wipe that empties the CAS in place *without* a prune call *and without* recreating the directory is still only caught for re-requested keys by the per-Hit presence check; no such path is known today (clean/delete recreate the directory; size management goes through prune).

### DIFF
--- a/cas-plugin/src/proxy.rs
+++ b/cas-plugin/src/proxy.rs
@@ -17,8 +17,8 @@ use std::sync::{Arc, Condvar, Mutex};
 use std::time::{Duration, Instant, UNIX_EPOCH};
 
 use crate::proxy_proto::{
-    read_request, write_response, Request, OP_PUBLISH, OP_RESOLVE, STATUS_ERROR, STATUS_HIT,
-    STATUS_MISS,
+    read_request, write_response, Request, OP_INVALIDATE, OP_PUBLISH, OP_RESOLVE, STATUS_ERROR,
+    STATUS_HIT, STATUS_MISS,
 };
 use crate::prefetch::Prefetcher;
 use crate::reapi::{self, ManifestEntry, Remote, RemoteConfig};
@@ -155,6 +155,11 @@ pub struct PathState {
     // the store was deleted and recreated under this long-lived proxy, so the
     // in-memory marks below are stale and must be dropped before they are trusted.
     generation: Mutex<Option<CasGeneration>>,
+    // Monotonic counter bumped by every invalidation (a detected wipe or a prune
+    // signal). A resolve snapshots it after its wipe check and only commits its
+    // known_local / resolved writes if it is unchanged, so a resolve that began
+    // under an older store can't reinsert stale marks after the maps were cleared.
+    gen_counter: AtomicU64,
     // key digest -> resolved outcome. A local publish updates its entry, so a
     // miss cached during planning turns into a hit once the local build
     // publishes it; misses also carry a timestamp so a key another machine
@@ -199,12 +204,27 @@ impl PathState {
     /// and re-materializes the full graph rather than trusting stale in-memory
     /// marks. Content-addressed and correctness-preserving, so clearing only
     /// forces re-work, never a wrong answer.
+    ///
+    /// The counter is bumped BEFORE the maps are cleared so that a concurrent
+    /// in-flight resolve (which checks the counter while holding the same map
+    /// lock it is about to write) either sees the new counter and skips its
+    /// write, or writes first and then has it cleared here — never inserts a
+    /// stale mark that survives the clear.
     fn invalidate(&self) {
+        self.gen_counter.fetch_add(1, Ordering::SeqCst);
         self.resolved.lock().unwrap().clear();
         for shard in &self.known_local {
             shard.lock().unwrap().clear();
         }
     }
+}
+
+/// Whether writes from a resolve that observed generation `observed` may still
+/// be committed: only if no wipe or prune advanced the path's `gen_counter`
+/// since. A stale resolve's known_local / resolved inserts describe a store that
+/// has been replaced, so they must be dropped rather than trusted.
+fn committable(observed: u64, current: u64) -> bool {
+    observed == current
 }
 
 // The proxy is single-process and owns these raw handles for its lifetime;
@@ -331,6 +351,7 @@ impl Proxy {
             cas,
             cas_path: cas_path.to_string(),
             generation: Mutex::new(cas_generation(cas_path)),
+            gen_counter: AtomicU64::new(0),
             resolved: Mutex::new(HashMap::new()),
             inflight: Mutex::new(HashSet::new()),
             inflight_cvar: Condvar::new(),
@@ -413,7 +434,13 @@ impl Proxy {
                 inflight = state.inflight_cvar.wait(inflight).unwrap();
             }
         }
-        let outcome = self.resolve_uncached(remote, state, key);
+        // Re-check the generation now that we hold the single-flight slot: a wipe
+        // during the wait must be caught before resolve_uncached trusts
+        // known_local. `observed` is snapshotted here so the write guard drops
+        // this resolve's marks if a wipe/prune advances the counter mid-resolve.
+        self.check_generation(state);
+        let observed = state.gen_counter.load(Ordering::SeqCst);
+        let outcome = self.resolve_uncached(remote, state, key, observed);
         {
             let mut inflight = state.inflight.lock().unwrap();
             inflight.remove(key);
@@ -427,6 +454,7 @@ impl Proxy {
         remote: &Remote,
         state: &'static PathState,
         key: &[u8],
+        observed: u64,
     ) -> Result<Option<Vec<u8>>, String> {
         let op_start = Instant::now();
         let phase = Instant::now();
@@ -461,7 +489,7 @@ impl Proxy {
         let phase = Instant::now();
         let missing: Vec<&ManifestEntry> = manifest
             .iter()
-            .filter(|entry| !self.is_local(state, &entry.llcas_digest))
+            .filter(|entry| !self.is_local(state, observed, &entry.llcas_digest))
             .collect();
         state
             .ms_filter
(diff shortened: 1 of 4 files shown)


cas-plugin/src/proxy.rs
```diff
@@ -51,11 +60,101 @@ enum Resolution {
...
+struct CasGeneration {
+    ino: u64,
+    // Birth time in nanos since the Unix epoch; 0 when the platform can't report
+    // it. Guards the (very unlikely) inode reuse when a directory is recreated.
+    birth_nanos: u128,
+}
+
+/// The CAS directory's current generation, or `None` if it does not exist
+/// (deleted and not yet recreated).
+fn cas_generation(cas_path: &str) -> Option<CasGeneration> {
```
pepicrft, line 126: One gap I would close: this only treats a wipe as a root directory generation change. If Xcode prunes or empties the Content Addressable Storage (CAS, https://en.wikipedia.org/wiki/Content-addressable_storage) in place, for example through `llcas_cas_prune_ondisk_data`, the generation stays the same and `known_local` can still skip blobs that were removed. A proxy invalidation signal on prune, or generation-scoped `known_local` entries, would make changed or uncached keys safe too.

cas-plugin/src/proxy.rs
```diff
@@ -393,10 +545,42 @@ impl Proxy {
...
+    /// Detects a wiped-and-recreated on-disk CAS (a deleted DerivedData under
+    /// this long-lived proxy) from a change in the CAS directory's identity and
+    /// drops the now-stale in-memory marks (`resolved`, `known_local`,
+    /// `publish_cache`) so a resolve re-probes and re-materializes authoritatively.
+    /// Called at the head of every resolve, so it covers uncached/changed keys
+    /// and parallel builds, not only re-requested cached Hits. The generation
+    /// lock is held across the invalidation so a concurrent resolve can't observe
+    /// the new generation as unchanged and filter against `known_local` while it
+    /// is being cleared.
+    fn check_generation(&self, state: &PathState) {
```
pepicrft, line 605: I would also tie in-flight resolves to the generation they observed. A resolve that started before a directory recreation can still finish after another thread advanced the generation and cleared the maps, then reinsert stale `known_local` or `resolved` entries for the new store. Carrying the observed generation through `resolve_uncached` and dropping or retrying writes if it changed would close that race.

pepicrft (approved):
Thanks, the direction makes sense and the tests cover the main regression. I left two follow-up comments around stale cache invalidation edges, but I do not think they need to block this change.

cas-plugin/src/proxy.rs
fortmarek: Closed via an OP_INVALIDATE proxy signal (commit 6406e16): the plugin's `llcas_cas_prune_ondisk_data` now clears its own `known_local` and tells the proxy to drop that path's `resolved`/`known_local`/`publish_cache`, so a later resolve re-fetches the pruned blobs. (Generation-scoped entries alone wouldn't help, since an in-place prune doesn't change the directory identity.)

cas-plugin/src/proxy.rs
fortmarek: Closed via a monotonic `gen_counter` (commit 6406e16): bumped by every invalidation (wipe or prune), a resolve snapshots it after its wipe check and commits a `known_local`/`resolved` write only if the counter is unchanged, checked under the same map lock the write takes. `invalidate()` bumps before clearing, so a racing insert is either skipped or cleared by the invalidation; a mid-resolve generation change drops the value and returns a miss, re-materialized on the next resolve.

PR #11722, opened 2026-07-08
perf(kura): promote old-segment artifacts in the background instead of on the read path

fortmarek:
## What changed

Serving an artifact that lives in an **Old-generation segment** no longer refreshes it inline on the read path. The serving path now answers straight from the Old segment and queues the artifact for a **background promotion worker** (new, spawned at boot alongside the other maintenance tasks), which runs the exact same refresh — re-validating existence and generation first — off the read path. The queue is deduplicated and bounded; dropping an entry is safe because promotion is a best-effort keep-alive (the same outcome as the pre-existing memory-pressure skip). The non-serving `fetch_artifact` path keeps its inline refresh.

## Root cause / why

Every read of an old-segment artifact ran `maybe_refresh_manifest`: a copy of that one artifact into the current segment plus a synchronous RocksDB write, **serialized on the store-global `segment_refresh_lock`, once per artifact**. One Xcode value-graph batch read can touch thousands of tiny old artifacts, so a single `BatchReadBlobs` paid thousands of serialized read+append+fsync cycles — and `batch_read_blobs`' own 16-way read buffering couldn't help, because all readers converge on the same lock.

This is the dominant term in warm Xcode-cache builds from dev machines over WAN against eu-central. Per-resolve instrumentation of a real warm build showed the top 10% of resolves (giant module-scan manifests, 3,500–15,000 entries of ~200-byte blobs) carrying **75% of all fetch time** at ~10.6ms/blob, while the p50 resolve was a healthy 70ms. Isolated probes exonerated everything else first: server reads of *fresh* blobs run at ~0.1ms/blob, concurrency and bandwidth (240 Mbps through kura) are fine — the cost appears exactly when the data has aged into the Old generation, which for a shared node happens within hours.

## Reproduction (local, deterministic)

Fresh local kura, author 4,000 × 200B blobs first, upload filler until the ring ages their segment to Old (`kura_segment_generation_count{generation="old"} 1`), then batch-read them:

| | read 4,000 tiny blobs from Old segment | promotions |
|---|---|---|
| before | **15.7s** (3.9 ms/blob; 0.05s when fresh) | 4,000 inline, `segment_refresh_duration_seconds_sum` = 250s of lock queueing |
| after | **0.2s**, all 4,000 blobs served | 4,000 in background, drained in ~30s |

A post-promotion read stays fast, and `kura_segment_refreshes_total` confirms the keep-alive still happens (4,000 refreshes either way — only *when* changed).

## Rollout safety

- Node-local behavior change only: response bytes/headers unchanged, no wire-format, storage-format, or replication changes; safe under mixed versions.
- Serving old-segment data without an inline refresh is the store's existing behavior under memory pressure (`allow_segment_refresh() == false`), so this generalizes an already-exercised path rather than introducing a new one.
- Reads racing a concurrent segment reclaim are safe as before: segments are unlinked, never truncated, so open handles stay readable, and a lost race degrades that lookup to a miss.
- The worker holds `segment_refresh_lock` per promotion exactly as the read path did, so write-path interactions are unchanged.

## Validation

- New unit tests: serving from an Old segment returns the old-segment manifest and enqueues (deduplicated) promotion; applying the promotion rewrites into the current segment; an end-to-end worker test (spawned worker + notify) promotes a served artifact within the timeout. Existing `fetch_artifact` inline-refresh test unchanged and passing (54 store tests total).
- `mise run clippy` and `mise run format -- --check` pass.
- Live validation on a local node with the shipped REAPI client, exactly the reproduction above, including verifying all 4,000 blobs were actually served (`found=4000/4000`) — an earlier iteration of the harness was invalidated by evicted blobs reading as fast empty responses, so the final numbers assert payload presence.

## Relation to other work

Complements #11719 (one-round-trip value-graph fetch): that PR removes the per-resolve WAN round-trip on the healthy p50 body; this one removes the old-data tail that instrumentation showed was ~75% of warm-build fetch time. Both are needed for warm Xcode-cache builds over WAN to approach the local-CAS floor.

### DIFF
--- a/kura/src/app.rs
+++ b/kura/src/app.rs
@@ -197,6 +197,7 @@ async fn run_with_config(
     spawn_multipart_janitor_task(state.clone());
     spawn_tmp_dir_metrics_task(state.clone());
     spawn_geoip_refresh_task(state.clone());
+    spawn_segment_promotion_task(state.clone());
 
     // When the node enrolled on boot, keep its peer certificate fresh in-process
     // so a short leaf does not require a restart.
@@ -451,6 +452,18 @@ async fn shutdown_signal() {
     let terminate = std::future::pending::<()>();
 
     wait_for_shutdown_signal(ctrl_c, terminate).await;
+}
+
+// Drains the store's read-path promotion queue: artifacts served from an
+// Old-generation segment are rewritten into the current segment here instead
+// of inline on the read path (see Store::run_promotion_worker).
+fn spawn_segment_promotion_task(state: Arc<AppState>) {
+    tokio::spawn(
+        async move {
+            state.store.run_promotion_worker().await;
+        }
+        .in_current_span(),
+    );
 }
 
 fn spawn_snapshot_task(state: Arc<AppState>) {
--- a/kura/src/store.rs
+++ b/kura/src/store.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::{BTreeMap, HashMap},
+    collections::{BTreeMap, HashMap, HashSet, VecDeque},
     path::{Path, PathBuf},
     pin::Pin,
     sync::{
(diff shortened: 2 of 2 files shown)


fortmarek:
Folded into #11719 alongside the one-round-trip inlining — together they cover both measured halves of the warm-over-WAN cost (p50 round-trips + the old-segment tail).

PR #11723, opened 2026-07-08
feat(server): surface runner job insights

fortmarek:
## What changed

This PR connects a Tuist Runner job detail page with the build and test insight data produced by that same GitHub Actions workflow run.

- Adds an Insights card to the runner job overview between CI Details and Metrics.
- Shows linked Builds and Tests sections with duration, Xcode cache hit rate, module cache hit rate, and selective testing hit rate when those values exist.
- Links the build/test rows in the Insights card to their existing build/test detail pages.
- Adds compact build/test chips under matching job steps so the Build and Run tests steps expose the scheme and duration directly.
- Keeps machine metric charts immediately below the Insights card and continues highlighting metric windows for relevant build/test steps.
- Updates local seeded runner data to use macOS-only, production-shaped CI runs with linked build insights, test insights, module cache, Xcode cache, and selective testing values.
- Adds a marketing changelog entry with a screenshot of the runner job Insights card.

## Screenshot

Runner job insights

## Why

Runner jobs, build insights, and test insights were useful on their own, but the runner job detail page did not make the relationship obvious. A CI job could show steps and machine metrics while the associated build/test insight data lived elsewhere. That made it harder to answer the natural debugging questions: which build or test did this job produce, what cache behavior did it see, and which step should I inspect next?

This keeps the first iteration focused on the runner job detail page only.

## Approach

The runner job page resolves the project from the job repository, then directly queries build/test runs for the same GitHub workflow run id. It derives aggregate cache/selective-testing summaries from the linked command events and renders those summaries in a dedicated Insights card. Step chips use simple step-name heuristics to associate build/test insights with their likely CI step.

The broader reciprocal surfaces were intentionally removed from this PR for now. A follow-up can add runner context back into build/test detail pages and build/test run list pages once the runner job details interaction settles.

## Impact

Users can inspect a runner job and immediately see the build/test outcomes, cache effectiveness, and selective-testing behavior tied to that CI run. The page also makes the relationship between steps and build/test details visible without expanding logs.

## Validation

- `mise exec -- mix test test/tuist_web/live/runner_job_live_test.exs test/tuist_web/live/build_run_live_test.exs test/tuist_web/live/build_runs_live_test.exs test/tuist_web/live/test_run_live_test.exs test/tuist_web/live/test_runs_live_test.exs test/tuist/marketing/changelog_entry_parser_test.exs`
- Manual browser pass against seeded data at `http://localhost:8520/tuist/runners/runs/4199008/jobs/4200008`.

Note: `server/mix.lock` has unrelated local dependency-resolution drift and remains intentionally unstaged.


server/lib/tuist/runners/jobs.ex
```diff
@@ -83,6 +87,65 @@ defmodule Tuist.Runners.Jobs do
...
+    |> where([test], test.ci_provider == "github")
+    |> where([test], test.ci_run_id == ^workflow_run_id)
+    |> order_by([test], desc: test.inserted_at, desc: test.ran_at)
+    |> ClickHouseRepo.all()
+    |> latest_runner_runs_by_id()
+    |> Enum.sort_by(&datetime_sort_key(&1.ran_at), :desc)
+  end
+
+  def command_events_for_runs(runs, kind) do
+    Enum.map(runs, fn run ->
```
pepicrft: Opening the runner job page now maps every linked build or test run through `get_command_event_by_*`, and each helper performs its own ClickHouse lookup. That gives this request path one query per linked run, which can get expensive on workflow runs with many build or test records. Can we batch this into one query per kind using `where: build_run_id in ^ids` and `where: test_run_id in ^ids`, then group by run id in Elixir?

server/lib/tuist/runners/jobs.ex
```diff
@@ -83,6 +87,65 @@ defmodule Tuist.Runners.Jobs do
     |> Map.new()
   end
 
+  def project_for_runner_job(account, %{repository: repository}) when is_binary(repository) do
+    case String.split(repository, "/", parts: 2) do
+      [_owner, project_handle] ->
+        case Projects.get_project_by_account_and_project_handles(account.name, project_handle) do
```
pepicrft: `repository` is a GitHub repository handle, but this resolves the project by assuming the Tuist project name is the repository name. Existing repository connection helpers support multiple projects per repository, so monorepos or projects named differently from the repository would show the empty insights state even when build or test rows exist for this workflow run. Should this resolve through repository connections and then match the run rows by `project_id` instead?

pepicrft (approved):
Approved with two review comments. The focused server tests passed locally, and the remaining comments are scoped to query shape and repository-to-project resolution.

server/lib/tuist/runners/jobs.ex
fortmarek: Addressed in 78e679a16c. `Jobs.command_events_for_runs/2` now batches command-event lookups per kind instead of calling `get_command_event_by_*` once per linked run. The build path queries with `build_run_id in ^ids`, the test path queries with `test_run_id in ^ids`, both scoped to the involved project ids and ordered by `ran_at` / `created_at` so we keep the newest event per run while returning results in run order. I also added coverage for duplicate events and missing events.

server/lib/tuist/runners/jobs.ex
fortmarek: Addressed in 78e679a16c. `Jobs.projects_for_runner_job/2` now resolves candidate projects through the repository VCS connections for the selected account instead of assuming the repo slug equals the Tuist project name. Build/test run queries accept that project list and match by `project_id`, and the insight links use each run’s project id so monorepos or differently named projects route correctly. I added tests for multiple projects connected to the same repository and for an unconnected project with a matching repo slug returning not found.

PR #11725, opened 2026-07-08
feat(kura): heartbeat-driven mesh membership — dynamic peer authority, re-enroll reactivation, and full re-bootstrap on rejoin

### LINKED ISSUE #11724: Kura: stale self-hosted mesh peers are never pruned and poison node outboxes
# Summary

Self-hosted Kura nodes that enroll into an account's mesh are **never removed** when they disappear. The enrollment row (`account_cache_endpoints` with technology `kura_self_hosted_peer`) is inserted by `Tuist.Kura.Mesh.register_node_endpoint/2` and nothing ever deletes it — there is no deregistration path, no TTL, and no liveness requirement (unlike `kura_registered_endpoints`, which are lease-based and swept by `ExpiredRegistrationsWorker`).

A stale row has two consequences on every node in that account's mesh, one cosmetic and one that eventually takes writes down:

1. **Log spam**: the discovery loop (`kura/src/replication/mod.rs`, `membership_task_loop`) dials every peer's `/_internal/status` every ~2s, emitting `peer status request failed for <dead peer>` ~30×/minute per pod. In the incident below this was ~98% of the pods' log volume.
2. **Outbox poisoning → write shedding**: every artifact write enqueues one replication message per peer (`enqueue_replication_for_artifact`), including the dead one. Outbox messages are deleted **only on successful delivery** (`process_outbox`); there is no TTL and no pruning for targets that left the mesh. Undeliverable messages therefore accumulate monotonically in RocksDB, and once `outbox_depth() >= outbox_max_depth` (100,000 by default, `kura/src/http.rs`) the node **rejects all writes** with `server is shedding writes while replication catches up`. A dead mesh peer is a slow-fuse write outage for the whole tenant.

# Evidence (production, tuist account)

- On **2026-07-02 15:07 UTC**, a node enrolled into the `tuist` account's mesh as `https://selfhost-e2e-test.tuist.dev:7443` (a self-hosting e2e test instance). The row is still in `account_cache_endpoints` (technology `kura_self_hosted_peer`) six days later. The hostname **no longer resolves in DNS**.
- All production `kura-tuist-*` pods receive that URL in their peer list (managed pods via `KURA_PEERS` injected by `Tuist.Kura.Provisioner.KubernetesController`, self-hosted nodes via the enrollment response `peers:` from `Tuist.Kura.Mesh.enroll_node/2`).
- Each pod logs `peer status request failed for https://selfhost-e2e-test.tuist.dev:7443: error sending request` every ~2s — over a 5-day Loki window this is >1M log lines in the `kura` namespace, drowning the signal (it materially slowed down diagnosing an unrelated readiness incident today).
- Every artifact write in the tenant since 2026-07-02 has left one permanently-undeliverable outbox message per node. No `write_rejected_outbox` events have fired yet, but the depth only ever grows toward the 100k shedding threshold.

# Why nothing cleans it up today

- `register_node_endpoint` inserts with `on_conflict: :nothing` — even re-enrollment doesn't refresh any liveness marker.
- `kura_registered_endpoints` (CLI endpoint advertising) has heartbeats + lease expiry + a sweeper, but the mesh peer row is a different table with none of that.
- Kura's `process_outbox` retries a message forever; removing the peer from the mesh (manually deleting the row) stops *new* enqueues but the already-queued messages for the dead target are still retried and never deleted.

# Proposed fix (two halves)

**Server — drop unresponsive mesh members after X minutes:**
- Reuse the existing registration heartbeat (nodes heartbeat every 60s, keyed by `node_id` = host of the node's peer URL) as the liveness signal for the mesh row: bump the `kura_self_hosted_peer` row's `updated_at` on heartbeat (throttled) and on enrollment renewal.
- Add a periodic worker that deletes `kura_self_hosted_peer` rows that have not been touched for X minutes **and** have no unexpired registration lease. The existing `peers_revision_suffix` in the Kura Kubernetes controller already cascades the change: peer-list changes alter the manifest revision, so managed pods roll with the corrected `KURA_PEERS`.

**Kura — cascade: drop outbox messages for targets that left the mesh:**
- In `process_outbox`, delete messages whose target is absent from the node's current replication targets (static peers + enrolled peers + discovered members), after the target has been continuously absent for a grace window (to ride out enrollment races and membership flaps), and never when the target set is empty (control-plane-unreachable safety).
- This is safe by design: a peer that later rejoins the mesh re-bootstraps the full dataset on enrollment, so dropped deltas are recovered.

Alternatives considered: a per-message outbox TTL (rejected — silently breaks replication completeness for peers that are down-but-still-registered), and the server actively probing peers (rejected — the control plane can't reach mTLS mesh endpoints on private networks; the outbound heartbeat is the health signal by design).

Follow-ups worth considering separately: rate-limit the `peer status request failed` warning, and let self-hosted nodes refresh their peer list on a shorter cadence than certificate renewal (currently up to 15 days).

esnunes:
Fixes #11724.

## What

Self-hosted Kura peers now prove liveness with a **mesh heartbeat**; the control plane deactivates silent peers and purges them after the peer-certificate lifetime. Mesh composition is fully dynamic: static config (`KURA_PEERS`) carries only platform-stable managed endpoints, and all volatile membership — additions **and removals** — flows through the heartbeat channel on both self-hosted and managed nodes. Membership is only ever created or restored by enrollment: a deactivated node is told `mesh_member: false` and recovers by re-enrolling (backoff-limited), which also triggers a **full re-bootstrap**. Kura nodes drop queued outbox messages for peers absent from the now-authoritative dynamic view immediately (observable as `dropped_stale_target` replication results) — the safety margin lives in one place, the server-side staleness window.

The two heartbeats are independent, one per plane:

| Signal | Plane | Purpose | URL |
|---|---|---|---|
| Registration heartbeat (unchanged) | public | advertise the client-facing cache endpoint for CLI routing + dashboard | explicit `KURA_REGISTRATION_URL` |
| Mesh heartbeat (new) | peer | prove membership liveness + carry the current peer list | derived: `{KURA_CONTROL_PLANE_URL}/_internal/kura/mesh/heartbeat` |

Per deployment topology:

| Topology | Before | After |
|---|---|---|
| **Managed** (control-plane-provisioned regional / per-account instances) | Self-hosted peer view frozen into `KURA_PEERS` at pod start; a self-hosted peer enrolling or leaving required re-rendering the manifest and **rolling the fleet**, and a dead peer kept its slot until the next roll — dialed every ~2s, outbox growing toward write shedding. | Pods fetch the dynamic peer view from the control plane at boot (**serving gated on the first successful fetch**) and every ~60s; self-hosted enroll/leave propagates at fetch cadence with no roll; messages for departed peers are dropped on the next outbox pass. In-cluster siblings keep arriving via headless-DNS discovery and are never pruned. Liveness stays Kubernetes' job — managed pods never enter the membership state machine. |
| **Self-hosted, bridged** (enrolled into the account's managed mesh) | Enrollment was a one-way door: the membership row was permanent, so a torn-down node stayed in every peer's mesh forever; the peer list refreshed only at certificate renewal (up to 15 days) or reboot. | Nodes prove liveness with a mesh heartbeat every ~60s, whose response also refreshes the peer list; a silent node is deactivated after 30 minutes and purged after the 30-day certificate lifetime; a returning node recovers automatically via backoff-limited re-enrollment plus a full re-bootstrap of the data it missed. |
| **Self-hosted, standalone** (no control plane) | Operator-managed static `KURA_PEERS` + own peer TLS; registration heartbeat (if configured) advertises the endpoint to CLI clients. | Unchanged: no mesh rows exist, so the sweep and mesh heartbeat are no-ops; static peers are never pruned; registration keeps working as before. |

**Server**
- `POST /_internal/kura/mesh/heartbeat`: single-statement `UPDATE` bumps the liveness marker of *active* rows only and returns `{mesh_member, peers, heartbeat_interval_seconds}` (cadence server-tunable without a node release). Deactivated/purged/never-enrolled → `mesh_member: false`; heartbeats never create or restore membership.
- `GET /_internal/kura/mesh/peers`: peers-only read for managed pods (deployment-level credential + tenant, or self-hosted credential). Managed pods must not enter the membership state machine — Kubernetes owns their liveness.
- Enrollment response additionally carries `managed_peers` (platform-stable endpoints) so nodes seed only those statically; the enrollment upsert reactivates a deactivated row and is the recovery path.
- Sweep worker (`*/5` cron): deactivates rows whose marker is older than 30 minutes, and purges rows deactivated past the 30-day certificate lifetime. The staleness window is the mesh's *entire* safety margin (there is no node-side grace), so it stays many missed heartbeats wide. Both statements are single `UPDATE`/`DELETE`s with `RETURNING` (no read-then-write races with concurrent heartbeats/enrollments), backed by new partial indexes, and skipped while the app has been up for less than a staleness window so a control-plane restart cannot mass-deactivate live peers before their next heartbeat.
- The Kura Kubernetes provisioner sets `KURA_MESH_PEERS_SYNC` on mesh-enabled instances (manifest revision bumped). The self-hosted peers digest **stays** in the manifest revision for now — see Rollout.

**Kura**
- Enrolled nodes heartbeat every ~60s using the control-plane relationship carried on the enrollment outcome; `mesh_member: false` triggers a recovery re-enrollment with exponential backoff (60s→5min cap; the backoff only resets on an *observed* `mesh_member: true`, so a server that keeps answering `false` despite successful re-enrollments cannot drive an enroll/bootstrap-clear loop at heartbeat cadence), which hot-reloads the new leaf, restores membership, and **clears local bootstrap progress** so the membership loop re-pulls the full dataset from every peer. Routine cert renewals do not touch bootstrap state.
- Bootstrap completions are **epoch-guarded**: a pass captures the bootstrap epoch at start, and `reset_bootstrap_progress` bumps it — so a pass already in flight when a recovery lands cannot re-mark its peer bootstrapped on completion (it may straddle the absence window, missing writes behind its key-ordered manifest cursor); the stale completion frees the slot and the peer gets a fresh pass. A recovery also **takes the node out of serving** until every known peer completes a clean pass under the new epoch (readiness implies complete data — the recovered node is the one whose data actually is incomplete, and every other node already re-enters the gate through the membership generation change its rejoin causes); the reset re-arms the readiness settle window like a generation change does, so serving cannot be re-marked before the gate is genuinely re-evaluated. Long re-bootstraps are otherwise safe: heartbeats keep the liveness marker fresh throughout (no re-deactivation mid-pull), and live replication resumes within one heartbeat of reactivation, so the re-bootstrap only has to cover the bounded absence window.
- Managed pods (`KURA_MESH_PEERS_SYNC`) fetch the dynamic peer view at boot and on cadence; **serving is gated on the first successful fetch** — a pod booting blind would accept writes without enqueuing replication for peers it cannot see, a silent and unrecoverable under-replication.
- Failure semantics: a failed heartbeat/fetch keeps the last-known peer view (never store on failure); outbox pruning never runs on an empty view, and — because every pruning protection is process-scoped while the outbox is persistent — never before the process's view has actually arrived (the first peers sync where one is configured, plus one completed membership pass), so neither a degraded control plane nor a fresh restart can trigger destructive action. A 200 heartbeat response missing `mesh_member` fails the decode into the same safe path rather than defaulting to the destructive `false`.
- Outbox pruning is **source-aware**: messages for a control-plane-managed target absent from the fetched view are dropped immediately — the removal is deliberate (the server withholds a peer only after a full staleness window of missed heartbeats) and backed by the re-bootstrap safety net, so no node-side grace or config knob is needed. Peers known *only* through discovery (in-cluster siblings via headless DNS, cross-region pods via the account peer Service) are treated like the static seeds and **never pruned within a process lifetime** — their absence usually means a network flap, and nothing re-bootstraps them afterwards, so dropping their messages would be silent under-replication. The protection is process-scoped (in-memory history): a genuinely removed pod is never rediscovered after the observer's next restart, and since enqueues stop within one membership tick of unreachability, its small frozen backlog is dropped after the next deploy.
- The heartbeat HTTP client has connect/request timeouts so a black-holed request cannot stall the loop; peer lists are compared sorted so server-side row order never registers as a membership change; empty outbox passes skip the target-set rebuild.

## Why / root cause

Enrollment was a one-way door: rows were permanent, so a torn-down self-hosted node (`selfhost-e2e-test.tuist.dev`, enrolled 2026-07-02 in the tuist account's mesh) kept receiving dial attempts from every production `kura-tuist-*` pod every ~2s (~98% of their log volume) and every artifact write left one permanently-undeliverable outbox message per node, growing toward the `outbox_max_depth` write-shedding cap — a dead mesh peer was a slow-fuse write outage. Full evidence in #11724.

Two structural issues made a liveness signal alone insufficient:

1. **Boot-time peers were immortal.** Enrollment froze the full peer list into immutable `config.peers`, and the replication target set unions static and dynamic views — so a departed peer known at boot could never be pruned without a restart. Static config now carries managed endpoints only.
2. **A reactivated node never recovered its missed writes.** Replication targets are computed at write time, so writes made while a peer was out of the mesh were never enqueued for it — the delta (including namespace delete tombstones) provably exists nowhere as outbox messages, and no retention policy can recover it. Only a full re-bootstrap on the returning node reconciles the gap, which is why recovery goes through re-enrollment: it is the one unambiguous local signal "I was out for an unknown window".

## Why this design

- **Restart-on-composition-change was rejected**: the node that must react on rejoin never observes a composition change locally, restarts drop in-flight traffic and cold caches, and flap scenarios would restart fleets. Kura already hot-reloads certificates in-process; peers follow the same principle.
- **Re-enroll-to-reactivate** keeps enrollment as the single membership door (heartbeats, a lower-trust signal, can only sustain what enrollment created) and gives the mandatory re-bootstrap a clean trigger, distinguished from routine cert renewals.
- **A dedicated heartbeat rather than reusing registration**: registration is optional and public-plane (standalone nodes register without enrolling; enrolled nodes may not register), so it cannot be a universal liveness signal; a per-plane signal covers every enrolled node by construction.
- **Re-enrolling frequently instead of heartbeating was rejected**: enrollment bundles identity (CSR → CA-signed cert, a Kubernetes secret read per call) with presence; merging them makes control-plane availability load-bearing for mesh mTLS.
- **Alternatives for the outbox**: a per-message TTL was rejected — it breaks replication completeness for down-but-live peers; a node-side grace window before dropping was rejected — it adds a config knob and a second timing regime that only duplicates the server-side staleness window. The wide server window + the discovered-peer exemption + full-re-bootstrap-on-rejoin is what makes immediate dropping safe.

## Rollout

- Ships without compatibility gating: **verified against production during review** (re-check at merge) — 0 `kura_self_hosted_clients`, 0 registered endpoints, and exactly 1 `kura_self_hosted_peer` row: the stale test node this PR exists to clean up. The first sweep (once the app has been up a staleness window) deactivates it — intended.
- Wire changes are additive (serde defaults on the node for list/interval fields, extra response fields on the server). An old node image never calls the mesh heartbeat endpoint, so it receives no signal when withheld: it is deactivated after the staleness window and rejoins at its next enrollment (boot or certificate renewal).
- **Deliberately deferred**: removing the self-hosted peers digest from the managed manifest revision. Managed kura images roll independently of server deploys, so dropping the digest before the fleet runs a peers-fetching image would break removal propagation to managed pods. Sequencing: this PR ships the pod-side fetch; a follow-up drops the digest once the fleet image floor fetches peers.
- Coexistence of the two managed discovery channels is safe by construction: the peers endpoint returns self-hosted peers only (exactly what `KURA_PEERS` carried), while in-cluster siblings keep arriving via headless-service DNS.

## Validation

- **kura**: full `cargo test` — 303/303, including new coverage: pruning fires for a departed boot-time peer once the dynamic view drops it (the regression test for the immortal-peers defect), recovery clears bootstrap progress so peers re-enter `peers_needing_bootstrap`, stale-epoch bootstrap completions don't count while fresh ones do, the peer-view gate withholds serving until the first fetch and lifts after it, pruning deferred until the first membership pass and the first peers sync, a response missing `mesh_member` failing the decode instead of defaulting to the destructive value, recovery backoff gating/growth/reset, sorted peer comparison, static-seed managed/fallback selection, response decoding tolerance, empty-view-never-prunes behavior, and discovered-peer messages never being dropped. `cargo clippy --all-targets` and `cargo fmt --check` clean. (Bazel is the CI gate; not run locally.)
- **server**: 205/205 across the kura tree (mesh, registrations, workers, provisioner, reconciler, regions, deployment) + oban config + mesh controller: heartbeat bumps active rows only, deactivated row → `mesh_member: false` with the record untouched, re-enrollment reactivates, registration heartbeats have no effect on membership, sweep deactivate/purge via single statements, uptime guard skips young apps, peers endpoint under both credential shapes, enrollment `managed_peers` split. `mix credo` clean; `mix excellent_migrations.check_safety` clean (concurrent partial indexes, additive column).

## Impact

After deploy, the stale production row is deactivated within ~35 minutes of the app being up a staleness window (30-minute window + 5-minute cron); every node's next heartbeat/fetch drops the dead peer from its targets, the log spam stops, and its queued outbox messages are dropped on the next outbox pass. The row purges 30 days after deactivation. Self-hosted peer enroll/leave no longer takes up to 15 days to propagate to other self-hosted nodes (heartbeat cadence instead of certificate renewal), and once the digest follow-up lands, it will no longer roll the managed fleet either.


kura/src/mesh_heartbeat.rs
```diff
@@ -0,0 +1,467 @@
...
+async fn maybe_recover_membership(state: &SharedState, recovery: &mut RecoveryBackoff) {
+    if !recovery.should_attempt(Instant::now()) {
+        return;
+    }
+
+    warn!("control plane reports this node is not a mesh member; re-enrolling to recover");
+    match crate::enrollment::renew().await {
+        Ok(outcome) => match crate::app::apply_renewed_enrollment(state, &outcome).await {
+            Ok(()) => {
+                state.reset_bootstrap_progress().await;
```
pepicrft, line 274: Recovery needs to take the node out of serving before clearing bootstrap progress. Right now this path clears the bootstrapped peer set, but `readiness_report` still derives readiness from the existing serving flag, and `maybe_mark_serving` returns early once serving is true. That means a node that was withheld from the mesh can continue accepting traffic during the full re-bootstrap window, even though the comment above says it missed writes that were never enqueued for it. Please clear `serving` or otherwise move the runtime back to joining here, and cover it with a readiness assertion in this recovery test.

server/lib/tuist/kura/mesh.ex
```diff
@@ -152,36 +173,123 @@ defmodule Tuist.Kura.Mesh do
...
+  Both statements re-check their predicate atomically (single
+  `UPDATE`/`DELETE` with `RETURNING`), so a heartbeat or enrollment landing
+  concurrently is never overwritten by a stale read.
+
+  Deactivation cascades on its own: the peer disappears from enrollment and
+  heartbeat responses, so every node's next heartbeat drops it from the
+  dynamic peer set and its queued outbox messages are pruned after the grace
+  window.
+  """
+  def sweep_stale_self_hosted_peers(opts \\ []) do
```
pepicrft, line 247: This sweep is not safe during a rolling deploy. It deactivates self-hosted mesh rows from `account_cache_endpoints.updated_at`, but only the new mesh heartbeat refreshes that timestamp. Existing self-hosted nodes from the previous image only send registration heartbeats, and this change keeps those independent from mesh membership. Once the server worker has been up past the staleness window, those live older nodes can be deactivated without ever receiving `mesh_member: false`, so they cannot recover until restarted or manually re-enrolled. Please either let registration heartbeats refresh mesh liveness during the rollout, or gate sweeping to rows that have already proven support for mesh heartbeats.

pepicrft (commented):
Two issues that need addressing before this can merge.

server/lib/tuist/kura/mesh.ex
esnunes: The mesh heartbeat only deactivates self hosted nodes that join a managed mesh. At the moment this feature isn't used by customers, not requiring a smooth rollout / backwards compatibility

kura/src/mesh_heartbeat.rs
esnunes: Agreed — keeping serving here violated the gate's own invariant (readiness implies complete data), and the recovered node was the only one skipping the gate on its rejoin while every peer re-enters it via the generation change. Fixed in 2d1f151f4c: `reset_bootstrap_progress` now clears `serving` and re-arms the readiness settle window (mirroring a generation change), so serving can't be re-marked before the gate is re-evaluated — e.g. while `known_peers` is momentarily empty. Serving resumes once every known peer completes a clean pass under the new bootstrap epoch. The recovery test now asserts the full cycle: serving → recovery → not serving (and `maybe_mark_serving` refuses to re-mark) → serving again only after a clean epoch-valid pass.

fortmarek (approved):
This looks good, thanks!

esnunes:
> dropped 379473 outbox message(s) for https://selfhost-e2e-test.tuist.dev:7443: no longer a replication target

PR #11734, opened 2026-07-09
fix(kura,cli): harden #11719 follow-ups — serving-path race tolerance, retry classification, wildcard-inline cost edges

esnunes:
## What changed

Follow-ups from an adversarial post-merge review of #11719 (wildcard value-graph inlining + background segment promotion). Nothing in #11719 needed rolling back — its safety rails (budget-bounded inlining, mixed-version interop, unlink-not-truncate) held up under review — but the review confirmed a set of edges where those rails didn't reach:

**kura**
- **Promotion-race tolerance on every serving path.** #11719 made the promote-then-evict race routine (every old-segment read feeds the promotion worker, whose appends rotate the ring and evict Old segments) but wired the tolerant retry into only the two REAPI blob-read helpers. ByteStream `Read` and the HTTP cache GET still opened the stale manifest: a lost race surfaced as `INTERNAL` (ByteStream) or a spurious 404 (HTTP) for an artifact alive at its promoted location. Bazel papers over the ByteStream case with its own retries (`INTERNAL` is retriable there, and a retried Read re-fetches the manifest), but the HTTP path silently turned warm hits into recompiles. A new `Store::open_artifact_reader_range_tolerating_promotion` mirrors the bytes-path helper — re-resolve once on open failure, serve relocations from the fresh manifest, report genuine evictions as misses — and both remaining paths now use it. (The other two suspected paths were verified safe during review: accelerated file serving falls back through the router with a fresh manifest, and action results are inline in RocksDB.)
- **Promotion worker supervision + visibility.** The worker is the only background task with a producer queue behind it: one panic (pre-existing invariant `expect`s deep in the refresh path are reachable) used to silently disable promotion for the process lifetime while `enqueue_promotion` pinned up to 262k queued ids. The worker is now respawned on panic; promotion failures log at `warn` with a `kura_promotion_failures_total` counter; queue depth is exported as `kura_promotion_queue_depth`.
- **README**: documents the `inline_output_files: ["*"]` ActionCache extension (best-effort budget semantics, explicit-path hard error, mixed-version behavior, and the warm-client egress trade-off), per kura/AGENTS.md's protocol-doc rule.

**cas-plugin**
- **Precise retry classification.** Retrying all `Internal`/`Cancelled` absorbed kura's graceful h2 GOAWAY rotation, but kura maps genuine storage faults to trailer-borne `Internal`, and tonic maps the client's own 60s `Endpoint::timeout` to `Cancelled` — so a deterministically failing server got 3 back-to-back attempts from every concurrent chunk task, and a hung server cost 180s instead of 60s per RPC. `Internal`/`Cancelled` now retry only when the status's error-source chain downcasts to a local h2/hyper transport error; trailer-borne server statuses (which carry no source) and timeouts stay terminal. A unit test constructs a real GOAWAY(`NO_ERROR`) `h2::Error` and proves the transport case still classifies as retryable. Retries after the first (immediate) one now back off briefly.
- **Publish probes skip the wildcard.** `upload_process`/`publish` use `get_action` purely as an existence probe (they compare only the first entry's digest); with the unconditional `"*"` hint, every probe *hit* made the server materialize the whole value graph — 16-way concurrent reads, claims on the shared materialization pool, and up to 64MB of billed egress — for bytes the publisher immediately discarded. The probe hit is exactly the common concurrent-CI case the fast-skip exists for. A new `probe_action` (same lookup, no inline hint) restores pre-#11719 probe cost.
- **Decode-cap headroom.** The ActionCache client's `max_decoding_message_size` (64MB) exactly equalled kura's maximum inline budget, which counts content bytes but not the few bytes/file of protobuf framing — a maximally packed response could exceed the cap, and tonic's `OutOfRange` is (rightly) not retryable, turning exactly the largest cached graphs into permanent misses. The window is narrow (needs a payload within KBs of exactly 64MB) but deterministic when hit; the cap is now 96MB, restoring the headroom the batch path deliberately kept (32MB chunks vs a 256MB CAS cap).
- **AGENTS.md** updated to the inline-with-fallback read model, `probe_action`, and the retry classification, per the root intent-layer rule; `get_action`'s doc comment now states the warm-store egress trade-off explicitly (deliberate and accepted — cold-store resolves, the dominant remote-hit case, waste nothing and save a WAN round-trip each).

## Why / how this was found

A structured multi-agent review of #11719 (10 finder angles, adversarial verification of each candidate against the code, ~45 candidates → 14 verified findings). The items in this PR are the findings the review confirmed and triage accepted; several plausible-sounding candidates were refuted during verification and are deliberately **not** addressed here (e.g. zero-byte-frame conflation — frames are ≥8 bytes by construction; the empty-batch guard — actually load-bearing). Also consciously deferred: e2e (shellspec) coverage for the wildcard extension — there is no ActionCache e2e coverage at all today and the Rust unit tests cover the behavior in-process — and promotion-worker batching (coalescing refreshes under one lock/WriteBatch), which is a larger change worth its own PR if `kura_promotion_queue_depth` shows sustained lag in production.

## Validation

- kura: `cargo test --lib` — 297 passed, including new `tolerant_reader_reresolves_when_a_concurrent_promotion_evicted_the_old_segment` / `tolerant_reader_reports_a_miss_when_the_artifact_was_actually_evicted`; `cargo clippy --all-targets -- -D warnings` clean.
- cas-plugin: `cargo test` — 29 passed, including new retry-classification tests (`transport_borne_internal_and_cancelled_are_retryable` builds a real h2 GOAWAY error through tonic's status conversion; `server_sent_internal_and_cancelled_are_terminal`); `cargo clippy --all-targets` introduces no new warnings (42 pre-existing on main, 42 after).
- Rollout safety: all changes are node-local behavior (no wire/on-disk format change); mixed-version peers and old/new clients interoperate unchanged — the tolerant open only changes error → success/miss classification on the serving node, and `probe_action` is a request-shape reduction old servers already accept.


fortmarek (approved):
Thanks! lgtm

PR #11738, opened 2026-07-09
feat(server): surface CI details in run details

fortmarek:
## What changed

Adds a CI Details card to build and test run detail pages when the run can be associated with a Tuist runner job. The card surfaces the workflow, job, runner profile, queue time, CI job duration, and matched step, and links those values back into the Tuist dashboard runner workflow/job/step pages instead of sending users directly to GitHub.

The runner job detail page now accepts a step deep link, expands the matching step on load, scrolls to it, and drops the step query param after the initial expansion so later tab navigation does not keep re-opening the step. Local seeds also include believable runner/job/profile data so the build and test detail UI can be inspected locally.

## Why

Now that Tuist owns both the runner execution data and the build/test insights, the build and test details can connect those two surfaces directly. This gives users enough CI context on the run detail page while keeping the deeper investigation path inside the dashboard.

## Marketing changelog

Added a product changelog entry with screenshot: `server/priv/marketing/changelog/2026.07.09-ci-details-build-test-runs.md`.

## Screenshot

CI Details card on a build run detail page

## Approach

- Added a CI Details card and responsive styling for the build/test details layout.
- Added local build/test runner context lookup logic so each detail page can derive the matching workflow, job, profile, queue time, CI job duration, and step.
- Added runner lookup support in `Tuist.Runners.Jobs`.
- Updated build/test run LiveViews to assign and render the CI Details card only when a relevant runner job is found.
- Updated runner job LiveView to expand and scroll to a linked step while removing the one-shot `step` query param afterward.
- Updated gettext templates, tests, local seeds, and the marketing changelog entry for the new dashboard strings and fixture data.

## Validation

- `mix format`
- `mix gettext.extract`
- `mix test test/tuist/marketing/changelog_entry_parser_test.exs test/tuist_web/live/build_run_live_test.exs test/tuist_web/live/test_run_live_test.exs test/tuist_web/live/runner_job_live_test.exs`
- `mix test test/tuist/marketing/changelog_entry_parser_test.exs`
- After merging the latest `origin/main`: `mix test test/tuist/marketing/changelog_entry_parser_test.exs test/tuist_web/live/build_run_live_test.exs test/tuist_web/live/test_run_live_test.exs test/tuist_web/live/runner_job_live_test.exs` (`59 passed`)
- Review follow-up: `TUIST_SERVER_CLICKHOUSE_RUNTIME_DIR=/tmp/tuist-clickhouse-ci-details-review mix test test/tuist_web/live/build_run_live_test.exs test/tuist_web/live/test_run_live_test.exs test/tuist_web/live/runner_job_live_test.exs` (`58 passed`)
- Captured only the CI Details card from the seeded build detail page at `http://localhost:8208/tuist/tuist/builds/build-runs/01910000-0000-7000-8000-000000000001` with headless Chrome and verified the card rendered with 3 columns.


pepicrft (commented):
Went through this closely. Two correctness issues I think are worth fixing before merge, plus two small cleanups.

## The deep-linked step never stays collapsed

`server/lib/tuist_web/live/runner_job_live.ex:850`

`maybe_expand_step` runs on every `handle_params`, and the `step` param survives live patches because the tab links build their target with `Query.put(@uri.query, "tab", ...)`.

So arriving from a CI Details deep link at `?tab=overview&step=5`, collapsing step 5, scrolling away, then clicking Metrics and back to Overview re-expands step 5 and pushes `scroll-to-element` again, yanking you back down to it. The collapse can never stick.

Reading the param once on mount, or dropping it from the query after the first expand, would fix it.

## `List.first(jobs)` can surface an unrelated job as the CI context

`server/lib/tuist_web/live/build_run_live.ex:192` and `server/lib/tuist_web/live/test_run_live.ex:178`

Both fall back to `List.first(jobs)` when no job name matches the needle. A workflow whose jobs are named `lint` and `deploy` still renders the CI Details card, populated from the earliest-started, entirely unrelated job. The Workflow/Job/Profile fields, the durations, the "View more" link and the step deep links all then point at the wrong job while claiming to be the CI context for this run.

Returning `nil` and hiding the card when nothing matches seems more honest than showing a confidently wrong one.

## Cleanups

The four CI helpers in `test_run_live.ex` (`normalize_ci_provider/1`, `parse_workflow_run_id/1`, `find_ci_job_by_name/2`, `find_ci_step_by_name/2`) are byte-for-byte copies of the ones in `build_run_live.ex`, and the two `with` context builders differ only in the name needle. Both live views already import `TuistWeb.Runs.CIContextCard`, so a single parameterized builder there would drop about thirty lines that will otherwise drift apart.

Minor: the `"tests"` needle in `["test", "tests"]` is dead, since `String.contains?(name, "test")` already matches anything containing `"tests"`.

pepicrft (approved):
Approving so this isn't blocked. Would still like the `List.first(jobs)` fallback and the step deep-link re-expand sorted, either here or in a follow-up.

pepicrft (approved):

PR #11739, opened 2026-07-09
docs(kura): propose range-digest anti-entropy to make bootstrap O(delta)

fortmarek:
## Why

Diagnosing why the tuist `eu-central` region wouldn't reach readiness after a reprovision, the final root cause turned out to be bootstrap performance, not connectivity: eu-central's `kura` log showed

```
bootstrap from scw-fr-par-0 failed: bootstrap timed out after 1800000 ms
```

Bootstrap walks a peer's **entire** manifest set (`/_internal/bootstrap/manifests`, 256/page, cursor-ordered) and pre-checks every artifact before it can serve — **O(peer dataset)** regardless of how in-sync the pair already is. The two nodes are ~99% in sync (~1.4M of ~1.4M artifacts already local) yet each pass still fetches **~5,652 manifest pages** (~15 min just to walk, measured) + ~1.4M local pre-checks, and never finishes within `KURA_BOOTSTRAP_TIMEOUT_MS` (30 min). Both nodes are same-metro (Paris) with 512 MB/s replication configured — so it's pure **per-artifact overhead**, not bytes or WAN. Production telemetry also shows the bootstrap **stalling** (`applied` flat) while live `replicate/artifact` runs at ~28/s, i.e. a likely second, orthogonal issue: live writes starving the bootstrap on the segment-append lock.

## What

A design proposal (`kura/docs/proposals/bootstrap-range-digest-anti-entropy.md`), no code:

**Range-based set reconciliation** over the `artifact_id`-sorted manifests CF. Both nodes summarize their manifest keyspace as per-prefix-bucket `(count, hash)` digests (hash folds `(artifact_id, version_ms)` so version bumps are caught); the joining node diffs digests, **skips matching buckets**, and enumerates **only divergent ranges** with the existing walk. Collapses thousands of serial page round-trips into one digest exchange + a few small range walks. Additive endpoint, negotiated with graceful fallback to today's full walk, mixed-version safe.

The doc also records the two complementary levers surfaced by the incident: a resumable manifest cursor (smaller, largely subsumed) and bootstrap-vs-live-replication **lock fairness** (orthogonal — recommend confirming with a profile before implementing).

## Status / next

Landing the design for review first. Happy to implement in a follow-up: `manifests_digest()` + range-scoped `manifests_page` in `store.rs`, the `/_internal/bootstrap/digest` endpoint + negotiation, the digest-diff-then-scoped-walk in `bootstrap_manifests_from_peer`, and tests (identical stores → zero ranges walked; single divergent artifact → one bucket; old-peer fallback).

### DIFF
--- a/kura/docs/proposals/bootstrap-range-digest-anti-entropy.md
+++ b/kura/docs/proposals/bootstrap-range-digest-anti-entropy.md
@@ -1,0 +1,129 @@
+# Proposal: range-digest anti-entropy for bootstrap
+
+## Problem
+
+When a node (re)joins the mesh it **bootstraps** each peer: it walks the peer's
+entire manifest set page by page (`/_internal/bootstrap/manifests`, 256/page,
+cursor-ordered by `artifact_id`), pre-checks every manifest against its local
+store, and fetches the ones it is missing. A peer is marked bootstrapped — and
+the node allowed to serve — only after one fully clean pass.
+
+This is O(peer's whole dataset) **regardless of how in-sync the two nodes
+already are**. In production the tuist `eu-central` ↔ `scw-fr-par` pair is ~99%
+in sync (the joining node already holds ~1.4M of the peer's ~1.4M artifacts),
+yet each bootstrap still:
+
+- fetches **~5,652 manifest pages** serially over the peer link (~15 min just to
+  walk, measured), and
+- runs **~1.4M local pre-checks**,
+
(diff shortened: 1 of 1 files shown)


fortmarek:
Closing the proposal — implementing the range-digest bootstrap directly per discussion. The design in this branch is carried into the implementation PR.

PR #11742, opened 2026-07-09
fix(infra): make kubectl logs/exec work on bare-metal Kura fleet nodes

fortmarek:
## What

Makes `kubectl logs` / `exec` / `attach` / `port-forward` work against the in-house-CAPI bare-metal Kura fleet nodes (Dedibox eu-central, Scaleway Elastic Metal `scw-fr-par`, OVH `us-east`/`us-west`), replacing the ad-hoc hostPath log-reader workaround.

The apiserver dials the kubelet on `:10250` directly. For these self-joined nodes that path is broken at **three independent layers**, each with a distinct symptom — diagnose by the error, not by guessing:

| `kubectl logs <fleet-pod>` symptom | Layer | Fix |
|---|---|---|
| `dial tcp: lookup <node> … no such host` | 1 — apiserver dials the unresolvable Hostname (`--kubelet-preferred-address-types` lacks `InternalIP`) | roll the control plane |
| `remote error: tls: internal error` | 2 — kubelet has no serving cert (`serverTLSBootstrap` + unapprovable ServiceAccount CSR) | re-provision the node (code fix here) |
| `401` / anonymous | 3 — kubelet has no `clientCAFile`, apiserver client cert untrusted | re-provision the node (code fix here) |

## Why / root cause (confirmed live against `tuist-k8s-production`, read-only)

**Layer 1 — address selection.** Running apiserver flag is `ExternalIP,Hostname,InternalDNS,ExternalDNS` (no `InternalIP`). Fleet nodes report a routable `InternalIP` (provider public IP) but no `ExternalIP` and a Hostname the Hetzner CP can't resolve, so the dial lands on Hostname → `no such host`. The InternalIP value is **already** in each Cluster CR (`ExternalIP,InternalIP,Hostname,…`, #11375) and the apply succeeded — but kubeadm bakes the apiserver flag into the static-pod manifest at CP init/join, and the running CP node predates the change and was never rolled. So this layer is a **control-plane rollout of already-correct config**, not a file change.

**Layer 2 — no serving cert.** With InternalIP preferred, the dial *reaches* the kubelet: metrics-server gets `remote error: tls: internal error` from the nodes' public IPs (**not** a timeout), so `:10250` is reachable and it's a TLS-layer failure, not a firewall. The self-join kubelet ran `serverTLSBootstrap: true`, but it authenticates as an operator-minted **ServiceAccount**, not `system:node:<name>`. The built-in serving-CSR approver only approves when requester == CSR CN == `system:node:<name>`, so a ServiceAccount requester never matches; the CSR stays pending and the kubelet presents no cert. (Hetzner nodes use real `system:node` identities, so theirs work.)

**Layer 3 — no client-CA trust.** Even with a serving cert, exec/logs would `401`: the self-join kubelet config omitted `authentication.x509.clientCAFile`, so the apiserver's `--kubelet-client-certificate` can't be verified → request is anonymous → denied. The ClusterClass sets this exact field on Hetzner nodes for the same reason.

## The fix

`controllers/linux/linux_cloudinit.go` (both the cloud-init and SSH-script forms), wired through the Dedibox / OVH / Elastic Metal controllers which already hold the node-identity CA:

- **Layer 2:** leave `serverTLSBootstrap` unset → the kubelet self-signs its `:10250` cert. Nothing verifies kubelet serving certs on this cluster (apiserver has no `--kubelet-certificate-authority`; metrics-server runs `--kubelet-insecure-tls`), so self-signed is accepted. **No enforced security is lost** — even the Hetzner nodes' CA-signed certs go unverified today; this only forgoes a future *option* to verify, which is already blocked cluster-wide by these same nodes.
- **Layer 3:** write the cluster CA to `/var/lib/kubelet/ca.crt` and set `clientCAFile`.

### Why self-signed over a CSR-approver
Discussed and chosen deliberately: the cluster verifies no kubelet serving certs today, so self-signed matches the existing posture with zero new workloads/RBAC. A `kubelet-csr-approver` would keep CA-signed certs but adds a cluster component and must special-case the non-`system:node` ServiceAccount requester. Revisit if kubelet serving-cert verification ever goes on the roadmap (most defensible here since these are public-IP nodes).

## Impact / rollout

- **Hetzner nodes unaffected** — they keep `serverTLSBootstrap: true` + their real `system:node` identity; this touches only the self-join linux path.
- **Layers 2 & 3 converge onto existing nodes automatically** — a second commit adds a narrow, zero-downtime **kubelet-config drift loop** for the Linux kinds (they previously had no re-push, unlike the macOS fleet). Each already-Ready node carries a `tuist.dev/kubelet-config-hash` annotation; when it doesn't match the current rendered config the controller SSHes a minimal re-push (rewrite `/var/lib/kubelet/{ca.crt,config.yaml}` + `systemctl restart kubelet`) — containerd/apt/`/data` mounts untouched, so running pods survive. So on **merge → prod cascade** (which builds + deploys the provider image), the fix lands on the four existing Dedibox/Scaleway-EM/OVH nodes on the next reconcile, **with no `kubectl delete machine`**. Re-provision still works as a force/fallback.
- **Layer 1 (apiserver flag) still needs a one-time control-plane rollout** for existing clusters — the `ExternalIP,InternalIP,Hostname,…` value is already in the Cluster CRs, but kubeadm bakes the apiserver flag at CP init/join, so a running CP that predates it keeps the old flag. New clusters get it by default. This is **not** auto-rolled from CI on purpose: rolling a (single-node) prod control plane as a side effect of a shared `infra/k8s/clusters/**` apply is too high-blast-radius to do from an untested bash step — it stays a deliberate operator action (runbook in `infra/cluster-api-provider-tuist/AGENTS.md`).
- Full runbook — per-layer symptom→fix table, the drift-loop convergence check, the CP-rollout commands, and end-to-end verification — is in `infra/cluster-api-provider-tuist/AGENTS.md`. The `docs/scaleway-elastic-metal-support.md` cross-cutting note (which previously claimed the address-types change alone "is what lets the apiserver reach those kubelets") is corrected to the full three-layer account.

## Validation

- `go build`, `go vet`, `go test ./...` pass.
- `TestRenderLinux_SelfSignedServingCertAndClientCA`: asserts `serverTLSBootstrap` is unset, `clientCAFile` + the on-disk CA are present and the CA is written before the kubelet config, and that a caller passing no CA renders byte-identically to before (both render forms).
- Drift-loop tests: the re-push script writes `ca.crt` before `config.yaml` + restarts kubelet and carries **none** of the disruptive bootstrap steps (`apt`/`containerd`/mounts/`swapoff`); the hash is stable and reflects the post-fix config; a stamped node is a no-op.
- Root cause verified live (read-only): the Hostname dial failure, the `tls: internal error` reachability signal, the missing apiserver flag, and the ServiceAccount identity model.


pepicrft:
Did a deep pass through the Linux bare-metal bootstrap and drift reconciliation. The implementation is coherent, and I validated it with:

- `mise exec go@1.25.0 -- go test ./...`
- `mise exec go@1.25.0 -- go test -race -count=1 ./controllers/linux`
- `mise exec go@1.25.0 -- go vet ./...`

All checks passed.

I found three non-blocking follow-ups worth addressing:

1. `desiredKubeletConfigHash()` does not include the current certificate authority bundle. Once a node is stamped, an authority rotation cannot cause `ca.crt` to be rewritten.
2. The drift repair builds host-key state but does not persist a newly observed fingerprint before stamping the node. If the original pin is missing, the repair trusts the first key without restoring the pin.
3. `AGENTS.md` says the dedicated boxes are left installed and are not reinstalled on release, while both deletion paths wipe and reinstall them.

I am approving because the immediate logs and exec repair is sound and the full provider suite and race checks pass. I would still capture the first two as follow-up fixes and correct the runbook.

pepicrft (approved):
The immediate repair is sound, and the focused, full, static, and race checks all pass. The follow-up observations are captured in the pull request conversation.

fortmarek:
Thanks for the deep pass @pepicrft. Addressed all three follow-ups in cf392f18e4:

1. **CA bundle in the drift hash.** `desiredKubeletConfigHash` now hashes `config.yaml` + the cluster CA bundle. The reconcile reads the node identity up front (idempotent, a few cached reads on the converged path) and folds `identity.CA` into the hash, so a CA rotation flips the hash and re-pushes a fresh `ca.crt` onto already-stamped nodes instead of trusting a stale one forever.

2. **Persist a newly observed fingerprint before stamping.** After the SSH, a host key TOFU'd on a first-contact box is now persisted via `SetMachineHostFingerprint` before the stamp, matching the bootstrap path. This is pin-safe: `HostKeyState.Observed()` returns the *pinned* value on a mismatch (the callback doesn't set `captured` when it returns `ErrHostKeyMismatch`), so `observed != known` is false there and a good pin is never overwritten. Runs even on SSH failure, since the key was still observed.

3. **Runbook correction.** Both the Dedibox and OVH `reconcileDelete` paths reinstall (wipe) the box back into the pool on release, so I fixed the two summary bullets and the "Release" paragraph that claimed the box was left installed with OS + key intact.

Validation: `go build ./...`, `go vet ./controllers/linux/...`, and `go test -race -count=1 ./controllers/linux/...` all pass. The no-op-when-stamped test now uses a fake-client `Manager` (the converged path reads the identity now), plus a new assertion that a CA change flips the hash.

PR #11743, opened 2026-07-09
feat(server): build-systems blog post with mermaid, admonitions, and device frames

pepicrft:
This adds the first post in a series exploring what a build system designed for coding harnesses could look like, together with the marketing rendering support the post needs.

The post starts from a simple observation, that every build system is the same graph underneath, and then builds a real example on top of it: a Rust core compiled for Apple and Android, with a Swift app and a Kotlin app that call into it, all built from plain shell scripts with a few `# once` annotations rather than any build DSL. It uses that example to show caching and early cutoff, then turns to what a naive graph does not capture, the toolchain it depends on, and closes on the questions we are chewing on around remote execution, reusability, authentication and attribution for agent clients, and the client that sits between a harness and the infrastructure. The companion example project lives at [tuist/once-example](https://github.com/tuist/once-example) and is linked from the post.

To render it, the marketing blog pipeline picked up a few reusable capabilities:

- Mermaid diagrams, with the runtime loaded only on pages that contain one and themed for light and dark.
- Admonitions through GitHub-style alert syntax (`> [!NOTE]` and friends), compiled at build time into the Noora alert component so callouts match the design system rather than a bespoke style.
- Side-by-side device frames for the app screenshots, done in CSS so they stay crisp and theme-independent.

There is also a rendering fix. `content_html/1` re-serializes each post through Floki to optimize images, and Floki drops whitespace-only text nodes on round-trip, which was collapsing the indentation inside highlighted code blocks. The code-heavy post made it obvious. The fix protects `<pre>` blocks across the Floki pass and restores them verbatim.

The two dependency lockfiles that were dirty in the worktree (`Package.resolved`, `server/mix.lock`) are unrelated to this work and were intentionally left out.

### How to test locally

Run the marketing site (`mise run dev` from `server/`) and open `/blog/2026/06/30/three-build-systems-one-graph`. Check that:

- The mermaid diagram near the top renders in both light and dark.
- The three `NOTE` callouts render as Noora alerts, including the one with the `pedro@tuist.dev` link.
- The shell and Rust code blocks keep their indentation.
- The two app screenshots show side by side inside iPhone and Android frames.


server/priv/marketing/blog/2026/06/30/three-build-systems-one-graph.md
```diff
@@ -0,0 +1,219 @@
...
+# once needs "./build-core-apple.sh"
+# once input "../apple/Sources/**/*"
+# once input "../core/include/core.h"
+# once fingerprint "swiftc --version"
+# once env "PATH"
+# once output "../apple/MyApp"
+# once cwd ".."
+set -euo pipefail
+
+sdk="$(xcrun --sdk iphonesimulator --show-sdk-path)"
```
fortmarek, line 156: I guess the version of `xcrun` and/or the sdk path should be ideally also modeled as inputs

fortmarek (approved):
What a great write-up 👏 Just a couple of nits 👇 

Image

The diagram strips a part of the labels.

The shadows of screenshots bleed into the text which I'd try to avoid:

Image

server/priv/marketing/blog/2026/06/30/three-build-systems-one-graph.md
pepicrft: Good catch. Added a second fingerprint for the SDK build version, so an Xcode bump invalidates the binary the same way a compiler bump does, plus a line in the post explaining why. It's exactly the kind of uncaptured input that section is about, so it reads better with it in.

pepicrft:
Thanks for the review 🙏 Both nits fixed:

- **Diagram labels.** The clipping was mermaid measuring label widths before Inter had loaded, so it sized the node boxes against a fallback font and the text overflowed once the real font swapped in. It now waits for `document.fonts.ready` before rendering, and I switched the graph to top-down so it fits the content column. Added a CSS safety net too so the SVG scales instead of overflowing.
- **Screenshot shadows.** Gave the device frames vertical margin and toned the shadow down so it stays off the surrounding text.

Preview is redeploying, should be up in a few minutes.

PR #11748, opened 2026-07-10
feat(server): add interactive runner shell access

fortmarek:
## What changed

- Added interactive shell access for runner jobs, including the server-side session lifecycle, browser/client WebSocket bridge, runner-agent WebSocket bridge, and shell broker wiring.
- Added runner-side shell agents for Linux and macOS runner images so a live job can attach to a PTY-backed shell.
- Added `tuist runner ssh <job-ref>` support in the CLI, including session creation and terminal client plumbing.
- Split the dashboard interactive surface into dedicated Terminal and VNC tabs, with terminal session request/start states and a Noora-aligned terminal surface.
- Added terminal rendering support for common PTY control sequences so erases, cursor movement, Backspace, Option+Backspace, and Command+Backspace behave like a terminal instead of leaking raw escape text.
- Added local dashboard seed data and a dev-only terminal simulation path on real runner job pages so UI and keyboard interactions can be tested locally without staging.
- Updated staging smoke/deploy workflow pieces so stopped Linux and macOS runner jobs can be dispatched and probed end to end.

## Why

VNC access gives visual debugging for macOS runners, but it does not cover the common need to quickly inspect a live runner from a shell, especially on Linux. This adds the complementary terminal path in both places people need it: directly from the dashboard when looking at a job, and from the CLI with `tuist runner ssh <job-ref>`.

The dashboard also needed Terminal and VNC to be separate concepts. Treating both as one generic interactive tab made the UI ambiguous and made local iteration awkward. The dedicated Terminal tab keeps the workflow direct while preserving VNC for macOS visual access.

## Approach

- Shell sessions reuse the existing interactive-session model with a new shell kind, token validation, active/disconnect tracking, and a small PubSub broker that connects browser clients to runner agents.
- Browser clients connect over an authenticated WebSocket using the session token; runner agents connect over the internal tunnel endpoint and relay PTY stdout/stdin plus resize and exit/status messages.
- The CLI creates a shell session for the requested job ref, opens the WebSocket endpoint, configures terminal raw mode locally, forwards stdin/resizes, and restores terminal state on exit.
- The dashboard renders a Terminal tab only when shell access is available, or when the local dev simulation flag is present. The terminal renderer now keeps a small screen model instead of appending raw bytes to a preformatted block.
- The local simulator intentionally stays lightweight. It is for UI and keyboard checks, not a full shell, but it supports common commands like `echo`, `pwd`, `ls`, `whoami`, `date`, and `clear` so the preview is not misleading.

## Bugfix details

- The initial browser terminal renderer appended raw PTY output directly, so shell editing sequences such as erase-line leaked as visible `[K` text and Backspace could leave control characters in the output. The renderer now interprets common control sequences and updates a screen buffer before rendering.
- The cursor previously blinked even after the terminal lost focus. It now only shows the active block cursor when the terminal itself is focused, matching the focus border.
- The local simulator previously reported `command not found` for simple `echo` usage. It now handles `echo` so local UI testing feels reasonable without pretending to be a full local PTY.

## Impact

- Authorized account members can open a live terminal for supported running runner jobs from the dashboard.
- CLI users can attach to a live runner shell with `tuist runner ssh <job-ref>`.
- Local development can exercise the actual runner job dashboard page with `?tab=terminal&terminal_simulation=1`, backed by seeded running Linux runner data.

## Validation

- `git diff --check`
- `mise exec node@24.14.0 -- node --check server/assets/app/js/RunnerShellTerminal.js`
- `mix test test/tuist_web/live/runner_job_live_test.exs test/tuist_web/controllers/api/runner_interactive_shell_session_controller_test.exs`
- `mix assets.build`
- `go test ./internal/podtemplate` from `infra/runners-controller`
- Local dashboard browser check on seeded running Linux job `4300002`: verified plain Backspace, Option+Backspace, Command+Backspace, `echo "hello"`, tighter terminal layout, and inactive cursor state when the terminal is blurred.

Not run locally: Xcode CLI build/test. The generated Xcode workspace is not present in this worktree, and this is opened as a draft for review plus CI validation.


pepicrft (commented):
I left inline comments on the validated runner shell findings.

infra/runners-controller/internal/podtemplate/podtemplate.go
```diff
@@ -349,6 +352,37 @@ func Build(pool *tuistv1.RunnerPool, podName, saName, dispatchURL, dispatchInter
...
+			Image:   pool.Spec.Image,
+			Command: []string{"sh", "-c"},
+			Args: []string{
+				"sed -i '/runner ALL=(ALL) NOPASSWD/d' /etc/sudoers || true; exec /usr/local/bin/runner-shell-agent",
+			},
+			Env:           shellEnv,
+			RestartPolicy: ptr(corev1.ContainerRestartPolicyAlways),
+			VolumeMounts: []corev1.VolumeMount{
+				{Name: "tuist-runner-token", MountPath: "/var/run/secrets/tuist-runner", ReadOnly: true},
+				{Name: "tuist-runner-jit", MountPath: jitMountPath, ReadOnly: true},
```
pepicrft, line 394: The Linux terminal is created in this `shell` sidecar, whose mounts stop at the token, claim file, and work volume. Because the `runner` container has a separate root filesystem, process namespace, and environment, the advertised "shell for a running job" cannot inspect the running process, job-level environment variables, packages installed by the workflow, or the Docker socket. That makes the debugging surface materially different from the job it claims to attach to. Can we keep the authenticated relay here but start the user-facing pseudoterminal inside the runner container, for example through a narrow shared local socket?

server/lib/tuist_web/websocks/runner_shell_client_web_sock.ex
```diff
@@ -0,0 +1,160 @@
...
+          reason: inspect(reason),
+          connection_id: connection_id
+        )
+
+        {:stop, {:session_activate_failed, reason}, {1011, "session unavailable"}, %{}}
+    end
+  end
+
+  @impl WebSock
+  def handle_in({payload, [opcode: :binary]}, %{session: session} = state) do
```
pepicrft, line 60: These inputs are accepted against the in-memory session forever after the handshake. The expiration worker only closes the database row, and neither endpoint schedules an absolute timer from `expires_at`; the 65-minute WebSocket setting is inactivity-based and active terminal traffic resets it. That lets an active shell remain usable after the one-hour grant expires. Please add an absolute expiry timer that closes both endpoints and a test that proves input stops at the boundary.

infra/linux-runner-image/cmd/runner-shell-agent/main.go
```diff
@@ -0,0 +1,915 @@
...
+		request.Header.Set("X-Tuist-Runner-Pod-Name", podName)
+	}
+	if podUID := os.Getenv("TUIST_RUNNER_POD_UID"); podUID != "" {
+		request.Header.Set("X-Tuist-Runner-Pod-Uid", podUID)
+	}
+	if pool := os.Getenv("TUIST_RUNNER_POOL"); pool != "" {
+		request.Header.Set("X-Tuist-Runner-Pool", pool)
+	}
+
+	response, err := directHTTPClient().Do(request)
```
pepicrft, line 261: `directHTTPClient()` creates a new client and custom transport on every two-second poll. Because the transport leaves `IdleConnTimeout` at zero, each successful keep-alive response can retain an idle connection and goroutines until the peer closes it. Reuse one client for the agent lifetime and close idle connections on shutdown. The identical macOS agent implementation needs the same change.

server/lib/tuist_web/websocks/runner_shell_agent_web_sock.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  @impl WebSock
+  def terminate(_reason, %{session: session}) do
+    Logger.info("runners: shell runner disconnected",
+      session_id: session.id,
+      workflow_job_id: session.workflow_job_id,
+      pod_name: session.pod_name
+    )
+
+    :ok = InteractiveShellBroker.broadcast_to_client(session.id, :runner_disconnected)
```
pepicrft, line 86: On the normal `shell_exit` path, this callback runs after the real `runner_exit` was already broadcast and immediately reports `runner_disconnected`. The browser then moves back to waiting and, ten seconds later, emits a second exit with status 255, overwriting the real status. Please suppress this notification after a reported exit, or close the client endpoint immediately after forwarding the real exit. A lifecycle test for zero and nonzero exits would protect the ordering.

cli/Sources/TuistRunnerCommand/Services/RunnerShellTerminalClient.swift
```diff
@@ -0,0 +1,201 @@
...
+
+        if ioctl(STDOUT_FILENO, request, &size) == 0, size.ws_col > 0, size.ws_row > 0 {
+            return TerminalSize(columns: Int(size.ws_col), rows: Int(size.ws_row))
+        }
+
+        return TerminalSize(columns: 80, rows: 24)
+    }
+}
+
+private struct RunnerShellControlMessage: Decodable {
```
pepicrft, line 140: The runner sends an exit status, but this type drops it and `shouldClose` reduces the control frame to a Boolean. Therefore `exit 42` still makes the local `tuist runner ssh` process return success. Please decode and propagate `status` so a nonzero remote exit becomes the command's process status, with Swift Testing coverage for zero, nonzero, malformed, and unrelated messages.

server/lib/tuist_web/live/runner_job_live.html.heex
```diff
@@ -640,18 +646,94 @@
     </.card>
   </div>
 
-  <div :if={@selected_tab == "interactive"} data-part="tab-panel">
+  <div :if={@selected_tab == "terminal"} data-part="tab-panel">
```
pepicrft, line 649: This is a new customer-facing dashboard capability, but the pull request does not add an entry under `server/priv/marketing/changelog/`. The server review guidance asks for a marketing changelog entry for features like this. Please add one with the required frontmatter and a terminal screenshot.

cli/Sources/TuistRunnerCommand/Services/RunnerShellTerminalClient.swift
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    func restore() {
+        guard !restored else { return }
+        tcsetattr(STDIN_FILENO, TCSANOW, &original)
+        restored = true
+    }
+}
+
+private final class StandardInputForwarder {
```
pepicrft: These two helpers do not need identity, inheritance, mutable stored state, or deinitialization. Please make `StandardInputForwarder` and `TerminalResizeObserver` structures, in line with the Swift review guidance. `TerminalRawMode` can remain a class because its deinitializer restores terminal state.

infra/runners-controller/internal/podtemplate/podtemplate.go
fortmarek: Addressed in the latest push. The authenticated relay stays in the `shell` sidecar, but the user-facing PTY now starts from the runner container through a narrow shared Unix socket (`TUIST_RUNNER_SHELL_SOCKET`). The runner container starts the PTY server before `run-job.sh`, so terminal sessions see the job container filesystem, process namespace, environment, installed packages, and Docker socket while the sidecar still owns server auth. I also updated the pod template tests and Linux runner image docs for that split.

server/lib/tuist_web/websocks/runner_shell_client_web_sock.ex
fortmarek: Addressed. The client WebSocket now schedules an absolute timer from `session.expires_at`; when it fires, it closes the DB session, notifies the runner endpoint, emits an exit frame to the browser/CLI, and closes the socket. I added WebSock coverage proving input is stopped at the expiry boundary.

infra/linux-runner-image/cmd/runner-shell-agent/main.go
fortmarek: Addressed for both Linux and macOS agent sources. `directHTTPClient()` now returns a process-lifetime client backed by a shared transport with `IdleConnTimeout`, and the agent closes idle direct-HTTP connections on shutdown instead of allocating a fresh transport per poll.

server/lib/tuist_web/websocks/runner_shell_agent_web_sock.ex
fortmarek: Addressed. The agent WebSocket state now tracks whether a shell exit was already reported; after forwarding the real `runner_exit`, `terminate/2` suppresses the follow-up `runner_disconnected` notification. I added lifecycle tests for both zero and nonzero exits so the browser/CLI keep the real status instead of falling back to 255.

cli/Sources/TuistRunnerCommand/Services/RunnerShellTerminalClient.swift
fortmarek: Addressed. The CLI terminal client now decodes `exit` control frames and throws `RunnerShellRemoteExitError` for nonzero remote statuses; `TuistCommand` maps that error to the process exit code without showing a generic error card. I added Swift Testing coverage for zero, nonzero, malformed, and unrelated messages, and enabled the `TuistRunnerCommandTests` target.

server/lib/tuist_web/live/runner_job_live.html.heex
fortmarek: Addressed. I added `server/priv/marketing/changelog/2026.07.15-runner-terminal-access.md` with the required frontmatter plus a dashboard terminal screenshot under `server/priv/static/marketing/images/changelog/2026.07.15-runner-terminal-access.png`.

pepicrft (approved):
I re-reviewed the addressed comments and verified the focused Go, server, and Swift tests. All required checks are green.

PR #11749, opened 2026-07-10
docs(server): add Tuist Runners documentation

fortmarek:

Adds product documentation for **Tuist Runners** — the managed macOS and Linux GitHub Actions runners — plus a deep-dive page on runner profiles, and registers both under a new "Runners" sidebar group.

### What changed

- `server/priv/docs/en/guides/features/runners.md` — the main Runners page.
- `server/priv/docs/en/guides/features/runners/profiles.md` — profiles deep-dive.
- `server/lib/tuist/docs_sidebar.ex` — new "Runners" group registering both pages.

### Why

Runners is a shipped, invite-only feature with no user-facing documentation yet. This gives customers a page to onboard from and, importantly, articulates the competitive positioning against providers like Blacksmith and Namespace.

The overview page:
- Leads with an **invite-only** availability note (the feature is gated behind the `:runners` toggle in production).
- Frames the **core differentiator**: other runner providers offer a colocated cache, but it is a CI-only island that only ever warms from other CI runs. Tuist Runners read and write the *same* module and Xcode cache used by developer machines and every other environment, and that cache is colocated on the runner's private network — colocated jobs receive an in-cluster `TUIST_CACHE_ENDPOINT` (see `runners_controller.ex` `dispatch_response/5`), so cache traffic stays internal instead of crossing the public internet.
- Documents getting started (`runs-on: tuist-<profile>`), the platform/shape catalog, macOS Xcode selection, and observability.

The profiles page documents profiles as account-scoped named machine shapes referenced from `runs-on`: default `linux`/`macos` profiles, fields, naming rules (immutability, reserved names), creation flow, examples, and the 10-per-account limit.

Content was verified against the implementation (`Tuist.Runners.Profile`, `Tuist.Runners.Catalog`, `server/config/config.exs`, `runners_controller.ex`). Shapes/Xcode versions are documented from the `config/config.exs` defaults with a note that the dashboard's New Profile form is authoritative (managed deploys override the catalog via Helm). Pre-GA interactive VNC/shell debugging sessions were intentionally left out.

### How to test locally

- Run the docs sidebar test, which builds and renders every registered slug (including the two new pages):

  ```bash
  cd server && mix test test/tuist/docs_sidebar_test.exs
  ```

  All 4 tests pass, including "all sidebar slugs correspond to actual docs pages".

- Optionally, start the server (`mise run dev` in `server/`) and visit the **Guides → Runners** section to review the rendered pages.


server/priv/docs/en/guides/features/runners.md
```diff
@@ -0,0 +1,57 @@
...
+        details="Choose a platform, size, and Xcode version with named machine profiles you reference from runs-on."
+        linkText="Manage profiles"
+        link="/guides/features/runners/profiles"/>
+</HomeCards>
+
+## Why Tuist Runners {#why-tuist-runners}
+
+Most CI runner providers make jobs faster by giving you beefier machines and a persistent cache, typically a shared volume mounted back into your CI runs. That helps, but a volume is just a disk: it carries whatever previous CI runs happened to leave on it. It isn't replicated to keep just the freshest artifacts close to where builds run, and it never reaches your developers' machines.
+
+Tuist Runners are different because the cache is the same cache your developers and every other environment already read from and write to. The <.localized_link href="/guides/features/cache/module-cache">module cache</.localized_link> and <.localized_link href="/guides/features/cache/xcode-cache">Xcode cache</.localized_link> that a teammate warms on their laptop, or that an earlier build produced, are a hit on the runner, and vice versa. There's no separate CI cache to warm up.
```
pepicrft: Wondering if we should make these parts of the docs less coupled to Xcode's cache and module cache, given that we support other build systems. 
Also, we mention that we make it fast in dev environments, but there are also other environments, like agents, so we can maybe talk about how there is more diversity of environments today, and it's crucial that the improvements are aimed at more than just the CI environments.

pepicrft (approved):

server/priv/docs/en/guides/features/runners.md
fortmarek: Decoupled from the module/Xcode caches (now links to the general cache overview and talks about "the cache"/"artifacts" regardless of build system), and reframed the section around the diversity of environments that build today — developer laptops, CI, and coding agents — making explicit the speedup isn't limited to CI.

PR #11752, opened 2026-07-10
feat(server): add configurable artifact retention

pepicrft:
Self-hosted Tuist operators can now opt into automatic artifact cleanup and choose a separate retention window for each supported artifact family.

## What changed

The server now reads six independently optional retention variables for cache artifacts, app previews, build archives, run artifacts, test attachments, and shard bundles. A grouped `server.artifactRetentionDays` Helm value renders those variables individually and leaves them absent when a family is not configured.

The existing retention workers now schedule only the configured artifact families and remove files from object storage without deleting the associated PostgreSQL or ClickHouse metadata. The public data-retention guide, self-hosting guide, data-export reference, deployment example, Helm documentation, and product changelog describe the new behavior and its exact scope.

## Why

Self-hosted instances could accumulate artifacts indefinitely even though the public retention guide described a maximum 30-day hosted window. Operators had no built-in way to apply their own storage policy and had to manage object cleanup outside Tuist.

## Root cause

Artifact-retention schedules were part of the hosted-only Oban configuration. Their windows were derived from hosted account plans and capped at 30 days, so self-hosted web processes never scheduled the cleanup workers.

## Approach

The implementation reuses the existing cleanup paths instead of introducing a second deletion system. Self-hosted schedules are opt-in by artifact family, while the hosted Tuist server continues using its plan-based policy unchanged.

The environment decides *whether* a retention cron is installed at boot. It does not travel in the job args: every self-hosted batch re-reads the current variables before deleting files, so removing a variable or widening a window protects files from already-queued work once the server restarts. An earlier revision of this branch also serialized the window into the cron args, which was dead payload that shadowed the real source of truth, so those args now carry only the `self_hosted` flag.

Paginated scans update and reschedule the same unique Oban job, which prevents multiple server replicas from branching the scan, and each page receives a fresh retry budget.

Cache cleanup covers instance-managed cache buckets, including matching objects whose account prefix no longer resolves to a current account. Accounts using account-specific custom cache storage remain excluded.

### Deletion-job uniqueness

The per-account deletion workers declare `unique: [keys: [:account_id], period: :infinity, states: [...]]` so that an account with a large expired backlog gets exactly one deletion chain walking it, however many nightly runs elapse while that chain is in flight.

While building the self-hosted path it became clear that `Oban.insert_all/1` does not honor that declaration. Oban's basic engine (the open-source one we run; bulk unique inserts are an Oban Pro feature) merges `on_conflict: :nothing` into a plain `Ecto.Repo.insert_all/3` and never takes the advisory lock that `Oban.insert/1` uses to run the uniqueness query. So the declared uniqueness silently did nothing on the bulk path.

The self-hosted branch was written to avoid this with per-job `Oban.insert/1`, but the hosted branch kept the pre-existing `Oban.insert_all/1`. That left the guarantee holding on one path and not the other. Both now insert one job at a time.

The alternative considered was keeping the bulk insert and pre-filtering each page against the in-flight jobs with an extra query. It would preserve one round trip per page instead of one per job, but it reimplements Oban's uniqueness check in application code and is only race-free because the scheduler happens to be a singleton today, which is an invariant nothing enforces. Inserting one at a time is what Oban's own documentation prescribes for the basic engine, and the cost is bounded: a page is at most `page_size` accounts times five workers, and the resulting jobs are each a database round trip anyway when they execute.

## Impact

The change is inert by default. When all retention variables are absent or blank, self-hosted deployments schedule no artifact-retention jobs and Helm emits no retention variables.

Once configured, artifact files older than their family window are deleted. Dashboard and analytics metadata remains available because PostgreSQL and ClickHouse rows are preserved. Package registry mirror objects, runner log archives, and account-specific custom cache buckets remain outside this policy.

On the hosted Tuist server, backlogged accounts no longer accumulate duplicate deletion chains issuing redundant object-storage deletes and database scans against the same rows.

## Validation

- Focused server, worker, documentation, and changelog tests: 134 passed.
- `MIX_ENV=test mix compile --warnings-as-errors`: passed.
- `mix format --check-formatted` for all changed Elixir files: passed.
- `mix credo`: no issues across 1,519 source files.
- `helm lint infra/helm/tuist`: passed.
- Helm rendering verified the default configuration emits zero retention variables, a partial configuration emits two, and a complete configuration emits all six.
- The uniqueness regression is pinned by a test on each path (`Tuist-hosted scheduling enforces deletion job uniqueness` and its self-hosted twin). Both were confirmed to fail when the insert is reverted to `Oban.insert_all/1`, which is what let us verify the bug was real rather than theoretical.

### How to test locally

1. Set one retention variable, such as `TUIST_CACHE_ARTIFACT_RETENTION_DAYS=30`, or configure the matching `server.artifactRetentionDays.cacheArtifacts` Helm value.
2. Restart a self-hosted web server in a production-like environment.
3. Confirm the cache-retention workers are present in the Oban schedule and workers for omitted families are absent.
4. Run the configured worker against an instance-managed bucket containing an object older than the window and one newer object.
5. Confirm only the expired object is removed and its associated database metadata remains available.

### DIFF
--- a/server/priv/docs/en/guides/server/data-retention.md
+++ b/server/priv/docs/en/guides/server/data-retention.md
@@ -28,35 +28,7 @@ The active account plan determines the Tuist Cloud retention window. If an accou
 
 ### Self-hosted instances {#self-hosted-artifact-files}
 
-Self-hosted artifact cleanup is opt-in and configured separately for each artifact type. Set an artifact type's environment variable to a positive integer number of days to enable its cleanup. Leaving a variable unset or blank disables cleanup only for that artifact type. Self-hosted windows are not capped at 30 days.
-
-The following example enables every supported artifact type with an independent retention window. Omit a variable or leave it blank to keep cleanup disabled for that artifact type.
-
-```bash
-TUIST_CACHE_ARTIFACT_RETENTION_DAYS=30
-TUIST_APP_PREVIEW_RETENTION_DAYS=30
-TUIST_BUILD_ARCHIVE_RETENTION_DAYS=60
-TUIST_RUN_ARTIFACT_RETENTION_DAYS=30
-TUIST_TEST_ATTACHMENT_RETENTION_DAYS=30
-TUIST_SHARD_BUNDLE_RETENTION_DAYS=14
-```
-
-The supported variables and their object-storage scope are:
-
-| Environment variable | Artifact files |
-| --- | --- |
-| `TUIST_CACHE_ARTIFACT_RETENTION_DAYS` | Xcode cache, legacy content-addressable storage cache, module cache, and Gradle cache files |
-| `TUIST_APP_PREVIEW_RETENTION_DAYS` | App preview builds and icons |
-| `TUIST_BUILD_ARCHIVE_RETENTION_DAYS` | Current and legacy build archives |
-| `TUIST_RUN_ARTIFACT_RETENTION_DAYS` | All objects stored under an expired run's artifact prefix, including result bundles, invocation records, and session archives |
-| `TUIST_TEST_ATTACHMENT_RETENTION_DAYS` | Test run attachments |
-| `TUIST_SHARD_BUNDLE_RETENTION_DAYS` | Test shard bundles |
-
-The cleanup jobs remove artifact blobs from object storage only. Associated PostgreSQL and ClickHouse metadata remains available for analytics and dashboards. The setting does not change database retention rules.
-
-Cache artifact cleanup scans the instance-managed cache buckets and skips accounts configured with account-specific custom cache storage. Matching cache objects whose prefix no longer resolves to a current account are cleaned with the configured window. Database-backed cleanup for app previews, current build archives, run artifacts, test attachments, and shard bundles uses the account's current storage configuration. Legacy build archive cleanup scans the instance-managed artifact bucket. Package registry mirror objects and runner log archives are outside this configurable policy. Runner log archives keep their separate 90-day retention window.
-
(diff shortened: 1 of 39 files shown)


server/priv/docs/en/guides/server/data-retention.md
```diff
@@ -7,11 +7,13 @@
 ---
 # Data retention {#data-retention}
 
-Tuist keeps downloadable artifacts for the windows below. Dashboard history and analytics can remain visible after artifact downloads expire. No artifact file is kept for more than 30 days.
+Tuist Cloud and self-hosted Tuist instances apply different artifact-retention policies. Tuist Cloud uses plan-based windows capped at 30 days. Artifact cleanup is disabled by default on self-hosted instances, where operators can opt in and configure a separate window for each supported artifact type.
```
fortmarek: Tuist Cloud 😓

fortmarek (approved):
Looks good apart from the "Tuist Cloud" mentions 😅

server/priv/docs/en/guides/server/data-retention.md
pepicrft: Good catch, fixed in f00d739.

You are right that it should not be there. I checked and `Tuist Cloud` does not appear anywhere else under `server/priv/docs`, and main's version of this guide just said `Tuist`. The only surviving uses in the repo are dated blog posts, one of which literally says `the server (formerly Tuist Cloud)`. So I reintroduced a retired brand name purely because I needed something to contrast with self-hosted.

Switched both this guide and `server/data-export.md` to `the hosted Tuist server`, which matches the wording already used in the cache and server self-hosting guides. The `#tuist-cloud-artifact-files` anchor became `#hosted-artifact-files`; it was new in this branch so there are no inbound links to break. Docs tests still pass.

PR #11753, opened 2026-07-10
fix(server): enforce dashboard authorization

pepicrft:
## What changed

- Scope bundle lookups to the selected project.
- Require project administrator authorization for bundle deletion.
- Require organization read authorization before loading the members page.
- Enforce invitation and member-role authorization in dashboard event handlers.
- Enforce test update authorization for test-case mutations.
- Enforce create, update, and delete authorization for automation mutations.
- Add a regression test for cross-organization members-page access.

## Why

The dashboard did not consistently apply the same project and organization authorization rules as the public interface. This allowed confirmed cross-project bundle reads and cross-organization member disclosure through normal dashboard routes.

## Root cause

Several LiveViews trusted route parameters or selected account state without checking the corresponding authorization policy. Some mutation handlers relied on the visibility of dashboard controls instead of enforcing authorization when events were received.

## Impact

The same fixes protect bundle data, organization member data, test metadata, and automation configuration. No destructive action was executed during validation.

## Approach

The existing authorization policy is reused at the dashboard boundaries. Bundle reads are additionally constrained by project identifier so a valid bundle identifier from another project cannot satisfy the lookup.

## Validation

- Headless Chrome reproduced the unpatched behavior:
  - A private bundle from `tuist/tuist` loaded at `tuist/public` and exposed the Delete bundle control.
  - An unrelated user opened `tuist/members` and saw member emails, roles, and management controls.
- With the fixes restored, the same browser sessions received Bundle not found and You are not authorized to perform this action.
- `mix test test/tuist_web/live/bundle_live_test.exs test/tuist_web/live/members_live_test.exs test/tuist_web/live/test_case_live_test.exs test/tuist_web/live/project_automations_live_test.exs`
- Result: 50 tests passed.
- `git diff --check`

## Advisory

[GHSA-557v-9rcm-gmh6](https://github.com/tuist/tuist/security/advisories/GHSA-557v-9rcm-gmh6)

### DIFF
--- a/server/lib/tuist/authorization.ex
+++ b/server/lib/tuist/authorization.ex
@@ -100,6 +100,11 @@ defmodule Tuist.Authorization do
       allow([:authenticated_as_account, scopes_permit: "project:bundles:read"])
       allow([:authenticated_as_account, scopes_permit: "project:bundles:write"])
     end
+
+    action :delete do
+      desc("Allows the admin of a project to delete a bundle.")
+      allow([:authenticated_as_user, user_role: :admin])
+    end
   end
 
   object :account do
--- a/server/lib/tuist_web/live/bundle_live.ex
+++ b/server/lib/tuist_web/live/bundle_live.ex
@@ -11,6 +11,7 @@ defmodule TuistWeb.BundleLive do
 
   alias Noora.Filter
   alias Tuist.Bundles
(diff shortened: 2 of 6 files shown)


pepicrft:
Closing this public pull request. The security fix will be coordinated through the private GitHub security advisory workflow.

PR #11760, opened 2026-07-10
feat(cli): cache the mapped Swift Package Manager dependency graph across generations

fortmarek:
## What changed

`tuist generate` re-ran the SwiftPM package-graph mapping on every invocation, even when nothing changed. The `Package.swift` manifest JSONs are already served from the manifest cache, but the output of `PackageInfoMapper` (the external dependency graph: targets, settings, module maps) was rebuilt from scratch each time. On the tuist repository's own graph (~105 packages) that is roughly 2-4s per generate on a prod-shape VM, visible in session logs as ~242 `[TuistLoader]` mapping lines with no subprocess activity.

This PR adds a `SwiftPackageManagerGraphCache` in `TuistLoader` that serializes the mapped `DependenciesGraph` (already `Codable`) into a new `SwiftPackageManagerGraphs` cache category, following the same pattern as `CachedManifestLoader`. When the cache key matches, `SwiftPackageManagerGraphLoader.load` returns the cached graph and skips package-info loading and mapping entirely.

## Cache key design

The mapping is deterministic in the following inputs, all of which are part of the key:

- `workspace-state.json` raw bytes. This pins the resolved dependency set: revisions of source-control checkouts, versions of registry downloads, artifacts, and prebuilts. Checkout/download contents are immutable for a given revision, so pinning is sufficient for remote and registry packages.
- The root package manifest content hash.
- A content fingerprint of every local (path-based) package. Local packages need more than their manifest because mapping inspects their disk layout: public headers scanning, custom module map detection, and resource globbing all happen at mapping time. Content hashes (not mtimes) keep the fingerprint stable across fresh checkouts, which matters for one-shot runner VMs restoring a per-account cache volume.
- Swift version, tuist version, manifest-loading environment variables, the sandbox flag, and the package/scratch directory paths (absolute paths are embedded in the mapped graph).

Package settings are compared with `Equatable` from the cached entry rather than hashed into the key: their JSON encoding is not deterministic across processes because they contain sets (`Destinations`) and non-string-keyed dictionaries (`[BuildConfiguration: ...]`), which encode in hash order. This was caught during validation when byte-identical entries still missed run-to-run.

## Side-effect safety

Mapping is not a pure function: it writes generated module maps (`<scratch>/tuist-derived/ModuleMaps`, `<package>/Derived`) and synthesizes XCFrameworks for static-library artifact bundles (`<scratch>/tuist-derived/XCFrameworks`). A cached graph referencing wiped derived files would produce broken projects, so the entry records the derived file listing at store time and a hit is only served while all recorded files exist. Wiping `.build` or the derived directories falls back to re-mapping, which recreates them.

Two churn sources are deliberately excluded from the local-package fingerprint because generation itself writes them into the package folder, and including them would invalidate the entry that stored them: the `Derived` directory and generated `*.xcodeproj`/`*.xcworkspace` contents. The latter was caught during validation: generating with a different focused-target set rewrites the generated projects inside local packages, which previously invalidated the cache even though the SPM graph is independent of the generated workspace shape.

Cache misses log their reason at debug level (incompatible entry, inputs changed, settings changed, missing derived files), so future regressions are diagnosable from session logs. Every failure in the cache (unhashable inputs, unavailable cache directory, undecodable entry) degrades to re-mapping; it never fails the load.

## Impact

- Warm `tuist generate` skips the mapping phase entirely (~2s warm / ~4s cold-boot on the tuist fixture per the original measurement; larger on bigger SPM graphs).
- The serialized graph (~900KB for the tuist repository) can ride the per-account cache volume so one-shot runner VMs skip the mapping too.
- `tuist clean` picks up the new `swiftPackageManagerGraphs` category automatically via `CaseIterable`.

## Validation

- New tests in `SwiftPackageManagerGraphLoaderTests`: a warm load reuses the cached graph without invoking the mapper and returns an equal graph; changing `workspace-state.json`, package settings, or a local package's file content re-maps; removing a recorded derived file re-maps. All 16 tests in the suite pass (11 pre-existing tests run with the cache neutralized via an unavailable cache directory, preserving their previous behavior).
- End-to-end with a debug build on the tuist repository: cold generate stores the entry and logs 242 mapping lines; subsequent generates log `Reusing the cached Swift Package Manager dependency graph` with zero mapping lines, including across focused-generation shape changes (`tuist generate tuist` vs `tuist generate tuist TuistLoaderTests`) and repeated runs.
- `mise run lint` passes.

### DIFF
--- a/cli/Sources/TuistCore/Cache/CacheCategory.swift
+++ b/cli/Sources/TuistCore/Cache/CacheCategory.swift
@@ -28,6 +28,9 @@ public enum CacheCategory: String, CaseIterable, RawRepresentable {
     /// uploaded by the last `tuist generate`.
     case generationMetadata
 
+    /// The mapped Swift Package Manager dependency graphs cache
+    case swiftPackageManagerGraphs
+
     public var directoryName: String {
         switch self {
         case .plugins:
@@ -48,6 +51,8 @@ public enum CacheCategory: String, CaseIterable, RawRepresentable {
             return "SelectiveTests"
         case .generationMetadata:
             return "GenerationMetadata"
+        case .swiftPackageManagerGraphs:
+            return "SwiftPackageManagerGraphs"
         }
     }
 }
--- a/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphCache.swift
+++ b/cli/Sources/TuistLoader/Loaders/SwiftPackageManagerGraphCache.swift
@@ -1,0 +1,264 @@
+import FileSystem
+import Foundation
+import Path
+import ProjectDescription
+import TuistConstants
+import TuistCore
+import TuistEnvironment
+import TuistLogging
+import TuistSupport
+
+/// The set of hashes that determine the mapped Swift Package Manager dependency graph.
+///
+/// Mapping is deterministic in these inputs:
+/// - `workspaceStateHash` pins the resolved dependency set, including revisions of source-control
+///   checkouts, versions of registry downloads, artifacts, and prebuilts. Checkout and download
+///   contents are immutable for a given revision/version, so hashing their pins is sufficient.
+/// - `rootManifestHash` and `localPackagesHash` cover the packages whose content can change
+///   without the resolution changing. Local packages are fingerprinted by file content (not
+///   only their manifest) because mapping inspects their disk layout: public headers, custom
+///   module maps, and resources.
+/// - `swiftVersion` and `environmentHash` cover the remaining mapper inputs; `packagePath` and
+///   `scratchDirectoryPath` are embedded in the mapped graph's absolute paths.
+///
+/// The package settings are part of the cached entry rather than this key: their JSON encoding is
+/// not deterministic across processes (they contain sets and non-string-keyed dictionaries), so
+/// they are compared with `Equatable` on load instead of by hash.
+struct SwiftPackageManagerGraphCacheKey: Equatable, Codable {
+    var tuistVersion: String
+    var swiftVersion: String
+    var workspaceStateHash: String
+    var rootManifestHash: String
+    var localPackagesHash: String
+    var environmentHash: String
+    var disableSandbox: Bool
+    var packagePath: String
+    var scratchDirectoryPath: String
+}
+
+private struct CachedSwiftPackageManagerGraph: Codable {
+    /// Note: please bump the version in case the cache structure is modified.
+    /// This ensures older cache versions are not loaded using this structure.
+    static let currentCacheVersion = 1
+
+    var cacheVersion: Int = currentCacheVersion
+    var key: SwiftPackageManagerGraphCacheKey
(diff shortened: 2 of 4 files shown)


fortmarek:
Addressed the cache-invalidation gap around pruned package roots in 0050af9196.

The cache key pinned remote/registry packages only through the `workspace-state.json` hash, and the recorded side-effect files only covered the derived directories. So if a checkout, registry download, or local package folder was pruned while `workspace-state.json` survived (a partial `.build` restore, or restoring the graph cache from an account volume without the checkouts), a hit would return a graph referencing missing source/binary roots. Before this cache existed that same state failed while loading the pruned package's manifest ("run tuist install"), so serving the cached graph silently regressed it to a broken project.

Fix: on a hit, verify every external project folder (the graph's `externalProjects` keys: checkouts, registry downloads, and local package paths) still exists, alongside the existing derived-file check. A missing root is treated as a stale entry and falls back to re-mapping, restoring the pre-cache behavior. The check reuses the already-decoded graph, so it only adds existence stats. New test `load_mapsAgain_whenAnExternalPackageFolderIsRemoved` covers it; the suite is at 17/17.

fortmarek:
Added a mapping-speed commit (5e6489eab5) after profiling the mapper itself, since caching only helps warm/repeat runs and the first (cold) generate still pays the full mapping.

Profiled on the tuist fixture (141 packages, 830 targets). The dominant cost turned out to be **contention, not filesystem work**. The mapping fans out a task per package, each fans out a task per target, and each target recursively globs its source tree for resources/headers/module maps (the resource glob itself spawns a task per subdirectory) — all unbounded. That nests into thousands of simultaneous walks. A single default-resource glob is ~3.4ms run serially but ~100-380ms under the concurrent thrash (~97% contention). On a fast multi-core machine the mapping wall is already ~0.5s, but the unbounded fan-out has a heavy tail that becomes the 2-4s seen on constrained CI.

Fix: cap the package-level `concurrentMap` at the active processor count. The bounded variant preserves input order, so the produced graph is byte-identical (17/17 loader tests still pass, cache still hits). Two pure static trait helpers moved to file scope to stay under the type-body lint limit.

Things I ruled out while profiling, to save the next person the detour: `SwiftVersionProvider.swiftVersion()` is **not** a subprocess herd (one coalesced `swift --version`, confirmed by counting spawns — an earlier reading that blamed it was wall-clock-around-await suspension inflation, not real cost), and base-path resolution is cheap. The real work is the resource globbing, which is inherent to translating SPM's convention-based layout into explicit targets and returns zero resources for every external target here but can't be skipped without changing semantics.

Caveat: I couldn't reproduce the 2-4s locally (this box is ~0.5s), so the fix is validated by the serial-vs-concurrent per-op delta and tail-latency elimination in a concurrency sweep, not an absolute wall-time drop on a prod-shape VM. Happy to split this into its own PR if you'd prefer to keep the cache change isolated.

PR #11769, opened 2026-07-11
fix(cli): generate scheme code coverage references for local Swift package targets

dogo:
Follow-up to #11399.

## Summary

#11399 allowed `codeCoverageTargets` in project schemes to reference targets from other projects, so `tuist generate` no longer fails linting for schemes that gather coverage for local Swift package targets (native SPM integration via `packages: [.local(path:)]`). However, two gaps remained for that scenario:

1. **The reference never reached the generated scheme.** `SchemeDescriptorsGenerator` resolves coverage targets through the graph, and local package targets are not part of it, so the reference was silently dropped. The generated scheme ended up with `codeCoverageEnabled = "YES"` but no `<CodeCoverageTargets>` list, meaning coverage was gathered for *all* targets instead of only the specified ones.
2. **`GraphLinter` still warned.** `lintSchemesUnknownTargets` uses `Scheme.targetDependencies()`, which includes `codeCoverageTargets`, producing a `Cannot find targets X (..) defined in Scheme` warning for perfectly valid references. This is the same root cause #11399 fixed in `SchemeLinter` via `nonCodeCoverageTargetDependencies`.

## Changes

- `SchemeDescriptorsGenerator`: when a coverage target is not found in the graph but its `projectPath` matches a local Swift package declared by a graph project, emit a package-style `XCScheme.BuildableReference`, with `ReferencedContainer` pointing at the package directory (relative to the workspace) and `BlueprintIdentifier`/`BuildableName`/`BlueprintName` set to the reference name. This matches what Xcode writes when the package product is selected manually in the scheme editor.
- `LocalPackageCoverageTargetsValidator` (new, part of `ManifestGraphLoader`): coverage references pointing at a declared local package path are validated against the actual products of that specific package, obtained by dumping its manifest (memoized per package, and only when such references exist, so there is no cost for projects that don't use this). Xcode exposes package coverage buildables by product name, so a name that is not a product of the referenced package, such as the backing target's name, a typo, a product of another package referenced through the wrong path, or a remote product under a local package path, is removed from the graph with a warning, and coverage falls back to all targets instead of producing a silently empty report.
- `GraphLinter.lintSchemesUnknownTargets`: skip references used *exclusively* as code coverage targets whose name matches a consumed product of a local package. A local package reference that also appears in another scheme action (build, run, etc.) keeps the warning, since the generator cannot resolve it there. This check, together with the equivalent one in the generator, acts as a backstop for graphs that don't go through manifest loading.
- `Scheme+Core` (TuistCore): `nonCodeCoverageTargetDependencies()` moved here from `SchemeLinter` (where #11399 introduced it as a private helper), so both `SchemeLinter` and `GraphLinter` share the same implementation, and `targetDependencies()` is now built on top of it.

## Testing

- `LocalPackageCoverageTargetsValidatorTests`: graph target references are kept without dumping packages, product references are kept, non-product names are dropped with a warning, a product of another package referenced through the wrong path is dropped with a warning, references outside local packages are left for the linter, and workspace schemes are validated too.
- `SchemeDescriptorsGeneratorCodeCoverageTests` (Swift Testing): the buildable reference emitted for a consumed product (`container:../Package`, blueprint identifier/name equal to the product name, `primary` buildable identifier); non-product names and references matching neither the graph nor a package are dropped.
- `GraphLinterTests`: no issues for a coverage reference naming a consumed product of a local package; the warning is preserved for non-product names, for local package references also used in other scheme actions, and for unknown targets.
- End-to-end: the `generated_ios_app_with_local_swift_package` fixture now has a product whose name differs from its backing target (`LibraryC` -> `LibraryCCore`) and a second package (`PackageB` with product `OtherLibrary`). The scheme references the product, the backing target name, and the other package's product through the wrong path. The `ios_app_with_local_swift_package_product_code_coverage` acceptance test asserts the generated `.xcscheme` keeps only the `LibraryC` product reference (`container:Packages/PackageA`).
- Validated against a real project using native SPM integration (`Project(packages: [.local(path: "../")])` with `codeCoverageTargets: [.project(path: "..", target: "MyPackageProduct")]`): `tuist generate` emits no warning and the generated `.xcscheme` contains the `<CodeCoverageTargets>` entry with `ReferencedContainer = "container:.."`.

### DIFF
--- a/cli/Sources/TuistGenerator/Generator/SchemeDescriptorsGenerator.swift
+++ b/cli/Sources/TuistGenerator/Generator/SchemeDescriptorsGenerator.swift
@@ -420,7 +420,7 @@ struct SchemeDescriptorsGenerator: SchemeDescriptorsGenerating {
             environments = environmentVariables(arguments.environmentVariables)
         }
 
-        let localPackagePaths = localPackagePaths(graphTraverser: graphTraverser)
+        let localPackageProducts = graphTraverser.consumedLocalPackageProducts()
         let codeCoverageTargets = try testAction.codeCoverageTargets
             .compactMap { (target: TargetReference) -> XCScheme.BuildableReference? in
                 if let graphTarget = graphTraverser.target(
@@ -433,10 +433,11 @@ struct SchemeDescriptorsGenerator: SchemeDescriptorsGenerating {
                         rootPath: rootPath
                     )
                 }
-                // Coverage targets can reference targets of local Swift packages, which are
+                // Coverage targets can reference products of local Swift packages, which are
                 // not part of the graph. Xcode resolves them through the package directory
-                // used as the referenced container.
-                if localPackagePaths.contains(target.projectPath) {
+                // used as the referenced container. Only names matching a consumed package
+                // product are emitted; other names would produce an empty coverage report.
+                if localPackageProducts[target.projectPath]?.contains(target.name) == true {
                     return packageTargetBuildableReference(target: target, rootPath: rootPath)
                 }
                 return nil
@@ -1103,27 +1104,13 @@ struct SchemeDescriptorsGenerator: SchemeDescriptorsGenerating {
         )
     }
 
-    /// Returns the paths of the local Swift packages declared by the graph projects.
-    ///
-    /// - Parameter graphTraverser: Tuist graph traverser.
-    /// - Returns: Set of local package directories.
-    private func localPackagePaths(graphTraverser: GraphTraversing) -> Set<AbsolutePath> {
-        Set(
-            graphTraverser.projects.values
-                .flatMap(\.packages)
-                .compactMap { package -> AbsolutePath? in
-                    guard case let .local(path) = package else { return nil }
-                    return path
-                }
-        )
-    }
-
-    /// Creates a buildable reference for a target that belongs to a local Swift package.
-    /// Package targets are not part of the generated projects, so the reference points to
-    /// the package directory the same way Xcode does when the target is selected manually.
+    /// Creates a buildable reference for a product of a local Swift package.
+    /// Package products are not part of the generated projects, so the reference points to
+    /// the package directory the same way Xcode does when the product is selected manually.
     ///
     /// - Parameters:
-    ///     - target: The target reference, with the package directory as its project path.
+    ///     - target: The target reference, with the package directory as its project path
+    ///       and a package product as its name.
     ///     - rootPath: Path to the project or workspace.
     private func packageTargetBuildableReference(
         target: TargetReference,
--- a/cli/Sources/TuistGenerator/Generator/SchemeDescriptorsGenerator.swift
+++ b/cli/Sources/TuistGenerator/Generator/SchemeDescriptorsGenerator.swift
@@ -420,18 +420,26 @@ struct SchemeDescriptorsGenerator: SchemeDescriptorsGenerating {
             environments = environmentVariables(arguments.environmentVariables)
         }
 
+        let localPackagePaths = localPackagePaths(graphTraverser: graphTraverser)
         let codeCoverageTargets = try testAction.codeCoverageTargets
             .compactMap { (target: TargetReference) -> XCScheme.BuildableReference? in
-                guard let graphTarget = graphTraverser.target(
+                if let graphTarget = graphTraverser.target(
                     path: target.projectPath, name: target.name
-                )
-                else { return nil }
-                return try testCoverageTargetReferences(
-                    graphTarget: graphTarget,
-                    graphTraverser: graphTraverser,
-                    generatedProjects: generatedProjects,
-                    rootPath: rootPath
-                )
+                ) {
+                    return try testCoverageTargetReferences(
+                        graphTarget: graphTarget,
+                        graphTraverser: graphTraverser,
+                        generatedProjects: generatedProjects,
+                        rootPath: rootPath
+                    )
+                }
+                // Coverage targets can reference targets of local Swift packages, which are
+                // not part of the graph. Xcode resolves them through the package directory
+                // used as the referenced container.
+                if localPackagePaths.contains(target.projectPath) {
+                    return packageTargetBuildableReference(target: target, rootPath: rootPath)
+                }
+                return nil
             }
 
         var macroExpansion: XCScheme.BuildableReference?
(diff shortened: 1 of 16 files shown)


cli/Sources/TuistGenerator/Linter/GraphLinter.swift
```diff
@@ -78,6 +88,13 @@ public struct GraphLinter: GraphLinting {
                     {
                         return false
                     }
+                    // Code coverage targets can reference targets of local Swift packages,
+                    // which are not part of the graph but are resolvable by Xcode.
+                    if codeCoverageTargets.contains(targetReference),
+                       localPackagePaths.contains(targetReference.projectPath)
```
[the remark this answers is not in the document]
dogo: Good catch. Addressed in 26d59ad99: the warning is now only suppressed for references used exclusively as code coverage targets, so a local package reference that also appears in another scheme action keeps the warning. I moved `nonCodeCoverageTargetDependencies` from `SchemeLinter` to the `Scheme+Core` extension in TuistCore so both linters share the same implementation, and added `lint_when_localPackage_reference_is_not_exclusive_to_codeCoverage` to cover this case.

cli/Tests/TuistGeneratorTests/Generator/SchemeDescriptorsGeneratorTests.swift
```diff
@@ -878,6 +878,95 @@ final class SchemeDescriptorsGeneratorTests: XCTestCase {
         XCTAssertEqual(codeCoverageTargetsBuildableReference.first?.buildableName, "App.app")
     }
 
+    func test_schemeTestAction_with_codeCoverageTargets_fromLocalPackage() throws {
```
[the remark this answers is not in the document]
dogo: Done in 26d59ad99: the new tests were rewritten with Swift Testing (`@Test`, `#expect`, `#require`) in a `SchemeDescriptorsGeneratorCodeCoverageTests` suite. The pre-existing XCTest cases in the file were left untouched.

cli/Sources/TuistGenerator/Generator/SchemeDescriptorsGenerator.swift
```diff
@@ -420,18 +420,26 @@ struct SchemeDescriptorsGenerator: SchemeDescriptorsGenerating {
...
+                        graphTarget: graphTarget,
+                        graphTraverser: graphTraverser,
+                        generatedProjects: generatedProjects,
+                        rootPath: rootPath
+                    )
+                }
+                // Coverage targets can reference targets of local Swift packages, which are
+                // not part of the graph. Xcode resolves them through the package directory
+                // used as the referenced container.
+                if localPackagePaths.contains(target.projectPath) {
```
pepicrft: **Resolve local package targets to real coverage buildables**

This accepts any name under a declared local package path and writes that name into the generated buildable reference. I reproduced the case where the package product and target names differ:

```swift
products: [.library(name: "LocalMathProduct", targets: ["LocalMathCore"])]
targets: [.target(name: "LocalMathCore")]
```

With `.project(path: "LocalPackage", target: "LocalMathCore")`, generation succeeds without a warning and tests pass, but the resulting coverage report has zero targets and zero executable lines. Xcode exposes the coverage buildable as `LocalMathProduct`; using that product name produces 88.89% coverage. Before this change, the target-name configuration was rejected with a warning and coverage fell back to all targets, so this turns a non-empty report into a silently empty report. A typo under a valid local package path fails the same way because the linter also validates only the path.

Could we resolve the package target to its product, or validate that the supplied name is an actual package coverage buildable, before suppressing the warning and setting `onlyGenerateCoverageForSpecifiedTargets`? Please add an end-to-end test where product and target names differ.

pepicrft (changes requested):
I reproduced the intended same-name scenario successfully, but the product-name mismatch below is a regression that can silently erase the entire coverage report. Requesting changes until package buildable resolution or validation covers that case.

cli/Sources/TuistGenerator/Generator/SchemeDescriptorsGenerator.swift
dogo: Thanks for catching this, you are right that writing an unvalidated name could silently empty the report. Addressed in 86d0c89d0: coverage references pointing at a local package are now validated against the package products consumed by the declaring project's targets (`.package(product:)` dependencies), since Xcode exposes package coverage buildables by product name. A name that is not a consumed product, like the backing target's name or a typo, is dropped and coverage falls back to all targets as before, and the GraphLinter warning is kept for it. I could not resolve target names to products because with native package integration the graph does not carry the package manifest, so validation against consumed products was the closest reliable source. Also extended the generated_ios_app_with_local_swift_package fixture with a product whose name differs from its backing target (LibraryC -> LibraryCCore) and added the ios_app_with_local_swift_package_product_code_coverage acceptance test asserting the product reference is kept and the target name is dropped.

pepicrft: Thanks, I retested `86d0c89d0`. The single-package product and target mismatch is fixed: an invalid backing-target name now produces a warning and falls back to all-target coverage, while the correct product name produces selective 88.89% coverage.

There is still a cross-package variant. `consumedLocalPackageProducts()` collects every `.package(product:)` dependency in a project and unions that entire set into every local package path:

```swift
for path in localPackagePaths {
    result[path, default: []].formUnion(products)
}
```

I reproduced this with two local packages. The project consumes `LocalMathProduct` from package A and `OtherMathProduct` from package B, while the coverage reference is `.project(path: "LocalPackageA", target: "OtherMathProduct")`. Generation succeeds without a warning, the scheme writes `OtherMathProduct` with `ReferencedContainer = "container:LocalPackageA"`, and the tests pass, but the coverage report contains zero targets and zero executable lines. The same implementation can also accept a remote package product under a local package path when a project declares both package types.

Could we preserve the exact product-to-package association? If the graph cannot prove that association, the safe behavior is to drop the reference, keep the warning, and preserve all-target fallback coverage. Please add a two-package acceptance case covering a product referenced through the wrong local package path.

dogo: You are right, the union heuristic could not prove the association. Addressed in 0017e1d06 by validating against the exact package: a new `LocalPackageCoverageTargetsValidator` runs as part of `ManifestGraphLoader` (after graph mappers, so it covers all generate/build/test flows) and, for each coverage reference pointing at a declared local package path, dumps that specific package's manifest and checks the name against its actual products. The dump is memoized per package and only happens when such references exist, so there is no cost for projects that don't use this. A name that is not a product of the referenced package, including a product of another package through the wrong path or a remote product under a local path, is removed from the graph with a warning and coverage falls back to all targets. The consumed-products check in the generator and linter stays as a backstop for graphs that don't go through manifest loading. Also added PackageB to the fixture with a coverage reference to its product through PackageA's path; the acceptance test asserts it is dropped and only the valid product reference reaches the xcscheme.

pepicrft (approved):

PR #11777, opened 2026-07-12
fix(server): harden security boundaries across server surfaces

pepicrft:
## What changed

This change closes resource-scoping gaps across the server surfaces reviewed during the security audit.

- Scope build, run, test, attachment, crash-report, bundle, and cache-endpoint lookups to the authenticated project or account.
- Reauthorize dashboard mutation events for organization members, bundles, and project automations, while hiding controls that the current user cannot invoke.
- Restrict account cache discovery to matching project and account subjects, including delegated user grants.
- Decode persistent cache entries with the safe term decoder and treat incompatible entries as cache misses.
- Upgrade dependencies with published security fixes and preserve the behavior affected by those upgrades, including Markdown rendering, content recompilation, large-number formatting, and timezone-data downloads.

## Why

Route-level authorization established that a subject could use a broad feature, but several handlers subsequently accepted globally unique identifiers without proving that the referenced record belonged to the selected project or account. Dashboard controls also relied on visibility in places where a forged event could still reach the mutation handler.

The dependency advisory scan additionally identified packages that needed newer releases. Those upgrades crossed several compatibility boundaries, so the migration includes explicit behavior-preserving adapters and regression coverage instead of only changing the lockfile.

## Root cause

Authorization was not consistently carried from the selected route resource into nested record lookups and event handlers. Globally unique identifiers made normal requests work while allowing a caller who knew another identifier to cross the intended tenant boundary.

Persistent cache values were decoded without the safe option. The first safe-decoding change then exposed a compatibility edge: an old value containing an unavailable atom could raise during a deployment instead of behaving like an expired entry.

The dependency upgrade also changed transport response shapes, Markdown output, compile-time content tracking, and number formatting in ways that compilation and the existing suite did not detect.

## Approach

The fix uses the existing central authorization policy and adds resource ownership constraints at the data-access boundary. Mutating dashboard events reauthorize immediately before the write, so stale sockets and forged events follow the same policy as the rendered interface.

Compatibility work was validated against the previous implementations. Invalid persistent values now rebuild under the existing cache lock, newsletter output is normalized to the previous document structure, production content files remain compiler dependencies, and timezone downloads use a compatible request adapter.

## Impact

Legitimate project members, organization administrators, project tokens, account tokens, and delegated user grants retain their existing access. Cross-project and cross-account identifiers now return a forbidden or not-found response instead of exposing or mutating another tenant's data.

Persistent cache corruption or deployment-era incompatibility becomes a cache miss rather than an application error. Dependency advisories are cleared without changing the rendered newsletter corpus or the daily report's number formatting.

## Validation

- `TUIST_DEV_ALL_LOCALES=1 mix test`: 5,560 tests passed.
- `MIX_ENV=prod mix compile --warnings-as-errors`: passed.
- `mix format --check-formatted`: passed.
- `mix credo`: passed with no issues.
- `mix hex.audit`: no retired packages or security advisories.
- `mix deps.unlock --check-unused`: passed.
- `mix deps.get --check-locked`: passed.
- `mise run security`: completed successfully; dependency and secret scans reported no findings.
- Local outbound transport smoke: passed for every upgraded request client and the timezone adapter.
- Newsletter corpus parity: 194 rendered fields checked with zero semantic mismatches.
- Large-number parity: 17 boundary cases checked with zero mismatches.
- Headless Chrome smoke runs covered bundle deletion, member management, and project automation permissions for administrator and regular-member sessions.


server/lib/tuist/authorization/checks.ex
```diff
@@ -81,6 +81,10 @@ defmodule Tuist.Authorization.Checks do
     false
   end
 
+  def accounts_match(%AuthenticatedAccount{issued_by: %User{} = user}, %Account{} = account) do
+    Accounts.owns_account_or_belongs_to_account_organization?(user, %{id: account.id})
```
fortmarek, line 85: **[P1] Avoid per-account membership queries**

This predicate performs a database-backed ownership/membership lookup. `Cache.account_cache_handles/2` invokes authorization inside an `Enum.filter`, and OAuth introspection computes the subject's grants before narrowing them to the requested account, so a user-issued token adds queries for every accessible account. Please derive or batch the authorized account IDs once, or scope grant generation to the introspected account before running this check.

server/lib/tuist_web/controllers/api/test_case_run_attachments_controller.ex
```diff
@@ -236,4 +252,15 @@ defmodule TuistWeb.API.TestCaseRunAttachmentsController do
       expires_at: System.system_time(:second) + expires_in
     })
   end
+
+  defp valid_test_run_id?(body_params, test_case_run) do
+    is_nil(body_params[:test_run_id]) or body_params.test_run_id == test_case_run.test_run_id
```
fortmarek: **[P3] Remove the redundant `test_run_id` validation**

The request schema accepts `test_case_run_id`, not `test_run_id`, and `create_attachment/4` already derives the stored and storage-key `test_run_id` from the scoped `test_case_run`. A supplied extra value cannot influence the write, so this check is dead defensive code.

server/lib/tuist_web/controllers/api/test_case_run_attachments_controller.ex
```diff
@@ -184,9 +184,25 @@ defmodule TuistWeb.API.TestCaseRunAttachmentsController do
   )
 
   def create(%{assigns: %{selected_project: project}, body_params: body_params} = conn, _params) do
-    attachment_id = UUIDv7.generate()
+    with {:ok, test_case_run} <-
+           Tests.get_test_case_run_by_id(body_params.test_case_run_id,
+             project_id: project.id,
+             preload: [:arguments]
```
fortmarek, line 190: **[P3] Use a targeted argument-membership lookup**

The relationship validation is necessary, but preloading every argument just to call `Enum.any?` makes attachment creation scale with the number of arguments in the run. A scoped existence lookup using the argument ID and test-case-run ID would preserve the boundary without loading the full collection.

server/lib/tuist_web/controllers/api/crash_reports_controller.ex
```diff
@@ -76,22 +76,34 @@ defmodule TuistWeb.API.CrashReportsController do
...
-      triggered_thread_frames: Map.get(body_params, :triggered_thread_frames, ""),
-      test_case_run_id: body_params.test_case_run_id,
-      test_case_run_attachment_id: body_params.test_case_run_attachment_id,
-      inserted_at: NaiveDateTime.utc_now()
-    }
+  def create(%{assigns: %{selected_project: project}, body_params: body_params} = conn, _params) do
+    with {:ok, test_case_run} <-
+           Tests.get_test_case_run_by_id(body_params.test_case_run_id,
+             project_id: project.id,
+             preload: [:attachments]
```
fortmarek, line 83: **[P3] Use a targeted attachment-membership lookup**

The attachment/run relationship must be verified, but preloading all attachments and scanning them makes each crash-report upload scale with the run's attachment count. A scoped existence lookup by attachment ID and test-case-run ID would enforce the same invariant more directly.

server/mix.exs
```diff
@@ -68,23 +68,24 @@ defmodule Tuist.MixProject do
       # (ArgumentError ":upgrade not a binary"). Without this pin, plug is an
       # unconstrained transitive dep and every `mix deps.get` re-resolves it to
       # the latest, churning the lock and forcing full recompiles.
-      {:plug, "1.19.2", override: true},
+      {:plug, "1.19.5", override: true},
```
fortmarek, line 74: **[P3] Update the Plug pin rationale**

The adjacent comment says Plug versions starting at 1.19.3 crash WebSocket upgrades with this Bandit version, while this line now pins 1.19.5. If that issue is fixed in 1.19.5, please update or remove the stale rationale; otherwise the new pin contradicts the documented constraint.

fortmarek (commented):
I found one query-scaling issue and a few places where the hardening can be simplified without weakening the security boundary.

fortmarek (approved):

server/lib/tuist_web/controllers/api/test_case_run_attachments_controller.ex
```diff
@@ -184,9 +184,25 @@ defmodule TuistWeb.API.TestCaseRunAttachmentsController do
   )
 
   def create(%{assigns: %{selected_project: project}, body_params: body_params} = conn, _params) do
-    attachment_id = UUIDv7.generate()
+    with {:ok, test_case_run} <-
+           Tests.get_test_case_run_by_id(body_params.test_case_run_id,
```
fortmarek, line 188: **[P1] Avoid a ClickHouse read-after-write race for new test runs**

Test-case runs and their arguments are inserted from `Tuist.Tasks.run_async/1` through independent ClickHouse buffers, which flush every five seconds in production. The create-test response returns the in-memory IDs immediately, and the CLI then starts attachment uploads concurrently. This new lookup can therefore return 404 for a valid run or argument that has not become visible yet; the CLI only logs the failure and does not retry, so the attachment and any corresponding crash report are lost. Tests mask the race because tasks are synchronous and buffer writes are write-through in the test environment. Please avoid depending on immediate ClickHouse visibility here—ensure the rows are visible before returning the test response, validate against another authoritative source/capability, or retry across the buffer interval. A more targeted ClickHouse query alone would not fix the race.

server/lib/tuist_web/controllers/api/cache_controller.ex
```diff
@@ -71,12 +72,32 @@ defmodule TuistWeb.API.CacheController do
...
+        if is_nil(account) or Authorization.authorize(:account_cache_read, subject, account) == :ok do
+          endpoints =
+            account_handle
+            |> Accounts.get_cache_endpoints_for_handle(technology(conn))
+            |> Enum.reject(&is_nil/1)
+
+          json(conn, %{endpoints: endpoints})
+        else
+          conn
+          |> put_status(:forbidden)
```
fortmarek, line 98: **[P2] Add the new forbidden response to the OpenAPI contract**

This operation now returns 403, but its `responses` map still declares only the 200 response. The generated Swift client consequently handles this as `.undocumented` and `GetCacheEndpointsService` surfaces a generic unknown-response error instead of an authorization error. Please declare the forbidden response and regenerate the server specification and Swift client.

fortmarek (commented):
Two additional backward-compatibility risks surfaced when tracing the new checks through the production write and generated-client paths.

server/lib/tuist_web/controllers/api/test_case_run_attachments_controller.ex
pepicrft: Good catch, this was going to lose every attachment in production.

Confirmed the whole chain: `create_test_case_runs` handed the runs and arguments to `TestCaseRun.Buffer` / `TestCaseRunArgument.Buffer` from inside `Tuist.Tasks.run_async`, the buffer only flushes on its interval (`flush_interval_ms`, 5s by default), and `InspectResultBundleService.uploadTestSummary` fans out to `uploadAttachments` immediately after `createTest` returns. The upload then hits this lookup, gets a 404, and the CLI only logs a warning, so the attachment and its crash report are gone. Tests hid it because `write_through_repo: true` and `Tuist.Tasks` is synchronous in the test env, exactly as you said.

Fixed in abf9a3ed26 by taking your first option: the two tables the upload endpoints read are now inserted and flushed before the response, and failures, repetitions and embedded attachments stay on the async path. The test run row itself is already written synchronously here (`IngestRepo.insert` in `create_test`), so this keeps the write pattern on this path consistent rather than introducing a new one.

Added a regression test in `tests_test.exs` that stubs `Tuist.Tasks.run_async` to a no-op, which stands in for an unflushed buffer, and asserts the run and its arguments are still readable. It fails on the previous commit and passes now.

server/lib/tuist_web/controllers/api/test_case_run_attachments_controller.ex
pepicrft: Right, it is dead. Removed in 24b327357f.

Worth recording why, because it is slightly worse than "the value is stripped": the schema has no `test_run_id` property, and with `additionalProperties` unset OpenApiSpex keeps unrecognised properties rather than dropping them, but under their original string key (`Cast.Object.cast_additional_property/2`). Only declared properties get atomised. So `body_params[:test_run_id]` reads the atom key and is always nil, whichever way the client sends it. Combined with `create_attachment/4` deriving both the stored and the storage-key `test_run_id` from the scoped run, there was no path for the supplied value to reach the write.

server/mix.exs
pepicrft: Stale rationale, you are right. Fixed in f25f091649.

The crash the comment describes came from the header validation plug 1.19.3 added to `Plug.Conn.inform/3`, which ran `:binary.match/2` straight on the header key and blew up on atom keys. 1.19.4 fixed it ("Allow atom header keys on `inform`"), and `Plug.Conn` now normalises with `to_string(key)` before validating, so 1.19.5 does not have the bug.

Rather than take the changelog's word for it I upgraded a real WebSocket connection against bandit 1.11.1 on plug 1.19.5 with Mint.WebSocket, and it completes with a 101.

The pin itself stays, since plug is otherwise an unconstrained transitive dep and every `mix deps.get` re-resolves it to the latest and churns the lock. The comment now says that instead of the obsolete crash rationale.

server/lib/tuist_web/controllers/api/cache_controller.ex
pepicrft: Agreed, fixed in 48f3820109. Declared the forbidden response on the operation, regenerated the spec and the client with `mise run generate-api-cli-code`, and `GetCacheEndpointsService` now handles `.forbidden` and throws with the server's message, following the same shape as `GetGenerationService`. Verified `TuistServer` still builds.

One thing to flag: the generator also wanted to pull in the cursor pagination that landed in 9a33f7f49d (#11634) without a matching spec regeneration. I kept that out of this branch, because it flips `total_count` to optional on `PaginationMetadata` and would change the Generations client types, which has nothing to do with this PR. So `server.yml` is still stale with respect to #11634 and someone should regenerate it separately.

server/lib/tuist/authorization/checks.ex
pepicrft: I dug into this one and I would rather leave it as is. Three things bound it more tightly than it looks:

The loop is not over "every accessible account" in an open-ended sense. `account_cache_handles/2` filters `accessible_accounts/1`, which for a user-issued token is the user's own account plus the accounts of the organizations they belong to (`cache.ex:138-150`). That is the user's membership set, typically a handful.

It runs once, not twice. Only `:cache_read` carries `:accounts_match`; `:cache_create` does not (`authorization.ex:119-137`), so the write bucket adds no lookups.

And it is not on a per-request path. `cache_grants/1` has exactly two callers, both in `Introspection` (`introspection.ex:14` and `:32`), and Kura caches the introspection result per token in its `principal_cache` with a 600s allow TTL by default (`kura/src/extension/mod.rs:1012`). So this is a few indexed queries per token per ten minutes, on a function that already authorizes once per accessible project (`project_cache_handles/3`), a strictly larger loop that predates this PR.

The batching version I tried is what put me off. To collapse it you have to decide the membership question in `Cache` and hand the answer to the filter, which pulls the decision back out of the policy and into the caller. That is the exact coupling this PR is removing, and getting it subtly wrong in an authz path is how the original hole appeared. I would rather pay the queries.

Happy to revisit if introspection ever shows up in traces, but the fix then is caching the grants, not weakening the check.

server/lib/tuist_web/controllers/api/test_case_run_attachments_controller.ex
pepicrft: I looked at this and I do not think a scoped existence query buys anything here.

The preload is not over the test run's arguments, it is over the arguments of the single `test_case_run` we already resolved: `ClickHouseRepo.preload(run, [:arguments])` follows the `has_many` on `test_case_run_id`, so the collection is the arguments of one test case. For a parameterized test that is single digits, and for the common case it is empty.

Both shapes also cost the same round trip. The preload is one extra ClickHouse query and an EXISTS on `(argument_id, test_case_run_id)` would be one extra ClickHouse query. The only difference is a handful of rows on the wire, and in exchange we would add a query function to the `Tests` context whose only caller is this check. Leaving it.

server/lib/tuist_web/controllers/api/crash_reports_controller.ex
pepicrft: Same reasoning as the argument-membership thread, so I left this one too.

`preload: [:attachments]` is scoped to the single `test_case_run` we just resolved, not to the test run, so it loads the attachments of one test case rather than the whole run's. And a scoped EXISTS lookup would be one extra ClickHouse query, which is exactly what the preload already costs. The gain would be a few rows on the wire in exchange for another context function that only this check calls.

The relationship check itself stays, and it now also benefits from the write-through fix in abf9a3ed26, since the run it validates against is guaranteed to be visible by the time the crash report arrives.

PR #11784, opened 2026-07-13
fix(cli): emit explicit cross-package target dependency edges in generated workspaces

### LINKED ISSUE #11770: Generated workspaces are missing cross-package target dependency edges
### What we observed

Tuist generates each SPM package in a workspace as its own Xcode project, and target dependencies inside a project become real dependency edges in the build description. But dependencies that cross package-project boundaries do not: dumping the llbuild manifest (the build description swift-build derives from the generated workspace) for the Tuist CLI workspace itself showed **zero target-dependency edges between package projects**. When a target in package B depends on a product of package A, the scheduler has no edge telling it to order A before B — ordering rests entirely on implicit mechanisms (product-name resolution, search paths, and scheduling timing).

This surfaced while evaluating swift-build's remote-cache machinery during the compilation-cache work on https://github.com/tuist/tuist/pull/11768, but the issue is a generator correctness gap independent of caching.

### Why it matters

On a cold build the gap is mostly masked: compilation is slow and wide, so upstream products tend to exist on disk by the time downstream planning looks for them. On a **warm cached build** the timing collapses — replayed tasks finish near-instantly — and the missing edges surface as races: a downstream target's dependency scan or driver planning can run before the upstream package's .swiftmodule exists on disk.

We reproduced the failure mode directly: with compilation caching configured so that dependency-scan tasks had to rely on on-disk products rather than cache read-through, builds of the generated workspace reproducibly lost targets to:

```
unable to resolve module dependency
```

With the CAS plugin's read-through enabled, scan tasks get their answers from the cache and do not need the sibling project's on-disk products, which masks the fragility — but the schedule is only correct by accident. Anything that shifts timing (different cache states, future Xcode/swift-build scheduling changes, remote-cache machinery modes) can expose it again.

### Expected behavior

Generation should emit explicit target-dependency edges across package projects — a target that links a product of another generated package project should carry a real dependency on that product's target, so the build description gives llbuild an actual ordering constraint instead of relying on implicit resolution.

### Reproduction / verification

1. Generate a workspace with cross-package product dependencies (the Tuist CLI workspace itself, `tuist generate`, is a ready-made case: its package projects depend on each other's products).
2. Dump the build description / llbuild manifest for a workspace build and inspect dependency edges between targets of different package projects — today there are none.
3. A fix is verifiable the same way: the edges should appear, and a warm compilation-cache build with scan read-through disabled should no longer hit `unable to resolve module dependency`.

fortmarek:
Resolves https://github.com/tuist/tuist/issues/11770

### What changed

Generation now emits explicit cross-project `PBXTargetDependency` edges when a target links a product of another generated SPM **package (external)** project. The existing `ForeignBuildCrossProjectDependencyGenerator` — the only code that ever wired cross-project container proxies — is generalized into `CrossProjectDependencyGenerator`; `foreignBuild()` aggregate handling is preserved as a subset.

### Why it changed / root cause

Tuist generates each SPM package as its own `.xcodeproj`. `TargetGenerator.generateTargetDependencies` only wires `PBXTargetDependency` for **same-project** dependencies (`GraphTraverser.directLocalTargetDependencies`, filtered by `$0.path == path`). A target in package B that depends on a product of package A was left entirely to Xcode's implicit dependency discovery (product-name matching across the shared build products dir), so the build description swift-build/llbuild derives had **zero** ordering edges between package projects.

That ordering was only correct by accident. On a cold build compilation is slow and wide, so upstream products tend to exist by the time downstream planning looks for them. On a **warm compilation-cache build** the timing collapses: replayed tasks finish near-instantly and a downstream target's dependency scan can run before the upstream package's `.swiftmodule` exists on disk, surfacing as `unable to resolve module dependency`. This was hit directly while evaluating swift-build's remote-cache machinery in #11768. The CAS plugin's read-through masks it (scan tasks get answers from the cache), but any timing shift can expose it again.

### Why this solution

The cross-project `PBXContainerItemProxy` + `PBXTargetDependency` machinery already existed and was correct; it was just gated to `foreignBuild()` aggregates. Generalizing it (rather than adding a parallel path) reuses the UUID-stabilization logic that eagerly encodes each dependency-bearing pbxproj so the consumer's proxy resolves to a stable remote UUID instead of a `TEMP_` placeholder.

Scope is deliberately limited to dependencies on **external** (SPM package) projects plus `foreignBuild()` aggregates. Cross-project edges between hand-written **local** projects keep relying on Xcode's implicit resolution, preserving existing behavior and avoiding churn to unrelated generated project structure.

### User/developer impact

Warm cached builds of workspaces with cross-package product dependencies no longer race on missing `.swiftmodule`s; the build description carries real ordering constraints instead of relying on implicit product-name resolution.

### How to test locally

Unit tests:

```
tuist generate TuistGenerator TuistGeneratorTests --no-open
  -only-testing "TuistGeneratorTests/CrossProjectDependencyGeneratorTests" \
```

New coverage: `generate_wiresCrossPackageProductDependency` (external package B → product of package A wires an explicit edge) and `generate_skipsLocalCrossProjectDependency` (local↔local stays implicit). All existing foreignBuild cases still pass.

End-to-end: build the `tuist` scheme, regenerate the CLI workspace, and inspect a package project:

```
grep -c "isa = PBXTargetDependency" \
  .build/registry/downloads/apple/swift-nio-ssl/2.37.0/swift-nio-ssl.xcodeproj/project.pbxproj
```

Before this change: `swift-nio-ssl.xcodeproj` had no cross-project edges. After: 8 explicit `PBXTargetDependency` edges (proxyType nativeTarget) to `swift-nio.xcodeproj` targets (NIO, NIOCore, CNIOBoringSSL, …) with stabilized remote UUIDs.

### Validation run

- All 335 `TuistGeneratorTests` pass (incl. the new external-package case and the local-stays-implicit guard).
- End-to-end regeneration of the Tuist CLI workspace confirmed the 8 new cross-project edges above.

### DIFF
--- a/cli/Sources/TuistGenerator/Generator/CrossProjectDependencyGenerator.swift
+++ b/cli/Sources/TuistGenerator/Generator/CrossProjectDependencyGenerator.swift
@@ -4,39 +4,53 @@ import TuistCore
 import XcodeGraph
 import XcodeProj
 
-/// Wires cross-project `PBXTargetDependency` edges for consumers that depend on a
-/// `foreignBuild()` aggregate located in another project.
+/// Wires explicit cross-project `PBXTargetDependency` edges for consumers that depend on a target
+/// located in another generated project.
 ///
-/// `foreignBuild()` aggregates produce no product under `BUILT_PRODUCTS_DIR`, so Xcode's implicit
-/// dependency discovery (which walks framework references and product paths) does not pick them
-/// up across project boundaries. Without an explicit cross-project `PBXTargetDependency`, the
-/// aggregate's build script is never invoked when the consumer is built, and the consumer keeps
-/// linking the stale artifact even after the foreign sources change.
+/// `TargetGenerator` only emits `PBXTargetDependency` for dependencies within the same project
+/// (`directLocalTargetDependencies`). Cross-project dependencies are otherwise left to Xcode's
+/// implicit dependency discovery (product-name matching across the shared build products dir),
+/// which gives `swift-build` / `llbuild` no explicit ordering constraint. That ordering is then
+/// only correct by accident: on a warm compilation-cache build replayed tasks finish near-instantly
+/// and a downstream target's dependency scan can run before the upstream `.swiftmodule` exists,
+/// surfacing as `unable to resolve module dependency`.
 ///
-/// For each cross-project edge `(consumer -> foreign-build aggregate)` this generator adds the
-/// following to the consumer's pbxproj:
+/// This generator closes that gap for two cases:
+///
+/// - Each generated SPM package becomes its own external project, so a target that links a product
+///   of another package project needs an explicit edge across the package-project boundary.
+/// - `foreignBuild()` aggregates produce no product under `BUILT_PRODUCTS_DIR`, so Xcode's implicit
+///   discovery (which walks framework references and product paths) does not pick them up across
+///   project boundaries at all. Without the explicit edge the aggregate's build script is never
+///   invoked when the consumer builds, and the consumer keeps linking the stale artifact.
+///
+/// Cross-project edges between hand-written local projects are intentionally left to implicit
+/// resolution to preserve existing behavior.
+///
+/// For each cross-project edge `(consumer -> dependency)` this generator adds the following to the
+/// consumer's pbxproj:
 ///
 /// - A `PBXFileReference` to the remote `.xcodeproj` (deduplicated per remote project).
 /// - A `PBXProject.projects` entry pairing that reference with an (empty) Products `PBXGroup`.
 /// - A `PBXContainerItemProxy` with `proxyType = .nativeTarget`,
 ///   `containerPortal = .fileReference(<remote .xcodeproj ref>)`, and
-///   `remoteGlobalID = .object(<remote aggregate>)`.
+///   `remoteGlobalID = .object(<remote target>)`.
 /// - A `PBXTargetDependency` wrapping the proxy, appended to the consumer target's
 ///   `dependencies`.
 ///
 /// XcodeProj's `ReferenceGenerator` assigns deterministic UUIDs lazily at encode time. The
 /// consumer pbxproj serializes the proxy's `remoteGlobalIDString` from the remote target's
-/// reference value, so we eagerly encode each aggregate-bearing pbxproj once up front to fix
+/// reference value, so we eagerly encode each dependency-bearing pbxproj once up front to fix
 /// its UUIDs. When the writer later writes the consumer pbxproj, the proxy resolves to the
 /// stable remote UUID instead of the temporary placeholder.
-protocol ForeignBuildCrossProjectDependencyGenerating {
+protocol CrossProjectDependencyGenerating {
     func generate(
         graphTraverser: GraphTraversing,
         projectDescriptors: [ProjectDescriptor]
     ) throws
 }
 
-struct ForeignBuildCrossProjectDependencyGenerator: ForeignBuildCrossProjectDependencyGenerating {
+struct CrossProjectDependencyGenerator: CrossProjectDependencyGenerating {
     func generate(
         graphTraverser: GraphTraversing,
         projectDescriptors: [ProjectDescriptor]
@@ -45,8 +59,8 @@ struct ForeignBuildCrossProjectDependencyGenerator: ForeignBuildCrossProjectDepe
         let edges = collectEdges(graphTraverser: graphTraverser)
         guard !edges.isEmpty else { return }
 
-        let aggregateProjectPaths = Set(edges.map(\.aggregateProjectPath))
-        for path in aggregateProjectPaths.sorted() {
+        let dependencyProjectPaths = Set(edges.map(\.dependencyProjectPath))
+        for path in dependencyProjectPaths.sorted() {
             guard let descriptor = descriptorsByPath[path] else { continue }
             _ = try descriptor.xcodeProj.pbxproj.dataRepresentation(outputSettings: PBXOutputSettings())
         }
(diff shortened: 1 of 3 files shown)


o-nnerb:
Not sure why this was closed without merging (no comment on the close), but wanted to flag: the minimal repro in #12589 (`swift-nio`'s `NIOCore` failing to resolve `DequeModule`/`Atomics`) is a clean instance of exactly the case this PR's description says it fixes — "each generated SPM package becomes its own external project, so a target that links a product of another package project needs an explicit edge." All three targets involved (`NIOCore`, `DequeModule`, `Atomics`) are real `PBXNativeTarget`s in their own generated projects; there's no synthesized-target complication (the class 2 case from the follow-up comments on #11770, e.g. `NIOHPACK` → `NIOCore` where `NIOCore` isn't a real target at all).

Given it's been ~6 weeks since this was closed with no further activity on #11770, and this PR's approach (generalizing `ForeignBuildCrossProjectDependencyGenerator` into `CrossProjectDependencyGenerator`) already covers the class-1 case that's hitting real users in practice (confirmed via #12589), would it make sense to reopen and land this on its own — even if the class-2 synthesized-target case still needs separate follow-up work? Happy to test this branch against the `tuist-nio-dequemodule-repro` repro if that'd help move it forward.

o-nnerb:
Tested this branch (`worktree-bridge-cse_01VqnRhGPYT8sYGJHYRfUb36`, commit `c0d430d7`) against the `tuist-nio-dequemodule-repro` minimal repro from #12589. Quick summary of the results in case it saves someone else the round trip:

**The fix itself works as described.** After `tuist generate --no-open` with this branch's binary, the generated `swift-nio.xcodeproj`'s `.pbxproj` does contain the new explicit edges — `PBXTargetDependency`/`PBXContainerItemProxy` from `NIOCore` to `Atomics` (in `swift-atomics.xcodeproj`) and to `DequeModule` (in `swift-collections.xcodeproj`), with `remoteGlobalIDString`s correctly matching the real native targets on the other side. So the generator change is doing exactly what the PR description says.

**But it doesn't fix the failure mode reported in #12589.** Running the exact repro command from that issue —

```
xcodebuild build -project SampleApp.xcodeproj -scheme "SampleApp" \
  -destination "generic/platform=iOS Simulator" \
  CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO
```

— still fails identically, on two independent clean-DerivedData runs:

```
.../swift-nio/Sources/NIOCore/AsyncChannel/AsyncChannelHandler.swift:15:8: error: unable to resolve module dependency: 'DequeModule'
.../swift-nio/Sources/NIOCore/AsyncSequences/NIOAsyncWriter.swift:15:8: error: unable to resolve module dependency: 'Atomics'
```

Interestingly, building the same generated output via `-workspace SampleApp.xcworkspace` instead of `-project SampleApp.xcodeproj` *does* succeed — but it also succeeds with stock, unpatched Tuist 4.205.0, so that success is unrelated to this PR; workspace-level builds of this repro were never broken.

So it looks like the correct pbxproj edges aren't sufficient when `xcodebuild` is invoked directly on the standalone `.xcodeproj` rather than through the workspace — presumably because the container-proxy edges only get resolved/scheduled in a workspace-level build description, not when a single generated project is built in isolation. That would mean this PR fixes the class of problem #11770 originally described (missing edges between package projects), but the specific failure mode in #12589 needs a separate fix on top, related to how standalone-project builds resolve cross-project references.

Happy to share the full build logs/pbxproj diffs if useful for debugging further.

o-nnerb:
**Edit:** dug further into this after the above. The generator change here is correct as far as it goes, but standalone `-project` builds need one more piece — opened a follow-up: **#12591**.

Short version: `xcodebuild -project SampleApp.xcodeproj` never even *attempts* to build `Atomics`/`DequeModule` — they don't appear in the build log at all, success or failure. `SampleApp.xcodeproj`'s own `PBXProject.projectReferences` only lists `swift-nio.xcodeproj` (its direct dependency); `swift-atomics.xcodeproj`/`swift-collections.xcodeproj` are only referenced from *inside* `swift-nio.xcodeproj`'s own `projectReferences` (added correctly by this PR). `xcodebuild -project` resolves one hop of nested subproject references from the primary project but doesn't recurse into a nested subproject's *own* `projectReferences` to discover further-nested ones — which is exactly why building via the generated `.xcworkspace` works (it lists all four projects as flat top-level members, sidestepping the recursion limit) while `-project` doesn't, and why that workspace success isn't actually this PR's doing (it also happens with unpatched stock Tuist).

#12591 expands every generated project's `PBXProject.projectReferences` to the full transitive closure of package projects reachable from it, not just direct one-hop dependencies — no new `PBXTargetDependency`, just the file-reference visibility `xcodebuild -project` needs to discover the project at all. With both this PR and #12591 applied, `xcodebuild -project SampleApp.xcodeproj -scheme "SampleApp" ...` succeeds on two independent clean-DerivedData runs against the `tuist-nio-dequemodule-repro` repro from #12589.

o-nnerb:
**Update:** #12591 no longer depends on this PR — turns out `main` had already grown its own `CrossProjectTargetDependencyGenerator` (covering `foreignBuild()` aggregates, macros, and resource bundles) independently of this PR, but never picked up this PR's core case: a plain external-package-to-external-package dependency (e.g. `NIOCore` → `Atomics`) still gets no edge at all on current `main`. #12591 now adds that missing edge directly against current `main`, plus the transitive-`projectReferences` fix from my comment above, as a single self-contained commit. Verified against the #12589 repro on top of current `main` (two clean-DerivedData runs, both green). Given that, this PR is probably safe to leave closed — its useful part is folded into #12591.

PR #11785, opened 2026-07-13
fix(infra): tart-kubelet GC must not reap a running VM (golden or orphan)

fortmarek:
## What changed

The tart-kubelet disk garbage collector now consults VM liveness before deleting. A `live()` guard sits in front of **both** `Tart.Delete` sites in `runOnce` — the golden-base branch and the orphan-clone branch — and skips (with a log line) any VM that currently has a `tart run` process.

- A probe error is treated as **live** (fail-safe): deleting a maybe-running VM is the harmful outcome, so an unreadable liveness signal skips the reap and lets the next 5-min pass retry. This mirrors the reconciler, which already treats an `IsRunning` error as "assume Running" to avoid flapping.
- Uses `IsRunning` (pgrep against `tart run <name>`), **not** the `VM.State` field — the codebase already documents `tart list`'s `State` as unreliable for backgrounded VMs (the on-disk state file isn't updated on exit).
- `IsRunning` shells out to `pgrep` rather than the fakeable Tart binary, so a small `IsRunning func` override was added to `Collector` (defaulting to `Tart.IsRunning`), mirroring the existing `Now func` test override. Production wiring in `cmd/tart-kubelet/main.go` leaves it nil and gets the real probe unchanged.

## Why

The GC decided reaps purely from Pod-reference + age. A golden base with no Pod referencing its digest and past `GoldenRetention` (24h) was deleted; a non-golden local VM with no backing Pod was deleted immediately. Neither branch checked whether a VM was actually running.

### Root cause

On the staging host-Kura mini (`m1@62.210.195.110`), `/var/log/tart-kubelet.log` showed the GC delete `tuist-golden-6fdcd0674dda678d` while it was in active use — booted directly by launchd and serving benchmarks — killing it and removing its directory mid-session. That golden had no k8s Pod on the node referencing its digest, so it was absent from the expected set, its in-memory retention clock had elapsed, and `keepGolden` returned "reap". The reconciler's design comment "the golden is never run; it's a stable copy-on-write base" is the assumption that made deletion safe in production. The benchmark harness violates it, and the GC had no defense — process state was simply never an input to the reap decision.

The fix is scoped to point (a) of the review: **never delete a running VM regardless of name**. It also hardens the orphan-clone branch, where a still-running orphan means the reconciler crashed mid-Stop and a hard delete is equally wrong.

## Not in scope (follow-ups from the same review)

- **(b)** Pinning a golden referenced by a pool/warm-pool spec that has no currently-scheduled Pod (the GC reads Pods + Store, not pool specs; today such a golden is protected only by the 24h retention).
- **(c)** A wall-clock min-age floor. Note this would not have prevented the incident — a running golden older than 24h still needs the liveness guard.
- `isNotFound` (in `internal/tart`) matches `"not found"` but not the `"does not exist"` string seen in the staging logs, which is why an absent golden surfaced as a "warm-path probe failed" rather than a clean cold-path miss. Worth reconciling against the deployed tart version separately.

## Validation

- New `TestRunOnceSparesRunningVMs` drives a real `runOnce` against a fake `tart` binary, with the golden seeded past its retention window so it *would* reap absent the guard:
  - running golden → spared
  - stopped golden → reaped
  - probe error → spared (fail-safe)
- `go build ./...`, `go vet ./internal/podagent/`, and `gofmt` clean.
- Full `internal/podagent` suite passes, including the untouched `TestKeepGolden` / `TestExpectedSet` regressions.

### DIFF
--- a/infra/tart-kubelet/internal/podagent/garbage.go
+++ b/infra/tart-kubelet/internal/podagent/garbage.go
@@ -60,6 +60,12 @@ type Collector struct {
 	// Now is overridable in tests; defaults to time.Now.
 	Now func() time.Time
 
+	// IsRunning probes whether a VM has a live `tart run` process, gating
+	// every reap so an in-use VM is never deleted mid-session. Overridable
+	// in tests (pgrep isn't stubbable via the Tart binary); defaults to
+	// Tart.IsRunning.
+	IsRunning func(ctx context.Context, name string) (bool, error)
+
 	mu sync.Mutex
 
 	// goldenSeen tracks the last time each golden base VM was seen
@@ -140,6 +146,10 @@ func (c *Collector) runOnce(ctx context.Context, aggressive bool) {
 				if c.keepGolden(vm.Name, expected, now, retention, aggressive) {
 					continue
 				}
+				if c.live(ctx, vm.Name) {
+					logger.Info("spared golden base: a tart run process is live", "name", vm.Name)
+					continue
+				}
 				if err := c.Tart.Delete(ctx, vm.Name); err != nil {
 					logger.Error(err, "delete stale golden base", "name", vm.Name)
 					continue
@@ -149,6 +159,10 @@ func (c *Collector) runOnce(ctx context.Context, aggressive bool) {
 				continue
 			}
 			if _, want := expected.vms[vm.Name]; want {
+				continue
+			}
+			if c.live(ctx, vm.Name) {
+				logger.Info("spared orphan VM: a tart run process is live", "name", vm.Name)
 				continue
 			}
 			if err := c.Tart.Delete(ctx, vm.Name); err != nil {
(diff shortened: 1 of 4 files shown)


infra/tart-kubelet/internal/podagent/garbage.go
```diff
@@ -151,6 +161,10 @@ func (c *Collector) runOnce(ctx context.Context, aggressive bool) {
 			if _, want := expected.vms[vm.Name]; want {
 				continue
 			}
+			if c.live(ctx, vm.Name) {
```
pepicrft, line 164: The liveness check used here ultimately runs `pgrep -f "tart run <name>"` without an argument boundary. That means a stopped orphan named `team-job` is considered live while `team-job-2` runs, so both the periodic collector and the aggressive disk reclamation path keep skipping the orphan. Both names are valid outputs of `VMNameForPod`. I reproduced this directly with `pgrep`: a process named `tart run tuist-golden-deadbeef2` matched a query for `tuist-golden-deadbeef`. Can we require an exact virtual machine name boundary and add a non-golden orphan regression test?

pepicrft (approved):
The incident fix is sound. I left one inline follow-up about exact process-name matching in the orphan cleanup path.

infra/tart-kubelet/internal/podagent/garbage.go
fortmarek: Good catch — fixed at the root in `IsRunning` (39e16b0). pgrep's `-f` match was unanchored, so I anchored the name to an argument boundary: `tart run <name>($| )`, with `regexp.QuoteMeta` keeping any metacharacter literal. This fixes the false positive for the periodic collector, the aggressive reclaim path, and the reconciler's two liveness reads. Added `TestIsRunningRequiresExactNameBoundary` (real `tart run <name>-2` decoy) and the non-golden-orphan `TestRunOnceReapsOrphanDespiteSimilarlyNamedRunningVM` (real `Tart.IsRunning`, asserts the stopped orphan is still reaped) — both fail against the unanchored code and pass now.

Issue #76, opened 2018-07-24
Website link to docs is broken


zdnk:
Link on website to docs does not work. Same for github etc.

pepibumur:
I double checked on the [website](https://tuist.io) and links seem to be working. Could you be more specific about the links that are not working for you?

pepibumur:
I'll close this use since we haven't got any input. If the problem is still there, feel free to open it.

PR #11788, opened 2026-07-13
feat(cli): account cache volumes for the macOS runner fleet (spec #76)

fortmarek:
Implements [Hive spec #76](https://hive.tuist.dev/specs/76), "Account cache volumes for the macOS runner fleet".

## What this does

macOS runner VMs are ephemeral: every job boots a fresh VM, so the Tuist cache starts cold and `tuist generate` re-pulls the account's binaries from the remote cache on every run. This adds a **per-account cache volume** on each Mac host so a job starts warm — without ever exposing one account's cache to another account's job on a shared host.

## Impact

On real staging hardware, a warm `tuist generate tuist ProjectDescription` runs in **~6.6 s vs ~22.5 s cold — ~3.4× faster**. Because that warm path is compute-bound (graph load, project generation), not cache-I/O-bound, the virtio-fs share retains **71–99 % of a block device's warmth** despite raw I/O being 2.5–16× slower (details in Validation). The cost side is negligible: materialization is a **~59 ms** APFS clonefile, and freshness convergence runs off the job-start path, so the cache adds **< 1 s to job start** in the common case (≤ 60 s worst case on a wedged host, then a cold-path fallback). And with the cache-signing grant the signatures are account-scoped, so a warm master's binaries validate as hits across every one of the account's VMs instead of being re-pulled per VM — the cold binary re-pull is what dominates a fresh VM's `tuist generate` today.

## How it works

The account a warm-pool VM will serve isn't known at boot (dispatch binds it only after the VM polls), so the cache is **materialized after dispatch**:

- A warm VM boots **generic**: tart-kubelet attaches an *empty* per-VM branch directory as a writable virtio-fs share at the cache root. No account data, no prediction.
- The server already stamps the pod's `tuist.dev/runner-account` label when it claims a job. The reconciler observes that label and **APFS-clonefiles that account's cache master tree into the VM's branch** (CoW, ~59 ms for a 2.4 GB / 6.3k-file tree), then writes a `cache-ready` marker.
- `dispatch-poll.sh` waits for `cache-ready` before starting the runner, so the guest never reads or writes the cache while the host is still clonefiling into it. The materialize is a clonefile, so this is normally sub-second; the wait is bounded (60 s ceiling) and, on timeout, the guest detaches to a local cold cache so a wedged host can't stall or corrupt the run.
- On job end the branch is promoted back to the account's master (job succeeded AND the cache changed) or discarded.

Because the account is known before any bytes are materialized, one account's cache can never reach a VM that runs another account's job, and every account on a host warms independently. Masters and branches are plain APFS directory trees (clonefile via `cp -c -R`), so there is no disk-image machinery. The host disk is fenced by a quota-bounded runner-cache volume, per-live-branch admission reservation, and watermark eviction; every space pressure degrades a job to the cold path rather than risking ENOSPC.

## Keeping masters fresh: converging toward a per-account HEAD

A host's master is otherwise only as fresh as the last job that ran *that account on that host*. To avoid a job on a lagging host paying remote misses for the delta, each account has a single canonical **volume HEAD** — a monotonic generation plus the inventory digest of its current warm set, last-writer-wins on promote (`runner_volume_heads`):

- On a successful, cache-changing job the guest archives the cache (a `ditto` zip, which preserves the signature xattrs — so the master is portable to any of the account's hosts without re-signing, exactly as the grant intends), uploads it to a presigned per-account master URL, and reports the new digest to bump the HEAD.
- Dispatch hands the next runner the account's HEAD (generation, digest, presigned download URL) via the response; the guest stages it into the status share.
- Dispatch hands the next runner the account's HEAD (generation, digest, presigned download URL); the guest stages it into the status share. **Off the job-start path**, the host fast-forwards its master toward that HEAD in the background — the running job already holds its own CoW branch, so the master refresh never touches it; it just makes the *next* job on that host start from the account's current warm set. The download is verified against the HEAD digest before it replaces the local master, and the host-side digest matches the guest's `cache_inventory` exactly, so the comparison is apples-to-apples.

All of this is best-effort and bounded: no HEAD, an already-current host, a digest mismatch, or any download/upload/extract failure leaves the host on its local master (the status quo — the job just pays a few remote misses for the delta). The HEAD report is bound to the account resolved from the Pod's server-stamped label, not the request body, so a runner can only advance the account it actually ran.

## The cache-signing grant, and why it is the core of this change

`ArtifactSigner` signs each cached binary with a payload validated on load. Historically that payload is the machine's MAC address, so a cache written by one VM fails validation in the next VM. A shared warm volume alone would therefore deliver warm manifests but every binary would still be rejected and re-pulled.

The grant fixes this. The server mints a short-lived EdDSA token scoped to the account (`scope=account-<id>`, `aud=tuist-runner-cache`, `iss=tuist-runners`) and delivers it to the job. When a valid grant is present, the CLI signs and validates against the account scope instead of the MAC. The scope is identical across all of the account's VMs, so a master clonefiled into a fresh VM validates as local hits. Verification is offline and any failure (no grant, bad signature, wrong audience/issuer, expired, or over-long lifetime) falls through to the MAC default, exactly like a developer machine.

Note on the signing model, since it is easy to misread: signatures are local only. `ModuleCacheRemoteStorage` strips the signature on upload and re-applies it on download into the local cache dir using the puller's scope. The remote holds bare bytes, so artifacts pushed from developer machines arrive unsigned and are re-signed by whichever runner pulls them. The grant is exactly what makes that re-signed local copy reusable across the account's VMs.

### Root cause of the server-side grant fix

`Environment.cache_grant_private_key` originally read only the encrypted secrets blob, which `decrypt_secrets/0` returns empty outside dev. The grant could therefore never be enabled in managed environments. Fixed by adding a `System.get_env("TUIST_CACHE_GRANT_PRIVATE_KEY")` fallback (the same pattern used for `database_url` and the license key) and wiring the key through External Secrets / 1Password.

## Components

- **tart-kubelet (`infra/tart-kubelet`)**: `VolumeManager` — allocate an empty branch at boot (with per-live-branch reservation), materialize the dispatched account's master into it (clonefile + atomic swap), converge a stale master toward the HEAD first, promote/discard on a dirty gate, watermark eviction, and a startup sweep of stale branches; darwin backend is `cp -c -R` + `df`. Prometheus metrics for outcome, warm/cold materialization, and convergence.
- **Runner image (`infra/runner-image/dispatch-poll.sh`)**: mounts the virtio-fs cache share, reads the host-staged `TUIST_CACHE_MAX_BYTES`, waits for `cache-ready`, reports the inventory-diff dirty marker, stages the account's HEAD for the host, and publishes its warm set as the new HEAD on a successful cache-changing job.
- **CAPI provider (`infra/cluster-api-provider-tuist`)** + **host bootstrap (`infra/macos-host-bootstrap`)**: provision the quota-bounded runner-cache volume and carry the config on both first-boot and drift-update paths, opt-in per host via `--runner-cache-volume-gib` (0 = off), with operator self-heal on a wedged tart-kubelet update.
- **Server (`server/lib/tuist/runners`)**: `CacheGrant.mint` (EdDSA JWT scoping cache signing to the account); `VolumeHeads` (a `runner_volume_heads` table) tracks each account's canonical volume HEAD — threaded into the dispatch response and bumped by a `POST /runners/volume-head` endpoint; `VolumeAffinities` (a `runner_volume_affinities` table + affinity-aware `pick_queued_top_k`) is a cold-start *optimization* — routing an account's jobs toward hosts that already hold its master — not a correctness mechanism.
- **TuistCacheEE (gitlink `2be9e2e`)**: offline `CacheSigningGrantVerifier` (with a 24 h `exp − iat` lifetime cap), grant-gated `ArtifactSignaturePayloadProvider`, and the byte-budget `CacheLocalStorage.clean(maxBytes:)` LRU prune.
- **Helm chart**: cache-volume operator flag plus the cache-grant private key wired through External Secrets / 1Password.

## Validation

Deployed to staging and validated end to end on real runner hardware.

- **Correctness**: volume provisions, mounts, and persists across jobs. Five host-side bugs were found by watching the box and fixed on this branch (CAPI drift path missing the config, `HostConfigHash` not re-rolling, APFS container resolution on a sealed `/`, and two tart share/disk suffix bugs).
- **Grant chain**: a real server-minted grant delivered to a runner verifies against the public key baked into the EE verifier, with all claims matching (`signature=VALID alg=EdDSA iss=tuist-runners aud=tuist-runner-cache scope=account-<id>`).
- **Cross-VM signature portability**: a runner test warmed a cache under the grant and re-generated; with the grant the binaries validate as hits, and with the grant removed (which reproduces a different VM's raw-MAC identity) the same binaries are rejected and re-pulled — proving the signature payload is account-scoped, not per-machine, so cross-VM reuse holds by construction.
- **virtio-fs is viable**: raw virtio-fs I/O is 2.5–16× slower than a block device, but `tuist generate`'s warm path is compute-bound (graph load, project generation), not cache-I/O-bound, so a warm generate over a virtio-fs cache **retained 71–99% of the block-device warmth** (6.6 s vs 6.5 s warm, vs 22.5 s cold). The signature xattr survives virtio-fs, and clonefile materialization is ~59 ms.

## Reviewer notes

- **Off by default in the chart; enabled in every managed env by these values.** The chart default is `macosFleet.runnerCacheVolume.gib: 0` (feature off, every VM boots cold), so it stays opt-in for self-hosted operators. The managed-env values turn it on: staging and canary at `gib: 40` (~2 masters at the 20 GiB default cap), production at `gib: 120` (~6 masters, so a production host keeps several accounts hot). The quota is an APFS ceiling that shares the container's free space — not a reservation — fenced by admission + watermark eviction, so space pressure degrades a job to the cold path rather than ENOSPC.
- **Warmth is still gated on the TuistCacheEE PR.** Enabling the values provisions the volumes and exercises materialization, manifest warmth, and HEAD convergence in all three envs — but the runners currently bake the released, grant-unaware CLI, so binaries still fail cross-VM validation and re-pull until it ships. The remaining path is: merge the EE PR, bump the `cli/TuistCacheEE` gitlink to the merged SHA (it points at branch commit `2be9e2e` today), release the CLI, rebake the runner image. Binary warmth lands when the rebaked image rolls out.
- **Rotate the cache-grant keypair for production.** The staging private key lives in the `tuist-k8s-staging` vault; the public key baked into EE is `KYXwBJW/2DSAldaOXMiaova/emzkOlBHMhYJkz63UA8=`.
- **Grant still needed under this model.** A master clonefiled from one VM into another (different raw MAC) only validates because the grant makes the signature account-scoped. A later simplification could drop per-artifact signing on the volume path entirely (the host is now the trust boundary), but that is out of scope here.
- **HEAD convergence rides `ditto` zip for the master archive**, which is expected to preserve the `tuist.cloud.metadata` signature xattr (proven for `ditto` tree copies; the archive round-trip is validated on staging when the feature ships). If it does not, convergence degrades to the status quo (unsigned artifacts miss and re-pull) — it never corrupts a cache. The archive is whole-master today; content-addressed delta sync is a follow-up. Server (`VolumeHeads`) and host (`MasterDigest`/`ReplaceMaster`) logic is unit-tested; the storage transport is staging-gated like the rest of the feature.

## Review findings addressed

First round:

- **Cross-account cache exposure** and **single-account-per-host** — fixed by materializing only after the account is known.
- **Concurrent branches don't reserve capacity** — admission reserves cap per live branch.
- **Kubelet restarts leak branches** — the startup sweep reaps stale branch dirs.
- **Committed long-lived grant + no lifetime cap** — the verifier rejects grants whose `exp − iat` exceeds a 24 h ceiling (requires `iat`, with a clock-skew allowance), and its public key is injectable so the test fixture is signed by a throwaway keypair rather than a real environment's key.
- **XCTest vs Swift Testing** — the new EE tests are converted to Swift Testing.

Second round (all six verified against the code):

- **[P0] Cross-account exposure after a *failed* dispatch.** The first round's "materialize only after the account is known" left a hole: `serve_claim` stamped the runner-account label (the materialize trigger) *before* minting the JIT, and `release_safely` re-queued a failed dispatch without removing it — so a Pod could carry account A's stale label, materialize A's cache, then win a committed claim for account B and run B's job on A's cache. Fixed by stamping the account label only *after* the dispatch fully commits (the owner label stays pre-mint for the egress NetworkPolicy), plus a defense-in-depth `SourceAccount == account` guard in `Finalize` that discards rather than promotes a mismatched branch.
- **[P1] Kubelet restart reaped live branches.** The startup sweep removed *every* branch, including one a surviving VM was still virtio-fs-mounting, and `recoverState` rebuilt entries with no volume state. `ReattachBranch` now reconstructs the attachment for recovered/adopted live VMs (so `Finalize` still promotes) and the sweep retains their branches, reaping only true orphans.
- **[P1] Guest cache-ready wait raced host materialization.** The 30 s guest wait was shorter than the host's 120 s convergence bound; on timeout the guest kept using the share while the host could still swap it. Fixed by taking convergence off the job-start critical path entirely: the host materializes from its fast local master (a CoW clonefile) and signals immediately, then fast-forwards the master toward HEAD **in the background** (the in-flight job holds its own CoW branch, so refreshing the master doesn't touch it — it just makes the next job on that host start fresher). The guest wait now only absorbs reconcile scheduling (60 s ceiling, normally sub-second) and, on timeout, detaches to a local cold cache so a late host swap can't corrupt the run. Worst-case job-start delay from the cache dropped from ~180 s to ~60 s, typically <1 s.
- **[P1] Affinity tolerance didn't bound head delay.** It compared the chosen candidate to the head, so a burst of affine jobs enqueued within the window could starve the head. It now returns the head once the head has itself waited longer than the tolerance (measured from now).
- **[P2] Affinity ran for volumeless fleets.** `record`/`select` ran on every claim, including Linux; now gated on `fleet_platform == :macos`.
- **[P2] Clean VM shutdown treated as job success.** `tart run`'s clean exit is the VM halting, not the job's conclusion, so `report_cache_dirty` now gates the dirty marker on the runner exit code (mirroring `report_volume_head`), carrying the job result to the host. (The runner rc catches infra/runner failures and cancellations; a step-failure where the runner still exits 0 promotes, which is acceptable since the artifacts are content-addressed and signature-validated.)

Third round:

- **[P0] Committed-but-abandoned Pod reused across accounts.** Post-commit stamping (2nd round) closed the *failed*-dispatch case, but a dispatch that *commits* and then loses its HTTP response leaves the runner never started while the Pod keeps polling — it could claim account B on the cache already materialized for A (the `SourceAccount` guard blocks the promote, not the guest *reading* A's cache). Fixed by making committed Pods one-shot: `check_not_stale` fetches the Pod first and 410s any Pod already carrying the account label, so the guest halts and the reconciler replaces it. A Pod that received its JIT runs in place and never polls again, so this only fires for the abandoned case.
- **[P1] HEAD digest/archive divergence.** The mutable per-account archive object and the separately-reported digest can point at different content. Convergence now verifies the downloaded archive's inventory digest equals the HEAD digest before installing it; on mismatch it keeps the local master. (Immutable, content-addressed archive keys are a tracked follow-up.)
- **[P1] HEAD publish could hang teardown.** The archive PUT and digest POST had no timeouts; both now carry connect/overall bounds so an object-storage stall can't keep the VM up and stall the warm pool.
- **[P1] Restart mistook idle branches for materialized.** The guest creates the (empty) `tuist` cache subtree at boot, and reattach treated that as the materialization signal — so a recovered idle VM skipped materialization and ran cold. The host now drops a `.host-materialized` marker and reattach keys off that.
- **[P2] HEAD staging raced the host read.** The guest stages the HEAD only after the dispatch response, which the server returns after the label patch that triggers convergence, so the background converge often read no HEAD and skipped. It now waits (bounded) for the staged HEAD.
- **[P2] Data-export doc.** `runner_volume_heads` and the `runner-volume-masters/{account_id}/tuist-cache.zip` archive are now documented in `data-export.md`.

Two third-round findings are **intentionally deferred** (not silent — flagged for reviewers):

- **Per-branch disk quotas (P1).** A tenant's branch is a writable share on the aggregate-quota volume; the advisory `TUIST_CACHE_MAX_BYTES` bounds only the Tuist CLI's own writes. The finding's suggested fix — per-branch APFS volumes — is incompatible with the design: `clonefile` (the ~59 ms CoW materialization that is the entire speed win) works only *within* one volume, so per-branch volumes would force a full byte copy of every master. Real enforcement would need a host-side branch-size monitor that stops a job whose branch exceeds its cap. Only Tuist cache writes land on the share (workflow checkouts/DerivedData go to the VM's own disk), and space pressure already degrades new branches to the cold path via admission + watermark eviction, so this is left as a documented limitation for v1.
- **`gib`-aware affinity capability (P2).** Affinity is gated on `fleet_platform == :macos`. A macOS fleet with `runnerCacheVolume.gib: 0` would still record benign affinity. All managed envs run `gib > 0`, so the platform gate is accurate for them; a precise per-host capability gate would need a new server config knob or a node-label round-trip, deferred as not worth it for a self-hosted-only, benign edge.

## Future improvement: shared versioned volume store

The most complete version of the above is a shared, versioned copy-on-write volume store: one current lineage per account that every job forks the latest of, so there is no per-host drift at all. That trades the per-host-copy model for a distributed CoW storage layer the whole fleet can fork from at local-disk speed — a real storage-infra investment (non-trivial on the Mac-mini fleet, which has no pooled-NVMe fabric), warranted only if per-host drift or per-host-copy storage cost become measured problems.

## Future improvement: runtime disk hot-plug (macOS 26)

The materialize-after-dispatch control flow is deliberately the same one a future block-device model would use. macOS 26 (WWDC 2026) added runtime storage hot-plug to Virtualization.framework via `VZXHCIController` (attach/detach mass storage to a running VM through `VZStorageDeviceAttachment`). Once `tart` exposes it, a generic warm VM can boot with no cache device and, on dispatch, CoW-clone the dispatched account's master and hot-attach it as a block device — same orchestration, block-device speed. Gated on `tart` support (2.32 has none today) and on measuring XHCI/USB mass-storage throughput against virtio-blk.

## How to test locally

The server and CLI unit tests run locally:

- Server: `mix test test/tuist/runners/volume_affinities_test.exs test/tuist/runners/cache_grant_test.exs test/tuist/runners_test.exs`
- tart-kubelet: `go test ./infra/tart-kubelet/...`
- CAPI provider: `go test ./infra/cluster-api-provider-tuist/...`
- host bootstrap: `go test ./infra/macos-host-bootstrap/...`

The volume substrate itself needs a real Mac host; it was validated on the staging runner fleet during development (correctness, cross-VM signature portability, and the virtio-fs numbers above).


server/lib/tuist/runners.ex
```diff
@@ -90,6 +94,91 @@ defmodule Tuist.Runners do
...
+  # one. Best-effort — any failure returns nil and the runner stays on its
+  # local master (the status quo).
+  defp volume_head_payload(account) do
+    key = volume_master_object_key(account.id)
+    head = VolumeHeads.get_head(account.id)
+
+    %{
+      generation: (head && head.generation) || 0,
+      digest: head && head.tree_digest,
+      download_url: Storage.generate_download_url(key, account, expires_in: @volume_master_url_ttl_seconds),
```
pepicrft: [priority 2] Restrict host downloads to trusted storage destinations

`volume_head_payload/1` builds the runner download location from configured object storage and sends it to the Mac host, whose downloader follows redirects. A misconfigured or compromised storage endpoint can point at localhost or a private address, turning this into a server-side request forgery primitive from every host. I reproduced the path with the real downloader against a loopback archive. Please use a fixed instance-controlled storage endpoint, or validate every resolved address and redirect target against an allowlist before the host fetches it.

infra/macos-host-bootstrap/bootstrap.go
```diff
@@ -1520,6 +1564,71 @@ func installVMCachePNInterface(ctx context.Context, client *ssh.Client, cfg Conf
...
+# off the always-real Data volume; fall back to a text parse for older macOS.
+CONTAINER=$(/usr/sbin/diskutil info -plist /System/Volumes/Data 2>/dev/null | /usr/bin/plutil -extract APFSContainerReference raw -o - - 2>/dev/null || true)
+if [ -z "${CONTAINER:-}" ]; then
+  CONTAINER=$(/usr/sbin/diskutil info / | awk -F'[[:space:]]*:[[:space:]]*' '/APFS Container:/{print $2; exit}')
+fi
+if [ -z "${CONTAINER:-}" ]; then
+  echo "could not determine APFS container for /System/Volumes/Data" >&2
+  exit 1
+fi
+sudo /usr/sbin/diskutil apfs addVolume "$CONTAINER" APFS "$VOL" -quota "${QUOTA_GIB}GiB"
```
pepicrft: [priority 1] Make the runner cache volume writable by tart-kubelet

The new Apple File System volume is created and mounted as root, but the launch daemon runs tart-kubelet as `cfg.SSHUser`. Nothing assigns ownership or creates a writable subdirectory, so the first branch allocation fails and every managed job falls back to a cold cache. The mounted-volume early return also prevents a later bootstrap run from repairing ownership. Please create and verify a runner-owned cache root on every run, including when the volume already exists.

infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
```diff
@@ -673,6 +684,20 @@ func (r *ScalewayAppleSiliconMachineReconciler) reconcileNormal(
...
+	// config the operator was trying to push when it exhausted its
+	// retries. If the desired host config has drifted since (configDrift
+	// is true against the stored hash the failure was recorded at), a new
+	// — typically fixed — config has landed, and it deserves its own retry
+	// budget rather than inheriting the old config's verdict. Clearing the
+	// terminal state here makes pushing a corrected operator image
+	// self-heal every host a bad rollout bricked, instead of requiring a
+	// manual status patch per host. A config that is still broken simply
+	// re-exhausts its retries and re-enters the terminal state.
+	if configDrift && terminalFailure {
```
pepicrft: [priority 1] Preserve the terminal retry cap for an unchanged broken config

This condition compares the desired host hash with the last successfully applied hash. While a broken desired config cannot be applied, the status hash remains old, so every reconcile still sees drift and clears the terminal failure and attempt count. The same bad config is then retried forever, defeating the cap. A focused regression test reproduced the reset on the next reconcile. Please persist the hash that exhausted its retries and clear the terminal state only when the desired hash changes from that failed hash.

infra/tart-kubelet/internal/podagent/reconciler.go
```diff
@@ -1046,6 +1129,11 @@ func (r *Reconciler) deleteByKey(ctx context.Context, namespace, name string) er
 	}
 	r.stopVNCForwarder(namespace, name, entry)
 
+	// Safety net: if the branch was not already finalized on a completion
+	// path (e.g. the Pod object vanished before we could observe its
+	// account), discard it so it never leaks. No-op once consumed.
+	r.finalizeVolume(entry, "", false)
```
pepicrft, line 1136: [priority 2] Preserve successful recovered branches during cleanup

Recovered store entries intentionally have `Run == nil`. When a recovered virtual machine is already stopped, reconciliation reaches `deleteByKey`, and this safety net finalizes with an empty account and `cleanExit == false`. That discards a successfully completed dirty branch instead of promoting it. A focused recovery regression test fails with the current behavior. Please recover enough completion and account state to finalize correctly, or avoid consuming the branch until the outcome is known.

server/lib/tuist/runners.ex
```diff
@@ -90,6 +94,91 @@ defmodule Tuist.Runners do
...
+         label when is_binary(label) <- get_in(pod, ["metadata", "labels", @account_label]),
+         {account_id, ""} <- Integer.parse(label) do
+      {:ok, account_id}
+    else
+      _ -> {:error, :account_unresolved}
+    end
+  end
+
+  defp volume_master_object_key(account_id) do
+    "runner-volume-masters/#{account_id}/#{VolumeHeads.reserved_tuist_cache()}.zip"
```
pepicrft: [priority 2] Commit the archive and digest atomically

Every job uploads to the same mutable object key and reports its digest in a separate request. Two jobs can interleave as upload A, upload B, report B, report A, leaving B's bytes paired with A's digest and generation. Hosts then keep redownloading or accept coordination metadata that does not describe the stored archive. Please publish to immutable generation-specific keys and atomically advance the head to that key and digest, or make the report conditional on the uploaded object version.

server/data-export.md
```diff
@@ -58,6 +58,9 @@ Sensitive authentication data (passwords, tokens) are excluded from exports.
...
+- **Runner cache volume master archives** (S3 `runner-volume-masters/{account_id}/tuist-cache.zip`): a per-account snapshot (a `ditto` zip that preserves the artifact-signature xattrs) of the account's warm Tuist cache set — compiled Manifests / ProjectDescriptionHelpers / Plugins and the binary-cache index a runner materializes to start a build warm. Published by a runner after a successful, cache-changing job and downloaded by a lagging host to converge toward the HEAD above. This IS customer-derived build cache content (no credentials). A single mutable object per account, overwritten by each publish (no version history); it is account-scoped under the `{account_id}` prefix and covered by the account's cache data lifecycle (export and deletion) like the other instance-managed cache buckets.
```
pepicrft: [priority 1] Add lifecycle deletion for runner volume archives

This says the new archive is covered by the account cache lifecycle, but the existing retention matcher only recognizes the current account, project, and artifact path shapes, and account deletion does not remove `runner-volume-masters/{account_id}/...`. A focused server regression test produced an empty deletion list for this key. The customer-derived build cache can therefore remain indefinitely after account deletion. Please add the prefix to retention and account-deletion cleanup, then cover both paths with tests.

pepicrft (commented):
I found seven issues that should be addressed before merging. The retry, recovered-branch, and retention findings were reproduced with focused failing regression tests. The account-wide cache trust and host-network findings were also validated with focused macOS and loopback probes. Six findings are inline. The submodule finding is below because GitHub cannot attach a parent-repository review comment to a line inside `cli/TuistCacheEE`.

[priority 1] Bind persisted artifact trust to the artifact bytes

The account-wide signature created by `ArtifactSignaturePayloadProvider` records only the account scope and does not bind the signature to the artifact bytes. I confirmed that modifying a file in place on macOS preserves the extended attribute. An untrusted pull-request job can therefore make the inventory dirty with a harmless cache addition, alter an already signed artifact, and have the host publish that tampered artifact as trusted for later jobs in the same account. Please sign a content digest, include an expiry, and separate trust contexts so untrusted jobs cannot create artifacts accepted by trusted jobs.

infra/macos-host-bootstrap/bootstrap.go
fortmarek: Fixed in the bootstrap script: it now enforces runner ownership of the cache-volume root (`diskutil enableOwnership` + `chown` to `cfg.SSHUser` + a `sudo -u <user> test -w` writability check) on **every** run, not just first provision — so an already-mounted, root-owned volume from an earlier bootstrap is repaired rather than skipped by the mounted-volume early return.

infra/tart-kubelet/internal/podagent/reconciler.go
fortmarek: Fixed: `podStatus`, on observing a recovered VM stopped, now finalizes its branch from the account label + the guest dirty marker (which exists only on a clean completion and is rc-gated for job success) before `deleteByKey`, treating the observed stop as a clean exit. A completed dirty branch is now promoted instead of discarded; the finalize is idempotent with the terminal path.

infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
fortmarek: Fixed: added a `FailedHostConfigHash` status field, set when a config exhausts its retry budget, and the self-heal (`shouldClearTerminalFailure`) now clears the terminal state only when the desired hash differs from that failed hash — not from the last-applied hash, which a broken (never-applied) config never advances, so the cap was reset every reconcile. Regression test covers same-hash-stays-terminal vs new-hash-retries.

server/lib/tuist/runners.ex
fortmarek: Fixed with a two-layer guard: the server now validates the generated download/upload URLs with a new `Tuist.URL.public_host_url?/1` (same private/loopback/link-local rejection as `public_url?/1`, but permits the presigned query string) before handing them out — a non-public host yields no HEAD (job stays on its local master) rather than a fetch against an internal address. And the host `curl` drops `-L`, so a presigned URL is fetched directly with no redirect-following, removing the redirect-based SSRF path.

server/data-export.md
fortmarek: Fixed: `Accounts.delete_account!/1` now explicitly deletes the whole `runner-volume-masters/{account_id}/` prefix from object storage (best-effort; a storage failure never fails the deletion), with tests for the purged prefix and for deletion still succeeding when the purge fails. I did not add it to the handle-based artifact retention: that scanner keys off the account *handle* as the first path segment, but this archive is keyed by `account_id`, so a retention target would treat every object as orphaned and delete live accounts’ caches. Account deletion is therefore the removal path, and `data-export.md` is corrected to say so.

server/lib/tuist/runners.ex
fortmarek: The unsafe outcome — a host adopting content that does not match the HEAD digest — is already prevented: `convergeMaster` verifies the extracted archive's inventory digest equals the HEAD digest before `ReplaceMaster`, and keeps the local master on mismatch. The residual (a diverged HEAD→object mapping causing a bounded, self-correcting re-download, never wrong content installed) I am deferring to a follow-up: the immutable generation/content-keyed object keys also make superseded archives accumulate, which needs its own cleanup, so I would rather not expand the protocol in this PR. Leaving this thread open for your call on the deferral.

fortmarek:
**On the artifact-trust finding (untrusted fork-PR jobs can poison the shared account cache):** confirmed, and taking it as a dedicated follow-up rather than a rushed patch here — misclassifying the trust boundary is itself the vulnerability, so it deserves care.

Planned approach: **gate writes to trusted jobs** — untrusted fork-PR jobs read the warm cache but never promote/publish to the shared master, with content-digest-bound signatures in EE as defense in depth. The blocker for a correct implementation is that fork/trust status is not in the current data model: the `workflow_job` webhook (`Dispatch.handle_webhook`) carries `head_branch`/`head_sha` but not the head repository, so "is this a fork" needs deriving (head_repository vs base, via the PR/workflow_run) and threading through dispatch → a Pod label → host `Finalize` (discard, do not promote) + guest (no HEAD publish) + `CacheGrant.mint` gating. Tracked as a focused change.

The other six review findings are addressed in this PR (each replied + resolved inline). The immutable-keys finding (#5) is scoped to the shipped host-side digest verification for v1 — which already blocks the unsafe outcome — with the generation-keyed rework deferred (noted on that thread).

_🤖 via [Claude Code](https://claude.com/claude-code)_

server/lib/tuist/runners.ex
pepicrft: The download side is fixed, but the paired upload path still follows redirects. `dispatch-poll.sh` runs `curl -fsSL ... -X PUT --upload-file ...`, and `-L` follows a redirect supplied by the storage endpoint.

I validated this with a local 307 redirect using the same curl flags. The redirected internal endpoint received the full 32,891-byte `PUT` request body. Because the threat model includes a compromised or misconfigured storage endpoint, validating only the initial host does not validate this redirect target.

Please remove redirect following from uploads as well. The initial host check is useful, but the storage-destination boundary is not fully closed while the guest upload can be redirected.

server/lib/tuist/runners.ex
pepicrft: I validated the shipped guard directly, and it does not prove that the downloaded content matches the head record. `inventoryDigest` hashes only sorted top-level entry names and never reads file bytes. The existing `TestTreeDigestMatchesMaster` writes different bytes under the same entry name and expects the digests to match; I ran that test and it passes.

An archive with the expected names but different bytes therefore passes `TreeDigest` and can replace the local master. The current artifact signature does not close this gap because it signs only the account scope, not the artifact bytes.

For archives with different names, the mismatch is also not globally bounded or self-correcting. Each later job can download and reject the same archive until another successful dirty publication repairs the mapping. I think this needs a content digest plus an immutable generation-specific object key atomically recorded with that digest.

server/data-export.md
pepicrft: This adds an attempt, but not durable deletion. `Accounts.delete_account!/1` removes the account before calling `purge_account_cache_masters/1`. A returned `{:error, reason}` is ignored, raised errors are rescued, and there is no retry record after the account and its storage configuration are gone.

I ran the focused test at `accounts_test.exs:3517`. It passes precisely when storage deletion raises and the account still disappears. Since the handle-based scanner cannot recognize this account-identifier prefix, that failure can leave the customer-derived archive indefinitely.

Please either complete the prefix deletion before removing the account and propagate failure, or persist a durable cleanup job or tombstone that retains enough storage context to retry until deletion succeeds.

pepicrft:
I validated the current follow-up implementation and explanations against the code and focused tests.

Current disposition:

- The cache-volume ownership fix is correct. Ownership repair and the writability check run for both new and already-mounted volumes.
- The terminal retry-cap fix is correct and has focused regression coverage.
- The recovered-branch fix is logically correct. The complete pod-agent test package passes, although this exact recovery scenario still lacks a focused regression test.
- The account-wide artifact-trust vulnerability is confirmed and remains unfixed in this pull request.
- The archive race remains unsafe because the reported digest covers entry names, not file contents.
- The storage-destination fix closes direct private downloads and download redirects, but the guest upload still follows redirects.
- Account deletion attempts archive cleanup but can permanently abandon it after a storage failure.

I do not think the feature should merge while enabled in managed environments. The safer merge path is to set `runnerCacheVolume.gib` to zero for staging, canary, and production, merge the dormant substrate if desired, then enable it after the trust boundary, content digest, immutable archive mapping, upload redirect, and durable deletion work lands.

For the planned trust fix, server-derived trust and host-enforced discard are the right foundation. `CacheGrant` should not be the authoritative write gate because it runs inside the guest that executes untrusted code and also supplies the account scope needed to validate warm entries. Shared-write denial must be independently enforced by the host and server.

Validation performed at the current head:

- 313 focused Elixir tests passed, with 2 excluded.
- The Apple File System bootstrap package passed.
- The complete pod-agent package passed.
- Focused retry-controller tests passed.
- A macOS probe confirmed modified bytes retain the trusted extended attribute.
- A redirect probe confirmed the upload follows a redirect and sends the complete request body.

server/lib/tuist/runners.ex
fortmarek: Good catch — fixed: the guest `report_volume_head` upload now uses `curl -fsS` (no `-L`) for both the presigned `PUT` and the HEAD-report `POST`, so neither follows a redirect. Combined with the download-side `-fsS` and the server-side `public_host_url?` host check, the storage-destination boundary is closed on both read and write.

server/lib/tuist/runners.ex
fortmarek: You're right, and my earlier "the unsafe outcome is already prevented" was wrong: `inventoryDigest`/`TreeDigest` hash sorted top-level entry names, not file bytes (that is by design for the cheap dirty-marker/HEAD comparison), so an archive with the same names but tampered bytes passes verification. This does not close the tampering gap. I am NOT claiming this as fixed — the content digest (over bytes) + immutable generation-specific object key atomically recorded with it is folded into the deferred security-hardening work alongside the trust boundary (task tracked). Leaving this thread open.

fortmarek:
**Trust boundary — implemented via fork-exclusion** (the direction we discussed).

Only a trusted (same-repo, non-fork) job touches the shared per-account cache:
- **Server** (`job_trusted?/2` in `serve_claim`): fetches the workflow run and treats the job as trusted only when `head_repository == repository` (base). **Fail-closed** — a missing run id, API error, unexpected shape, or any exception ⇒ untrusted ⇒ cold. Untrusted jobs get **no cache-signing grant** (cache stays MAC-local, not account-portable), **no volume HEAD** (no convergence; the guest has no upload URL so it cannot publish), and a `tuist.dev/runner-cache-untrusted` label.
- **Host** (tart-kubelet): honors the label — `maybeMaterializeVolume` skips materialize and leaves `SourceAccount` empty, so `Finalize` discards the branch. An untrusted job can neither read the warm master nor promote into it.

Enforcement is **host + server, not the grant** — per your point that `CacheGrant` runs inside the untrusted guest and supplies the read scope, so it can't be the write gate. With forks excluded, the master only ever holds trusted content, so the **content-digest + immutable-archive-key** work becomes deferred hardening rather than a merge gate (defense against a storage-layer compromise). Upload-redirect is fixed. The remaining open item from your list is **durable account-deletion cleanup**.

_🤖 via [Claude Code](https://claude.com/claude-code)_

infra/tart-kubelet/internal/podagent/volume_reconcile.go
```diff
@@ -0,0 +1,321 @@
...
+		return // not dispatched yet — nothing to materialize
+	}
+
+	// Fork-exclusion: an untrusted job never touches the shared cache. Skip
+	// materialize (the guest runs cold on its isolated, empty branch) and leave
+	// SourceAccount empty so Finalize's SourceAccount==account guard discards the
+	// branch — the job can neither read the account's warm master nor promote
+	// into it. Mark materialized so this doesn't re-enter, and still signal
+	// cache-ready so the guest stops waiting and runs.
+	if pod.Labels[runnerCacheUntrustedLabel] == "true" {
```
pepicrft, line 124: [priority 1] Preserve the untrusted decision across tart-kubelet recovery

The normal path leaves `SourceAccount` empty here, but both recovery paths reconstruct it unconditionally from the ordinary `tuist.dev/runner-account` label. They do not check `tuist.dev/runner-cache-untrusted`.

I reproduced this with a focused host test:

1. Allocate a branch and leave `SourceAccount` empty, matching this untrusted branch.
2. Write attacker-controlled cache content.
3. Simulate a tart-kubelet restart with `ReattachBranch` followed by the exact `RunnerAccountFromPod` assignment used by `recoverState`.
4. Complete the job successfully with a dirty cache marker.

The test returned `promoted`, and the attacker-controlled branch became account 42’s master. The adoption path in `reconciler.go` performs the same unconditional assignment. This requires tart-kubelet to restart while the untrusted job is active, but it does not require a storage compromise. The volume also contains unsigned compiled project-description helpers that are reused when their hash directory exists, so attacker-controlled content can reach a later trusted job.

Please make both recovery paths preserve the untrusted decision and add a focused recovery regression test.

server/lib/tuist/runners.ex
```diff
@@ -377,19 +562,57 @@ defmodule Tuist.Runners do
...
+  # the stamp (mint or ClickHouse error → release_safely re-queues the job while
+  # the Pod keeps polling) leave a stale account on a Pod that later wins a claim
+  # for a DIFFERENT account — the host would then run the second account's job on
+  # the first account's materialized cache. Post-commit stamping guarantees only
+  # the account the VM actually runs ever triggers materialization. Best-effort:
+  # a stamp failure degrades to a cold (unmaterialized) job, never a
+  # wrong-account one.
+  defp stamp_account_label(namespace, pod_name, account, trusted) do
+    labels = %{@account_label => Integer.to_string(account.id)}
+    labels = if trusted, do: labels, else: Map.put(labels, @cache_untrusted_label, "true")
```
pepicrft, line 601: [priority 2] Reject volume-head reports from untrusted pods

This patch deliberately stamps both the account label and the untrusted label, but the volume-head report path resolves authorization only from `tuist.dev/runner-account`. It never rejects `tuist.dev/runner-cache-untrusted`.

I reproduced this with a focused controller test:

1. Return a successful Kubernetes token review for a pod carrying both labels.
2. Submit an attacker-chosen digest to `/api/internal/runners/volume-head`.
3. Read the persisted volume head.

The endpoint returned status 204 and stored the attacker-chosen digest as generation 1. Workflow code can reach this path because the runner is an administrator with passwordless privilege elevation, the projected credential is readable inside the guest, and it remains valid for up to eight hours. Withholding the upload address prevents direct archive-byte upload, but the fork can still corrupt shared coordination metadata, force repeated downloads, and generate database writes.

Please reject the untrusted label in the report authorization path and cover it with a focused endpoint test.

pepicrft (commented):
Two security boundaries remain after validating the fork-exclusion follow-up. Both findings below were reproduced with focused temporary regression tests. The tests were removed after validation and the worktree is clean. I am withholding approval because the Priority 1 path allows untrusted cache content to become shared after host recovery.

infra/tart-kubelet/internal/podagent/volume_reconcile.go
fortmarek: Fixed in 730a056876. Both recovery paths (`recoverState` and the createPod adoption path) now go through a shared `ReattachVolumeForPod` helper that only reconstructs `SourceAccount` for trusted pods, so an untrusted (fork) branch stays discard-only after a kubelet restart and can never be promoted into the account master. Added a focused Go recovery regression test (`TestReattachVolumeForPodPreservesUntrusted`: untrusted branch keeps an empty SourceAccount and Finalize discards it; `TestReattachVolumeForPodTrustedPromotes` guards the trusted side).

server/lib/tuist/runners.ex
fortmarek: Fixed in 730a056876. `account_id_for_sa/2` now rejects any pod carrying `tuist.dev/runner-cache-untrusted` before resolving the account, fail-closed (`{:error, :cache_untrusted}` → the controller falls through to `401`). So a compromised fork job can no longer advance the account's shared HEAD by POSTing a promote, even though its SA token is valid. Covered by a focused `account_id_for_sa/2` test group (trusted resolves, untrusted rejected, missing-label unresolved).

pepicrft:
I revalidated `730a056876`. The recovery and untrusted volume-head reporting findings are fixed. Two earlier findings remain open:

- [Archive and digest atomicity](https://github.com/tuist/tuist/pull/11788#discussion_r3578135203): content verification hashes entry names rather than file bytes. This remains explicitly unfixed and deferred.
- [Durable archive deletion](https://github.com/tuist/tuist/pull/11788#discussion_r3578135209): cleanup is best effort after account deletion, ignores storage failures, and has no durable retry.

I am approving this pull request, but the approval does not mark these findings as resolved. Please keep both threads open and ensure the deferred work is tracked.

pepicrft (approved):
Approved with the two remaining findings explicitly called out in the pull request discussion. This approval does not resolve or withdraw those findings.

server/lib/tuist/runners.ex
fortmarek: Deferred as agreed; not fixed in this PR. Tracked in #11840 (immutable generation-keyed objects + atomic HEAD advance / version-conditional report). Leaving this thread open per your request.

server/data-export.md
fortmarek: Deferred as agreed; not fixed in this PR. Tracked in #11841 (add the prefix to retention + durable, retrying account-deletion cleanup, tests for both paths, and a data-export.md update). Leaving this thread open per your request.

PR #11791, opened 2026-07-13
fix: stop orphaned Hetzner volumes (kura reclaim + grace-period PV reaper)

fortmarek:
Stops the tuist-workloads Hetzner project from accumulating orphaned block volumes, which had grown to ~36 TB / ~€2k per month (~95% of the project's storage bill). This PR is the prevention; the pre-existing debris was cleaned up out-of-band (documented at the end).

Two complementary code changes.

## 1. kura-controller: reclaim data volumes at teardown

The kura-controller now flips each of an instance's data PVs to `reclaimPolicy: Delete` before it tears the PVC down (new `reclaimDataVolumes` helper), called from both teardown paths:
- the deletion finalizer branch (instance removed → StatefulSet GC'd → `whenDeleted: Delete`), and
- `reconcileStaleDataStorage` (storage-class / size / pinned-node change → StatefulSet + PVCs recreated). Reclamation runs *before* the StatefulSet delete here: `whenDeleted: Delete` owner-references the PVCs to the StatefulSet, so a foreground delete could GC them (and release their PVs) before we list them.

Adds the `persistentvolumes` `update;patch` RBAC the flip needs (kubebuilder marker + the `*-node-reads` ClusterRole).

## 2. Chart: grace-period reaper for orphaned Retain PVs

The kura fix only covers kura. CNPG (and any operator we don't own) still deletes PVCs on rebuild/restore/failover and leaves the `hcloud-volumes` PV `Released` under the `Retain` default — that is how the tuist Postgres cluster's June rebuilds left ~2 TB of orphaned 100 GB volumes. `templates/released-pv-reaper.yaml` is a CronJob (shipped to every cluster) that reaps PVs which have been `Released` longer than `graceHours` by flipping reclaim to `Delete` and deleting them, so the CSI driver deletes the backing Hetzner volume.

Safe by construction:
- A PV is only `Released` once its PVC is actually deleted; a scaled-down but still-wanted volume keeps its PVC `Bound` (StatefulSet `whenScaled: Retain`) and is never touched.
- The grace window (default 7d) is the recovery window for an accidental deletion — a `Released`/`Retain` PV can still be re-bound by hand within it.
- Scoped to `reclaimPolicy: Retain` + `storageClass: hcloud-volumes`, so node-local (`scw-local-nvme`, already `Delete`) and non-CSI volumes are never touched.
- Each candidate is re-fetched and re-validated (full predicate + UID) immediately before acting, so a PV recovered or replaced between the scan and the delete is skipped (`kubectl delete` does no resource-version check).
- `dryRun` (default `true`) only logs what it would reap; arm per-env after reviewing a run's logs.

`Retain` stays the cluster default so a botched database operation is still recoverable — this just stops the debris accumulating forever, rather than the riskier alternative of making the StorageClass `Delete`.

No `PrometheusRule`: managed clusters run no Prometheus Operator (they scrape via Alloy into Grafana Cloud), so a CRD would break the render; alert on the reaper's log lines or `kube_persistentvolume_status_phase{phase="Released"}` in Grafana Cloud instead.

## Why not just flip the StorageClass to Delete

Tempting one-liner, but wrong for databases: `Retain` is the safety net that keeps data around after a botched CNPG failover/restore. `Delete` would destroy it the instant the PVC is removed. Keeping `Retain` + reaping deliberately after a grace period preserves the safety net while ending the accumulation.

## One-time cleanup already performed (out-of-band)

The ~249 pre-existing orphans were deleted manually before this prevention landed. Method and result, for the record:

- **Cross-checked every detached volume against all 7 CAPI clusters** in the project (tuist, tuist-canary, tuist-staging, tuist-preview, atlas-production, hive-production, once-production — each backs its own CSI volumes in the shared Hetzner project). A volume was only deleted if its PV was bound to no PVC in any of the seven.
- **249 volumes deleted, 0 failures; project 265 → 16.** All 16 remaining map to a live Bound PVC (15 attached + one scaled-down-but-bound Postgres replica).
- **~36 TB reclaimed, ~€2,083/month (~€25k/year).**
- Done in batches by risk: per-account kura cache meshes (169), scaled-down staging dev meshes needing PVC/PV cleanup first (12), verified database orphans incl. the tuist Postgres relics (51), and the separate-cluster set once those clusters were reachable (17).
- **Retroactive safety check: zero live PVCs across all 7 clusters reference any deleted volume** — confirming nothing live was reaped.

## Validation

- `go build`, `go vet`, full `go test ./...` kura-controller suite pass. New tests: `TestKuraInstanceReconcileDeletionReclaimsDataVolumes`, `TestKuraInstanceReconcileStaleStorageReclaimsOldVolume`.
- `helm template` renders the whole chart and the reaper cleanly; `enabled: false` gate verified.
- Reaper filter + UID re-validation unit-tested against mock PVs: reaps only the old `Released`/`Retain`/`hcloud-volumes` PV, and keeps `Bound`, recently-released, `Delete`-policy, other-StorageClass, missing-timestamp, and stale-UID (recovered/recreated) PVs.

### DIFF
--- a/infra/helm/tuist/templates/released-pv-reaper.yaml
+++ b/infra/helm/tuist/templates/released-pv-reaper.yaml
@@ -38,7 +38,6 @@ apiVersion: v1
 kind: ServiceAccount
 metadata:
   name: {{ $name }}
-  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "tuist.labels" . | nindent 4 }}
     app.kubernetes.io/component: released-pv-reaper
@@ -77,7 +76,6 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ $name }}
-  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "tuist.labels" . | nindent 4 }}
     app.kubernetes.io/component: released-pv-reaper
@@ -148,7 +146,6 @@ apiVersion: batch/v1
 kind: CronJob
 metadata:
   name: {{ $name }}
-  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "tuist.labels" . | nindent 4 }}
     app.kubernetes.io/component: released-pv-reaper
--- a/infra/helm/tuist/templates/released-pv-reaper.yaml
+++ b/infra/helm/tuist/templates/released-pv-reaper.yaml
@@ -1,0 +1,176 @@
+{{- if .Values.pvReaper.enabled }}
+{{- /*
+Grace-period garbage collector for orphaned Hetzner block volumes.
+
+The hcloud-volumes StorageClass provisions PVs with reclaimPolicy Retain (see
+infra/k8s/mgmt/bootstrap/hcloud-csi-values.yaml) — a deliberately conservative
+default so a botched operator action can't destroy data. The cost is that any
+operator that deletes a PVC (CNPG on a cluster rebuild/restore/failover, a
+StatefulSet teardown, a namespace removal) strands the underlying Hetzner volume
+forever: the PV goes Released and nothing reaps the block storage. That is how
+this project accumulated ~36 TB / ~€2k-per-month of dead volumes.
+
+The kura-controller now reaps its own volumes at teardown, but we don't own
+CNPG's code, so Postgres (and any future operator) needs an out-of-band sweep.
+This CronJob is that backstop: once per run it reaps every PV that has been
+Released for longer than graceHours by flipping its reclaim policy to Delete and
+deleting it, which lets the CSI driver delete the Hetzner volume.
+
+Why this is safe:
+  * A PV only becomes Released when its PVC was actually deleted. A scaled-down
+    but still-wanted volume keeps its PVC Bound (StatefulSet whenScaled: Retain),
+    so it is never Released and never touched here.
+  * The grace window is the recovery window: within graceHours a Released Retain
+    PV can still be re-bound by hand, so an accidental PVC deletion is
+    recoverable. Only volumes abandoned for longer are reaped.
+  * Scoped to storageClass hcloud-volumes and reclaimPolicy Retain, so it never
+    touches node-local (scw-local-nvme, already Delete) or non-CSI volumes.
+  * dryRun (default) only logs what it would reap. Review a run's logs in Loki /
+    Grafana Cloud, then set pvReaper.dryRun=false per env to arm it.
+
+There is intentionally no PrometheusRule here: managed clusters have no
+Prometheus Operator (they scrape via Alloy into Grafana Cloud). Alert on the
+reaper's log lines or on kube_persistentvolume_status_phase{phase="Released"} in
+Grafana Cloud instead.
+*/ -}}
+{{- $name := include "tuist.componentName" (dict "root" . "component" "released-pv-reaper") }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ $name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "tuist.labels" . | nindent 4 }}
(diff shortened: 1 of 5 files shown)


pepicrft:
I checked this against Kubernetes 1.34.6, the external storage provisioner 6.2.0 running in production, Hetzner's chart defaults, and CloudNativePG's volume ownership behavior.

The grace-period approach is sound. Kubernetes explicitly supports using the last phase transition time to clean up volumes that have remained Released for a defined period, and the Kura path correctly changes the reclaim policy before deleting the claim.

I think we should make one correction before turning off dry mode: after changing a Released volume from Retain to Delete, do not immediately delete the Kubernetes persistent-volume object. The external provisioner is responsible for deleting the Hetzner volume first and then removing the Kubernetes object. Since these volumes used Retain while bound, the provisioner's deletion finalizer is not guaranteed to be present. The immediate deletion can therefore win the race and leave the Hetzner volume orphaned.

My suggested follow-ups are:

- Remove the explicit kubectl delete operation and let the external provisioner complete reclamation.
- Make the policy patch conditional on the resource version fetched during revalidation.
- Remove the delete permission from the reaper's cluster role.
- Default the Hetzner-specific reaper to disabled in the general chart and enable it explicitly in managed environments.
- Use a kubectl version supported by the Kubernetes server. The image currently contains 1.31.1 while production is 1.34.6.
- Report volumes that remain Released with a Delete policy, since that indicates a provisioner failure.

The current dry-mode configuration makes this safe to merge, but I would keep it in dry mode until the deletion ordering is corrected. A separate one-time Hetzner inventory will also be needed for cloud volumes whose Kubernetes objects have already disappeared, since this reaper cannot discover those.

pepicrft (approved):
Approved with the follow-up above. The current configuration remains in dry mode, so it is safe to merge. Please correct the deletion ordering before enabling actual reclamation.

infra/helm/tuist/templates/released-pv-reaper.yaml
```diff
@@ -0,0 +1,196 @@
...
+metadata:
+  name: {{ $name }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "tuist.labels" . | nindent 4 }}
+    app.kubernetes.io/component: released-pv-reaper
+data:
+  reap.sh: |-
+    #!/bin/sh
+    set -eu
```
esnunes: Since the reaper's only observability is its log lines (no PrometheusRule, by design), the `set -eu` without `pipefail` (released-pv-reaper.yaml:87) is worth tightening. In `candidates=$(kubectl get pv -o json | jq ...)` the pipeline's exit status is jq's, so if `kubectl get pv` fails (API blip, throttle, discovery error) jq just gets empty input, `count` is 0, and the Job logs `0 PV(s) …` and exits 0 — a broken run looks identical to a healthy empty one.

Adding `set -o pipefail` (so a kubectl failure aborts and fails the Job) would surface it via failedJobsHistoryLimit / log alerting. Cheap one-liner.

esnunes (approved):
Really solid piece of work — the problem is well diagnosed, the two-layer approach (kura-controller reclaims what it owns, the CronJob backstops what it doesn't) is the right shape, and the safety engineering is genuinely careful: keeping Retain as the cluster default, the grace window as a recovery buffer, the dryRun-by-default rollout, the UID re-validation, and the commit-3 reorder that closes the stale-storage teardown race. The controller side in particular I couldn't fault — the reclaim's label List provably matches the production PVCs, the finalizer is only removed after the flip succeeds, and it's cleanly idempotent.

I left one small, non-blocking suggestion inline about adding `set -o pipefail` to the reaper's scan, so a failed `kubectl get pv` can't silently look like a healthy empty run — since the logs are the only observability channel. Otherwise this looks great to me. Nice cleanup.

PR #9692, opened 2026-03-03
fix(cli): propagate module map flags to configuration-level setting overrides

pepicrft:
## Summary

The `ModuleMapMapper` adds `-Xcc -fmodule-map-file=` flags to `target.settings.base["OTHER_SWIFT_FLAGS"]` so the Swift compiler can find ObjC modules from dependencies. However, Xcode applies settings in order: base, then configuration. If a configuration sets `OTHER_SWIFT_FLAGS` without `$(inherited)` (which is common, e.g. `"-D DEBUG -D FEATURE"`), the configuration value completely replaces the base value and all module map flags are lost. This causes `Unable to find module dependency` errors when building.

This fix makes the mapper also append module map flags to any configuration that overrides `OTHER_SWIFT_FLAGS`, `OTHER_CFLAGS`, or `HEADER_SEARCH_PATHS`. Configurations that don't set these keys are left untouched since they correctly inherit from base.

> [!NOTE]
> If you are hitting this issue, a quick workaround is to add `$(inherited)` to your configuration-level `OTHER_SWIFT_FLAGS`. For example: `"OTHER_SWIFT_FLAGS": "$(inherited) -D DEBUG -D FEATURE"`. This lets the base flags (where Tuist injects the module map paths) flow through to each configuration.

Resolves https://github.com/tuist/tuist/issues/9676

## Test plan

- [x] Added unit test verifying module map flags are propagated to configuration-level overrides
- [x] Existing `ModuleMapMapper` tests still pass
- [x] Verified end-to-end with the reporter's fixture project: the `Unable to find module dependency: 'FirebaseCrashlytics'` (and similar) errors are gone

PR #11797, opened 2026-07-13
fix(cli): avoid duplicating module map flags in inheriting configurations

Loupehope:
### What changed

Avoid duplicating generated dependency module-map flags and header search paths in a target configuration when that configuration already inherits the target’s base settings through `$(inherited)`.

The mapper continues to add the generated `<Target>-deps.modulemap` flag and transitive dependency header search paths to the target base settings. A configuration with `$(inherited)` receives those values from base and is left unchanged.

### Why

Tuist 4.202.1 consolidates transitive dependency module maps into a single `<Target>-deps.modulemap`. The mapper was still adding that same map to both base settings and every configuration that explicitly contained OTHER_SWIFT_FLAGS, OTHER_CFLAGS, or HEADER_SEARCH_PATHS.

For configurations using `$(inherited)`, this produced duplicate `-fmodule-map-file` arguments and duplicate header search paths in the final Xcode build settings. For large Swift targets, repeated compiler arguments can noticeably increase the Planning Swift module phase.

This complements https://github.com/tuist/tuist/pull/9692. That change correctly propagates module-map flags to configuration-level overrides that do not inherit base settings. This change preserves that behavior: flags are still added when an overridden setting does not contain `$(inherited)`, and are skipped only when they are already inherited from base.

### User impact

Generated projects no longer duplicate dependency module-map flags and header search paths in inheriting build configurations. Targets with configuration-specific flags such as Debug continue to receive dependency module maps through their base settings.

### Validation

Added coverage that verifies the base settings of a dependent target receive the generated A-deps.modulemap flags and B header search path, while a Debug configuration containing $(inherited) remains unchanged. Existing coverage continues to verify that non-inheriting configuration overrides receive the required flags.


Loupehope:
Thanks for tuist! @fortmarek @pepicrft
I hope the pull request will be considered.

pepicrft (approved):

fortmarek (approved):
Looks good, thanks!

PR #11799, opened 2026-07-13
feat(cli): support package traits in native integration

pepicrft:
## What changed

Native packages declared through `Project.packages` can now select traits alongside each package declaration. The project description models new declarations as `Package.Dependency`, with a typed dependency kind and a `Set<Trait>` that includes `.defaults`.

Trait selections flow through manifest loading, the project graph, Xcode project generation, and existing Xcode project loading. The change also updates XcodeProj to [version 9.14.0](https://github.com/tuist/XcodeProj/releases/tag/9.14.0) and adds a command-line fixture whose local package only compiles when its selected trait is written to the generated project.

## Why

Xcode 26.4 stores selected package traits on local and remote package references. Tuist's native package integration could not express or preserve that configuration, which forced users to maintain otherwise generated Xcode projects manually.

The project description should also feel familiar to users who already declare dependencies with [Swift Package Manager](https://www.swift.org/package-manager/). Keeping Tuist's model aligned gives future package additions a clear reference instead of introducing another similar but incompatible abstraction.

## Root cause

The project model did not carry the package-reference `traits` field. The previous XcodeProj version also predated support for reading and writing it.

## Approach

Package traits use the existing `Project.packages` label and package-construction syntax:

```swift
let project = Project(
    name: "App",
    packages: [
        .package(path: "Package", traits: ["NativeIntegration"]),
    ]
)
```

`Package.Dependency` follows the Swift Package Manager shape. Each value contains a file-system, source-control, or registry kind and a typed trait set. Source-control and registry requirements remain separate, and source-control factories support the familiar `from`, `exact`, `branch`, and `revision` labels. Trait encoding is sorted so equivalent sets serialize deterministically.

The default trait set is `[.defaults]`. An empty set disables all default traits, while string literals select named traits. A selection containing exactly `.defaults` is normalized to an omitted Xcode trait field, preserving the generated-project behavior of existing declarations.

The canonical `Project` storage contains only `[Package.Dependency]`. The previous `Package` enum is a computed compatibility view, and its deprecated, disfavored project initializer converts legacy values at the boundary. Custom decoding accepts legacy payloads, while conflicting legacy and canonical representations are rejected.

The existing public `TargetDependency.package` and `XcodeGraph.Package` enum case shapes are unchanged. This preserves existing construction and pattern matching. Trait selections cross the graph through optional project metadata keyed by normalized package identity, then reconnect to Xcode package references during generation and import.

Because the public target dependency does not carry package identity, each package dependency hash conservatively includes every explicit project trait selection. The fingerprint uses sorted structured encoding to avoid stale cache entries, ordering sensitivity, and delimiter collisions. Projects without explicit traits retain their previous hash input.

## Impact

Users can configure Swift package traits inline for dependencies in Tuist's native Xcode package integration with an interface aligned with Swift Package Manager. Existing enum construction, pattern matching, serialized graph values, and generated projects remain compatible. Explicitly typed legacy package arrays receive a deprecation warning that points to `Package.Dependency`. Projects that select or disable traits require Xcode 26.4 or later.

## Validation

- `mise run cli:lint --fix`
- `mise x -- tuist generate tuist ProjectDescription TuistLoader TuistLoaderTests TuistGenerator TuistGeneratorTests TuistHasher TuistHasherTests XcodeGraph XcodeGraphTests XcodeGraphMapper XcodeGraphMapperTests TuistKit TuistKitTests --no-open`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing:TuistLoaderTests/PackageManifestMapperTests -only-testing:TuistHasherTests/DependenciesContentHasherPackageTraitsTests -only-testing:XcodeGraphTests/PackageTests -only-testing:TuistGeneratorTests/ProjectDescriptorGeneratorTests -only-testing:XcodeGraphMapperTests/PBXProjectMapperTests -only-testing:XcodeGraphMapperTests/XCPackageMapperTests -only-testing:TuistKitTests/DumpServiceProjectTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- The seven focused suites passed 56 tests.
- The native package traits fixture generated with the current binary, wrote `NativeIntegration` to the local package reference, and built successfully with the trait compiler condition enabled.

The compatibility coverage includes legacy enum representation, legacy manifest decoding, source-control and registry conversion, exact, branch, and revision factories, major and minor version spelling preservation, deterministic trait encoding, project import and generation, hash invalidation, collision resistance, and end-to-end native package generation.


cli/Sources/TuistGenerator/Generator/ProjectDescriptorGenerator.swift
```diff
@@ -313,6 +313,7 @@ struct ProjectDescriptorGenerator: ProjectDescriptorGenerating {
 
         Logger.current.debug("Processing \(project.packages.count) packages for project \(project.name)")
         for package in project.packages {
+            let traits = project.packageTraits?.first(where: { $0.package == package })?.traits
```
fortmarek: **[P1] Include package traits in consumer target hashes**

The native Swift package module itself is not cached, but Tuist-owned cacheable targets that consume native package products—and selective testing—still use the target content hash pipeline. `DependenciesContentHasher` currently reduces a `.package` dependency to its product and type, so switching traits can change the package API or implementation while leaving the consumer target hash unchanged. That can reuse a stale binary or incorrectly skip tests. Please fold a deterministic representation of the relevant trait selection into those hashes and add coverage where only the traits differ.

cli/Tests/TuistAutomationAcceptanceTests/BuildAcceptanceTests.swift
```diff
@@ -412,6 +412,14 @@ struct BuildAcceptanceTestSwiftPMPrebuiltMacro {
         )
         #expect(try await fileSystem.exists(prebuiltDirectory, isDirectory: true))
 
+        // Prebuilt SwiftSyntax libraries require a valid signed manifest. Package resolution falls back to sources
+        // when the manifest cannot be validated, leaving no prebuilt integration for this test to exercise.
+        let prebuiltModules = try await fileSystem.glob(
+            directory: prebuiltDirectory,
+            include: ["**/*-apple-macos.*"]
+        ).collect()
+        guard !prebuiltModules.isEmpty else { return }
```
fortmarek: **[P2] Do not silently pass when prebuilts are absent**

When this glob returns no modules, the test function returns normally, so Swift Testing reports success even though generation, the project assertions, and the build were never exercised. A regression in prebuilt resolution can therefore make this acceptance test stop covering the behavior without showing as a failure or skip. Please keep this scenario deterministic and require the modules, or surface the condition as an explicit skip and test fallback behavior separately.

fortmarek (commented):
I found two issues worth addressing before merging.

cli/Sources/ProjectDescription/PackageConfiguration.swift
```diff
@@ -0,0 +1,80 @@
+/// A Swift package dependency and its optional trait selection.
+///
+/// When `traits` is `nil`, the package uses its default traits. An empty array disables all default traits.
+public struct PackageConfiguration: Codable, Equatable, Sendable {
```
fortmarek: **[P1] Align the new dependency API with SwiftPM**

This adds a second public abstraction for package dependencies, but shapes it as a wrapper around Tuist’s legacy `Package` enum with raw `[String]?` traits. SwiftPM already models this concept as [`Package.Dependency`](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/Runtimes/PackageDescription/PackageDependency.swift), with a dependency kind and typed traits (including `.defaults`). Since this is new API, can we make the canonical Tuist type follow that model—for example, a `PackageDependency` value containing `kind` and `traits`—and keep the existing `Package` enum and `[Package]` initializer as deprecated conversion facades? That preserves existing enum pattern matching while avoiding another similar-but-different package API that will be harder to converge later.

fortmarek (commented):
One additional API-design finding after comparing the proposed public model with SwiftPM.

cli/Sources/TuistGenerator/Generator/ProjectDescriptorGenerator.swift
pepicrft: Fixed in 9ff17ddffe. `DependenciesContentHasher` now includes an order-independent representation of every explicit project package-trait selection when hashing a native package dependency. The representation includes the package identity and normalized trait names, so changing only traits invalidates consuming targets while reordering selections or trait names does not. `DependenciesContentHasherTests` covers both behaviors.

cli/Tests/TuistAutomationAcceptanceTests/BuildAcceptanceTests.swift
pepicrft: Fixed in 6199695317. The prebuilt-module guard now calls `Test.cancel(...)`, so an unavailable signed SwiftSyntax prebuilt is reported as an explicitly skipped test instead of a passing test. When modules are present, generation, project assertions, and the build continue unchanged. The focused acceptance test passed locally and exercised the full path.

cli/Sources/ProjectDescription/PackageConfiguration.swift
fortmarek: This thread was marked resolved, but the finding has not been addressed at the current head (`6d613c4`). `PackageConfiguration` is still a wrapper around the legacy `Package` enum with raw `[String]?` traits, and `Project` still splits it into parallel `packages` and `packageTraits` collections. The SwiftPM API-alignment concern therefore remains unresolved.

pepicrft: Addressed in `6184e59209` and `fb63f4b1dc`. I replaced `PackageConfiguration` with `Package.Dependency`, following Swift Package Manager's model with a typed dependency `kind`, `Set<Trait>`, and `.defaults`. The canonical `Project(packages:)` overload now accepts `[Package.Dependency]`, while the existing `Package` enum and `[Package]` overload remain as the deprecated, disfavored compatibility path.

The separate public `packageTraits` model is gone. Manifest mapping now consumes each `Package.Dependency` directly for both its package identity and traits, with the conversion to Xcode's string representation limited to the graph boundary.

I also added coverage for legacy enum pattern matching and serialization round trips, alongside the typed kind, trait defaults, graph mapping, and native package traits acceptance coverage. The focused loader suite passes all four tests, the related generation, hashing, and project import suites pass all 44 tests, and the native traits acceptance fixture generates and builds successfully.

pepicrft:
The bundle-size check is not an isolated regression from this pull request. Its 14.8 MB baseline is `main` commit `d5bee9b1a9`, last built on July 9. Since then, `main` has added 423 lines and removed 242 lines across command-line modules linked into the app, but those changes did not trigger the App workflow. This pull request is therefore the first app build to include those already-merged changes.

The fresh device build reproduced 18.1 MB, which is within the 17.2 MB to 18.9 MB range already recorded by recent `main` builds. The assets and resources are unchanged; the reported delta is entirely in the monolithic executable. The package-trait implementation itself remained within roughly 1 KB of install size across its major API revisions.

I consider the size alert safe to accept. Auto-merge is already enabled once the required approval and remaining checks complete.

cli/Sources/ProjectDescription/Package+Dependency.swift
```diff
@@ -0,0 +1,163 @@
...
+    /// A Swift package dependency and its enabled traits.
+    public struct Dependency: Codable, Equatable, Sendable {
+        /// The type of package dependency.
+        public enum Kind: Codable, Equatable, Sendable {
+            /// A dependency located at a local path.
+            case fileSystem(path: Path)
+            /// A dependency hosted in a source control repository.
+            case sourceControl(location: String, requirement: Package.Requirement)
+            /// A dependency published to a package registry.
+            case registry(id: String, requirement: Package.Requirement)
```
fortmarek: **[P1] Keep registry requirements type-safe like SwiftPM**

`Package.Requirement` also contains `.branch` and `.revision`, so this public case permits invalid combinations such as `.registry(id: "scope.package", requirement: .branch("main"))`; the current conversion path then carries that combination into the generated package reference instead of rejecting it. SwiftPM makes this state unrepresentable by using distinct [`SourceControlRequirement` and `RegistryRequirement`](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/Runtimes/PackageDescription/PackageDependency.swift) types, with registry requirements limited to versions and ranges. Please mirror that separation in the new canonical API and keep the all-purpose `Package.Requirement` only on the deprecated compatibility model.

fortmarek (commented):
One remaining API-model finding after re-reviewing the SwiftPM-aligned dependency changes.

.github/workflows/app.yml
```diff
@@ -6,6 +6,7 @@ on:
       - main
     paths:
       - "app/**"
+      - "cli/Sources/**"
```
fortmarek: why? I would strongly advise not to do this. We will run the app checks unnecessarily often.

cli/Sources/ProjectDescription/Project.swift
```diff
@@ -69,6 +69,8 @@ public struct Project: Codable, Equatable, Sendable {
     public let options: Options
     /// The Swift Packages used by the project.
     public let packages: [Package]
+    /// Swift package dependencies using the Swift Package Manager-aligned model.
+    public let packageDependencies: [Package.Dependency]?
     /// The targets of the project.
     public let targets: [Target]
```
fortmarek, line 98: **[P1] Keep a single package source of truth**

Exposing both `packages` and `packageDependencies` makes the same concept two public sources of truth. The constructors keep them synchronized, but `Project` is `Codable`, so a payload can contain conflicting collections, and the loader must decide which one wins; it also leaves users with two package APIs indefinitely. Please retain one canonical dependency collection and derive the compatibility representation from it. If strict read-side source compatibility is required, the canonical `[Package.Dependency]` storage can remain private/SPI for `TuistLoader`, while the existing public `packages: [Package]` becomes a deprecated computed view and custom `Codable` handles old payloads. Otherwise, make `packages: [Package.Dependency]` canonical and retain only the deprecated `[Package]` initializer. Either approach avoids adding a second public package property.

fortmarek (commented):
One remaining public API design finding around the package migration model.

cli/Sources/XcodeGraph/Sources/XcodeGraph/Models/Project.swift
```diff
@@ -53,6 +53,9 @@ public struct Project: Hashable, Equatable, CustomStringConvertible, CustomDebug
     /// Project swift packages.
     public var packages: [Package]
 
+    /// Trait selections for native packages. A missing package uses its default traits.
+    public var packageTraits: [PackageTraitSelection]?
```
fortmarek: **[P1] Keep trait selections attached to their package**

`packages` and `packageTraits` are independently mutable, which makes invalid states representable. The generator reconnects them through an equality lookup and treats a missing match as the package defaults, so any graph transformation or consumer that updates one collection without the other can silently generate a project with the wrong traits. SwiftPM keeps selected traits on the dependency value and the final enabled traits on the resolved package. Please preserve that association in XcodeGraph as well—for example, make `Project.packages` contain a package dependency/reference value that owns both the package kind and its traits—so there remains a single source of truth.

cli/Sources/TuistHasher/DependenciesContentHasher.swift
```diff
@@ -163,9 +163,14 @@ public struct DependenciesContentHasher: DependenciesContentHashing {
                 )
             }
         case let .package(product, type, _):
+            let packageTraits = try packageTraitsFingerprint(graphTarget.project.packageTraits)
```
fortmarek: **[P2] Hash only the traits of this package dependency**

`packageTraitsFingerprint(graphTarget.project.packageTraits)` includes every package trait selection in the project for each `.package` dependency. Changing traits for package A therefore also changes the dependency hash of a target that only consumes package B, invalidating unrelated cache entries. Please carry enough package identity with the dependency (or resolve the product back to its package) to hash only its effective trait selection, and add a regression test showing that changing an unrelated package's traits leaves this target hash unchanged.

fortmarek (commented):
Two follow-up findings on keeping package trait state associated and scoping its cache impact.

cli/Sources/XcodeGraph/Sources/XcodeGraph/Models/Project.swift
pepicrft: Fixed in `58c6ca4f93`. `XcodeGraph.Package` now owns its optional trait selection directly, so `Project.packages` is the only graph-level source of truth. Manifest mapping, Xcode project import, generation, graph traversal, and hashing now preserve that association without a parallel lookup. Packages without traits retain their previous encoded representation, and the focused loader, generator, Xcode project mapper, and package compatibility tests pass.

cli/Sources/TuistHasher/DependenciesContentHasher.swift
pepicrft: Fixed in `58c6ca4f93`. The additive `.package(product:package:)` target dependency overload carries the package identity into `XcodeGraph`. `DependenciesContentHasher` now fingerprints only the matching package’s normalized trait selection. Existing dependencies without an identity retain the conservative all-package fallback. `DependenciesContentHasherPackageTraitsTests` verifies that related trait changes invalidate the consumer, ordering does not matter, unrelated package trait changes leave the consumer hash unchanged, and the legacy fallback still invalidates.

cli/Sources/ProjectDescription/Package+Dependency.swift
pepicrft: Fixed in `c3333d8701`. `Package.Dependency.Kind` now uses separate `SourceControlRequirement` and `RegistryRequirement` types. Registry requirements contain only exact versions and version ranges, while branch and revision requirements exist only for source control dependencies. `PackageManifestMapperTests.packageDependencyUsesDistinctSourceControlAndRegistryRequirements` verifies both conversion paths.

.github/workflows/app.yml
pepicrft: Fixed in `171370b068`. I removed `cli/Sources/**` from the application workflow push and pull request filters. Command-line source changes continue to run the command-line workflow, while the application workflow is limited to application and project-definition inputs, avoiding full application checks for unrelated command-line changes.

cli/Sources/ProjectDescription/Project.swift
pepicrft: Fixed in `4e00f9e8ba`. `Project` now stores exactly one private `PackageStorage` value. The existing `packages` property is a computed compatibility view, while trait-aware dependencies are visible only to `TuistLoader` through a restricted import boundary. Custom `Codable` preserves legacy payloads and rejects payloads whose package representations conflict. `PackageManifestMapperTests` covers legacy round trips, trait-aware round trips, mapping, and conflict rejection, and the existing project dump test passes.

cli/Sources/ProjectDescription/Project.swift
```diff
@@ -59,6 +59,26 @@
 /// )
 /// ```
 public struct Project: Codable, Equatable, Sendable {
+    private enum PackageStorage: Equatable, Sendable {
```
fortmarek: **[P2] Keep dependencies as the single canonical package representation**

`PackageStorage` preserves which initializer was used, leaving `Project` with two internal representations and construction-history-dependent encoding. The deprecated initializer can instead map its `[Package]` values to `[Package.Dependency]` with default traits, while `packages` remains a computed legacy projection. We should keep `[Package.Dependency]` as the sole storage and preserve the legacy API only at the initializer/property boundary.

cli/Sources/ProjectDescription/TargetDependency.swift
```diff
@@ -114,6 +114,14 @@ public enum TargetDependency: Codable, Hashable, Sendable {
     ///   - condition: condition under which to use this dependency, `nil` if this should always be used
     case package(product: String, type: PackageType = .runtime, condition: PlatformCondition? = nil)
 
+    /// The representation used by `package(product:package:type:condition:)`.
+    case packageWithIdentity(
```
fortmarek: **[P2] Represent package identity on the existing package case**

`packageWithIdentity` is not a distinct dependency kind; it is the existing package dependency with additional ownership information. Introducing another case forces every downstream switch to duplicate `.package` behavior and leaks a compatibility workaround into the public model. Please add an optional package identity to the existing `.package` case and use `nil` for the deprecated product-only form. Although changing its associated values can affect callers that pattern-match it, that usage should be uncommon, and adding this new public case already affects exhaustive switches.

fortmarek (commented):
Two follow-up API-modeling findings.

cli/Sources/ProjectDescription/Project.swift
pepicrft: Agreed. I checked the current Swift Package Manager implementation and changed this in `801b60c5be`. `Project` now stores only `[Package.Dependency]`; the deprecated `[Package]` initializer maps each value into that representation with default traits, and `packages` is a computed legacy projection. Encoding no longer depends on which initializer was used, while decoding still accepts payloads that contain only the legacy `packages` field. I added coverage for local, source-control, and registry conversion plus legacy payload decoding.

cli/Sources/ProjectDescription/TargetDependency.swift
pepicrft: Agreed. The current Swift Package Manager model also uses one product-dependency case with an optional package identity internally. In `801b60c5be`, I removed `packageWithIdentity` and added `package: String? = nil` to the existing `.package` case across the manifest and graph models. Existing construction syntax continues to compile, and a compatibility test confirms the previous encoded `.package(product:type:condition:)` value decodes with `package == nil`. The only source-level caveat is callers pattern-matching the associated values, which matches the tradeoff described here and avoids a second public enum case.

cli/Sources/ProjectDescription/Project.swift
pepicrft: Addressed in `99a8837e59`. `Project` now stores only `[Package.Dependency]`. The legacy `packages` property is a computed projection, and the deprecated `[Package]` initializer converts at the compatibility boundary. Custom decoding accepts legacy-only payloads and rejects conflicting legacy and canonical representations. `PackageManifestMapperTests` covers legacy decoding, source-control and registry conversion, round trips, and conflict rejection.

cli/Sources/ProjectDescription/TargetDependency.swift
pepicrft: I changed course after compiling the existing pattern-matching form. Adding a defaulted fourth associated value still makes `case let .package(product, type, condition)` fail to compile, so the optional identity is source-breaking. In `99a8837e59`, I restored the original public package cases in both the project description and graph models. Explicit trait selections now cross the graph as optional project metadata, and `46638d6a0c` conservatively includes every explicit selection in a structured dependency fingerprint. This preserves existing construction, pattern matching, and encoded package values while still preventing stale consumer hashes. The focused model, mapping, generation, and hashing suites cover those compatibility paths.

PR #11805, opened 2026-07-13
feat(cli): enable SwifterPM resolution by default

pepicrft:
### What changed

The Tuist [command-line interface](https://en.wikipedia.org/wiki/Command-line_interface) now uses [SwifterPM](https://github.com/tuist/swifterpm) by default for package resolution and updates. Setting `TUIST_USE_SWIFTERPM=0` keeps using Swift Package Manager, providing an explicit compatibility escape hatch.

The development environment no longer sets `TUIST_USE_SWIFTERPM=1`, since the opt-in is redundant once SwifterPM is the default.

### Why

SwifterPM resolution was introduced behind an opt-in while the integration matured. Making it the default gives every Tuist user the accelerated resolution path without requiring environment configuration, while retaining the native Swift Package Manager path for projects that need it.

### Approach

The existing environment variable remains the source of truth to avoid introducing a second configuration mechanism. An unset value now enables SwifterPM, an explicit truthy value continues to enable it, and an explicit false-like value disables it.

The controller tests use deterministic environment dictionaries and cover the new unset default plus the resolution and update opt-out paths.

### User and developer impact

Users without `TUIST_USE_SWIFTERPM` configured will start resolving packages with SwifterPM. Existing `TUIST_USE_SWIFTERPM=1` configurations remain valid, and `TUIST_USE_SWIFTERPM=0` selects Swift Package Manager without any other migration.

### How to test locally

Generate the focused workspace and run the controller tests:

```sh
tuist generate tuist TuistSupport TuistSupportTests ProjectDescription --no-open
  -only-testing TuistSupportTests/SwiftPackageManagerControllerTests \
  CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" \
  COMPILATION_CACHE_ENABLE_CACHING=NO
```

The suite executes 14 tests with no failures. Formatting and linting also pass:

```sh
mise run cli:lint --fix
```

### DIFF
--- a/cli/Sources/TuistSupport/SwiftPackageManager/SwiftPackageManagerController.swift
+++ b/cli/Sources/TuistSupport/SwiftPackageManager/SwiftPackageManagerController.swift
@@ -274,7 +274,9 @@ public struct SwiftPackageManagerController: SwiftPackageManagerControlling {
     }
 
     private var swifterPMIsEnabled: Bool {
-        Environment(variables: environmentVariables()).isVariableTruthy(Constants.EnvironmentVariables.useSwifterPM)
+        let variables = environmentVariables()
+        guard variables[Constants.EnvironmentVariables.useSwifterPM] != nil else { return true }
+        return Environment(variables: variables).isVariableTruthy(Constants.EnvironmentVariables.useSwifterPM)
     }
 
     private func swifterPMRequest(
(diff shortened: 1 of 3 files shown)


pepicrft:
Closing this for now. Before we make SwifterPM resolution the default, we want to validate that it works reliably across many projects. Please try the opt-in and share feedback in the community thread: https://community.tuist.dev/t/better-tuist-install-package-resolution-when-working-with-git-worktrees/1001/1

PR #11816, opened 2026-07-13
fix(kura,cli): serve action-cache snapshots from a dedicated index instead of scanning the namespace

fortmarek:
> [!NOTE]
> Stacked on #11794; GitHub will retarget to `main` when that merges. Draft until the large-namespace scale test finishes locally and the fix is validated against a production-scale namespace end to end.

## Problem

Every action-cache snapshot fetch for tuist/tuist timed out, on every attempt, indefinitely. A local two-rep bench of the view-heal flow (#11787) showed `snapshot_hits=0` across 10,629 resolves with the endpoint itself healthy (`/up` in 160ms), making warm remote builds RTT-bound on the per-key ladder — the exact traffic the snapshot exists to remove. Rep 2 measured 836s against a 411s no-cache baseline.

## Root cause

Three compounding server-side issues in the snapshot serve path:

1. **The reconcile's enumeration was O(namespace), not O(action cache).** `action_cache_manifests` walked the namespace's entire artifact index with a manifest point-read per row just to filter out non-action-cache entries. Blobs outnumber action-cache entries roughly a thousand to one in real namespaces, so a cold build did millions of sequential RocksDB point reads — tens of minutes. The 100k entry cap bounded memory, not the scan.
2. **The first request awaited that build inline.** Every client fetch ran into its own deadline (the plugin's 180s) while the build churned, and since the reconcile only logs on completion, the server looked silent the whole time. The stale-kick design re-paid the full scan every 60 seconds once warm.
3. **The client treated the resulting timeout as permanent.** A failed fetch parked the snapshot in `Absent` for the 1-hour retry interval that was designed for "server has no snapshot support", so a single slow build disabled the snapshot for entire build sessions.

Also found while reading the path: a reconcile panic leaked its `builds`-map entry, permanently bricking snapshots for that namespace until pod restart (every later request resolved to the dead shared future).

## Change

**kura**
- New `action_cache_index` column family, ordered newest-first by write time (the row key embeds `!version_ms` so a forward prefix scan needs no sort and stops at the entry cap). Maintained at publish (segment and inline paths), delete, expiry sweep, and namespace delete. Namespaces written before the CF existed backfill lazily: one legacy scan whose rows and completion marker commit in a single batch, so a crash mid-backfill just redoes it. Rows whose manifest is gone or has moved to a different version self-heal on scan.
- `action_cache_manifests` reads that index: a reconcile now touches at most the snapshot's entry cap (100k) of rows plus their manifest point-reads, instead of the whole namespace.
- Cold serves (no cached index) wait at most 15s for the build, then answer `UNAVAILABLE` while it continues in the background; the next fetch serves from the completed index. Already-indexed and small namespaces complete inline and keep their one-round-trip semantics — the runner flow validated in the snapshot bench is unaffected.
- Build tasks are panic-guarded, with the builds-map cleanup outside the guard so it always runs.
- Reconciles log build start plus `scan_ms`/`load_ms`/`gate_ms` phase timings (the Loki breadcrumb this incident lacked), and the presence gate yields every 1024 checks instead of parking a runtime worker for the duration of a cold gate.

**cas-plugin**
- `Absent` now carries a `retry_after`: fetch *errors* retry after 60s (the server's `UNAVAILABLE` while building, timeouts, transport errors — all transient by nature), while a definitive not-found keeps the 1-hour interval. No status-string sniffing; the error/none distinction already encodes it.

## Why this shape

- A dedicated index over "return `NOT_FOUND` immediately on cold serves": deployed clients retry `Absent` after an hour, so a bare fast-fail would bench the snapshot for entire builds on today's fleet. The bounded 15s wait keeps the existing contract for every namespace whose build is actually fast, which after backfill is all of them.
- `UNAVAILABLE` over `NOT_FOUND` for "still building": `NOT_FOUND` is the existing wire contract for "server has no snapshot support" and must keep meaning that for mixed-version rollouts.
- Backfill marker + atomic batch over gradual population: the index must be complete before capped scans are correct; entries not in the index would be invisible to snapshots forever.

Rollout-safe in both directions: old clients treat `UNAVAILABLE` like any error (their behavior today), old servers still answer not-found, new CF creation is additive at DB open, and no wire or on-disk formats change.

## Validation

- All 341 kura lib tests pass, including the pre-existing scan-semantics tests (newest-first cap, mid-scan shed) now running against the indexed implementation — the first call in those tests exercises the backfill path, the second the indexed path.
- New store tests: index rows maintained across overwrite (old version row replaced), expiry deletes, and writes landing after backfill.
- New paused-time serve test: a cold serve during a stalled build answers `UNAVAILABLE` instead of hanging, and the same build completes and serves after the stall clears.
- `cargo clippy --all-targets` clean on kura; zero new findings on cas-plugin (48 pre-existing, unchanged).
- The `--ignored` large-namespace scale test passes against the indexed path (build boundedness held; ~25 min wall, dominated by populating the fixture namespace).
- Pending before undraft: an end-to-end rerun of the view-heal bench against a deployment of this branch to confirm `snapshot_hits` recover on tuist/tuist.


fortmarek:
End-to-end validation against production (kura@0.16.4, tuist-eu-central-1) is complete — this closes the gate the draft was held on.

**Snapshot serves at tuist/tuist scale.** First fetch after the rollout landed a 48MiB snapshot (38,779 keys / 350,236 nodes) about a minute after the pods became ready — the payload sits exactly at the recency-window ceiling, shedding oldest keys as designed. The one-time backfill plus index build completed in under a minute where the old full-namespace scan ran for tens of minutes and timed out every fetch. During the rollout window the client rode the LB 503s on the new 60s error-retry and recovered without intervention.

**Two-rep view-heal bench (dev machine over WAN, cold DerivedData per rep, tuist scheme):**

| scenario | wall | resolve mix |
|---|---|---|
| no-cache baseline | 411s | — |
| before this PR (snapshot timing out) | 836s | remote_hits=10613, snapshot_hits=0, per-key ladder |
| after (kura 0.16.4) | **336s** | **snapshot_hits=7525 (71%)**, remote_hits=2999, **misses=0** |

Remote Xcode CAS over WAN goes from 2x net-negative to 18% below the no-cache baseline on this workload. Rep 1 (the healing populate) also dropped 762s → 433s with a 66% snapshot-hit mix, and the view-refresh queue drained to zero in 7 ticks. Zero CAS errors in either rep.

PR #11820, opened 2026-07-14
perf(kura,cli): compress the snapshot, coalesce demand fetches, and serve stale during rebuilds

fortmarek:
> [!NOTE]
> Draft pending clean perf validation. The demand-coalescer win is client-side and measurable against production; the compression win needs a branch-kura deploy (production is kura@0.16.4, which lacks the server side), same gate #11816 used. All three changes are tested and the client's old-server compat path is verified live.

## Context

After #11816 landed snapshots at tuist/tuist scale, a cold warm-remote Xcode-CAS build over the WAN measured **336s against an 87s local-CAS floor** — a real win (below the 411s no-cache baseline) but still ~4x the floor. Attribution of the remaining gap, from the proxy's per-interval counters on a cold rep:

- **~2,200 thread-seconds on the per-key ladder** for keys shed from the snapshot's 48 MiB recency window. tuist/tuist is written by all of CI; the wire body encodes newest-first, so the tail of older-but-still-live keys a given build needs falls out and each one costs a WAN round trip.
- **~5,700 thread-seconds fetching ~7,000 objects on demand**, one `BatchReadBlobs` round trip per object. When a compiler worker outruns the background materializer it self-heals through `OP_FETCH_OBJECT`, and those reads were issued one at a time.

And a third issue surfaced while validating: on a namespace this hot, a *cold* client got repeated `UNAVAILABLE` because the snapshot reconcile kept running (and exceeding the 15s cold-serve wait). This PR attacks all three.

## 1. Snapshot compression (kura + cli)

The server always emits the body zstd-wrapped in a `TSNZ` envelope. It includes keys up to a larger content budget (144 MiB uncompressed) and enforces the 48 MiB ceiling on the **compressed** wire size; a body that compresses worse than budgeted is re-encoded with a smaller budget scaled from the observed ratio, bounded to a few tries that can only converge on a provably-safe minimum, so it always terminates. The same wire budget now carries ~3x the keys, so the shed tail — and the per-key ladder it caused — shrinks toward nothing.

**No negotiation.** The snapshot protocol has only shipped in canary, so no stable client needs the uncompressed body, and an old client hitting a new server just returns `None` from decode and degrades to the per-key path — it never breaks. The one compat that matters is the reverse: kura is a rolling mesh (0.16.4 pods serve the old `TSNP` body mid-roll) and self-hosted nodes lag, so the **client decodes both** `TSNZ` and bare `TSNP` (with an allocation guard on the declared length). Verified live — this branch's client against production kura fetched and decoded the uncompressed snapshot.

## 2. Demand-fetch coalescing (cli)

Concurrent `OP_FETCH_OBJECT` blob reads for one instance now join a shared `BatchReadBlobs` through a leader/follower batcher. A leader drains everything pending into one batch while followers accumulate during its round trip, so the batch size self-clocks to the worker concurrency with no fixed linger on the steady-state path — the ~7,000 single reads collapse into far fewer batches. Each result is a one-shot mailbox, so two workers demanding the same digest at once may each fetch it (a rare redundant read) but never see a wrong or dropped result.

## 3. Serve the cached view during a reconcile (kura)

The serve fast path returns a cached (slightly-stale) view immediately and kicks a background reconcile only when the view ages past the freshness window — but the reconcile **removed** the index from the cache for its whole duration. Any serve that landed during a reconcile found nothing, fell to the cold path, and answered `UNAVAILABLE`; on a hot namespace where reconciles run back-to-back that window is a large fraction of the time. Now the reconcile builds against a **clone** and leaves the live index in place (swapping the result in on success, leaving it untouched on failure), so serves hit the stale-but-present view throughout. A fresh namespace's first-ever build still takes the cold path once. Cost: one extra index-sized copy per reconcile, bounded by the 100k-entry cap.

## Validation

- kura: full suite (343) + clippy green; tests round-trip the `TSNZ` envelope, and assert a stale serve returns the cached view (not `UNAVAILABLE`) while a reconcile is outstanding.
- cli: full plugin suite (32) green, no new clippy findings; tests decode both snapshot forms (rejecting torn/oversized payloads), and drive the coalescer under 8 concurrent threads (every hash fetched exactly once) plus missing/failure routing.
- Live compat: this branch's client fetched production's uncompressed `TSNP` and decoded it fine.
- Preliminary coalescer signal (against production, snapshot available): the populate leg went **433s → 382s** with ~7,480 demand fetches coalesced. A clean cold-slate rep and the compression win both need a branch-kura deploy for an uncontaminated number, tracked as the undraft gate.

### DIFF
--- a/cas-plugin/src/proxy.rs
+++ b/cas-plugin/src/proxy.rs
@@ -51,6 +51,17 @@ const SNAPSHOT_DELTA_INTERVAL: Duration = Duration::from_secs(120);
 const SNAPSHOT_FETCH_WAIT: Duration = Duration::from_secs(20);
 const SNAPSHOT_FULL_INTERVAL: Duration = Duration::from_secs(30 * 60);
 const SNAPSHOT_RETRY_INTERVAL: Duration = Duration::from_secs(60 * 60);
+// Ceiling on a compressed snapshot's declared uncompressed size, so a torn or
+// hostile length prefix cannot drive an unbounded allocation. Comfortably
+// above the server's content budget (144 MiB); a legitimate body never
+// approaches it.
+const SNAPSHOT_DECOMPRESS_MAX_BYTES: usize = 512 << 20;
+// A leader that finds itself alone lingers this long before firing, giving a
+// near-simultaneous straggler time to join its batch. Once batches flow the
+// followers accumulating during each round trip carry the coalescing, so this
+// only pays out at the very start and end of a demand burst; kept far below a
+// WAN round trip so a lone fetch is barely delayed.
+const DEMAND_BATCH_LINGER: Duration = Duration::from_millis(3);
 // Failed fetches retry much sooner than definitive not-found: the server
 // answers UNAVAILABLE while it builds a large namespace's snapshot index,
 // and timeouts/transport errors are transient by the same token.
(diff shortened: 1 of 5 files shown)


fortmarek:
Clean combined validation against production is in — this closes the undraft/validation gate.

kura 0.16.5 (this PR's server side) is serving tuist/tuist on the eu-central pods, so both halves — compression AND coalescing — were measured together against a real populated namespace for the first time.

**Compression, proven at the protocol level:** the snapshot now carries **72,243 keys in a 47.9 MiB compressed wire**, versus ~38,779 keys in a 50.3 MiB uncompressed wire on 0.16.4 — ~1.9x the key coverage in the same wire budget, exactly as designed (the ceiling moved to the compressed size). Snapshot hit rate rose 71% → 75% and the per-key ladder shrank (remote_hits 2999 → 2664).

**Two-rep view-heal bench (dev machine over WAN, cold DerivedData, tuist scheme):**

| build | wall | vs 87s local-CAS floor |
|---|---|---|
| no cache | 411s | 4.7x |
| #11816 (snapshots only, kura 0.16.4) | 336s | 3.9x |
| **this PR (compression + coalescing, kura 0.16.5)** | **282s** | **3.2x** |

Cold-slate rep 2: 282s, **snapshot_hits=7995 / 10,674 resolves (75%), zero misses, zero CAS errors**. The wall dropped 336s → 282s even though demand-fetch volume was similar (~7,600) — the coalescer batched those reads into far fewer round trips, and compression cut the shed-key ladder. Rep 1 (heal populate) settled at 40s on its incremental pass with 7,989 snapshot hits.

The remaining 282s-vs-87s gap is the WAN transfer of the working-set closure the build actually needs — bandwidth-bound now rather than round-trip-bound, so eager closure prematerialization (pulling the snapshot's referenced graphs at high parallelism during the scheduling-bound early phase) is the next lever, bounded ultimately by the home-link envelope.

PR #11825, opened 2026-07-14
fix(cli): use native Xcode macro dependencies

pepicrft:
## What changed

Tuist now models source-built Swift macros through Xcode's native macro dependency mechanism.

- Macro implementation targets declare Xcode's host-build and macro implementation settings.
- Source macro consumers rely on target dependencies instead of generated compiler flags.
- Precompiled and cached macros use `SWIFT_LOAD_BINARY_MACROS`.
- Cross-project source macro dependencies receive explicit target dependency edges.
- The macro executable copy build phase has been removed.

## Why

The previous workaround copied a macro executable into a hard-coded Debug output path and pointed every consumer at that path. Release builds therefore depended on the Debug layout, multiple targets could write the same output, and script sandboxing and incremental build behavior were difficult to reason about.

## Root cause

Tuist treated source-built macros like detached binary plugins. Xcode's build system already understands source macro implementations when the producer target advertises `SWIFT_IMPLEMENTS_MACROS_FOR_MODULE_NAMES` and is reachable through the target dependency graph. Because Tuist did not emit that metadata, it compensated with a shell copy and `-load-plugin-executable`.

## Approach

The implementation follows the behavior in Apple's [Swift Build](https://github.com/swiftlang/swift-build) and [Swift Package Manager](https://github.com/swiftlang/swift-package-manager):

- Source macro targets advertise their module through native build settings and build for `$(HOST_PLATFORM)`.
- Xcode derives the executable path from the dependency target and propagates the macro implementation transitively.
- Binary-only macros remain explicit through `SWIFT_LOAD_BINARY_MACROS`.

Tuist continues to generate macro targets as command-line-tool products. A dedicated host-build-tool product was evaluated, but it made the generated macro test scheme ineligible for the local macOS destination with the current project model. Keeping the existing product type limits this change to dependency semantics.

## Impact

Source macros no longer need a shell phase, a Debug-specific path, or direct compiler flags. Debug simulator builds, Release device archives, cross-project macro propagation, script sandboxing, the documented macro test wrapper, and cached macro consumption all work locally. Cached macros use Xcode's native binary macro setting.

## Validation

- Generated the focused Tuist workspace with `tuist generate tuist TuistCore TuistCoreTests TuistGenerator TuistGeneratorTests ProjectDescription --no-open`.
- Passed 158 `GraphTraverserTests` and 99 focused generator tests with `xcodebuild test`.
- Built a three-project cross-project macro fixture for an iOS simulator in Debug with user script sandboxing enabled.
- Archived the same fixture for a generic iOS device in Release.
- Built this branch with the cache extension enabled and warmed fresh Debug and Release `MacroHost` artifacts with zero local and remote hits.
- Generated a focused `Feature` workspace where the source target had no build dependencies and `MacroHost` was available only as a cached universal executable.
- Built the cached fixture for both simulator architectures in Debug and for a generic iOS device in Release, then archived the Release framework.
- Confirmed Xcode translated `SWIFT_LOAD_BINARY_MACROS` into the configuration-specific cached executable path and that the produced binaries contained the expected `cached-native: 42` expansion.
- Passed the documented `MyMacrosTests` macOS macro test workflow.
- Ran `mise run cli:lint --fix` and `git diff --check`.

### DIFF
--- a/cli/Sources/TuistCore/Graph/GraphTraverser.swift
+++ b/cli/Sources/TuistCore/Graph/GraphTraverser.swift
@@ -27,7 +27,7 @@ public class GraphTraverser: GraphTraversing {
     private let graph: Graph
     private let conditionCache = ConditionCache()
     private let conditionalTargets: Set<GraphDependency>
-    private let swiftPluginExecutablesCache = GraphCache<GraphDependency, Set<String>>()
+    private let precompiledSwiftMacroExecutablesCache = GraphCache<GraphDependency, Set<String>>()
     private let systemFrameworkMetadataProvider: SystemFrameworkMetadataProviding =
         SystemFrameworkMetadataProvider()
     private let targetDirectTargetDependenciesCache: ThreadSafe<[GraphTarget: [GraphTarget]]> =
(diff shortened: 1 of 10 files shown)


fortmarek (approved):
as long as the acceptance tests pass, I'm good

PR #11826, opened 2026-07-14
feat(registry): route production traffic to the standalone frontend

pepicrft:
## What changed

The production registry ingress now keeps `registry.tuist.dev` as its routing and certificate identity. The domain-name controller publishes `registry-origin.tuist.dev` only as a resolution alias, and the registry router uses Cloudflare's origin-resolution override to reach that address while preserving the canonical request hostname.

The standalone registry frontend is always attempted first. If its health check fails, its request fails, or it returns a server error, the worker falls back to the nearest healthy regional cache origin. Canonical `/swift` requests are rewritten to the compatibility path only when a cache fallback serves them.

The server continues advertising `https://registry.tuist.dev/api/registry/swift` during the cutover because both the cache origins and standalone frontend serve it. Infrastructure and registry documentation now describe the resolution alias, fallback behavior, and direct-ingress smoke test.

## Why

The standalone registry reader and server-owned sync writer are already running in production, but public requests still terminate on regional cache nodes. This change starts serving through the standalone frontend without turning its single production deployment into a new availability bottleneck or exposing the resolution alias as a second application identity.

## Root cause

`registry.tuist.dev` is bound to the Cloudflare Worker route, so the worker needs a separate domain-name record to resolve the Kubernetes ingress without recursively invoking its public route. Using that alias as the request hostname would also make it a directly usable registry endpoint.

The initial cutover also removed all nine regional cache origins. Because the production standalone frontend currently has one replica in one cluster, an origin or cluster incident would have required a manual worker rollback instead of failing over automatically.

## Approach

The ingress rule and certificate stay on `registry.tuist.dev`. An annotation-only domain-name record maps `registry-origin.tuist.dev` to the ingress address without changing the ingress application host. The worker explicitly keeps same-zone fetches on the private origin path and uses the resolution override, so the request hostname and certificate server name remain canonical. A normal request to the alias does not match the ingress rule or certificate identity.

The worker retains the standalone frontend as the primary origin and restores the previous geographically ordered cache list as fallback. Health checks cover the primary resolution alias and every cache node. Cache fallback rewrites `/swift` to `/api/registry/swift`, the route the cache service supports, while leaving legacy requests unchanged.

## Impact

Production traffic moves to the standalone registry frontend while retaining automatic regional fallback throughout the soak period. Existing and newly configured clients remain compatible, and the resolution alias is not a separate registry application hostname.

The safe rollout order remains chart first and worker second. If the alias is not ready when the worker deploys, the primary health check fails and cache fallback continues serving traffic. Reverting the worker restores the previous cache-only routing independently of the registry deployment.

## Validation

- The TypeScript compiler check passed for `infra/registry-router`.
- A Wrangler dry deployment accepted the worker bundle, compatibility flag, and bindings.
- Focused worker behavior assertions confirmed canonical-host origin resolution, canonical-path rewriting for cache fallback, and fallback after a primary server error.
- The repository's standard Helm chart lint passed.
- The production Helm render confirmed that the ingress rule and certificate use `registry.tuist.dev`, while only `registry-origin.tuist.dev` is published by the domain-name annotation.
- Git whitespace validation passed.

### How to test locally

```sh
cd infra/registry-router
npm ci
npm exec tsc -- --noEmit
npx --yes wrangler@latest deploy --dry-run
cd ../..
helm lint infra/helm/tuist -f infra/helm/tuist/values-ci.yaml
  -f infra/helm/tuist/values-ci.yaml
```


infra/helm/tuist/values-managed-production.yaml
```diff
@@ -76,15 +74,11 @@ server:
...
-    # registry.tuist.dev is still fronted by the registry-router Worker
-    # against the cache origins, which only serve that prefix. Flip to
-    # https://registry.tuist.dev/swift once the Worker cuts over to the
-    # standalone frontend (which serves /swift).
+    # Full registry address advertised at /.well-known/registry.json so the
+    # Tuist command-line interface configures Swift Package Manager against
+    # the standalone registry frontend.
     - name: TUIST_REGISTRY_URL
-      value: https://registry.tuist.dev/api/registry/swift
+      value: https://registry.tuist.dev/swift
```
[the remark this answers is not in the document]
pepicrft: Fixed in c28174f218. `infra/helm/tuist/values-managed-production.yaml` now keeps the advertised address at `https://registry.tuist.dev/api/registry/swift` until the Worker cutover is live. That prefix is mounted by both `cache/lib/cache_web/router.ex` and the standalone `registry/lib/tuist_registry_web/router.ex`, so it works during the chart-first phase and after the Worker moves to `registry-origin.tuist.dev`.

I also rendered the production Helm values and confirmed the ingress remains `registry-origin.tuist.dev` while the advertised registry address remains on the compatibility path.

infra/helm/tuist/values-managed-production.yaml
```diff
@@ -7,19 +7,17 @@
...
+  # without recursing through its registry.tuist.dev route.
   enabled: true
   publicHost: registry.tuist.dev
   serverUrl: https://tuist.dev
   deployEnv: prod
   sentryEnv: prod
   poolSize: "10"
   ingress:
-    host: registry.tuist.dev
+    host: registry-origin.tuist.dev
```
fortmarek: **[P2] Avoid making the Worker’s upstream a second public registry hostname**

Changing the ingress host to `registry-origin.tuist.dev` creates a publicly resolvable, TLS-enabled registry endpoint solely because the Worker cannot fetch `registry.tuist.dev` without matching its own route again. This makes the routing workaround part of the production architecture: requests can reach the registry through an undocumented hostname that bypasses the Worker’s health decision and may not receive the same Cloudflare rules, observability, or protections as the canonical host. It also adds another DNS record and certificate that must be provisioned and preserved during deployments and rollbacks.

The origin should not need its own public application identity. Prefer routing `registry.tuist.dev` directly to the ingress and using a Cloudflare Origin Rule for the legacy `tuist.dev/api/registry/*` path. If the Worker must remain, use Cloudflare origin-resolution controls so requests retain `registry.tuist.dev` as their Host/SNI and treat any required resolution alias as infrastructure-only. Otherwise, this PR should explain why a publicly reachable bypass is required and how access and Cloudflare policy parity are enforced.

fortmarek (commented):
Flagging an architectural concern with the new public origin hostname.

infra/registry-router/src/index.ts
```diff
@@ -82,36 +37,29 @@ async function proxyToOrigin(request: Request, host: string): Promise<Response |
...
 async function handleRequest(request: Request, env: Env): Promise<Response> {
-  const coords = requestCoordinates(request);
-  const origins = coords ? sortedByDistance(coords.lat, coords.lon) : ORIGINS;
-
-  for (const { host } of origins) {
-    if (!(await isOriginHealthy(host, env))) continue;
-
-    const response = await proxyToOrigin(request, host);
-    if (response) return response;
+  if (!(await isOriginHealthy(ORIGIN_HOST, env))) {
```
fortmarek: **[P1] Retain automatic fallback while the standalone frontend is a single failure domain**

This replaces the previous loop over nine independently health-checked cache origins with a single origin and returns `502` whenever that origin is marked unhealthy or its request fails. Production currently inherits `registry.replicaCount: 1` and has no registry PodDisruptionBudget, so a registry pod or node failure—and any wider ingress or cluster incident—now makes both `registry.tuist.dev/*` and the legacy `tuist.dev/api/registry/*` entry point unavailable. The scheduled health check detects that outage, but with no alternate origin it cannot mitigate it; it only changes the failure into an immediate Worker-generated 502.

Keep the cache origins as fallback during a defined soak/rollback period, or make the standalone frontend highly available across failure domains and provide another automatic fallback before sending all production traffic to it. Otherwise this cutover materially reduces availability from multi-region failover to a single Kubernetes origin that requires manual Worker rollback during an incident.

fortmarek (commented):
Flagging the remaining availability regression in the registry cutover.

infra/helm/tuist/values-managed-production.yaml
pepicrft: Fixed in 798c4a1cfa. The production ingress rule and certificate host are back on `registry.tuist.dev`. `registry-origin.tuist.dev` is now created only by the annotation-only domain-name record and is passed to Cloudflare's [`resolveOverride`](https://developers.cloudflare.com/workers/runtime-apis/request/#requestinitcfproperties) setting, while the request address and `Host` header remain canonical. `global_fetch_private_origin` is explicit in `wrangler.toml` so that same-zone fetches reach the origin instead of invoking the public Worker route again.

A normal request to the alias no longer matches the ingress rule or certificate identity, so it is not a second registry application endpoint. The Wrangler dry deployment passed, and the production Helm render confirmed canonical routing and certificate hosts with only the resolution annotation using the alias.

infra/registry-router/src/index.ts
pepicrft: Fixed in 798c4a1cfa. The standalone frontend remains the primary origin, and all nine geographically ordered cache origins are restored as automatic fallback during the soak period. The scheduled health check now covers the standalone resolution alias and every cache node; request failures and server errors on the primary continue through the healthy cache list.

`/swift` requests are rewritten to `/api/registry/swift` only for cache fallback because that is the compatibility surface mounted by the cache service. Focused behavior assertions covered an unhealthy primary, a primary `503`, and canonical-path fallback. The TypeScript compiler check and Wrangler dry deployment also passed.

infra/registry-router/src/index.ts
```diff
@@ -11,8 +11,11 @@ interface Env {
 const HEALTH_CHECK_TTL = 120;
 const HEALTH_CHECK_TIMEOUT = 5000;
 const EARTH_RADIUS_KM = 6371;
+const PUBLIC_REGISTRY_HOST = "registry.tuist.dev";
+// Cloudflare keeps this alias out of the request host and uses it only to resolve the ingress address.
+const ORIGIN_RESOLVE_HOST = "registry-origin.tuist.dev";
```
fortmarek, line 16: [P2] Keep the Worker and origin alias temporary

The new `registry-origin.tuist.dev` alias and `resolveOverride` logic make sense only while the Worker provides regional cache fallback during the cutover. They should not become the permanent registry architecture. Once the standalone frontend has sufficient replicas and disruption protection, `registry.tuist.dev` can be a normal proxied DNS record whose underlying origin is the Kubernetes ingress, eliminating both the Worker and the resolution-only hostname. The legacy `tuist.dev/api/registry/*` URL does not require a Worker either: DNS cannot route by path, but the existing `tuist.dev` ingress already supports `server.ingress.extraPaths`, so that prefix can be transparently routed to the registry service while every other path continues to the server. This preserves compatibility without redirects, avoids cross-host authentication behavior, and removes the Worker KV health state, cron checks, path rewriting, fallback policy, and extra DNS identity from the steady-state request path. Please document and preferably encode this as an explicit post-soak removal step rather than presenting the Worker and `registry-origin.tuist.dev` as the long-term design.

fortmarek (approved):

PR #11829, opened 2026-07-14
feat(server): add source-backed Tuist answers

pepicrft:
## What changed

This adds a private Rust service that exposes bounded file listing, text search, and line-range reads over an immutable Tuist source revision. The hosted [Model Context Protocol](https://modelcontextprotocol.io/) server now offers `search_tuist_code`, `list_tuist_files`, and `read_tuist_file`, plus an `ask_tuist` prompt that accepts a question, finds relevant public material, and uses implementation and tests as the source of truth for current behavior.

The managed deployment builds the source-search image from the same commit as the server, deploys it behind a cluster-only service, and advertises the tools and prompt only when that service is configured.

## Why

Public material provides explanations and terminology, but accurate answers can also depend on defaults, feature gates, call sites, error handling, or focused tests that only exist in the repository. A small source-reading surface lets agents answer Tuist questions from that evidence without exposing shell execution or treating the repository as an unbounded resource.

## Approach

The `ask_tuist` prompt is organized around answering the question directly. It searches public sources for context and consults the code only when the answer depends on actual behavior, explicitly avoiding an unnecessary codebase tour.

The service uses fixed ceilings for concurrent work, request size, duration, traversal depth, files and bytes scanned, regular-expression complexity, readable file size, and returned content. Partial searches report explicit truncation metadata so callers can narrow their next request.

Each image contains a full immutable commit hash and emits revision-pinned GitHub source links. The workload runs as an unprivileged user with a read-only filesystem, no service-account token, processor and memory limits, ingress restricted to the server, and no network egress. Preview and self-hosted deployments do not advertise the tools.

## Impact

Hosted agents can now answer Tuist questions with public context and revisioned source evidence. Existing project tools and self-hosted behavior are unchanged. The managed deployment gains one small private workload and one image built for each deployed server revision.

## Validation

- `mise x rust@1.94.1 -- cargo fmt --all -- --check`
- `mise x rust@1.94.1 -- cargo clippy --all-targets -- -D warnings`
- `mise x rust@1.94.1 -- cargo test --all-targets` with 10 tests passing
- Elixir formatting checks for every changed Elixir source and test file
- Focused Elixir tests for the environment, server registration, prompt, client, and tools with 65 tests passing
- Helm chart linting with codebase search enabled
- Managed production and preview Helm renders, including verification that previews omit codebase search
- Workflow configuration parsing
- Docker image build and live checks for search, immutable source files, traversal rejection, request-size enforcement, and graceful shutdown


codebase-search/src/lib.rs
```diff
@@ -0,0 +1,1319 @@
...
+        let mut stats = SearchStats::default();
+        let mut returned_text_bytes = 0usize;
+        let mut truncation_reason = None;
+
+        let mut walker = WalkBuilder::new(start_path);
+        walker
+            .hidden(false)
+            .require_git(false)
+            .follow_links(false)
+            .max_depth(Some(self.limits.max_traversal_depth));
```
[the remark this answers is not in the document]
pepicrft: Fixed in a5600d97ee. `Codebase::search` and `Codebase::list_files` now install `WalkBuilder::sort_by_file_name` before file, byte, entry, result, and output limits are evaluated. The new `truncated_searches_follow_path_order` and `truncated_listings_follow_path_order` tests create `z.rs` before `a.rs` and verify capped responses return `a.rs`. Formatting, Clippy, and all 12 Rust tests pass.

.github/workflows/server-production-deployment.yml
```diff
@@ -1929,6 +1930,18 @@ jobs:
...
             MIX_ENV=prod
             APT_CACHE_BUST=${{ github.sha }}
+      - uses: docker/build-push-action@v6
+        name: Build and push codebase search image
+        with:
+          context: ./codebase-search
+          file: ./codebase-search/Dockerfile
+          push: true
+          tags: |
+            ${{ env.REGISTRY }}/${{ env.CODEBASE_SEARCH_IMAGE_NAME }}:${{ steps.tag.outputs.image_tag }}
```
fortmarek, line 1941: [P1] Make the new GHCR package pullable before enabling it

`ghcr.io/tuist/codebase-search` does not currently exist, and GitHub creates a container package as private on first publication by default. The managed chart supplies no image pull secret, so the first codebase-search pod will enter `ImagePullBackOff` and the atomic rollout will fail. Pre-create and publicize the package, automate its visibility, or configure pull credentials before enabling the workload. GitHub reference: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#pushing-container-images

.github/workflows/server-deployment.yml
```diff
@@ -919,6 +933,7 @@ jobs:
             --namespace "$NAMESPACE" --create-namespace \
             "${helm_values_args[@]}" \
             --set server.image.tag="$IMAGE_TAG" \
+            --set codebaseSearch.image.tag="$IMAGE_TAG" \
```
fortmarek: [P1] Preserve pinned-image rollbacks

When `image_tag` is supplied, the build job is skipped, but this assigns that server tag to the codebase-search image too. Rolling back to any server tag created before this service existed therefore references a missing image and fails the atomic Helm rollout. A separate `codebase_search_image_tag`, a known-good fallback, or explicit legacy-tag handling would preserve the documented rollback path.

codebase-search/src/lib.rs
```diff
@@ -0,0 +1,1377 @@
...
+            current_line += 1;
+            if current_line < start_line {
+                continue;
+            }
+            if lines.len() >= max_lines {
+                has_more = true;
+                break;
+            }
+
+            let text = bounded_bytes_line(&buffer, self.limits.max_line_bytes);
```
fortmarek: [P2] Surface per-line truncation

`bounded_bytes_line` clips lines above `max_line_bytes`, but that clipping does not affect `has_more` or `truncation_reason`. A long final or single line can therefore return incomplete source with `truncated=false`; search matches and context have the same behavior. Please add line-level truncation metadata or propagate a global truncation reason, with coverage for a single long line.

server/lib/tuist/mcp/server.ex
```diff
@@ -61,16 +67,28 @@ defmodule Tuist.MCP.Server do
...
       tools: tools(),
-      prompts: @prompts
+      prompts: prompts()
     )
   end
 
   defp tools do
-    if Environment.tuist_hosted?(), do: @hosted_tools ++ @tools, else: @tools
+    hosted_tools = if Environment.tuist_hosted?(), do: @hosted_tools, else: []
+    codebase_tools = if codebase_search_enabled?(), do: @codebase_tools, else: []
```
fortmarek, line 90: [P2] Add a product changelog entry

These newly registered hosted MCP tools and prompt are publicly observable product behavior, but the diff has no `server/priv/marketing/changelog/*.md` entry. Please add a short Product entry with `title`, `category`, and `pull_request` frontmatter.

fortmarek (approved):
Generally looks good—the bounded source-reading surface and deployment isolation are thoughtfully put together. Approving with four inline follow-ups worth addressing.

.github/workflows/server-production-deployment.yml
pepicrft: Confirmed. I published `ghcr.io/tuist/codebase-search:sha-bfe3f9576a66`, and the package now exists, but its package endpoint still reports `visibility: private`. GitHub does not expose package visibility changes through `gh`, so I am leaving this thread open until the package setting is switched to public. The managed chart still has no pull secret, so resolving this before that switch would hide the remaining rollout blocker.

.github/workflows/server-deployment.yml
pepicrft: Fixed in `904b211788`. `server-deployment.yml` now accepts a distinct `codebase_search_image_tag`. Fresh builds pass their revision-specific search tag through the reusable deployment workflow, while a pinned server rollback without an explicit search tag discovers and preserves the currently deployed codebase-search image. Both Helm render and upgrade paths use the resolved search tag. The changed workflows pass actionlint validation.

codebase-search/src/lib.rs
pepicrft: Fixed in `1f2f4d74cd`. Search-match and context clipping now marks the response with `truncated: true` and `truncation_reason: line_byte_limit`. `read_file` reports the same condition even for a single final line, without inventing a `next_start_line` when there is no later line. I added focused coverage for clipped match lines, context lines, and a single long read line; all 15 service tests and strict Rust lint pass.

server/lib/tuist/mcp/server.ex
pepicrft: Added `server/priv/marketing/changelog/2026.07.16-source-backed-tuist-answers.md` in `bfe3f9576a`. It uses Product frontmatter with pull request 11829 and explains the `ask_tuist` prompt, revision-pinned source evidence, and the bounded read-only source tools.

PR #11830, opened 2026-07-14
feat(server): run-centric runner dashboard with cancel-run support

fortmarek:
Makes the runner dashboard run-centric and adds the ability to cancel a running workflow run, modelled on the workflows → runs → jobs hierarchy.

### What changed

**Workflow detail now lists runs.** The workflow detail page (`/runners/workflows/:owner/:repo/:name`) previously listed individual jobs; it now lists **workflow runs** (grouped by `workflow_run_id`), newest activity first. Each run row shows:
- a rolled-up **status badge** — `Running` while any job is still in progress, otherwise the run's conclusion (failure > cancelled > success > skipped);
- branch, commit (linking out to the GitHub Actions run), job count, duration, last activity;
- a **Cancel** button on in-progress runs, shown **only when the account's GitHub App installation granted `actions: write`**.

**Cancel a run.** Clicking Cancel authorizes the new `:runners cancel` action and calls GitHub's cancel-run API (`POST /repos/:owner/:repo/actions/runs/:id/cancel`). GitHub cancels every job in the run and the `workflow_job.completed` webhooks reconcile our rows — we deliberately mutate no local state so the dashboard reflects GitHub's authoritative transition.

**Status on the workflows list.** Each row on the workflows list now carries a status badge derived from that workflow's latest run.

**Job → workflow link.** The job detail's CI Details card links up to its parent workflow.

### How it's built

- **Data** (`Tuist.Runners.Jobs`): `list_workflow_runs/4` + `count_workflow_runs/4` roll `runner_jobs` up per `workflow_run_id`, including in-progress runs, with the run-level status/conclusion. `list_workflows_for_account/2` now also returns the latest run's status. Both validated against production data.
- **GitHub** (`Tuist.GitHub.Client`): `cancel_workflow_run/3`; `installation_permissions/1` + `installation_has_actions_write?/1`, read from the permissions map GitHub already returns on the (cached) installation token — no extra round-trip.
- **Business logic** (`Tuist.Runners`): `cancel_workflow_run/3` + `can_cancel_workflow_runs?/1` resolve the account's installation and gate on `actions: write`.
- **Authorization**: new `:runners cancel` action (any org member; admin; ops).
- **Web**: workflow-detail runs table + gated Cancel + cancel event; status column on the workflows list; workflow link on the job detail.

### Permission rollout (required for Cancel to appear)

Cancelling a run needs the Tuist GitHub App to hold `actions: write` (it currently has `actions: read`). Detection is per-installation, so nothing breaks in the meantime — installations without write simply don't see the Cancel button. To enable it:
- **github.com App**: add `Actions: Read and write` in the App settings. Existing installs then get GitHub's "review updated permissions" prompt; the button appears for an org once it approves.
- **GHES manifest** (`github_app_manifest_controller.ex`): would need `actions: "write"` added for parity (that manifest is currently missing the runner permissions/events entirely — a separate GHES-runners gap, not addressed here).

### Notes / follow-ups
- Runs link out to the GitHub Actions run for now; an internal run-detail view (run → its jobs) is a natural follow-up.
- The workflows-list badge reflects the latest run's status; a per-run history strip (à la the reference UI) could come later.

### Validation
`mix test` across the runner + GitHub suites: **420 passed**. New/updated tests cover the runs query, run listing/scoping/pagination, the in-progress badge, Cancel visibility gating + the cancel event, the workflows-list status badge, and the job→workflow link. `mix credo` clean; `.pot` template updated (no `.po` edits).


server/lib/tuist_web/live/runner_workflow_run_live.ex
```diff
@@ -0,0 +1,221 @@
...
+     |> assign(:jobs_count, length(jobs))
+     |> assign(:can_cancel_runs, Runners.can_cancel_workflow_runs?(account))}
+  end
+
+  @impl true
+  def handle_event("cancel_run", _params, socket) do
+    %{selected_account: account, current_user: current_user, repository: repository, workflow_run_id: run_id} =
+      socket.assigns
+
+    with :ok <- Authorization.authorize(:runners_read, current_user, account),
```
pepicrft: Workflow cancellation is a destroy operation, but this authorizes it with `:runners_read`. That read action also allows regular account members and Tuist operators. Could we add a dedicated runner destroy action and use it both for rendering the button and handling the event? That keeps read-only access from being able to cancel a customer's workflow.

server/lib/tuist_web/live/runner_workflow_live.ex
```diff
@@ -313,14 +264,66 @@ defmodule TuistWeb.RunnerWorkflowLive do
...
+  def run_path(account_name, run_id) when is_binary(account_name) and is_integer(run_id) and run_id > 0 do
+    "/#{account_name}/runners/runs/#{run_id}"
+  end
+
+  def run_path(_, _), do: nil
+
+  # Duration for a run row: elapsed-so-far for an in-progress run,
+  # else the completed span. The SQL `duration_ms` is only meaningful
+  # once the run has a completed job, so guard on a positive value.
+  def run_duration_ms(%{status: "in_progress", enqueued_at: enqueued}), do: DateFormatter.ms_since(enqueued)
```
pepicrft: Could we base elapsed duration on the earliest non-null `started_at` instead of `enqueued_at`? A run whose only job is queued is rendered as Running here and accumulates queue time as Duration, while the same run's detail page correctly displays `–` for both Started and Duration.

I reproduced this locally in headless Chrome. Commit `5be0494` showed 8 minutes 33 seconds in the workflow table, but the run detail showed the only job was queued with no start time or duration.

Queued workflow duration mismatch

server/lib/tuist_web/live/runner_workflow_live.ex
```diff
@@ -313,14 +264,66 @@ defmodule TuistWeb.RunnerWorkflowLive do
...
+  def run_duration_ms(_), do: 0
+
+  @doc """
+  Public GitHub Actions run URL for a run row — the drill-down target
+  from the runs list (we don't host an internal run view yet).
+  """
+  def github_run_url(repository, run_id) when is_binary(repository) and is_integer(run_id) do
+    case String.split(repository, "/", parts: 2) do
+      [owner, name] when owner != "" and name != "" ->
+        "https://github.com/#{owner}/#{name}/actions/runs/#{run_id}"
```
pepicrft: The external link is fixed to github.com even when the account's runner installation targets GitHub Enterprise Server. I confirmed the runner path supports enterprise installations end to end: dispatch carries the installation host into just-in-time runner registration and the webhook resolver distinguishes installations by host. Could we derive this link from the account installation and add coverage for both github.com and GitHub Enterprise Server?

pepicrft (approved):
I ran the focused runner and GitHub test suites, built the assets, and exercised the new pages in headless Chrome at desktop and narrow widths. The implementation looks solid overall. I left three comments that I think are worth revisiting.

server/lib/tuist_web/live/runner_workflow_live.ex
fortmarek: Fixed: the runs list now measures in-progress elapsed time from the earliest job `started_at` (surfaced from the runs rollup) instead of `enqueued_at`, so a run whose only job is still queued reads `–` like the detail page. Covered in `runner_workflow_live_test.exs`.

server/lib/tuist_web/live/runner_workflow_live.ex
fortmarek: Fixed: the run/job/repository links now derive their host from the account installation's `client_url` via `VCS.github_base_url_for_account/1` (github.com, or the GitHub Enterprise Server host), instead of hardcoding github.com. Applied it to all three runner links (run, job, repo), with github.com + GHES coverage in `runner_workflow_run_live_test.exs` and `vcs_test.exs`.

server/lib/tuist_web/live/runner_workflow_run_live.ex
fortmarek: Done: added a dedicated `runners` `:cancel` action allowing account members + admins (not read-only tokens or Tuist operators), and gated both the Cancel button and the `cancel_run` handler on it. Operators keep read access but can no longer cancel. Covered in `authorization_test.exs` (member/admin allowed, non-member + operator denied).

PR #11834, opened 2026-07-14
feat(kura): export RSS anon/file split and jemalloc stats

esnunes:
## What

Adds five process-memory gauges to Kura, sampled in the existing 5s snapshot task:

- `kura_process_resident_anon_bytes` — `RssAnon` from `/proc/self/status`: private heap/stacks (where a leak shows; not reclaimable without swap).
- `kura_process_resident_file_bytes` — `RssFile`: mmap'd segments + code (reclaimable page cache).
- `kura_jemalloc_allocated_bytes` — jemalloc live allocations (`stats.allocated`): unbounded growth = real leak.
- `kura_jemalloc_resident_bytes` — jemalloc resident physical memory (`stats.resident`): far exceeding `allocated` = allocator retention, not a leak.
- `kura_jemalloc_retained_bytes` — jemalloc virtual memory purged back to the OS but not unmapped (`stats.retained`): confirms whether decay purging is returning memory.

jemalloc stats are read via a new `tikv-jemalloc-ctl` dependency, with the `stats` feature enabled on the allocator. **Observability only** — no change to the allocator config or the memory-pressure controller.

## Why

A production `mise r compile` against `tuist-eu-central-1.kura.tuist.dev` was returning:

```
WARNING: Remote Cache: RESOURCE_EXHAUSTED: action result needs 408 bytes
but only 0 bytes remain in the REAPI materialization budget
```

Root cause: `MaterializationBudget` seeds its per-request budget from `reapi_response_budget_bytes()`, which returns `0` under `MemoryPressure::Critical`. One replica (`kura-tuist-eu-central-1-1`) had climbed to ≥85% of its 2Gi limit (Critical), so every REAPI read — even a 408-byte ActionResult — was rejected. Bazel surfaces that gRPC status as a non-fatal warning and falls back to a cache miss; builds stay correct but uncached. Confirmed in Grafana: `kura_memory_pressure_state=2` for a sustained window, `kura_memory_actions_total{action="reapi_materialization_rejected"}=2364` on `-1` vs `0` on `-0`.

The follow-up question was whether that replica has a **memory leak** — it sat at ~1.5 GiB while idle. The behavioral evidence (working-set sawtooth that releases 400+ MiB, boots straight back to the same level after a restart, flat floor, no OOM-kills) argues *against* a classic leak and *for* a large resident working set on the hot replica plus jemalloc high-water retention (default config — `background_thread` off — so freed pages aren't purged while idle).

But that couldn't be **confirmed** remotely:
- `MemoryController` reads total `VmRSS` (`src/app.rs`), which can't separate reclaimable file-backed mmap pages from held anonymous heap.
- Grafana Cloud adaptive-metrics had aggregated away the `pod` label on the existing byte gauges.
- `kubectl exec`/`logs`/pod-proxy all fail on the dedibox bare-metal fleet nodes (streaming/network path to the kubelet is unreachable), so `/proc/self/status` and the in-process jemalloc stats are unreadable from outside.

These gauges ship the decisive numbers through the one telemetry path that works (local Alloy scrape → Prometheus).

## Why these five (and not more)

They cover every branch of the leak decision tree and no more:
- `anon` vs `file` → reclaimable segment cache vs held heap.
- `jemalloc allocated` vs `resident` → real leak vs allocator retention.
- `retained` → verifies the eventual `background_thread` mitigation returns memory.

Dropped `jemalloc active` (fragmentation — a less-likely cause, one-liner to re-add), `metadata` (tiny bookkeeping), and `mapped` (redundant with the existing `kura_process_virtual_memory_bytes`).

## Impact

Diagnostic only. No allocator or pressure-controller behavior changes; purely additive metrics, safe under rolling updates. `kura_jemalloc_*` and the RSS split populate wherever jemalloc is the allocator (all non-msvc targets) / on Linux respectively.

## Validation

- `cargo build`, `cargo clippy --all-targets -- -D warnings` (clean), `cargo test --lib` (342 pass).
- Bazel `mise run compile` (CI-gated path) — rules_rs resolves the new crate from `Cargo.lock`; `MODULE.bazel.lock` updated accordingly.
- New tests: `jemalloc_stats_snapshot_reads_live_allocator_stats` exercises the real mallctl path (epoch refresh + typed reads); `resident_rss_splits_into_anon_and_file` (Linux-gated) checks `RssAnon`/`RssFile` parsing and the `anon + file ≤ VmRSS` invariant.

- If adaptive-metrics later aggregates the new series, add them to the keep-list so `-1` stays distinguishable from `-0`.
- Once the split confirms retention: enable jemalloc `background_thread` (via `MALLOC_CONF`) and/or raise the 2Gi limit for the tuist-eu-central instances.
- Consider having the pressure controller act on `RssAnon` (or subtract mmap'd file RSS) instead of total `VmRSS`.


kura/src/metrics.rs
```diff
@@ -669,11 +679,36 @@ impl Metrics {
...
+            process_resident_file_bytes.clone(),
+        );
         registry.register(
             "kura_process_virtual_memory_bytes",
             "Process virtual memory size in bytes",
             process_virtual_memory_bytes.clone(),
         );
+        registry.register(
+            "kura_jemalloc_allocated_bytes",
+            "jemalloc bytes in live application allocations (stats.allocated); unbounded growth indicates a leak",
```
pepicrft: Non-blocking: Could we keep these descriptions factual and avoid classifying the cause? `stats.allocated` can grow because of legitimate live data, `stats.resident - stats.allocated` also includes metadata and fragmentation, and `stats.retained` represents virtual mappings that may not have associated physical memory. I would describe only what each counter measures. If we need stronger decay diagnostics later, we can also export active, dirty, and muzzy page statistics.

pepicrft (approved):
The implementation looks good overall. The sampling remains best-effort, keeps the existing memory-pressure behavior unchanged, and has focused coverage for the allocator and metric rendering. I left one non-blocking note about keeping the metric descriptions neutral.

PR #11836, opened 2026-07-14
feat(server): add runner concurrency limits

fortmarek:
## Motivation

Tuist Runners currently admits work without an account-level resource budget. Claiming a job reserves its ownership, but it does not limit the aggregate CPU or memory held by one account. Since macOS and Linux have different shapes and fleet capacity, a noisy account can consume disproportionate shared capacity and leave neighboring workloads waiting.

This change restores concurrency controls as resource budgets instead of a flat job count. CPU and memory are enforced independently per platform so admission reflects the actual runner shape being requested.

## What changed

- Added one `runner_concurrency_limits` row per account and platform, with deliberately low defaults for the current trial-heavy workload:
  - macOS: 12 vCPU / 28 GB
  - Linux: 32 vCPU / 64 GB
- Backfilled both platform rows for existing accounts and transactionally creates them for new user and organization accounts.
- Snapshotted platform, vCPU, and memory onto queued jobs and active claims.
- Moved claim reservation and the capacity decision into one `Claims.attempt/5` transaction. `Concurrency.fits?/3` is now a pure predicate rather than a database-backed API whose safety depends on caller ordering.
- Serialized admission by locking the indexed `(account_id, platform)` limit row with Ecto's `FOR UPDATE SKIP LOCKED` support.
- Made lock acquisition non-blocking. When an account/platform scope is already admitting work, dispatch immediately skips that account for the current poll and considers other queued work without consuming the lost-race retry budget.
- Inserted the uniqueness-sensitive claim reservation before acquiring the account/platform row lock. A concurrent primary-key or `pod_name` conflict can therefore wait without occupying that account's admission lock; after acquisition, the lock covers only the indexed usage aggregate, pure capacity decision, and transaction completion.
- Added a unique live-claim index on `pod_name`, so claim conflicts—not an application-level preflight query—enforce one active claim per pod across accounts and platforms.
- Persisted runner resource metadata in ClickHouse and derived exact hourly peak concurrency with an event sweep over claim/completion intervals.
- Added a platform selector and hourly peak CPU/memory charts to the Runners page. Red bars identify intervals where the configured limit was reached.
- Added ops controls for custom per-account platform limits.
- Updated the runner documentation, data-export inventory, seeds, Gettext templates, and product changelog with a focused dashboard screenshot.

## Root cause and approach

The original admission boundary had two related problems. `Concurrency.check_available/2` appeared safe as a standalone API but only remained atomic when called after a separate lock inside `Claims.attempt/5`. Concurrent callers that did not preserve that hidden ordering could read the same usage and both admit work. A blocking lock also meant a burst for one account could leave many transactions waiting while occupying or waiting for database-pool connections.

`Claims.attempt/5` now owns the complete OLTP boundary: validate the request, insert an uncommitted claim reservation, try to lock the account/platform limit row, aggregate live usage including that reservation, and run the pure capacity predicate. An over-limit or busy result rolls the reservation back; an admitted result commits it. PostgreSQL releases the row lock automatically at transaction completion.

The limit table is both the configuration model and the lock namespace. Its unique `(account_id, platform)` index gives every admission scope a concrete, collision-free row, avoiding encoded advisory keys. `FOR UPDATE SKIP LOCKED` returns immediately when another admission transaction holds that row, preserving fairness without moving accounting into a separate mutable counter that could drift from live claims. If a required limit row is missing, admission fails closed and rolls back the reservation.

The reservation is deliberately inserted before the row lock. PostgreSQL may wait on an uncommitted workflow-job or pod uniqueness conflict, but that wait no longer prevents unrelated work for the same account/platform from acquiring the admission lock. Once acquired, only the indexed aggregate and in-memory decision remain before commit or rollback.

The previous pod preflight was also serialized only by account, so simultaneous attempts for the same pod could race when they targeted different accounts or platforms. A database unique index closes that race at the actual write boundary while still allowing pod-name reuse after the live claim is deleted.

## Impact

Jobs are never admitted beyond an account's platform CPU or memory budget; they remain queued until a claim completes and releases capacity. Bursts for one busy account/platform no longer stack blocking admission transactions or prevent dispatch from considering other accounts. Linux and macOS admission for the same account can proceed independently.

Limits can be adjusted without changing runner profiles or fleet definitions, and the dashboard makes sustained limit pressure visible.

## Validation

- Focused concurrency, claims, schema, and ops LiveView suite: **50 passed**
- Accounts and runner dispatch suites: **288 passed, 2 excluded**
- Concurrent admission suite passed **100 consecutive runs** (400 race-focused tests), covering an eight-way burst against the 12-vCPU/28-GB macOS limit
- Concurrent coverage verifies immediate busy-row rejection, same-account cross-platform independence, cross-account independence, cross-account pod uniqueness, and that an intentionally blocked unique insert does not hold the account/platform lock
- New migration applied from a fresh database and completed a down/up round trip successfully
- `mix compile --warnings-as-errors`
- Changed-file `mix credo --strict`: no new findings; one pre-existing design suggestion remains in `Accounts`
- `mix excellent_migrations.check_safety`: no warnings for the new migration; repository-wide historical warnings remain
- Earlier affected runner, ops, controller, LiveView, and changelog parser suites: **181 passed**
- `mix assets.build`: app, marketing, docs, API docs, and Noora bundles built successfully
- `git diff --check`
- Local visual verification of macOS/Linux concurrency charts, limit coloring, legends, date ranges, platform switching, and ops controls


server/priv/docs/en/guides/features/runners.md
```diff
@@ -27,6 +27,14 @@ Adopting them is usually a one-line change to your workflow:
...
 ```
 
+## Concurrency limits {#concurrency-limits}
+
+Each account has independent vCPU and memory concurrency limits for macOS and Linux. Before Tuist admits a job, it checks the job's <.localized_link href="/guides/features/runners/profiles#machine-shapes">machine shape</.localized_link> against all jobs currently running on that platform. The job starts only when both its vCPU and memory fit within the remaining capacity; otherwise it stays queued until capacity becomes available. Work on one platform never consumes the other platform's capacity.
+
+The **Concurrency** section of the Runners dashboard shows the peak admitted vCPU and memory usage for every hour. A red bar means usage reached the configured limit at some point during that hour. Usage cannot exceed the line because jobs that would take either resource over its limit remain queued.
+
+Your current limits are shown alongside each chart. If your workflows regularly reach them and spend time queued, [contact us](mailto:contact@tuist.dev) with your account handle, platform, and expected parallel workload. We'll review the required capacity and help raise the limits.
+
```
fortmarek, line 37: we will also need to align on the default concurrency for folks on the Pro plan, but that's something to figure out down the line before runners become GA.

server/priv/repo/migrations/20260714120000_add_runner_resource_concurrency_limits.exs
```diff
@@ -0,0 +1,66 @@
...
+    # excellent_migrations:safety-assured-for-next-line check_constraint_added
+    create constraint(:runner_concurrency_limits, :runner_concurrency_limits_positive_memory,
+             check: "memory_gb > 0"
+           )
+
+    # Backfill the new table with the default policy before new account creation starts writing rows.
+    # excellent_migrations:safety-assured-for-next-line raw_sql_executed
+    execute(
+      """
+      INSERT INTO runner_concurrency_limits
```
pepicrft, line 73: **High: Backfill accounts created by old replicas during rollout**

This backfill runs in the pre-upgrade migration, while production keeps the two old server replicas serving traffic until the rolling update replaces them. An old replica can create an account after this `INSERT` commits without creating its `runner_concurrency_limits` rows. The missing rows are permanent: dispatch returns `:concurrency_limit_missing`, and the runners and operations pages call `Map.fetch!` on them. Could we make this invariant database-backed or run a second idempotent backfill after old writers are gone?

server/priv/repo/migrations/20260714120000_add_runner_resource_concurrency_limits.exs
```diff
@@ -0,0 +1,66 @@
...
+      # Zero-sized Linux defaults keep them loadable by Ecto while allowing
+      # new claims to carry their exact platform-specific resources.
+      # excellent_migrations:safety-assured-for-next-line column_added_with_default
+      add :platform, :string, null: false, default: "linux"
+
+      # excellent_migrations:safety-assured-for-next-line column_added_with_default
+      add :vcpus, :integer, null: false, default: 0
+
+      # excellent_migrations:safety-assured-for-next-line column_added_with_default
+      add :memory_gb, :integer, null: false, default: 0
```
pepicrft, line 94: **Medium: Account for claims written by old server replicas**

These defaults make existing claims and any claims inserted by old replicas after migration appear as Linux with zero resources. During the rolling update, new replicas sum zero for those active claims and can admit more work beyond the account budget. Could we deploy the schema first and enable resource admission only after all writers persist resources, or derive resources for live claims during the transition?

server/lib/tuist/runners/concurrency.ex
```diff
@@ -0,0 +1,495 @@
...
+  defp active_intervals_query(normalized_platforms, start_dt, end_dt, linux_default, macos_default) do
+    from(job in subquery(normalized_platforms),
+      where: job.platform != "",
+      select: %{
+        platform: job.platform,
+        active_from: fragment("greatest(?, ?)", job.claimed_at, ^start_dt),
+        active_until: fragment("least(ifNull(?, ?), ?)", job.completed_at, ^end_dt, ^end_dt),
+        vcpus:
+          fragment(
+            "if(? > 0, toInt64(?), if(? = 'linux', ?, ?))",
```
pepicrft: **Medium: Resolve legacy history from the fleet shape**

When the stored resource fields are zero, this falls back to the current platform default after `fleet_name` has been dropped. A historical non-default Linux job is consequently reported as the current default shape. Since shape pools existed before this migration, the 30-day charts understate processor and memory usage for those jobs. Could the normalized rows retain `fleet_name` and resolve known catalog pool shapes before using the platform default?

server/lib/tuist/runners/catalog.ex
```diff
@@ -169,6 +169,47 @@ defmodule Tuist.Runners.Catalog do
...
+        pool_name(%{platform: :linux, vcpus: shape.vcpus, memory_gb: shape.memory_gb}) == fleet_name
+      end)
+
+    case shape do
+      nil -> legacy_linux_resources(fleet_name)
+      shape -> {:ok, %{platform: :linux, vcpus: shape.vcpus, memory_gb: shape.memory_gb}}
+    end
+  end
+
+  defp legacy_linux_resources("linux-" <> _fleet_name), do: default_resources(:linux)
```
pepicrft, line 248: **Medium: Preserve rollout support for operator-defined Linux pools**

Helm renders `runnersFleetLinux.pools` with the same component prefix that `fleet_platform/1` recognizes, but this fallback accepts only old names beginning with `linux-`. A pre-migration queued job in a supported custom pool therefore returns `:invalid_resources`; dispatch returns an internal error without excluding the job, so the oldest job pins that pool. Could this fallback obtain the custom pool's configured resources instead of rejecting every prefixed non-catalog pool?

pepicrft (commented):
I found four rollout and accounting issues that I think we should address before merging.

server/lib/tuist/runners/concurrency.ex
fortmarek: Addressed in bf483fe3c4. The normalized ClickHouse rows now retain `fleet_name`, and the hourly interval query maps known Linux fleet names to the exact configured shape before falling back to the platform default. I added a regression with a zero-resource historical job on a non-default Linux shape and assert that both the vCPU and memory series use that shape.

server/priv/repo/migrations/20260714120000_add_runner_resource_concurrency_limits.exs
fortmarek: Addressed in bf483fe3c4. Admission now detects zero-resource claims written by old replicas and derives their platform, vCPU, and memory from the configured fleet catalog before calculating reserved capacity. This means a legacy macOS claim whose new columns contain the database defaults is counted against macOS with its real shape instead of as zero-sized Linux work; an unresolvable legacy fleet still fails closed. The regression inserts that exact legacy row and verifies a subsequent macOS claim is rejected with the derived 6-vCPU/14-GB usage.

server/priv/repo/migrations/20260714120000_add_runner_resource_concurrency_limits.exs
fortmarek: Addressed in bf483fe3c4. The migration now installs an `AFTER INSERT` trigger on `accounts` before running the backfill, so an account written by an old replica receives both platform-limit rows in the same database transaction. The trigger, backfill, and application-side inserts are all idempotent via `ON CONFLICT DO NOTHING`. I added a regression that inserts an account directly through `Repo` (bypassing the new account-creation flow) and asserts both default limits, and I rehearsed the migration down/up cycle locally.

server/lib/tuist/runners/catalog.ex
fortmarek: Addressed in bf483fe3c4. Helm now serializes each operator-defined Linux pool's name and effective pod CPU/memory into `TUIST_RUNNER_LINUX_POOLS`; runtime config parses it into the catalog, which resolves the exact prefixed fleet name for queued jobs and legacy live claims. Partial CPU/GiB units round up so admission cannot understate a reservation. Coverage includes the JSON boundary and exact custom-pool resolver, and I rendered the chart with a custom 6500m/20000MB pool to verify the injected value.

server/lib/tuist/runners/claims.ex
```diff
@@ -58,62 +60,163 @@ defmodule Tuist.Runners.Claims do
...
+  defp positive_integer?(value), do: is_integer(value) and value > 0
+  defp non_empty_binary?(value), do: is_binary(value) and value != ""
+
+  defp try_lock_concurrency_limit(account_id, platform) do
+    limit_query =
+      from(limit in ConcurrencyLimit,
+        where: limit.account_id == ^account_id and limit.platform == ^platform
+      )
+
+    case Repo.one(from(limit in limit_query, lock: "FOR UPDATE SKIP LOCKED")) do
```
pepicrft, line 137: **Non-blocking question: preserve the rollout limit guarantee**

An old replica still serializes claims with `pg_advisory_xact_lock`, while this version locks the per-platform limit row. I reproduced an overlap where the old transaction inserted a zero-resource macOS claim but had not committed. The new transaction acquired this row, saw only its own `%{vcpus: 12, memory_gb: 28}` reservation, and committed. The old transaction then committed another `%{vcpus: 6, memory_gb: 14}`, leaving a derived total of `18/42` against a `12/28` limit. This is confined to the mixed-version rollout, so I do not think it should block this change, but the description's guarantee starts only after old replicas drain. Would it be worth documenting this or tracking staged activation if strict rollout enforcement matters?

server/lib/tuist/runners/concurrency.ex
```diff
@@ -0,0 +1,529 @@
...
+  Returns aggregate active-claim resources for each platform.
+  """
+  def usage_by_platform(account_id) when is_integer(account_id) do
+    usage =
+      Claim
+      |> where([claim], claim.account_id == ^account_id)
+      |> group_by([claim], claim.platform)
+      |> select([claim], {
+        claim.platform,
+        %{vcpus: sum(claim.vcpus), memory_gb: sum(claim.memory_gb)}
```
pepicrft, line 68: **Non-blocking question: derive legacy claims in live summaries**

Admission now derives zero-resource claims from their fleet, but this query still groups by the stored platform and sums the zero-valued fields. I inserted the same legacy macOS row and this aggregate reported Linux `0/0`, while admission derived macOS `6/14`. The Runners page can therefore show zero current use until old claims complete. This is also confined to rollout and is non-blocking, but could we reuse the claim normalization here or explicitly accept the short-lived display inconsistency?

pepicrft (approved):
Validated the fixes on the latest head against the four original threads. The account trigger, legacy resource derivation, historical fleet mapping, and operator-defined pool configuration address the reported failures. I reproduced the two short-lived mixed-version caveats noted inline. Neither persists after old replicas drain, so I consider them non-blocking.

server/lib/tuist/runners/claims.ex
fortmarek: > This is confined to the mixed-version rollout, so I do not think it should block this change, but the description's guarantee starts only after old replicas drain. 

I don't think this matters much given runners are now generally available. If we overcommit resources for a little bit, that's totally fine

server/lib/tuist/runners/concurrency.ex
fortmarek: same as above, I don't think it's super important how we handle the state during the rollout which will be fast.

PR #11839, opened 2026-07-14
fix(server): keep preview alive and surface migration failures

pepicrft:
## What

Preview deploys fail because the server migration Job fails (`failed: 7/1`, i.e. all 7 attempts), and **nothing has ever printed why**. This makes the failure visible and stops it taking the whole preview environment down with it.

- Drops `--wait-for-jobs` from `helm upgrade`.
- Adds a dedicated **Wait for migrations** step that blocks on the migration Job and, on failure, prints its log and `describe`.

## Why the failure was invisible

Three things conspired:

1. `helm upgrade` ran with `--wait`, which only waits for **Deployments** to go ready, not for **Jobs** to complete. Preview runs the migration as a regular Job (`server.migrationJob.asHook=false`), so nothing blocked on it. Seeding then ran against a half-migrated database, which is the long-standing and very confusing `Table tuist.build_runs does not exist` error. The database was not missing a migration, it was still being migrated.

2. Adding `--wait-for-jobs` (#11835) made Helm finally *notice* the Job had failed. But it combines badly with `--atomic`: the failed upgrade is rolled back and the release is **uninstalled**, which deletes the migration pod. The only copy of the real error goes with it, and the preview is torn down entirely, so we traded a confusing error for no environment and no explanation.

3. The `Diagnostics on failure` step only ever dumped Kura. It never looked at the migration Job.

## What this changes

Blocking on the Job moves into an explicit step, so seeding still cannot run against a half-migrated database, but a migration failure now:

- **leaves the release standing**, so the preview environment still comes up and is usable, and
- **prints the migration Job's log and describe output**, so the next failing deploy actually tells us what is wrong.

## What this deliberately does *not* claim

It does not fix the migration Job itself, and I do not want to pretend otherwise.

The migrations are **not** the bug. I replayed them locally in exactly the order the release uses (`Tuist.Release.migrate` runs all of `Tuist.Repo`, then all of `Tuist.IngestRepo`) against a fresh Postgres **and** a fresh ClickHouse, mirroring a preview's embedded databases:

```
>>> PHASE 1: Postgres migrations
>>> Postgres done
>>> PHASE 2: ClickHouse migrations
>>> ALL MIGRATIONS APPLIED CLEANLY
```

All of them apply cleanly, backfills included. I also checked that `ensure_database_schema/1` and the `grant_*_role/1` calls in `Tuist.Release.migrate/0` are no-ops for the ClickHouse repo (guarded by `when repo == Tuist.Repo`), so the release path does not do anything extra that my replay skipped.

So whatever breaks the Job is specific to the in-cluster environment, and it cannot be diagnosed from outside the cluster while the evidence is being deleted on failure. This PR is what makes it legible. Once a deploy runs with this, the migration error will be in the log and the real fix can follow.

## Validation

- `preview-deploy.yml` parses as valid YAML (10 jobs).
- Step order confirmed: `helm upgrade --install` -> `Wait for migrations` -> `Smoke test` -> `Seed preview account`, so nothing seeds ahead of migrations.
- `--wait-for-jobs` no longer appears in the helm invocation (only in the comment explaining why it is absent).
- The wait keys off the Job's `Complete` / `Failed` conditions rather than raw counts, so Job retries within `backoffLimit` are not mistaken for terminal failure.
- Reproduced the full preview migration sequence locally against fresh Postgres + ClickHouse to establish that the migrations themselves are sound (see above).

Found while getting #11743 back to green. Follows #11831 (preview ClickHouse memory) and #11835.

### DIFF
--- a/.github/workflows/preview-deploy.yml
+++ b/.github/workflows/preview-deploy.yml
@@ -696,11 +696,15 @@ jobs:
         # lifecycle. Its static shape lives in values-preview-kura.yaml; only
         # the per-preview identity is set here.
         #
-        # --wait-for-jobs is load-bearing: --wait on its own only waits for the
-        # Deployments to go ready, not for Jobs to complete. Preview runs the
-        # migration as a regular Job (server.migrationJob.asHook=false), so
-        # without this Helm returns success while migrations are still running
-        # and the seed step below writes into a half-migrated database.
+        # Deliberately NOT --wait-for-jobs. It combines badly with --atomic:
+        # a failing migration Job fails the upgrade, --atomic then uninstalls
+        # the release, and the migration pod (the only place the actual error
+        # is written) is deleted before anything can read it. The preview is
+        # torn down and the failure is unexplainable.
+        #
+        # The "Wait for migrations" step below blocks on the Job instead, so
+        # the seed still never runs against a half-migrated database, but a
+        # failure leaves the release standing and prints the migration log.
         run: |
           set -euo pipefail
           helm upgrade --install "$RELEASE" "$HELM_CHART_PATH" \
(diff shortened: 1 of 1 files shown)


pepicrft:
Security note added after review: printing the migration Job log naively would have leaked the preview Postgres password into a public Actions log.

Ecto logs the SQL of every migration it runs, and the Postgres-to-ClickHouse backfills interpolate the credentials straight into that SQL:

```
... FROM postgresql('host:port', 'db', 'command_events', 'user', 'password')
```

I confirmed this by replaying the ClickHouse migrations locally against a Postgres configured with a known password: it shows up verbatim 4 times in the migration output. `tuist/tuist` is public, so an unfiltered `kubectl logs` would have published it.

The log and describe output now go through a redactor that masks the `postgresql(...)` credential argument, `postgres://user:pass@` URLs, and bare `password:` fields. Verified against the captured log: 4 occurrences before, 0 after.

Worth flagging separately: **this is a pre-existing leak, not one this PR introduces.** The credentials are in the migration Job's log wherever it runs, including staging and production, where they land in cluster logging. That is a smaller blast radius than a public CI log, but it is still a database password sitting in plaintext logs. The clean fix is for the backfill migrations to stop interpolating the password into logged SQL (for example by running the statement through `execute(fn -> ... end)`, whose function form is not logged). Happy to open that as a follow-up.

PR #11842, opened 2026-07-14
feat(infra): track a per-deploy dispatch generation on RunnerPools

esnunes:
## What

Adds a per-deploy "dispatch generation" to the RunnerPool CRD and has the controller record when it changes:

- **`spec.dispatchGeneration`** — the chart stamps this on every pool (macOS xcode pools and Linux shape pools, all render paths) from the server release identifier (`.Values.server.image.tag`), so it changes on every server deploy.
- **`status.dispatchRolledAt`** (+ `status.observedDispatchGeneration`) — the reconciler stamps the moment the generation changes, a direct mirror of the existing `observedImage`/`imageRolledAt` image-roll bookkeeping.

## Why

This is the first, **dormant** increment of a fix for registered-idle runner staleness. A runner that has claimed a job and registered with GitHub, but is still waiting for the scheduler to assign it work, sits "registered-idle" — often for hours. Everything the server staged at claim time (the JIT, and the dispatch-provided `cache_endpoint_url`) is frozen for that whole window. When a server deploy changes dispatch behavior, those pre-deploy-claimed runners keep running jobs with the **old** config.

That is exactly what happened on 2026-07-11: #11767 changed what dispatch stages; the server was correct within minutes, but jobs kept landing on runners whose claims predated the deploy and ran without the new cache endpoint for ~3 hours.

The eventual fix recycles registered-idle Pods whose claim predates the last deploy. That needs a boundary — *"when did dispatch behavior last change?"* — which is what `dispatchRolledAt` provides.

## Why this shape

`dispatchGeneration`/`dispatchRolledAt` deliberately mirror the existing `observedImage`/`imageRolledAt` pattern, so the mechanism is familiar and the future drain can reuse the same image-roll pacing machinery. Sourcing the generation from `.Values.server.image.tag` ties the boundary to **server code** deploys (which is what changes dispatch behavior), not runner-image rolls (already handled separately by the image-roll path).

## Scope / safety

Inert on its own — **no Pod is drained in this PR**. With no chart-stamped generation the spec value is empty and never differs from the empty observed value, so the controller does nothing; once the chart wires it, the controller only records a timestamp. No change to Pod lifecycle.

Intent node (`infra/runners-controller/AGENTS.md`) updated to describe the new field as dormant groundwork.

## Validation

- `go build ./...` and `go vet ./api/...` clean.
- Controller tests pass, including two new reconcile tests: stamps `dispatchRolledAt` when the generation first appears; leaves it zero when the generation is empty.
- `helm template` renders `dispatchGeneration` onto every macOS xcode pool and Linux shape pool from `.Values.server.image.tag` (verified value flows through as `sha-deadbeef` in a test render).

## Next (for context, not in this PR)

- A third-state "busy" signal so the controller can tell a registered-idle Pod from one running a job (open fork: Linux readiness probe vs server-stamped label).
- The drain itself: recycle registered-idle Pods that claimed before `dispatchRolledAt`, paced by the existing roll-concurrency cap.

### DIFF
--- a/infra/helm/tuist/crds/tuist.dev_runnerpools.yaml
+++ b/infra/helm/tuist/crds/tuist.dev_runnerpools.yaml
@@ -182,6 +182,9 @@ spec:
                         minimum of 1 so a rollout always makes progress. Lower
                         keeps more of the warm pool serving during a roll at
                         the cost of a slower rollout.
+                dispatchGeneration:
+                  type: string
+                  description: Opaque token the chart stamps from the server release identifier; changes on every server deploy. The controller records when it changes in status.dispatchRolledAt.
             status:
               type: object
               properties:
@@ -197,6 +200,13 @@ spec:
                 observedImage:
                   type: string
                   description: spec.image value the controller last recorded an imageRolledAt for. Bumped together on every image change.
+                observedDispatchGeneration:
+                  type: string
+                  description: spec.dispatchGeneration value the controller last recorded a dispatchRolledAt for. Bumped together on every change, mirroring observedImage/imageRolledAt.
+                dispatchRolledAt:
+                  type: string
+                  format: date-time
+                  description: Timestamp the controller last observed spec.dispatchGeneration change (the last server deploy). A future drain uses it as the boundary before which a registered-idle Pod's claim-time dispatch config is stale.
                 imageRolledAt:
                   type: string
                   format: date-time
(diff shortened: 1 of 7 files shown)


esnunes:
Pivoting to the simpler in-pod idle-registration TTL watchdog (per discussion): it reaps zombies, bounds staleness, and covers macOS uniformly with far less surface than the controller deploy-drain, and it leans on the controller's existing reap→gap-fill to replace a recycled pod with a fresh claimer. This deploy-boundary trigger (dispatchGeneration/dispatchRolledAt) is shelved but preserved in this PR's history — recover it if we ever need event-driven (sub-TTL) convergence. The TTL work reuses this branch.

PR #11847, opened 2026-07-14
feat(server): meter public cache egress

pepicrft:
## What changed

This separates operational cache traffic from billable cache usage and introduces the first Stripe meter for public cache egress.

- Cache usage now lives under `Cache > Usage` and keeps the existing egress, ingress, request, project, and regional views.
- Billing usage now lives under `Billing > Usage` and shows only billable public Internet egress as a cumulative graph for the current billing period.
- Kura usage rollups now record their factual network path as `public_internet`, `private_network`, or `unknown`.
- The daily billing worker reports the previous day's billable egress bytes to the Stripe `cache_egress_byte` meter.
- Development seeds include public and collocated runner-cache traffic so both pages have representative data.

## Why

Kura nodes are now collocated with managed runners. Traffic between those private runner caches and nearby runners is operationally useful, but it must not contribute to customer charges. Traffic delivered from Tuist-managed cache infrastructure over the public Internet needs a stable billing boundary that can be reconciled with Stripe.

The two usage surfaces serve different purposes. Cache usage explains how the cache is behaving, while Billing usage explains the quantity that may appear on an invoice.

## Root cause

The original billing boundary used the serving region as a proxy for the network path. That worked while public and private deployments were separate, but it encoded deployment placement rather than the fact we want to bill: whether the response traveled over the public Internet.

There was also no dedicated billable query or Stripe meter event, and customer-operated Kura submissions could provide their own traffic-plane value. Customer-controlled classification is not a trustworthy input for deciding whether Tuist incurred public egress.

## Approach

Each Kura deployment now configures the network path it serves. Public cache instances emit `public_internet`, collocated runner-cache instances emit `private_network`, and standalone or unclassified instances default to `unknown`. The value is stored on every usage rollup and included in its deterministic event identifier.

The billable query requires egress direction, the public client traffic plane, and `network_path = public_internet`. Region names no longer determine billing. Private and unknown paths, peer traffic, and customer-operated traffic fail closed and are excluded.

During a mixed-version rollout, the server classifies older managed events that do not contain the field from the deployment topology. Explicit invalid values become `unknown`. Usage submitted with a customer-operated credential is always rewritten to the customer-operated plane and unknown network path, regardless of the payload.

The Stripe worker sends one pre-aggregated daily byte total with a deterministic event identifier and idempotency key. The billing graph uses the active Stripe subscription period start when available and falls back to the current calendar month for accounts without an active subscription. The graph is cumulative and the period is enforced internally, so the page does not expose a date control.

## Impact

Private runner-cache traffic remains visible under Cache usage for operational diagnosis but does not contribute to Billing usage or Stripe. Existing usage rows receive the safe `unknown` default and are not retroactively billed. The former top-level `/:account/usage` path permanently redirects to `/:account/cache/usage`.

The new Stripe meter can collect usage before a price is attached. This change does not start charging customers by itself.

### Billing usage

Billing usage with cumulative public Internet egress

### Cache usage

Cache usage with operational traffic widgets, graph, and regional breakdown

## Validation

- `mise run format -- --check`
- `mise run compile`
- `mise run test-unit`: 347 passed, one existing scale test ignored
- `mise run clippy`
- Focused server tests: 122 passed
- Strict Credo analysis on the 11 changed server source and test files: no issues
- Kura and Tuist Helm chart linting and template rendering
- Headless Chrome verification of both local pages, including the cumulative billing-period graph and the absence of widgets, ingress, requests, and date controls on Billing usage

### How to test locally

1. From `server`, load the development seed data and start the server.
2. Sign in with the seeded account.
3. Open `/tuist/cache/usage` and confirm operational traffic includes both public and private seeded paths.
4. Open `/tuist/billing/usage` and confirm only public Internet egress appears as a cumulative graph from the start of the current billing period.


server/lib/tuist_web/components/app_layout_components.ex
```diff
@@ -409,30 +409,48 @@ defmodule TuistWeb.AppLayoutComponents do
...
         selected={String.starts_with?(@current_path, ~p"/#{@selected_account.name}/webhooks")}
       />
-      <.sidebar_item
+      <% cache_path = ~p"/#{@selected_account.name}/cache" %>
+      <% cache_usage_path = ~p"/#{@selected_account.name}/cache/usage" %>
+      <.sidebar_group
         :if={
           FeatureFlags.kura_enabled?(@selected_account) and
             Authorization.authorize(:account_update, @current_user, @selected_account) == :ok
         }
```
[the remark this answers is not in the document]
pepicrft: Fixed in `48340611f5`. The Cache sidebar group now checks `Authorization.authorize(:projects_read, ...)`, which matches `UsageLive.mount/3`, so regular organization members can discover and open `/cache/usage`. The group-level Cache settings navigation remains conditional on `:account_update`, preserving the existing admin restriction. I also ran `mix test test/tuist_web/live/usage_live_test.exs test/tuist_web/live/billing_usage_live_test.exs`; all 15 tests passed.

server/lib/tuist_web/router.ex
```diff
@@ -1050,8 +1051,9 @@ defmodule TuistWeb.Router do
       live "/webhooks/:id", WebhookLive
       live "/webhooks/:id/events/:attempt_id", WebhookEventLive
       live "/cache", CacheLive
+      live "/cache/usage", UsageLive
       live "/billing", BillingLive
-      live "/usage", UsageLive
+      live "/billing/usage", BillingUsageLive
```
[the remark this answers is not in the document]
pepicrft: I am not adding a changelog entry in this pull request because this surface is still a controlled rollout. `BillingUsageLive.mount/3` raises `NotFoundError` unless `FeatureFlags.kura_enabled?/1` succeeds, and on the hosted service that function requires the explicit `:kura` toggle for the account. This pull request does not make the feature broadly available. Both `server/priv/AGENTS.md` and the Tuist Elixir review guidance exclude account-gated or otherwise unannounced functionality from the product changelog. The changelog entry should be added when the Kura billing usage surface is made generally available.

fortmarek:
image

I think this can use some work. Right now, we're just showing basically the same graphs as in the existing cache usage. This should be focused on $$$, something more akin to https://tuist.grafana.net/a/grafana-cmab-app/usage?from=now%2F1M&to=now&compareFrom=now-1M%2FM&compareTo=now-1M where we spell out how much a user got billed in the past and how much they will get billed next month.

I realize we don't have the price, so this page and everything billing related for cache should be behind a dedicated new feature flag.

server/lib/tuist/billing/workers/sync_customer_stripe_meters_worker.ex
```diff
@@ -1,32 +1,41 @@
...
-      Sentry.Context.set_extra_context(%{
-        customer_id: customer_id,
-        date: date
-      })
-    end
-
     {:ok, account} = Accounts.get_account_from_customer_id(customer_id)
 
-    {:ok, _} = Billing.update_remote_cache_hit_meter(customer_id, idempotency_key)
+    meters = ["remote_cache_hit", "cache_egress"]
```
fortmarek: **[P1] Gate Kura billing independently from Kura availability**

This unconditionally schedules `cache_egress` for every Stripe-backed account. The only UI gate is currently `:kura`, so there is no account-level kill switch for meter submission and no way to roll Kura billing out independently. Please introduce an account-scoped flag such as `FeatureFlags.kura_billing_enabled?/1` and apply it here, to the Billing → Usage navigation/mount guard, and when the corresponding Stripe price is attached. The network-path ingestion and aggregation can remain active for shadow validation while billing is disabled.

server/lib/tuist_web/live/billing_usage_live.html.heex
```diff
@@ -0,0 +1,59 @@
+<div id="usage">
+  <div data-part="page-header">
+    <div data-part="header">
+      <h1 data-part="title">{dgettext("dashboard_usage", "Metered usage")}</h1>
+      <span data-part="subtitle">
+        {dgettext(
+          "dashboard_usage",
+          "Billable public cache traffic for this account. Private runner traffic is excluded."
```
fortmarek: **[P1] Do not call this traffic billable without contract-aware gating**

The meter worker submits cache egress regardless of plan, while Enterprise subscriptions are negotiated flat, invoice-billed contracts and may not include metered public egress. Self-hosted deployments also satisfy `kura_enabled?/1`, even though their customer-operated traffic is deliberately excluded. As written, those accounts can see a “billable” page that does not reflect their agreement. Gate this surface with the separate account-level Kura-billing toggle—disabled by default for Enterprise and self-hosted—or present it as informational “metered public egress” and model chargeability separately.

server/lib/tuist/billing.ex
```diff
@@ -122,6 +123,47 @@ defmodule Tuist.Billing do
...
+  """
+  def update_cache_egress_meter(
+        %Account{customer_id: customer_id} = account,
+        idempotency_key,
+        now \\ Tuist.Time.utc_now()
+      )
+      when is_binary(customer_id) do
+    start_of_yesterday = now |> Timex.shift(days: -1) |> Timex.beginning_of_day()
+    end_of_yesterday = now |> Timex.shift(days: -1) |> Timex.end_of_day()
+    bytes = Usage.billable_egress_bytes(account.id, start_of_yesterday, end_of_yesterday)
```
fortmarek: **[P1] Reconcile late Kura rollups before finalizing the day**

This takes one snapshot of yesterday from a daily midnight billing job, but Kura closes 60-second windows and delivers its durable outbox asynchronously. The final 23:59 window—or any temporarily delayed rollup—can arrive after this job succeeds, leaving the Stripe total permanently partial or skipped. Report a sufficiently lagged day with reconciliation, or submit closed rollups continuously with deterministic identifiers, so successful early execution cannot finalize incomplete usage.

server/lib/tuist/billing.ex
```diff
@@ -122,6 +123,47 @@ defmodule Tuist.Billing do
...
+      path = Stripe.OpenApi.Path.replace_path_params("/v1/billing/meter_events", [], [])
+      usage_date = start_of_yesterday |> DateTime.to_date() |> Date.to_iso8601()
+
+      []
+      |> Stripe.Request.new_request(%{"Idempotency-Key" => "#{idempotency_key}-cache-egress-byte"})
+      |> Stripe.Request.put_endpoint(path)
+      |> Stripe.Request.put_params(%{
+        event_name: "cache_egress_byte",
+        identifier: "#{customer_id}-cache-egress-byte-#{usage_date}",
+        timestamp: DateTime.to_unix(start_of_yesterday),
```
fortmarek: **[P1] Preserve usage timestamps across subscription-period boundaries**

This assigns the entire previous calendar day's aggregate to midnight. Stripe subscription billing anchors can occur at arbitrary times, so on a renewal day traffic from both sides of the anchor is attributed to a single billing period. Preserve hourly/raw rollup timestamps when reporting, or split the aggregate at the active subscription boundary before creating meter events.

fortmarek (changes requested):
Requesting changes for the remaining billing correctness and rollout blockers:

- Kura billing needs an account-scoped rollout switch independent from the existing Kura availability flag.
- Enterprise and self-hosted billing semantics are not represented by the new "billable" surface.
- A midnight snapshot can permanently miss asynchronously delivered Kura rollups.
- Assigning a whole calendar day to midnight cannot preserve arbitrary Stripe subscription boundaries.
- The existing open P1 on `SyncCustomerStripeMeterWorker` also remains: pass the intended reporting date through the parent and every child meter job. This must cover `remote_cache_hit` and `llm_token` as well as `cache_egress`, since all three currently recompute "yesterday" at execution time.

The earlier `data-export.md` gap is fixed on the current head: it now documents the per-rollup `kura_usage_events` data, including `network_path`, retention, and billing usage.

server/lib/tuist/billing/workers/sync_customer_stripe_meter_worker.ex
```diff
@@ -0,0 +1,38 @@
...
+    end
+  end
+
+  defp update_meter("remote_cache_hit", customer_id, idempotency_key) do
+    Billing.update_remote_cache_hit_meter(customer_id, idempotency_key)
+  end
+
+  defp update_meter("cache_egress", customer_id, idempotency_key) do
+    {:ok, account} = Accounts.get_account_from_customer_id(customer_id)
+    Billing.update_cache_egress_meter(account, idempotency_key)
```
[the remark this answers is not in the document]
pepicrft: Fixed in `0011884932`. `SyncStripeMetersWorker` now chooses the reporting date once and passes it through both per-customer workers. Cache-hit, Kura egress, and language-model token queries, identifiers, and Stripe timestamps all use that stored date, so queue delay or retry after midnight cannot change the usage window. The worker tests also cover an explicit backfill date and all three meter paths.

server/lib/tuist/billing/workers/sync_customer_stripe_meters_worker.ex
pepicrft: Fixed in `0011884932`. `FeatureFlags.kura_billing_enabled?/1` is a dedicated hosted, account-scoped gate. It now controls Kura meter scheduling, Billing Usage navigation and mount authorization, and attachment of the optional `kura_usage` Stripe price category. Kura ingestion remains independent for shadow validation, and tests cover enabled, disabled, and customer-operated deployment behavior.

server/lib/tuist_web/live/billing_usage_live.html.heex
pepicrft: Fixed in `0011884932`. The Billing Usage route and sidebar now require the dedicated account-scoped `:kura_billing` flag. That flag is off by default for every account, does not inherit Kura availability, and always evaluates false on customer-operated deployments. Enterprise accounts therefore do not see or submit billable Kura usage unless their contract is represented by explicitly enabling this separate gate.

server/lib/tuist/billing.ex
pepicrft: Fixed in `0011884932`. Billing now selects closed Kura rollups by their actual ingestion date, so a delayed durable-outbox delivery is picked up after it arrives. After Stripe accepts a rollup, its deterministic Kura event identifier is recorded in the new `kura_billing_events` PostgreSQL ledger; later deliveries are excluded even after ClickHouse compaction. The migration, data-export documentation, and delayed-redelivery tests are included.

server/lib/tuist/billing.ex
pepicrft: Fixed in `0011884932`. Kura egress is no longer collapsed into one calendar-day event at midnight. Each immutable rollup is submitted with its original `window_start` timestamp and a deterministic hashed identifier, so Stripe assigns usage on the correct side of an arbitrary subscription boundary. The billing test asserts the exact source timestamp sent to Stripe.

pepicrft:
@fortmarek I agree that a billing page should ultimately show historical charges and a forecast rather than duplicate cache traffic charts. We do not have an account contract price source in this change, so I did not invent a cost projection. In `0011884932` I put the page, navigation, meter submission, and optional Stripe price attachment behind the new account-scoped `:kura_billing` flag. It is disabled by default and always disabled for customer-operated deployments. This keeps the traffic view available only for controlled billing validation; it should not be enabled broadly until the price source and dollars-focused experience are modeled.

pepicrft:
@fortmarek The five blockers from the changes-requested review are addressed on the current head:

- `FeatureFlags.kura_billing_enabled?/1` is a hosted, account-scoped rollout switch independent from Kura availability. It gates meter scheduling, the Billing Usage navigation and mount, and attachment of the optional Stripe price.
- Enterprise accounts are disabled by default and require an explicit account flag. Customer-operated deployments always evaluate the billing flag as false.
- `Billing.update_cache_egress_meter/3` selects Kura rollups by ingestion date and records accepted event identifiers in the PostgreSQL `kura_billing_events` ledger, so delayed outbox delivery is reconciled without duplicate reporting.
- Each Kura rollup is submitted with its original `window_start` timestamp and deterministic identifier, preserving arbitrary Stripe subscription boundaries.
- `SyncStripeMetersWorker`, `SyncCustomerStripeMetersWorker`, and `SyncCustomerStripeMeterWorker` carry one explicit `usage_date` through remote cache hit, Kura egress, and language-model token reporting. Retries after midnight cannot change the reporting window.

These changes are in `0011884932`. The current head also includes the generated Gettext template updates and continuous integration fixes. The full Server test job, Kura Bazel and end-to-end suites, Gradle Cache Acceptance, Gettext, Credo, Security, migrations, and the targeted billing tests are green. GitHub reports 33 passing checks and no failures or pending checks.

Could you take another look?

fortmarek:
> This keeps the traffic view available only for controlled billing validation; it should not be enabled broadly until the price source and dollars-focused experience are modeled.

Why don't we set _some_ price, so the page can actually reflect the $ usage amount instead of shipping a page in a shape that we don't want to have long-term? Then shipping billing will be just about flipping the flag and setting the right amount.

pepicrft:
@fortmarek Agreed. I picked a provisional price of $0.10 per gigabyte and reshaped the page around cost in `848601059f`.

The Billing Usage page now shows the previous cache bill, cost to date, and an estimated next cache bill projected through the current billing period. The graph is cumulative cost rather than bytes, and the page makes the active unit price visible. Private runner traffic remains excluded from every amount.

The price lives in the existing `Tuist.Billing` unit price map, which is already the source that must stay aligned with Stripe. Changing the launch price is therefore one code value plus the matching Stripe price, while enabling the account remains a feature flag change.

I validated this with 46 focused billing and LiveView tests, strict Credo analysis on the changed Elixir files, Gettext extraction, and a full asset build. I could not attach a browser screenshot because the in-app browser control was unavailable in this session.

PR #11848, opened 2026-07-14
feat(server): add account token dashboard

fortmarek:
## What changed

- Added a dashboard Tokens settings page for both personal and organization accounts.
- Added listing, creation, scoped project restrictions, expiration parsing, one-time token reveal, copy support, and revoke actions for account tokens.
- Added a dedicated account-token detail route at `/settings/tokens/:token_id` for metadata, effective permissions, project access, and revocation.
- Added a masked token hint that shows the stable token prefix plus the beginning of the token ID, then masks the rest.
- Persisted optional `account_tokens.token_last_four` metadata for newly issued tokens without storing the token value.
- Seeded development personal and organization accounts with existing account tokens, including an organization token restricted to specific projects.
- Added the Tokens settings tab, updated dashboard gettext templates, and documented the new stored token hint in `server/data-export.md`.
- Capped the create-token modal to the viewport, made its body scrollable, kept closed modals hidden, and replaced typed project handles with a project checklist that defaults to all selected.
- Refined the create-token modal with project-name-only rows, no MCP preset, Account/Project scope select-all rows, fixed header/footer dividers around the scrollable body, no select-all hover highlight, and clearer spacing between scope cards.
- Updated scope selection so presets check their effective permission rows, write scopes imply the matching read scope, and detail pages show those effective read permissions too.

## Why

Account tokens could be managed through the CLI, but the dashboard had no equivalent workflow. Rotating tokens, reviewing existing tokens, and managing personal or organization account credentials should be possible from the same account settings surface where other account-level access is managed.

The dashboard also needs to help users identify an existing token without exposing the secret. The plaintext token is still shown only once and is never persisted. The list/detail hint now uses the known generated-token prefix and token ID prefix, then masks the rest of the value.

The create flow should also keep permission state honest. If a preset grants a set of permissions, those permission rows should visibly reflect that. If a write permission is selected, the corresponding read permission should be selected as well because write-only access is not meaningful in this permission model.

## Approach

- Reused the existing `Accounts.create_account_token/2`, authorization policies, and account-token scope model instead of adding a parallel token flow.
- Added a LiveView settings page under account settings so the same UI works for personal and organization accounts.
- Kept the tokens index as a compact list and made each whole table row navigate to a full detail page. The detail page owns expanded metadata, effective permissions, project restrictions, and revocation.
- Added a scoped `Accounts.get_account_token/3` lookup so detail pages load by token ID without exposing tokens across accounts.
- Used Noora `card` + `card_section` components for the detail content and Noora tables for permissions and project access instead of custom card-like containers or bespoke detail rows.
- Expanded preset scopes, such as `ci`, through the same authorization helper used at runtime so the detail table shows the effective permissions rather than the stored preset shorthand.
- Loaded account projects into the create-token modal and replaced manual project-handle entry with a `Select all projects` checkbox plus individual project checkboxes. Keeping every project selected creates an all-projects token; selecting only some projects creates a project-restricted token.
- Added Account and Project select-all scope rows using the same indeterminate checkbox pattern as webhook event groups. Presets remain explicit, and the MCP scope remains backend-valid but is no longer offered as a dashboard preset.
- Kept preset storage compact while deriving checkbox state from effective permissions. When a user customizes a preset-derived permission row, the modal expands the preset into explicit fine-grained scopes and then applies the edit.
- Normalized custom write selections with their corresponding read scope before saving. Presets still save as presets unless the user customizes them.
- Persisted `token_last_four` on new account and SCIM tokens through the existing account-token schema.
- Kept create-token modal behavior local to the account-token page CSS: the content becomes flex only while `data-state="open"`, preserving Noora/Zag closed-state hiding, and fixed header/footer dividers are rendered outside the scrollable body.

## Bugfix details

During local QA and browser review, the first dashboard pass had a few visible issues:

- The create-token modal could exceed the viewport when the scope list was long, leaving actions hard to reach. The modal content now has a viewport-based max height and a scrollable body while the header/footer stay visible.
- A first pass applied `display: flex` directly to modal content, which could override the closed hidden presentation. The flex display now applies only when the modal content has `data-state="open"`.
- The modal's top and bottom dividers originally lived inside the scrollable content. They are now fixed as the header/footer boundary while only the body scrolls.
- The token table attempted to fit name, token hint, scopes, projects, expiration, last-used, created, and actions into one row. At a 1016px-wide viewport, several cells truncated. The table now shows only high-signal fields.
- The generate action appeared in its own empty card section. It now lives in the Account tokens card header.
- The token hint exposed too much of the generated identifier and then showed the trailing characters. It now shows the initial token ID prefix with dot-style masking and no trailing token characters.
- The first detail pass was bolted onto the list page. It is now a full detail route with its own page layout and back navigation.
- Only the token name was clickable and its cell was visually misaligned. The table now uses Noora's whole-row navigation and the standard text cell alignment.
- The token detail route initially kept the Settings title and tabs, used custom card styling, and made Revoke destructive. The route now omits the settings tab chrome, uses Noora cards/sections, and renders Revoke as a secondary action.
- The token detail permissions/project access areas initially used custom list rows inside the cards. They now use Noora tables and built-in table cells, which also fixed the numeric project-row DOM ID crash found by the LiveView test.
- The permissions table initially showed preset groups as a Category column. It now drops that column and expands `ci` into the actual effective write permissions.
- The create-token form initially required manually typing project handles, then briefly used split all/specific buttons. It now uses the existing select-all checkbox pattern with project checkboxes, defaults to all projects selected, shows only project names because the account prefix is implicit, and validates that at least one project remains selected when restricting access.
- Account and Project scope groups previously required checking every fine-grained permission manually. They now have group-level select-all rows.
- The MCP preset appeared in the dashboard create flow even though it is not relevant to normal token creation. It has been removed from the dashboard picker while keeping backend support intact.
- Preset selection previously left the underlying permission rows visually unchecked. Presets now check their effective permission rows while still saving compactly.
- Selecting a write scope previously allowed a write-only token to be created from the dashboard. Custom write selections now automatically include the corresponding read scope.

## Developer impact

This brings account-token dashboard behavior to parity with the CLI capabilities while keeping the existing account-token model and authorization boundaries. The development seeds now make it easier to inspect both unrestricted and project-restricted token states locally.

## Validation

- `mise exec -- mix ecto.migrate`
- `mise exec -- mix format lib/tuist/accounts.ex lib/tuist_web/live/account_token_helpers.ex lib/tuist_web/live/account_tokens_live.ex lib/tuist_web/live/account_tokens_live.html.heex lib/tuist_web/live/account_token_live.ex lib/tuist_web/live/account_token_live.html.heex test/tuist/accounts_test.exs test/tuist_web/live/account_tokens_live_test.exs priv/repo/seeds.exs`
- `mise exec -- mix format lib/tuist_web/live/account_token_helpers.ex lib/tuist_web/live/account_tokens_live.ex lib/tuist_web/live/account_tokens_live.html.heex test/tuist_web/live/account_tokens_live_test.exs`
- `mise exec -- mix gettext.extract`
- `mise exec -- mix test test/tuist_web/live/account_tokens_live_test.exs`
- `mise exec -- mix test test/tuist/accounts_test.exs:3104 test/tuist/accounts_test.exs:3194 test/tuist/accounts_test.exs:3247 test/tuist/accounts/account_token_test.exs test/tuist/scim_test.exs`
- `mise exec -- mix assets.build`
- `git diff --check`
- Confirmed no `.po` translation files changed.
- Verified locally in the browser at `/tuist/settings/tokens`: compact index, whole-row navigation, no inline detail panel, dedicated detail page without settings tabs, Noora detail cards/sections, permissions and project access rendered as Noora tables, secondary revoke action, leading-prefix bullet-masked token hints, aligned name cell, project-restricted seed token rendering, expanded `ci` effective permission rows with no Category column, select-all project checklist in the create-token modal, project-name-only rows, no all/specific buttons, no project-handle textarea, no MCP preset, Account/Project scope select-all rows, fixed header/footer dividers around the scrollable modal body, no select-all hover highlight, extra spacing between scope cards, CI preset rows visibly checked by default, and write scope clicks selecting their matching read scope.


server/lib/tuist_web/live/account_tokens_live.ex
```diff
@@ -0,0 +1,465 @@
...
+        order_by: [:name],
+        order_directions: [:asc],
+        page: 1,
+        page_size: 500
+      })
+
+    projects
+  end
+
+  defp ensure_can_create(%{assigns: %{can_create_tokens?: true}}), do: :ok
```
pepicrft: Please re-authorize immediately before changing account credentials. `can_create_tokens?` is evaluated only during mount, and `ensure_can_create/1` trusts that cached value for the lifetime of the LiveView socket.

I reproduced this locally by opening the page as an organization administrator, demoting that user to a regular member, and then submitting the create event. A fresh authorization check correctly returned forbidden, but this handler still created the token. The token detail page has the same issue for revocation through its cached `can_delete_tokens?` value.

Please call `Authorization.authorize/3` inside both mutation events against the current account instead of relying on mount-time authorization state.

server/lib/tuist_web/live/account_tokens_live.ex
```diff
@@ -0,0 +1,465 @@
...
+  end
+
+  defp list_account_projects(account) do
+    {projects, _meta} =
+      Projects.list_projects(%{
+        filters: [%{field: :account_id, op: :==, value: account.id}],
+        order_by: [:name],
+        order_directions: [:asc],
+        page: 1,
+        page_size: 500
```
pepicrft: This silently widens project access for accounts with more than 500 projects. The query loads only the first 500 projects, initializes `selected_project_ids` from that partial result, and later treats every loaded project being selected as `all_projects: true`.

I validated that an account with 501 projects renders only 500 individual project rows. Leaving the defaults selected therefore creates an unrestricted token that also covers the omitted project and future projects. Please load every project, paginate the selection without losing state, or represent unrestricted access as an explicit choice instead of inferring it from a partial page.

pepicrft (commented):
I found two security-sensitive issues that should be addressed before merging. Both are called out inline.

I also confirmed that the existing `server/data-export.md` finding about the nonexistent stored `token_last_four` field remains valid. The current dashboard hint is derived from the token identifier prefix, and there is still no schema field or migration for that suffix.

Local validation:

- The focused account-token, account, authorization, and LiveView tests passed: 367 passed, 2 excluded.
- Two temporary reproduction tests passed for creating and revoking tokens after an organization administrator is demoted. The temporary tests were removed after validation.
- The repository security task passed. The vulnerability and secret scan was clean for the scanned manifests, and the source analyzer only reported existing baseline locations.
- Focused formatting passed.
- Strict Credo reports one new readability failure at `account_tokens_live.ex:84` because the line is 121 characters. The other reported warnings and suggestions are pre-existing.
- The fully seeded application started locally, authentication succeeded, and `/tuist/settings/tokens` returned a successful response with the account-token creation surface.
- I could not complete screenshot verification because the in-app browser control was unavailable in this session.

server/lib/tuist_web/live/account_tokens_live.ex
fortmarek: Addressed in `AccountTokensLive` and `AccountTokenLive`. Create and revoke now call `Authorization.authorize/3` at mutation time instead of trusting the mount-time `can_create_tokens?` / `can_delete_tokens?` assigns. I added LiveView coverage for demoting an organization admin after mount and verified the create/revoke mutations no longer succeed.

server/lib/tuist_web/live/account_token_helpers.ex
```diff
@@ -0,0 +1,227 @@
...
+          },
+          %{
+            scope: "account:registry:write",
+            label: dgettext("dashboard_account", "Registry write"),
+            description: dgettext("dashboard_account", "Publish packages to the account registry.")
+          },
+          %{
+            scope: "account:scim:write",
+            label: dgettext("dashboard_account", "SCIM write"),
+            description: dgettext("dashboard_account", "Provision organization members through SCIM.")
```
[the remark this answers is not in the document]
fortmarek: Addressed. SCIM is no longer exposed in the generic account-token dashboard picker, and generic account tokens continue to use only the user-creatable scope set. SCIM tokens stay on the dedicated SCIM token path/format instead of being created through `Accounts.create_account_token/1`.

server/data-export.md
```diff
@@ -19,7 +19,7 @@ Sensitive authentication data (passwords, tokens) are excluded from exports.
 - Organization memberships and roles (user, organization, role, and timestamps)
 - Account billing information and subscriptions
 - Account-level cache settings, including storage region, custom cache endpoint enablement, and cache write policy (`members_and_tokens` or `tokens_only`)
-- API tokens, SCIM-scoped account tokens, and project tokens (existence, scopes, names, timestamps, and last-used metadata only; token values and hashes are excluded)
+- API tokens, SCIM-scoped account tokens, and project tokens (existence, scopes, names, timestamps, last-used metadata, and masked-preview hints such as the stored account-token `token_last_four` suffix only; token values and hashes are excluded)
```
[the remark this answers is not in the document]
fortmarek: Addressed in `server/data-export.md`. The export documentation no longer claims a stored `token_last_four` account-token suffix and now documents only persisted token metadata while explicitly excluding token values, hashes, and generated secret previews.

server/lib/tuist_web/live/account_tokens_live.ex
fortmarek: Addressed in `AccountTokensLive`. Project access now recursively loads every project page using the existing Flop metadata instead of only page 1, so the default selected-project state is complete before it can be interpreted as `all_projects: true`. I added a LiveView test with 501 projects to verify a project beyond the first page appears in the selector.

pepicrft (approved):
I revalidated the review fixes locally. The focused account token LiveView suite passes, and the authorization and project pagination regressions are covered.

PR #11851, opened 2026-07-15
ci(server): smoke test PromEx metrics endpoint

pepicrft:
## What changed

The existing server image build now runs a production-release smoke test that starts the packaged Plug Cowboy and PromEx applications, starts the real `Tuist.PromEx` supervisor, requests `/metrics`, requires a successful response, and stops the supervisor cleanly.

The existing Plug Cowboy declaration is also placed next to PromEx with a comment explaining why it remains a direct dependency even though the Phoenix endpoint runs on Bandit.

## Why

Marek's #11850 reverted #11777 and restored the dependencies that unblock deployment. The incident still exposed a gap in the pull request checks: compilation and production image assembly both succeeded after the runtime adapter used by PromEx had been removed.

The packaged image should prove that its standalone metrics listener can actually start and serve metrics before it can be deployed.

## Root cause

#11777 removed Plug Cowboy under the assumption that Bandit had replaced it completely. Bandit serves the Phoenix endpoint, but the standalone PromEx metrics listener still invokes `Plug.Cowboy.child_spec/1`.

Because that path runs only when metrics are enabled at runtime, compilation and image assembly did not detect the missing dependency.

## Approach

Reuse the image already loaded by the Docker build job and run it with the production deployment environment, hosted mode, web mode, and Prometheus metrics enabled. The smoke test starts the actual `Tuist.PromEx` module and verifies its listener by requesting `/metrics` on port 9091.

This catches both missing packaged dependencies and failures in the real PromEx configuration without booting the full application, which would require PostgreSQL, ClickHouse, object storage, and license validation.

## Impact

There is no runtime dependency change relative to `main`; #11850 already restored Plug Cowboy. Future changes that break the standalone metrics listener will fail the server image check before deployment. The smoke test reuses the existing image and adds only its startup and request time.

## Validation

- Parsed `.github/workflows/server.yml` successfully.
- Ran the smoke test against a locally available packaged server image and received a successful `/metrics` response.
- The `Smoke test PromEx metrics server` GitHub Actions step passed against the image built by this pull request.


.github/workflows/server.yml
```diff
@@ -407,6 +407,46 @@ jobs:
             exit "$status"
           fi
           echo "Docker build completed"
+      - name: Smoke test production server PromEx metrics
```
pepicrft: @tuist/engineering ended up smoke-testing this particular one cause the full production-image boot would also require setting up other dependencies and a license to make it work.

fortmarek: Could we boot the built image and wait for `/ready` or `:9091/metrics` instead? This check defines a second PromEx module with empty plugins, so it does not exercise `Tuist.PromEx`, the default container command, or the actual supervision tree. A real image boot would have caught the missing `Plug.Cowboy` dependency and future release-only boot regressions through the same path production uses. If wiring PostgreSQL, ClickHouse, storage, and a CI license is too expensive for this job, I would prefer reducing this to a direct assertion that `:plug_cowboy` is present and startable in the release rather than maintaining a synthetic PromEx configuration here.

.github/workflows/server.yml
```diff
@@ -407,6 +407,21 @@ jobs:
...
+            --env TUIST_DEPLOY_ENV=prod \
+            --env TUIST_HOSTED=1 \
+            --env TUIST_OBJECT_STORAGE_PROVIDER=azure_blob \
+            tuist/tuist \
+            /app/bin/tuist eval '
+              :web = Tuist.Environment.mode()
+              :prod = Tuist.Environment.env()
+              {:ok, _} = Application.ensure_all_started(:plug_cowboy)
+              true = Code.ensure_loaded?(Plug.Cowboy)
+              true = function_exported?(Plug.Cowboy, :child_spec, 1)
```
fortmarek: Could we make this prove that the packaged metrics component actually runs? These assertions establish that `plug_cowboy` is loadable and exports `child_spec/1`, but they never invoke the PromEx path, bind port 9091, or request `/metrics`.

A broader check can still avoid the full application dependencies by starting the real packaged component and scraping it:

```elixir
{:ok, _} = Application.ensure_all_started(:plug_cowboy)
{:ok, _} = Application.ensure_all_started(:prom_ex)
{:ok, supervisor} = Tuist.PromEx.start_link([])
{:ok, _} = Application.ensure_all_started(:inets)
{:ok, {{_, 200, _}, _, _}} = :httpc.request(~c"http://127.0.0.1:9091/metrics")
:ok = Supervisor.stop(supervisor)
```

That exercises the production `Tuist.PromEx` configuration, plugin/storage initialization, the actual `Plug.Cowboy.child_spec/1` call, listener startup, and a real response without requiring PostgreSQL, ClickHouse, storage, or license validation. This is non-blocking because the current check still catches the original missing-dependency regression.

fortmarek (approved):
The dependency restoration and packaged-release check correctly cover the original missing Plug Cowboy regression. I left one non-blocking suggestion to exercise the actual PromEx listener and `/metrics` endpoint for broader runtime confidence.

PR #11857, opened 2026-07-15
feat(cli): persist the Xcode CAS across runner VMs via a per-account disk image

fortmarek:
> Stacks on #11996 (fast-forward last-writer-wins for the runner cache volume). Composes with #11788 (account cache volumes) and #11853 (persistent CAS on developer machines). Off by default.

## Purpose

Persist the **Xcode compilation cache (CAS)** across ephemeral macOS runner VMs, so a from-scratch build on a fresh VM resolves against a warm store instead of re-ingesting the whole action-closure every job.

The CAS is **folded into the per-account cache image**: a `CompilationCache.noindex` store directory beside `tuist/` inside the one image #11996 already clones master↔branch, promotes, and converges across hosts. So the CAS rides that whole machinery — CoW clone, fast-forward-LWW promote, generation compare-and-swap, cross-host HEAD, convergence — with no image, mount, promote gate, or eviction path of its own.

## Root cause: why the CAS needs a block device, not the bare share

llcas's on-disk store (`UnifiedOnDiskCache`) uses memory-mapped file locking, and `mmap` (MAP_SHARED) over Apple's Virtualization.framework virtio-fs faults with **SIGBUS**. A CAS pointed straight at the branch share crashes the Xcode build system at `ValidateCAS` (`unexpected service error: The Xcode build system has crashed`).

Verified directly on a staging Mac mini, isolated from the build system:

| store | `llvm-cas --check-lock-files` |
|---|---|
| local disk | rc=0 |
| virtio-fs share | **rc=138 (SIGBUS)**, reproduced on 2 stores |

Trivial single-shot `--ingest` / `--cat` / `--validate` deceptively succeed on virtio-fs — only the locking path a real concurrent build exercises faults, so a naive smoke test would wrongly conclude it works. The space in "My Shared Files" was ruled out (local-disk-with-space passes; virtio-fs via a space-free symlink still crashes).

The cache image sidesteps this: the guest `hdiutil attach`es it off the share and gets a real block-device APFS volume whose block layer does plain read/write to the backing file (no mmap of the shared file), so llcas works. The CAS store lives **inside** that image, so it inherits the block device for free — which is why folding is possible at all.

## The approach: fold the CAS into the cache image

When a job is dispatched, the host clonefiles the account's master image into the VM's branch and the guest attaches it at `/Users/runner/.tuist-cache-volume`. The store is just a subdir of that mount:

- `<mount>/tuist/` — the binary cache → `TUIST_XDG_CACHE_HOME`
- `<mount>/CompilationCache.noindex/` — the CAS → `COMPILATION_CACHE_CAS_PATH`

`setup_cas_store` (called from `attach_cache_image`, gated on a host-staged `cas-enabled` marker) creates the store, writes an xcconfig, and exports `XCODE_XCCONFIG_FILE`. There is no separate attach, detach, promote, or success gate: the CAS rides the binary cache's clone/promote/converge and its quiesced-detach guard.

## Why fold (and what it costs)

The CAS *was* a second sibling image with its own runner-success promote gate, host-local only. Folding it into #11996's image is a net **−602 lines** and buys real correctness: the CAS inherits #11996's **generation compare-and-swap** (a stale, slow job can't clobber a fresher master — refuse-to-regress) and its **cross-host HEAD + convergence**, neither of which the separate host-local CAS had. It also dissolves three prior review findings (the detach-promote gate, CAS-only-master eviction, and the concurrency LWW race are now the binary cache's, which already handle them).

The one real cost is upload cadence. #11996 promotes-and-uploads the whole image on the `dirty` gate at guest teardown, so with the CAS in the image, a compile-only job's CAS growth now rides that path (see the digest below), and the upload includes CAS bytes. We accept this and keep it honest two ways: **keep the volume sized reasonably** so the upload stays fast, and **measure it** — the guest times the teardown HEAD PUT (which blocks VM halt and slot reclaim) and the host records it as `tart_kubelet_cache_volume_upload_seconds`. If that histogram's tail grows, the volume cap is the lever.

## The digest: how a compile-only job still persists its CAS

#11996's `Finalize` promotes only when the inventory digest changed (`dirty`). A compile-only job that gets binary-cache *hits* but compiles new code changes only the CAS — and llcas grows partly by **appending to fixed files**, so an entry-name digest wouldn't notice. So the inventory digest now includes a **`~cas.bytes/<N>` line** — the store's total logical size. CAS growth flips the digest → dirty → the whole image promotes.

Size is a coarse *content proxy* (two different CAS states of equal total size collide → at worst an occasional miss, fine for a cache). Because this digest is what a converging host uses to verify a downloaded master, it must be **byte-identical** host- and guest-side: the host sums regular-file `st_size` under the store (`dirLogicalSize`), the guest does `find … -type f -exec stat -f %z`, and the `~` prefix sorts the line last under `LC_ALL=C` on both. This is convergence-critical (a bug would break the *binary* cache too), so it is covered by `TestInventoryDigestMatchesGuestScript` (byte-for-byte) and the real-`hdiutil` `TestDarwinImageInventoryDigest`.

## What changed

- **tart-kubelet VolumeManager** (`internal/podagent/volume.go`): removed the separate-CAS machinery (`casImageName`, `masterCASImage`/`branchCASImage`, `materializeCASImageLocked`, `FinalizeCAS`, `promoteCASImageLocked`, `casReserveBytes`); `allMastersLocked`/admission are back to the single image. `inventoryDigest` takes the mount root and appends the `~cas.bytes` line via `dirLogicalSize`. `casEnabled` now only gates a `cas-enabled` status marker.
- **Runner image** (`dispatch-poll.sh`): removed `attach_cas_image` / `detach_cas_image` / `report_runner_ok`. `setup_cas_store` points the compiler at `<mount>/CompilationCache.noindex` via `XCODE_XCCONFIG_FILE` — an overrides layer that every `xcodebuild` invocation honors (including a plain `xcodebuild build` against a project Tuist never generated). It sets `COMPILATION_CACHE_CAS_PATH` / `KEEP_CAS_DIRECTORY=YES` / a host-staged `LIMIT_PERCENT` (the CAS's coordinated share of the image — see [Shared budget](#reviewer-notes)), `#include`s a workflow's own xcconfig **last** (clean opt-out), and deliberately does **not** set `COMPILATION_CACHE_ENABLE_CACHING`. `cache_inventory` gained the matching CAS size line; `report_volume_head` times the upload. When the feature is turned off, `reclaim_cas_if_disabled` drops the stale store from the image at teardown so masters stop cloning/uploading dead CAS bytes.
- **CLI** (`XcodeCacheSettingsProjectMapper`): on **Xcode 27+**, enables compilation-cache path prefix mapping — see [DerivedData sensitivity](#deriveddata-path-sensitivity). (The runner CAS *path* is injected via xcconfig above, not this mapper.)
- **main.go**: `--cache-volume-cas-gib` is now the CAS's byte **budget within** the shared cache image (there's no separate image to size): it sets the CAS's share of `--cache-volume-cap-gib`, and the binary cache gets the rest minus reserve. `0` = off.
- **metrics.go**: `tart_kubelet_cache_volume_upload_seconds` (upload duration) and `tart_kubelet_cache_volume_fill_percent` (post-job image fill %, the ENOSPC-pressure signal).

## Security: cross-account defenses

Folding removes the separate guest-writable CAS image, and with it the CAS-specific symlink-swap surface. The CAS now rides the binary cache's #11996 defenses: the branch image promotes only when `SourceAccount == account` (the defense-in-depth half of #11788's cross-account guard) and only on an accepted generation fast-forward; a failed run or torn/undetached image is discarded, not promoted.

## DerivedData path sensitivity

Xcode's compilation-cache keys are **not** sensitive to the CAS location — measured, 4 builds changing one variable at a time (Xcode 26.5 / Swift 6.3.2, 40-file package):

| CAS path | DerivedData path | result |
|---|---|---|
| casA (cold) | dd | 0 hit / 28 miss — populates |
| casA | dd | **28 hit / 0 miss** — baseline |
| **casB** (store copied elsewhere) | dd | **28 hit / 0 miss** |
| casA | **dd2** | **4 hit / 24 miss** |

Moving the store is free (row 3) — exactly what this does when it clonefiles a master into a branch and mounts it at a fixed path. But the **DerivedData path IS in the key** (row 4): it leaks in via DD-derived frontend flags (`-module-cache-path`, `-I <dd>/Build/Products/...`, `swift-overrides.hmap`, …).

**Consequence:** on Xcode ≤ 26 (our current fleet), warm hits need the DerivedData path stable *job to job on a runner* — which holds for a normal workflow (deterministic checkout, fixed `-derivedDataPath`); a workflow randomizing DD per job reduces the hit rate.

**This PR fixes it for Xcode 27+.** Swift 6.4 / Xcode 27 is the first build system to implement source/build directory prefix mappings (`$(PROJECT_DIR)=/^src`, `$(PROJECT_TEMP_DIR)=/^derived`, `$(BUILT_PRODUCTS_DIR)=/^built`, `$(WORKSPACE_DIR)=/^workspace`; upstream `swiftlang/swift-build@c0d1ca4`, gated behind `[CLANG|SWIFT]_ENABLE_[PROJECT_]PREFIX_MAPPING` in `@015bbff`). Those settings have **no `DefaultValue`** — off until set — so Xcode 27 alone changes nothing; the mapper turns them on when the selected Xcode is ≥ 27. Caveats: enabling prefix mapping **changes every cache key**, but only at the 26→27 boundary where the compiler version already invalidated the cache, so no *extra* rebuild; and local↔CI sharing needs both sides on 27 with mapping on. Xcode 26.5 has neither the settings nor the mappings (verified against its `Swift.xcspec`), and enabling the flags by hand on 6.3.2 does not recover the hits — no workaround before 27, just the DD-stability requirement.

## Validation

- **The warm-hit benefit and SIGBUS root cause were established on a staging Scaleway Mac mini** (macOS 26.3.2, arm64): a build pointed at a block-device CAS store via `COMPILATION_CACHE_CAS_PATH` (exactly what `setup_cas_store` emits) goes **cold 77 miss → warm 77 hits / 0 miss, both BUILD SUCCEEDED** (~6.9s warm vs ~21s cold; sparse store 188 MB on disk, not its logical cap). The build is indifferent to whether that store is its own image or a subdir of the cache image — both are block-device APFS — so the number carries to the folded design.
- **Host-side**: full tart-kubelet suite green under `-race`, including `TestInventoryDigestMatchesGuestScript` (byte-for-byte host↔guest digest, now with the CAS size line) and the real-`hdiutil` `TestDarwinImageInventoryDigest`. `go build ./...`, `vet`, `gofmt` clean; darwin + linux build.
- **Guest**: `bash -n` + `shellcheck -S error` on `dispatch-poll.sh`.
- **Remaining**: a full folded-path run on the staging dispatch loop (label stamp → materialize → promote → converge → next job) with the daemon deployed.

## Reviewer notes

- **Off by default.** `--cache-volume-cas-gib` defaults to 0: no `cas-enabled` marker, the compilation cache stays VM-local, the binary cache is untouched.
- **Digest scheme change, self-healing.** Including the CAS size changes the inventory digest for *all* masters (binary-only get `~cas.bytes/0`), so any existing HEAD digest-mismatches once and re-uploads — one cold round per account. #11996 isn't in production, so there is likely nothing to migrate.
- **Shared budget (operator-tuned).** The CAS and binary cache share `--cache-volume-cap-gib`, so their two independent pruners (the CLI LRU via `TUIST_CACHE_MAX_BYTES`, llcas via `COMPILATION_CACHE_LIMIT_PERCENT`) get a **coordinated split** the host derives from `--cache-volume-cas-gib`: the CAS gets that budget's share (`LIMIT_PERCENT ≈ cas-gib/cap-gib`), the binary cache gets the rest, and a filesystem reserve of `max(2 GiB, 5%)` stays free — so they can't both fill the one image to ENOSPC. The reserve is `max(absolute, proportional)`, not a flat percent, because APFS headroom and single-build pruner overshoot scale absolutely (floor binds below a 40 GiB cap, percent above). One `cacheImageSplit` computes both halves so they can't drift; an oversized `cas-gib` is clamped so the binary cache keeps a ≥10% slice. With the CAS off the binary cache gets the full ~80%. `tart_kubelet_cache_volume_fill_percent` (post-job fill, sampled at teardown) is the signal for whether the reserve/split needs tuning — its tail near 100 is real pressure.
- **Intentionally not in this PR:** config plumbing of the flags through CAPI / helm / host-bootstrap; idle/cadence-gated cross-host upload (a follow-up that would let the CAS distribute on its own cadence rather than the binary cache's dirty cadence); real large-project perf.

### How to test locally

Host-side logic runs on Linux CI with the injected fake backend:

```
cd infra/tart-kubelet && go test ./internal/podagent/
```

The host↔guest digest consistency (the convergence-critical part) and, on a real Mac, the `hdiutil` round-trip:

```
cd infra/tart-kubelet
go test ./internal/podagent/ -run 'InventoryDigest|DarwinImageInventoryDigest' -v
```


fortmarek:
## End-to-end staging validation: a real cache build uses the volume

Ran the full chain on a staging Scaleway Mac mini (macOS 26.3.2 host, Xcode 26.0.1 guest), with the real host + guest code, and confirmed a real compilation-cache build's CAS store physically lives on the volume.

**Job 1 — attach + build + prove on-volume**
- Real `attach_cas_image` (sourced from the committed `dispatch-poll.sh`) mounted the CAS image as a **block device**: `/dev/disk9s1 on /Users/…/xcode-cas (apfs)`, distinct from the VM root `/dev/disk3s1s1`, and exported `TUIST_COMPILATION_CACHE_DIR`.
- A cache-enabled `xcodebuild` ran with `COMPILATION_CACHE_CAS_PATH=<dir>/tuist_tuist` — the **exact string the mapper emits** for handle `tuist/tuist` (locked by the new unit test).
- Proof it used the volume: the store (`builtin`/`generic`, 33 MB) populated at `<mount>/tuist_tuist`, and `df` on that path reports the **disk-image device `/dev/disk9s1`, not the VM root** — with **no CAS in DerivedData**. `report_runner_ok 0` wrote the `runner-ok` marker; real `detach_cas_image` unmounted.

**Host promote/materialize** — real `cp -c`: promote (branch→master) 0.025 s, materialize (master→branch clone) 0.024 s, both CoW.

**Job 2 — warm reuse from the volume** — reattached the cloned image at the fixed mount, wiped DerivedData, rebuilt: **14 hits / 0 miss / BUILD SUCCEEDED in 6.1 s**, 100% of the CAS hits served from the volume.

### The one substitution (and why)
The CAS path was set to the mapper's exact output rather than produced by a CLI-built `tuist generate`, because building the `tuist` binary in this worktree (empty EE submodule, offline) is a known rabbit hole. That single link — the mapper emitting `COMPILATION_CACHE_CAS_PATH=<dir>/<handle>` — is instead locked by a new unit test (`map_whenRunnerCASDirSet_pointsCASAtTheVolumeMount`, runs in CI). So: the mapper's emission is proven by test, and a real build with that identical setting is proven on hardware to put the CAS on the volume with warm cross-VM reuse.

Still not exercised: a full fleet rollout (helm/CAPI wiring of `--cache-volume-cas-gib` + a runner-image rebake + a real GitHub dispatch), which remains the separate deploy step flagged in the PR.

infra/runner-image/dispatch-poll.sh
```diff
@@ -194,13 +204,22 @@ VOLUME_HEAD_UPLOAD_URL_MINT_ENDPOINT="${VOLUME_HEAD_REPORT_URL}/upload-url"
...
   [ -n "${CACHE_MOUNT}" ] || { echo "none"; return 0; }
   local root="${CACHE_MOUNT}/tuist"
+  local cas="${CACHE_MOUNT}/${CAS_STORE_DIR}"
+  # Total logical bytes of the folded CAS store. llcas grows by APPENDING to fixed
+  # files, so its entry names don't reliably change on growth — the size does — so
+  # a compile-only job (binary cache clean, CAS grown) still flips the digest and
+  # promotes. MUST match the host's inventoryDigest/dirLogicalSize EXACTLY: regular
+  # files only, logical st_size (stat -f %z), summed.
+  local cas_bytes
+  cas_bytes=$( { find "${cas}" -type f -exec stat -f %z {} + 2>/dev/null || true; } | awk '{s += $1} END {printf "%d", s + 0}' )
```
pepicrft: `cache_inventory` reduces the entire Xcode compilation cache to a logical-byte total. Once pruning starts replacing entries at a steady size, a real cache change can compare equal and never promote. More importantly, two concurrent branches with different contents but the same total produce the same tree digest and upload to the same supposedly immutable object key before the fast-forward decision. The rejected writer can therefore overwrite the accepted writer's image, and the host verification repeats the same size-only calculation, so it cannot detect the substitution. Please use a stable content identity for the compilation cache, or upload each attempt under a unique immutable key and bind only the accepted key.

infra/runner-image/dispatch-poll.sh
```diff
@@ -351,6 +372,85 @@ probe_cache_share() {
...
+  fi
+  rm -f "${store}/.writable" 2>/dev/null || true
+  {
+    printf 'COMPILATION_CACHE_CAS_PATH = %s\n' "${store}"
+    printf 'COMPILATION_CACHE_KEEP_CAS_DIRECTORY = YES\n'
+    # Bound the store to the host-computed share of the image so llcas prunes
+    # before the image can hit ENOSPC. This is coordinated with the binary cache's
+    # TUIST_CACHE_MAX_BYTES (each gets the same percent, the rest is reserve) so
+    # the two pruners cannot both fill the one shared image.
+    printf 'COMPILATION_CACHE_LIMIT_PERCENT = %s\n' "${cas_pct}"
```
pepicrft: `COMPILATION_CACHE_LIMIT_PERCENT` is not a percentage of the image capacity. Swift Build computes it from the current cache database size plus the filesystem's free space: https://github.com/swiftlang/swift-build/blob/5a49bfa5d4d7c4fbf1bea6e140481ba0818d676a/Sources/SWBTaskExecution/DynamicTaskSpecs/CompilationCachingDataPruner.swift#L289-L309

For the tested 20-gibibyte image with 8 gibibytes requested for compilation caching, the code emits 40 percent. Once the binary cache occupies its 10-gibibyte budget, the percentage denominator is approximately 10 gibibytes, so Xcode prunes toward roughly 4 gibibytes instead of 8. Please use `COMPILATION_CACHE_LIMIT_SIZE` with the exact byte budget.

infra/tart-kubelet/cmd/tart-kubelet/main.go
```diff
@@ -123,6 +124,12 @@ func main() {
 	flag.IntVar(&cacheVolumeCapGiB, "cache-volume-cap-gib", envIntOr("TART_KUBELET_CACHE_VOLUME_CAP_GIB", 20),
 		"Provisioned capacity (GiB) of each per-account cache master image. The image is sparse, so this is a "+
 			"ceiling, not an allocation; the runner-cache-root quota is the real aggregate bound.")
+	flag.IntVar(&cacheVolumeCASGiB, "cache-volume-cas-gib", envIntOr("TART_KUBELET_CACHE_VOLUME_CAS_GIB", 0),
```
pepicrft, line 127: This option defaults to zero, but the managed provisioning path never passes it. The host bootstrap only emits the cache root and total image capacity, while the provider and chart expose no compilation-cache budget field. Managed staging and production hosts therefore keep the feature disabled and never write the enablement marker. Please thread the budget through the chart, provider, bootstrap configuration, launch configuration, and their tests.

pepicrft (approved):
Approving the overall direction and implementation shape. I found three issues that should be addressed before merging: the compilation-cache identity can collide for different equal-size states, the configured byte budget is expressed using percentage semantics with a different denominator, and the managed fleet has no configuration path that enables the feature.

The focused Swift and Go tests pass, but these are cross-component behavior gaps that the current tests do not exercise.

infra/runner-image/dispatch-poll.sh
fortmarek: Replaced the total-size line with one `~cas/<relpath>\t<size>` line per store file — a real content identity (llcas names record files by their content hash; the append-mostly files are distinguished by size), so two branches with different contents can no longer produce the same digest/immutable key, and a steady-size change is still detected. The host `casInventoryLines` and the guest `find … | stat` are now cross-checked byte-for-byte by a new darwin test (`TestInventoryDigestMatchesGuestPipeline`) that runs the real shell pipeline — it immediately caught that BSD `stat` does not expand `\t` in its format string, so the separator is now a real tab built with `$(printf '\t')`.

infra/runner-image/dispatch-poll.sh
fortmarek: Switched to `COMPILATION_CACHE_LIMIT_SIZE` with the exact byte budget from `cacheImageSplit` (the `cas-enabled` marker now carries bytes, not a percent). An absolute bound is invariant to how full the binary cache is, so the CAS no longer prunes toward less than intended as the shared image fills.

infra/tart-kubelet/cmd/tart-kubelet/main.go
fortmarek: Threaded `--cache-volume-cas-gib` through the chart (`macosFleet.runnerCacheVolume.casGib` + the capi-scaleway template), the provider (flag → HostConfigHash input + reconciler), and the host bootstrap (Config field → launchd plist arg), with a plist-arg test and a HostConfigHash test so an omitted budget cannot make a roll look applied while the feature stays off. Defaulted off — actually enabling it in a managed env stays a separate rollout.

PR #11864, opened 2026-07-16
fix(server): resolve runner claim↔execution mismatch (attribution + reaping)

### LINKED ISSUE #11862: Runners are bound to their claimed job, not the one GitHub actually runs (claim↔execution mismatch)
## Summary

The server binds a runner to the workflow job it **claimed** — but GitHub assigns jobs to runners **independently** (by label) and may run a *different* job on that runner, or none at all. The server never learns the difference, so a runner is recorded as running a job it may not be running. This one **claim↔execution mismatch** surfaces as three distinct failures: **zombie runners**, **concurrency over-charging**, and **metrics mis-attribution**.

## Why the mismatch happens

1. On a dispatch poll, the server FIFO-claims a specific `workflow_job_id` and mints a JIT runner registered with the pool's **shared** labels (`runner_labels + dispatch_label`) in the org's default runner group. The claimed `workflow_job_id` is **never sent to GitHub** — nothing job-specific is.
2. **GitHub assigns queued jobs to *any* label-eligible runner** — the assignment is not bound to the server's claim. So the runner minted "for" job X is frequently *not* the one GitHub runs X on (it may run a sibling's job, or none).
3. There is **no `workflow_job.in_progress` handler** — only `queued`/`waiting`/`completed` are ingested (~33% of `workflow_job` traffic is dropped) — so the server never learns which runner GitHub actually assigned a job to, and can't correct the binding.

The runner↔job binding recorded at **claim/mint time** — across `runner_claims`, `runner_jobs.status` (→ `running` at JIT mint, *not* at job start), `runner_sessions`, and metrics — is therefore a **guess that GitHub does not honor**.

## The three symptoms

**1. Zombie / stuck-`running` runners.** A runner recorded as `running` (server-side and in Kubernetes) while executing **nothing** — for minutes to hours (observed: 6h; one macOS pod 8 days). It holds warm-pool capacity, counts as "busy" for autoscaling, and freezes the config it staged at claim time (e.g. `TUIST_CACHE_ENDPOINT`) — so if it later picks up a *different* job, that job runs with stale config (root of #11767).

**2. Concurrency over-charging.** The per-account concurrency limits (#11836) admit a claim only if its shape keeps the account within its vCPU/memory budget, counting **active claims** (both `claimed` and `running`, no lifecycle filter). A stuck/mis-attributed claim consumes that budget, so an account can hit **`:concurrency_limit_reached` while running fewer jobs than its limit allows** — real work throttled below capacity for phantom claims.

**3. Machine-metrics mis-attribution.** The metrics endpoint resolves a pod to its **claimed** `workflow_job_id` (`Claims.by_pod_name` → `JobMetrics.record`). If GitHub ran a *different* job on that pod, the CPU/memory chart on the *claimed* job's detail page shows the **other** job's run — and the job that actually ran there gets no metrics.

## Steps to reproduce (simulated)

1. Two jobs **J1** and **J2** queue for the same pool with identical `runs-on` labels → two `workflow_job.queued` webhooks; the server enqueues both.
2. Two warm runner pods **A** and **B** poll `/dispatch`. The server FIFO-claims **J1 → A**, **J2 → B**; both flip to `running` (JIT minted); both register with GitHub with **identical** labels.
3. A new push fires `concurrency: cancel-in-progress`, cancelling **J2** → one real job (J1), two registered runners.
4. GitHub assigns **J1 to pod B** (any eligible runner) → B runs J1 (say, a 45-min build).
5. **Pod A is stranded**: it registered, its claimed job J1 runs on B, no other job for it → runs nothing, yet its claim and `runner_jobs` row stay `running` for J1's whole duration. Pod A isn't cleaned up, even if the desired number of runners reduces to `0` because from kubernetes point-of-view it is still _running_.

> **This example uses only two jobs for clarity — real workflows are far larger.** A single workflow commonly fans out into **dozens** of jobs (matrix builds, many parallel jobs) that queue together with identical labels, so **many** runners register and **many** can be stranded **simultaneously**, each recorded as `running` while executing nothing. A stranded runner stays that way **until GitHub happens to assign it *some* job** (which it then runs — potentially with stale claim-time config), **or lingers idle-but-`running` for hours (observed up to 8 days) if none ever does**.

*(Proposed solutions — reaping and attribution — are in the comments below.)*

fortmarek:
Resolves https://github.com/tuist/tuist/issues/11862

> [!NOTE]
> Review surfaced a trust-boundary gap around fork PRs that this PR deliberately does **not** close. It pre-exists on `main` — capability issuance in `serve_claim/5` is untouched here, so nothing is introduced or widened — and fork PRs are not a pattern we expect our customers to use. Tracked separately; see [Open items](#open-items).

## What & why

The server binds a runner to the `workflow_job` it **claimed**, but GitHub assigns jobs to runners **independently** by label and may run a *different* job on that runner, or none. The server never learns the difference, so a runner is recorded as running a job it may not be running. This one claim↔execution mismatch surfaces as **zombie runners**, **concurrency over-charging**, and **metrics mis-attribution**.

The fix has two halves, per the issue's proposal comments: *attribution* (learn the binding from GitHub rather than guessing) and *reaping* (idle-TTL + gap-fill).

### Attribution — learn the runner→job binding from GitHub

`workflow_job.in_progress` was dropped at the webhook door (~33% of `workflow_job` traffic). It is the first event carrying the `runner_name` GitHub actually placed the job on — the only real-time proof of the binding.

- Ingest `in_progress`; record the executed `workflow_job_id` onto the claim and the durable session, keyed by `runner_name` and **scoped to the account resolved from the delivery's App installation**. New nullable `executed_workflow_job_id` on `runner_claims` / `runner_sessions`; NULL until GitHub proves execution — that absence is itself the "never assigned work" signal.
- Machine metrics resolve through the **proven** binding (`RunnerSessions.executed_job_for_pod/1`) and refuse to guess. Resolving via the claim would chart an idle runner's samples on a job running elsewhere, and would drop the real samples of a pod running a job it never claimed.
- The mismatch is emitted as webhook telemetry (`action=in_progress`, `outcome=matched|mismatch|unknown_runner`), making the problem measurable for the first time.

### Concurrency — release capacity by executor, not by claim

Completion released `where workflow_job_id == J` — the claim of whoever *claimed* J, not whoever *ran* it. With A claiming J1, B claiming J2 and GitHub running J1 on B, J2's completion freed B's slot **while B was still executing J1**, and A's idle timeout then freed J1's claim too — dropping the account to zero live reservations while a runner was working, admitting work above the limit.

- Completion now releases by executor (`Claims.complete_by_runner_name/2`, account-scoped). A job cancelled while queued carries no `runner_name`, so nothing is released; the claiming pod is still alive and may still be handed a sibling, and its slot is reclaimed when the pod stops or by `OrphanedRunnersWorker`.
- Job-keyed `Claims.complete/1` remains **only** for the recovery workers, where a GitHub-side terminal status has already proven the claim stale.
- `POST /pods/stopped` releases the claim held by a stopped pod, surfaced as `recovery` telemetry (`stranded_claim_released`).

### Reaping — idle watchdog

A pod's `tuist.dev/runner-pool-owner` label is stamped at **claim**, not execution, so a runner that registered but was handed no job reads as busy and lingers (observed up to 8 days). The reliable signal is in the pod: the runner fires `ACTIONS_RUNNER_HOOK_JOB_STARTED` the instant it gets work.

- The runner images supervise `./run.sh --jitconfig` and terminate it after `TUIST_RUNNER_IDLE_TIMEOUT_SECONDS` if no job arrives. The job-start signal is **latched irreversibly in the supervising process** — the hook cancels the watchdog outright (it runs before any workflow step) and the watchdog stands down permanently on first sighting. `/tmp` is workflow-writable, so a job deleting the marker must never be able to make itself look idle; the workflow can only ever make the watchdog give up earlier, never fire on a live runner.
- New `RunnerPoolSpec.IdleTimeoutSeconds` (CRD + chart, all four pool blocks including both Linux ones).

## Rollout

**The watchdog is enabled automatically**, not gated behind an operator override: `runnersFleet.idleTimeoutSeconds` and `runnersFleetLinux.idleTimeoutSeconds` default to **300s**, so every pool in every managed env gets it on deploy (verified: 13/13 pools per env render `idleTimeoutSeconds: 300`). Per-pool / per-shape / per-xcodeOverride values win; `0` disables.

300s bounds only the post-claim window between a runner registering and GitHub assigning it work (happy path is sub-30s). A warm pod still polling for a claim has no runner container yet and is unaffected, so this adds no warm-pool churn.

<a name="open-items"></a>
## Open items from review

- **Trust boundary (deferred, pre-existing).** Not introduced or widened by this PR. A trusted claim hands the runner `cache_signing_grant`, `volume_head`, and a trusted pod label that makes tart-kubelet materialize *and promote* the shared cache volume. A fork job landing on that runner inherits all three, breaking the fail-closed fork exclusion. This cannot be fixed by pinning the runner to the job — GitHub matches any job whose labels are a subset of the runner's, and fork PR jobs run in the base repo with identical `runs-on`. The correct fix is issuing capabilities only once the binding is proven (JOB_STARTED-hook callback writing `$GITHUB_ENV`, volume label flipped on proven trust). Deferred: fork PRs are not a pattern we expect customers to use. Note the grant cannot be fixed by scrubbing `$GITHUB_ENV` in the hook — steps inherit the runner process env, so a job can read `/proc/<runner_pid>/environ` regardless.
- **Durable rebind.** `runner_jobs` keeps the mint-time pod/runner/start, so on a mismatch the customer-facing job row and interactive-session target point at the wrong runner, and session `repository`/`workflow_name` stay stale.
- **Metrics loss before binding.** Samples posted before `in_progress` is processed are dropped. A retriable status does not help — both samplers are fail-open with no retry — so this needs a pod-keyed server buffer or sampler retention.

## Related

Runner stealing is an industry-wide consequence of static labels rather than a Tuist defect — RunsOn names it, and Depot's architecture (webhook → warm standby pool → ephemeral JIT runner → static label) is identical to ours. https://github.com/tuist/tuist/issues/11875 proposes opt-in run-scoped labels so customers who want deterministic assignment can avoid stealing entirely.

## Impact

Fixes metrics mis-attribution outright; makes zombies and over-charging both measurable and self-correcting. No customer-facing API change. Migration adds two nullable columns + two partial `runner_name` indexes.

## How to test locally

Server (from `server/`):

```
mix ecto.migrate
mix test test/tuist/runners/ \
  test/tuist_web/controllers/runner_pods_controller_test.exs \
  test/tuist_web/controllers/runner_job_metrics_controller_test.exs \
  test/tuist_web/controllers/webhooks/github_controller_test.exs
```

Runners-controller (from `infra/runners-controller/`):

```
go build ./... && go test ./...
```

Runner image scripts:

```
bash -n infra/linux-runner-image/run-job.sh
bash -n infra/runner-image/dispatch-poll.sh
```

Chart renders the field on every pool, both fleets:

```
helm template infra/helm/tuist -f infra/helm/tuist/values.yaml \
  --set runnersFleet.runnerImageSemver=9.9.9 --set runnersFleetLinux.shapeRunnerImage=img \
  --show-only templates/runner-pool.yaml | grep -c 'idleTimeoutSeconds: 300'
```

### Validation run

- 438 server tests green, including regression coverage for the executor-keyed release, cross-account `runner_name` collisions (both directions), execution-proven metrics, and the idle-runner-never-attributed case.
- Full server suite: 5679/5685; the 6 failures are pre-existing and environmental (uncompiled `ko` locale, blog content, redirect config) with no overlap with this diff.
- Full runners-controller Go suite green.
- Watchdog behaviour verified against the reported repro: firing the hook then deleting the marker + pid file leaves the runner at exit 0 (was 143); a runner never handed a job is still reaped at 143.
- `helm template` against the real managed values: 13/13 pools per env carry `idleTimeoutSeconds: 300` in production, canary and staging.


infra/helm/tuist/values.yaml
```diff
@@ -1365,6 +1365,23 @@ macosFleet:
...
+  # label-eligible runner, so a runner whose job landed on a sibling
+  # otherwise waits indefinitely, holding warm-pool capacity that reads
+  # as busy (its owner label is stamped at claim, not at execution).
+  #
+  # 300s only bounds the post-claim window between registering and
+  # being handed work (the happy path is sub-30s), and a warm Pod still
+  # polling for a claim has no runner container yet, so this adds no
+  # warm-pool churn. The value matches the stale-claim /
+  # orphan-recovery thresholds.
+  idleTimeoutSeconds: 300
```
pepicrft, line 1384: The pull request description says the watchdog is off by default and requires an operator override, but this value enables it at 300 seconds for every enabled macOS pool. The Linux default does the same. If rollout gating is intended, these defaults should be zero; otherwise the description needs to state that the watchdog is enabled automatically.

infra/linux-runner-image/run-job.sh
```diff
@@ -50,12 +50,56 @@ if [ -s "${CACHE_ENDPOINT_PATH}" ]; then
...
+./run.sh --jitconfig "${jit}" --disableupdate &
+runner_pid=$!
+
+# Forward pod-deletion SIGTERM to the runner so it deregisters cleanly.
+trap 'kill -TERM "${runner_pid}" 2>/dev/null || true' TERM INT
+
+if [ "${idle_timeout}" -gt 0 ] 2>/dev/null; then
+  (
+    sleep "${idle_timeout}"
+    if [ ! -e "${JOB_STARTED_MARKER}" ] && kill -0 "${runner_pid}" 2>/dev/null; then
```
pepicrft: The marker is writable by the workflow and is checked only after the full timeout. A long-running job that removes `/tmp/tuist-runner-job-started`, including a broad `/tmp` cleanup, is then treated as idle and terminated. I reproduced this by invoking the start hook, deleting the marker, and keeping the runner alive; the watchdog killed it with status 143. The job-start signal needs to be latched irreversibly in the supervising process, or the hook should cancel the watchdog immediately. The macOS script has the same pattern.

server/lib/tuist/runners/claims.ex
```diff
@@ -320,20 +320,77 @@ defmodule Tuist.Runners.Claims do
...
+  A job cancelled while still queued carries no `runner_name` (no
+  runner ever ran it); nothing is released, because the Pod that
+  claimed it is still alive and may still be handed a sibling job. Its
+  slot is reclaimed when that Pod stops (idle timeout / scale-down) or
+  by `OrphanedRunnersWorker`.
+
+  Idempotent. Returns the number of claims released.
+  """
+  def complete_by_runner_name(runner_name) when is_binary(runner_name) and runner_name != "" do
+    {count, _} = Repo.delete_all(from(c in Claim, where: c.runner_name == ^runner_name))
```
pepicrft: This deletion is global by `runner_name`, even though the webhook is authenticated to a specific GitHub App installation and therefore a specific account. Another organization controls its own self-hosted runner names, so a name collision can release another account's live claim. Please resolve the account from the installation and include `account_id` in this deletion and in both execution-attribution lookups.

server/lib/tuist/runners/dispatch.ex
```diff
@@ -196,27 +203,103 @@ defmodule Tuist.Runners.Dispatch do
...
+
+  # Binds the runner→job proof onto the live claim (drives metrics
+  # attribution while the job runs) and the durable session (the
+  # backstop that outlives the pod). The claim and session agree in
+  # the common case; when they disagree — or when only one is still
+  # present — we surface `:mismatch` so it's never silently the
+  # weaker `:matched`.
+  defp record_execution(runner_name, executed_workflow_job_id, source) do
+    claim_outcome = Claims.record_execution(runner_name, executed_workflow_job_id)
+    session_outcome = RunnerSessions.record_execution(runner_name, executed_workflow_job_id)
```
pepicrft: The real binding is written only to the PostgreSQL claim and session. `runner_jobs` was already marked running at mint time with the claimed pod and runner, so a mismatch leaves the customer-facing job row, its start time, and the interactive-session target pointing at the wrong runner. The session's repository and workflow fields also remain from the claimed job. Can we make this a single rebind operation that updates the job row and session metadata from the proven execution?

server/lib/tuist/runners/dispatch.ex
```diff
@@ -196,27 +203,103 @@ defmodule Tuist.Runners.Dispatch do
...
+    outcome = combine_attribution(claim_outcome, session_outcome)
+
+    case outcome do
+      :mismatch ->
+        Logger.info("runners: claim/execution mismatch (source=#{source})",
+          runner_name: runner_name,
+          workflow_job_id: executed_workflow_job_id
+        )
+
+        {:ok, :mismatch}
```
pepicrft, line 261: This detects the mismatch only after the runner has already received cache capabilities derived from the claimed job's `job_trusted?` result. If GitHub places an untrusted fork job on a runner minted from a trusted claim, that job inherits the account cache signing grant and, on macOS, the shared cache volume. Logging the mismatch cannot restore the fail-closed boundary once the job has started. We should avoid issuing job-specific shared cache capabilities from a speculative claim, or otherwise bind the runner to the exact job before exposing them.

server/lib/tuist_web/controllers/runner_job_metrics_controller.ex
```diff
@@ -75,7 +84,7 @@ defmodule TuistWeb.RunnerJobMetricsController do
   def create(conn, %{"pod_name" => pod_name} = params) when is_binary(pod_name) and pod_name != "" do
     with :ok <- RunnerPodAuth.authenticate(conn, pod_name),
          {:ok, samples} <- parse_samples(params) do
-      case Claims.by_pod_name(pod_name) do
+      case RunnerSessions.executed_job_for_pod(pod_name) do
```
pepicrft, line 88: This change makes samples posted before the `in_progress` worker records execution return a successful 204 response without being stored. Both samplers start immediately and do not resend successful posts, while webhook processing is asynchronous, so every job can lose its initial samples and a short job can lose the entire series. Please buffer samples by pod or return a retriable response and make the samplers retain the batch until the execution binding exists.

pepicrft (changes requested):
There are several correctness and trust-boundary gaps that need addressing before this can land. The most important one is that a runner can receive shared cache capabilities based on one claimed job and then execute a different, untrusted job. The inline comments also cover account scoping, durable rebinding, idle watchdog behavior, metrics loss, and rollout documentation.

infra/linux-runner-image/run-job.sh
fortmarek: Fixed in cf611bc. The job-start signal is now latched irreversibly in the supervising process: the hook cancels the watchdog outright (it runs before any workflow step, so job code never sees a live watchdog), and the watchdog polls and stands down permanently on first sighting instead of sampling the marker once at the deadline — so a later `/tmp` wipe is inert. Both scripts. Reproduced your case: hook fired then marker+pidfile deleted now exits 0 instead of 143, and a never-assigned runner is still reaped.

server/lib/tuist/runners/claims.ex
fortmarek: Fixed in cf611bc. `Claims.record_execution/3`, `Claims.complete_by_runner_name/2` and `RunnerSessions.record_execution/3` now take the account resolved from the delivery's App installation and filter on it; a delivery that resolves to no account touches no runner state at all. Regression tests cover a colliding `runner_name` across two accounts in both directions.

infra/helm/tuist/values.yaml
fortmarek: Fixed the description, not the defaults — enabling it automatically is intended. The PR body was written when the field was gated and went stale when I enabled it after finding the reaper was undeployable (your other comment). It now states plainly that the watchdog is on by default at 300s for every pool in every managed env, with per-pool/shape/xcodeOverride overrides and `0` to disable, plus why 300s adds no warm-pool churn (it bounds only the post-claim register→assign window; a pod still polling has no runner container).

server/lib/tuist/runners/dispatch.ex
fortmarek: Not fixed — you are right and it needs a decision, so I am not papering over it. Confirmed the blast radius: a trusted claim hands over `cache_signing_grant` (sign artifacts *as the account*), `volume_head`, and a trusted pod label that makes tart-kubelet materialize **and promote** the shared volume, so a fork job inherits all three.

Worth flagging: it cannot be fixed by binding the runner to the exact job. GitHub matches a runner to any job whose labels are a subset of the runner's, so extra/unique labels do not restrict assignment, and fork PR jobs run in the base repo with identical `runs-on` — runner groups cannot separate them either. So the only correct fix is your first option: never issue job-specific capabilities from a speculative claim, and move issuance behind the proven binding (JOB_STARTED-hook callback writing `$GITHUB_ENV`; volume label flipped only once `in_progress` proves trust, which moves materialization into the hook).

That is a cache-warmth vs fail-closed tradeoff I did not want to make unilaterally. Options: (1) fail-closed interim — stop issuing grant/volume/trusted-label speculatively, restores the boundary now, every job runs cold until the follow-up; (2) post-binding issuance as above; (3) split — land attribution/reaping/scoping and fix the boundary separately. I lean (1) now + (2) as a follow-up, since this is a security property. Which do you want?

server/lib/tuist/runners/dispatch.ex
fortmarek: Not fixed — confirmed and tracked as an open item in the description. `runner_jobs` is still written at mint time, so on a mismatch the customer-facing row, its `started_at`, the `pod_name`/`runner_name` binding and the interactive-session target all point at the wrong runner, and the session's `repository`/`workflow_name` stay from the claimed job. Agreed the right shape is a single rebind from the proven execution covering the CH row and the session metadata, not just the two PG columns. Holding it until the trust-boundary decision above, since that may move where the binding is established.

server/lib/tuist_web/controllers/runner_job_metrics_controller.ex
fortmarek: Not fixed — confirmed, and worth noting a retriable response alone would not work: both samplers are explicitly fail-open with no retry (`metrics-sampler.sh`: "a bad sample or failed POST is skipped"), so a 503 drops the batch just as silently. It needs either a pod-keyed server-side buffer flushed when the binding lands, or sampler retention (an image change) — I did not want to pick between those without your view, given the buffer has to survive the webhook landing on a different node than the POST. Tracked as an open item in the description.

server/lib/tuist/runners/dispatch.ex
fortmarek: Deferring this rather than fixing it here — flagging explicitly so it is a decision on the record, not an oversight.

Two things that inform it. First, **this PR does not introduce or widen the gap**: capability issuance in `serve_claim/5` is untouched, so the exposure is exactly what it is on `main` today. Second, we do not expect our customers to be using fork PRs, so the practical exposure is narrow, and Runners are still invite-only.

Worth recording what I found while digging, since it constrains the fix:

- The gap cannot be closed by pinning the runner to the job. GitHub matches a runner whose labels are a **superset** of `runs-on`, so extra labels on our runner do not restrict anything, `generate-jitconfig` takes no job parameter, and fork PR jobs run in the base repo so runner groups cannot separate them either. I checked the runner **scale set** protocol as an alternative: `GenerateJitRunnerConfig` also takes no job id, and `JobAssigned` carries no runner name — the binding only appears in `JobStarted`, after the fact. Same property.
- This is an industry-wide consequence of static labels, not a Tuist defect. RunsOn calls it ["runner stealing"](https://runs-on.com/guides/troubleshoot/); Depot's architecture (webhook → warm standby pool → ephemeral JIT runner → static label) is identical to ours.
- So the real fix is issuing capabilities only once the binding is proven — no grant or account label at claim, both delivered on `in_progress`. Note the grant specifically **cannot** be fixed by scrubbing `$GITHUB_ENV` in the hook: steps inherit the runner process env, so a job can read `/proc/<runner_pid>/environ` regardless.

Related: trust is a property of the workflow **run**, not the job (`job_trusted?` resolves `get_workflow_run(run_id)`), so all jobs in a run share a verdict. That makes run-scoped labels dissolve this for opt-in users — filed as https://github.com/tuist/tuist/issues/11875 — though being opt-in it can never be the boundary on its own.

Leaving this thread open for you to weigh in on the deferral rather than resolving it myself.

pepicrft (approved):
The account-scoping and watchdog issues are fixed. The remaining trust-boundary, durable-rebind, and pre-binding metrics gaps are explicitly documented as deferred follow-up work and do not block this change. The trust issue is pre-existing and is not widened by this implementation.

server/lib/tuist/runners/dispatch.ex
```diff
@@ -194,28 +201,138 @@ defmodule Tuist.Runners.Dispatch do
...
       # redelivery cannot resurrect the completion row.
-      :ok = Claims.complete(workflow_job_id)
+      #
+      # Release by EXECUTOR, not by the completed job's id: the Pod that
+      # claimed this job is not necessarily the one that ran it. Freeing
+      # by job id would release a slot still held by a runner executing
+      # someone else's claim, under-counting the account's live runners.
+      # A job cancelled while queued has no runner_name — nothing ran, so
+      # nothing is released; the claiming Pod keeps its slot until it
+      # stops (idle timeout) or `OrphanedRunnersWorker` recovers it.
```
esnunes: The comment on the completed path says a cancelled-while-queued job's claim is reclaimed by the idle timeout "or `OrphanedRunnersWorker` recovers it" — I don't think the worker can reach it: this same path calls `Jobs.complete/2`, which flips the ClickHouse row to `completed`, and the worker only lists rows still in `running` (`Jobs.list_orphaned_running/1`); `StaleClaimsWorker` excludes `running` claims by design. In practice the idle-timeout → pod-stop path is what releases this class, so the behavior is fine with the watchdog on — but I'd correct the comment so the recovery story isn't misleading for the next reader.

infra/linux-runner-image/run-job.sh
```diff
@@ -50,12 +50,84 @@ if [ -s "${CACHE_ENDPOINT_PATH}" ]; then
...
+      echo "$(date -u +%FT%TZ) run-job: no job assigned within ${idle_timeout}s; terminating idle runner"
+      kill -TERM "${runner_pid}" 2>/dev/null || true
+    fi
+  ) &
+  watchdog_pid=$!
+  printf '%s' "${watchdog_pid}" >"${WATCHDOG_PID_FILE}" 2>/dev/null || true
+fi
+
+wait "${runner_pid}"
```
esnunes, line 147: One thing I noticed in the exec→supervise switch: bash's `wait` returns immediately (128+15) when a trapped signal arrives, so on pod deletion the trap forwards TERM to run.sh, `wait` unblocks, and the script — the container's main process — exits within milliseconds; the runtime then SIGKILLs the still-running runner. The "so it deregisters cleanly" intent of the trap doesn't materialize: under `exec`, run.sh was the main process and had the full terminationGracePeriod to cancel/report the job and deregister. A second `wait "${runner_pid}"` after the first (the standard re-wait pattern, e.g. `while kill -0 "$runner_pid" 2>/dev/null; do wait "$runner_pid"; rc=$?; done`) restores the grace window. As written the pod's exit code on deletion is also always 143, regardless of the runner's real status.

infra/linux-runner-image/run-job.sh
```diff
@@ -50,12 +50,84 @@ if [ -s "${CACHE_ENDPOINT_PATH}" ]; then
...
+
+./run.sh --jitconfig "${jit}" --disableupdate &
+runner_pid=$!
+
+# Forward pod-deletion SIGTERM to the runner so it deregisters cleanly.
+trap 'kill -TERM "${runner_pid}" 2>/dev/null || true' TERM INT
+
+if [ "${idle_timeout}" -gt 0 ] 2>/dev/null; then
+  (
+    waited=0
```
esnunes, line 113: The latch rework closes the marker-deletion hole, but I think a narrow assignment race at the deadline survives it: ACTIONS_RUNNER_HOOK_JOB_STARTED fires when the Worker process starts the job — a second or more after the Listener has already acknowledged the assignment — so a job handed over in the last seconds of the idle window can still be TERMed as "idle" after GitHub has irrevocably placed it, and as far as we could determine, an ephemeral JIT runner dying post-acknowledgment marks the job failed rather than re-queuing it. Low probability (assignment must land in the final ~1–3s of the budget), but at fleet scale it becomes a background trickle of confusing one-off failures correlated with the deadline. A cheap way to close it completely: at expiry, kill only if the marker is absent *and* no `Runner.Worker` process exists — the Worker exists from the moment the Listener dispatches the job, before the hook runs, so it's the authoritative "got work" signal (one pgrep at the deadline).

server/lib/tuist_web/controllers/runner_job_metrics_controller.ex
esnunes, line 88: The attribution switch trades misattribution for silent truncation: samples now resolve against a binding that only exists after the in_progress webhook is delivered AND drained through Oban, and every earlier sample is 204'd with no retry — so each job's startup ramp depends on webhook latency, and under GitHub delivery delays or queue backlog, short jobs can chart entirely empty (the old claim-based path attributed from t=0). For the mismatch case the new behavior is strictly better; the cost lands on the matched majority. Since samples already arrive keyed by pod_name, would it be feasible to ingest unconditionally and resolve pod→job at binding/read time, so a late binding backfills instead of dropping data it can never recover? If the loss window is an accepted trade-off for now, a "samples dropped awaiting attribution" counter would at least make it observable. As a follow-up, the in-pod JOB_STARTED marker could complement this — it can't name the job, but it tells the ingest path "a job is running here, binding pending", so buffering (or at least that dropped-samples metric) can be gated on real work rather than applied to every idle pod.

server/lib/tuist/runners/claims.ex
```diff
@@ -320,20 +320,80 @@ defmodule Tuist.Runners.Claims do
...
-  Called from the `workflow_job.completed` webhook to free the
-  cap slot the instant GitHub tells us the job finished — without
-  this the slot stays occupied until `StaleClaimsWorker` sweeps
-  it (~5 min later), which is a real UX issue for a customer who
-  just freed a slot and wants to claim the next workflow_job.
-
-  Idempotent — repeated webhook deliveries are a no-op. Returns
-  `:ok` whether or not a row existed.
+
+  **Recovery path only.** This releases the claim of whichever Pod
```
esnunes, line 324: Under the mismatch this PR models, OrphanedRunnersWorker can still free a busy runner's slot: if pod B claimed J2 but is executing J1 (its claim carries `executed_workflow_job_id = J1`), and J2 sits queued on GitHub past the 5-min threshold, the worker's `handle_gh_status("queued")` path calls `Claims.release(J2, claimed_at)` — deleting pod B's row mid-job. When J1 later completes, `complete_by_runner_name("runner-b")` is a no-op (row already gone), so the account under-counts a live runner for J1's remaining runtime. The PR fixed exactly this class on the webhook path but the worker's queued path never consults the new column — skipping (or deferring) claims whose `executed_workflow_job_id` is set would close it with data you're already writing. (There's a partial coincidental compensation from the sibling's stranded claim until it's reaped, but it breaks down once the idle watchdog fires or the two claims differ in shape.)

server/lib/tuist_web/controllers/webhooks/github_controller.ex
```diff
@@ -229,13 +229,15 @@ defmodule TuistWeb.Webhooks.GitHubController do
...
-  @dispatchable_workflow_job_actions ~w(queued waiting completed)
+  # the same job as queued. `in_progress` is the first event carrying
+  # the `runner_name` GitHub actually placed the job on — the only
+  # real-time proof of the runner↔job binding, which the worker binds
+  # onto the runner's claim and session so metrics attribute to the
+  # job that ran. Any remaining action falls through to the catch-all
+  # `:ignored` branch in the worker; short-circuiting those here
+  # removes the Oban.insert (and its Postgres write) for events the
+  # worker would discard anyway.
+  @dispatchable_workflow_job_actions ~w(queued waiting in_progress completed)
```
esnunes: Re-admitting in_progress at the webhook door restores the per-event Oban insert the old comment on this line said was deliberately removed under burst load — and it now applies to every installation, including VCS-only customers whose jobs run on GitHub-hosted runners and can never match a claim (each event: one oban_jobs INSERT, one worker run, two lookups ending in :unknown_runner). Since `workflow_job.labels` is already in the payload at the door, would it be worth enqueueing in_progress only when a label looks like a Tuist runner label? That keeps the binding for the jobs that can actually match while preserving the old savings for the ~33%-of-traffic case the comment was written about.

server/lib/tuist/runners/claims.ex
```diff
@@ -477,4 +544,52 @@ defmodule Tuist.Runners.Claims do
...
+  emit it as telemetry:
+
+    * `:matched` — GitHub ran the job the claim was minted for.
+    * `:mismatch` — GitHub ran a *different* job on this runner than
+      the one claimed (the claim↔execution mismatch we're measuring).
+    * `:unknown_runner` — no live claim carries this `runner_name`
+      (a dropped/late webhook, or the claim already completed). The
+      durable session binding is the backstop for this case.
+  """
+  def record_execution(runner_name, executed_workflow_job_id, account_id)
```
esnunes, line 589: As far as I can trace, nothing reads `runner_claims.executed_workflow_job_id` — the metrics path reads the session's copy, and claim release keys on runner_name/pod_name — so the claim column is write-only, and its only effect is producing the second matched/mismatch verdict that `combine_attribution/2` merges (the same verdict the session computes against the same claimed id). Two coherent directions: drop it (the session becomes the single home of the binding; combine_attribution, the second write, one migration column, and a data-export paragraph all disappear), or make it earn its keep as the orphan worker's busy-signal (see the OrphanedRunnersWorker note — that's the one consumer that would genuinely want it). As-is it's a second copy that can silently disagree with the session, since the two writes aren't transactional. Not a blocker either way.

server/priv/repo/migrations/20260716120000_add_runner_execution_attribution.exs
```diff
@@ -0,0 +1,44 @@
...
+    # reported runner_name. Partial: the default "" (pre-mint) is
+    # never looked up and would bloat the index.
+    # excellent_migrations:safety-assured-for-next-line index_not_concurrently
+    create index(:runner_claims, [:runner_name], where: "runner_name <> ''")
+
+    # Durable attribution backstop: the session outlives the pod, so
+    # a `completed` webhook can still bind the runner after a fast
+    # job's pod is gone.
+    # excellent_migrations:safety-assured-for-next-line index_not_concurrently
+    create index(:runner_sessions, [:runner_name], where: "runner_name <> ''")
```
esnunes: A planner subtlety on the new indexes: the lookups are parameterized equality (`runner_name = $1` through Ecto's prepared statements), and Postgres can't prove `$1 <> ''` when building a generic plan, so the partial index is invisible to generic plans — the planner either keeps custom-planning every execution or caches a generic plan that can't use the index. On runner_claims (tiny) this is noise, but runner_sessions is append-only and grows forever, and this is its hottest new query path. The partiality also excludes almost nothing, since every session row gets a real runner_name at open. A plain `create index(:runner_sessions, [:runner_name])` would be generic-plan-safe at essentially the same size (the claims one can stay partial or go plain, it barely matters there).

server/test/tuist/runners/dispatch_test.exs
```diff
@@ -338,6 +362,138 @@ defmodule Tuist.Runners.DispatchTest do
...
+               Dispatch.handle_webhook(
+                 completed_payload(id: 4850, runner_name: "runner-exec", steps: []),
+                 1
+               )
+
+      assert_receive {:released, "runner-exec", ^account_id}
+    end
+
+    test "skips backstop attribution when the completed payload has no runner_name (cancelled-while-queued)" do
+      stub(Claims, :complete, fn _ -> :ok end)
```
esnunes: A handful of small cleanups, none blocking: (1) the completed-path tests still stub `Claims.complete/1` (4 sites, one added by this PR), which this PR removed from that path — dead stubs asserting a false dependency; (2) `record_execution/3`'s `source` param has a single call site passing a literal; (3) the hook/watchdog scaffolding is copy-pasted across the two image scripts and already diverges in three details (shebang, comments, TERM trap only on the Linux side) — a hook script baked into both images next to vitals.sh would keep the fleets' reaping behavior from drifting; (4) the metrics controller test hand-rolls `Repo.insert!(%RunnerSession{...})` instead of `RunnerSessions.open/1`, the path dispatch actually uses; (5) the "prefer open session, else latest closed" policy now exists twice in runner_sessions.ex (`session_for_runner/2` vs `latest_for_pod/1`).

esnunes (approved):
Really thorough work on a genuinely gnarly distributed-state problem — the claim/execution framing, the executor-keyed release with the J1/J2 test, and the passive scale-down philosophy all read as the right shape, and the PR description (now with the open-items section) made the review a pleasure. The latch + account-scoping fixes both look solid, including the two-directional collision tests.

These notes are on top of the review already on the PR — nothing here duplicates the trust-boundary blocker or the durable-rebind item, and the metrics-loss note extends what's already tracked. They cluster around three themes:

1. Recovery coverage. The cancelled-while-queued class isn't actually reachable by OrphanedRunnersWorker as the code comment suggests (`Jobs.complete/2` flips the ClickHouse row out of that worker's scan set) — with the watchdog on by default the pod-stop path bounds it, so the inline note just asks for the comment to be corrected. Relatedly, the orphan worker's queued-path release is still job-keyed and mismatch-blind, and can free a runner the new column proves busy.

2. Supervision semantics. The exec→supervise switch needs a re-wait after the TERM trap (as written, the runner loses its graceful-deregistration window and the pod always exits 143 on deletion), and the watchdog still has a narrow assignment race at the deadline that the latch rework doesn't cover — checking for a live Runner.Worker process at expiry would close it.

3. Attribution robustness. record_execution can report a binding it didn't persist (a swallowed update error on the session side, a select/update race on the claim side); a single UPDATE…RETURNING per store fixes both and halves the webhook path's queries.

Plus smaller notes: in_progress is enqueued for every installation with no label pre-filter, the partial indexes can't serve generic plans on the append-only sessions table, the claim-side executed column is currently write-only (unless the orphan worker starts consulting it), three AGENTS.md nodes went stale, and a few test/script cleanups.

infra/linux-runner-image/run-job.sh
fortmarek: Fixed in 5551cac. Right on both counts — and this was a regression I introduced with the exec→supervise switch. Added the re-wait loop (`wait` once, then re-wait while `kill -0` succeeds), so the runner keeps the full terminationGracePeriod to cancel/report and deregister, and `rc` is its real exit code rather than a blanket 143.

infra/linux-runner-image/run-job.sh
fortmarek: Fixed in 5551cac, using your `Runner.Worker` suggestion — the deadline check is now `marker absent && ! pgrep -f "Runner.Worker" && kill -0 runner`. The Worker existing from Listener-dispatch (before the hook) is exactly the authoritative signal the marker lacks, and one pgrep at expiry is cheap. Applied to both images. The failed-not-requeued consequence is what made it worth closing rather than accepting.

server/lib/tuist/runners/claims.ex
fortmarek: Fixed in 5551cac. Confirmed the trace: `handle_gh_status("queued")` judged the claim purely by the claimed job's GitHub status and would delete a live runner's row mid-job, after which the executor's `complete_by_runner_name` no-ops. Added `Claims.executing?/1` (is `executed_workflow_job_id` set) and the worker now skips those, letting the executor's completion or pod-stop free the slot. Test covers the skip.

server/lib/tuist/runners/claims.ex
fortmarek: Took your second direction in 5551cac — the column now earns its keep as exactly the consumer you identified: `OrphanedRunnersWorker` reads it via `Claims.executing?/1` to avoid releasing a busy runner (your other comment). You were right that it was write-only and that the duplicate verdict could silently disagree, since the two writes aren't transactional; the orphan-worker gate is a real read that the session's copy can't serve, because that path starts from a claimed workflow_job_id.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
fortmarek: Fixed in 5551cac. `in_progress` is now gated at the door on a `tuist-` label prefix, so VCS-only installations keep the old savings and the binding survives for jobs that can actually match. Test covers a `ubuntu-latest` in_progress not enqueueing. Good catch that I had silently reverted the very thing the comment on that line documented.

server/priv/repo/migrations/20260716120000_add_runner_execution_attribution.exs
fortmarek: Fixed in 5551cac — `runner_sessions.runner_name` is now a plain index, with the reasoning recorded in the migration. The generic-plan point is one I would not have caught: parameterized equality means Postgres can't prove `$1 <> `, so the partial index was invisible on the one table here that grows forever, while excluding essentially nothing since every session gets a real runner_name at open. Made the claims one plain too for consistency.

server/lib/tuist/runners/dispatch.ex
fortmarek: Fixed in 5551cac — comment corrected. You're right that the worker can't reach this class: this path's `Jobs.complete/2` flips the ClickHouse row to `completed` and `list_orphaned_running/1` only lists `running`. It now states the idle-timeout → pod-stop → `release_by_pod_name/1` path explicitly, and why the worker cannot.

server/lib/tuist_web/controllers/runner_job_metrics_controller.ex
fortmarek: Partially addressed in 5551cac — took the counter, not the buffering. Added `metrics_dropped_awaiting_attribution` (recovery telemetry, measured in samples) so the loss is visible instead of silent, and the trade-off is written down at the drop site.

Not buffering yet, deliberately: samples arrive at whichever node handles the POST while the webhook drains through Oban on another, so an in-memory buffer would lose exactly the data it exists to save — it needs a pod-keyed durable staging table, which is its own change. Your framing that the cost lands on the matched majority while the benefit is the mismatch minority is the right way to see it, and it argues for doing that properly rather than bolting on something node-local.

Also agreed on the JOB_STARTED marker as the gate — "a job is running here, binding pending" is a signal we now have in-pod anyway (this push uses it plus `Runner.Worker` for the watchdog), so it would let buffering apply to real work rather than every idle pod. Tracking both with pepicrft's metrics thread rather than expanding this PR further.

server/test/tuist/runners/dispatch_test.exs
fortmarek: Mostly done in 5551cac: (1) removed 3 dead `Claims.complete/1` stubs; (2) dropped `record_execution`'s `source` param; (4) the metrics test now goes through `RunnerSessions.open/1`; (5) "prefer open, else latest closed" is expressed once (`prefer_open_then_latest/1`).

(3) I skipped and want to flag why rather than leave it looking overlooked — you're right the scaffolding is duplicated and already diverging, but a genuinely shared script isn't reachable from both builds: the Linux image `COPY`s from `infra/linux-runner-image/` (Docker build context) and the macOS image's packer provisioner reads `${path.root}` = `infra/runner-image/`, so a single file needs a build-context change, and copying it into both dirs just moves the drift. I did close the behavioural divergence you noticed (both now have the Worker-process check), but the structural fix wants its own change where the image builds can actually be verified.

PR #11866, opened 2026-07-16
feat(kura,cli): scope the Xcode CAS snapshot to trunk, and ingest it ahead of the build

fortmarek:
The action-cache snapshot kura serves is a size-capped, newest-first view of a whole namespace. Feature-branch churn evicts trunk entries from it, and because a resolve answers from that snapshot, a fresh trunk build does not just get slower: it **loses hits outright and recompiles work the cache already holds** (measured: 464 of ~930 tasks lost this way under pollution).

This makes the snapshot mean one thing (**what trunk looks like, as CI built it**) and adds a layer that materializes that trunk closure into the local CAS off the build's critical path. It also fixes a separate regression that had been keeping C and Objective-C out of the cache entirely.

### The two decisions everything else follows from

**The snapshot is trunk-only.** An entry tagged with another branch is not in it, and neither is an untagged one. `None` used to mean "in the trunk baseline", for backward compatibility with entries published before the tag existed. That made every way of *not knowing* resolve itself the least safe way: silently, into the one view this scoping exists to keep clean. And there are more of those ways than the legacy one it was chosen for. An untagged entry is still stored, still served per key, and the first publisher to name a branch claims it.

**The branch comes from CI, and only from CI.** Not from the checkout. That makes CI the only publisher whose branch has to be right, and CI is exactly where a checkout cannot answer anyway, its HEAD being detached.

The second decision deleted more code than it added, and deleted a class of bug rather than fixing it: a registered source root could be moved, renamed, replaced by a different repository, or shared by two worktrees of one project, and every one of those quietly attributed a build to the wrong branch. None of it is reachable from a value the CI job told us about itself.

**The trade, stated plainly:** a project that publishes locally rather than from CI gets a permanently empty trunk view. It keeps per-key resolves and its local CAS; it loses the snapshot's breadth. That is the cost of the snapshot meaning one thing.

### What changed

**kura: a branch per action-cache entry, and a trunk-scoped snapshot**
- `ArtifactManifest` gains a `branch: Option<String>` (`serde(default, skip_serializing_if)`), threaded through `PersistedManifestRecord`/`PersistArtifactSpec`. Additive on purpose: manifests are serde_json, so an old node ignores the field and a new node defaults it, which is what the mixed-version overlap of a rolling deploy needs.
- `UpdateActionResult` stores the branch from request metadata; the snapshot serve filters by the trunk, under a compound `(namespace, trunk)` cache key. **No trunk asked for = today's behavior, byte for byte.**
- **Trunk-sticky tags, under the artifact lock.** A feature build recomputing an action shared with trunk republishes an identical-but-byte-wobbled result, and must not steal the key. The decision is a read-modify-write, so it happens inside the persist, under the lock, from a read taken there: an unlocked probe let a feature publish resume with a precomputed tag and overwrite a trunk publish that landed in between.
- **A publisher that names no branch cannot erase one.** An older node, or one whose context could not be resolved, is saying it cannot tell, not that the entry has none.
- **Damping compares the tag it would resolve, not just the bytes.** Two builds computing the same action produce the same bytes and differ only in the tag, so damping on the value alone silently dropped exactly the reclaim this design depends on.
- **An index row carries its branch in its KEY**, so a trunk-scoped scan rejects feature churn without a point-read. In the key rather than the value because a node that predates it reads the version from a fixed offset and the artifact id from the value, and never parses what sits between: its rows survive a rollback intact, which the README validates on the same persistent volumes.
- **An empty index is told apart from a stale one** by a per-namespace write generation. Empty is ambiguous, because "nothing to show" and "built before anything was published" look identical, and both wrong answers are expensive: serving a stale-empty view costs a client every key it came for (it fetches once per session), and rebuilding a correctly-empty one is a namespace scan per build.
- Branch survives mesh replication; trunk indexes are maintained eagerly.

**cas-plugin: CI's branch, the project's trunk, and the proxy as the enforcer**
- The branch **never touches a build setting**. Anything baked into the compiler invocation risks entering the `-cache-compile-job` key, which would split the cache per branch and stop a feature build ever hitting trunk: the exact opposite of the goal.
- Trunk is a **server decision**, taken from the project's configured default branch. Which branch is trunk is a property of the project, not of how this machine cloned it: a fork, a mirror, or a clone whose remote head was never set all get it wrong locally.
- **The upload policy is enforced at the proxy**, because it is the only place that sees the whole build. The plugin checks it too, but from a compiler option, and swift-build's `CASOptions` carries a plugin path and **no plugin options**: the CAS it creates for Clang caching never sees the project's answer and defaults to uploading. Every publication crosses `enqueue_publish`, from either lane and from the sweeper, so that is where `upload: false` is made to mean it. A refused record is deleted rather than left for the next sweep.
- **Tags are bound when a record is accepted**, not when it is uploaded. Publishing is asynchronous and durable, and the upload runs after the queue wait, the existence probe and the closure's transfers, which is typically after the producing build has exited. A sidecar carries the answer forward so a sweeper cannot re-tag it with whatever the registry says by then.
- **A refresh is distinct per (instance, action, branch, trunk)**, and a claim that does not become work is released. Keyed on the action alone, a feature job's refresh permanently suppressed the later trunk hit for that key, disabling the reclaim path outright on a shared runner.
- **The CAS handle is restatted before a demand fetch**, not only on a resolve. Nothing orders those, so after a DerivedData wipe a fetch is often the first thing to touch a handle bound to the deleted store, and clang does not survive `missing object`.
- **Background refreshes hold off while the machine is busy**, and a demand fetch counts as busy: the resolves all land during planning, so a long compile phase is nothing but fetches, and the machine would otherwise read as idle exactly when it is most bandwidth-bound.
- **No trunk ingestion on CI.** Ingestion pays for itself only where the machine can warm before a build; CI has no such window, and our runners get a warm CAS from an attached volume.

**cli**
- Generation sets `COMPILATION_CACHE_REMOTE_SERVICE_PATH` on the kura path. See the measurement below for why this is not cosmetic.
- `tuist setup cache` records the project's trunk, the CI job's branch, and the upload policy, and swaps the file in with `rename`. The proxy re-reads it on a timer, and any instant where it is missing reads as "no projects", which in turn reads as "upload".
- The branch is reported on GitHub pushes, not just pull requests: `GITHUB_HEAD_REF` is a PR-only variable, and a detached checkout left CI, the publisher the trunk view is built from, untagged.

### Measured: the clang regression

The gap that made this worth chasing was attributed before it was fixed: of an ingested build's 378 misses, **100% were clang** and **Swift was 0**. Two earlier readings of that were wrong: first "key drift", then "an Apple limitation" (documented in a3405d68d2 and superseded here). The keys are stable and it is not Apple's to fix.

clang does not load a CAS plugin, so left alone it caches only into Xcode's local builtin CAS. The build system covers clang itself, but only where it considers a remote cache present, and it decides that purely by `remoteServicePath != nil` (`CASOptions.hasRemoteCache`). That one flag gates uploading a clang or module output (`ClangCompileTaskAction`, `PrecompileClangModuleTaskAction`) and requesting the materialize-key task that fetches one back (`maybeRequestCachingKeyMaterialization`). We never set it on the kura path. **The legacy daemon always did**, so this was a regression the plugin path introduced, not a wall.

Isolated A/B, own socket/registry/kura port, fresh kura per arm, full DerivedData wipe before each build, both populate runs at 0 hit / 930 miss with `snapshot_hits=0` (proving the kura really was fresh):

| Replay after a full DerivedData wipe | Cached | Missed | Wall |
|---|---|---|---|
| Without `COMPILATION_CACHE_REMOTE_SERVICE_PATH` | 259 | 671 | 35.4s |
| With it | **930** | **0** | **28.3s** |

2,136 keys resolved with 2,136 remote hits and 0 misses: a complete replay from remote onto an empty CAS. The first attempt at this A/B varied two settings at once; the table is the single-variable re-run.

Our plugin *consumes* the matching `remote-service-path` option rather than forwarding it, so the flag flips without Xcode's own remote client, the slow path this plugin replaced, ever taking the socket.

### Measured: trunk scoping and ingestion

With the trunk closure ingested, the build stops talking to the network at all. From the proxy's own counters, at 50ms RTT, after a full DerivedData wipe:

```
resolves=2136  snapshot_hits=2136  remote_hits=0  demand_fetched=0  misses=0
```

Every resolve answered from the snapshot already in memory, no object fetched on demand, and the compiler reporting 930 hits / 0 misses. No compiles and no network: that is a local CAS replay, which is the floor by construction rather than by measurement.

Counters rather than a stopwatch, deliberately. An earlier version of this section quoted wall times (32.0–34.5s against a 17.4–17.5s floor) from a harness I no longer trust: re-running it produced 22.0s and 38.6s for the *same* floor arm, a 75% swing on unchanged code, and in one run the floor came out slower than a cold build. Two flaws explain it, and both invalidate the comparison rather than just adding noise:

- **The "pre-ingested" arm never was.** `pending=6437` objects were still queued when the build started. The build did not wait for them, but the proxy saturated disk and network in the background while it ran. Lifting the node budget to force a fuller ingest made that arm *slower*, which is the tell.
- **The floor arm is a different workload.** It reports 687 hit remarks against the ingested build's 930, consistently. Dropping `Build` products does not reproduce the same task set, so the two were never comparable.

**The honest cost, since the counters do not show it:** ingestion competes with the machine while it runs. It is off the build's critical path in the sense that the build never blocks on it, but it is not free, and a build that starts while it drains pays for the contention. Background refreshes are gated on the machine being idle; the initial ingest is not.

**No CI numbers are claimed.** Every measurement here populated and replayed on the same machine. That is the real shape of a DerivedData wipe, not of a fresh one. `<Module>_vers.c` stubs and Swift keys embed the DerivedData path, so CI likely wants a pinned `derivedDataPath` before any of this is worth measuring there.

### Why this shape

Keys are orders of magnitude lighter than bytes: ~100k of them ride one round trip in a ≤48MB frame, where the equivalent bytes are GBs. The key cache is the breadth layer (cold-start coverage in one fetch), scoping keeps that capped budget spent on entries a trunk build can actually hit, and ingestion buys the last gap to floor, affordable *only* because it is trunk-scoped. None substitutes for another.

### Rollout

Every entry published before this is untagged, so the first trunk-scoped views come back empty and those builds run cold. They refill themselves: a per-key hit queues a refresh under the tags of the build that took it, so a namespace re-tags over its next few builds with nothing re-uploaded. The kura side is header-gated and additive, so it can go out ahead of any client that sets them.

### Testing

kura 362 tests, cas-plugin 50, clippy clean on both. The CLI's mapper, setup-cache and GitController suites pass.

Each fix here is verified by mutation rather than by a green suite: the change is reverted, the test must fail, and it must be the right test. That is how several of these were found to be testing nothing, including one case where `-only-testing` with a Swift Testing function name ran zero tests and reported success.

### Reviewer notes

- The kura changes are header-gated and additive; absent headers preserve today's behavior exactly.
- Two mapper tests that asserted `COMPILATION_CACHE_REMOTE_SERVICE_PATH == nil` on the kura path are inverted here on purpose: they were pinning the regression as intended behavior.
- Setup still writes the source root into the registry's second column and the proxy steps over it. Dropping the column would shift every column after it, and a long-lived agent from before this change would read the trunk out of the branch's position.
- The index needs no backfill. A row from before the branch reports it as unknown and asks the manifest, as it always did, and the next publish supersedes it, so the migration rides along with the republishes that re-tagging needs anyway.

### DIFF
--- a/cas-plugin/src/proxy.rs
+++ b/cas-plugin/src/proxy.rs
@@ -868,44 +868,48 @@ impl DemandCoalescer {
     }
 }
 
-/// A git context read from an instance's source root, memoized on a TTL.
-struct GitContext {
+/// What setup recorded for an instance, memoized on a TTL so a publish does not
+/// re-read the registry file.
+struct SourceContext {
     read_at: Instant,
-    /// The current checked-out branch (`git rev-parse --abbrev-ref HEAD`), or
-    /// `None` on a detached HEAD (CI) — the env override covers that case.
-    branch: Option<String>,
-    /// The repo's default branch (`origin/HEAD`), used as the trunk to scope the
-    /// snapshot to, or `None` when the remote head is unknown.
     trunk: Option<String>,
+    ci_branch: Option<String>,
 }
 
-/// How long a derived git context is reused before the proxy re-reads HEAD, so
-/// a branch switch is picked up within seconds without forking git per publish.
+/// How long a recorded context is reused before the proxy re-reads the registry,
+/// so a project set up after startup is picked up within seconds.
 const GIT_CONTEXT_TTL: Duration = Duration::from_secs(15);
 
-/// The (branch, trunk) pair `git_context` resolves, cloned out from under the
+/// The (branch, trunk) pair `source_context` resolves, cloned out from under the
 /// cache lock.
-struct GitBranches {
+struct SourceBranches {
     branch: Option<String>,
     trunk: Option<String>,
 }
 
 /// What `tuist setup cache` recorded for an instance.
+///
+/// Note what is NOT here: the checkout. Nothing about a publish is read from the
+/// working copy any more, which is what makes a moved, renamed, or duplicated one
+/// unable to mis-attribute a build. Setup still writes the source root into the
+/// registry's second column and this deliberately ignores it, so a proxy from
+/// before this change keeps parsing the columns after it.
 struct RegisteredSource {
-    /// The project's checkout, which the branch of each publish is read from.
-    root: PathBuf,
-    /// The branch `tuist setup cache` saw, recorded ONLY on CI, whose checkout is
-    /// detached and so tells `git rev-parse` nothing. A launchd agent does not
-    /// inherit the job's environment, so the provider's branch variable is
-    /// unreachable from here and the command inside the job has to hand it over.
-    /// Never recorded off CI, where reading git live is both possible and correct
-    /// across a `git checkout`.
+    /// The branch `tuist setup cache` saw in the CI job's environment, recorded
+    /// ONLY on CI. A launchd agent does not inherit the job's environment, so the
+    /// provider's branch variable is unreachable from here and the command inside
+    /// the job has to hand it over.
+    ///
+    /// `None` off CI, and that is the design rather than a gap: the snapshot is
+    /// what trunk looks like as CI built it, so CI is the only publisher whose
+    /// branch has to be right. A local publish goes out untagged, stays out of
+    /// every trunk view, and is still stored and served per key.
     ci_branch: Option<String>,
-    /// The project's default branch, as the server knows it. Authoritative over
-    /// the checkout's `origin/HEAD`: which branch is trunk is a property of the
-    /// project, not of how this machine happened to clone it (a fork, a mirror,
-    /// or a clone whose remote head was never set all get it wrong locally).
-    /// `None` when setup could not reach the server.
+    /// The project's default branch, as the server knows it. Which branch is
+    /// trunk is a property of the project, not of how this machine happened to
+    /// clone it (a fork, a mirror, or a clone whose remote head was never set all
+    /// get it wrong locally). `None` when setup could not reach the server, which
+    /// means no scoping: what a client too old to ask for it already gets.
     trunk: Option<String>,
 }
 
@@ -935,7 +939,7 @@ pub struct Proxy {
     active_instances: Mutex<HashSet<String>>,
     // instance -> the last git context read from its source root, refreshed on
     // a short TTL so per-publish tagging is a cache hit, not a git fork.
-    git_cache: Mutex<HashMap<String, GitContext>>,
+    source_cache: Mutex<HashMap<String, SourceContext>>,
     paths: Mutex<HashMap<String, &'static PathState>>,
     publisher: Prefetcher,
     // Resolves/publishes that arrived with no declared instance and no primed
@@ -1017,7 +1021,7 @@ impl Proxy {
                     .map(|path| load_sources(&sources_path_for(path)))
                     .unwrap_or_default(),
             ),
-            git_cache: Mutex::new(HashMap::new()),
+            source_cache: Mutex::new(HashMap::new()),
             registry_path,
             paths: Mutex::new(HashMap::new()),
             publisher: Prefetcher::new(),
@@ -1904,14 +1908,15 @@ impl Proxy {
     /// where they are used (the upload, which runs after the queue wait, the
     /// existence probe and the closure's blob transfers, tens of seconds over
     /// a WAN link, and mostly AFTER the build that produced them has exited).
-    /// Resolving there reads whatever is checked out by then, so a `git checkout`
-    /// straight after a build permanently re-tags that build's still-draining
-    /// outputs with the new branch: a trunk build's outputs land tagged
-    /// `feature` and drop out of the trunk view they belong in.
     ///
-    /// Binding at accept costs nothing: `git_context` is memoized on
-    /// GIT_CONTEXT_TTL, so this forks git at most once per TTL per instance,
-    /// never per publish.
+    /// Resolving there would read whatever the registry says by then, and the
+    /// registry moves: a shared CI runner rewrites it on every job's setup. So
+    /// job A's still-draining outputs would be tagged with job B's branch, and a
+    /// trunk build's outputs would land tagged `feature` and drop out of the
+    /// trunk view they belong in.
+    ///
+    /// Binding at accept costs nothing: the context is memoized, so this reads
+    /// the registry at most once per TTL per instance, never per publish.
     fn enqueue_publish(&self, cas_path: &str, instance: &str, record_path: &str) {
         let (branch, trunk) = self.record_tags(instance, record_path);
         let mut item = Vec::with_capacity(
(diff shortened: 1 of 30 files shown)


cas-plugin/src/proxy.rs
```diff
@@ -1557,13 +2054,13 @@ impl Proxy {
         // DIFFERENT value (a genuine local recompute) still publishes.
         if let Some(Resolution::Hit(value)) = state.resolved.lock().unwrap().get(&record.key) {
             if value == &record.value_digest {
-                let _ = std::fs::remove_file(&record_path);
+                remove_record(&record_path);
```
pepicrft: Could we keep tagged records on this path? I reproduced this with the current code by preloading `resolved` with the same key and value, then publishing a record tagged `main`. The record was deleted here and `publish` was never reached, so the identical-manifest retagging path below cannot reclaim an entry that is still tagged with a feature branch. Making this shortcut conditional on both tags being absent, or making the resolved cache tag-aware, would preserve that reclaim path.

cli/Sources/TuistKit/Services/Setup/SetupCacheCommandService.swift
```diff
@@ -24,29 +26,182 @@ enum SetupCacheCommandServiceError: Equatable, LocalizedError {
...
+            sourcesPath = try AbsolutePath(
+                validating: Environment.current.casProxySocketPath().pathString + ".registry.sources"
+            )
+        }
+
+        // Read every other project back: this rewrites the whole file, so anything
+        // lost here is a project silently losing its policy. A registry we cannot
+        // decode therefore fails the command rather than being written over with
+        // just this project, which would erase every other one's.
+        var entries: [String: RegisteredSource] = [:]
```
pepicrft: Could we serialize this read-modify-write with a cross-process file lock? I ran two processes through this exact read, decode, upsert, stage, and rename sequence, with both reads completed before either rename. The final registry retained only one new project. Atomic rename protects readers, but the last writer still drops the other project's trunk and upload policy.

kura/src/store.rs
```diff
@@ -1332,16 +1383,30 @@ impl Store {
...
-                action_cache_index_key(&manifest.namespace_id, manifest.version_ms, action_hash),
+                action_cache_index_key(
+                    &manifest.namespace_id,
+                    manifest.version_ms,
+                    action_hash,
+                    manifest.branch.as_deref(),
+                ),
                 artifact_id.as_bytes(),
             );
+            self.bump_action_cache_generation(&manifest.namespace_id);
```
pepicrft: Could we bump the generation after `write_batch_sync` succeeds? With a temporary runtime barrier here, I observed generation `1` while the trunk scan was still empty. After the write committed, the scan contained the entry but the generation remained `1`. A snapshot can therefore stamp the pre-commit view as current.

kura/src/store.rs
```diff
@@ -3915,6 +4145,61 @@ fn outbox_message_key(message: &OutboxMessage) -> String {
...
+        // publisher that could tell recorded, or a node too old to send the
+        // header would untag a trunk key and drop it out of the trunk view by
+        // republishing it.
+        (Some(existing), _) if branch.is_none() => existing.branch.as_deref(),
+        _ => branch,
+    }
+}
+
+/// Whether a manifest belongs in a trunk-scoped snapshot: entries tagged with
+/// the trunk branch and untagged/legacy entries (no branch) form the trunk
```
pepicrft: This description appears inverted. The runtime trunk-filter test confirms that an untagged entry is excluded, and `branch_in_trunk` only accepts `Some(trunk)`. Could we update this to say that only entries tagged with the trunk branch form the scoped baseline?

pepicrft (approved):
Validated this at runtime on `1860fbed`. The content-addressable storage plug-in reproduction confirmed that a tagged same-value record is discarded before retagging, the two-process registry reproduction confirmed a lost update, and the Kura barrier confirmed that the generation can advance before the database commit. The full Kura suite passed, including the trunk-scope behavior, and the worktree is restored clean. I left focused inline follow-ups for those cases and the contradictory comment. Approving the overall direction.

cas-plugin/src/proxy.rs
fortmarek: Done in f15586b4. The churn-skip now fires for every re-put except a trunk publish (`branch == trunk`): a trunk build's re-put is the reclaim path, so it must reach `publish` even when the value matches, since the entry it matches may still be feature-tagged. Feature/local/untagged re-puts stay free; only post-merge trunk builds pay the round trip, which is where the reclaim belongs. Extracted the decision into `is_redundant_reput` and unit-tested it (including your exact case: trunk re-put + matching value must not be dropped).

cli/Sources/TuistKit/Services/Setup/SetupCacheCommandService.swift
fortmarek: Done in f15586b4. The read-modify-write is now serialized with an advisory `flock` on a `<registry>.sources.lock` sidecar, held across the whole read → decode → upsert → stage → rename. The second setup waits and upserts onto the first's result instead of dropping it. Released on descriptor close, including on a crash, so a killed setup can't wedge the next.

kura/src/store.rs
fortmarek: Fixed in f15586b4. The `bump_action_cache_generation` now runs after `write_batch_sync` succeeds, so the generation only advances once the index row is committed and visible to a scan. A concurrent serve can no longer stamp the pre-commit, row-less view as current.

kura/src/store.rs
fortmarek: Fixed in f15586b4. Corrected the doc to say only entries tagged with the trunk branch form the scoped baseline; untagged is excluded (matching `branch_in_trunk`, and its own sibling doc two lines down).

PR #11879, opened 2026-07-16
fix(server): pin the Kura CAS budget to the region's declared storage size

fortmarek:
## What changed

`KURA_CAS_CAPACITY_BYTES` is now derived from the disk envelope each Kura region declares and injected through `extension_env/1`. That envelope is normally `storage_size`; a region whose claim bounds nothing can override it with the new `cas_capacity_size`, which the runner-cache region uses to keep the ~137Gi per account it already holds. `@manifest_revision` is bumped so the rollout worker re-applies the manifest to instances that already exist.

## Why

Found while investigating a live eu-central outage (2026-07-16): the single Dedibox node backing the region sat at `DiskPressure=True`, which evicts running pods **and** blocks rescheduling. Because it is a one-node fleet nothing could move, so the region lost its demux, its ingress and every tenant instance. The `platform-kura-eu-central-ingress-nginx-controller` DaemonSet could then never go Ready, so `helm --wait` on the Platform chart failed and the whole production deploy cascade died with it.

## Root cause

When `KURA_CAS_CAPACITY_BYTES` is unset, Kura sizes its CAS segment ring from `libc::statvfs()` on the data dir (`kura/src/store.rs`, `resolve_segment_ring_limits`), defaulting to `CAS_CAPACITY_DEFAULT_DISK_PERCENT` (50%). Nothing in the server or the kura-controller ever set it.

Every managed region is backed by the local-path provisioner, where a volume is a plain directory on the node's shared disk. `statvfs` therefore reports **the whole box**, not the PVC's declared size — and on ext4 nothing enforces the PVC's capacity either, so the `50Gi` claim was never more than a label. Each replica budgeted 50% of the entire node, and the two replicas a region runs on its single node over-committed it two to one.

The node's own confirmation, straight from a replica's startup log:

```
"resolved CAS segment ring limits" ... "capacity_bytes":491773755392
```

That is 458 GiB — exactly 50% of the 984 GB filesystem — per replica, times two replicas, on one disk.

The box crosses kubelet's `imagefs.available<15%` eviction threshold long before either replica reaches its budget, so Kura's ring rotation never gets to evict anything and kubelet evicts the region instead. Two details worth recording:

- The filesystem hosting the volumes also hosts `/data/containerd` and `/var/lib/kubelet` (all bind mounts of one `md2p1`), which is why a cache-sizing bug surfaces as *imagefs* pressure.
- The `FreeDiskSpaceFailed` / `ImageGCFailed` events on the node are a red herring. The node had 4 images totalling 0.34 GB and reported `only found 0 bytes eligible to free`; the "attempted to free 49.4 GB" figure is just 5.022% of the filesystem, the default 85%→80% image-GC band.

Measured on the box: 732G of the 733G in use was Kura's segment data, and 733G of that belonged to Tuist's own two dogfooding replicas (379G + 354G). The four paying tenants co-located in the region were using ~12MB each and were evicted anyway.

## How the budget is derived

`KURA_CAS_CAPACITY_BYTES` bounds the **segment ring only**, not the data dir. The ring shares its volume with two things the cap does not cover: the controller points `KURA_TMP_DIR` at `<data dir>/tmp`, so upload staging is inside the claim, and RocksDB's index sits beside it with no budget of its own.

So the declared size is treated as the whole envelope and the fixed reserves are subtracted from it:

- the tmp dir's ceiling (`KURA_TMP_DIR_MAX_BYTES`, never set by us, so Kura's 8 GiB default applies)
- one `MAX_SEGMENT_BYTES` (512 MiB), which a rotation appends before evicting the oldest segment
- 3% for the index, which tracks entry count rather than bytes (measured ~1.2% of resident segment bytes on a production instance)

Subtracting rather than taking a percentage matters, and an earlier revision of this PR got it wrong: the reserves are **fixed sizes, not a share**, so any percentage that fits a large volume overruns a small one. Subtracting also makes the declared number mean something useful on a local-path volume — the envelope *is* the per-replica footprint, which is exactly what the box arithmetic below needs. The flat 80% quietly understated it (a 40Gi ring, but ~49Gi actually on disk once staging and the index are counted). When what remains is not worth a ring, the env is omitted entirely and Kura's own default applies.

To be precise about the stakes: today this changes nothing for a live region, because every region that actually runs is local-path, where the claim is unenforced and only the box total binds. The one region declaring an enforced 20Gi claim, `hetzner-staging-runners`, is dead config — its Hetzner `kura` node pool no longer exists, so its pods have been unschedulable and its PVCs unbound for over a month (see below). The subtraction is what keeps the rule correct rather than accidentally correct.

## Why this fix

The declared envelope is what the region already asks for, so deriving the budget from it makes actual usage match the stated contract instead of silently tracking whatever box the volume landed on. It is also correct on a CSI-backed volume, where the claim is enforced.

**Why `cas_capacity_size` rather than just raising `storage_size`.** The first revision carried the runner-cache budget by setting `storage_size: 200Gi`, on the reading that a size change is inert for existing instances (`reconcileStatefulSet` preserves `existingVolumeClaimTemplates`, and `staleDataStorageReason` recreates on a storage *class* change, not a size one). That reading was incomplete and the approach was wrong: `reconcileStatefulSet` also calls `reconcileDataPersistentVolumeClaims` on every pass, which patches each live data PVC up to `spec.storageSize` whenever the desired size is larger. `scw-local-nvme` sets `allowVolumeExpansion: false` (verified in the chart and live in production), so the API would reject every resize and leave all three existing runner-cache instances failing to reconcile. Keeping the budget on its own field leaves the claim untouched. The budget still defaults to `storage_size`, so the safe behaviour is the default and the override is only for regions where the claim is a fiction.

Other alternatives considered:

- **Setting the env by hand on the CR.** Rejected: the server writes `"extraEnv" => extension_env(region)` wholesale and the kura-controller owns the StatefulSet, so any manual edit is reconciled away.
- **Dividing the disk by co-located replicas inside Kura.** Rejected: a replica cannot see how many peers share its filesystem, and it would keep the budget coupled to the box instead of to the declared volume.
- **Capping only the 2-replica regions.** Rejected: it leaves the budget silently box-derived everywhere else, so the same bug returns the moment a region gains a replica or moves to a smaller box.
- **Cutting the runner cache to the nominal 50Gi.** Rejected: at 1 replica the 50%-of-disk default was never over-committed, and that region has been running near 137Gi per account for a long time. Enforcing a number that never applied would shrink a healthy cache ~3.5x for no reason. `150Gi` reproduces roughly what it already holds; `200Gi` would have grown it and left the box unable to take a fourth account.

## Resulting headroom

| Region | budget source | ring/replica | envelope/replica | replicas | tenants | total on box | box | used | tenant ceiling |
|---|---|---|---|---|---|---|---|---|---|
| `eu-central` | storage_size 50Gi | 40Gi | 50Gi | 2 | 5 | 537GB | 984GB | **54.6%** | ~8 |
| `scw-fr-par-runners` | cas_capacity_size 150Gi | 137Gi | 150Gi | 1 | 3 | 483GB | 909GB | **53.2%** | ~5 |
| `us-east` | storage_size 50Gi | 40Gi | 50Gi | 2 | 2 | 215GB | 942GB | **22.8%** | ~7 |
| `us-west` | storage_size 50Gi | 40Gi | 50Gi | 2 | 2 | 215GB | 942GB | **22.8%** | ~7 |

Against 85% at the time of the incident. These are worst case: the envelope assumes every replica holds a full ring *and* hits its staging ceiling at once, so steady state sits lower (eu-central ~45%). Every region keeps `storage_size: 50Gi`, so no PVC resize is attempted anywhere.

## Impact

- eu-central stops evicting itself, and us-east / us-west / ca-east carry the same latent bug today — they simply have not taken enough traffic to trip it yet. `scw-fr-par-runners` is the natural experiment: 1 replica, so 50% of the disk was always a safe plateau, and it sits stably at 48% with rotation working.
- Tuist's eu-central cache is bounded at 40Gi per replica instead of growing to 458 GiB, so its dogfooding traffic can no longer evict paying tenants off the box.

## Residual risk (not addressed here)

- **The cap is per instance, so the box total scales with tenant count** — see the ceiling column. eu-central is fine at 5 tenants and would approach the eviction line near ~8; `scw-fr-par-runners` tops out near ~5 accounts. That is now linear and predictable rather than a fixed 200% over-commit, but adding accounts past a region's ceiling needs a second box, not a bigger envelope. Each region also remains a **single node** with no failover, so a full box is still a full-region outage.
- `reconcileDataPersistentVolumeClaims` patches a PVC without checking whether the storage class allows expansion. That is a latent footgun independent of this PR: any future `storage_size` increase on a local-path region wedges its instances the same way. Worth fixing in the controller separately.
- The 8 GiB staging reserve mirrors Kura's `DEFAULT_TMP_DIR_MAX_BYTES` rather than reading it, because we never set `KURA_TMP_DIR_MAX_BYTES`. If that default moves, this constant has to move with it.
- **`hetzner-staging-runners` is dead and should be deleted**, separately from this PR. Its `kura` node pool (Hetzner Falkenstein) is gone — staging's Linux runners moved to bare metal (`runners-linux`) — but the region spec stayed behind and keeps minting instances that can never schedule: 12 KuraInstances on staging, 33-35 days old, pods `FailedScheduling` against a `node.cluster.x-k8s.io/pool: kura` that does not exist, PVCs stuck Pending. Nothing is provisioned so it costs nothing, but it is the last region pinning `hcloud-volumes` into the region table; removing it makes every Kura region bare-metal (OVH/Dedibox/Scaleway EM) and drops the only enforced-claim case from this code path. Left out here because it touches `kura_test.exs` fixtures and needs the orphaned staging instances reaped.

## Validation

- `mix test test/tuist/kura/` — 191 passed. New cases cover the derived value, the override path, omission when a region declares none, a regression test asserting the runner region budgets from `cas_capacity_size` while `storageSize` stays `50Gi`, an assertion that a 20Gi enforced volume still fits ring + staging + rotation, and omission when the envelope is too small to carve a ring from.
- `mix format --check-formatted` and `mix credo` clean on the changed files.
- Verified against the live fleet: caps rendered from the real region configs and checked against each box's actual disk size (table above); `allowVolumeExpansion: false` confirmed on the live `scw-local-nvme` StorageClass; the three live runner-cache PVCs confirmed still at 50Gi; `KURA_TMP_DIR=/var/cache/kura/tmp` confirmed inside `KURA_DATA_DIR`.
- The incident itself was resolved by reclaiming the 733G (Tuist's own cache only); `DiskPressure` cleared after kubelet's 5-minute transition period, the demux, the regional ingress and all four tenant instances rescheduled, and the eu-central hosts serve again behind a valid certificate.

### DIFF
--- a/server/lib/tuist/kura/provisioner/kubernetes_controller.ex
+++ b/server/lib/tuist/kura/provisioner/kubernetes_controller.ex
@@ -410,16 +410,26 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   # single node over-commit it (2 replicas x 50% of the disk = 100% of it). The
   # box then crosses kubelet's imagefs eviction threshold long before any replica
   # reaches its own budget, so Kura's ring rotation never gets to evict and the
-  # node evicts the whole region instead. Pin the budget to the volume the region
-  # actually asked for so the ring stays inside the PVC contract on any backing.
+  # node evicts the whole region instead.
+  #
+  # Budget from the size the region declares. That is normally storage_size, so
+  # the ring stays inside the claim on a class that enforces it; a region whose
+  # claim bounds nothing (local-path) can override with cas_capacity_size rather
+  # than inflate storage_size, which the controller would try to apply to the
+  # live PVCs.
   defp cas_capacity_env(%Regions{} = region) do
-    with size when is_binary(size) <- storage_size(region),
+    with size when is_binary(size) <- cas_capacity_source(region),
          {:ok, bytes} <- parse_storage_quantity(size) do
       [env_var("KURA_CAS_CAPACITY_BYTES", Integer.to_string(cas_capacity_bytes(bytes)))]
     else
       _ -> []
     end
   end
+
+  defp cas_capacity_source(%Regions{provisioner_config: %{cas_capacity_size: size}}) when is_binary(size) and size != "",
+    do: size
+
+  defp cas_capacity_source(%Regions{} = region), do: storage_size(region)
 
   # Kura appends a new segment before evicting the oldest one, so a ring sized to
   # the entire volume can still run it full mid-rotation. Kura reserves the same
--- a/server/lib/tuist/kura/provisioner/kubernetes_controller.ex
+++ b/server/lib/tuist/kura/provisioner/kubernetes_controller.ex
@@ -19,7 +19,7 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   alias Tuist.Kura.Server
 
   @namespace "kura"
-  @manifest_revision "2026-07-09-mesh-peers-sync-v1"
+  @manifest_revision "2026-07-16-cas-capacity-v1"
   @manifest_revision_annotation "tuist.dev/kura-manifest-revision"
   @impl true
   def provision(%{name: handle}, %Regions{} = region, %Server{}) do
@@ -397,9 +397,55 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
         "KURA_CONTROL_PLANE_CLIENT_ID",
         Environment.kura_control_plane_client_id()
       ) ++
+      cas_capacity_env(region) ++
       mesh_peers_sync_env(region) ++
       telemetry_env(region)
   end
+
+  # With KURA_CAS_CAPACITY_BYTES unset, Kura sizes its CAS segment ring from
+  # statvfs() on the data dir. Every managed region is backed by the local-path
+  # provisioner, where a volume is a plain directory on the node's shared disk,
+  # so statvfs reports the whole box instead of the PVC's declared size: each
+  # replica budgets a fraction of the box, and replicas co-located on a region's
+  # single node over-commit it (2 replicas x 50% of the disk = 100% of it). The
+  # box then crosses kubelet's imagefs eviction threshold long before any replica
+  # reaches its own budget, so Kura's ring rotation never gets to evict and the
+  # node evicts the whole region instead. Pin the budget to the volume the region
+  # actually asked for so the ring stays inside the PVC contract on any backing.
+  defp cas_capacity_env(%Regions{} = region) do
+    with size when is_binary(size) <- storage_size(region),
+         {:ok, bytes} <- parse_storage_quantity(size) do
+      [env_var("KURA_CAS_CAPACITY_BYTES", Integer.to_string(cas_capacity_bytes(bytes)))]
+    else
+      _ -> []
+    end
+  end
+
+  # Kura appends a new segment before evicting the oldest one, so a ring sized to
+  # the entire volume can still run it full mid-rotation. Kura reserves the same
+  # headroom itself (CAS_CAPACITY_MAX_DISK_PERCENT); apply it here against the
+  # volume size, which is the denominator it cannot infer on a local-path volume.
+  defp cas_capacity_bytes(storage_bytes), do: div(storage_bytes * 80, 100)
+
+  defp parse_storage_quantity(value) do
+    case Integer.parse(value) do
+      {quantity, suffix} when quantity > 0 ->
+        case storage_multiplier(String.trim(suffix)) do
+          nil -> :error
+          multiplier -> {:ok, quantity * multiplier}
+        end
+
+      _ ->
+        :error
+    end
+  end
+
+  defp storage_multiplier(""), do: 1
+  defp storage_multiplier("Ki"), do: 1024
+  defp storage_multiplier("Mi"), do: 1024 * 1024
+  defp storage_multiplier("Gi"), do: 1024 * 1024 * 1024
+  defp storage_multiplier("Ti"), do: 1024 * 1024 * 1024 * 1024
+  defp storage_multiplier(_), do: nil
 
   # Managed pods fetch the account's self-hosted peer list from the control
   # plane at boot and on cadence, so a self-hosted peer joining or leaving
--- a/server/lib/tuist/kura/provisioner/kubernetes_controller.ex
+++ b/server/lib/tuist/kura/provisioner/kubernetes_controller.ex
@@ -21,6 +21,14 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   @namespace "kura"
   @manifest_revision "2026-07-16-cas-capacity-v1"
   @manifest_revision_annotation "tuist.dev/kura-manifest-revision"
+  # Mirrors Kura's DEFAULT_TMP_DIR_MAX_BYTES (kura/src/constants.rs): 4 x
+  # MAX_REPLICATION_BODY_BYTES, itself 4 x MAX_SEGMENT_BYTES. We never set
+  # KURA_TMP_DIR_MAX_BYTES, so this default is what upload staging can reach
+  # inside the data volume. Keep in sync if either constant moves.
+  @kura_tmp_dir_max_bytes 8 * 1024 * 1024 * 1024
+  # Kura's MAX_SEGMENT_BYTES: the one extra segment a ring rotation appends
+  # before evicting the oldest one.
+  @kura_max_segment_bytes 512 * 1024 * 1024
   @impl true
   def provision(%{name: handle}, %Regions{} = region, %Server{}) do
     {:ok, instance_name(handle, region)}
@@ -419,8 +427,9 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   # live PVCs.
   defp cas_capacity_env(%Regions{} = region) do
     with size when is_binary(size) <- cas_capacity_source(region),
-         {:ok, bytes} <- parse_storage_quantity(size) do
-      [env_var("KURA_CAS_CAPACITY_BYTES", Integer.to_string(cas_capacity_bytes(bytes)))]
+         {:ok, bytes} <- parse_storage_quantity(size),
+         capacity when is_integer(capacity) <- cas_capacity_bytes(bytes) do
+      [env_var("KURA_CAS_CAPACITY_BYTES", Integer.to_string(capacity))]
     else
       _ -> []
     end
@@ -431,11 +440,27 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
 
   defp cas_capacity_source(%Regions{} = region), do: storage_size(region)
 
-  # Kura appends a new segment before evicting the oldest one, so a ring sized to
-  # the entire volume can still run it full mid-rotation. Kura reserves the same
-  # headroom itself (CAS_CAPACITY_MAX_DISK_PERCENT); apply it here against the
-  # volume size, which is the denominator it cannot infer on a local-path volume.
-  defp cas_capacity_bytes(storage_bytes), do: div(storage_bytes * 80, 100)
+  # KURA_CAS_CAPACITY_BYTES budgets the CAS segment ring only, but the ring is
+  # not the only thing in the data dir: the controller points KURA_TMP_DIR at
+  # <data dir>/tmp, so upload staging shares the volume, and RocksDB's index sits
+  # beside it with no budget of its own. Size the ring against what is left after
+  # them rather than taking a flat percentage — the tmp budget is a fixed 8 GiB,
+  # so a percentage that fits a 50Gi volume overruns a 20Gi one.
+  #
+  # Reserves, in order: the tmp dir's own ceiling; one extra segment, which a
+  # rotation appends before it evicts the oldest one; and a few percent for the
+  # RocksDB index, which tracks entry count rather than bytes (measured ~1.2% of
+  # resident segment bytes on a production instance, so 3% is slack).
+  defp cas_capacity_bytes(storage_bytes) do
+    usable = storage_bytes - @kura_tmp_dir_max_bytes - @kura_max_segment_bytes
+
+    if usable > 0 do
+      div(usable * 97, 100)
+      # Too small to carve a ring out of once staging is reserved. Emit nothing
+      # and let Kura fall back to its own statvfs default, which on a volume this
+      # size is both enforced and more conservative than anything derived here.
+    end
+  end
 
   defp parse_storage_quantity(value) do
     case Integer.parse(value) do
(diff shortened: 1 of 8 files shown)


server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -397,10 +405,83 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
...
+  # so a percentage that fits a 50Gi volume overruns a 20Gi one.
+  #
+  # Reserves, in order: the tmp dir's own ceiling; one extra segment, which a
+  # rotation appends before it evicts the oldest one; and a few percent for the
+  # RocksDB index, which tracks entry count rather than bytes (measured ~1.2% of
+  # resident segment bytes on a production instance, so 3% is slack).
+  defp cas_capacity_bytes(storage_bytes) do
+    usable = storage_bytes - @kura_tmp_dir_max_bytes - @kura_max_segment_bytes
+
+    if usable > 0 do
```
esnunes, line 473: Really solid derivation overall — one interaction with the runtime I wanted to flag. Kura clamps the segment-ring total to a legacy floor of 5 segments (2.5GiB): `resolve_segment_ring_limits` does `.clamp(floor.total_segments(), MAX_DESIRED_SEGMENTS)` (kura/src/store.rs:3697, floor = 1+2+2 segments of 512MiB). So any emitted `KURA_CAS_CAPACITY_BYTES` below ~2.5GiB gets silently raised to a 2.5GiB ring. `cas_capacity_bytes/1` emits whenever `usable > 0`, which leaves a band of declared sizes (roughly 8.5Gi–11.1Gi) where the runtime will use more disk than the derivation promised — the invariant the 20Gi test asserts (ring + tmp ceiling + rotation inside the volume) doesn't hold there on an enforced class. Today the only region in that band is `local-controller` (10Gi, kind local-path, unenforced), so nothing breaks live — it's latent. A cheap guard would be to emit only when `div(usable * 97, 100) >= 5 * @kura_max_segment_bytes`, or at least a comment acknowledging the floor.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -397,10 +405,83 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
...
+      div(usable * 97, 100)
+      # Too small to carve a ring out of once staging is reserved. Emit nothing
+      # and let Kura fall back to its own statvfs default, which on a volume this
+      # size is both enforced and more conservative than anything derived here.
+    end
+  end
+
+  defp parse_storage_quantity(value) do
```
esnunes, line 504: Small robustness thought: when `parse_storage_quantity/1` can't parse a value (a fractional "1.5Ti", an SI "500M", a typo), `cas_capacity_env/1` silently emits nothing and the region falls back to the statvfs box-derived budget — the exact behaviour this PR is fixing. Since region specs are compile-time constants, this feels like a case for failing loudly (raise while building the manifest, or at least a warning log) rather than a silent fallback. Related: there's no upper bound, and a value over 2^64 would render a string Kura's config rejects ("must be a valid u64"), which would crash the pod at boot. Definitely not a blocker — every current value parses fine — just guarding the regression path.

server/lib/tuist/kura/regions.ex
```diff
@@ -175,16 +175,18 @@ defmodule Tuist.Kura.Regions do
...
-  #     kata Pods, so it serves `:linux` only. The staging Mac minis
-  #     CAN reach it over the tailnet, but they sit in Scaleway fr-par
-  #     — their cache lives in `scw-fr-par-runners`, the same region
-  #     spec production uses, so staging exercises the co-located
-  #     topology rather than a WAN-crossing one.
+  #     handing its URL to a Linux runner would route cache traffic
+  #     across the WAN, which is worse than the public ingress it's
+  #     meant to replace.
+  #
+  # Linux runners have no private cache region. `hetzner-staging-runners`
```
esnunes, line 182: Heads-up on sequencing the reap of the 12 orphaned staging instances relative to this merge: with the spec gone, `Regions.fetch/1` returns `:not_found` for those rows, which has two consequences worth knowing. (1) The app can still soft-reap the DB rows — `reconcile_destroying_server/1` treats the region-level `:not_found` from `Provisioner.current_image_tag/1` like a missing CR and marks them destroyed — but it can no longer reach the staging cluster's KuraInstance CRs/PVCs at all, since every `Provisioner` call dispatches through `Regions.fetch/1` and the cluster context for that region went with the deleted spec; that cleanup is kubectl-only now. (2) Until they're reaped, those rows stop being filtered as private in the dashboard (`Regions.private?(Regions.get(...))` is now false), so they'll render in staging's cache/account-settings pages with the raw region id as label. Nothing to change in the diff — just suggesting the reap happens around the merge.

server/lib/tuist/kura/regions.ex
```diff
@@ -198,6 +200,22 @@ defmodule Tuist.Kura.Regions do
...
+      # the box tops out near five, so raising this or adding accounts past that
+      # needs a second box, not a bigger number.
+      #
+      # Kept separate from storage_size because raising the claim is not a no-op:
+      # the controller patches existing PVCs up to spec.storageSize on every
+      # reconcile, and scw-local-nvme sets allowVolumeExpansion: false, so the API
+      # would reject each resize and wedge the instances that already exist. Only
+      # meaningful where the claim is a fiction — leave it unset on a class that
+      # enforces the claim, so the ring stays inside the volume.
+      cas_capacity_size: "150Gi",
```
esnunes: Tiny naming thought, definitely not a blocker: `cas_capacity_size` holds the whole disk envelope (ring + staging + index), but the value that actually lands in `KURA_CAS_CAPACITY_BYTES` is the envelope minus reserves — for the 150Gi setting, a 137GiB ring. The comments explain it well; the name alone suggests it sets the ring itself. Something like `disk_envelope_size` / `data_budget_size` might spare a future reader the misread.

esnunes (approved):
This is a really strong fix — the root-cause analysis is thorough and the PR description is a model writeup. I traced the change end to end and verified the key claims: the mirrored constants match kura/src/constants.rs (8GiB DEFAULT_TMP_DIR_MAX_BYTES, 512MiB MAX_SEGMENT_BYTES), the kura-controller does place KURA_TMP_DIR inside the data dir and nothing managed sets KURA_TMP_DIR_MAX_BYTES, the test arithmetic is exact, and the 150Gi envelope lands on a 274-segment (137GiB) ring — right on what the pool holds today. The cas_capacity_size/storage_size split correctly avoids the PVC-resize wedge, and the deleted region's Regions.fetch call sites all degrade gracefully (the reconciler's "no longer in the catalog" path, nil-safe private?/1, label fallbacks).

The inline notes are mostly small. The one I'd act on: Kura floors the ring at 5 segments (2.5GiB), so the "emit nothing when too small" threshold has a narrow band of declared sizes where the emitted budget understates actual use — latent today (only the local kind region sits in it), cheap to guard. The rest are suggestions: loud-fail on unparseable quantities, sequencing the staging-orphan reap around the merge (the app can no longer reach their CRs), and a naming nit on cas_capacity_size. One deploy note, nothing to change: the revision bump re-applies fleet-wide in about one reconciler tick, so pods roll across all regions together — same pattern as the 07-09 mesh-peers bump. One doc note: the PR body's residual-risk section still says the hetzner-staging-runners deletion was "left out here", but the last commit does include it — worth a quick description refresh.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
fortmarek: Confirmed and fixed in a340275a1e — I reproduced the band independently: 9Gi and 10Gi both derive a sub-floor budget, get clamped to a 2.5GiB ring, and land at 11.00GiB against the volume. `cas_capacity_bytes/1` now emits only when the budget clears `@kura_segment_ring_floor_bytes`, so the value we declare is the one Kura honours verbatim. One wrinkle your note surfaced: guarding *alone* would have regressed `local-controller` (the only in-band region), since emitting nothing sends it to statvfs and sizes the dev ring at half the developer's disk — worse than the 2.5GiB clamp it gets today. So it now declares a 12Gi envelope and emits a bounded 3GiB ring. A new test asserts every registered region clears the floor and fits its envelope.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
fortmarek: Agreed and fixed in a340275a1e — the silent fallback landed on exactly the statvfs behaviour this PR exists to prevent, which made it the worst possible failure mode. `parse_storage_quantity!/2` now raises an ArgumentError naming the region and the value; specs are compile-time constants, so a typo fails in CI rather than in production. I kept the genuinely-absent case (`size` nil, for self-hosted peers that carry their own disk) as a legitimate no-emit, so only unparseable values are loud. On the u64 bound: rather than a magic ceiling constant, the new every-region test asserts each derived capacity is u64-representable, which covers the overflow path without inventing a limit.

server/lib/tuist/kura/regions.ex
fortmarek: No code change, but flagging that the reap is already done, in the order you suggest: I deleted all 12 stranded CRs on staging ahead of the merge (27→15 instances, 0 leftover pods/PVCs, other regions intact). Your point (1) matched what I hit — the routine staging context 403s on `kurainstances` delete and `rpc` fails there with `node is not alive`, so it was kubectl-only via break-glass regardless. Worth recording that the trap is real: `kura-kura-smoke-staging-dedicated-eu-central-1` merely *contains* "staging" and lives in eu-central, so I matched on `spec.region` rather than the name; it survived. Your point (2) is the one I had not considered — I had verified the leftover `kura_servers` rows cannot *crash* anything (both `Regions.get` sites in `mesh.ex` have `_ -> nil` fallbacks, `Provisioner` short-circuits on `Regions.fetch`, and the UI renders the region as a plain string), but visibility is a separate claim from safety and you are right that `private?` no longer filters them. They are staging-only and cosmetic, so I have left them rather than widen this PR; happy to purge them separately.

server/lib/tuist/kura/regions.ex
fortmarek: Taken in a340275a1e — renamed to `disk_envelope_size` throughout (9 sites). Your read was exactly right and the numbers back it: the 150Gi setting derives a 137.25GiB budget and a 137.0GiB ring, so the field was a whole envelope wearing the ring's name. I also extended the comment at the scw-fr-par-runners override to say outright that the ring is what is left after the reserves, since the name alone was doing too much work there.

PR #11888, opened 2026-07-16
feat(kura): make membership churn and discarded bootstrap completions observable

esnunes:
## What

Makes the silent membership/bootstrap transitions observable, three instruments plus logs:

- **`kura_membership_generation`** (gauge) — the membership view generation, exported every second by `sync_runtime_metrics`. A climbing value means the peer set is flapping.
- **`kura_membership_peer_changes_total{change="discovered"|"lost"}`** (counter) — peers entering/leaving the membership view.
- **`kura_bootstrap_completions_discarded_total`** (counter) — completed bootstrap passes that didn't count toward the readiness gate because the peer left the view or the epoch was reset mid-pass.
- **Logs at the same sites**: `info!` on discovered and lost peers — membership changes are routine (peers drop on every rolling deploy/scale-down), so the `{change="lost"}` counter is the alerting primitive; `warn!` on discarded pass completions, which is the actual anomaly.

## Why

During the 2026-07-16 production incident, `kura-tuist-eu-central-1` sat at 0/2 ready for 4+ hours while both replicas re-downloaded the legacy scw peer's **static** ~2 GB / 85k-artifact dataset **~16 times** (1.38M bootstrap fetches served, all 200s) — and emitted **zero log lines** doing it. Every loud failure path (`bootstrap from … incomplete/failed … will retry`) was ruled out via Loki; what remained were the silent paths in the heartbeat-driven membership machinery (#11725):

- `apply_membership` (`src/state.rs`) rebuilds `known_peers` from each 2.5s `/status` poll and applies it unconditionally — **one missed poll** deletes the peer's `bootstrapped` mark and re-enters it into the bootstrap gate when it answers again;
- `note_bootstrap_succeeded` discards a finished pass if the peer isn't in `known_peers` at completion time (or the epoch was reset);
- neither path logged, and `generation` — the one value that directly reveals the churn — was not exported.

This is self-sustaining under load: a peer busy serving ~100 req/s of bootstrap fetches occasionally misses a poll → flap → marks dropped, fresh passes spawn → sustained load → more misses. Rolling updates never trigger it (one pod bootstraps at a time); a simultaneous cold boot of both replicas (taint eviction after a node flap) did.

Full diagnosis: [incident report](https://artifacts.nunes.dev/a/019f6c40-269d-739b-9deb-40b56d13969c/kura-incident-report.html) (tuist.dev domain access).

## Why logs *and* metrics

During the incident the reliable signal for the stuck replicas was their **container logs** (verified flowing through Loki), so these log lines make the diagnosis available the moment an instrumented image reaches even one stuck replica:

```
{cluster="tuist-production", namespace="kura", service_name="kura"}
  | pod=~"kura-tuist-eu-central-1-.*" |~ "membership changed|completion.*discarded"
```

The `kura_membership_generation` gauge also enables an alert on sustained generation growth.

## Rollout safety

Additive only: new metrics, new log lines, one internal function now returns `bool`. No wire-format, on-disk, or protocol changes; safe under one version skew.

## Validation

- `cargo test --lib` — 346 passed (includes new state-path tests for the membership counters and the discarded-completion path, plus render assertions with exact metric values)
- `cargo clippy --all-targets -- -D warnings` — clean
- `mise run format -- --check` (rules_rust rustfmt aspect) — clean

### DIFF
--- a/kura/src/state.rs
+++ b/kura/src/state.rs
@@ -11,6 +11,8 @@ use tokio::{
     time::{Duration, Instant},
 };
 
+use tracing::{info, warn};
+
 use crate::{
     analytics::Analytics,
     bandwidth::BandwidthLimiter,
@@ -255,13 +257,16 @@ impl ReadinessState {
         true
     }
 
-    fn note_bootstrap_succeeded(&mut self, peer: &str, epoch: u64) {
+    fn note_bootstrap_succeeded(&mut self, peer: &str, epoch: u64) -> bool {
         self.bootstrap_inflight_peers.remove(peer);
         // A completion from a pass started before the last progress reset does
         // not count as bootstrapped; the peer re-enters peers_needing_bootstrap
         // and gets a fresh pass.
         if epoch == self.bootstrap_epoch && self.known_peers.contains(peer) {
             self.bootstrapped_peers.insert(peer.to_string());
+            true
+        } else {
+            false
         }
     }
 
@@ -358,7 +363,7 @@ impl AppState {
     ) -> MembershipUpdate {
         let known_peers = peer_nodes.keys().cloned().collect::<BTreeSet<_>>();
 
-        {
+        let membership_update = {
             let mut readiness = self.readiness.lock().await;
             let membership_update = readiness.apply_membership(
                 members,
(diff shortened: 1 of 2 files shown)


kura/src/state.rs
```diff
@@ -255,13 +257,16 @@ impl ReadinessState {
         true
     }
 
-    fn note_bootstrap_succeeded(&mut self, peer: &str, epoch: u64) {
+    fn note_bootstrap_succeeded(&mut self, peer: &str, epoch: u64) -> bool {
```
pepicrft, line 260: Could we track each bootstrap attempt with an identity or captured membership generation? A lost peer is removed from `bootstrap_inflight_peers`, and rediscovery can spawn a second detached task while the first one still runs or waits for the semaphore. Because the permit is acquired inside the spawned task, the semaphore bounds active transfers but not queued tasks. An older completion can also remove the marker for the newer attempt here. This behavior predates this observability change, so I do not consider it blocking for this pull request, but it should be handled in a follow-up to keep churn resource-bounded.

pepicrft (approved):
Validated the new metrics and state transitions locally. The added in-process resources are bounded in the current call graph: one gauge, one counter, and exactly two fixed membership label values. The targeted Bazel state tests and the metric rendering test pass.

One follow-up is called out inline for the pre-existing detached bootstrap task queue under repeated membership flaps.

The description should also be updated: lost-peer logs are now `info`, not `warn`. The statement that unready pods drop out of metrics scraping is not supported by the current repository configuration. Controller-managed and chart-managed Kura pods already carry direct scrape annotations, and annotation discovery is enabled. If production behaved differently, it would be useful to link the live target evidence.

kura/src/state.rs
esnunes: I took a note about this and I will address in a follow up PR adding more observability to the bootstrap/replication/rollout of kura

PR #11892, opened 2026-07-17
test(cli): temporary cache-volume warm-path validation (do not merge)

fortmarek:
**Throwaway — do not merge, do not review.** Delete once the per-account runner cache volume (spec #76) is validated.

## Why

Validating the cache-volume fix ([#11884](https://github.com/tuist/tuist/pull/11884)) needs a job that actually runs `tuist cache` — it's the only one that moves downloaded xcframeworks into `tuist/Binaries/<hash>`, the exact write that failed in production:

```
Failed to download VendoredCSystem … "VendoredCSystem.xcframework" couldn't be moved to "64f3307f…"
```

But `cli.yml`'s `Cache` job is gated:

```yaml
if: github.ref == 'refs/heads/main'
```

So **no pull request ever runs it.** A throwaway PR without this harness would go green while exercising nothing — a false green, which is the failure mode that has already bitten this workstream three times (a cold staging run declared "warm validated", a root-only chmod, a root-only write probe).

This mirrors `cli-cache` step for step minus the main-only gate, so the warm path can be proven on a throwaway branch instead of on the team's CI.

## What it does

1. **Run 1 (cold):** no master on the mini yet → `tuist cache` populates `tuist/Binaries` → promoted to the host's master.
2. **Run 2 (warm):** volume affinity lands it on the same mini → the host clonefiles that master in → `tuist cache` must move artifacts into the **cloned** `Binaries/`. That's the write that broke.

It also dumps the share's ownership/mode and write-probes `Binaries/` directly before the build, so a failure names the cause rather than surfacing as an opaque CLI error — the original bug was diagnosed only from a CLI message that misreported a failed `mkdir` as "parent directory doesn't exists" and sent the investigation to the wrong layer entirely.

### DIFF
--- a/.github/workflows/tmp-cache-volume-validation.yml
+++ b/.github/workflows/tmp-cache-volume-validation.yml
@@ -1,0 +1,104 @@
+# TEMPORARY — DO NOT MERGE.
+#
+# Validation harness for the per-account runner cache volume (spec #76).
+#
+# Why this exists: cli.yml's `Cache` job is gated `if: github.ref ==
+# 'refs/heads/main'`, so no pull request ever runs `tuist cache`. That is the
+# only job that moves downloaded xcframeworks into `tuist/Binaries/<hash>` —
+# precisely the write that failed in production when the cache volume was
+# enabled:
+#
+#     Failed to download VendoredCSystem … "VendoredCSystem.xcframework"
+#     couldn't be moved to "64f3307f…"
+#
+# Validating the fix on a PR without this job would exercise nothing and report
+# a false green. This mirrors cli-cache step for step, minus the main-only gate,
+# so the warm cache path can be proven on a throwaway branch instead of on the
+# team's CI.
+#
+# Delete this file once the volume is validated.
+name: TMP Cache Volume Validation
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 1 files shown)


fortmarek:
Closing: validation complete, and it found a blocker deeper than the permission fix.

**The permission fix ([#11884](https://github.com/tuist/tuist/pull/11884)) works.** With a real warm master (230 binaries, promoted by a live job), the guest saw `whoami: runner (502:20)` — confirming the uid mismatch — every subtree materialized `drwxrwxrwx`, and `couldn't be moved` never appeared. The tree-wide chmod does what it claims.

**But `tuist cache` still fails on a warm share:**

```
NSCocoaErrorDomain 512 "Resources couldn't be copied to Algorithms.framework"
  NSUnderlyingError = NSPOSIXErrorDomain Code=62 "Too many levels of symbolic links"
  dest: <share>/tuist/Binaries/<hash>/Algorithms.xcframework/macos-arm64_x86_64/…/Resources
```

The master held **230 binaries and zero symlinks**, and every cached framework was an **iOS/simulator slice** — those are flat. **macOS frameworks are versioned bundles** (`Resources -> Versions/Current/Resources`), and copying one onto the virtio-fs share ELOOPs. That's why the cold job passed (iOS slices only) and this one failed the moment it needed a macOS slice.

Same family as the earlier virtio-fs NO-GO for CAS. `gib` is back to 0 on production; the fix is deployed and dormant.

PR #11893, opened 2026-07-17
feat(infra): git-track the Product Usage dashboard and point it at live Postgres via PDC

fortmarek:
> No linked issue. This started from the Product Usage dashboard showing empty or stale panels.

## What changed

Two related pieces, both under `infra/`.

**The Product Usage dashboard is now git-tracked.** `infra/grafana-dashboards/product-usage.json` wraps the dashboard (UID `80989bed-af2e-4a2e-98c7-be8b9f30462a`) in the `dashboard.grafana.app/v2` envelope so Git Sync manages it like the other dashboards. `metadata.name` is the existing UID, so this adopts the current dashboard rather than creating a duplicate.

Beyond tracking what was already there, it adds 22 panels so each feature has a consistent trio of views (usage over time, adoption over time, heavy users), plus a new Tuist Runners section:

| Feature | Added | Source |
| --- | --- | --- |
| Module cache | hits over time, active projects, top accounts | ClickHouse `command_events` |
| Xcode cache (CAS) | operations, bytes transferred, active projects | ClickHouse `cas_events` |
| Inspected builds | runs (CI vs total), active projects and accounts, top accounts | ClickHouse `build_runs` |
| Previews | previews, active projects and accounts | Postgres `previews` |
| Bundles | active projects and accounts | Postgres `bundles` |
| Tests | runs (CI vs total), active projects and accounts | ClickHouse `test_runs` |
| Tuist Runners (new) | jobs by outcome, adoption, compute minutes, jobs by fleet, top accounts | ClickHouse `runner_jobs` |

Also in the dashboard: a `Granularity` variable (day/week/month) now drives every over-time panel, the default range moved from 6h to 30d, the Package Registry section is removed (no longer tracked), and two panels that read Postgres tables which are empty in every environment were fixed. `build_runs` and `command_events` now live only in ClickHouse, so "Accounts by Build Runs" was removed as redundant with its ClickHouse equivalent, and "Warm new accounts" was rebuilt on ClickHouse.

**Grafana can now reach the live production Postgres.** Adds the Grafana Cloud Private Data source Connect (PDC) agent as a chart component (`infra/helm/tuist/templates/pdc-agent.yaml`), enabled per environment, so Grafana's built-in Postgres datasource can query the in-cluster CNPG database as the existing CNPG-managed `tuist_ops_ro` role (login + `pg_read_all_data`).

## Why

The dashboard's Postgres datasource, named "Production Database" (`denpzjx6h7280b`), actually points at a legacy Supabase instance that no longer receives writes. That produced a quiet and fairly dangerous failure mode: every time-filtered Postgres panel rendered "No data", while un-filtered Postgres panels happily showed that stale database's all-time totals, and the ClickHouse panels (a live datasource) worked fine. The dashboard therefore looked mostly healthy while reporting from a dead database, including the panel used for billing conversations.

The fix is to point Grafana at the live CNPG database. Grafana Cloud sits outside our network and CNPG is not publicly exposed, hence the tunnel.

## Why this approach over the alternatives

- **Have the server push metrics instead.** Rejected. Metrics only exist from the moment you start emitting them, so every over-time panel would render empty for past ranges and fill in gradually, and backdated samples are not practically backfillable. Metrics also cannot carry the entity-level tables (top accounts, billing contacts) without a high-cardinality and PII explosion, which is the same pattern that has spiked our Grafana metrics cost before. Those tables are much of the point of a BI dashboard.
- **A server-side read-only SQL endpoint consumed by the Infinity datasource plugin.** Built and then reverted within this branch (`31355a8`, reverted by `87d97f8`); it stays in history if PDC ever proves unworkable. It bought no security: it ran its SQL as `tuist_ops_ro` anyway, so a leaked endpoint token and a leaked `tuist_ops_ro` password grant identical read-all access. It cost a third-party plugin dependency, a rewrite of roughly 15 panels' SQL (Grafana macros to `${__from}` interpolation, long to wide format), a 200-row cap inherited from `Tuist.Ops.Database`, and a permanently public read-all-SQL endpoint.
- **PDC plus the built-in Postgres datasource (chosen).** No plugin, no panel rewrites (the existing `$__timeFilter`, `$bucket`, and `time_series` SQL works unchanged), no row cap, and it reuses a least-privilege role that already exists. The cost is one always-on component and a tunnel from Grafana Cloud into the cluster.

## Notable details

- **One PDC network per environment.** Grafana treats every agent on a network as interchangeable (that is its HA model), so sharing a network across clusters would let a canary query be routed through the production agent, where an in-cluster hostname resolves in the wrong cluster or not at all. Each environment gets its own network and its own independently revocable token.
- **`cluster` and `hostedGrafanaId` are stack identifiers, not secrets**, so they live in the chart (`prod-eu-west-2` / `1028740`). Only the signing token is a secret and comes from 1Password via ESO as `PDC_AGENT/password`.
- **Enable order matters.** The agent cannot start without its token, so an environment only flips `pdcAgent.enabled: true` once `PDC_AGENT` exists in its vault. Guards fail the render if it is enabled without an image tag, cluster, or hostedGrafanaId, matching the kura-controller convention of never running a component on `:latest`.
- **The image tag has no `v` prefix.** Upstream's GitHub release is `v0.0.61` but the published Docker Hub tag is `0.0.61`.
- Datasources target the `-ro` replica service so dashboard queries never contend with the primary.

## Validation

- **Staging is deployed and proven end to end.** The ExternalSecret reports `SecretSynced`, the pod is `1/1 Running`, and the agent holds its tunnel: `Authenticated to private-datasource-connect-prod-eu-west-2.grafana.net using "publickey"` followed by `This is Grafana Private Datasource Connect!`.
- The staging Postgres datasource returns **Database Connection OK**, and a real BI panel renders live rows through the tunnel (verified by rendering the panel image, not just the connection test).
- `tuist_ops_ro` is confirmed reconciled by CNPG in staging: `managedRolesStatus.byStatus.reconciled` includes it, and a `passwordStatus` entry proves the password was actually applied. `OPS_DATABASE_PASSWORD` exists in all three vaults.
- Every dashboard query was validated read-only against production Postgres and ClickHouse before authoring.
- `helm template` across all three environments: the agent is absent when disabled, renders with the shared stack identifiers and each environment's own token reference when enabled, and its guards fire on missing values.
- The dashboard resource is valid JSON with all 34 layout element references resolving and no orphan elements.

## Impact

Canary and production have their agents enabled here but are not deployed yet, so their datasources will fail Save and Test until this merges and deploys. That is expected rather than a misconfiguration. Staging is already running.

### How to test locally

The chart changes render without a cluster:

```
  -f infra/helm/tuist/values-managed-staging.yaml \
  --set runnersFleet.runnerImageSemver=26.5 \
  --set runnersFleetLinux.shapeRunnerImage=x \
  --set registry.enabled=false \
  --set kuraController.image.tag=x \
  --show-only templates/pdc-agent.yaml
```

Swap the overlay for `values-managed-canary.yaml` or `values-managed-production.yaml` to check those, and drop `pdcAgent.enabled` to confirm the template disappears.

The dashboard file is a Git Sync resource, so to inspect the raw dashboard extract the spec:

```
jq '.spec' infra/grafana-dashboards/product-usage.json
```

After an environment deploys, verify its agent:

```
kubectl get externalsecret,deploy,pods -n tuist-staging | grep pdc
kubectl logs -n tuist-staging deploy/tuist-tuist-pdc-agent | tail
```

## Follow-ups

- Repoint the dashboard's Postgres panels to the `Tuist Production DB` datasource once production's agent is deployed. No SQL rewrites are needed since these queries were always written against a Postgres datasource.
- Delete the stale `denpzjx6h7280b` "Production Database" datasource. Leaving something named that pointing at a dead database is exactly what produced the original silent wrong-data problem.
- Consider PDC's "limit hosts" setting to pin each tunnel to its Postgres host and port, so the tunnel cannot reach arbitrary in-cluster services.


infra/grafana-dashboards/product-usage.json
```diff
@@ -0,0 +1,5771 @@
...
+                      "version": "v0",
+                      "datasource": {
+                        "name": "denpzjx6h7280b"
+                      },
+                      "spec": {
+                        "dataset": "postgres",
+                        "editorMode": "code",
+                        "format": "table",
+                        "rawQuery": true,
+                        "rawSql": "SELECT\n  \"Accounts\".\"name\" || '/' || \"Projects\".\"name\" AS \"project_identifier\",\n  COUNT(*) AS \"count\"\nFROM\n  \"public\".\"bundles\"\n  LEFT JOIN \"public\".\"projects\" AS \"Projects\" ON \"public\".\"bundles\".\"project_id\" = \"Projects\".\"id\"\n  LEFT JOIN \"public\".\"accounts\" AS \"Accounts\" ON \"Projects\".\"account_id\" = \"Accounts\".\"id\"\nWHERE\n  $__timeFilter(\"public\".\"bundles\".\"inserted_at\")\nGROUP BY\n  \"project_identifier\"\nORDER BY\n  \"count\" DESC,\n  \"project_identifier\" ASC;",
```
pepicrft, line 258: This fails at runtime against the current PostgreSQL schema with `relation "public.bundles" does not exist`. `20260505120000_drop_pg_bundles_table_and_replication_flag.exs` explicitly dropped this table after bundles moved to ClickHouse. The same problem affects the bundle queries on lines 369, 2410, and 3825. These aggregations need to use the ClickHouse data source and join project or account metadata where needed.

infra/grafana-dashboards/product-usage.json
```diff
@@ -0,0 +1,5771 @@
...
+                      "version": "v0",
+                      "datasource": {
+                        "name": "dexgs9hv7rjswd"
+                      },
+                      "spec": {
+                        "editorType": "sql",
+                        "format": 1,
+                        "queryType": "table",
+                        "range": true,
+                        "rawSql": "SELECT toStartOfInterval(enqueued_at, INTERVAL 1 $bucket) AS time,\n  count() AS \"Total\",\n  countIf(conclusion = 'success') AS \"Success\",\n  countIf(conclusion = 'failure') AS \"Failure\",\n  countIf(conclusion = 'cancelled') AS \"Cancelled\"\nFROM runner_jobs\nWHERE $__timeFilter(enqueued_at)\nGROUP BY time\nORDER BY time"
```
pepicrft: This reads physical lifecycle rows from a `ReplacingMergeTree`. I reproduced one job with queued, claimed, running, and completed inserts while merges were paused: this query reported `Total = 4`, while the repository's `argMax(..., updated_at)` logical view reported `Total = 1`. The raw `build_runs`, `test_runs`, and `test_suite_runs` queries have the same issue. In the reproduction, build and test totals doubled, and the test-suite join doubled as well. Please collapse to the latest version before aggregating, using the existing runner analytics `argMax` pattern and `FINAL` or an equivalent latest-row subquery for build and test data.

infra/helm/tuist/templates/pdc-agent.yaml
```diff
@@ -0,0 +1,111 @@
...
+{{- end }}
+{{- if not .Values.pdcAgent.hostedGrafanaId }}
+{{- fail "pdcAgent.hostedGrafanaId is required — take it from Grafana Cloud > Connections > Private data source connections > Configuration Details." }}
+{{- end }}
+# Grafana Cloud Private Data source Connect (PDC) agent.
+#
+# Grafana Cloud is outside our network, so it cannot reach the in-cluster CNPG
+# Postgres directly. This agent dials out to Grafana Cloud and terminates a
+# tunnel, which lets a Grafana *Postgres* datasource query CNPG as the
+# CNPG-managed `tuist_ops_ro` role (login + pg_read_all_data). That keeps the
```
pepicrft, line 17: `tuist_ops_ro` is a member of `pg_read_all_data`, so it is not least privilege for this dashboard. I reproduced that membership in PostgreSQL 16 and confirmed that the role could read synthetic user tokens, encrypted passwords, project tokens, and custom-storage credentials. A leaked Grafana data-source credential would expose every current and future table. Please add a dedicated Grafana role with access limited to the required views and columns.

infra/helm/tuist/templates/pdc-agent.yaml
```diff
@@ -0,0 +1,111 @@
...
+        app.kubernetes.io/component: pdc-agent
+    spec:
+      securityContext:
+        runAsUser: 30000
+        runAsGroup: 30000
+        fsGroup: 30000
+      containers:
+        - name: pdc-agent
+          image: "{{ .Values.pdcAgent.image.repository }}:{{ .Values.pdcAgent.image.tag }}"
+          args:
```
pepicrft, line 87: The pinned `0.0.61` image reports that `-permit-domains` defaults to `all`, and no `PermitRemoteOpen` restriction is passed here. The tunnel therefore reaches any host and port visible from this pod, not only PostgreSQL. Please restrict it to the CloudNativePG read-only service on port 5432. Grafana documents the restriction here: https://grafana.com/docs/grafana-cloud/connect-externally-hosted/private-data-source-connect/configure-pdc/#restrict-accessible-endpoints

infra/grafana-dashboards/product-usage.json
```diff
@@ -0,0 +1,5771 @@
...
+                      "datasource": {
+                        "name": "denpzjx6h7280b"
+                      },
+                      "spec": {
+                        "format": "table",
+                        "range": true,
+                        "rawSql": "SELECT \n  a.name AS account_name,\n  a.billing_email,\n  a.current_month_remote_cache_hits_count,\n  ROUND((a.current_month_remote_cache_hits_count / 200.0) * 100, 1) AS percent_of_limit,\n  CASE \n    WHEN a.current_month_remote_cache_hits_count >= 200 THEN 'EXCEEDED - Needs billing'\n    WHEN a.current_month_remote_cache_hits_count >= 180 THEN 'CRITICAL - 90%+ of limit'\n    WHEN a.current_month_remote_cache_hits_count >= 150 THEN 'HIGH - 75%+ of limit'\n    WHEN a.current_month_remote_cache_hits_count >= 100 THEN 'MEDIUM - 50%+ of limit'\n    ELSE 'LOW - Under 50% limit'\n  END AS billing_priority,\n  a.current_month_remote_cache_hits_count_updated_at AS last_updated\nFROM public.accounts a\nLEFT JOIN public.subscriptions s ON a.id = s.account_id AND s.status = 'active'\nWHERE s.account_id IS NULL\n  AND a.current_month_remote_cache_hits_count > 0\nORDER BY a.current_month_remote_cache_hits_count DESC"
```
pepicrft: The application treats both `active` and `trialing` as current subscriptions. I ran this exact predicate against PostgreSQL with one active account, one trialing account, and one free account: it returned both the trialing and free accounts as unsubscribed. Changing the join predicate to `s.status IN ('active', 'trialing')` returned only the free account. The organization query on line 1738 needs the same correction.

infra/helm/tuist/templates/pdc-agent.yaml
```diff
@@ -0,0 +1,111 @@
...
+metadata:
+  name: {{ include "tuist.componentName" (dict "root" . "component" "pdc-agent") }}
+  labels:
+    {{- include "tuist.labels" . | nindent 4 }}
+    app.kubernetes.io/component: pdc-agent
+spec:
+  # The agent multiplexes every PDC-routed datasource query over one outbound
+  # tunnel; a second replica buys no throughput and just doubles the
+  # connections Grafana Cloud sees. Upstream ships 1 replica for the same reason.
+  replicas: 1
```
pepicrft: The comment is incorrect about additional replicas providing no benefit. Grafana load-balances across agents and fails over when one disconnects, and recommends at least three agents in production. With one replica, a node drain or process failure disconnects every PostgreSQL-backed panel. Please make the replica count configurable and set production to at least three. Reference: https://grafana.com/docs/grafana-cloud/connect-externally-hosted/private-data-source-connect/scalability-and-security/#scale-with-additional-agents

pepicrft (changes requested):
Runtime validation found six issues that need addressing before merge. The reproductions used PostgreSQL 16, ClickHouse 25.8 with the repository's table engines and separate state inserts, the pinned `grafana/pdc-agent:0.0.61` image, and the rendered managed Helm overlays. The managed Kubernetes contexts required fresh Pomerium authentication, so I did not bypass that boundary and did not independently recheck the staging deployment claims in the description.

infra/grafana-dashboards/product-usage.json
fortmarek: Confirmed and fixed: added `FINAL` to the 11 panels reading `build_runs`, `test_runs`, `test_suite_runs`, and `runner_jobs` (all `SharedReplacingMergeTree`); `command_events`/`cas_events` are plain `SharedMergeTree` and untouched. On production the real inflation is ~1.02x, not 2x (last hour of `runner_jobs`: 162 raw vs 155 actual) — the doubling came from your merges-paused repro — but the error is unbounded and worst on fresh data, so this was a genuine bug. `FINAL` now agrees with `uniqExact(workflow_job_id)`.

infra/grafana-dashboards/product-usage.json
fortmarek: Correct and fixed — both panels now use `s.status IN ('active', 'trialing')`, matching `Billing.get_current_subscription/1`, `RetentionPolicy`, and the ops LiveViews. Worth noting for accuracy: it's latent rather than active today — production currently has 0 trialing subscriptions, so old and new predicates both return the same 58 accounts. It would have misfired the moment someone started a trial.

infra/helm/tuist/templates/pdc-agent.yaml
fortmarek: You're right and the comment was wrong — Grafana load-balances across agents and fails over, so one replica made a node drain take every PDC-routed panel with it. `pdcAgent.replicas` is now configurable, production renders 3, staging/canary stay at 1 (single dashboard, brief gap acceptable). Comment rewritten to say replicas are availability, not throughput.

infra/helm/tuist/templates/pdc-agent.yaml
fortmarek: Confirmed via the docs you linked and fixed: the agent applies no destination restriction by default, so the tunnel could reach any host:port from its pod. Added `pdcAgent.permitRemoteOpen`, rendering `-ssh-flag=-o` + `-ssh-flag=PermitRemoteOpen=<host>:5432` pinned to each env's CNPG read-only service (verified live: `tuist-tuist-pg-ro` in `tuist-staging`/`tuist-canary`/`tuist`). Unset leaves behaviour unchanged. Validating the flag form against the live staging agent in a deploy now.

infra/grafana-dashboards/product-usage.json
fortmarek: Not changing this — I believe the premise is incorrect. `20260505120000_drop_pg_bundles_table_and_replication_flag.exs` does not exist on `main` (`git grep` for `drop_pg_bundles` / any `drop table bundles` across `server/priv/repo/migrations/` returns nothing; the nearest `20260505*` migration is `add_baseline_established_at_to_automation_alerts`). The newest bundles migration is `20260504120000_add_replicated_to_ch_to_bundles`, which *adds* `replicated_to_ch` rather than dropping anything. And the table is live in production right now: `SELECT to_regclass('public.bundles')` returns an OID, so these panels do not fail.

Happy to move them if you disagree — bundles does exist in ClickHouse as a `SharedMergeTree`, so a PG→CH migration is plausibly the direction of travel. But I'd rather not rewrite working queries on the basis of a migration I can't find. If you have a reference for it, point me at it and I'll take another look.

infra/helm/tuist/templates/pdc-agent.yaml
fortmarek: Agreed, and my "least privilege" wording was wrong — `tuist_ops_ro` has `inRoles: pg_read_all_data`, so the datasource credential can read every table including `users.hashed_password`, account/project tokens, and storage credentials. That's a real exposure for a credential stored in Grafana Cloud.

I've partially mitigated it here by pinning the tunnel to the CNPG read-only service (#3602000239), but that bounds the tunnel, not the credential's reach — your point stands.

Not fixing it in this PR, deliberately: a dedicated role means a new CNPG managed role + password Secret, a `tuist-grafana-ro-grants.sql` with per-table SELECTs, a 1P item per env, and recreating all three datasources under the new role. That's a self-contained change that shouldn't ride along with the dashboard, and I'd rather not silently half-do it. Flagged to @fortmarek to decide whether to land it before or after this merges.

fortmarek: Reversing my earlier call — @fortmarek asked for this to land here, and it was the strongest point in the review. Implemented as `tuist_grafana_ro`: a CNPG-managed login role with **no `inRoles`**, so it inherits nothing, plus `infra/cnpg/tuist-grafana-ro-grants.sql` granting an exact allowlist of the columns the dashboard reads.

Went **column-level, not table-level**, because your point is sharper than table grants can fix: every table this dashboard touches sits beside a secret in the same row. `users` is the clearest — the dashboard needs only `created_at`, while the table also holds `encrypted_password`, `token`, and `reset_password_token`. Same for `projects.token`, `accounts.s3_secret_access_key`, `organizations.oauth2_encrypted_client_secret`. `ALTER DEFAULT PRIVILEGES` also revokes future tables, so a new migration can't quietly widen it the way `pg_read_all_data` does. Passwords are 1Password-generated per env (`GRAFANA_DATABASE_PASSWORD`), so each env's credential is distinct and independently revocable.

fortmarek: Final state — permissions are now genuinely limited to what the dashboard needs, and my earlier reply predates half of this.

**The role.** `tuist_grafana_ro`: CNPG-managed, `login: true`, and **no `inRoles`** — it inherits nothing, so it starts with zero table access instead of `pg_read_all_data`.

**Column-level, not table-level.** Your finding is sharper than table grants can fix: every table this dashboard reads sits beside a secret in the same row, so table-level `SELECT` would still leak. Concretely withheld: `users.encrypted_password`, `users.token`, `users.reset_password_token`, `users.email`, sign-in IPs, `projects.token`, `projects.slack_webhook_url`, `accounts.s3_secret_access_key`, `accounts.s3_access_key_id`, `organizations.oauth2_encrypted_client_secret`, `subscriptions.default_payment_method`. `users` ends up with exactly one granted column (`created_at`, for the signup curve).

**Applied declaratively, not by runbook.** `Tuist.Release.migrate/0` → `grant_grafana_role`, gated on `TUIST_DATABASE_GRAFANA_ROLE` (chart sets it only for managed CNPG). Every migrate re-asserts the allowlist, so it tracks the schema instead of drifting behind a psql step nobody re-runs; `infra/cnpg/tuist-grafana-ro-grants.sql` is demoted to a bootstrap/restore fallback, matching the processor/registry roles.

Two properties worth calling out: the list is bracketed by `REVOKE ALL ON ALL TABLES` first — revoking table privileges also drops that table's column privileges, so removing a column here actually removes access on the next migrate — and `ALTER DEFAULT PRIVILEGES … REVOKE ALL` last, so a table added by a future migration can't silently become readable by a credential a third party holds. That's the specific failure mode `pg_read_all_data` has.

**Evidence.** Tests assert the exact column set, that no granted column is a secret, and that the fallback SQL hasn't drifted from the Elixir statements (8/8 pass). Staging has the role created and its 1Password-generated password synced (`SecretSynced/True`); the grants apply on the next migrate, currently deploying. Both error directions are covered: a *wrong* column name fails the migrate job rather than passing silently, and a static check across all 21 Postgres panel queries confirms no *missing* column. Passwords are generated per env, so each credential is distinct and independently revocable.

fortmarek:
## Review addressed — 5 of 6 fixed, 1 refuted

Thanks, this was a genuinely useful review. Summary so reviewers don't have to reassemble it from threads. Each has a reply inline.

| # | Finding | Outcome |
|---|---|---|
| [3602000235](https://github.com/tuist/tuist/pull/11893#discussion_r3602000235) | `tuist_ops_ro` has `pg_read_all_data` | ✅ **Fixed** — new `tuist_grafana_ro`, no `inRoles`, column-level allowlist, applied on every migrate |
| [3602000230](https://github.com/tuist/tuist/pull/11893#discussion_r3602000230) | Raw ReplacingMergeTree reads over-count | ✅ **Fixed** — `FINAL` on 11 panels |
| [3602000239](https://github.com/tuist/tuist/pull/11893#discussion_r3602000239) | Tunnel unrestricted | ✅ **Fixed** — `PermitRemoteOpen` pinned per env |
| [3602000245](https://github.com/tuist/tuist/pull/11893#discussion_r3602000245) | 1 replica is a SPOF | ✅ **Fixed** — configurable, production 3 |
| [3602000242](https://github.com/tuist/tuist/pull/11893#discussion_r3602000242) | `trialing` treated as unsubscribed | ✅ **Fixed** — `IN ('active','trialing')` |
| [3602000223](https://github.com/tuist/tuist/pull/11893#discussion_r3602000223) | `public.bundles` dropped | ❌ **Refuted** — see below |

**The role finding was the best one** and drove the largest change. It's also worse than stated: table-level `SELECT` wouldn't have been enough either, because every table the dashboard reads sits beside a secret in the same row (`users.encrypted_password`, `projects.token`, `accounts.s3_secret_access_key`, `organizations.oauth2_encrypted_client_secret`). Hence column-level. `users` now has exactly one granted column.

**On the refuted one:** `20260505120000_drop_pg_bundles_table_and_replication_flag.exs` doesn't exist on `main` — no migration drops that table, the newest bundles migration *adds* `replicated_to_ch`, and `to_regclass('public.bundles')` returns a live OID in production. Happy to reconsider with a reference.

**Two magnitude corrections**, since both repros were synthetic: the ReplacingMergeTree inflation is ~1.02x in production, not 2x (`runner_jobs` last hour: 162 raw vs 155 actual) — real and worth fixing, but merges collapse most of it; and the subscription bug is latent, since production currently has **0** trialing subscriptions, so both predicates return the same 58 accounts.

**Also caught by deploying rather than trusting review:** my first `PermitRemoteOpen` attempt CrashLooped the agent (`panic: index out of range [1]` — the parser does `SplitN(flag, " ", 2)` then indexes `[1]`, so it must be one arg with the space inside). Fixed and verified: staging is `1/1 Running` with the tunnel up and restricted.

Re the note about not rechecking staging: fair, that boundary is the right call. Current state is agent up + tunnel restricted + role created + password synced; column grants land with the in-flight deploy.

_🤖 Generated with [Claude Code](https://claude.com/claude-code)_

pepicrft (approved):
Confirmed the addressed findings against the current head, including runtime checks for ClickHouse lifecycle deduplication, subscription handling, the restricted Private Data Source Connect tunnel, replica rendering, and the dedicated PostgreSQL role.

The remaining bundle-query migration mismatch and first-deployment role ordering are valid.

PR #11894, opened 2026-07-17
fix(infra): reclaim macOS host slots and detect wedged runner jobs

fortmarek:
Found while investigating a `tuist-macos` job that stayed queued on GitHub for **4h08m** while macOS runners sat Running and idle. Two detection gaps that let it run unnoticed, and three defects wasting host slots on the 9-slot macOS fleet.

## What actually caused the 4h

Not the floor defects — those are capacity hygiene, and they're fixed here because they're real, not because they caused this. The allocator gave 26-5 `desired = 1` for the entire four hours and total desired was 8 of 9, so **a host was spare**. `claims_count` stayed 0. The job was never starved of an allocation or of capacity.

The wait happened between `spec.replicas = 1` and a booted VM, and **every signal was green throughout**:

- `tart_kubelet_pod_provision_delay_seconds` logged **zero provisions at 01:00 and 02:00, then exactly one at 03:00** — 26-5's Pod finally starting at 02:48 after ~4h. It's observed on the path to `Run`, so a Pod that never boots contributes nothing.
- Its top bucket was **600s**, and prod p99 sits on that ceiling through every busy hour — the "rising p90 = pull wave" signal it was added for cannot rise past its own ceiling.
- The Node kept heartbeating **Ready**, and `kube_pod_status_unschedulable` stayed **0**: the Pod *was* scheduled. It just never booted.

It was **not** a dropped webhook. A JIT config is minted only after `pick_queued` returns a candidate read from ClickHouse, so no row would mean the runner never registers and the job never starts — yet it did start, on a 26-5 runner. Its sibling, created in the same webhook batch, dispatched in 15s.

## 1. Two gauges for the states nothing could see

**`tuist_runners_queue_oldest_age_seconds{fleet}`** (server) — age of the oldest still-queued job. `queue_length` measures depth: a fleet steadily serving arrivals sits at 1 forever and looks identical to one job wedged for hours. `job_queue_time_milliseconds` is emitted **on claim**, so an unclaimed job emits nothing, and it tops out at 300s — even retrospectively a 4h wait and a 6m wait are indistinguishable (p99 pins at exactly 300). Rides the existing `queue_length` event and poll, so no extra ClickHouse query per tick.

**`tuist_runners_pool_oldest_pending_pod_age_seconds{pool}`** (controller) — age of the oldest un-`Running` Pod, the boot path's equivalent. Rides the reconciler's existing Pod list and its metrics defer, so no extra API call. Timestamps are tracked per Pod name rather than as a running max: reaping the oldest has to reveal the next-oldest, not leave a stale peak.

Both clamp negatives so a backwards clock can't read as healthy, and both report 0 rather than skipping so they drain instead of holding a stale peak.

**Alert "Runner queue age"** (folder Alerts, group Runners, uid `afsd33dx5fhmoc`): `max by (fleet) (tuist_runners_queue_oldest_age_seconds{env="production"}) > 1800` for 5m. `max by (fleet)` is required, not stylistic — polling gauges are emitted by every server replica and prod runs 3, so a bare aggregate triple-counts. Threshold from measured behaviour: healthy waits peak at ~3.3m, the worst day's p90 was ~11m, the incident ran 106m then 248m. It would have fired at ~23:15 instead of never. Sits NoData → OK until this ships, then activates on its own.

Also widens `tart_kubelet_pod_provision_delay_seconds` and `..._vm_provision_work_seconds` to 1200/1800/3600 so observations past ten minutes stop being censored. Nothing queries them today, so no dashboard or alert changes.

## 2. `minWarmPoolFloor` was double-counted

`DesiredReplicas` counted it twice: as a lower bound inside `floor = max(MinWarmPoolFloor, P95ConcurrentLastHour)`, and again as additive slack in `desired = max(load, floor) + MinWarmPoolFloor`. An idle pool sized to `2 * MinWarmPoolFloor`.

Dropping it from the max is enough — because the slack is additive, `desired >= MinWarmPoolFloor` already holds unconditionally. Every other documented behaviour is preserved. The only changed test is the case whose name already said what it should be: *"idle with no history uses MinWarmPoolFloor"* now wants 3, not 6.

## 3. `minWarmPoolFloor: 0` was silently becoming 1

`omitempty` on a field whose CRD default is `1`, so an explicit `0` was dropped from the serialized spec and re-defaulted by the apiserver. Not latent: `RunnerPoolReconciler` Updates the **whole typed object** when it adds the drain finalizer, so the flip lands on a pool's first reconcile and sticks for its lifetime.

The chart renders `0` for 26.5/26.4.1/26.3/26.0.1; production reported **1** for three of them, flat across 16h and several deploys. The pattern is what the mechanism predicts: 26.5 escaped because its `0` arrived after its finalizer already existed, and 26.6 kept its `3` because non-zero values aren't dropped. Only `0` is affected.

`scaleDownCooldownSeconds` has the same shape (default 300, `minimum: 0`) and is fixed alongside. `enabled` and `maxReplicas` default to their zero values and keep `omitempty`. The rule is documented on the struct and pinned by a test that fails with the old tag. No CRD regeneration needed — `omitempty` is a Go serialization concern and optionality comes from the explicit `+optional` markers.

## 4. The 26.6 floor was sized on a false premise

A floor doesn't follow the catalog default on its own: `tuist-macos` resolves through each account's protected `macos` profile, which stores `xcode_version` at account creation. Moving `default: true` to 26.6 in #11822 didn't move where existing accounts' jobs landed — 143 of 147 macOS jobs in a 10-hour window went to **26-5**, none to 26-6. The floor was warming a pool nothing dispatched to.

Dropped to **1** (canary and staging already use 1). The tuist account's profile has since been moved to 26.6 and dispatch followed within the hour, so the default now genuinely carries the traffic and the floor is correctly placed. The comment records the ordering constraint: bumping the catalog default only redirects *new* accounts, so existing profiles have to move in the same change.

## Fleet budget impact

| pool | configured floor | effective floor | target | allocated |
|---|---|---|---|---|
| macos-26-6 | 3 | 3 | **6** | **4** |
| macos-26-0-1 | 0 | **1** (drift) | 2 | 1 |
| macos-26-3 | 0 | **1** (drift) | 2 | 1 |
| macos-26-4-1 | 0 | **1** (drift) | 2 | 1 |
| macos-26-5 | 0 | 0 | 1 | 1 |

Targets summed to **13 against 9 hosts**. Reserved floors go from an effective **12 of 9 slots** to **1**, leaving 8 for the allocator.

## Validation

- `go build ./...`, `go vet ./...`, `go test ./...` pass across the controller and tart-kubelet.
- `mix test test/tuist/runners/prom_ex_plugin_test.exs` — 10 passed, including the oldest-of-two-jobs case and the empty-queue zero.
- `helm template` with production values renders 26.6=1 and the rest 0.
- Two tests were verified to **fail** against the bug they guard, not merely pass against the fix: the `omitempty` serialization test, and the reap-reveals-next-oldest test (which reports `4h0m0s, want 30s` against a running max).

## Known gap

`tuist_runners_pool_replicas_observed` has its `fleet` label aggregated away in Grafana Cloud (`fleet: "<aggregated>"`), so desired-vs-observed per pool isn't queryable. The new Pod-age gauge covers the case that mattered here, but that aggregation is worth revisiting separately.

### DIFF
--- a/infra/runners-controller/api/v1alpha1/runnerpool_types.go
+++ b/infra/runners-controller/api/v1alpha1/runnerpool_types.go
@@ -143,14 +143,28 @@ type RunnerPoolSpec struct {
 // its own struct so an absent block (the v1 default) keeps
 // `RunnerPoolSpec` byte-identical to its pre-autoscaling shape on
 // the wire — no `autoscaling: null` noise on every macOS pool.
-// A field here may carry `omitempty` only when its CRD default equals
-// its Go zero value. `omitempty` drops the zero value from the
-// serialized spec, which the apiserver cannot tell from "unset" and so
+// CRD defaults for the two optional fields whose zero value is
+// meaningful. Kept in sync with the `default:` markers in
+// crds/tuist.dev_runnerpools.yaml; the accessors below apply them when a
+// field is unset so the apiserver default and the in-process default
+// never disagree.
+const (
+	defaultMinWarmPoolFloor         int32 = 1
+	defaultScaleDownCooldownSeconds int32 = 300
+)
+
+// A field here may keep the plain `int32` + `omitempty` only when its
+// CRD default equals its Go zero value; `enabled` (default false) and
+// `maxReplicas` (default 0) qualify. The two fields whose default is
+// non-zero use `*int32` instead: `omitempty` on a plain `int32` drops a
+// deliberate 0, which the apiserver cannot tell from "unset" and so
 // replaces with the CRD default — and the controller serializes the
-// whole spec whenever it adds the drain finalizer (runnerpool_controller.go).
-// `enabled` (default false) and `maxReplicas` (default 0) are safe;
-// the fields below that declare `minimum: 0` with a non-zero default
-// are not.
+// whole spec whenever it adds the drain finalizer
+// (runnerpool_controller.go). A pointer keeps the distinction: nil
+// serializes to nothing (so the apiserver default applies, and a typed
+// client that never set the field gets it), while a non-nil 0 serializes
+// as 0 (so a deliberate zero survives the round-trip). Read them through
+// the OrDefault accessors, never the raw pointer.
 type RunnerPoolAutoscaling struct {
 	// Enabled flips the autoscaling reconciler on for this pool.
 	// When false, the controller leaves `spec.replicas` alone.
@@ -167,10 +181,11 @@ type RunnerPoolAutoscaling struct {
 	// and uncontended Linux pools honor it as a floor; real load
 	// (claimed + queued) is always funded above it.
 	//
-	// No `omitempty`: 0 ("this pool holds no warm capacity") is a valid
-	// and useful setting, and the CRD default is 1.
+	// Pointer so a deliberate 0 ("this pool holds no warm capacity")
+	// survives serialization; nil means unset and defaults to 1. Read
+	// via MinWarmPoolFloorOrDefault.
 	// +optional
-	MinWarmPoolFloor int32 `json:"minWarmPoolFloor"`
+	MinWarmPoolFloor *int32 `json:"minWarmPoolFloor,omitempty"`
 
 	// MaxReplicas is the hard ceiling on the autoscaler-driven
 	// `spec.replicas` value. 0 disables autoscaling-driven scale
@@ -182,10 +197,32 @@ type RunnerPoolAutoscaling struct {
 	// waits between successive scale-down actions for this pool.
 	// Anti-thrash guard.
 	//
-	// No `omitempty`: 0 ("scale down as soon as demand drops") is a
-	// valid setting, and the CRD default is 300.
+	// Pointer so a deliberate 0 ("scale down as soon as demand drops")
+	// survives serialization; nil means unset and defaults to 300. Read
+	// via ScaleDownCooldownSecondsOrDefault.
 	// +optional
-	ScaleDownCooldownSeconds int32 `json:"scaleDownCooldownSeconds"`
+	ScaleDownCooldownSeconds *int32 `json:"scaleDownCooldownSeconds,omitempty"`
+}
+
+// MinWarmPoolFloorOrDefault returns the configured floor, or the CRD
+// default when the field is unset (nil). The apiserver normally fills
+// the default in on admission, so nil is the belt-and-suspenders path —
+// but reading through this accessor means a nil never reaches the
+// allocator as a 0 and silently zeroes a pool's warm floor.
+func (a *RunnerPoolAutoscaling) MinWarmPoolFloorOrDefault() int32 {
+	if a == nil || a.MinWarmPoolFloor == nil {
+		return defaultMinWarmPoolFloor
+	}
+	return *a.MinWarmPoolFloor
+}
+
+// ScaleDownCooldownSecondsOrDefault returns the configured cooldown, or
+// the CRD default when unset (nil).
+func (a *RunnerPoolAutoscaling) ScaleDownCooldownSecondsOrDefault() int32 {
+	if a == nil || a.ScaleDownCooldownSeconds == nil {
+		return defaultScaleDownCooldownSeconds
+	}
+	return *a.ScaleDownCooldownSeconds
 }
 
 // RunnerPoolRollout carries the image-roll throttle knob. Its own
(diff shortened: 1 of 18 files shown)


server/lib/tuist/runners/prom_ex_plugin.ex
```diff
@@ -230,6 +230,17 @@ defmodule Tuist.Runners.PromExPlugin do
             description: "Queued workflow jobs per fleet (ClickHouse runner_jobs).",
             measurement: :count,
             tags: [:fleet]
+          ),
+          # Rides the `queue_length` event: both values come out of one
+          # ClickHouse scan. Depth alone can't distinguish a queue that
+          # is never empty because arrivals are served promptly from one
+          # job wedged for hours — both sit at 1.
+          last_value(
```
pepicrft, line 248: The new queue-age metric can retain a stale non-zero value after a fleet disappears. If a pool is deleted while it still has a queued row, the fleet remains in stats and emits an increasing age. Once the final row leaves queued, the fleet is in neither stats nor active_fleets, so this loop emits nothing and last_value keeps the previous value indefinitely. That can leave the Runner queue age alert firing after the queue has cleared. Please track previously emitted queue fleets, like execute_pool_replicas_telemetry_event/0 already does, and emit one final zero for fleets that disappear. A two-poll regression test covering queued to cleared after pool deletion would pin the behavior.

infra/runners-controller/api/v1alpha1/runnerpool_types.go
```diff
@@ -158,8 +166,11 @@ type RunnerPoolAutoscaling struct {
 	// contention to admit another shape's real queued work. macOS pools
 	// and uncontended Linux pools honor it as a floor; real load
 	// (claimed + queued) is always funded above it.
+	//
+	// No `omitempty`: 0 ("this pool holds no warm capacity") is a valid
+	// and useful setting, and the CRD default is 1.
 	// +optional
-	MinWarmPoolFloor int32 `json:"minWarmPoolFloor,omitempty"`
+	MinWarmPoolFloor int32 `json:"minWarmPoolFloor"`
```
pepicrft: Making these scalar fields always serialize preserves explicit zero, but it also removes the distinction between zero and unset. A typed Go client that builds RunnerPoolAutoscaling{Enabled: true, MaxReplicas: 9} without setting minWarmPoolFloor or scaleDownCooldownSeconds now sends both as zero, so the Kubernetes defaults of 1 and 300 never apply. Helm renders both fields today, but the optional/default contract of the type is still broken for other typed clients. Please use pointers or another optional representation, and test both cases: omitted values remain absent while explicit zero values are serialized.

pepicrft (approved):
Approving the overall change. I left two inline follow-ups around metric draining and preserving unset values in the typed custom resource.

server/lib/tuist/runners/prom_ex_plugin.ex
fortmarek: Fixed in cca5956. The queue poll now tracks previously-emitted fleets in the process dictionary and emits one final zero for any that disappear — the same seen-set pattern `execute_pool_replicas_telemetry_event/0` uses. Covers the count gauge too, which had the same latent staleness. Added the two-poll regression test you described: a fleet with a queued row but no pool emits count=1, then after the row completes and the pool stays gone the next poll emits count=0 / oldest_age_seconds=0.

infra/runners-controller/api/v1alpha1/runnerpool_types.go
fortmarek: Fixed in 21b36f9. Both fields are now `*int32` with `omitempty`, so nil serializes to nothing (apiserver default applies, and a typed client that never set the field gets it) while a non-nil 0 serializes as 0. Reads go through `MinWarmPoolFloorOrDefault` / `ScaleDownCooldownSecondsOrDefault`, which apply the CRD default for a nil field so a nil never reaches the allocator as a raw 0. deepcopy hand-edited for the two pointers (no controller-gen in this module). Tests cover both directions: explicit 0 reaches the wire, unset stays absent, and the decode round-trip lands helm's `minWarmPoolFloor: 0` as non-nil 0 while an absent field stays nil.

server/lib/tuist/runners/prom_ex_plugin.ex
fortmarek: Fixed in cca5956: the queue poll now tracks previously-emitted fleets in the process dict and emits a final zero for any that disappear (same pattern as `execute_pool_replicas_telemetry_event/0`), covering both the count and age gauges, with the two-poll queued→cleared-after-pool-deletion regression test.

infra/runners-controller/api/v1alpha1/runnerpool_types.go
fortmarek: Fixed in 21b36f9: both fields are now `*int32` with `omitempty` (nil→omitted so the CRD default applies, non-nil 0→serialized), read via `…OrDefault()` accessors; tests cover explicit-zero-serialized, unset-absent, and the decode round-trip.

PR #11897, opened 2026-07-17
test(cli): temporary virtio-fs symlink probe (do not merge)

fortmarek:
Throwaway diagnostic for spec #76. **Do not merge or review** — deleted once the cache-volume design is settled.

Measures *why* `tuist cache` ELOOPs on a warm cache volume, and whether the proposed disk-image fix works, **before** any of it is written in Go.

Every check runs against both the virtio-fs share **and** a local-disk control, so a failure is attributable to the share rather than the probe. It prints `readlink` output because a symlink whose stored target comes back wrong is exactly what produces ELOOP — that would pin the mechanism.

It also exercises the proposed fix in place: create a sparse APFS image on the share, `hdiutil attach -owners off`, rebuild the framework layout inside, and measure throughput.

### DIFF
--- a/.github/workflows/tmp-virtiofs-symlink-probe.yml
+++ b/.github/workflows/tmp-virtiofs-symlink-probe.yml
@@ -1,0 +1,132 @@
+# TEMPORARY — DO NOT MERGE. Delete once the cache-volume design is settled.
+#
+# Pins down why `tuist cache` dies on a warm cache volume, and whether the
+# proposed fix actually works — before any of it gets written in Go.
+#
+# Observed in production (spec #76, volume enabled):
+#
+#   NSCocoaErrorDomain 512 "Resources couldn't be copied to Algorithms.framework"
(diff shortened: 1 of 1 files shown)


fortmarek:
Probe complete — it disproved my hypothesis and validated the fix. Results in the linked session; summary:

**Symlinks are NOT the problem.** On the virtio-fs share, `ln -s` succeeds, `readlink` returns the exact target, resolution works, and reads through the link work — identical to the local-disk control.

**The actual failure is extended attributes on symlinks:**
```
cp: Bar.framework/Resources: could not copy extended attributes to <share>/Bar.framework/Resources
```
`cp -R` (and FileManager.copyItem, which is what CacheLocalStorage.store calls) copies xattrs; doing that against a symlink on virtio-fs fails, and surfaces as ELOOP through Foundation.

**The disk-image fix works, with room to spare:** `hdiutil create` + `attach -owners off` on the share both succeed, the full framework bundle copies in cleanly, and block I/O runs at **1.57 GB/s** (512MB in 0.34s). `-owners off` also makes the host/guest uid mismatch irrelevant.

PR #11902, opened 2026-07-17
feat(server): support auth.md agent setup for Gradle builds

pepicrft:
This makes the natural “my Android project is slow to build, optimize it with Tuist” workflow reliable for unauthenticated coding agents. The server now implements the [auth.md protocol](https://workos.com/auth-md), teaches agents the complete Gradle integration sequence, and exposes enough read-back tooling to prove that the optimization works.

### What changed

- Implemented deployment-local auth.md discovery, anonymous and service-authenticated registration, signed identity assertion exchange, browser claims, credential polling, token revocation, provider events, credential expiry, audit records, rate limits, and signing-key rotation.
- Preserved the legacy agent registration routes while moving new clients onto the protocol-defined credential lifecycle.
- Replaced the standalone claim-page markup with Tuist's existing authentication shell and Noora design-system components. Completion and repeated-visit states follow the established device-connection pattern without separate status badges.
- Added `get_gradle_integration_guide` and `list_accounts` to the [Model Context Protocol](https://modelcontextprotocol.io/) server, moved the Gradle workflow into one shared guide, and advertised durable initialization instructions.
- Fixed the Gradle plugin so the server address resolved from `tuist.toml` reaches shared configuration and background uploads, and made authentication failures print the exact server-specific login command.
- Updated the plugin version to `0.10.0`, expanded the Gradle and agent documentation, and added isolated local benchmark settings.
- Added a public product changelog entry for the complete agent-authentication lifecycle.

### Why

Current headless clients do not consistently expose an unauthenticated protocol-server response to the model. Without deployment-local discovery, they can drift to stale hosted documentation, start generic browser authorization, infer account details, or edit the Android project before the required user handoff.

The goal is a deterministic workflow in which the agent discovers authentication from the configured Tuist deployment, asks for unavoidable user input, creates the correct Tuist project, installs the Gradle plugin, and verifies remote-cache behavior before reporting success.

### Root cause

The existing agent registration flow predated the full auth.md credential lifecycle. It could guide a custom client, but it did not expose the protocol's signed assertion exchange, standard revocation and provider-event routes, durable credential records, or a complete deployment-local document.

The Gradle instructions were also split between prompts and hosted documentation. Separately, the plugin published its extension address before resolving `tuist.toml`, allowing background work to fall back to the hosted server after a local build.

### Approach

The implementation treats the local auth.md document and discovery metadata as authoritative, retains compatibility routes, and associates a claimed anonymous credential with the confirming user only at the protocol-server boundary. It never derives a user's email address from provider, Git, environment, or session data.

The Gradle prompt and tool now share one guide. Agents must list available accounts before project creation, authenticate the Tuist command-line interface separately from the protocol server, use deterministic foreground uploads for verification, and confirm results through Tuist read-back tools.

### Impact

- New deployments apply one database migration for agent credentials and registration audit fields. The data-export inventory documents the added records.
- The protocol server advertises version `1.14.0`; the Gradle plugin and generated setup guidance use version `0.10.0`.
- Existing legacy agent clients remain supported.
- Local development can assign an isolated object-storage port and disable code reloading during long benchmark runs.

### Validation

- `mise x -- mix test test/tuist/accounts/workers/expire_agent_registration_worker_test.exs test/tuist/mcp test/tuist/oauth/clients_test.exs test/tuist_web/controllers/agent_auth_controller_test.exs test/tuist_web/controllers/auth_md_protocol_controller_test.exs test/tuist_web/controllers/mcp_controller_test.exs test/tuist_web/controllers/well_known_controller_test.exs test/tuist_web/plugs/authentication_plug_test.exs test/tuist_web/rate_limit/agent_auth_test.exs test/tuist_web/rate_limit/mcp_test.exs`: 184 passed.
- `TUIST_SERVER_TEST_PORT=4817 mise x -- mix test test/tuist/docs test/tuist/docs_test.exs test/tuist_web/controllers/docs_login_flow_test.exs test/tuist_web/live/docs_live_test.exs test/tuist_web/router_test.exs`: 64 passed.
- `TUIST_SERVER_TEST_PORT=4818 mise x -- mix test test/tuist_web/controllers/agent_auth_controller_test.exs test/tuist_web/controllers/auth_md_protocol_controller_test.exs`: 17 passed after adopting the shared authentication design.
- `TUIST_SERVER_TEST_PORT=4818 mise x -- mix test test/tuist/marketing/changelog_entry_parser_test.exs`: 3 passed.
- `MIX_ENV=test TUIST_SERVER_TEST_PORT=4818 mise x -- mix run --no-start`: confirmed that the new public changelog entry loads with pull request `11902`.
- `mise x -- mix assets.build`: successful.
- `./gradlew test --no-daemon --console=plain`: successful.
- `mise x -- mix format --check-formatted`: successful.
- `mise run lint`: successful.
- `git diff --check`: successful.

### Browser verification

[Headless Chrome](https://developer.chrome.com/docs/chromium/headless) exercised the real local flow with the seeded development user. It followed the sign-in redirect, confirmed that the claim page did not reveal the agent's six-digit code, submitted the code, reached the success page, repeated the visit to verify idempotency, polled the credential, initialized the protocol server, revoked the credential, and confirmed that the next request returned `401`.

Before authorization:

Claim page before authorization

After authorization:

Claim page after authorization

Repeated visit:

Already-authorized claim page

### Agent benchmark

Claude Code, Codex, OpenCode, and Pi each completed local auth.md discovery, anonymous registration, signed assertion exchange, Gradle-guide discovery, and the exact user handoff when the harness required endpoint inspection before hosted documentation or repository edits.

The earlier post-claim run used Google's [Now in Android](https://github.com/android/nowinandroid) project. All four clients created a Tuist project, installed plugin `0.10.0`, completed upload and clean read-back builds, and verified server-side remote-cache hit rates between 99.6% and 99.8%. These timings are execution evidence rather than a fair client-performance ranking because dependency state and machine load differed between runs.

### How to test locally

1. From `server`, apply the migration and start the development server with `mise run dev`.
2. Request `/mcp` without a credential, follow the advertised `auth_md` document, register, exchange the signed assertion, and start a browser claim.
3. Sign in as `tuistrocks@tuist.dev`, enter the six-digit claim code, poll for the claimed credential, initialize the protocol server, then revoke the credential and verify that reuse returns `401`.
4. In an Android project, configure the local server and project handle in `tuist.toml`, add plugin `0.10.0`, and run `tuist auth login --url http://localhost:8080`.
5. Run one upload build followed by a clean read-back build, then use the protocol tools to verify the build and representative remote task hits.


server/lib/tuist/accounts/agent_auth.ex
```diff
@@ -32,9 +35,404 @@ defmodule Tuist.Accounts.AgentAuth do
...
+      {:error, :expired_token}
+    else
+      {:error, :authorization_pending}
+    end
+  end
+
+  def revoke_protocol_access_token(token) when is_binary(token) do
+    with %JOSE.JWT{fields: %{"jti" => jti}} <- JOSE.JWT.peek_payload(token),
+         %AgentAuthCredential{} = credential <-
+           AgentAuthCredential |> Repo.get_by(jti: jti) |> Repo.preload(:agent_registration),
```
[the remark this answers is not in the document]
pepicrft: Good catch. Fixed in 556f1cc7ce. `revoke_protocol_access_token/1` now runs `Tuist.Guardian.decode_and_verify/1` and reads the `jti` from the verified claims instead of `JOSE.JWT.peek_payload`, so revocation requires a genuinely Tuist-signed token rather than just a known `jti`. Since the `jti` shows up in audit events and logs, a leaked one could otherwise revoke a live session over the unauthenticated `/oauth2/revoke`. Added a regression test that forges a JWT carrying a live credential's `jti` and asserts the session stays active, then confirms revocation still works with the real token.

server/lib/tuist_web/plugs/authentication_plug.ex
```diff
@@ -101,10 +107,38 @@ defmodule TuistWeb.AuthenticationPlug do
         TuistWeb.Authentication.put_current_user(conn, user)
 
       %AuthenticatedAccount{} = subject ->
-        assign(conn, :current_subject, subject)
+        conn
+        |> assign(:current_subject, subject)
+        |> put_claimed_agent_user(subject)
```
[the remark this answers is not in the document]
pepicrft: I dug into this one and I don't think it denies access in practice, so I'm leaving the precedence as is.

The claimed token's `AuthenticatedAccount` isn't a bare "mcp scope" subject. `issue_protocol_access_token/1` signs it with `user_id` set to the confirming user, and `Tuist.Guardian.resource_from_claims/1` rebuilds it with `issued_by: <that user>`, `all_projects: true`, and `scopes: ["mcp"]`. The `mcp` scope expands to `project:builds:read` and the other `project:*:read` scopes via the `AccountToken` scope groups, so `build:read` matches `allow([:authenticated_as_account, scopes_permit: "project:builds:read"])`.

That check runs `scopes_permit/3`, which for a `Project` also calls `project_access_permitted/2`. With `issued_by: %User{}, all_projects: true` that delegates to `owns_account_or_belongs_to_account_organization?(user, project.account)`, so authorization resolves to the confirming user's real account plus org membership, including private projects in orgs they belong to. `current_subject` winning over `current_user` still evaluates the user's actual reach, not a scope-only grant. The controller test also asserts `issued_by == user.id` on the claimed token.

Happy to revisit if you can point at a project-access shape where the user has a role but `owns_account_or_belongs_to_account_organization?` returns false.

server/lib/tuist/accounts/agent_auth.ex
```diff
@@ -740,6 +1241,445 @@ defmodule Tuist.Accounts.AgentAuth do
...
+      _ -> {:error, :invalid_grant}
+    end
+  end
+
+  defp protocol_registration_user(_registration), do: {:error, :invalid_grant}
+
+  defp ensure_claim_pollable(%AgentRegistration{status: status}) when status in [:expired, :revoked],
+    do: {:error, :expired_token}
+
+  defp ensure_claim_pollable(%AgentRegistration{status: :claimed}), do: :ok
```
fortmarek: **[P1] Stop accepting the claim token after its outer expiry**

This claimed fast path skips `claim_token_expires_at` entirely. Claimed registrations are not expired by `ExpireAgentRegistrationWorker`, so anyone retaining the original `clm_` value can keep polling indefinitely and mint fresh access tokens and v2 assertions long after the outer claim window closes. That contradicts the advertised `expired_token` behavior. Please enforce the outer expiry here, or introduce a separately bounded/consumable result-retrieval grace period.

server/lib/tuist/accounts/agent_auth.ex
```diff
@@ -740,6 +1241,445 @@ defmodule Tuist.Accounts.AgentAuth do
...
+    {:ok, Map.keys(events)}
+  end
+
+  defp validate_security_events(_claims), do: {:error, :invalid_request}
+
+  defp revoke_provider_delegation(claims, audience) do
+    now = Time.utc_now()
+    registrations = list_revocable_provider_registrations(claims, audience)
+
+    Enum.each(registrations, fn registration ->
```
fortmarek, line 1688: **[P1] Make SET replay recording and revocation atomic**

`receive_protocol_event/2` records the SET `jti` during `verify_jwt/4` before reaching this loop, but these credential, registration, and audit writes run outside a transaction. A transient database failure or node exit after the JTI commit can leave credentials active, while the provider's retry is rejected as `replay_detected`. Please commit the replay marker and all revocation writes atomically, or persist a retryable processing state; a bulk update/`Ecto.Multi` would also avoid the request-path N+1.

server/lib/tuist_web/controllers/oauth/token_controller.ex
```diff
@@ -6,13 +6,67 @@ defmodule TuistWeb.Oauth.TokenController do
...
   def token(%Plug.Conn{} = conn, _params) do
     oauth_module().token(conn, __MODULE__)
   end
 
+  def revoke(%Plug.Conn{} = conn, %{"token_type_hint" => token_type_hint}) when token_type_hint != "access_token" do
+    oauth_error(conn, "unsupported_token_type", "token_type_hint must be access_token when provided.")
+  end
+
+  def revoke(%Plug.Conn{} = conn, %{"token" => token}) when is_binary(token) and token != "" do
+    :ok = Accounts.revoke_protocol_agent_access_token(token)
```
fortmarek: **[P1] Revoke the other OAuth token families too**

This route is now advertised as the authorization server's RFC 7009 revocation endpoint, but this helper only recognizes tokens backed by `AgentAuthCredential`. Existing authorization-code access and refresh tokens are stored through Boruta, so this returns 200 without invalidating them; the preceding clause also rejects `refresh_token` outright. Please keep the protocol-credential path and delegate other access/refresh tokens to the Boruta revocation flow with the appropriate client checks.

fortmarek (changes requested):
Found three blocking authentication and revocation issues. Details are inline on the affected code paths.

server/lib/tuist/accounts/agent_auth.ex
pepicrft: Agreed, fixed in 64fb2c94cb. You're right that nothing closed the window: `ExpireAgentRegistrationWorker` only expires registrations still in `:pending` (`expire_protocol_registration/1` matches on that status), so once claimed the `clm_` value kept minting tokens forever.

`ensure_claim_pollable/1` now runs `claim_token_expired?/1` on the claimed path too, so the poll returns `expired_token` after the 30 minute window. I went with enforcing the outer expiry rather than adding a grace period, since the registration keeps its `:claimed` status and already-issued credentials stay valid on their own TTLs plus the identity-assertion refresh path. Only the claim token stops minting new ones.

I also removed the assertion in the anonymous flow test that expired `claim_token_expires_at` and then expected the poll to still succeed, since that was encoding exactly the behavior you're flagging. Replaced it with a dedicated test that completes a claim, expires the window, and asserts the poll returns `expired_token`.

server/lib/tuist/accounts/agent_auth.ex
pepicrft: Good catch, fixed in c3ff1ea93b. Confirmed the ordering: `record_jti/1` runs inside `verify_jwt/4` and committed on its own, then `revoke_provider_delegation/2` did its credential, registration, and audit writes with no transaction around them. So a failure in the middle left credentials active while the provider's retry hit `replay_detected`.

I added a `defer_jti` option to `verify_jwt/4` so the security event path skips the marker during verification and records it itself inside a `Repo.transaction/1` that also wraps the revocation writes. Verification and the JWKS fetch stay outside the transaction so we're not holding a database connection across HTTP. The revocation path raises on failure (the `{:ok, _} = Repo.update()` match and `insert_event!/3`), so a partial run rolls the marker back and the retry is accepted.

On the N+1: I left the per-registration loop as is for now. `list_revocable_provider_registrations/2` is scoped to one issuer, subject, and audience, so it's realistically one or a few rows rather than a bulk set, and folding it into an `Ecto.Multi` mostly buys shape rather than round trips here. Happy to convert it if you'd rather have it uniform.

server/lib/tuist_web/controllers/oauth/token_controller.ex
pepicrft: Agreed, fixed in f887c1f0af. Confirmed the gap: `/oauth2/revoke` is what `well_known` advertises as `revocation_endpoint`, and the app runs a full authorization server (authorize, token, introspect, register), so Boruta-backed access and refresh tokens were reaching a handler that only knew about `AgentAuthCredential` and returned 200 without invalidating them.

`revoke_protocol_access_token/1` now returns `:ok` when it handled a Tuist-signed protocol credential and `:not_found` otherwise. The controller keeps the protocol path directly, since those are public-client tokens with no client secret to present, and delegates everything else to `oauth_module().revoke(conn, __MODULE__)` with `revoke_success/1` and `revoke_error/2` implemented off `Boruta.Oauth.RevokeApplication`. I dropped the `token_type_hint` rejection since RFC 7009 treats the hint as advisory and Boruta handles both families.

One behavior change worth flagging: a token that isn't a protocol credential now goes through Boruta's client checks, so a request without `client_id` gets `invalid_request` instead of a blanket 200. That's the RFC 7009 path for a request that hasn't authenticated its client, but it does mean unknown tokens no longer silently return 200. I updated the forged-token test to assert this, and it still asserts the live session stays active.

fortmarek (approved):

server/lib/tuist/accounts/agent_auth.ex
pepicrft: Following up on the N+1 half of this, since my earlier answer was hand-waved and I wanted to actually check it.

I went to convert the loop to bulk statements, then wrote a test first and it disproved my own premise. I had assumed each id_jag exchange inserts a new claimed registration, which would make the set grow unboundedly per subject. It doesn't: `create_protocol_registration/1` calls `get_user_by_delegation/2`, and when an existing claimed delegation is found for the issuer, subject, and audience it reuses that registration and only issues a new assertion. `create_protocol_provider_registration/1` runs solely when no delegation exists yet. Two exchanges for the same subject produce one row, which is why `get_delegation_registration!/2` can get away with `order_by desc, limit 1` (defensive against step-up and concurrent-insert races rather than steady-state growth).

So the loop is over ~1 row, not a bulk set, and I backed the bulk refactor out rather than trade changeset-based writes for `update_all`/`insert_all` to optimize an N of one. What I did keep is a test pinning the reuse, in c80f4110a7, because nothing asserted it and that assumption is exactly what keeps the loop off a request-path N+1. If the registration path ever stops reusing, that test fails and flags the loop for bulking.

Conclusion is the same as before, but the earlier "realistically one or a few rows" was a guess and this is the actual reason. The atomicity fix in c3ff1ea93b is unchanged.

PR #11903, opened 2026-07-17
feat(server): meter runner usage by machine type

pepicrft:
## What changed

Runner usage is metered as normalized compute units, one Stripe Meter per platform, each normalized to its own baseline machine.

Runner billing sessions persist the platform, processor count, memory allocation, the billing multiplier the session was admitted under, and GitHub's own execution window for the workflow job. Aggregation scales the job's exact runtime by that frozen multiplier and reports compute-unit milliseconds to `runner_linux_compute_unit_milliseconds` or `runner_macos_compute_unit_milliseconds`.

Meter reporting uses a three-level worker flow:

- the daily scheduler chooses one half-open period, from yesterday at midnight up to today at midnight
- each customer worker receives those boundaries, splits them at every service-period boundary falling inside them, snapshots all meter values per resulting window, and fans out one child job per meter
- each child carries the immutable value and its own window, then reports only that Stripe Meter with a period-specific identifier and idempotency key

Runner meters are included automatically without a new feature flag. Configured runner metered Prices are also added to new, changed, and enterprise subscriptions, and existing runner subscription items survive plan changes untouched.

## Why

Runner runtime cannot be billed as an undifferentiated minute, because the cost depends on the machine specification. But a Meter and Price per exact machine shape does not scale either: the initial nine shapes would consume nine of Stripe's [20 classic subscription items](https://docs.stripe.com/billing/subscriptions/quantities), and every future shape would need a Meter, a Price, environment configuration, and a backfill across existing subscriptions.

Multiplying runtime by an immutable machine factor is metering rather than reimplementing pricing. Stripe keeps ownership of the unit's currency amount, tiers, discounts, taxes, and credits, while Tuist reports how much compute the customer consumed. A subscription then carries two runner items no matter how large the catalog grows.

The reporting workers also needed the service period to be chosen once and carried through. The earlier helpers derived "yesterday" independently when they executed, so a delayed child could cross midnight and report a different day. And because a UTC-day aggregate can straddle an arbitrary subscription-item boundary, a single timestamp cannot describe it correctly: the aggregate has to be split at the boundary first.

The cancellation boundary is the case that makes this load-bearing. At `cancel_at_period_end` there is no following invoice to receive shifted final-day usage, so usage stamped after the subscription ended is simply never billed.

The runner-specific feature flag was unnecessary. Reporting a Meter event is safe before the matching Price is configured or attached, and the managed runner Price mappings remain empty until the Stripe rollout is ready.

Prepaid access needs to make a minute on every machine cheaper by the same percentage. Reporting gross usage at each machine's normal rate, then applying a monetary Stripe Credit Grant, preserves those relative prices and avoids maintaining a second credit ledger in Tuist.

### The compute unit

Each platform is normalized to its own baseline machine, and one compute unit is one minute on that baseline: Linux 2 vCPU / 8 GB, macOS 6 vCPU / 14 GB. Within a platform, shapes are weighted by resources:

```
effective_multiplier = resource_multiplier * platform_multiplier
```

The resource weighting is cost-based rather than equal to vCPU count: roughly two thirds of a machine's cost is CPU and one third memory, giving `(8 * vcpus + memory_gb) / 24`, exact at the baseline and linear from there.

There is no platform coefficient. Each platform's premium lives in its own Stripe Price, quoted directly per baseline machine-minute, so the macOS Price describes the real 6 vCPU / 14 GB machine rather than a hypothetical Mac at the Linux baseline.

That separation is what makes the rate card workable. A single global unit forces one Price to fix both platforms at a hardcoded ratio, so choosing a macOS per-minute rate silently determines every Linux rate. It also collapses prepaid terms: a monetary credit grant scopes to a Price, so one Price means one prepaid discount percentage across every machine, while one Price per platform lets Linux and macOS prepaid terms differ.

`runner_sessions` carries a `billing_multiplier` column written at open time, so changing the catalog cannot reprice usage that already happened, and macOS economics can move independently for future sessions without a separate Stripe Price. Rows created before the migration fall back to the current catalog, which is the one case where a rate-card change can move an old session, and it is unavoidable since those rows never recorded what they were admitted under. Raw shape and raw milliseconds stay on the row, so analytics keep full fidelity.

The new runner session columns are nullable so the migration does not rewrite historical rows. During rollout, historical sessions resolve their specification from the fleet catalog. Usage with an unresolvable historical fleet is omitted with a warning, keeping failure biased toward undercharging.

### What is billable

Billing measures the workflow job, not the Pod that ran it. A Pod boots a VM before the job can start and holds the host through post-job cache work and teardown afterwards; on a staging smoke run that was ten minutes of boot around a six-second job. That overhead is Tuist's to optimize rather than the customer's to pay for, and charging for it would hide it, since a slower boot would quietly become revenue instead of a cost.

So `runner_sessions` records GitHub's `started_at` / `completed_at` for the job from the `workflow_job.completed` webhook, and billing intersects the period with that window. The session's own bounds are unchanged and still drive capacity signals, orphan detection, and the autoscaler, which all want Pod occupancy.

A session bills nothing until both bounds are recorded: an in-flight job, a job cancelled while queued, a partial or inverted payload, and a Pod whose completion webhook never arrived all contribute zero. A missing window is not evidence that the customer's work ran, so charging nothing keeps the module's bias toward undercharging. It also retires the six-hour max-lifetime clamp, which only existed to bound an orphaned Pod billing against the wall clock. The trade-off is that a running job contributes zero until it completes, so a mid-period usage figure lags by whatever is still in flight.

### Service periods and reporting

`Billing.usage_windows/3` derives the account's service-period boundaries and slices the reporting day at every one that falls strictly inside it. Boundaries come from `current_period_start`, `cancel_at`, and `ended_at` together, so both a renewal and an end of service are covered, and a day carrying both produces three windows. Discovery reads the account's most recent subscription regardless of status, because at `cancel_at_period_end` the local row is no longer active or trialing yet its final cycle is exactly the one being reported.

Each event is then stamped one second inside the half-open end of its own window, so it lies wholly within one service period and carries exactly the usage that period earned.

An unknown boundary is deliberately not treated as no boundary. `usage_windows/3` returns `{:ok, windows}` or `{:error, reason}`, and a Stripe failure fails the parent job rather than snapshotting the day as one window: the snapshot is permanent, since a child never re-queries usage, so a wrong guess cannot be corrected later. The parent's retries are capped at 12 attempts, about 11 hours with Oban's default backoff, so the chain finishes inside the invoice-finalization grace period. Only a genuine no-subscription account falls back to one whole window, and it has nothing to invoice against.

Child jobs are capped at 12 attempts for a different reason: Stripe only guarantees meter-event identifier uniqueness for at least 24 hours and may prune request idempotency keys after 24 hours, so a request that succeeded remotely but lost its response must not be retried past that window. A duplicate-identifier rejection resolves to `{:ok, :already_reported}` rather than an error, since it means Stripe already has the event.

The Stripe event identifier includes the customer, meter name, window start, and window end, so duplicate jobs for the same window cannot double-report usage.

### Subscription items

Stripe Price identifiers live in a top-level `runners` configuration map keyed by meter event name. Subscription creation and plan changes attach every configured runner Price, while subscription plan detection deliberately ignores that non-plan configuration entry.

Plan changes preserve existing runner subscription items instead of replacing them. In Stripe's [classic billing mode](https://docs.stripe.com/billing/subscriptions/usage-based/manage-billing-setup#update-prices-mid-cycle), removing a metered Price stops the eventual invoice from reflecting usage accumulated on that item, and a newly added metered item only captures usage after it was added. Since runner Prices are plan-independent and usually unchanged by the plan change, a blanket delete-and-re-add would discard runner usage already accrued in the cycle. Both `update_plan/1` and `upgrade_to_enterprise/2` now keep every existing item whose Price is a configured runner Price, delete only the rest, and add only the runner Prices missing from the subscription.

## Impact

Every managed environment ships in the reporting-only state: both meter event names are present in `server.stripe.prices.runners` with empty Price ids. Usage flows to the Meters, where it can be inspected against real runner activity, and nothing can be charged. No Price exists to bill against, and both subscription-item builders skip empty ids, so no subscription can carry a runner item even through a plan change or a new signup.

Deployment therefore starts reporting in staging, canary, and production, and starts charging nowhere. Billing begins only when a real Price id is filled in and the item is deliberately attached.

Reported volume starts near zero and ramps. Sessions that already exist have no recorded job window and never will, so only sessions opened after the deploy report anything.

Existing runner-enabled subscriptions still need the runner Price items added once, deliberately, when pricing is ready. New subscriptions and future plan changes receive every configured runner Price automatically, and keep the ones they already have.

Adding a new machine shape to an existing platform needs no Stripe work at all: it reports into that platform's existing meter at its own multiplier. Only a genuinely new platform needs a Meter, Price, and config entry.

One prerequisite applies before merge regardless of runner billing. This change also moves the existing `remote_cache_hit` and language-model meters from ingestion-time stamping to service-period stamping, so the invoice-finalization grace period must be set to exceed the gap between period close and the daily job. Without it, boundary-day cache usage can be stamped inside a period that has already finalized and go unbilled.

## Stripe rollout

Create one `Runner compute` Product and one summed Meter per event below.

```text
runner_linux_compute_unit_milliseconds
runner_macos_compute_unit_milliseconds
```

Each Meter sums the `value` payload key and maps customers by ID from `stripe_customer_id`.

Rollout is two ordered steps per platform, each independently verifiable.

**Reporting.** With the Meter created, add its event name to each managed environment's `server.stripe.prices.runners` map with an empty Price id:

```yaml
runners:
  runner_linux_compute_unit_milliseconds: ""
  runner_macos_compute_unit_milliseconds: ""
```

Real usage then accrues on the Meter, where it can be inspected in Stripe against known runner activity. Nothing can be charged: no Price exists, and both subscription-item builders skip empty ids, so no subscription carries the item.

**Billing.** Create one monthly recurring metered Price per Meter, transform the quantity by dividing by 60,000 and rounding up, then set that platform's amount per baseline machine-minute: per Linux 2 vCPU / 8 GB minute, and per macOS 6 vCPU / 14 GB minute. Fill the Price ids into the same map and add the items to existing runner-enabled subscriptions.

Each platform moves through these steps independently.

The invoice-finalization [grace period](https://docs.stripe.com/billing/subscriptions/usage-based/configure-grace-period) has to be set to exceed the gap between period close and the daily reporting job before runner Prices go live. It is a Billing settings value rather than an API parameter, so it cannot be configured from code.

When prepaid packages are introduced, a paid prepaid invoice should create a Stripe Credit Grant scoped to the runner metered Prices. Usage reporting stays identical for prepaid and pay-as-you-go customers: the discount lives in how over-funded the grant is, not in a second Price, so there is no second credit ledger in Tuist. Scoping a grant to one platform's Price is what lets prepaid terms differ between Linux and macOS.

## Validation

Verified end to end against staging, not only in tests.

**Stripe setup.** Both Meters created in the staging, canary, and production accounts, each summing the `value` payload key and mapping customers by id from `stripe_customer_id` — matching exactly what `report_meter_event/6` posts, and the same shape as the long-standing production `remote_cache_hit` meter.

**Reporting-only state.** Staging's `runners` map carries both meter keys with empty Price ids, and the running pod was confirmed to carry that in `TUIST_STRIPE_PRICES`. No Price exists, so nothing could be charged while this was exercised.

**A real job.** A macOS smoke job ran on the staging fleet from 14:55:02 to 14:55:08 UTC — six seconds on the 6 vCPU / 14 GB machine, whose Pod existed for roughly ten minutes around it.

**The result.** Reporting that window produced `runner_macos_compute_unit_milliseconds = 6000` on the child job, and Stripe's meter event summary for the customer reads `aggregated_value=6000.0`.

Six thousand is the whole point: six seconds of job time, not the ten minutes the Pod held the host. macOS 6 vCPU / 14 GB is the macOS baseline, so compute-unit milliseconds equal elapsed job milliseconds and the figure is directly checkable against GitHub's own job timestamps.

**Two bugs this caught that tests did not.** Both are unreachable from the nightly path, which only ever reports a window that has already closed, and both surfaced immediately when reporting a window that was still open:

- the event was stamped one second before `period_end`, which is in the future for an open window, and Stripe rejects a future-dated meter event outright
- clamping that timestamp to the current time then made the parameters differ between attempts under an idempotency key derived from the customer, meter, and period, which Stripe also rejects

Both are fixed with regression tests: the timestamp is clamped to the reporting instant, and the reporting instant is the job's `inserted_at` so every retry stamps identically.

**Also run.**

- `mix format --check-formatted`
- runner, billing, billing worker, and command-event suites: 605 passed (one failure under some orderings is a pre-existing order-dependent flake in `OrphanedRunnersWorkerTest`, which passes in isolation and moves with the seed)
- `mix credo --strict`: no findings in changed files; existing repository-wide notices remain
- rendered the staging and production deployment templates and confirmed the `TUIST_STRIPE_PRICES` payload carries the `runners` map
- migration safety checker reported no warning for the new migrations

No browser screenshots are included because this changes billing persistence, reporting, configuration, and documentation without changing a web interface.


server/lib/tuist/billing.ex
```diff
@@ -224,11 +300,30 @@ defmodule Tuist.Billing do
...
+  end
+
+  defp runner_subscription_items(available_prices) do
+    available_prices
+    |> Map.get("runners", %{})
+    |> Enum.sort_by(&elem(&1, 0))
+    |> Enum.flat_map(fn
+      {_meter_event_name, price_id} when is_binary(price_id) and price_id != "" -> [%{price: price_id}]
+      _ -> []
+    end)
```
fortmarek, line 519: **[P1] Model runner consumption as normalized compute units before creating the Stripe catalog.**

Attaching every machine-specific Price to every subscription makes each shape a permanent Stripe billing item. The initial nine runner shapes already consume nine of Stripe's 20 classic subscription items, in addition to the plan's flat and other metered items. Every future shape would also require a Meter, Price, environment configuration, and a backfill across existing subscriptions.

Multiplying runtime by an immutable machine factor is metering rather than reimplementing pricing: Tuist can report `elapsed_milliseconds * billing_multiplier` to one `runner_compute_unit_milliseconds` Meter while Stripe continues to own the unit's currency amount, tiers, discounts, taxes, and credits. The multiplier should be cost-weighted (including memory and the macOS premium), not simply equal to vCPU count. Persist the multiplier or a rate-card version on the runner session so changing the catalog cannot reprice usage that has already occurred, while retaining raw shape and milliseconds internally for analytics. If Linux and macOS economics need independent control, two platform meters are still preferable to one meter per exact shape.

Stripe documents the 20-item classic subscription limit here: https://docs.stripe.com/billing/subscriptions/quantities

server/lib/tuist/billing.ex
```diff
@@ -99,27 +100,98 @@ defmodule Tuist.Billing do
...
+    |> Stripe.Request.new_request(%{"Idempotency-Key" => identifier})
+    |> Stripe.Request.put_endpoint(Stripe.OpenApi.Path.replace_path_params("/v1/billing/meter_events", [], []))
+    |> Stripe.Request.put_params(%{
+      event_name: event_name,
+      identifier: identifier,
+      payload: %{
+        value: value,
+        stripe_customer_id: customer_id
+      }
+    })
```
fortmarek, line 304: **[P1] Attribute the event to the service period in which usage occurred.**

Without `timestamp`, Stripe assigns yesterday's aggregate to ingestion time. That can move usage across invoice cycles or a mid-cycle Price change. More importantly, when `cancel_at_period_end` is reached there is no following invoice to receive the shifted final-day usage, so it can go unbilled; the assumption in the function comment that a following invoice always exists is unsafe.

Please report timestamps within the source usage interval and configure enough invoice-finalization grace for the reporting delay. Because a UTC-day aggregate can straddle an arbitrary subscription-item boundary, simply using `period_start` is not fully correct either—the aggregate should be split at billing boundaries or reported at a finer granularity.

Stripe references: https://docs.stripe.com/billing/subscriptions/usage-based/how-it-works and https://docs.stripe.com/billing/subscriptions/usage-based/configure-grace-period

server/lib/tuist/billing/workers/sync_customer_stripe_meter_worker.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Tuist.Billing.Workers.SyncCustomerStripeMeterWorker do
+  @moduledoc """
+  Reports one snapshotted customer meter value for the billing period
+  chosen by the parent worker.
+  """
+  use Oban.Worker
```
fortmarek: **[P2] Bound retries to Stripe's deduplication window.**

This uses Oban's default 20 attempts, whose exponential backoff can span more than 13 days, while Stripe only guarantees Meter Event identifier uniqueness for at least 24 hours and may prune request idempotency keys after 24 hours. A request that succeeded remotely but lost its response can therefore be retried after both protections expire and double-report usage.

Please keep automatic retries within the 24-hour window and use durable local delivery/reconciliation state for later recovery. Duplicate-event responses should also be treated as already delivered rather than retried indefinitely.

References: https://docs.stripe.com/api/billing/meter-event/create, https://docs.stripe.com/api/idempotent_requests, and https://oban.hexdocs.pm/Oban.Worker.html

fortmarek (changes requested):
The session persistence and immutable worker period are good foundations, but I think the Stripe model needs revision before runner Prices are enabled. The inline findings cover the billing unit model, service-period attribution, and retry/idempotency guarantees.

server/lib/tuist/billing.ex
pepicrft: Agreed, and changed. Runner usage is now metered as normalized compute units with one meter per platform, so a subscription carries two runner items regardless of how many shapes the catalog grows to.

One unit is the 2 vCPU / 8 GB baseline. The factor is cost-weighted rather than vCPU count: roughly two thirds of a machine's cost is CPU and one third memory, so it comes out to `(8 * vcpus + memory_gb) / 24`, exact at the baseline and linear from there. `Catalog.billing_multiplier/3`.

I kept the multiplier purely about resources and left the macOS premium in the per-platform Stripe Price, which is why I went with your two-platform-meter option rather than a single global meter. Folding a platform premium into the multiplier would have pushed pricing into what should stay metering, and it would have made Linux and macOS economics impossible to move independently.

On repricing: `runner_sessions` now has a `billing_multiplier` column written at open time, so changing the catalog cannot touch usage that already happened. Rows created before the migration fall back to the current catalog, which is the one case where a rate-card change can move an old session, and it is unavoidable since those rows never recorded what they were admitted under. Raw shape and raw milliseconds stay on the row, so analytics keep full fidelity.

`Tuist.Runners.Billing.compute_units_by_platform/4` is the new aggregation. Covered by tests for the weighting, the frozen multiplier, and the legacy fallback.

server/lib/tuist/billing.ex
pepicrft: You are right, and the `cancel_at_period_end` case is what convinces me. I dropped the timestamp on purpose earlier, reasoning that a `period_start` stamp risks landing usage in an invoice that already finalized while ingestion time always lands in the open period and only defers boundary-day usage by a cycle. That argument quietly assumes the subscription keeps renewing. When it does not, there is no following invoice and the final day is simply lost.

So the timestamp is back, and your point that `period_start` alone is not fully correct either is handled by splitting first rather than by stamping harder. `Billing.usage_windows/3` looks up the subscription's renewal instant and, when it falls inside the day being reported, snapshots each side separately. Each event is then stamped one second inside the half-open end of its own window, so it lies wholly within one service period and carries exactly the usage that period earned. A renewal is the only boundary that can land inside a one-day window and there can be at most one, so this stays one Stripe lookup and at most two sets of jobs.

On the grace period: it is a Billing settings value rather than an API parameter, so I could not configure it in code. It needs to be set to exceed the gap between period close and this daily job before runner Prices go live. I noted that requirement in the `report_meter_event/5` docs so it does not get lost.

Fallback is the whole undivided window when there is no active subscription or Stripe is unreachable. Over-reporting into the current period beats dropping a customer's day, and an account with no subscription has nothing to invoice against.

server/lib/tuist/billing/workers/sync_customer_stripe_meter_worker.ex
pepicrft: Both halves are done now.

Retries are capped at 12 attempts. With Oban's default `attempt^4 + 15` backoff the twelfth attempt fires about 11 hours after the first, so the whole retry chain finishes inside Stripe's 24 hour window even with jitter, against more than 13 days at the default of 20. Anything failing past that is dropped rather than risking a duplicate once both the identifier uniqueness and the idempotency key have expired.

Duplicate responses are now treated as delivered: `report_meter_event/5` resolves them to `{:ok, :already_reported}` instead of returning an error. Stripe does not document a stable error code for this, so the match is on the identifier-conflict message shape and everything else stays an error. Both paths have tests.

I did not add durable local delivery state. The value being reported is already snapshotted by the parent worker and the identifier is deterministic from `(customer, meter, period)`, so a retry reports the same number under the same identifier and there is nothing extra to reconcile against. If we do see drops past the 24 hour bound in practice, a reconciliation pass is the right follow-up, but I would rather add it against real evidence than build the state now.

server/lib/tuist/billing.ex
fortmarek: Thanks — this is internally coherent, but I don't think the metering/pricing boundary quite holds. The multiplier already encodes pricing policy: the two-thirds CPU / one-third memory weighting is cost-based rather than a purely physical measurement. Adding a platform coefficient would complete that same rate card rather than introduce a different concern.

With separate platform Prices, a “compute unit” has a different monetary value on Linux and macOS. The macOS Price is also defined against a hypothetical 2 vCPU / 8 GB Mac; the actual 6 vCPU / 14 GB machine consumes `31 / 12` (about `2.5833`) of those units per minute.

Unless we specifically need independent Stripe discounts, credits, contracts, or invoice line items for Linux and macOS, I think one global Meter and one Price is cleaner and closer to the conventional CI billing-unit model:

```
effective_multiplier = resource_multiplier * platform_multiplier
```

The effective multiplier is already frozen on the session, so macOS economics can still move independently for future sessions without repricing usage that already happened. Do we have a concrete requirement for those platform-specific Stripe levers? If not, I'd prefer the single-meter model. If we keep two meters, I'd instead define platform-specific baselines — Linux 2/8 = one Linux unit and macOS 6/14 = one macOS unit — so the macOS Price does not describe a machine that does not exist.

server/lib/tuist/billing.ex
fortmarek: **[P1] Boundary discovery still loses the cancellation boundary and silently misattributes outages.**

The timestamp-and-split direction is right, but the current lookup does not cover two important cases. At `cancel_at_period_end`, the subscription's `current_period_start` is the beginning of the final cycle; once the local row becomes canceled, `get_current_active_subscription/1` returns nothing at all. Neither path discovers the cancellation *end* inside the reporting day, so the daily aggregate remains whole and is stamped near the end of the UTC day — potentially after the subscription ended — and the final usage can still miss the final invoice.

A transient `Stripe.Subscription.retrieve/1` failure is also collapsed to `nil`. On a renewal day that permanently snapshots an unsplit value and attributes the pre-renewal portion to the post-renewal period instead of retrying the lookup.

Please derive boundaries from the metered subscription item that covers the reporting window, including canceled or past-due subscriptions, and distinguish “no applicable subscription” from a Stripe error. The latter should make the parent retry within the configured invoice-finalization grace period rather than treating an unknown boundary as no boundary.

server/lib/tuist/billing.ex
```diff
@@ -157,14 +312,18 @@ defmodule Tuist.Billing do
...
+      |> Enum.map(&%{price: &1})
 
     flat_prices =
       available_prices[plan]["flat_monthly"]
+      |> List.wrap()
       |> Enum.map(&%{price: &1, quantity: 1})
       |> Enum.take(1)
 
-    usage_prices ++ flat_prices
+    usage_prices ++ runner_subscription_items(available_prices) ++ flat_prices
```
fortmarek, line 437: **[P1] Plan changes can discard already-accrued runner usage.**

Including the runner Prices in `subscription_items` feeds them into both update paths above, which delete every existing subscription item and then add fresh ones. With Stripe's classic billing mode, removing a metered Price prevents the eventual invoice from reflecting usage accumulated on that item, while a newly added metered item only captures usage after it was added. A mid-cycle plan or enterprise change can therefore discard runner usage already accrued in the cycle — even though the runner Prices are plan-independent and may not have changed at all.

Please preserve the existing runner subscription item IDs across plan changes, or explicitly adopt and validate a billing mode / scheduled transition that invoices the removed item's prior usage before including runner items in the blanket replacement. Stripe documents the classic-mode behavior here: https://docs.stripe.com/billing/subscriptions/usage-based/manage-billing-setup#update-prices-mid-cycle

fortmarek (changes requested):
The normalized metering direction is much better, but two remaining paths can still lose billed usage: blanket replacement of runner subscription items during mid-cycle plan changes, and incomplete service-period boundary discovery. I added the plan-change finding inline and followed up in the existing boundary thread.

server/lib/tuist/billing.ex
fortmarek: You are right, and the plan-independence is exactly what makes it worse: a customer switching Air to Pro has not touched their runner configuration at all, yet the blanket replacement would drop the runner item and take the cycle's accrued runner usage with it.

Fixed by preserving the item IDs rather than changing billing mode. `reconcile_subscription_items/2` splits the subscription's current items into those whose Price is a configured runner Price and everything else. The runner items are neither deleted nor re-added, so they keep their Stripe item IDs and the usage already recorded against them. Only the non-runner items are deleted, and only runner Prices that are not already on the subscription get added. Both `update_plan/1` and `upgrade_to_enterprise/2` go through it.

That also keeps the "attach configured runner Prices to changed subscriptions" behaviour intact: a subscription that predates the runner rollout still picks the runner items up on its next plan change, it just does not lose the ones it already had.

I did not go the billing-mode route. Adopting flexible billing mode or a scheduled transition would change invoicing for every existing subscription to fix a case that item preservation handles directly, and it would need its own validation pass against live subscriptions.

Two tests cover it: a plan change on a subscription that already carries the Linux runner item asserts the item is absent from the update payload entirely while only the missing macOS Price is added, and the enterprise switch asserts both runner items survive.

server/lib/tuist/billing.ex
fortmarek: Both gaps are real and both are fixed.

The cancellation end was the worse of the two, and you identified the reason it slipped through: discovery started from `get_current_active_subscription/1`, so at `cancel_at_period_end` it found the final cycle but never its end, and once the row flipped to canceled it found nothing at all. Discovery now starts from the account's most recent subscription row regardless of status, which is exactly the row that covers the reporting window on the day service ends. Boundaries are collected from `current_period_start`, `cancel_at`, and `ended_at` together: `cancel_at` is what `cancel_at_period_end` schedules, and `ended_at` is what Stripe stamps once the subscription actually ends, so the day before and the day after the cancellation lands are both covered.

Splitting is now general rather than single-cut. Every boundary that falls strictly inside the window becomes a cut, and the window is sliced into consecutive half-open sub-windows. A day carrying both a renewal and a cancellation produces three windows. The pre-cancellation portion is stamped one second inside the end of the final period, so it lands on the final invoice.

On the transient failure: `usage_windows/3` returns `{:ok, windows}` or `{:error, reason}` instead of collapsing an error to `nil`. An unknown boundary is not the same as no boundary, and the parent worker's snapshot is permanent since a child never re-queries usage, so a wrong guess cannot be corrected later. The parent now fails the job on error and enqueues nothing. Its retries are capped at 12 attempts, about 11 hours with Oban's default backoff, so the retry chain finishes inside the invoice-finalization grace period the reporting delay already has to fit within.

Only the genuine no-subscription case is still treated as no boundary, and it reports the whole window as one. An account with no subscription has nothing to invoice against.

Tests cover the scheduled cancellation, the already-canceled subscription, a day carrying both a renewal and a cancellation, and the parent worker refusing to enqueue anything when the lookup fails.

server/lib/tuist/billing.ex
fortmarek: Agreed, and changed to the single-meter model.

You are right that the boundary was drawn in the wrong place. The resource weighting is already cost-based rather than a physical measurement, so calling it metering while calling the platform coefficient pricing was not a real distinction. Both are rate card, and the compute unit only means something if it is worth the same amount wherever it came from.

`Catalog.billing_multiplier/3` is now `resource_multiplier * platform_multiplier`, reporting into one `runner_compute_unit_milliseconds` meter with one Price. `resource_multiplier/2` stays exposed separately for analytics that want to compare shapes independently of the platform premium.

We have no concrete requirement for platform-specific Stripe discounts, credits, contracts, or invoice line items. The one thing I did want to keep was the ability to move macOS economics without repricing usage that already happened, and the frozen `billing_multiplier` on the session gives that on its own, which is what made the second meter unnecessary rather than merely undesirable.

macOS carries a 4x coefficient. Apple Silicon capacity costs several times equivalent Linux capacity per resource unit: the hardware is dedicated, a host serves one tenant at a time, and none of it can be oversubscribed. The real 6 vCPU / 14 GB machine then bills at `62/24 * 4`, about 10.33 baseline units per minute, which lands on the conventional macOS-to-Linux ratio in CI pricing. Flagging that number explicitly since it is the rate-card decision rather than a derivation, and it is the one thing worth confirming before the Price goes live. It only affects sessions opened after a change.

A side effect worth noting: a new machine shape, or a new platform, now needs no Stripe work at all. It reports into the existing meter at its own multiplier.

fortmarek: Coming back to this one: working through the actual rate card convinced me you were right to push on the boundary, and also that the condition you set for preferring a single meter turns out not to hold. So I have gone to two meters, with the platform-specific baselines you suggested.

Your framing was "unless we specifically need independent Stripe discounts, credits, contracts, or invoice line items". Two things showed we do.

Pricing the macOS machine fixed every Linux price. With one global unit there is one Price, so choosing a per-minute macOS rate determines the Linux rate at a hardcoded ratio, and the only lever left is the platform coefficient. Setting one number and silently getting eight others is not a rate card anyone can reason about.

Prepaid is the harder one. Prepaid access is a monetary credit grant, and a grant scopes to a Price. With a single Price, prepaid is necessarily one discount percentage across every machine type. With one Price per platform, Linux and macOS prepaid terms can move separately. That is exactly the "independent credits" case.

On the baselines: I took your suggestion rather than keeping the old 2/8-for-both definition. Linux 2/8 is one Linux unit, macOS 6/14 is one macOS unit, both exactly 1.0 at their own baseline. The platform coefficient is gone entirely, and each Price is now quoted directly per real baseline machine-minute, so no Price describes a machine that does not exist. Within a platform the resource weighting is unchanged, so a shape that doubles still doubles.

What I kept from the single-meter round: normalized units rather than a meter per shape. Adding a shape to an existing platform still needs no Stripe work, and a subscription carries two runner items, not nine. Only a genuinely new platform needs a Meter, Price, and config entry.

Nothing has shipped, so there is no repricing or backfill to worry about: the runner price map is empty in all three managed environments and no `billing_multiplier` rows exist anywhere.

`Catalog.billing_multiplier/3`, `Runners.Billing.compute_units_by_platform/4`, and `meter_event_name/1`. The config map was already keyed by meter event name, so it took two keys with no plumbing change.

server/lib/tuist/runners/dispatch.ex
```diff
@@ -314,15 +314,46 @@ defmodule Tuist.Runners.Dispatch do
...
+    case RunnerSessions.record_execution(runner_name, workflow_job_id, account_id, job_execution_window(job)) do
+      {:error, changeset} -> {:error, {:session_execution_write_failed, inspect(changeset.errors)}}
+      _outcome -> :ok
+    end
+  end
+
+  # GitHub's own bounds for the job, which is what the customer is
+  # charged for. The Pod that ran it boots before this window and tears
+  # down after it, and that overhead is ours rather than theirs. Missing
+  # or unparseable timestamps yield an empty window, which bills nothing.
```
fortmarek, line 349: this is a product decision worth knowing about. We'd rather err in the favor of the custom when possible and they would most likely compare what they see in GitHub with what we report.

fortmarek (approved):

PR #11906, opened 2026-07-17
fix(infra): unstrand tart pods on macOS machine deletion (vm-cleanup finalizer)

### LINKED ISSUE #11889: cluster-api-provider-tuist: macOS machine deletion strands tart pods in Terminating (vm-cleanup finalizer never removed)
## Summary

When a `ScalewayAppleSiliconMachine` is deleted (fleet churn, scale-down, MHC remediation), its tart pods can be left stuck in `Terminating` **forever**: the `tart-kubelet.tuist.dev/vm-cleanup` finalizer can only be removed by the tart-kubelet on that host, and the host is gone. Nothing else ever strips it.

Observed in production on 2026-07-16: the runners fleet churned and 4 pods in `tuist-runners` got wedged, all PodGC-deleted at the same instant (`DisruptionTarget: DeletionByPodGC — "PodGC: node no longer exists"`), all bound to Node objects that no longer exist. `kubectl delete` on them hangs indefinitely (the watch-stream `INTERNAL_ERROR` people see while waiting is unrelated proxy noise).

## Root cause chain

1. `tart-kubelet` adds `tart-kubelet.tuist.dev/vm-cleanup` to every pod it runs, and is the **only** actor that removes it after VM teardown (`infra/tart-kubelet/internal/podagent/reconciler.go:38-46`, removal at `:1102-1108`).
2. `reconcileDelete` in `infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go:895-1000` releases the server (Stage 1), deletes credentials (Stages 2-3), and explicitly deletes the Node object (Stage 4, `:969-979`) *because "the host is gone, so the kubelet can't deregister itself"* — but never touches the pods bound to that node.
3. PodGC then notices the missing node, marks the pods `Failed`, and issues a grace-period-0 delete — which hangs on the finalizer forever. The tart-kubelet reconciler comment at `reconciler.go:198-199` describes this exact terminal state.

The same logic that justifies Stage 4 (the kubelet can't deregister its Node) applies to the finalizer: the kubelet can't strip it either.

## Proposed fix

**Stage 4.5 in `reconcileDelete`**: after deleting the Node, list pods with `spec.nodeName == machine.Name` that carry the `vm-cleanup` finalizer and strip it (patch finalizers). There is no VM left to leak — the host has been released and reinstalled back into the Scaleway pool by Stage 1, which is precisely what the finalizer exists to protect against on a *live* host.

Belt-and-braces (optional, covers manual `kubectl delete node` and any non-CAPI node removal): a small periodic sweep in the same controller manager that strips the finalizer from pods where `deletionTimestamp` is set and the referenced Node is `NotFound`.

Notes:
- The finalizer string would be referenced from capt; either share a constant with `tart-kubelet` or duplicate it with a comment pointing at `podagent.PodFinalizer`.
- Needs RBAC for `pods` list/patch in the capt manager if not already present.
- `ScalewayElasticMetalMachine` (Linux) nodes run the standard kubelet and don't use this finalizer — no change needed there, though the generic sweep is harmless if shared.

## Alternatives considered

- **Sweep in runners-controller**: only covers `RunnerPool` pods; the builders/macos fleets run other tart pods (e.g. xcresult processor), which would still wedge.
- **Standalone controller**: works, but an extra deployable for a small, machine-lifecycle-coupled concern that capt already owns half of (it deletes the Node for the same reason).

## Remediation for currently stuck pods

Three pods are still wedged in production `tuist-runners` (a fourth was already manually cleared):

```
tuist-tuist-runner-pool-macos-26-3-runner-0034e37b    (node ...mndbc-h5dkz, gone)
tuist-tuist-runner-pool-macos-26-4-1-runner-2156b082  (node ...mndbc-k6kwt, gone)
tuist-tuist-runner-pool-macos-26-6-runner-5558f001    (node ...mndbc-hz6vv, gone)
```

Manual clear: `kubectl -n tuist-runners patch pod <pod> --type=merge -p '{"metadata":{"finalizers":null}}'`

Safe because the hosts were released back to the Scaleway pool and reinstalled — there is no VM to orphan.

esnunes:
## What changed

Adds a **Stage 4.5** to `ScalewayAppleSiliconMachineReconciler.reconcileDelete`: after deleting the departing host's Node object, the controller lists Pods bound to that node and strips tart-kubelet's `tart-kubelet.tuist.dev/vm-cleanup` finalizer from them. The strip is **best-effort with a bounded retry budget** — it never wedges the machine's deletion.

Closes #11889.

## Why

When a `ScalewayAppleSiliconMachine` is deleted (fleet churn, scale-down, MHC remediation), its tart pods could be left stuck in `Terminating` **forever**.

### Root cause

1. `tart-kubelet` stamps the `vm-cleanup` finalizer on every Pod it runs and is the **only** actor that removes it — after it tears the backing Tart VM down (`infra/tart-kubelet/internal/podagent/reconciler.go`).
2. `reconcileDelete` releases the Scaleway host (Stage 1) and deletes the Node object (Stage 4, *because the vanished kubelet can't deregister itself*) — but never touched the Pods bound to that node.
3. Once the host is gone, so is its tart-kubelet. PodGC notices the missing Node, marks the Pods `Failed`, and issues a grace-period-0 delete — which then hangs on the finalizer forever. `kubectl delete` on such a pod hangs indefinitely.

The exact same reasoning that already justifies Stage 4 (the vanished kubelet can't deregister its Node) applies to the finalizer: the vanished kubelet can't strip it either. Observed in production on 2026-07-16, where fleet churn wedged 4 pods in `tuist-runners`.

### Why this solution

Stage 1 already released **and reinstalled** the host back into the Scaleway pool, so there is no VM left for the finalizer to protect against — stripping it is safe and lets the Pods finalize. Doing it in capt (rather than a runners-only sweep or a standalone controller) is the natural home: capt already deletes the Node for the same "host is gone" reason and already owns tart-kubelet's lifecycle (bootstrap, config, rolling updates), and it covers every tart pod on the fleet (runners, xcresult processor, …), not just `RunnerPool` pods.

The scoped strip is deliberately narrow — only the `vm-cleanup` finalizer, only on the deleted node, only after Stage 1 released+reimaged the host. Alternatives (drain-before-delete, force-delete, remove-any-finalizer, self-clean in tart-kubelet) were considered and rejected: they either miss the dead/unreachable-host case that caused the bug, are exactly what PodGC already does, or would clobber other finalizers (volume/IP/mesh) that guard resources that are **not** moot.

### Best-effort, not fail-closed

The sweep does **not** block machine deletion indefinitely. On error, while the machine has been deleting for less than `podFinalizerCleanupRetryBudget` (5m), it requeues so a transient API failure — or a lagging `pods … patch` RBAC grant — gets several shots. Past that budget it emits a Warning event, logs, and **proceeds** to drop the machine finalizer.

Rationale: blocking forever would let a single un-patchable Pod wedge the machine's deletion, and with it the owning MachineDeployment's scale-down/rollout — a strictly worse failure than the original bug (one localized stranded pod). Since Stage 1 already released the host, proceeding leaks no capacity; the worst case degrades back to the original symptom, now loudly evented for follow-up. The budget is measured off the machine's `DeletionTimestamp`, so no new status field / CRD regen is required.

> The optional periodic sweep for manual `kubectl delete node` mentioned in the issue is intentionally **not** included here — this PR closes the reported machine-lifecycle path; the belt-and-braces sweep can follow separately if we decide we want it.

## Implementation notes

- **Finalizer constant** is duplicated from `podagent.PodFinalizer` (capt is a separate Go module and `podagent` is an `internal` package, so it can't be imported; importing tart-kubelet just for the string would drag its whole dependency graph into the build). To guard against silent cross-module drift, **both** copies are pinned to the literal by value tests (`TestPodVMCleanupFinalizerValue` in capt, `TestPodFinalizerValue` in tart-kubelet) with reciprocal comments, so a rename on either side fails CI.
- **Listing** goes through the uncached `APIReader` (`mgr.GetAPIReader()`) with a `spec.nodeName` field selector, which the API server serves natively — a single scoped read that returns only the doomed Pods, avoiding a cluster-wide Pods informer just for delete-time cleanup.
- **Patch, not Update**, drops the finalizer via `client.MergeFrom`. That JSON merge patch replaces the whole `metadata.finalizers` array, which is safe **only because there is no concurrent finalizer writer** (tart-kubelet, the sole other actor on this Pod's finalizers, is gone with the host). The sweep attempts every matching Pod and joins per-pod errors so one bad Pod can't hide the rest.
- **RBAC**: adds `pods … patch` to the kubebuilder marker, the capt manager `ClusterRole` in the Helm chart (`get,list,watch` → `+patch`), and the `config/rbac` intent mirror. The ClusterRole and the operator Deployment ship in the **same Helm release** with a values-sourced image, and Helm applies ClusterRoles before Deployments — so the grant lands no later than the operator rollout. With the best-effort behavior above, even a GitOps split that lags RBAC only delays the strip instead of wedging deletion.

`ScalewayElasticMetalMachine` (Linux) nodes run the standard kubelet and don't use this finalizer — no change there.

## Validation

- `go build ./...` (capt + tart-kubelet) — clean
- `go vet ./controllers/macos/ ./cmd/manager/` — clean
- `go test ./controllers/macos/` — passes, including:
  - `TestReconcileDelete_UnwedgesTerminatingPod` — reproduces the bug: a Pod force-deleted (DeletionTimestamp set) but held open by the finalizer is finalized and disappears after Stage 4.5.
  - `TestReconcileDelete_StripsStrandedPodFinalizers` — scoping: only Pods on the deleted node with the finalizer are touched; other nodes and unrelated finalizers are left intact.
  - `TestReconcileDelete_PodFinalizerStripFailureWithinBudgetRequeues` — a failing strip within the budget requeues (30s) and **keeps** the machine finalizer.
  - `TestReconcileDelete_PodFinalizerStripFailureBeyondBudgetProceeds` — a persistently failing strip past the budget **proceeds** (machine finalizer removed, no requeue, `PodFinalizerCleanupGaveUp` event).
  - `TestPodVMCleanupFinalizerValue` — drift guard.
- `go test ./internal/podagent/ -run TestPodFinalizerValue` (tart-kubelet) — drift guard.

## Remediation for currently-stuck pods

Independent of this rollout, any pods still wedged can be cleared manually (safe — the hosts were released and reinstalled, so there is no VM to orphan):

```
kubectl -n tuist-runners patch pod <pod> --type=merge -p '{"metadata":{"finalizers":null}}'
```

### DIFF
--- a/infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
+++ b/infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
@@ -67,6 +67,15 @@ const (
 	// advertises for dashboard VNC sessions through the per-Mac Tailscale
 	// egress Service.
 	DashboardVNCRelayPort = 5900
+
+	// podFinalizerCleanupRetryBudget bounds how long reconcileDelete keeps
+	// retrying the Stage 4.5 stranded-Pod finalizer strip before it gives up
+	// and lets the machine delete proceed anyway. It exists so a transient API
+	// error — or a `pods … patch` RBAC grant that lands after the operator
+	// image rolls — gets several shots, without ever wedging the machine's
+	// deletion (and the owning MachineDeployment's rollout) forever. See
+	// stripStrandedPodFinalizers for why proceeding is safe.
+	podFinalizerCleanupRetryBudget = 5 * time.Minute
 )
 
 // ScalewayAppleSiliconMachineReconciler reconciles ScalewayAppleSiliconMachine objects.
@@ -1013,15 +1022,34 @@ func (r *ScalewayAppleSiliconMachineReconciler) reconcileDelete(
 	// Terminating (see podagent.PodFinalizer). Stage 1 already released and
 	// reinstalled the host, so there is no VM left for the finalizer to
 	// protect. Do it after the Node delete so we only ever strip Pods whose
-	// host we've confirmed gone. Blocking (requeue on error) like the other
-	// stages: a stranded Pod is precisely the failure this stage exists to
-	// prevent, so we don't drop the machine finalizer until the sweep runs
-	// clean.
+	// host we've confirmed gone.
+	//
+	// Best-effort, deliberately NOT fail-closed: while we're still within
+	// podFinalizerCleanupRetryBudget of the deletion, requeue on error so a
+	// transient API failure or a lagging `pods … patch` RBAC grant (the
+	// operator image can roll before the ClusterRole update lands) gets
+	// another shot. Once that budget is spent, event + log and PROCEED to drop
+	// the machine finalizer instead of retrying forever. Blocking here
+	// indefinitely would let a single un-patchable Pod wedge the machine's
+	// deletion — and with it the owning MachineDeployment's scale-down /
+	// rollout — which is a strictly worse failure than the bug this stage
+	// fixes. Proceeding leaks nothing (Stage 1 already released the host); the
+	// worst case degrades back to the original symptom, a Pod left in
+	// Terminating, now loudly evented for an operator / the periodic reclaim.
 	if err := r.stripStrandedPodFinalizers(ctx, machine.Name); err != nil {
-		logger.Error(err, "strip stranded pod finalizers; will retry")
-		r.Recorder.Eventf(machine, corev1.EventTypeWarning, "PodFinalizerCleanupFailed",
-			"strip vm-cleanup finalizer from stranded pods on node %s: %v (will retry)", machine.Name, err)
-		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+		withinBudget := !machine.DeletionTimestamp.IsZero() &&
+			time.Since(machine.DeletionTimestamp.Time) < podFinalizerCleanupRetryBudget
+		if withinBudget {
+			logger.Error(err, "strip stranded pod finalizers; will retry")
+			r.Recorder.Eventf(machine, corev1.EventTypeWarning, "PodFinalizerCleanupFailed",
+				"strip vm-cleanup finalizer from stranded pods on node %s: %v (will retry)", machine.Name, err)
+			return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+		}
+		logger.Error(err, "strip stranded pod finalizers exhausted retry budget; proceeding with machine deletion",
+			"budget", podFinalizerCleanupRetryBudget)
+		r.Recorder.Eventf(machine, corev1.EventTypeWarning, "PodFinalizerCleanupGaveUp",
+			"strip vm-cleanup finalizer from stranded pods on node %s still failing after %s; proceeding with machine deletion, pods may remain Terminating: %v",
+			machine.Name, podFinalizerCleanupRetryBudget, err)
 	}
 
 	// Stage 5: drop the per-machine Tailscale egress Service if the
@@ -1054,8 +1082,15 @@ func (r *ScalewayAppleSiliconMachineReconciler) reconcileDelete(
 // The list goes through the uncached APIReader with a spec.nodeName field
 // selector, which the API server serves natively — a single scoped read that
 // returns only the doomed Pods, no cluster-wide Pods informer required. Each
-// match is patched (not Updated) to drop only the finalizer, so a concurrent
-// PodGC status write can't lose a resourceVersion race against us.
+// match is patched (not Updated) to drop the finalizer. The JSON merge patch
+// MergeFrom emits replaces the whole metadata.finalizers array, so this is
+// safe only because there is no concurrent finalizer writer — tart-kubelet,
+// the sole other actor that touches this Pod's finalizers, is gone with the
+// host. PodGC still races us, but only on the Pod's status/deletion, which the
+// merge patch doesn't carry, so it can't clobber or be clobbered here.
+//
+// Every match is attempted even if one fails; the per-Pod errors are joined so
+// a single un-patchable Pod can't hide progress on (or failures of) the rest.
 func (r *ScalewayAppleSiliconMachineReconciler) stripStrandedPodFinalizers(
 	ctx context.Context,
 	nodeName string,
@@ -1067,6 +1102,7 @@ func (r *ScalewayAppleSiliconMachineReconciler) stripStrandedPodFinalizers(
 		return fmt.Errorf("list pods on node %q: %w", nodeName, err)
 	}
 
+	var errs []error
 	for i := range pods.Items {
 		pod := &pods.Items[i]
 		if !controllerutil.ContainsFinalizer(pod, PodVMCleanupFinalizer) {
@@ -1078,12 +1114,13 @@ func (r *ScalewayAppleSiliconMachineReconciler) stripStrandedPodFinalizers(
 			if apierrors.IsNotFound(err) {
 				continue
 			}
-			return fmt.Errorf("strip vm-cleanup finalizer from pod %s/%s: %w", pod.Namespace, pod.Name, err)
+			errs = append(errs, fmt.Errorf("strip vm-cleanup finalizer from pod %s/%s: %w", pod.Namespace, pod.Name, err))
+			continue
 		}
 		logger.Info("stripped stranded vm-cleanup finalizer",
 			"pod", pod.Namespace+"/"+pod.Name, "node", nodeName)
 	}
-	return nil
+	return errors.Join(errs...)
 }
 
 // reader returns the uncached APIReader when wired, else the cached client.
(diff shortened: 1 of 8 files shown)


infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
```diff
@@ -999,6 +1073,67 @@ func (r *ScalewayAppleSiliconMachineReconciler) reconcileDelete(
...
+		return fmt.Errorf("list pods on node %q: %w", nodeName, err)
+	}
+
+	var errs []error
+	for i := range pods.Items {
+		pod := &pods.Items[i]
+		if !controllerutil.ContainsFinalizer(pod, PodVMCleanupFinalizer) {
+			continue
+		}
+		patch := client.MergeFrom(pod.DeepCopy())
```
pepicrft, line 1111: Could we make this patch concurrency-safe? `client.MergeFrom` replaces the complete `metadata.finalizers` array and does not include a resource-version precondition. Finalizers are shared, so another controller can change the list between the uncached read and this patch. We can then drop a newly added finalizer or attempt to restore one that was just removed. `MergeFromWithOptions(..., client.MergeFromWithOptimisticLock{})` plus the existing retry path would prevent the lost update. It would also be useful for the preservation test to put both finalizers on the same Pod and simulate a concurrent change.

infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
```diff
@@ -978,6 +1013,45 @@ func (r *ScalewayAppleSiliconMachineReconciler) reconcileDelete(
...
+	// operator image can roll before the ClusterRole update lands) gets
+	// another shot. Once that budget is spent, event + log and PROCEED to drop
+	// the machine finalizer instead of retrying forever. Blocking here
+	// indefinitely would let a single un-patchable Pod wedge the machine's
+	// deletion — and with it the owning MachineDeployment's scale-down /
+	// rollout — which is a strictly worse failure than the bug this stage
+	// fixes. Proceeding leaks nothing (Stage 1 already released the host); the
+	// worst case degrades back to the original symptom, a Pod left in
+	// Terminating, now loudly evented for an operator / the periodic reclaim.
+	if err := r.stripStrandedPodFinalizers(ctx, machine.Name); err != nil {
```
pepicrft, line 1039: Could we gate this cleanup on knowing that a reinstall was actually accepted? `ReleaseToPool` currently treats every `scw.TransientStateError` from `ReinstallServer` as success, even when `CurrentState` is not `reinstalling`. If it returns `updating`, this path deletes the identity and Node and strips the Pod finalizer even though the host was not wiped. The server has already been renamed into the adoption pool, so it can return Ready with the previous Tart virtual machines still present. I think `ReleaseToPool` should only swallow `CurrentState == "reinstalling"` (or return a disposition that this stage verifies), with a regression test for a non-reinstall transient state.

pepicrft (approved):
Approving the overall direction. I left two inline findings that should be addressed before merge: make Pod cleanup contingent on an accepted host reinstall, and make finalizer removal safe under concurrent updates.

fortmarek (approved):

esnunes:
Closing: the symptom this PR was written for is handled on `main` by a different (reactive) fix, and the branch has drifted out of scope.

**What handles it now.** #12828 (`8206b9787e`, 2026-09-03) added a sweep to the runners-controller: once a runner Pod has been deleting for longer than its grace period plus a margin *and* its `spec.nodeName` resolves to a `NotFound` Node, `forceReapRunner` strips the Pod's finalizers instead of retrying a force-delete forever (`infra/runners-controller/controllers/runnerpool_controller.go`, `terminationTimedOut` + `nodeMissing` in `provisioning.go`). That loop runs over every Pod in a pool regardless of OS, so it covers exactly the `tuist-runners` Pods that wedged in #11889 — the reported failure now self-heals a few minutes after PodGC gives up, with no operator action.

It is not the design this PR argued for — it cleans up after the fact rather than preventing the strand at the source, and it deliberately leaves a finalizer alone while the owning node is still present — but it removes the operational problem, and no wedged tart Pods have shown up since.

**Why not rebase this instead.** Two things have changed under the branch:

- It conflicts with `main` after ~2 months of capt churn.
- It is now under-scoped. #13022 added `RackAppleSiliconMachine`, whose `reconcileDelete` deletes the Node object for the same "the host's kubelet cannot deregister itself" reason (`rackapplesiliconmachine_controller.go`). A Stage 4.5 that only exists on the Scaleway controller would leave the same hole open on the rack fleet, so the source-side fix would now want a shared helper called from both.

**What is genuinely left.** `reconcileDelete` still never touches the Pods bound to the Node it deletes, so tart Pods that are *not* owned by a `RunnerPool` — today that is the xcresult processor on the macOS fleet — can still be stranded by a machine deletion, as can any Pod after a manual `kubectl delete node`. The impact is small: a Deployment excludes deleting Pods from its active count, so a replacement is scheduled immediately and the stranded Pod is inert leftover state on a Node that no longer exists, clearable with

```
kubectl -n <ns> patch pod <pod> --type=merge -p '{"metadata":{"finalizers":null}}'
```

That residual case is not worth carrying a conflicted branch for. If it ever bites, the fix should be written fresh against the current two-controller layout.

PR #11911, opened 2026-07-17
fix(search): stop the GitHub issues indexer from silently emptying the index

pepicrft:
## What

Reworks `search/bin/index-github` so a bad or empty run can no longer wipe the live `github-issues` search index. It now builds into a fresh collection, verifies the import, and only then atomically points the `github-issues` alias at it.

## Why

While testing the source-backed answers work (#11829), I noticed the `issues` source of `search_tuist` returns nothing for every query, including broad single words like `cache`, while docs, forum, and releases all return results. The MCP query path is fine (`Tuist.MCP.Search` queries the right collection with valid fields). The `github-issues` collection is simply empty.

Tracing it back, the collection is rebuilt daily by the `typesense-indexers` CronJob, which runs `index-github`. The old script did a destructive rebuild:

1. `DELETE` the live `github-issues` collection.
2. Recreate it.
3. Bulk-import in batches.

The import hits TypeSense's `documents/import`, which returns HTTP 200 even when every per-document write fails, so `curl -sf` never caught it. The script counted successes, printed `Indexed 0/N`, and exited 0. An empty rebuild was a silent success, and because it had already deleted the previous collection, search went dark with nothing to flag it. Releases and DocC still worked, which also tells us `index-github` exited 0 rather than hard-failing (otherwise `set -e` in `run-indexers.sh` would have aborted before the releases step).

I could not determine the exact trigger for the last empty rebuild from here, since that needs the CronJob run logs, but the structural problem is clear: the pipeline can empty the collection and call it success.

## What changed

- Build into a timestamped collection (`github-issues-<epoch>`) and only swap the `github-issues` alias to it once the import is verified. TypeSense alias upserts are atomic, so searches never see a missing or half-built collection, and a failed build leaves the previous data serving traffic.
- Fail the run instead of swapping when the fetch returns nothing, or when fewer than `MIN_IMPORT_RATIO` (default 0.9) of the prepared documents actually import. This is the check that was missing.
- Handle non-2xx GitHub responses explicitly and fix the rate-limit path. The old fetch used `curl -sf`, which turned a 403 into `http_code="000"` and skipped the backoff branch entirely; it now reads the real status and backs off on 403/429.
- Log the first error from any batch with per-document failures, so a partial import is visible rather than a bare count.
- One-time migration from the existing real `github-issues` collection to an alias, plus cleanup of stale `github-issues-*` builds after each swap.

No change is needed on the MCP or website side: both query the collection by the name `github-issues`, which now resolves through the alias.

## Impact

This does not repopulate the currently-empty live index by itself. It stops the silent-empty failure going forward, and the next successful CronJob run (or a manual trigger) will rebuild and swap in a populated collection. Until then, `search_tuist`'s `issues` source stays empty.

This is the observability gap made concrete: the tooling advertises issues search and nothing noticed it had gone dark. A follow-up worth doing is a doc-count check or alert on each collection after indexing, and `index-github-releases` looks like it shares the same destructive-rebuild shape and deserves the same treatment.

## Verification

`bash -n` passes and the executable bit is preserved. The existing `search` test only builds the image and health-checks TypeSense, so it is unaffected. The alias swap, the import-ratio guard, and the migration path still need a run against a live TypeSense to exercise end to end, which I could not do from my environment. I kept the collection schema and the document shape identical to before so the swap is a drop-in for the existing `github-issues` consumers.

### DIFF
--- a/search/bin/index-github
+++ b/search/bin/index-github
@@ -1,7 +1,14 @@
 #!/bin/bash
 # Indexes all GitHub issues and pull requests from tuist/tuist into TypeSense.
 # Paginates through the GitHub API, handles rate limiting, and bulk-imports
-# documents into the "github-issues" TypeSense collection in batches of 250.
+# documents into a freshly built collection, then atomically points the
+# "github-issues" alias at it.
+#
+# The live index is never emptied by a bad run. A run that fetches nothing,
+# hits an unexpected GitHub status, or imports too few documents aborts before
+# the swap and leaves the previous collection serving traffic. Only a verified
+# build replaces it.
+#
 # Requires TYPESENSE_API_KEY and TYPESENSE_HOST environment variables.
 set -euo pipefail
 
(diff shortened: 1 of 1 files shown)


fortmarek (approved):

pepicrft:
Thanks, both good catches. Addressed in the latest commit:

**P1, the legacy cutover.** The migration no longer risks permanently dropping the index. The freshly built collection is verified before anything else is touched and stays on disk as the recoverable copy. The legacy delete happens immediately before the swap, the alias `PUT` is retried with backoff to absorb a transient failure, and if it still fails we exit non-zero without cleanup so the next run just re-points the alias to the already-built collection instead of rebuilding.

**P2, secondary rate limits.** We now read `Retry-After` when GitHub omits `X-RateLimit-Reset`, and fall back to a conservative 60s (with a numeric guard) instead of the old 1s, so a secondary limit no longer gets hammered.

PR #11918, opened 2026-07-17
fix(infra): grant Grafana role after provisioning (hotfix)

pepicrft:
## What changed

The Grafana database grants now run in a dedicated Helm post-install and post-upgrade job. Schema migrations remain a pre-upgrade gate, while the new job reuses the release grant logic after CloudNativePG has received the role declaration.

The job is limited to managed CloudNativePG deployments, uses the migration database owner, and retries while the operator reconciles the new role. Self-hosted deployments are unchanged.

## Why

The production rollout for the cache authorization hotfix was blocked by an unrelated database role ordering problem. A single deployment needs to be able to declare the role and grant its restricted column access without a manual bootstrap or a second deployment.

## Root cause

The Product Usage dashboard change added `tuist_grafana_ro` to the regular CloudNativePG manifest and immediately referenced it from the pre-upgrade migration hook. Helm runs pre-upgrade hooks before applying regular resources, so the migration job attempted the grant before CloudNativePG knew about the role. PostgreSQL returned `42704 undefined_object`, and the atomic upgrade rolled back.

## Approach

The role dependency is now expressed through Helm lifecycle ordering:

- The pre-upgrade job migrates the schema and grants roles that already exist.
- Helm applies the CloudNativePG role declaration and its credential resource.
- The post-upgrade job applies the exact Grafana column allowlist.

The grant remains a required hook, so reconciliation or permission failures still fail the release instead of silently leaving broader or incomplete access.

## Impact

The hotfix deployment can now create and grant the Grafana role in one rollout. Existing environments reassert the same least-privilege grants after every deployment.

## Validation

- `mise exec -- mix format lib/tuist/release.ex lib/tuist/environment.ex`
- `mise exec -- mix test test/tuist/release_test.exs` completed with 8 passing tests.
- The complete production Helm chart rendered successfully with 7,525 lines.
- Focused renders confirmed that the pre-upgrade migration no longer receives `TUIST_DATABASE_GRAFANA_ROLE` and the new job is a post-install and post-upgrade hook.
- `git diff --check`

`helm lint` still encounters the existing `dedibox-fleet.yaml` document-separator error; the same production values render successfully with `helm template`.


pepicrft:
Closing this duplicate because Marek has already opened the fix.

PR #11920, opened 2026-07-17
fix(server): stop server ServiceAccount UID churn causing intermittent K8s 401s

fortmarek:
## What changed

The server's Kubernetes ServiceAccount (`tuist-server`) is no longer rendered as a Helm hook. It is now a stable, normal release resource, so its UID stays constant across deploys. Supporting changes remove a now-obsolete force-rollout workaround, harden the Kura reconciler's self-heal, and automate the one-time SA adoption the switch requires.

- **`serviceaccount.yaml`** — the server SA drops the `helm.sh/hook: pre-install,pre-upgrade` + `hook-delete-policy: before-hook-creation` annotations (previously added when `server.migrationJob.asHook=true`). As a normal resource with `resource-policy: keep`, Helm patches it in place instead of delete+recreate.
- **`server-deployment.yaml` / `processor-deployment.yaml`** — removed the `tuist.dev/server-service-account-revision: {{ .Release.Revision }}` pod annotation.
- **`server/lib/tuist/kura/reconciler.ex`** — `project_server` now re-applies a `not_found` KuraInstance for a present-intent server regardless of the latest deployment's status (was gated on `:succeeded`).
- **`.github/workflows/server-deployment.yml`** — idempotent, self-skipping pre-`helm upgrade` step that adopts the existing SA into the release (see below).
- **`server/test/tuist/kura/reconciler_test.exs`** — rewrote the test that encoded the old stranding behavior to assert self-heal.

## Why

Production saw intermittent HTTP 401s from the apiserver on **every** server→Kubernetes call: the Kura reconciler's GET/apply of `KuraInstance` CRs (which failed Kura deployments into `status=failed`) and, far more loudly, runner-dispatch `POST /tokenreviews` (thousands/min across all replicas). Both paths authenticate with the same in-cluster server SA token.

### Root cause

`before-hook-creation` **deletes and recreates the SA on every `helm upgrade`, minting a new UID each deploy**. Running server and processor pods (both use the server SA) still present projected (bound) tokens whose `uid` claim is the *previous* SA UID. Kubernetes validates a bound token's `uid` claim against the current SA object, so the apiserver rejects them until kubelet re-mints the token on its refresh cycle (`expirationSeconds: 3607`, ~48 min). Reading the token fresh from disk each request doesn't help — the on-disk token itself carries the stale UID.

The apiserver's own logs name it exactly:

```
authentication.go:75 "Unable to authenticate the request"
  err="[invalid bearer token, service account UID (1c7ae…) does not match claim (26d8a…)]"
```

This explains every symptom: intermittent, all accounts/regions/instances, all replicas, self-heals via kubelet refresh, recurs each deploy. Verified live — the SA was recreated mid-investigation (`16:13:47Z`, new UID) and pods started before that recreation presented the stale UID while a pod started after it did not. The bare 401 body is just `"Unauthorized"`; the reason exists only in the apiserver logs.

### Why this fix over the alternatives

- The migration Job **already** has a dedicated throwaway hook SA (`server-migrate`) created specifically to keep this churn away from long-running pods — its comment says the churn "must never happen to an SA that long-running pods hold projected tokens for." The decoupling was half-done; un-hooking the server SA completes it. First-install ordering is unaffected.
- The `server-service-account-revision` annotation was a workaround that force-rolled the whole server every deploy so pods would re-mint tokens after the SA churned. With a stable UID it is dead weight and pointless rollout churn.
- A **client-level retry-on-401 was deliberately not added**: an immediate retry reads the same stale on-disk token, so it cannot help this failure mode.

### Reconciler hardening

A failed deployment never flips back to `:succeeded` on its own, so a transient 401 that marked a deployment `:failed` could **permanently** disable CR self-heal: if the CR was later deleted out-of-band it was never recreated and the instance stranded (this is what happened to `kura-tuist-scw-fr-par`, account_id=3). Re-applying a missing CR for a present-intent server is idempotent drift correction; a genuinely broken rollout surfaces its own error each tick instead of the instance silently disappearing.

## SA adoption (automated, one-time)

The live SA was created by a Helm hook and lacks the `meta.helm.sh/release-*` ownership annotations, so Helm's import check would reject adopting it as a normal resource on the first upgrade after this switch. This is handled by a pre-`helm upgrade` step in `server-deployment.yml` (the `workflow_call` target the production cascade reuses for **every** environment — important because the cascade deploys canary first, whose SA is equally unannotated, so a production-only manual step would not gate it). The step stamps the ownership annotations + `managed-by=Helm` label on the live `tuist-server` SA and strips the stale hook annotations, preserving the UID.

It is a one-time migration guard: it **self-skips with no API write** once a cluster has been adopted (Helm re-stamps the annotations itself thereafter), tolerates first installs where the SA does not exist, and is safe to delete once every managed cluster has adopted (tracked as a follow-up). Purely-manual adoption was considered but is ordering-sensitive: the current chart still hook-deletes the SA, so any old-chart deploy after a manual annotate would wipe it; running the step inside the fix deploy applies it at exactly the right instant.

## User / operator impact

Kura deployments stop failing on transient apiserver auth; runner dispatch stops its 401 flood; deploys no longer force a full server rollout every time. Stranded Kura instances self-heal on the next reconciler tick once this ships.

## How to test locally

- `cd server && mix test test/tuist/kura/reconciler_test.exs` — 24 pass, including the rewritten vanished-CR self-heal test.
- `cd infra/helm/tuist && helm template tuist . --set server.enabled=true --set server.serviceAccount.create=true --set server.migrationJob.asHook=true --show-only templates/serviceaccount.yaml` — SA renders with `resource-policy: keep` and **no** `helm.sh/hook` annotations; `server-migrate` hook SA is preserved. Repeat with `asHook=false` for parity.
- `actionlint .github/workflows/server-deployment.yml` — no new findings.

### DIFF
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -819,6 +819,37 @@ jobs:
           # so dependent pods (Pending or otherwise) go with the Job.
           echo "$stale_jobs" | xargs -r kubectl -n "$NAMESPACE" delete job \
             --ignore-not-found --cascade=foreground --timeout=2m
+      - name: Adopt server ServiceAccount into the Helm release
+        # The server ServiceAccount used to render as a Helm *hook*
+        # (before-hook-creation), which churned its UID every deploy and
+        # broke bound-token auth. It is now a normal release resource, but a
+        # hook-created object carries no `meta.helm.sh/release-*` ownership
+        # annotations, so the first `helm upgrade` after the switch would
+        # fail its import check with "invalid ownership metadata". Stamp the
+        # ownership annotations + managed-by label (and strip the now-stale
+        # hook annotations) so helm adopts the live SA in place, preserving
+        # its UID. Idempotent: once the SA is a tracked resource this is a
+        # no-op, and it tolerates first installs where the SA does not exist
+        # yet. Applies to every environment because the production cascade
+        # deploys canary first — a production-only manual step would not gate
+        # it. SA name is `server.serviceAccount.name` in
+        # values-managed-common.yaml.
+        run: |
+          set -euo pipefail
+          SA_NAME=tuist-server
+
+          if ! kubectl -n "$NAMESPACE" get serviceaccount "$SA_NAME" >/dev/null 2>&1; then
+            echo "ServiceAccount $SA_NAME not present yet (first install); nothing to adopt."
+            exit 0
+          fi
+
+          kubectl -n "$NAMESPACE" annotate serviceaccount "$SA_NAME" \
+            "meta.helm.sh/release-name=$HELM_RELEASE_NAME" \
+            "meta.helm.sh/release-namespace=$NAMESPACE" --overwrite
+          kubectl -n "$NAMESPACE" label serviceaccount "$SA_NAME" \
+            app.kubernetes.io/managed-by=Helm --overwrite
+          kubectl -n "$NAMESPACE" annotate serviceaccount "$SA_NAME" \
+            helm.sh/hook- helm.sh/hook-weight- helm.sh/hook-delete-policy-
       - name: helm upgrade --install
         # App secrets never pass through --set: the chart runs in
         # `server.managedSecrets: true` mode, which reads DATABASE_URL /
--- a/.github/workflows/server-deployment.yml
+++ b/.github/workflows/server-deployment.yml
@@ -820,20 +820,23 @@ jobs:
           echo "$stale_jobs" | xargs -r kubectl -n "$NAMESPACE" delete job \
             --ignore-not-found --cascade=foreground --timeout=2m
       - name: Adopt server ServiceAccount into the Helm release
-        # The server ServiceAccount used to render as a Helm *hook*
-        # (before-hook-creation), which churned its UID every deploy and
-        # broke bound-token auth. It is now a normal release resource, but a
-        # hook-created object carries no `meta.helm.sh/release-*` ownership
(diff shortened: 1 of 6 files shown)


pepicrft (approved):
Validated this against Helm 4.2.0 in a disposable Kubernetes cluster.

The overall approach is correct. Keeping the long-lived ServiceAccount as a normal Helm resource fixes the token invalidation at its source, the dedicated migration ServiceAccount preserves first-install ordering, and the Kura reconciliation change restores the expected desired-state behavior.

One non-blocking robustness improvement is worth following up on in `.github/workflows/server-deployment.yml`. The adoption guard checks only `meta.helm.sh/release-name` before skipping, while Helm also requires `meta.helm.sh/release-namespace` and `app.kubernetes.io/managed-by=Helm`. I reproduced a partial failure where the first annotation write succeeded, the label was absent, the next run skipped adoption, and Helm still rejected the ServiceAccount.

Helm's built-in `--take-ownership` is the cleaner one-time migration here. In the reproduction it preserved the ServiceAccount identifier, removed the stale hook annotations, and remained adopted when the rest of the upgrade was forced to roll back. I would replace the three metadata writes with a guarded conditional ownership flag, fail safely for an unexpected existing owner, and document the one-time flag for self-hosted installations that previously enabled `server.migrationJob.asHook=true`.

This does not invalidate the root fix, and the implementation addresses the reported incident. Approving.

PR #11933, opened 2026-07-19
fix(cli): false-positive diagnostics in tuist edit projects

Kyle-Ye:
## What changed

- Keep manifest source files visible in the editable project while excluding project, workspace, config, and package manifests from compilation where appropriate.
- Disable module definition for those non-buildable manifest targets.
- Add a cached `SwiftVersionProvider.swiftDefaultLanguageModeVersion()` probe that asks the selected toolchain which Swift language mode it uses when no mode is specified.
- Use that language mode for the edit project's `SWIFT_VERSION` setting.
- Add focused mapper and provider coverage.

The existing generated-project `defaultSwiftVersion` behavior remains unchanged; this adjustment is scoped to projects created by `tuist edit`.

## Why

The edit project modeled manifests as static-framework sources. Xcode consequently tried to compile script-style manifest files as library code, producing false diagnostics such as “expressions are not allowed at the top level.”

The edit project also derived `SWIFT_VERSION` from the compiler's product version. A compiler version and its default Swift language mode are different concepts, so this could make Xcode diagnose manifests with concurrency semantics that differ from the unversioned `swift` and `swiftc` invocations Tuist uses to evaluate manifests and compile helpers.

Excluding non-buildable manifest sources and matching the selected toolchain's default language mode keeps Xcode's editor diagnostics aligned with Tuist's actual manifest-loading behavior.

## Impact

Developers opening a project generated by `tuist edit` should no longer see the false top-level-expression and Swift concurrency diagnostics covered by this change. Manifest files remain navigable and editable in Xcode.

## Validation

Validated with Xcode 26.6 and a locally built Tuist binary:

- `TuistSupportTests/SwiftVersionProviderTests`
- `TuistKitTests/ProjectEditorMapperTests`
- Tuist CLI build
- `tuist edit --permanent` plus Manifests workspace builds for DemoKit and OpenSwiftUI/Example

### DIFF
--- a/cli/Sources/TuistKit/ProjectEditor/ProjectEditorMapper.swift
+++ b/cli/Sources/TuistKit/ProjectEditor/ProjectEditorMapper.swift
@@ -60,9 +60,9 @@ struct ProjectEditorMapper: ProjectEditorMapping {
         projectDescriptionSearchPath: AbsolutePath
     ) async throws -> Graph {
         Logger.current.notice("Building the editable project graph")
-        // Tuist evaluates manifests and compiles helpers with `xcrun swift`/`swiftc`, whose default language mode
-        // is Swift 5.
-        let swiftVersion = Version(5, 0, 0).description
+        // The Swift compiler version and its default language mode can differ. Match edit targets to the language mode
+        // used by the unversioned `swift` and `swiftc` invocations that evaluate manifests and compile helpers.
+        let swiftVersion = try await SwiftVersionProvider.current.swiftDefaultLanguageModeVersion()
 
         let pluginsProject = try await mapPluginsProject(
             pluginManifests: editablePluginManifests,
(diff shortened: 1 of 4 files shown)


cli/Sources/TuistKit/ProjectEditor/ProjectEditorMapper.swift
```diff
@@ -60,7 +60,9 @@ struct ProjectEditorMapper: ProjectEditorMapping {
         projectDescriptionSearchPath: AbsolutePath
     ) async throws -> Graph {
         Logger.current.notice("Building the editable project graph")
-        let swiftVersion = try await SwiftVersionProvider.current.swiftVersion()
+        // The Swift compiler version and its default language mode can differ. Match edit targets to the language mode
+        // used by the unversioned `swift` and `swiftc` invocations that evaluate manifests and compile helpers.
+        let swiftVersion = try await SwiftVersionProvider.current.swiftDefaultLanguageModeVersion()
```
Kyle-Ye, line 65: Or we can just hard code "5.0" here.

One of the issue is because "try await SwiftVersionProvider.current.swiftVersion()" return a compiler version, but we will later use swiftVersion as the xcodeproj's `SWIFT_VERSION` which is actually language mode version instead of compiler version.

pepicrft (approved):

PR #11934, opened 2026-07-19
fix(infra): make the runner cache-volume master converge across hosts

fortmarek:
## What this fixes

The per-account macOS runner cache volume (spec #76) is enabled in production (`gib: 80`) and mechanically sound — the disk image mounts, promotes, and clones back correctly. But it delivers **zero measurable benefit**: across 20h of traffic the one account our macOS fleet serves (tuist/tuist) recorded **0 CLI local cache hits** on every command.

A field inspection of all 9 production hosts (reading `$TUIST_XDG_CACHE_HOME/tuist/Binaries` from inside jobs) found the cause: the account's full 266-entry master exists on **exactly one host**; four other hosts mounted an **empty** master, and three booted with **no share at all**.

The master isn't spreading. This PR fixes the two bugs that stop cross-host **convergence** from completing, and documents the separate boot-time no-share bug (#3) as a follow-up.

## Root cause — convergence never completes

Convergence (a behind host fast-forwarding its master to the account HEAD) works like this: dispatch hands the host the HEAD `{generation, digest, download_url}`; the host downloads the master object, recomputes its inventory digest, and **only adopts it if the recomputed digest equals the HEAD digest** ([volume_reconcile.go:266](infra/tart-kubelet/internal/podagent/volume_reconcile.go)). On mismatch it keeps its own (empty) local master. Two independent bugs made that check fail every time:

### 1. The master object was mutable and clobbered (commit: content-addressed keys)

The master was uploaded to a **single fixed key** `runner-volume-masters/<account>/tuist-cache.zip`. Concurrent jobs for the same account all PUT to that one key, so the object gets overwritten out from under the digest a prior promote published as HEAD. A converging host then downloads bytes whose inventory ≠ the HEAD digest → mismatch → keeps its empty master. The existing code even anticipated this ("*interleaved jobs for the same account overwriting the object… immutable, content-addressed keys are the fuller fix and a tracked follow-up*").

**Fix:** make the object content-addressed and immutable — `runner-volume-masters/<account>/tuist-cache/<digest>.image`. A concurrent promote of a different digest writes a different object and can never clobber the one the HEAD points at; the download for a HEAD always returns exactly the bytes that produced its digest.

Because the upload key now depends on the new digest (known only at job end, not at dispatch), the guest mints its upload URL at promote time via a new SA-authenticated endpoint `POST /runners/volume-head/upload-url {tree_digest}`, PUTs the image, **then** bumps the HEAD — so the object always exists before any converging host can read the new HEAD. The digest is validated server-side as 40-char SHA-1 hex (format match + traversal-safe key component), and the minted URL gets the same public-host SSRF check as the download URL.

### 2. Host and guest computed the inventory digest differently (commit: digest parity)

Even with immutable keys, the digest-verify would still fail because the two sides disagreed on how to compute the digest:

- The **host** (`inventoryDigest`, Go) used `os.ReadDir`, which returns dotfiles (`.DS_Store`, in-flight `.tmp*`); the **guest** (`cache_inventory`, bash) uses `ls -1` (no `-a`), which omits them. Any dotfile in a cache subtree diverged the digests. → Host now skips dotfiles.
- The **guest** sorted with the ambient locale; the **host** sorts byte-wise (`sort.Strings`). Mixed-case / punctuated hash names could order differently. → Guest now uses `LC_ALL=C sort`.

Both fixes are required together: immutable keys make the *right object* fetchable, digest parity makes the verify *pass*.

## Changed surfaces

| Layer | File | Change |
|---|---|---|
| Server | `runners.ex` | content-addressed `volume_master_object_key/2`; download URL derived from HEAD digest; new `volume_master_upload_url/2` (hex-validated + SSRF-guarded) |
| Server | `runners_controller.ex`, `router.ex` | new `POST /runners/volume-head/upload-url` action + route |
| Guest | `dispatch-poll.sh` | `report_volume_head` mints upload URL → PUT → bump; `cache_inventory` pins `LC_ALL=C sort` |
| Host | `volume.go` | `inventoryDigest` skips dotfiles |
| Docs | `data-export.md` | new key format + image type |

## Validation

- Server: `mix test` on `runners_test.exs` + `runners_controller_test.exs` — **41 passed**, incl. new tests for `volume_master_upload_url/2` (valid digest → content-addressed key; non-hex/empty/short/uppercase rejected before any storage call; non-public host rejected; missing account) and the new endpoint (200 / 422 invalid digest / 401 no bearer). `mix credo` clean, compiles `--warnings-as-errors`.
- Host: `go build`, `go vet`, and an extended `TestInventoryDigestMatchesGuestScript` proving a `.DS_Store` no longer moves the digest.
- Guest: `bash -n`.

**Not yet validated end-to-end on staging/prod.** This changes the promote/converge protocol (dispatch payload drops `upload_url`; guest makes an extra call), so guest and server must ship together, and it wants a real cross-host convergence check on staging before landing. Opening as a draft for that reason.

## #3 (investigated, separate follow-up): 3/9 hosts boot with no share

This is a **distinct** bug from convergence and is **not** fixed here. The cache share is attached at **VM boot** from `AllocateBranch`, independent of dispatch/Materialize ([reconciler.go:492-498](infra/tart-kubelet/internal/podagent/reconciler.go)); Materialize failures still leave an (empty) share, so a *totally missing* share is always a boot-time attach failure. Ranked causes:

1. **Feature not enabled on those hosts** — launchd plist missing `--runner-cache-root` because the rolling drift reconcile hasn't re-rendered/reloaded it, or the host re-bootstrapped from pre-enablement config. Tell: no `per-account cache volumes enabled` startup log and no `tart_kubelet_cache_volume_*` metrics.
2. **Backing 80 GiB volume unmounted/unprovisioned** — the host rebooted and the APFS volume didn't auto-remount (there is **no startup mount check or retry** in tart-kubelet — `installRunnerCacheVolume` is idempotent by a *mounted* check and assumes auto-mount), or provisioning errored on a host that joined after enablement. Then every `AllocateBranch` fails on `df` → `CacheVolumeSkipped` event, no share.
3. Runner-cache root not owned by the runner user (branch `mkdir`/`chmod` EPERM).
4. `errNoRoom` admission decline under disk pressure — **silent** (no log/event).

Proposed follow-up fix: a startup existence/mount check with bounded retry (fail loud if `--runner-cache-root` is set but not mounted), and a log line on the `errNoRoom` path so this stops being invisible.

## Also worth a look (out of scope)

The resume hook flagged `ueberauth_apple 0.6.1` as **CRITICAL CVE-2026-55954** (account takeover via missing ID-token claim validation). Unrelated to this PR but should be triaged.


fortmarek:
Addressed all three review findings (4 commits):

**[P1] Validate reported HEAD digests** — `1762ba3`. You're right, I'd only validated the digest on the new upload-url endpoint, not on the bump path that the download key is actually derived from. `report_volume_head` now applies the same `^[a-f0-9]{40}$` validation before `bump_head`; an invalid digest is rejected (`:error` → 422) and never persisted. Tests cover `../../etc/passwd`, `a/b`, empty, nil, and uppercase.

**[P1] Retention for superseded images** — `6e9fa4e` + `f9ac729`. On a HEAD bump that supersedes an old digest, `report_volume_head` schedules `PruneVolumeMasterWorker` with a delay equal to the presigned-URL TTL (6h) — so the object survives exactly as long as any dispatch that already handed out its download URL could still be converging to it, then is deleted. The worker re-checks the HEAD and skips the delete if the digest is (again) the live master (content-addressed keys mean a re-promoted identical set reuses the same object, so a naive delete could drop the current master). This bounds steady-state storage to the current master plus whatever was superseded within the grace window — matching the old single-object footprint in practice. I went with scheduled-delete-on-supersession rather than a periodic list-and-sweep to avoid provider-specific object-listing code and the mid-download races that come with timestamp-based pruning. Enqueue is best-effort (logs, never fails the report); a persistent delete failure falls back to the account-deletion prefix sweep.

**[P2] Preserve existing HEADs during migration** — `b5fcfdc`. Migration `20260720120000` clears the `runner_volume_heads` rows, since their pre-change digests reference the old `tuist-cache.zip` key that the new scheme never writes (dispatch would hand out a phantom `<digest>.image` URL → 404). After deploy, dispatch returns no download URL until the first cache-changing job re-establishes a valid content-addressed HEAD. Only convergence pointer state is reset (rebuilt on the next promote); on-disk local masters are untouched. I chose reset over copy/migrate the old objects because the one real fleet account's `.zip` is exactly the clobbered object the digest wouldn't match anyway, so a copy would be rejected by the digest-verify regardless.

Validation: `mix test` runners + controller = **46 passed** (new tests for the bump-path validation, the scheduled prune, and the re-promote guard); `mix credo` clean; migration passes `excellent_migrations.check_safety` and applies.

Residual (noted): an object uploaded by a promote that then crashes before the HEAD bump is never referenced and never superseded, so it isn't reclaimed by this path — a periodic backstop sweep is a smaller follow-up.

PR #11936, opened 2026-07-20
fix(infra): recover wedged macOS runner hosts + cache-volume observability + JIT host-unwedge

fortmarek:
## What and why

Started as a triage of the per-account macOS runner cache volume (spec #76): 3 of 9 prod `runners-fleet-mndbc-*` hosts were reported booting VMs with **no** `/Volumes/My Shared Files/cache` share. This PR carries the fixes that came out of it, plus a bootstrap hotfix surfaced while remediating.

### Triage result

Not a runtime volume bug. The two hosts with the no-share signature (`m42wr`, `xs6fx`) are **SSH-wedged**: their CAPI drift roll fails with `dial tcp <ip>:22: i/o timeout`, they hit terminal `TartKubeletUpdateExceededRetries`, and `shouldClearTerminalFailure` only clears when `desired != failed` (both are the current hash), so they are stranded on the pre-current config `3c1d0bb6` and never received the cache-volume machinery. The other 7 hosts are converged and healthy. Key gotcha for future triage: the `tart_kubelet_cache_volume_*` metric families are registered unconditionally, so their *presence* proves nothing. The CAPI `hostConfigHash` and the metric *values* are the decisive signals.

## Changes

**1. tart-kubelet observability + startup guard** (`infra/tart-kubelet/`)
- Startup mount check with bounded retry: if `--runner-cache-root` is set but the path is not a mounted volume, `Start` logs loudly and retries before sweeping/evicting against a bare mountpoint, instead of failing every `AllocateBranch` silently. New `volumeBackend.isMounted` (darwin = `st_dev`-vs-parent mountpoint check, which `df` cannot fool).
- New gauge `tart_kubelet_cache_volume_root_mounted` (1/0), refreshed each reconcile tick, disambiguates "unmounted volume" from "feature disabled" (the exact ambiguity that made this triage need the CAPI hash).
- `errNoRoom` admission declines now log + increment `tart_kubelet_cache_volume_admission_declined_total`. Previously a host wedged under disk pressure looked identical to an idle one.

**2. Bootstrap hotfix** (`infra/macos-host-bootstrap/bootstrap.go`) — the urgent one
- `renderSSHReachabilityScript` tee'd into `/usr/local/bin` without creating it. On the first-boot `Run()` path it executes before `installTart` (the step that does the `mkdir`), so a freshly imaged pool server has no `/usr/local/bin` and the whole bootstrap fails on the `install ssh reachability` step. The converged fleet never hit this (they got that step later via the drift path). Surfaced live when a recreated host could not bootstrap.
- Fix: `sudo mkdir -p /usr/local/bin` at the top of the script, plus a regression assertion.

**3. JIT elevation: host recovery without break-glass** (`infra/helm/pomerium/templates/access-tiers.yaml`)
- New `tuist-fleet-unwedge` ClusterRole (machines delete; scalewayapplesiliconmachines(/status) patch; nodes patch for cordon; pods/eviction for drain), bound **directly** to the per-env `tuist-<env>-write` elevation group rather than aggregated into `edit`. delete-machines is sensitive enough that it must not ride on the generic `edit` tier, where a future RoleBinding could silently inherit it. Every machine in the workload cluster is MachineSet-owned, so a delete can only recreate a worker host. Fleet scaling / teardown stays break-glass.

## Validation

- `infra/tart-kubelet`: `go test ./...`, `go vet`, `gofmt`, linux cross-build all green (new tests for the mount check, admission metric, and darwin `isMounted`).
- `infra/macos-host-bootstrap`: `go build`/`vet`/`test`/`gofmt` green; regression assertion added.
- `infra/helm/pomerium`: `helm template` renders; confirmed `edit` is bound to exactly one subject (`tuist-<env>-write`) in prod today.

## Deploy notes

- Change-sets 1 + 2 ship via the **capi-scaleway** component (includes `infra/tart-kubelet/**` and `infra/macos-host-bootstrap/**`). Deploying the operator with change-set 2 is what recovers the stuck replacement host and makes future host recreates safe.
- Change-set 3 ships via the **pomerium** chart.

Draft: opened for review while the stuck prod hosts are being remediated with break-glass in parallel.


pepicrft:
I did a deep pass over the current head, including the merge from main.

The earlier review findings are addressed:

- Cache allocation now refuses an unmounted root, so it cannot spill cache images onto the boot volume.
- The mount wait runs before state recovery on the successful-mount path.
- The enabled metric distinguishes a disabled feature from an enabled but unmounted root.
- The fresh-host bootstrap now creates `/usr/local/bin` before installing the secure shell reachability helper.

I also verified the Apple File System mount detection against a real temporary mounted volume, rendered the Helm chart, inspected the production Machine ownership model, and ran the complete Go tests, static analysis, race detection, formatting checks, repeated focused tests, and the Linux cross-build. The relevant GitHub checks are green.

One narrow follow-up remains worth tracking. If `AwaitMountedRoot` exhausts its bounded retries and the volume mounts later, recovery may have built an empty retained set before `Start` sweeps branches. The safest follow-up is to keep the volume manager inactive until restart after a timeout, or rerun recovery before activating a late mount.

Given that this requires the volume to remain unmounted past startup and then appear after recovery, while this pull request materially improves the current operational failure modes, I do not think it should block the fix.

pepicrft (approved):
Approved after the deep review. The late-mount lifecycle edge case is documented in the conversation as a focused follow-up.

PR #11942, opened 2026-07-20
perf(server): bound rolling alert queries by key range

pepicrft:
## What changed

Rolling test alert evaluations now respect their configured cadence, coalesce work into one job per project and cadence, and split changed test identifiers into bounded contiguous primary-key ranges before querying ClickHouse.

Only established rolling metric alerts use the ingestion-scoped path. Calendar-window alerts and rolling alerts without a baseline remain on the periodic full-evaluation path, preventing both job shapes from evaluating the same alert concurrently.

Compatible alerts that use the same rolling window and aggregate column share a measurement query. Their evaluation cursors are advanced with one monotonic bulk update, so an alert that is already further ahead cannot be moved backwards by a shared evaluation.

## Why

The [slow ClickHouse query alert](https://tuist.grafana.net/alerting/grafana/ffeb6l2ax5qtcf/view?orgId=1) recorded the rolling test query 37,515 times over 24 hours. The rule measures the 99th percentile latency of each query and fires above 1.5 seconds. Atlas showed the targeted query reaching approximately 3.4 seconds.

Reducing execution frequency was useful, but it did not address the individual query latency measured by the alert. The query also needed to read less data on every execution.

## Root cause

Ingestion-driven evaluations were delayed only by the ClickHouse flush interval and ignored each alert's configured cadence. Their uniqueness key was scoped to an individual alert, so alerts for the same project independently repeated compatible work.

The rolling aggregate tables are ordered by project and test identifier, with a sparse primary index. A single identifier set sampled across a large project spans most of that key space, forcing ClickHouse to read nearly every relevant granule even when the set contains relatively few tests. Increasing the identifier batch size amortized executions but retained this project-wide access pattern.

## Approach

The changed-test query already returns identifiers in the rolling table's primary-key order. The worker now preserves that order and selects a range size that produces at least four nonempty ranges whenever four or more identifiers change, capped at 2,000 identifiers per range. Sets with fewer than four identifiers use point lookups. Each resulting set covers a contiguous section of the observed identifier order, allowing the primary index to prune granules between ranges.

Cadence enforcement and project-level coalescing remain because they remove redundant evaluations. Alerts with compatible rolling measurements share the aggregate merge, while reliability stays separate because it reads a different aggregate column.

A shared alert predicate now defines path ownership. The enqueue path and project-scoped worker accept only established rolling metric alerts, while the periodic scheduler handles calendar-window alerts and rolling baselines.

Shared cursor advancement uses the later of the stored cursor and the evaluated cursor for every alert. The bulk update also avoids one database update per alert.

## Impact

The production-shaped Atlas plan for a 256-identifier range selected 10 of 514 granules. Across a fixed five-minute window containing 14,895 changed tests, all eight contiguous ranges completed between 0.27 and 0.60 seconds. Every sampled execution was below the 1.5-second alert threshold. Production monitoring after deployment will confirm the resulting 99th percentile latency.

Alert thresholds, transitions, recovery behavior, and the configured cursor lookback duration remain unchanged. No data migration is required.

## Validation

- `mix test test/tuist/automations_test.exs test/tuist/automations/alerts/alert_test.exs test/tuist/automations/monitors/flaky_tests_monitor_test.exs test/tuist/automations/workers/alert_evaluation_worker_test.exs test/tuist/automations/workers/automation_scheduler_test.exs`
  - 136 tests passed.
- `mix compile --warnings-as-errors`
- Targeted `mix credo --strict`
  - No issues found across the relevant automation source and test files.
- Formatting and `git diff --check` passed.
- Atlas returned identical result counts and digests when the same identifier set was evaluated as one set and as four disjoint subsets.
- Atlas reproduced both production flakiness thresholds from the shared measurement result.

### DIFF
--- a/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
+++ b/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
@@ -64,7 +64,7 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
       %{test_case_ids: test_case_ids, cursor: cursor} = Automations.recent_test_case_run_changes_for_alert(alert)
 
       test_case_ids
-      |> Enum.chunk_every(Automations.scoped_evaluation_chunk_size())
+      |> Automations.scoped_evaluation_ranges()
       |> Enum.each(&evaluate_and_execute(alert, &1))
 
       {:ok, _alert} = Automations.update_alert_scoped_evaluation_cursor(alert, cursor)
@@ -86,12 +86,10 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
         Automations.recent_test_case_run_changes_for_alerts(established_alerts)
 
       test_case_ids
-      |> Enum.chunk_every(Automations.scoped_evaluation_chunk_size())
+      |> Automations.scoped_evaluation_ranges()
       |> Enum.each(&evaluate_alert_group(established_alerts, &1))
 
-      Enum.each(established_alerts, fn alert ->
-        {:ok, _alert} = Automations.update_alert_scoped_evaluation_cursor(alert, cursor)
-      end)
+      {:ok, _updated_count} = Automations.advance_alert_scoped_evaluation_cursors(established_alerts, cursor)
     end
 
     :ok
--- a/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
+++ b/server/lib/tuist/automations/workers/alert_evaluation_worker.ex
@@ -21,6 +21,24 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
   @recovery_candidate_batch_size 500
 
   @impl Oban.Worker
+  def perform(%Oban.Job{
+        args: %{
+          "project_id" => project_id,
+          "cadence_seconds" => cadence_seconds,
+          "evaluate_recent_test_case_runs" => true
+        }
+      }) do
+    alerts =
+      project_id
+      |> Automations.list_alerts()
+      |> Enum.filter(fn alert ->
+        alert.enabled and Alert.recovery_ledger?(alert) and
+          Alert.cadence_seconds(alert.cadence) == cadence_seconds
+      end)
+
+    evaluate_recent_test_case_runs_and_execute(alerts)
+  end
+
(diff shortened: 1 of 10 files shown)


server/lib/tuist/automations/workers/alert_evaluation_worker.ex
```diff
@@ -21,6 +21,24 @@ defmodule Tuist.Automations.Workers.AlertEvaluationWorker do
...
+          "project_id" => project_id,
+          "cadence_seconds" => cadence_seconds,
+          "evaluate_recent_test_case_runs" => true
+        }
+      }) do
+    alerts =
+      project_id
+      |> Automations.list_alerts()
+      |> Enum.filter(fn alert ->
+        alert.enabled and Alert.recovery_ledger?(alert) and
```
fortmarek: Could we avoid including `last_days` alerts in this project-scoped path, or otherwise dedupe it against `AutomationScheduler`? The scheduler continues to enqueue established `last_days` alerts by `alert_id`, while this job is unique by `project_id`/`cadence_seconds`. Since those uniqueness shapes cannot conflict, sustained ingestion can produce both a scheduled full evaluation and this scoped evaluation for the same alert each cadence. Beyond duplicate ClickHouse work, concurrent runs can both observe the pre-transition ledger and execute the same trigger/recovery action. A regression test that enqueues both paths would make the intended coordination explicit.

fortmarek (commented):
I found one potential scheduling interaction to address.

fortmarek (approved):

server/lib/tuist/automations/workers/alert_evaluation_worker.ex
pepicrft: Good catch. I restricted the project-scoped path to established rolling metric alerts through `Alert.scoped_evaluation?/1`. Both `Automations.enqueue_flaky_alert_evaluations/2` and the project-scoped `AlertEvaluationWorker` now use that predicate, while `AutomationScheduler` uses the inverse ownership rule for calendar-window alerts and rolling alerts whose baseline is not established.

I also added a regression test in `AutomationSchedulerTest` that invokes both enqueue paths for a calendar-window alert and asserts that only the scheduler's per-alert job exists. The complete focused automation suite passes with 136 tests, along with compilation, formatting, targeted Credo, and diff checks. Fixed in e153c0aa80.

PR #11943, opened 2026-07-20
perf(kura): build RocksDB as a Bazel cc_library instead of the build script

fortmarek:
## What changed

RocksDB is now compiled as a first-class Bazel `cc_library` (`//bazel/third_party/rocksdb`) from the sources vendored in the `librocksdb-sys` crate, instead of being cold-compiled inside the crate's cargo build script. The build script is redirected via `ROCKSDB_LIB_DIR`/`ROCKSDB_STATIC` so it still runs `bindgen` against the crate's vendored headers but skips `build_rocksdb()`.

New files, all under `kura/`:
- `bazel/third_party/rocksdb/rocksdb_lib_sources.bzl` — the exact 328-file RocksDB `LIB_SOURCES` list.
- `bazel/third_party/rocksdb/lib_dir.bzl` + `BUILD.bazel` — materialize `librocksdb.a` into a directory artifact for `ROCKSDB_LIB_DIR`.
- `MODULE.bazel` — the injected `cc_library` and the `librocksdb-sys` build-script redirect.

## Why

RocksDB was compiled from source inside the `librocksdb-sys` cargo build script on every build. That action shells out to `cc`, so it is opaque to Bazel and caches unreliably — a **remote-cache miss**. In the `build-kura-runtime` deploy job it ran locally for **~14 min** (the build ended `928 processes: 420 remote cache hit, ..., 50 processwrapper-sandbox`, and this was one of the misses), dominating the compile step of a ~26-min job.

This is the same move already made in this repo for `lz4` (`@lz4`): moving `-sys` C/C++ compiles out of build scripts into hermetic `cc_library` targets. (Kura applied the identical pattern to Lua/mlua too, until #12264 dropped the Lua dependency entirely.) RocksDB is now the last `-sys` crate still compiling in its build script. A hermetic `cc_library` is deterministic, cross-compilable via the Bazel cc toolchain, and served from the remote cache — build once, download thereafter.

## Why 10.4.2 (and why not the BCR module)

`librocksdb-sys` runs `bindgen_rocksdb()` **unconditionally against the headers vendored in the crate**; `ROCKSDB_LIB_DIR` only redirects linking, not bindgen. So the linked library must match the vendored headers. Building from the crate's own vendored source guarantees that automatically.

`rocksdb 0.24.0` / `librocksdb-sys 0.17.3+10.4.2` (RocksDB 10.4.2) is the latest released crate. The Bazel Central Registry `rocksdb` module tops out at 9.11.2 (an open PR adds 11.1.2, but it is stalled on a CLA and is ahead of any released crate), so it can't be paired without an ABI mismatch. Hence we compile the crate's vendored 10.4.2 sources rather than depending on the BCR module.

## Implementation notes for reviewers

- The `srcs`/`includes`/`local_defines`/`copts` mirror the crate's `build.rs` exactly: the curated `LIB_SOURCES` list (a glob does **not** reproduce it — it wrongly pulls `db_bench_tool.cc`/`ldb.cc` mains, `port/win/*`, and the deliberately-omitted `range_tree` lock manager), LZ4-only backend, `-std=c++17`, `-msse2` on x86_64, `OS_MACOSX` + `-include cstdint` on macOS. This keeps the compiled ABI identical to the bindgen'd headers.
- The crate ships a pre-generated `build_version.cc` (git sha/tag baked in), so it's compiled as-is — no `expand_template`.
- The `cc_library` is injected into the crate's package via `additive_build_file_content` so its includes resolve against the vendored `rocksdb/` tree; that content does not inherit our module's apparent repo names, so it uses canonical `@@rules_cc+` / `@@lz4+` / `@@platforms` / `@@//`.
- The now-inert `DEP_LZ4_INCLUDE` wiring is kept as a cold-build fallback (used only if someone forces `ROCKSDB_COMPILE`).

## Validation (macOS arm64)

- `bazel build //:kura`: RocksDB compiles as parallel `cc` actions in ~4.8 min cold (vs ~19 min on the 2-vCPU CI runner); the `librocksdb-sys` build script no longer compiles the C++.
- `bazel test //:kura_lib_test`: **passes** — confirms the FFI bindings still match the linked 10.4.2 lib (a version/ABI mismatch would surface here as a link error or crash).
- `bazel test //...`: **344 passed, 0 failed**. `mise run clippy`: clean.
- Rebuild is a full cache hit (0 recompiles), confirming the action is deterministic and remotely cacheable.
- `--config=macos-x86_64-cross` analyzes cleanly (the cross target the release pipeline uses).

Not exercised locally: the linux/amd64 compile (the `build-kura-runtime` runner target) — the defines mirror `build.rs`'s linux branch but only run in CI. The separate ~4.3-min `moby/buildkit` Docker Hub pull in that job is out of scope.


esnunes (approved):
Approving — the change is correct and the win is real. I reproduced it as an A/B on the same 24-core Linux box, both sides a fresh clone with no remote or disk cache:

| Branch | Wall | Critical path | Sandboxed actions |
|---|---|---|---|
| `main` (a6f0ff6) | 882.7s | 876.3s | 404 |
| this PR | 229.0s | 177.0s | 1065 |

`main` sits at `857 / 863` for over 11 minutes on the `librocksdb-sys` build script alone, at load average 1.6 across 24 cores — `rules_rs` hands build scripts `NUM_JOBS=1`, so cc-rs compiles all 328 files serially. That's the whole story, and it's why this also helps the 2-vCPU runner disproportionately.

Also verified, since these are the parts that break silently: `rocksdb_lib_sources.bzl` is byte-identical to the crate's `rocksdb_lib_sources.txt` minus `util/build_version.cc`; no optimization regression (cc-rs appends the env `CXXFLAGS` *after* its own `-O3` — see `get_compiler`, "Set flags from the environment (do this last…)" — so Bazel's `-O2` already won before this PR); `-msse2` alone matches the previous baseline because nothing sets `-Ctarget-cpu`, so `CARGO_CFG_TARGET_FEATURE` never enabled sse4.2/pclmul either; and RocksDB compiles exactly once — the exec-configuration output tree holds 0 objects on both macOS and Linux. `bazel build //:kura` and `bazel test //:kura_lib_test` both pass on macOS arm64, and the build plus a binary smoke test pass on linux/amd64.

One non-blocking suggestion, for a follow-up if you prefer:

**`ROCKSDB_LIB_DIR` doesn't need to hold the real archive, and pointing it at an empty directory would take RocksDB off the critical path.** `try_to_find_and_link_lib` only reads env vars — it prints `-L`/`-l` and returns, never opening `librocksdb.a`. So the `lib_dir` dependency makes the build script (and therefore bindgen, the `librocksdb-sys` rlib, and everything downstream of it) wait for the complete C++ compile, when the link is already satisfied by the `deps` edge. Handing it an empty declared directory would let bindgen and the C++ compile overlap; on my Linux run the critical path was 177s of a 229s build, so there's headroom there. It would also drop the second 44 MB copy of the archive that `lib_dir` currently materializes and pushes to the remote cache on every build.

The argument for keeping it as-is is robustness: if `rules_rs` ever resolves `-lrocksdb` through that search path rather than through `deps`, a real archive is the safe thing to find. Either way is fine by me — worth one line in `lib_dir.bzl` recording which of the two it is, so the next person doesn't have to re-derive it from `build.rs`.

fortmarek:
Thanks for the thorough review and the A/B numbers.

I tried the empty-`ROCKSDB_LIB_DIR` idea and it actually fails the build — the `librocksdb-sys` rlib compile errors with:

```
error: could not find native static library `rocksdb`, perhaps an -L flag is missing?
```

So `rules_rs` resolves the build script's emitted `-l static=rocksdb` through the `ROCKSDB_LIB_DIR` search path at the rlib compile stage, **not** through the `deps` edge — the materialized archive is load-bearing, and the robustness case you raised is the actual behavior rather than a hypothetical. Because the rlib genuinely needs to find `librocksdb.a`, the critical-path / redundant-copy win isn't reachable without breaking the link, so I kept the real archive.

Documented the finding and the exact failure mode in `lib_dir.bzl` (14faf99) so it doesn't get re-derived from `build.rs`, and dropped the now-stale `//bazel/third_party/lua` reference (main removed the Lua dependency in #12264). Re-verified `bazel build //:kura` + `bazel test //:kura_lib_test` green on macOS arm64 after the change.

PR #11947, opened 2026-07-20
fix(app): export the shared device preview ad-hoc so it can be installed on devices

fortmarek:
## What

The Tuist iOS app preview shared to tuist.dev was signed for App Store distribution, which cannot be installed on a device over the air. This exports the shared device preview with an ad-hoc (`release-testing`) method instead, while keeping the App Store Connect submission path unchanged.

## Root cause

`mise/tasks/app/bundle-ios.sh` produced `build/Tuist.ipa` with export `method: app-store-connect` (Apple Distribution certificate plus the "Tuist App Store Connect" profile, `get-task-allow: false`, no provisioned devices). App Store distribution builds can only be vended through the App Store or TestFlight; installing one via the preview's over-the-air flow fails with "This app cannot be installed because its integrity could not be verified."

The same script is reused by the App Store Connect submission (`mise/tasks/app/upload-ios.sh`), so it must keep producing an app-store-connect build for that path. The ad-hoc profile ("Tuist App Ad Hoc") was already being downloaded in the script but never used, and `app/Project.swift` declared the ad-hoc profile specifier for the release configuration, so ad-hoc was the intent for the shared build.

## Change

- `bundle-ios.sh` takes an `EXPORT_METHOD` env var (default `app-store-connect`, so the submission path is unchanged) and selects the matching provisioning profile.
- The `Device Build` job in `.github/workflows/app.yml` sets `EXPORT_METHOD: release-testing` (ad-hoc), using the "Tuist App Ad Hoc" profile.

## Impact

The shared device preview of the Tuist app becomes installable over the air on devices registered in the ad-hoc profile. Ad-hoc builds only install on registered devices, which is inherent to ad-hoc distribution.

### How to test locally

This needs the release signing assets, so it is exercised in CI rather than locally: run the `App` workflow (push, dispatch, or a team-member PR) and confirm the Device Build job exports with `EXPORT_METHOD: release-testing`, then install the resulting preview on a device registered in the "Tuist App Ad Hoc" profile.

Validated: triggered the workflow on this branch, the Device Build job exported ad-hoc and shared successfully, the resulting `.ipa` is signed with the "Tuist App Ad Hoc" profile (6 provisioned devices, was 0 before), and it installed end to end on a registered device from a staging preview.

### DIFF
--- a/.github/workflows/app.yml
+++ b/.github/workflows/app.yml
@@ -232,6 +232,8 @@ jobs:
           github.event_name == 'push' ||
           github.event_name == 'workflow_dispatch' ||
           (github.event_name == 'pull_request' && needs.check-team-membership.outputs.is-team-member == 'true')
+        env:
+          EXPORT_METHOD: release-testing
         run: mise run app:bundle-ios
       - name: Inspect TuistApp
         if: |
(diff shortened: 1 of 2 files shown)


fortmarek:
Folded into #11948 together with the server download fix and the Cloudflare change.

Issue #11955, opened 2026-07-21
kura: a slow cold bootstrap is indistinguishable from a hang — both-unready 2-pod region 503s for the whole pull, with observability gaps


esnunes:
## Summary

On 2026-07-21 the `tuist`-tenant eu-central Kura mesh (`kura-tuist-eu-central-1`, a 2-pod StatefulSet on the Scaleway dedibox pool) served **503 for hours** because both replicas were `0/1` not-Ready. `/ready` gates on bootstrap *completion*, and both replicas were doing a **cold bootstrap of a large dataset** (~160K+ manifests) at ~50 artifacts/sec, which takes ~1h+.

The point of this issue is **not** that a cold node is slow. It's that **a slow-but-progressing bootstrap is indistinguishable from a hang** with the observability we currently have — which turned a one-line "let it finish" into hours of misdiagnosis and several unnecessary (and counterproductive) production interventions.

## What was actually happening

- Each pod restarts cold (little/no local index), computes manifest-bucket divergence against its peers (logged once per peer), then fetches the diverged artifacts.
- Confirmed only via the **source peer's server-side** metric: `kura-tuist-scw-fr-par-0` served `/_internal/bootstrap/artifacts/{id}` at a sustained **~44–63 req/s for 43+ minutes** (plus ~1.5–3 `/_internal/bootstrap/manifests` pages/s) — active, convergent download.
- pod-1 reached ~62% of buckets matched; pod-0 was the long pole (~5%). Dataset ≈ 160,384 manifest index entries on the source, so a full cold pull is ~tens of minutes to 1h+ at that rate.
- Because it's a **2-pod region and both were cold at once, there were zero serving endpoints → the whole tenant region 503s for the entire bootstrap.**

## Why it was nearly impossible to tell "progressing" from "hung"

1. **Bootstrap emits no progress after the divergence scan.** It logs `bootstrap from <peer>: N/4096 buckets diverged` once per peer, then nothing until completion. 38+ minutes of total log silence looked exactly like a hang.
2. **These pods can't be `kubectl logs` / `exec` / `debug`'d.** The dedibox bare-metal nodes have an unresolvable kubelet hostname (`dial tcp: lookup …dedibox-fleet-… : no such host`), so every kubelet-proxied stream fails.
[...]

PR #11956, opened 2026-07-21
feat(kura): make bootstrap progress observable

esnunes:
Works towards #11955 — implements the improvement that issue ranks highest value: making bootstrap progress observable.

## What changed

- **Per-peer bootstrap progress tracking.** A new `BootstrapProgress` tracker (in `src/state.rs`) records the pass's phase (`namespace_tombstones` → `digest_reconcile` → `manifest_fetch`) plus counters: tombstones applied, diverged buckets total/completed, manifest pages fetched, artifacts applied and failed, and elapsed time. Each bootstrap pass registers its tracker in an `AppState` registry keyed by peer URL, and deregisters when the pass completes or is abandoned.
- **Periodic INFO progress log.** The bootstrap task now emits a progress line every 30 seconds (a new select branch next to the no-progress watchdog), e.g. `bootstrap from <peer> in progress: 120/4096 diverged buckets (~2.9%), 51230 artifacts applied (~48/s), 0 failed, 231 pages fetched, 1071s elapsed`. When the peer predates the digest endpoint (full-walk fallback), the bucket denominator is omitted rather than invented.
- **Progress on `/ready` and `/status/rollout`.** Both bodies now carry a `bootstrap_progress` array with the same per-peer snapshots, so an operator (or the rollout gate's data source) can confirm forward progress with one request even when logs are inconvenient to reach.
- The tracker also carries the existing no-progress watchdog tick counter, replacing the bare `AtomicU64` — watchdog semantics are unchanged (page fetches and applied artifacts tick it, exactly as before).
- Docs: `docs/architecture.md` replication section and the README feature list now describe the progress surface.

## Why

In the 2026-07-21 incident (#11955), both replicas of a 2-pod region cold-bootstrapped a ~160K-manifest dataset at ~50 artifacts/s (~1h+). During that window the pass was completely silent after the one-shot `N/4096 buckets diverged` line, and the pods — being NotReady — were absent from the Service endpoints Prometheus scrapes, so `kura_bootstrap_*` metrics didn't exist either. "Slow but converging" and "hung" were externally indistinguishable; confirming progress required inferring it from the *source peer's* `/_internal/bootstrap/*` request rate, and the ambiguity triggered counterproductive interventions (restarting an in-flight bootstrap).

Logs and status endpoints are exactly the two channels that survive that failure mode: structured logs flow to Loki regardless of readiness (and independent of the dedibox kubelet-hostname issue that breaks `kubectl logs`), and `/ready` / `/status/rollout` are reachable in-cluster (the rollout gate already relies on exactly that assumption).

### Why this shape

- A registry on `AppState` (rather than counters local to the bootstrap task) is what lets the HTTP handlers see the numbers; the periodic logger doubles down on it so both surfaces report identical figures.
- The per-peer tracker is re-created per pass, so a retried pass restarts its counters instead of accumulating misleading totals across attempts.
- Bucket completion percent is an approximation (buckets vary in size) and is labeled as such (`~`); it's the only denominator known up front, fixed by the digest exchange before the walk starts.
- The 30s log interval is a constant, not config: it's an observability heartbeat, not a tuning knob, and avoids a Helm values change.

Rollout safety: additive only — no wire-format, on-disk, or protocol changes; the new JSON fields are additions to existing bodies, so mixed-version meshes and older clients are unaffected.

The issue's other proposals (scraping not-Ready pods via PodMonitor, shrinking the both-unready 503 window, disambiguating drain-time fs aborts) are intentionally not in this PR.

## Validation

- `mise run test-unit` (Bazel): 373 passed, 0 failed — includes new tests:
  - snapshot counters/percent math and watchdog tick aggregation,
  - `describe()` output for each phase, including the no-digest fallback,
  - registry lifecycle (register, sorted snapshots, per-pass reset, deregister),
  - `/ready` reporting an in-flight pass's `bootstrap_progress` and returning `[]` after it ends,
  - existing watchdog tests extended to assert the tracker deregisters on both completion and abandonment.
- `mise run clippy` — clean (warnings as errors).
- `mise run format -- --check` — clean.

### DIFF
--- a/kura/README.md
+++ b/kura/README.md
@@ -183,6 +183,7 @@ Replication is leaderless and eventually consistent:
 
 - 🔁 local writes become durable together with their outbox work
 - 🌍 peers bootstrap by pulling manifests, tombstones, and artifact bodies, reconciling only the diverging ranges via a per-bucket manifest digest exchange
+- 📈 in-flight bootstraps log a progress line every 30 seconds and report per-peer counters (phase, buckets completed, artifacts applied/failed) as `bootstrap_progress` on `/ready` and `/status/rollout`, so a long cold pull is distinguishable from a hang
 - 🔎 DNS discovery can expand the peer set automatically
 - 🧠 the outbox is processed incrementally so queue depth does not blow up heap usage during backlog
 
--- a/kura/docs/architecture.md
+++ b/kura/docs/architecture.md
@@ -105,6 +105,8 @@ Before walking a peer's manifests, bootstrap runs **range-digest anti-entropy** 
 
 A per-peer bootstrap runs under a **no-progress watchdog** rather than a fixed total-runtime cap. `KURA_BOOTSTRAP_TIMEOUT_MS` is the maximum time the bootstrap may go *without* forward progress (a fetched page or applied artifact); a bootstrap that keeps advancing runs to completion however long that takes, and only a genuinely stalled one is abandoned and retried. A single wall-clock cap could never let a large cold pull finish — it would be killed and restarted from scratch every window — so a node whose backlog exceeds one window's worth of transfer would stay `NotReady` indefinitely.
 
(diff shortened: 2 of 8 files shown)


esnunes:
no longer needed, PR https://github.com/tuist/tuist/pull/12171 already makes bootstrap observable

PR #11960, opened 2026-07-21
feat(cli): make network retry policy configurable

pepicrft:
## What changed

The command-line client now resolves [Hypertext Transfer Protocol (HTTP)](https://developer.mozilla.org/en-US/docs/Web/HTTP) retry behavior from a shared policy. It adds two environment variables:

- `TUIST_HTTP_MAXIMUM_RETRY_COUNT` controls retries after the initial attempt, defaults to `3`, and is capped at `10`.
- `TUIST_HTTP_RETRY_BASE_DELAY_IN_MILLISECONDS` controls the initial exponential backoff delay, defaults to `100`, and is capped at `30000`.

The request middleware and server retry provider use the same policy. Tests cover defaults, overrides, disabled retries, invalid and extreme values, exponential backoff, jitter, and computed delay limits. The debugging guide, intent documentation, and product changelog describe the new behavior.

## Why

Self-hosted operators may raise network timeouts so slow but progressing cache transfers can complete. A fixed retry count makes that longer timeout multiply the worst-case wait, so operators need to tune retries and backoff together.

## Root cause

`DelayProvider` passed `1_000_000` to `Task.sleep(nanoseconds:)`, producing a one-millisecond base delay despite the intended 100 milliseconds. Its jitter was similarly too small to spread retries after simultaneous failures. The retry provider also hardcoded three retries, resulting in up to four attempts with no override.

## Approach

A shared policy keeps both retry paths aligned. It preserves three retries and a 100-millisecond base delay as defaults, accepts zero retries, and falls back safely for malformed environment values. It caps retry count at 10 and every computed jittered delay at 30 seconds, including exponential or arithmetic overflow, so an extreme configuration remains recoverable.

## Impact

Existing users keep the intended defaults. Operators can reduce retries when using longer resource timeouts or increase the base delay to spread retries under congestion. No migration is required.

## Validation

Twenty-eight focused tests passed across five suites, and the command-line executable built successfully.

### How to test locally

- `mise run cli:lint --fix`
- `tuist generate tuist TuistHTTP TuistHTTPTests TuistServer TuistServerTests ProjectDescription --no-open`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistHTTPTests/HTTPRetryPolicyTests -only-testing TuistServerTests/DelayProviderTests -only-testing TuistServerTests/RetryProviderTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistHTTPTests/RetryMiddlewareTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `xcodebuild build -workspace Tuist.xcworkspace -scheme tuist CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY="" COMPILATION_CACHE_ENABLE_CACHING=NO`
- `cd server && mix test test/tuist/marketing/changelog_entry_parser_test.exs`

### DIFF
--- a/cli/Sources/TuistHTTP/HTTPRetryPolicy.swift
+++ b/cli/Sources/TuistHTTP/HTTPRetryPolicy.swift
@@ -1,0 +1,61 @@
+import Foundation
+import TuistEnvironment
+
+public struct HTTPRetryPolicy: Sendable {
+    public static let defaultMaximumRetryCount = 3
+    public static let defaultBaseDelayMilliseconds: UInt64 = 100
+
+    public let maximumRetryCount: Int
+    public let baseDelayMilliseconds: UInt64
+
+    public init(
+        maximumRetryCount: Int? = nil,
+        baseDelayMilliseconds: UInt64? = nil,
+        environment: [String: String] = Environment.current.variables
+    ) {
+        let resolvedMaximumRetryCount = maximumRetryCount ?? Self.maximumRetryCount(environment: environment)
+        let resolvedBaseDelayMilliseconds = baseDelayMilliseconds ?? Self.baseDelayMilliseconds(environment: environment)
+
+        self.maximumRetryCount = max(0, resolvedMaximumRetryCount)
+        self.baseDelayMilliseconds = Self.validatedBaseDelayMilliseconds(resolvedBaseDelayMilliseconds)
+    }
+
+    public func delay(for retry: Int) -> UInt64 {
+        let baseDelayNanoseconds = baseDelayMilliseconds * 1_000_000
+        guard baseDelayNanoseconds > 0 else { return 0 }
+
+        let exponent = max(0, retry)
+        guard exponent < UInt64.bitWidth else { return .max }
+
+        let exponentialDelay = baseDelayNanoseconds.multipliedReportingOverflow(by: UInt64(1) << exponent)
+        guard !exponentialDelay.overflow else { return .max }
+
+        let jitter = UInt64.random(in: 0 ... baseDelayNanoseconds)
+        let delay = exponentialDelay.partialValue.addingReportingOverflow(jitter)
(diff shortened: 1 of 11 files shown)


cli/Sources/TuistHTTP/HTTPRetryPolicy.swift
```diff
@@ -0,0 +1,61 @@
...
+        self.maximumRetryCount = max(0, resolvedMaximumRetryCount)
+        self.baseDelayMilliseconds = Self.validatedBaseDelayMilliseconds(resolvedBaseDelayMilliseconds)
+    }
+
+    public func delay(for retry: Int) -> UInt64 {
+        let baseDelayNanoseconds = baseDelayMilliseconds * 1_000_000
+        guard baseDelayNanoseconds > 0 else { return 0 }
+
+        let exponent = max(0, retry)
+        guard exponent < UInt64.bitWidth else { return .max }
```
Hoontopia: That value flows straight into Task.sleep(nanoseconds:), and UInt64.max nanoseconds is roughly 585 years, so the process hangs silently instead of failing.
  
Both variables are clamped at the bottom with max(0, ...) but not at the top. Malformed values fall back to the defaults, while well formed but extreme ones pass through.

Clamping to a ceiling, say 30 seconds, instead of saturating to .max would keep a misconfiguration recoverable.

Hoontopia (approved):

PR #11968, opened 2026-07-21
fix(server): bind shard downloads to exact plans

pepicrft:
## What changed

Shard matrices now carry the exact shard plan identifier from the build job into every test job. `tuist test`, `tuist xcodebuild test`, and `tuist xcodebuild test-without-building` accept it through `--shard-plan-id` or `TUIST_SHARD_PLAN_ID`.

The server resolves a shard by that exact identifier, scoped to the selected project. Reference-based lookup remains available for older clients. The generated server client, provider examples, documentation, and regression coverage have been updated accordingly.

CircleCI keeps emitting its existing continuation parameters by default. It only adds `shard-plan-id` after `TUIST_CIRCLECI_SHARD_PLAN_ID_ENABLED` is set to `true`, allowing workflows to declare the new parameter before opting in.

## Why

Customers can create several sharded test plans in parallel under the same GitHub workflow run. Each test job must download products from the exact plan created by its corresponding build job.

## Root cause

Legacy shard reads resolved the newest plan matching a run-level reference. When parallel matrix roots reused that reference, a later plan could replace the plan selected by an earlier test job. The job would then request an artifact that belonged to another plan and might not exist yet.

## Approach

The immutable plan identifier is propagated through generated provider outputs and sent with shard requests. It takes precedence on the server, while the existing reference remains a compatibility fallback.

GitHub matrix output retains the existing `shard` array and adds the plan identifier through `include`, so workflows that only consume the shard indices continue to work.

CircleCI is deliberately opt-in because its continued configuration must declare every continuation parameter. This avoids changing the payload for existing workflows while giving users a documented migration path to exact-plan selection.

Malformed plan identifiers are rejected during request validation before reaching the database.

## Impact

Workflows that propagate the exact plan identifier select the correct artifacts when several plans run concurrently. Existing reference-only workflows keep running without changes, but remain exposed to the concurrent-plan race until they carry the identifier from the build job into each test job. CircleCI users must opt in after declaring the new continuation parameter.

### How to test locally

- `mix test test/tuist/shards_test.exs test/tuist_web/controllers/api/shards_controller_test.exs`
- `mix format --check-formatted test/tuist_web/controllers/api/shards_controller_test.exs`
- `xcodebuild test -quiet -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistKitTests/ShardMatrixOutputServiceTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistKitTests/ShardMatrixOutputServiceTests -only-testing TuistKitTests/ShardServiceTests -only-testing TuistKitTests/XcodeBuildTestCommandServiceTests -only-testing TuistKitTests/CommandEnvironmentVariableTests -only-testing TuistKitTests/TestServiceTests/test_run_testWithoutBuilding_passesShardArchivePathToShardService CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `mise run generate-api-cli-code` from `server/`
- `mise run cli:lint`
- `git diff --check`


server/lib/tuist_web/controllers/api/shards_controller.ex
```diff
@@ -233,6 +233,11 @@ defmodule TuistWeb.API.ShardsController do
         required: true,
         description: "The shard plan reference."
       ],
+      shard_plan_id: [
+        in: :query,
+        schema: %Schema{type: :string, format: :uuid},
+        description: "The exact shard plan identifier. When present, it takes precedence over the reference."
```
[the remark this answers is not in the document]
pepicrft: Fixed in 594e1b5390. The shard operation now documents the 400 Bad Request response, the generated server contract and Swift client expose it, and GetShardService maps the response payload to badRequest(message) so callers receive the validation message. The focused controller tests pass.

cli/Tests/TuistKitTests/Services/TestServiceTests.swift
```diff
@@ -5737,6 +5741,7 @@ final class TestServiceTests: TuistUnitTestCase {
             try await testRun(
                 path: path,
                 action: .testWithoutBuilding,
+                shardPlanId: "plan-123",
```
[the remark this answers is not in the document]
pepicrft: Fixed in bfa87f8003. The touched sharding coverage now lives in TestServiceShardingTests as a Swift Testing test with the temporary-directory and mocked-dependencies traits, uses #require for the temporary directory, and is backed by a dedicated value-type fixture. The focused xcodebuild test passes.

cli/Sources/TuistServer/Services/GetShardService.swift
```diff
@@ -52,7 +55,8 @@ public struct GetShardService: GetShardServicing {
                 project_handle: handles.projectHandle,
                 reference: reference,
                 shard_index: shardIndex
-            )
+            ),
+            query: .init(shard_plan_id: shardPlanId)
```
fortmarek, line 59: **[P1] Treat an empty shard plan ID as absent**

The documented CircleCI continuation parameter defaults to `""` and is always exported as `TUIST_SHARD_PLAN_ID`. String environment parsing preserves that empty value, so this line serializes `shard_plan_id=`. The server's UUID validation rejects it before the controller can take its `plan_id != ""` fallback, breaking the intended config-first CircleCI migration and reference-only fallback. Normalize blank values to `nil` before constructing the query (or conditionally avoid exporting the variable), and cover the empty environment value.

cli/Sources/TuistKit/Services/Sharding/ShardMatrixOutputService.swift
```diff
@@ -114,32 +134,35 @@ public struct ShardMatrixOutputService: ShardMatrixOutputServicing {
...
 
-    private func writeCodemagicOutput(indices: [Int], cmEnvPath: String) async throws {
+    private func writeCodemagicOutput(indices: [Int], shardPlanId: String, cmEnvPath: String) async throws {
         let outputPath = try AbsolutePath(validating: cmEnvPath)
         let existing = (try? await fileSystem.readTextFile(at: outputPath)) ?? ""
         let matrixJSON = "{\"shard\":\(indices)}"
         try await fileSystem.writeText(
-            existing + "TUIST_SHARD_MATRIX=\(matrixJSON)\nTUIST_SHARD_COUNT=\(indices.count)\n",
+            existing +
+                "TUIST_SHARD_MATRIX=\(matrixJSON)\nTUIST_SHARD_COUNT=\(indices.count)\nTUIST_SHARD_PLAN_ID=\(shardPlanId)\n",
```
fortmarek: **[P2] Carry the Codemagic plan ID across workflows**

Writing `TUIST_SHARD_PLAN_ID` to `CM_ENV` only makes it available to subsequent steps in the current workflow. The documentation defines `build-shards` and each `test-shard-*` as separate workflows, and those test workflows only set `TUIST_SHARD_INDEX`, so the exact plan ID never reaches them and they remain on reference lookup. Please add an explicit cross-workflow transport/consumption path and update both examples accordingly. Codemagic documents the scope here: https://docs.codemagic.io/yaml-basic-configuration/using-environment-variables/

cli/Sources/TuistKit/Services/Sharding/ShardMatrixOutputService.swift
```diff
@@ -114,32 +134,35 @@ public struct ShardMatrixOutputService: ShardMatrixOutputServicing {
...
-    private func writeBitriseOutput(indices: [Int], deployDir: String) async throws {
+    private func writeBitriseOutput(indices: [Int], shardPlanId: String, deployDir: String) async throws {
         let deployPath = try AbsolutePath(validating: deployDir)
         let outputPath = deployPath.appending(component: ".tuist-shard-matrix.json")
         if try await fileSystem.exists(outputPath) {
             try await fileSystem.remove(outputPath)
         }
-        let matrixJSON = "{\"shard\":\(indices),\"shard_count\":\(indices.count)}"
+        let matrixJSON =
+            "{\"shard\":\(indices),\"shard_count\":\(indices.count),\"shard_plan_id\":\"\(shardPlanId)\"}"
```
fortmarek, line 161: **[P2] Consume the Bitrise plan ID in downstream workflows**

Adding the ID to this JSON file does not put it in the test workflows' environments. The documented Bitrise pipeline neither shares/reads the file nor exports `TUIST_SHARD_PLAN_ID`; every `test-shard-*` workflow still sets only the index, so the race this PR fixes remains. Wire the value through Pipeline Intermediate File Sharing or shared pipeline variables and show the corresponding consumption in both examples: https://docs.bitrise.io/en/bitrise-ci/workflows-and-pipelines/build-pipelines/configuring-a-bitrise-pipeline

fortmarek (commented):
Three issues remain in the exact-plan rollout: blank CircleCI identifiers are sent as invalid UUIDs, and the Codemagic and Bitrise outputs are not carried into the documented downstream test workflows.

cli/Sources/TuistEnvKey/EnvKey.swift
```diff
@@ -198,6 +198,8 @@ public enum EnvKey: String, CaseIterable {
     case testShardGranularity = "TUIST_TEST_SHARD_GRANULARITY"
     case testShardIndex = "TUIST_SHARD_INDEX"
     case testShardReference = "TUIST_SHARD_REFERENCE"
+    case testShardPlanId = "TUIST_SHARD_PLAN_ID"
+    case testShardCircleCIPlanIdEnabled = "TUIST_CIRCLECI_SHARD_PLAN_ID_ENABLED"
```
fortmarek: **[P2] Make the CircleCI plan-ID migration explicit instead of adding a rollout flag**

This provider-specific environment variable creates a permanent out-of-band handshake between the CLI and the user-owned continuation config: an updated config with the flag off silently keeps the race-prone behavior, while an old config with the flag on fails because CircleCI rejects the undeclared parameter. CircleCI sharding users can update their continuation configuration in the same PR that bumps Tuist, so please remove this key, always emit `shard-plan-id`, make the writer's plan ID non-optional, and document the intentional workflow migration prominently. That leaves one clear contract instead of four possible flag/config states.

fortmarek (commented):
One additional CircleCI migration finding.

cli/Sources/TuistServer/Services/GetShardService.swift
pepicrft: Fixed in a5c7679. `ShardService.shard()` now normalizes an empty `shardPlanId` to `nil` before using it for the reference fallback or passing it to `GetShardService`. Added a test (`shard_emptyPlanId_isTreatedAsAbsent`) that verifies the CI-derived reference is used when the env var is an empty string.

Files: `cli/Sources/TuistKit/Services/Sharding/ShardService.swift`, `cli/Tests/TuistKitTests/Services/Sharding/ShardServiceTests.swift`

cli/Sources/TuistKit/Services/Sharding/ShardMatrixOutputService.swift
pepicrft: Fixed. Codemagic's `CM_ENV` is workflow-scoped, so writing `TUIST_SHARD_PLAN_ID` there never reached the separate `test-shard-*` workflows. Removed the dead write from `writeCodemagicOutput` and documented that Codemagic falls back to reference-based shard lookup (the `CM_ENV` limitation is noted in both sharding guides).

Codemagic runs each workflow on a separate instance with no built-in cross-workflow variable sharing, so there is no simple mechanism to transport the dynamic plan ID between the build and test workflows.

Files: `ShardMatrixOutputService.swift`, both `test-sharding/*.md` guides

cli/Sources/TuistKit/Services/Sharding/ShardMatrixOutputService.swift
pepicrft: Fixed. The `.tuist-shard-matrix.json` file (which includes `shard_plan_id`) is still written to `BITRISE_DEPLOY_DIR` and uploaded as a build artifact via `deploy-to-bitrise-io`. Updated both docs to show the full consumption path in the test workflows: `pull-intermediate-files` downloads the artifact, then `jq` extracts the plan ID and `envman` exports it as `TUIST_SHARD_PLAN_ID` before the test command runs.

Files: both `test-sharding/*.md` guides

cli/Sources/TuistEnvKey/EnvKey.swift
pepicrft: Agreed. Removed `TUIST_CIRCLECI_SHARD_PLAN_ID_ENABLED`, made `writeCircleCIOutput` take a non-optional plan ID, and always include `shard-plan-id` in the continuation parameters. The docs now show the `shard-plan-id` parameter declared directly in the continued config without the flag. Users update their continuation config and Tuist version in the same change, leaving one clear contract.

Files: `EnvKey.swift`, `ShardMatrixOutputService.swift`, both `test-sharding/*.md` guides

PR #11969, opened 2026-07-21
fix(server): improve hosted runner dispatch resilience

pepicrft:
Related runner failure: <https://github.com/tuist/tuist/actions/runs/29766261140/job/88433129102>

Related Grafana alert: <https://tuist.grafana.net/alerting/grafana/afsd33dx5fhmoc/view?orgId=1>

## What changed

Runner registration now prefers the organization endpoint and falls back to the repository endpoint for personal GitHub accounts. Repository endpoint failures are scoped according to their impact: a missing administration permission excludes the account, a not-found response excludes only that repository, and transient failures exclude only the workflow job. Failed candidates are requeued without another ClickHouse read, and the same polling runner tries the next eligible candidate within a bounded retry loop.

Flaky-test rolling alert queries now run with two threads and a one-gigabyte memory limit. Alert evaluation also has a dedicated Oban queue with one worker per server pod.

## Why

macOS jobs were waiting even when runner hosts appeared idle. Production logs showed a personal account repeatedly reaching the organization runner endpoint, receiving a not-found response, and being selected again. This prevented the available macOS 26.5 runner from progressing to other eligible work.

At the same time, the Grafana alert showed memory-heavy ClickHouse aggregation competing with runner lifecycle queries.

## Root cause

Personal GitHub accounts cannot register organization-level runners, but the server only attempted the organization endpoint. When registration failed, dispatch requeued the candidate and returned, so the oldest broken candidate could monopolize every subsequent poll.

The alert queries merged large rolling aggregate states without query-level resource limits and shared the default background-job queue. This allowed several expensive evaluations to run concurrently on each server pod.

## Approach

The organization endpoint remains the preferred path because it uses the narrower organization runner permission. A not-found response triggers a repository-scoped fallback when the queued candidate contains a repository handle.

Dispatch exclusions are kept local to the bounded polling attempt. This lets another repository from the same account continue when one repository is unavailable, while preventing an unbounded loop across broken repositories. Account-wide permission failures still skip all work for that account during the attempt.

ClickHouse limits are applied only to the rolling alert queries, and queue isolation bounds aggregate concurrency without slowing unrelated background work.

## Impact

Personal accounts can receive repository-level runners after accepting the requested GitHub App permission. A broken repository or installation no longer blocks unrelated queued jobs from using the same runner. Existing organization registrations continue through the organization endpoint.

The alert queries have predictable processor and memory usage, leaving more ClickHouse headroom for runner state transitions. There is no schema migration or customer migration.

This does not change the current per-version macOS pool allocation, so idle capacity reserved by one operating-system pool is still not directly reusable by another pool.

### How to test locally

- `env MIX_ENV=test mise exec -- mix do deps.loadpaths --no-deps-check + test test/tuist/github/client_test.exs test/tuist/runners/jobs_test.exs test/tuist/runners_test.exs`: 139 tests passed.
- Focused alert-monitor tests passed as part of the broader server validation.
- Formatting and `git diff --check` passed.

### DIFF
--- a/server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
+++ b/server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
@@ -59,6 +59,12 @@ defmodule Tuist.Automations.Monitors.FlakyTestsMonitor do
   @default_rolling_window_size 100
   @recent_runs_bucket_sizes [100, 250, 500, 750]
 
+  # Merging the rolling aggregate states is memory-heavy and memory grows with
+  # parallelism. Keep this limit local to these queries so concurrent alert
+  # evaluations leave headroom for runner lifecycle writes and other reads.
+  @rolling_query_max_threads 2
+  @rolling_query_max_memory_bytes 1024 * 1024 * 1024
+
   def evaluate(alert, test_case_ids \\ nil) do
     trigger_config = alert.trigger_config
     threshold = trigger_config["threshold"] || 10
@@ -396,6 +402,8 @@ defmodule Tuist.Automations.Monitors.FlakyTestsMonitor do
       )
     )
     WHERE length(recent_runs) > 0
+    SETTINGS max_threads = #{@rolling_query_max_threads},
+             max_memory_usage = #{@rolling_query_max_memory_bytes}
     """
 
     %{rows: rows} =
@@ -525,6 +533,8 @@ defmodule Tuist.Automations.Monitors.FlakyTestsMonitor do
     )
     WHERE length(recent_runs) > 0
       AND #{rolling_having_expr(monitor_type)} #{rolling_comparison_op(comparison)} {threshold:Float64}
+    SETTINGS max_threads = #{@rolling_query_max_threads},
+             max_memory_usage = #{@rolling_query_max_memory_bytes}
     """
 
     params = maybe_put_test_case_ids(%{project_id: project_id, threshold: threshold * 1.0}, test_case_ids)
--- a/server/lib/tuist/github/client.ex
+++ b/server/lib/tuist/github/client.ex
@@ -16,6 +16,8 @@ defmodule Tuist.GitHub.Client do
   alias Tuist.VCS.Repositories.Content
   alias Tuist.VCS.Repositories.Tag
 
+  require Logger
+
   @doc """
   Lists repositories for a GitHub app installation with pagination support.
   Returns {:ok, %{meta: %{next_url: ...}, repositories: [...]}} format similar to Flop.
@@ -352,18 +354,17 @@ defmodule Tuist.GitHub.Client do
 
   @doc """
   Generates a just-in-time runner configuration for an ephemeral
-  GitHub Actions self-hosted runner registered at the org level.
+  GitHub Actions self-hosted runner.
   The returned `encoded_jit_config` is what the in-VM
   `./run.sh --jitconfig` consumes; runners registered this way
   are single-shot and auto-cleaned by GitHub after they exit.
 
-  Org-scoped (vs. repo-scoped) intentionally — the repo-scoped
-  endpoint requires the GH App to hold `administration: write`
-  on the repo, which grants access to settings, secrets,
-  collaborators, and many other unrelated capabilities. The
-  org-scoped endpoint requires only
-  `organization_self_hosted_runners: write`, a targeted scope
-  that does only what the name implies.
+  The organization-scoped endpoint remains the preferred path because it
+  only requires the targeted `organization_self_hosted_runners: write`
+  permission. Personal GitHub accounts cannot use that endpoint, so a 404
+  falls back to the repository-scoped endpoint when
+  `:repository_full_handle` is present. That endpoint requires the GitHub
+  App installation to have accepted `administration: write`.
 
   Runners registered at the org level are usable by any repo in
   the org subject to the runner-group's repo allowlist. The
@@ -371,10 +372,11 @@ defmodule Tuist.GitHub.Client do
   `attrs` to register the runner into a restricted group.
 
   See: https://docs.github.com/en/rest/actions/self-hosted-runners#create-configuration-for-a-just-in-time-runner-for-an-organization
+  See: https://docs.github.com/en/rest/actions/self-hosted-runners#create-configuration-for-a-just-in-time-runner-for-a-repository
   """
   def generate_jit_config(installation, org, attrs) do
     api_url = installation_api_url(installation)
-    url = "#{api_url}/orgs/#{org}/actions/runners/generate-jitconfig"
+    org_url = "#{api_url}/orgs/#{org}/actions/runners/generate-jitconfig"
 
     body = %{
       "name" => Map.fetch!(attrs, :name),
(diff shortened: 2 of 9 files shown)


server/lib/tuist/runners.ex
```diff
@@ -1012,21 +1084,42 @@ defmodule Tuist.Runners do
...
           account: account.name,
           org: github_org,
           runner: runner_name,
-          reason: inspect(reason)
+          reason: inspect(reason),
+          workflow_job_id: candidate.workflow_job_id
         )
 
-        {:error, :github_mint_failed}
+        {:error, {:github_mint_failed, :workflow_job}}
```
esnunes, line 1122: One case I wanted to flag on the retry-scope mapping: `mint_jit`'s catch-all (runners.ex:1114) scopes an org-endpoint 403 / transport error / 5xx to `:workflow_job`, which decrements the attempt budget. Because a requeued job keeps its `enqueued_at` and `pick_queued_top_k` orders `ASC`, an org that installed the app *without* `organization_self_hosted_runners: write` (or a GHES host that's down) and has ≥16 jobs at the head of the queue would burn the whole 16-attempt budget every poll on those jobs — and the first healthy co-tenant behind them never gets reached. Next poll re-attempts the same 16 from scratch, since the exclusion seed (`Claims.workflow_job_ids_for_fleet`) only holds still-claimed rows and we already released them. That looks like the monopolization the PR set out to prevent, shifted from 1 job to ≥16. Would it make sense to classify account-wide org failures (org 403 "not accessible by integration", transport/GHES-down, 5xx) as `:account`, so a broken installation is excluded once per poll instead of one job at a time?

server/lib/tuist/github/client.ex
```diff
@@ -383,8 +385,51 @@ defmodule Tuist.GitHub.Client do
...
+  end
+
+  defp generate_repository_jit_config(repository_full_handle, body, token, api_url) do
+    case String.split(repository_full_handle, "/", parts: 2) do
+      [owner, repository] when owner != "" and repository != "" ->
+        repository_url = "#{api_url}/repos/#{owner}/#{repository}/actions/runners/generate-jitconfig"
+
+        case post_jit_config(repository_url, body, token, api_url) do
+          {:error, {:http, 403, response_body}} ->
+            {:error, {:repository_administration_permission_required, 403, response_body}}
```
esnunes, line 417: Small robustness thing on the repo-endpoint error mapping: `generate_repository_jit_config` maps every repo 403 to `:repository_administration_permission_required` → whole-account exclusion + a `Logger.error` about the administration permission (client.ex:416, runners.ex:1094). GitHub also returns 403 for secondary/primary rate limits, and the shared retry only backs off 429/5xx (retry.ex:26), so a rate-limited 403 gets read as a permanent permission problem — the account is skipped for the tick and (with the requeue-at-head behavior) the throttled installation keeps getting hit. Could we peek at the body/headers (a "rate limit" message, `Retry-After`, `X-RateLimit-Remaining: 0`) and treat those as transient instead of escalating to `:account`? When it genuinely is a permission 403, account-scope is right — it's just the rate-limit case that gets over-scoped.

server/lib/tuist/runners.ex
```diff
@@ -494,97 +500,121 @@ defmodule Tuist.Runners do
...
+  defp handle_claim_attempt({:error, reason}, %{fleet_name: fleet_name}) do
+    Logger.warning("runners: dispatch_for_sa failed",
+      reason: inspect(reason),
+      fleet: fleet_name
+    )
+
+    {:error, reason}
+  end
+
+  defp handle_serve_claim({:error, {:github_mint_failed, exclusion_scope}}, context)
```
esnunes, line 591: Heads-up on a cost/observability angle (not a correctness bug): there's no backoff for a persistently-failing mint — the requeue keeps `enqueued_at`, so the job returns to the queue head and every idle Pod re-probes it each poll, and `:account`-scope retries don't decrement the budget (runners.ex:667). Two side effects I'd watch: `record_claimed` fires `:job_claim` with `outcome: "ok"` and a `queue_time_ms` that grows every poll for a job that never actually dispatches (jobs.ex:396), so one stuck installation skews claim-throughput and queue-time percentiles; and the `claimed`→`queued` broadcasts flap the account's LiveView. A short per-job/installation cooldown (a "not before" stamp, or excluding a mint-failed account for a few polls) would bound the load and the metric skew.

server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
```diff
@@ -396,6 +402,8 @@ defmodule Tuist.Automations.Monitors.FlakyTestsMonitor do
       )
     )
     WHERE length(recent_runs) > 0
+    SETTINGS max_threads = #{@rolling_query_max_threads},
```
esnunes, line 405: Worth double-checking the sizing here: these two rolling queries now pin `max_memory_usage = 1 GiB` (flaky_tests_monitor.ex:405/536), which is 6× below the read repo's 6 GiB default (environment.ex:1148), and there's no `max_bytes_before_external_group_by`, so a query that genuinely needs >1 GiB hard-fails (MEMORY_LIMIT_EXCEEDED) rather than spilling. Since `query!` raises and the memory need is deterministic, all 3 Oban attempts would fail and the alert would silently stop evaluating for that project. Is 1 GiB comfortably above the observed peak for the largest projects (a 750/1000-run window over tens of thousands of active test cases)? If any project sits in the 1–6 GiB band it would regress from slow-but-working to never-evaluating — enabling spill, or catching MEMORY_LIMIT_EXCEEDED and marking the project unevaluable with a metric, would make that failure mode explicit. (Also: this cap is per-query-per-server, so it bounds one pod but not the cluster — `max_memory_usage_for_user` would if a global bound was the goal.)

server/lib/tuist/github/client.ex
```diff
@@ -383,8 +385,51 @@ defmodule Tuist.GitHub.Client do
...
+          response ->
+            response
+        end
+
+      _ ->
+        {:error, {:invalid_repository_full_handle, repository_full_handle}}
+    end
+  end
+
+  defp post_jit_config(url, body, token, api_url) do
```
esnunes, line 431: One resilience angle for a GitHub brownout: `post_jit_config` sets no explicit `receive_timeout` (client.ex:431), and this PR turns "one mint per poll" into up to 16 (`@max_claim_attempts_per_dispatch`), each of which can make two GitHub calls (token + JIT) over the shared `https://api.github.com` Finch pool (20 connections, application.ex:455) that also serves token minting and webhook verification. If GitHub slows down, a single poll could serialize many slow calls and contend that pool, so the slowdown could spread to token/webhook handling. A tight explicit `receive_timeout` plus a per-poll mint-time budget (separate from the 16-candidate cap) would contain it; isolating JIT minting on its own pool is the bigger-hammer option. Not a blocker — flagging the failure mode.

esnunes (approved):
Really thorough PR — the write-up, the scoped-exclusion design, and the test coverage for the single-broken-candidate paths are all excellent. I verified a few things that hold up well: the candidate-map requeue writes a complete 24-column row (no field loss), the release-before-retry ordering is race-safe (the Postgres claim is held across the requeue and gated on `completion_recorded?`), and the Oban queue move doesn't orphan in-flight jobs. Since it's already merged, everything below is follow-up rather than a merge blocker.

The theme across the top findings is that the retry loop's *exclusion scope is derived from the HTTP status + which endpoint returned it, not from the actual failure cause* — and two cases fall on the wrong side of that. An account-wide org failure (missing `organization_self_hosted_runners: write`, GHES-down, 5xx) is scoped per-job, so ≥16 such head-of-queue jobs can still monopolize a fleet (the very thing the PR targets, shifted from 1 job to 16); and a repo-endpoint rate-limit 403 is scoped to the whole account as if it were a permission denial. Below those: no cross-poll backoff means broken installations get re-probed every poll (with some telemetry/UI side effects), the 1 GiB ClickHouse cap can hard-fail large-project alerting rather than spill, and the mint call has no explicit timeout now that it runs up to 16× per poll. Details inline.

PR #11973, opened 2026-07-21
feat(infra): make runner dispatch starvation distinguishable from saturation

pepicrft:
## What

Adds three series to the runners-controller so a starved runner pool is distinguishable from a saturated one:

- `tuist_runners_autoscaler_claimed_jobs{pool}` and `tuist_runners_autoscaler_queued_jobs{pool}`, the server's two demand signals published unsummed
- `tuist_runners_pool_idle_replicas{pool}`, alive current-image Pods with no `tuist.dev/runner-pool-owner`, meaning warm capacity able to take work right now

Plus a dashboard panel for the resulting signature and an AGENTS.md section explaining how to read it.

No behavior change. Metrics only.

## Why

Today's macOS runner outage is the motivation. Dispatch was starved by a broken repository (fixed server side in #11969), and for over an hour the fleet looked like this: GitHub jobs queued, warm macOS Pods `Running` and polling, nothing being claimed.

While debugging it I could not answer "is this capacity or is this dispatch" from any exported metric, and that ambiguity is what made the outage take as long as it did to characterize. Working through the existing series, each one conflates the two states:

- `tuist_runners_pool_phase_replicas` counts a warm idle Pod and a Pod running a customer job identically. Both are `Running`.
- The autoscaler consumes `claimed + queued`, and `RecordAllocation` receives that sum but only uses it to derive the warm deficit. The sum itself is never exported, and it stays flat while work drains normally, because a job moving from queued to claimed does not change it.
- `tuist_runners_pool_oldest_pending_pod_age_seconds` only sees Pods that never booted. During this outage the Pods had booted fine. They just never received work.
- `Oldest queued job age by fleet` and its `Runner queue age` alert do fire here, but they fire identically whether the pool is starved or simply out of hosts. They tell you something is wrong without telling you which lever to pull.

The invariant worth alerting on is that queued work and idle warm capacity should never coexist for long. An idle Pod polls dispatch continuously, so queued work should reach it within a poll interval. Sustained overlap means dispatch is not serving the pool despite free capacity, and adding hosts will not help.

That gives two cleanly separated states:

| State | queued | idle | Fix |
|---|---|---|---|
| Saturated | > 0 | 0 | Capacity |
| Starved | > 0 | > 0 | Server side |

## Approaches considered

**Fix the reconciler's replica accounting instead.** My first read of the incident was that the controller counts unschedulable `Pending` Pods as alive and idle, inflating perceived capacity. Reading the code, that is wrong and I dropped it. `alive` has to count `Pending`, otherwise the gap-fill creates a fresh Pod every tick for one that is merely waiting on a node, and the pool runs away. The count is correct as written.

**Make pool sizing capacity aware.** Also already done. `AllocateFleet` splits a shared budget in three priority tiers and reclaims other pools' speculative warm buffers to admit real queued work. Tier one deliberately grants load in full past capacity so the excess goes `Pending` as the operator's "add a host" signal. That is intentional design, not a bug.

**Export a single composite starvation gauge.** Rejected. It would bake a threshold and a policy decision into the controller. Exporting the two primitives lets the dashboard and any future alert own that, and the primitives are independently useful.

**Add an alert rule.** Out of scope here. There is no alerting as code anywhere in the repo, only dashboards, so introducing that pattern is a separate decision. The panel description spells out the alert expression for whoever wires it up in Grafana.

So the change is deliberately narrow: add the missing primitives, do not touch dispatch or scheduling semantics.

## Implementation notes

`idleCount` is settled before the scale-down loop rather than after. That loop can return early on a failed delete, and defaulting the gauge to 0 on that path would read as "no warm capacity available" and mask the exact signal this adds. It is decremented per successful delete and published from the existing `defer`, so every return path reports the count that survived the reconcile.

`RecordDemand` is called from `desiredForPool`, which runs on every autoscaler tick including when the fleet allocator falls back to the per-pool target, so the demand signals stay live while allocation is degraded.

## Validation

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean
- `go test ./...` passes across all packages
- Confirmed the new reconciler test is not vacuous: removing the `RecordIdleReplicas` call makes `TestReconcilePublishesIdleReplicas` fail with `idle replicas = 0, want 2`, and it passes again once restored
- Tests cover the claimed Pod not counting as idle capacity, the gauge draining to 0 on a fully claimed pool rather than holding a stale sample, negative clamping, and series cleanup on pool deletion
- Dashboard JSON validated by round-trip parse, inserted as a pure addition with 61 added lines and 0 deletions so no unrelated formatting churn

Verified against production while diagnosing: the pool that eventually picked up the stuck job reported 8 idle Pods with `queued: 8` held flat for 25 minutes across autoscaler ticks, which is precisely the overlap this panel would have surfaced.

One thing worth a reviewer's eye: the panel's PromQL uses `label_replace` plus `or` plus `min by (pool)` to take a per pool minimum across two metrics. It is a standard construction but I could not render it against Grafana from here, so it is worth confirming on the dashboard after deploy.


pepicrft:
Eduardo's Slack observation caught a real bug in this PR, pushed a fix in d7b9cf2.

He pointed at two pods that had been alive for over 3h:

```
macos-26-6-runner-e58803e6   1/1   Running   3h1m    ...gjf8p
macos-26-6-runner-dc051555   0/1   Pending   3h15m   <none>
```

The Pending one has no node and no VM. It is unclaimed and unowned, so `isIdle` returns true for it, which means my new `pool_idle_replicas` gauge was counting it as "warm capacity able to take work right now".

That inverts the exact signal the PR is for. During today's incident there were 10 such pods on the macOS fleet. A pool starved of Mac minis would have reported ~10 idle pods sitting on queued work, which is the fingerprint of dispatch failing, so the panel would have pointed at the wrong cause with high confidence. Worse than not having it.

Fixed by gating on OS, the same asymmetry `oldestPendingPodAge` already encodes:

- darwin: only `Running` counts, since Pending means the VM is not up
- linux: `Pending` counts, since that is where a warm dispatch poller spends its whole idle life (the poller is an init container, so kubelet holds the pod Pending for its entire idle life)

Two tests added for both directions. Confirmed non-vacuous: with the guard removed the darwin test reports 3 instead of 1, which is precisely the bug.

On Eduardo's actual question, whether those pods should terminate: as far as I can tell from the code, no, and nothing will terminate either of them.

- `IdleTimeoutSeconds` does not apply. Its doc comment is explicit that it only bounds the post-claim window between a runner registering and being handed work. A pod still polling for a claim has no runner container yet and is unaffected.
- The Running unclaimed one is warm pool by design. It is only reaped on scale-down (overflow) or an image roll.
- The Pending one is the `AllocateFleet` "excess goes Pending as the operator's add-a-host signal" case. It counts toward `alive`, so `gap` is 0 and no replacement is made, which is correct here since there is no capacity for one anyway.

So the 3h15m Pending pod is tolerated rather than intended. It should be showing bright red on `Oldest un-booted Pod age by pool`, whose threshold is 30m.

infra/runners-controller/controllers/runnerpool_controller.go
```diff
@@ -595,6 +616,28 @@ func isIdle(pod *corev1.Pod) bool {
...
+// failure.
+//
+// The test is OS-dependent for the same reason oldestPendingPodAge is
+// darwin-only. On a Tart pool, Pending means the VM isn't up, so only
+// Running is real capacity. On a Linux pool, Pending is the healthy
+// steady state — the dispatch poller is an init container and kubelet
+// holds the Pod in Pending for as long as it runs — so a warm Linux
+// runner is Pending for its whole idle life and must still count.
+func isWarmCapacity(pod *corev1.Pod, pool *tuistv1.RunnerPool) bool {
+	if pool.Spec.OS != "darwin" {
```
fortmarek: [P1] Require an actively running Linux poller

`Pending` on Linux is not synonymous with warm polling. An unscheduled Pod, or one still waiting to start its init container, is also `Pending`, has no owner label, and `pollerTerminated` returns false when no poller status exists. Since this branch returns true for every non-darwin Pod, the capacity-exhausted Linux case still publishes `idle > 0` and is classified as starvation. Check that the `poller` init status is actually `Running` (or otherwise prove it is actively polling). The added Linux test currently creates Pending Pods with no init status, so it models the unscheduled case while expecting it to count.

infra/runners-controller/controllers/autoscaler_controller.go
```diff
@@ -230,6 +230,12 @@ func (r *AutoscalerReconciler) desiredForPool(
 	// on its own series, not inferred from alive-vs-desired.
 	metrics.RecordAllocation(pool.Name, signals.Claimed+signals.Queued, knobs.MinWarmPoolFloor, perPool, allocated)
 
+	// Publish the demand signals unsummed as well. The allocator only
+	// needs claimed+queued, but that sum is flat while work drains
+	// normally (queued -> claimed), so it cannot distinguish a pool
+	// serving its backlog from one wedged with the same backlog.
+	metrics.RecordDemand(pool.Name, signals.Claimed, signals.Queued)
```
fortmarek, line 237: [P1] Separate concurrency-blocked queue depth from dispatchable work

`signals.Queued` comes from `Jobs.queued_count_by_fleet/1`, which includes jobs that remain queued after `Claims.attempt/5` rejects them because the account reached its platform concurrency limit. Dispatch excludes that account only for the current poll; the job stays queued. Therefore `queued > 0 && idle > 0` is a valid steady state when all queued accounts are capped, not evidence that dispatch is broken. Fleet-level `claimed` cannot resolve this because the caps are account- and platform-wide and may be consumed by claims in another pool. Export dispatchable versus concurrency-blocked queue depth, or include another admission-control signal, before treating this overlap as starvation.

infra/grafana-dashboards/runners.json
```diff
@@ -1559,6 +1559,67 @@
...
+          "tooltip": {
+            "mode": "multi",
+            "sort": "desc"
+          }
+        },
+        "targets": [
+          {
+            "datasource": "$datasource",
+            "editorMode": "code",
+            "expr": "min by (pool) (\n  label_replace(max by (pool) (tuist_runners_autoscaler_queued_jobs{pool=~\".*$platform.*\", env=\"$env\"}), \"signal\", \"queued\", \"\", \"\")\n  or\n  label_replace(max by (pool) (tuist_runners_pool_idle_replicas{pool=~\".*$platform.*\", env=\"$env\"}), \"signal\", \"idle\", \"\", \"\")\n)",
```
fortmarek: [P2] Require both signals before taking the minimum

The `or` union retains the lone series when the other operand is absent, and `min by (pool)` then returns that value. Static or autoscaling-disabled pools deliberately clear the demand series while the RunnerPool reconciler continues publishing idle capacity, so `idle = 5` with no queued series charts as 5 starved replicas. Gate the result on the presence of both operands or coalesce a missing operand to zero.

infra/runners-controller/controllers/runnerpool_controller.go
```diff
@@ -168,8 +168,13 @@ func (r *RunnerPoolReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			phaseReplicas.add(p)
 		}
 	}
+	// Settled below, once scale-down has run; the deferred publish reads
+	// the final value so every return path reports the idle count that
+	// actually survived this reconcile.
+	idleCount := 0
 	defer func() {
 		metrics.RecordPodPhases(pool.Name, phaseReplicas.pending, phaseReplicas.running, phaseReplicas.unknown)
+		metrics.RecordIdleReplicas(pool.Name, idleCount)
```
fortmarek, line 186: [P2] Do not publish zero before the count is initialized

This defer starts with `idleCount = 0`, while `idleWarm` is assigned only after the terminal-Pod reap loop. If reaping a terminal Pod fails, reconcile returns early and publishes 0 even when warm Pods were discovered, masking the signal on an error path the surrounding comments intend to preserve. Compute the warm count before any early-return mutations, or skip publication until it has been initialized.

fortmarek (commented):
The macOS Pending-Pod correction in d7b9cf2 addresses the reported case. I found four remaining edge cases that keep `queued > 0 && idle > 0` from being a reliable starvation fingerprint; details inline.

pepicrft:
All four addressed in 740383e. Thanks, the first one is a real inversion and the test point was exactly right.

**P1, Linux poller.** You're correct that I only fixed half the bug. `isWarmCapacity` returned true for every non-darwin Pod, so an unscheduled Linux Pod counted as available capacity, and a Linux pool merely out of hosts read as starved. Same inversion I'd just fixed for darwin, still live on the other platform.

Now gated on the `poller` init container being actively `Running`, which is only true once the Pod has a node and kubelet has started it.

And you were right that my test asserted the wrong behaviour: it built Pending Pods with no init status, so it modelled the unscheduled case while expecting it to count. Confirmed by applying the fix first, which made that test fail. Replaced with three that separate the cases: running poller counts, unscheduled does not, still-waiting does not. Mutation-checked by reverting to `return true`, which fails the latter two.

**P1, concurrency-blocked queue depth.** Agreed, and this is the one I'd hold the PR on. I hit the same thing in production a few hours after opening this: `macos-26-5` sat at `claimed: 0, queued: 8` flat for 25 minutes with idle Pods in the same pool, which is exactly your valid-steady-state case, not a fault.

The fix is server-side in #11981, which caps each account's contribution at its remaining headroom so `queued` becomes dispatchable-only. So there's a real ordering dependency: **#11981 has to land before this**, or the panel is wrong by construction. I've documented that caveat in the panel description and AGENTS.md rather than leaving it implicit.

Worth noting it still won't be perfect afterwards: a repository whose token mint fails also leaves jobs queued that dispatch skips, and `excluded_repositories` is per-call in-memory state with nothing to join against, so that slice needs durable state before it can be excluded too.

**P2, PromQL.** Right. `or` keeps the lone series and `min` charts it alone, so a pool that clears its demand series while still publishing idle capacity reads as starved. Now gated with `and on (pool)` against both operands.

**P2, uninitialised publish.** Also right, and I'd reasoned about this exact hazard for the scale-down loop and missed the reap path one level up. The warm count is now computed in the earlier pass that has no early returns, so it is always initialised before the defer is registered.

Full suite green, gofmt/vet clean.

fortmarek (approved):

infra/runners-controller/controllers/runnerpool_controller.go
pepicrft: Fixed in 740383e. You were right that the Linux branch was the same inversion I had just fixed for darwin. `isWarmCapacity` now asks whether the `poller` init container is actively `Running`, which is only true once the Pod has a node and kubelet has started it. The old Linux test modelled the unscheduled case while asserting it counts, so it is gone, replaced by three that separate running, unscheduled, and still-waiting pollers.

infra/runners-controller/controllers/autoscaler_controller.go
pepicrft: Agreed, and this is the finding that changed the merge order. The overlap is only a fault once `queued` counts dispatchable work, so I split the actual fix into #11981, which caps each account contribution at its remaining concurrency headroom, and landed that first. It is merged now. This PR carries the caveat in the panel description and AGENTS.md, updated in 7af0f1c to name #11981 as the reason the fingerprint holds rather than as a condition still pending.

infra/grafana-dashboards/runners.json
pepicrft: Fixed in 740383e. The expression is now gated with `and on (pool)` against both operands, so a pool that clears its demand series while idle capacity keeps publishing drops out instead of charting its idle count as starvation.

infra/runners-controller/controllers/runnerpool_controller.go
pepicrft: Fixed in 740383e. The warm count is computed in the first pass over the Pod list, the one with no early returns, so every return path publishes a real value. The reap loop can no longer bail before the counter is assigned.

PR #11977, opened 2026-07-21
feat(server): project test case state from the event ledger

pepicrft:
First of three pull requests moving `state` and `is_flaky` off `test_cases`. **This one is deliberately behaviour neutral** and exists so the release that follows can ship without a window where a state change can be lost.

The three have to merge and deploy in order, one deploy each.

## Why

We had a report of an automation muting a test case where the mute did not stick: the dashboard showed `muted` and the history recorded the event, yet the stored state was still `enabled` and the test kept running.

`test_cases` is a ClickHouse `ReplacingMergeTree(inserted_at)` shared by two very different writers. Test report ingestion owns the run derived columns and carries `state` and `is_flaky` forward. The control plane, a user action or an automation's `change_state`, owns those two columns.

Ingestion reads every test case's current state once at the start of a report but stamps `inserted_at` later, per module. A mute landing inside that window is written with a *lower* version than the ingestion row carrying the pre-mute value forward, so `FINAL` picks `enabled` and the mute silently disappears. It does not recover on its own either: the automation skips test cases that already have a `triggered` alert event, so it never retries.

## What changed

This adds `test_case_states` and a materialized view that maintains it from `test_case_events`.

The ledger is written by the control plane on the same path that sets the state, and ingestion never touches it, so it is the one copy the race could not corrupt. Deriving the state from it removes the second writer entirely rather than keeping two writers in agreement, which is a weaker property: an earlier iteration of this change wrote the state and the event separately, two unsynchronised inserts that could drift.

### One column per row

Each row records only the column its event affected and leaves the other NULL, and reads resolve each column with its own `argMaxIf(..., isNotNull(...))`.

This is the crux. Sharing a row is what the original bug was: writing one column carried the other along, so a `marked_flaky` event would overwrite a concurrent mute. Collapsing to one row per test case (a `ReplacingMergeTree` keyed on the test case) would reintroduce exactly that. Keeping the rows separate makes it impossible by construction rather than something we have to order correctly.

### `project_id` on `test_case_events`

Every read of the projection is project scoped so it can ride the `(project_id, test_case_id)` sort prefix, and a materialized view only sees the rows of the insert that triggered it, so it cannot join back to `test_cases` for the answer. The column has to be on the event row.

Backfilled in place with a dictionary lookup, the same shape as `20260609120000_denormalize_project_id_on_cas_outputs`, and cheap here: the table is ~1.6M rows and ~60 MiB. It takes a Postgres advisory lock, because unlike that precedent this one drops the dictionary afterwards, and a pod dropping it while another is mid mutation would fail that pod's migration and crash it on boot.

## Approach

Deriving the state from `test_case_events` on every read was considered and rejected on cost. Measured on the busiest project: 115ms reading 1.96M rows from the event table, against 33ms and 330k rows from a dedicated projection. It also scales with event history, which grows forever, rather than with test case count, and the quarantined list is on the hot path for every CI run through the command line interface and Gradle plugin.

The projection stays small because only state and flaky events land in it: ~129k rows against ~2.2M test cases.

## Rollout

Creating a materialized view does not replay history, and that is load bearing here rather than incidental. Every event already in the table stays out of the projection, and the release that flips reads over seeds the starting values explicitly. If creation did replay history it would restore ~600 mutes from February and March that teams have been running without for five months.

Reads and the ingestion carry-forward are unchanged, so nothing observable moves in this release.

## Validation

- All three migrations apply against a real ClickHouse from the existing schema, including the dictionary backed mutation.
- Verified the projection end to end on a scratch database with the real table definitions: a mute resolves to `muted`, an unmute back to `enabled`, and, the case that matters most, a flaky event landing after a mute leaves the mute intact. A design that wrote both columns per row fails that last one.
- Confirmed explicitly that creating the view does not backfill, by inserting historical events first and checking the projection stayed empty. The decision to leave the February and March mutes alone depends on it.
- `mix test test/tuist/tests_test.exs`: 239 passing. Also ran the analytics, automations, and test case API suites: 116 passing.
- `mix credo` reports only the pre-existing "Raw sql executed" warning that every ClickHouse migration trips.

### How to test locally

```sh
cd server
mix test test/tuist/tests_test.exs
```

The behaviour is pinned by `a flaky event does not clobber the state set by an earlier mute`.


server/lib/tuist/tests.ex
```diff
@@ -963,6 +964,8 @@ defmodule Tuist.Tests do
 
       updated_test_case = Map.merge(test_case, filtered_attrs)
 
+      write_test_case_state(updated_test_case)
```
fortmarek: **[P1] Keep events as the control-plane source of truth**

Could we avoid introducing another application-written state snapshot here? `update_test_case/3` already records the corresponding transitions in `test_case_events`, and ingestion never writes that log. This new `ReplacingMergeTree` also retains a lost-update case: concurrent callers can both read `{enabled, false}`, then write `{muted, false}` and `{enabled, true}`; the later row replaces the whole earlier row. Resolving `state` and `is_flaky` independently from their latest events avoids both the ingestion race and this control-plane race. We should first audit/backfill missing historical events and benchmark project-listing queries, but if those pass, events should be authoritative and any compact current-state table should be a derived projection rather than a second dual-written source of truth.

fortmarek (commented):
One architectural concern:

server/lib/tuist/tests.ex
pepicrft: Good catch on the lost update, it's real. Reproduced it against the new table:

```
writer A reads {enabled,false}, mutes  -> writes {muted,   false}
writer B reads {enabled,false}, flags  -> writes {enabled, true}
resolved: state="enabled" is_flaky=true
```

The mute is gone. Each row carries both columns, so the later row wins for both even though `argMax` resolves them independently. Worth saying this race already exists on `test_cases` today for the same reason, but you're right that I'm carrying it forward instead of fixing it.

On events being authoritative, I think you're more right than the comment claims. Look at `20260211150000_backfill_missing_unquarantined_events`: PR #9397 was this same class of bug, and the interesting part is the note that ingestion reset `is_quarantined` "without creating an unquarantined event". So when ingestion clobbers state, the events log does not follow it. Which means right now, for every test case this bug has silently unmuted, `test_case_events` still says `muted` while `test_cases.state` says `enabled`.

That has a consequence for this PR I had not spotted. My backfill seeds from `test_cases FINAL WHERE state != 'enabled'`, so it misses exactly the test cases the bug already broke. An events-sourced backfill would recover them.

Two things I want to measure before making events authoritative for reads, both of which you already called out:

1. `test_case_events` has no `project_id` and is ordered by `(test_case_id, event_type, id)`. Resolving state for the test cases listing is project scoped, so it becomes a filter over a large id set rather than a sort key prefix scan. That's fine for the flaky page, which already does this in `currently_flaky_test_case_ids_subquery`, but the test cases listing is the hot one. For scale, from production: the busiest project has 296k rows in `test_cases` and the current listing returns in about 120ms.
2. History completeness. There are already two synthetic backfills in the log (`backfill_first_run_events` and `backfill_missing_unquarantined_events`), so it is partly a reconstruction rather than a pure record of intent. I want to see how far the events derived state and the snapshot actually diverge before trusting it.

I'll run both and post the numbers here. If they hold up I'm happy to make events the source of truth and reduce this table to a derived projection, or drop it entirely.

One product question that falls out either way, and I don't want it decided as a migration side effect: sourcing state from events means test cases that were silently unmuted months ago start being muted again. That is correct per the user's original intent, but it will surprise people whose suites have been green because of the bug. We should choose that deliberately rather than have it ship inside a backfill.

pepicrft: Ran both against production. The audit backs you up hard, the benchmark does not, and I think together they point at a third shape.

**Audit.** Cross tab of `test_cases FINAL.state` against the state implied by each test case's last `muted`/`unmuted`/`skipped`/`unskipped` event, over 1,267,484 test cases:

| snapshot | last event | count |
|---|---|---|
| enabled | none | 1,255,987 |
| enabled | unmuted | 5,078 |
| muted | muted | 4,855 |
| enabled | unskipped | 715 |
| **enabled** | **muted** | **626** |
| skipped | skipped | 200 |
| muted | unmuted | 12 |
| **enabled** | **skipped** | **8** |
| skipped / muted / other | mixed | 3 |

The bolded rows are the bug. 634 test cases were muted or skipped, ingestion silently reverted them, and no `unmuted` event was ever written. Events say 5,691 test cases should be quarantined right now, the snapshot says 5,070, so roughly **11% of all intended quarantines are silently not in effect**. Same story for flakiness: 465 test cases where the snapshot says not flaky but the last event is `marked_flaky`.

So you are right that events are the only representation ingestion has never corrupted, and my backfill sourcing from `test_cases` would have permanently discarded those 634. That is the strongest argument in the thread and I had it backwards in the PR description when I said the lost mutes cannot self correct.

**Benchmark.** Test cases listing for the busiest project (296k test cases), measured through `system.query_log`:

| shape | duration | rows read | memory |
|---|---|---|---|
| current (no join) | 25 ms | 196,212 | 85.8 MiB |
| `test_case_states` join | 33 ms | 329,926 | 71.8 MiB |
| events derived | 115 ms | 1,962,615 | 110.2 MiB |

Events derived is 4.6x slower and reads 10x the rows, because `test_case_events` has no `project_id` and is sorted by `(test_case_id, event_type, id)`, so project scoping degrades into a filter over a large id set instead of a prefix scan. That is the hot listing, so I do not want to put it there.

**Where that leaves me.** I think your last sentence is the answer, and the numbers say which half goes where. Events stay authoritative, the compact table becomes a derived projection rather than a second source of truth:

1. Backfill the projection from the events log, not from `test_cases`. That is what recovers the 634.
2. Keep reads on the projection, so the listing stays at 33 ms rather than 115 ms.
3. Derive each column in the projection from its own event stream instead of from a snapshot read. That also kills the lost update you spotted, because two concurrent writers emit two independent events and neither overwrites the other's column.

That is a real change to these PRs rather than a tweak, so I would rather agree the shape with you before I rewrite them. Does the split above match what you had in mind?

One thing I do not want to decide unilaterally: point 1 means test cases silently unmuted months ago start being muted again on deploy. Correct per the user's original intent, but 634 tests going quiet at once will surprise people whose suites have been green because of this bug. Options are to restore them all, restore only recent ones, or notify the affected projects first. I lean toward restoring all of them and telling the affected accounts, but that is a product call.

PR #11984, opened 2026-07-21
fix(server): stop ClickHouse insert failures from killing shared ingestion buffers

pepicrft:
Two independent production issues found while investigating test runs that were sitting on `processing` for hours. Neither turned out to be the cause of that stall (that was a capacity shortfall in the xcresult processor queue, tracked separately), but both are real and one of them is why the investigation took as long as it did.

## What changed

**Ingestion buffer survives ClickHouse insert failures.** `Tuist.Ingestion.Buffer` is one named process per destination table, shared by every producer on the node. `do_flush` called `IngestRepo.query!`, so any ClickHouse error raised inside `handle_call/3` and killed the process, taking every unrelated caller blocked on it down as well. It now uses the non-bang `query/3` and hands the failure back to the caller. `flush/2` re-raises, so the requesting job still fails and retries exactly as before; the difference is that everyone else is untouched.

**Bounded waits.** Both `insert/3` and `flush/2` called with `:infinity`. A ClickHouse that was slow rather than failing could park every producer on a node indefinitely, producing no crash, no exception and no telemetry. They now use a configurable timeout (`clickhouse.flush_timeout_ms`, default 60s).

**Redundant `GROUP BY` removed from `test_case_runs_validated_on_branch_mv`.** The view aggregated on every insert block into `test_case_runs`.

**Per-query write memory ceiling.** `Tuist.IngestRepo` now sets `max_memory_usage` (`clickhouse.write_max_memory_usage_bytes`, default 2 GiB).

**`Logger.error` now reaches Sentry.** `Sentry.LoggerHandler` was started without `capture_log_messages`, which defaults to false, so the handler only ever reported crashes.

## Why

The buffer crash is Sentry issue TUIST-37Z: 3643 occurrences, marked escalating. Each one killed a shared process and failed an unknown number of concurrent, unrelated inserts alongside the one that actually went wrong.

The Sentry gap is more insidious. Two paths in `TestsController` report through `Logger.error`: one when an xcresult job cannot be enqueued, the other when the run cannot then be marked as failed. Both leave a test run stuck on `processing` forever, and neither produced a single Sentry event. During the investigation, the absence of errors was reasonably read as evidence that nothing was failing, which sent things down the wrong path for a while. Any silent failure mode that leaves user-visible state wedged should be visible.

## Root cause

The buffer crashes trace to this ClickHouse error:

```
Code: 241. DB::Exception: (total) memory limit exceeded: would use 18.00 GiB,
maximum: 18.00 GiB. OvercommitTracker decision: Query was selected to stop by
OvercommitTracker: While executing AggregatingTransform: while pushing to view
default.test_case_runs_validated_on_branch_mv
```

`test_case_runs_validated_on_branch_mv` was defined with `GROUP BY project_id, git_branch, test_case_id`, which makes ClickHouse build an `AggregatingTransform` over every insert block pushed into `test_case_runs`, synchronously, on the write path, for every producer.

Two things then made a bad situation worse. The ceiling that was hit is `(total)`, the server-wide limit, not a per-query one, so OvercommitTracker killed whichever query it happened to select, including inserts with no relationship to this view. And `ClickHouseRetry.with_retry` only rescues `Mint.TransportError` and `DBConnection.ConnectionError`, so a `Ch.Error` propagated straight out of `query!` and through the `GenServer` callback.

## Approach

Dropping the `GROUP BY` is safe and was the cheapest correct fix. The view's target table is a `ReplacingMergeTree` ordered by exactly those three columns, so duplicates collapse on merge, and the read path in `Tests.test_case_ids_with_successful_default_branch_run/3` already applies `distinct` to tolerate rows that have not merged yet. The aggregation was never load-bearing; removing it turns a per-insert aggregation into a plain projection.

The write-path memory ceiling mirrors what the read path has had since it was introduced. The comment at `config/runtime.exs:289` already claimed that "writes and backfills go through IngestRepo, which sets its own per-query limits where needed", but no such limit existed. With one, an oversized insert fails on its own as a retryable `(for query)` error rather than walking the whole server up to its global ceiling.

For the buffer, the alternative was to keep raising and accept the crash. That was rejected because the blast radius is the problem, not the failure itself: one bad insert should fail one job. The flush now returns a tagged tuple, and the public functions re-raise so no call site had to change. Flushes with no caller to answer to (the periodic tick, a full buffer, shutdown) log and drop the batch, which is what the previous raise already did in practice once the supervisor restarted with an empty buffer, minus the collateral damage.

`ClickHouseRetry.with_retry_result/2` was added rather than reworking `with_retry/2`, because the existing function can only observe failures that are raised and is used from many other call sites that depend on that.

Sentry ships with `rate_limiting: [max_events: 300, interval: 1 minute]`. Promoting every error log to an event is exactly the kind of change that behaves fine until an incident, and a saturated database connection pool already produces tens of thousands of events a week unaided. Dropping past a cap is a better failure mode than amplifying an outage into a Sentry outage.

## Impact

No API or user-facing behaviour changes. A failed flush fails the same job it would have failed before, so retry semantics are unchanged.

The migration replaces a view only. `test_case_runs_validated_on_branch` keeps all of its data. Between the `DROP` and the `CREATE` a concurrent insert would not produce marker rows; a missing marker only makes the flaky-test quarantine check treat a case as unvalidated, which is the conservative direction, and the marker reappears the next time that case runs on the default branch.

Sentry event volume will rise, since error-level logs that were previously discarded now become events. The rate limit bounds the worst case.

The `mix.lock` bump is isolated in its own commit so it can be dropped independently. `mix deps.get` resolves `elixir_make` and `lazy_html` above the pinned versions because the declared ranges allow it. This is pre-existing drift, not an intentional upgrade, but the tree will not compile while the lockfile and the fetched dependencies disagree.

## Validation

- `mix compile --warnings-as-errors` clean
- `mix test test/tuist/ingestion/ test/tuist/clickhouse_retry_test.exs test/tuist/tests/`: 273 passed
- `mix test test/tuist_web/controllers/api/tests_controller_test.exs`: 24 passed
- `mix credo --strict` on the touched files: no new findings (the single reported issue is pre-existing at `environment.ex:579`)
- `mise run format` applied

New tests cover the behaviour that regressed: that a failing flush raises to the caller while leaving the buffer alive, that the buffer keeps serving callers afterwards, that a failing periodic flush drops the batch without crashing, that a stalled flush gives up instead of waiting forever, and that `with_retry_result/2` retries transport errors but not ClickHouse query errors.

The migration was applied against a real ClickHouse and the resulting view definition confirmed to have no `GROUP BY`:

```
CREATE MATERIALIZED VIEW test_case_runs_validated_on_branch_mv
TO test_case_runs_validated_on_branch (...)
AS SELECT project_id, git_branch, assumeNotNull(test_case_id) AS test_case_id
FROM test_case_runs
WHERE (test_case_id IS NOT NULL) AND (status = 'success') AND (is_flaky = false)
```

### How to test locally

```sh
cd server
mix ecto.migrate -r Tuist.IngestRepo
mix test test/tuist/ingestion/buffer_test.exs test/tuist/clickhouse_retry_test.exs
```

To confirm the view no longer aggregates on insert, against your local ClickHouse:

```sh
curl -s "http://localhost:8123/" --data-binary \
  "SELECT create_table_query FROM system.tables
   WHERE name = 'test_case_runs_validated_on_branch_mv' FORMAT TSV"
```

The returned definition should contain no `GROUP BY` clause.

To exercise the buffer path end to end rather than through the write-through repository that tests use, set `config :tuist, Tuist.Ingestion.Bufferable, write_through_repo: false` in `config/test.exs` and re-run the ingestion suite.


pepicrft:
Closing. This was opened while chasing an incident it turned out to have nothing to do with.

The investigation ended somewhere else entirely: test runs were showing `processing` for hours because of an overnight demand spike, not a failure. Arrivals outran throughput between 00:00 and 06:00 UTC on 2026-07-21 and built a ~1,343 job deficit, driven mainly by one customer's nightly pipeline more than doubling. Nothing in this pull request would have prevented or shortened that.

Two things in here are still real, and are being dropped by closing it rather than being resolved:

**TUIST-37Z** (3,643 events, escalating). A `Ch.Error` raised out of `IngestRepo.query!` inside `Ingestion.Buffer.handle_call/3` kills the buffer process. Each buffer is a single named process shared by every producer on the node, so one bad insert also evicts every unrelated caller blocked on it. The trigger is `test_case_runs_validated_on_branch_mv` aggregating on the write path and exhausting the server `(total)` memory ceiling, which then lets OvercommitTracker kill unrelated queries.

**The Sentry blind spot.** `Sentry.LoggerHandler` runs without `capture_log_messages`, so every `Logger.error/1` in the codebase is invisible. That includes both paths in `TestsController` that fire when an xcresult job cannot be enqueued or its run cannot be marked failed, each of which leaves a run stuck on `processing` indefinitely. That gap actively slowed down today's investigation, because the absence of errors read as evidence that nothing was failing.

The branch is `chore/tests-are-getting-stuck-at` and the commits are intact, so reopening is a click if either becomes worth doing on its own merits.

PR #11992, opened 2026-07-22
fix(runner-image): install Metal Toolchain in Xcode base images

qasim:
## Why

Xcode 16 and earlier bundled the Metal compiler toolchain. [Starting with Xcode 26](https://developer.apple.com/documentation/updates/xcode), Apple made it an optional component downloaded only when needed. Tuist's macOS Xcode base image downloads platform runtimes but not this component, so jobs compiling `.metal` files fail because Xcode cannot execute `metal`.

## What changed

- Run `sudo xcodebuild -downloadComponent MetalToolchain` after downloading all platform runtimes.
- Fail the Packer build unless `xcrun metal --version` can execute the installed compiler.
- Record the component in the Layer 1 image inventory.

The runner and xcresult processor images both inherit this Layer 1 image, so their newly built VMs receive the toolchain without downloading it during jobs. Existing published images remain unchanged until rebuilt.

## Before merging

- [ ] Run the Xcode 26.6 Layer 1 build on a bare-metal `vm-image-builder`, confirm the Metal smoke check passes, and verify the refreshed Tart image publishes successfully.

## Validation

- `packer fmt -check -diff infra/macos-xcode-image/macos-xcode.pkr.hcl`
- `packer validate -syntax-only infra/macos-xcode-image/macos-xcode.pkr.hcl`
- `git diff --check`
- Confirmed Xcode 26 supports the component download and `xcrun metal --version` executes the locally installed compiler.

### DIFF
--- a/infra/macos-xcode-image/AGENTS.md
+++ b/infra/macos-xcode-image/AGENTS.md
@@ -52,8 +52,9 @@ ghcr.io/tuist/tuist-xcresult-processor:<server-semver>
 ```
 
 Why split this layer out: the Xcode install is ~30 min of work
-(unxip, license accept, runFirstLaunch, downloadAllPlatforms). If
-that lived in the runner-image / xcresult-processor packer files,
+(unxip, license accept, runFirstLaunch, Metal Toolchain, and
+downloadAllPlatforms). If that lived in the runner-image /
+xcresult-processor packer files,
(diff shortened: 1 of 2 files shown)


fortmarek (approved):
Thanks a lot for the PR! Looks good

PR #11996, opened 2026-07-22
feat(infra): fast-forward last-writer-wins for the runner cache volume

fortmarek:
## What this is

Reconciliation for the per-account macOS runner cache volume, redone as **fast-forward last-writer-wins**. A promote replaces the account's master wholesale, but the HEAD only advances if the promoting job built on the current version — a compare-and-swap on a monotonic generation. Spans host (tart-kubelet), server (HEAD bump), and guest (runner image).

> Branch history note: this branch first implemented an **object-level union** (commits `ae80edd`, `6ea91ff`, `41144854`). We then pivoted away from it — the union kept every master an ever-growing superset and needed an object-LRU it never had. The later commits **remove the union** and replace it with fast-forward LWW. Review the branch as a whole; the union commits are superseded.

## Why fast-forward LWW over the union

The original bug: promote and converge did a whole-image **replace**, so a job whose end-of-job cache carried fewer objects than the master (a `test` job with manifests but no binaries) clobbered a richer master, and convergence spread the emptiness fleet-wide — 0 local hits across 177 jobs.

The union fixed data-loss but at a cost: masters only ever grew, so it needed object-level LRU as a hard follow-up, plus attach-both-images + per-object `ditto` + an off-lock staging dance to keep the slow merge from blocking `Materialize`.

Fast-forward LWW is simpler and self-bounding. A promoted branch replaces the master wholesale, so the **CLI's own `TUIST_CACHE_MAX_BYTES` prune** shapes every promote and the master stays bounded for free — no object-LRU. The tradeoff is that a concurrent loser's delta is dropped (rebuilt next run), which for this fleet's low same-account concurrency is cheap. This became viable once cross-host convergence was fixed (content-addressed keys, PR #11934), since fast-forward progress depends on hosts converging to the tip.

## How it works

- **Generation as the version marker.** The host tracks its local master's HEAD generation in a `master.generation` sidecar (replacing the inventory-digest + converged-head markers), honored only while the master image exists.
- **Host `InstallMaster`** — a whole-image CoW clone + atomic swap, gated so the master only ever moves to a *newer* generation. This is the whole of reconciliation now: no `hdiutil` attach for the install, no per-object `ditto`, no `mergeMu`/off-lock dance (all removed).
- **The compare-and-swap lives in the server.** `Materialize` captures the base generation the branch was cloned from; the guest relays it at promote; `VolumeHeads.bump_head/5` advances the HEAD `G→G+1` only if it's still at that base (or inserts gen 1 if base 0 and no HEAD). Accept → returns the new generation; stale base → `:conflict` (409).
- **The guest relays a distinct outcome.** It captures the report's HTTP status explicitly (no `curl -f`, which would collapse a 409 and a transport error) and writes `accepted <gen>` / `conflict` (409) / `error` (upload/network/control-plane failure). On `accepted` the host's `Finalize` installs the branch as its local master at that generation; `conflict`/`error` discard and re-converge. So local master and HEAD advance together, and an outage is never mistaken for cross-host contention.
- **Converge** = download HEAD, verify its digest, `InstallMaster` at the HEAD generation. Plain whole-image replace, safe because HEAD is now strictly monotonic.
- A rejected promote's uploaded object (the guest uploads before the compare-and-swap) is recorded in a `runner_volume_master_orphans` table and reclaimed by `PruneVolumeMasterOrphanWorker` after the URL-TTL grace. The accept path forgets the orphan the instant the same digest becomes HEAD, so a digest a later job commits is never deleted, and a never-HEAD digest (which has no download URL) is reclaimed rather than leaking until account deletion.

## User/developer impact

Masters stay bounded (CLI prune shapes them) instead of growing without a cap. Behind or losing hosts don't contribute their delta on that run but re-converge and rebuild cheaply. No behaviour change for jobs beyond warmth.

## Validation

- **Host**: full tart-kubelet suite green under `-race`, including a generation fast-forward/refuse-regression test, a rejected-promote-discards test, and a real macOS `hdiutil` `imageInventoryDigest` round-trip. `darwin` + `linux` builds, `gofmt`, `vet` clean.
- **Server**: `bump_head/5` (cold-establish, cold-reject-when-HEAD-exists, warm fast-forward, stale-base reject), `report_volume_head/4` (accepted generation, stale reject, orphan-prune-on-reject), and controller status mapping (200/409/422). `mix compile --warnings-as-errors`, `credo`, `format` clean.
- **Guest**: `bash -n` clean; `shellcheck` clean on the changed functions.

Do **not** deploy on merge — team decides after review. Production remains on the pre-existing behaviour until this ships.


server/lib/tuist/runners.ex
```diff
@@ -243,24 +243,45 @@ defmodule Tuist.Runners do
...
+          # Do NOT schedule a prune of the losing upload here. Content-addressed
+          # keys mean a rejected digest can be, or later become, the live HEAD (a
+          # concurrent or subsequent accepted promote of the same inventory reuses
+          # the exact object), and download URLs are minted only for the current
+          # HEAD's digest — so a conflict-clocked prune could delete an object that
+          # is HEAD or whose in-flight converge URLs are still valid, or race an
+          # upload→bump. Cleanup stays tied to an observed supersession (the
+          # accepted branch above): a losing digest that never becomes HEAD is a
+          # true orphan, reclaimed by the account-deletion sweep rather than on a
+          # guess made at conflict time.
```
pepicrft: Rejected promotions now leak uploaded master objects for active accounts. The guest uploads the image before this compare-and-swap, but every conflict here leaves its digest object permanently because only accepted supersessions enqueue the prune worker and account deletion may never happen. Under contention, each rejected cache-changing job can add another multi-gigabyte object with no retention path. This also contradicts the pull request description and validation, which say rejected uploads are pruned. Please give each upload attempt a safely reclaimable key or lease, or persist orphan metadata for a sweeper, so conflicts do not accumulate indefinitely.

infra/tart-kubelet/internal/podagent/volume_reconcile.go
```diff
@@ -340,7 +389,18 @@ func (r *Reconciler) finalizeVolume(entry *Entry, actualAccount string, cleanExi
...
+	// by the guest. Zero means no promote or a rejected bump — Finalize then
+	// discards the branch rather than moving the local master off the lineage.
+	entry.Volume.PromotedGeneration = readPromotedGeneration(entry.VolumeStatusDir)
+
+	// For a promote-eligible job (it did cache-changing work for its own account),
+	// record whether the server accepted the fast-forward — the reject-rate signal.
+	// Read-only, failed, and account-mismatched jobs never promote, so they are
+	// excluded to keep the ratio meaningful.
+	if succeeded && dirty && actualAccount != "" && entry.Volume.SourceAccount == actualAccount {
+		RecordVolumePromote(entry.Volume.PromotedGeneration > 0)
```
pepicrft: A missing promoted generation does not exclusively mean that the server rejected a stale base. It also occurs when upload-address minting, the object upload, the report request, response parsing, or the status-share write fails. This records every one of those failures as rejected, while the dashboard interprets the ratio solely as cross-host contention. A storage or control-plane outage will therefore look like cache races. Please relay a distinct accepted, conflict, or failed result and increment rejected only for an actual status code 409 response.

infra/grafana-dashboards/runners.json
```diff
@@ -1620,6 +1620,549 @@
...
+            "fields": "",
+            "values": false
+          },
+          "textMode": "auto"
+        },
+        "targets": [
+          {
+            "datasource": "$datasource",
+            "editorMode": "code",
+            "expr": "100 * sum(rate(tart_kubelet_cache_volume_promote_total{result=\"rejected\",env=\"$env\"}[$__rate_interval])) / sum(rate(tart_kubelet_cache_volume_promote_total{env=\"$env\"}[$__rate_interval]))",
```
pepicrft: Healthy zero-rejection periods render as No data here. Counter-vector label series are created lazily, so until any process records a rejection the numerator has no rejected series. Dividing that empty result by a non-empty accepted rate still returns an empty result, even while promotions are happening. Please initialize both result labels when registering the metric, or coalesce a missing numerator to zero while preserving No data when the denominator is zero.

pepicrft (approved):
I approve the overall direction and implementation. The inline comments should be addressed before merging, especially the unbounded retention of rejected uploads.

server/lib/tuist/runners.ex
fortmarek: Fixed in a4a53ac. Rejected promotes now record the uploaded digest in a new `runner_volume_master_orphans` table and schedule `PruneVolumeMasterOrphanWorker` after the URL-TTL grace to reclaim the object. The accept path forgets the orphan the instant the same digest becomes HEAD, so a digest that later commits is never deleted; the delete only fires while the digest is still an orphan (never accepted) and not the current HEAD, and the row is forgotten only after a confirmed delete so a storage error retries. A never-HEAD digest has no download URL minted for it, so reclaiming it is safe. PR description/validation updated to match.

infra/tart-kubelet/internal/podagent/volume_reconcile.go
fortmarek: Fixed in a139087. The guest now captures the HTTP status explicitly (dropping `curl -f`, which collapsed a 409 and a transport error into one failure) and relays `accepted <gen>` / `conflict` (409) / `error`. The host maps conflict→rejected and upload/network/report/parse/write failures→a distinct `error` bucket, so `rejected` is incremented only for a real 409 and the reject-rate divides by accepted+rejected only (error excluded).

infra/grafana-dashboards/runners.json
fortmarek: Fixed in a139087. All three promote-result series (accepted/rejected/error) are initialized at metric registration, and the panel now coalesces a missing numerator to zero (`(sum(rate(...rejected...)) or vector(0)) / sum(rate(...accepted|rejected...))`) — so it reads 0% during healthy zero-rejection activity and only shows No data when the accepted+rejected denominator is empty.

PR #11999, opened 2026-07-22
fix(server): harden dashboard rate limiting

pepicrft:
The server now accepts Cloudflare's `CF-Connecting-IP` header only when it contains a valid client Internet Protocol address and the request arrived directly from Cloudflare or through the trusted private ingress with a final Cloudflare hop. Requests outside that trust boundary continue to fall back to `X-Forwarded-For` and then the connection address. See [Cloudflare's request header reference](https://developers.cloudflare.com/fundamentals/reference/http-headers/) and [published network ranges](https://www.cloudflare.com/ips/) for the source data.

Dashboard limits are scoped to the matched route pattern, request method, and requester. Authenticated requests use the current user identifier, while anonymous requests use the client address. Hosted production, staging, and development environments allow 60 requests per route and minute, while canary allows 300. Individual routes can override that default.

The [Model Context Protocol](https://modelcontextprotocol.io) retains its separate defaults of 120 requests per minute and 600 in canary because concurrent authenticated sessions share a subject-level bucket.

A crawler could overwhelm a public dashboard route because the application trusted the first `X-Forwarded-For` value, which a client can influence when Cloudflare appends its own value. The previous shared ceiling of 1,000 requests per minute also did not reflect the lower request volume expected from human-facing dashboard routes.

Cross-server counters remain eventually consistent. The cluster broadcast now excludes the local listener, preventing the serving node from counting the same request once from the broadcast and again from the direct hit.

### How to test locally

From `server/`:

```sh
mise exec -- mix test test/tuist_web/remote_ip_test.exs test/tuist_web/rate_limit/in_memory_test.exs test/tuist_web/rate_limit/mcp_test.exs
```

The targeted suite passes with 17 tests and no failures. `git diff --check` also passes.

### DIFF
--- a/server/lib/tuist/environment.ex
+++ b/server/lib/tuist/environment.ex
@@ -1207,16 +1207,30 @@ defmodule Tuist.Environment do
   end
 
   @doc """
+  Returns the bucket size for the dashboard route rate limiter.
+
+  The default values are:
+  - 300 requests per route for canary environments
+  - 60 requests per route for other environments (production, staging, dev)
+  """
+  def dashboard_rate_limit_bucket_size(secrets \\ secrets()) do
+    case get([:dashboard_rate_limit, :bucket_size], secrets) do
+      bucket_size when is_binary(bucket_size) -> String.to_integer(bucket_size)
+      _ -> if can?(), do: 300, else: 60
+    end
+  end
+
+  @doc """
   Returns the bucket size for the MCP rate limiter.
 
   The default values are:
-  - 600 requests for canary environments
-  - 120 requests for other environments (production, staging, dev)
+  - 300 requests for canary environments
+  - 60 requests for other environments (production, staging, dev)
   """
   def mcp_rate_limit_bucket_size(secrets \\ secrets()) do
     case get([:mcp_rate_limit, :bucket_size], secrets) do
       bucket_size when is_binary(bucket_size) -> String.to_integer(bucket_size)
-      _ -> if can?(), do: 600, else: 120
+      _ -> if can?(), do: 300, else: 60
     end
   end
 
--- a/server/lib/tuist/environment.ex
+++ b/server/lib/tuist/environment.ex
@@ -1224,13 +1224,13 @@ defmodule Tuist.Environment do
   Returns the bucket size for the MCP rate limiter.
 
   The default values are:
-  - 300 requests for canary environments
-  - 60 requests for other environments (production, staging, dev)
+  - 600 requests for canary environments
+  - 120 requests for other environments (production, staging, dev)
   """
   def mcp_rate_limit_bucket_size(secrets \\ secrets()) do
     case get([:mcp_rate_limit, :bucket_size], secrets) do
       bucket_size when is_binary(bucket_size) -> String.to_integer(bucket_size)
-      _ -> if can?(), do: 300, else: 60
+      _ -> if can?(), do: 600, else: 120
     end
   end
 
--- a/server/lib/tuist_web/rate_limit/in_memory.ex
+++ b/server/lib/tuist_web/rate_limit/in_memory.ex
@@ -49,12 +49,14 @@ defmodule TuistWeb.RateLimit.InMemory do
     end
   end
 
-  def rate_limit(%Plug.Conn{} = conn, _opts) do
+  def rate_limit(%Plug.Conn{} = conn, opts) do
     if Environment.tuist_hosted?() do
       scale_ms = to_timeout(minute: 1)
-      limit = 1_000
+      limit = opts[:limit] || Environment.dashboard_rate_limit_bucket_size()
+      route = route_pattern(conn)
+      key = "dashboard:#{conn.method}:#{route}:ip:#{TuistWeb.RemoteIp.get(conn)}"
 
-      case hit(TuistWeb.RemoteIp.get(conn), scale_ms, limit) do
+      case hit(key, scale_ms, limit) do
         {:allow, _count} ->
           conn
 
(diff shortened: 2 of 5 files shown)


server/lib/tuist/environment.ex
```diff
@@ -1206,17 +1206,31 @@ defmodule Tuist.Environment do
...
+
+  The default values are:
+  - 300 requests per route for canary environments
+  - 60 requests per route for other environments (production, staging, dev)
+  """
+  def dashboard_rate_limit_bucket_size(secrets \\ secrets()) do
+    case get([:dashboard_rate_limit, :bucket_size], secrets) do
+      bucket_size when is_binary(bucket_size) -> String.to_integer(bucket_size)
+      _ -> if can?(), do: 300, else: 60
+    end
```
fortmarek, line 1220: **uncertain: should we correct the local PubSub double count before setting this to 60?**

`InMemory.hit/4` calls `Phoenix.PubSub.broadcast/3` and then `Local.hit/4`. The local `Listener` is subscribed to that same topic and handles the broadcast with `Local.inc/3`; a regular PubSub broadcast includes same-node subscribers. That means a request contributes one broadcast increment on every node plus another increment on the node serving it.

With production's 2–5 replicas, balanced traffic would exhaust a configured bucket of 60 after roughly 40–50 real requests; a single-node or skewed workload can exhaust it around 30. The current tests mock `FixWindow.hit/5`, so they don't exercise the listener interaction.

Could we ignore broadcasts originating on the current node (or publish only to remote nodes) and add an integration test around the real listener? Until then, starting the dashboard default at 120 would provide safer headroom.

server/lib/tuist/environment.ex
```diff
@@ -1206,17 +1206,31 @@ defmodule Tuist.Environment do
...
+  - 300 requests for canary environments
+  - 60 requests for other environments (production, staging, dev)
   """
   def mcp_rate_limit_bucket_size(secrets \\ secrets()) do
     case get([:mcp_rate_limit, :bucket_size], secrets) do
       bucket_size when is_binary(bucket_size) -> String.to_integer(bucket_size)
-      _ -> if can?(), do: 600, else: 120
+      _ -> if can?(), do: 300, else: 60
     end
   end
```
fortmarek, line 1235: **uncertain: should MCP retain its separate 120-request default?**

The incident described in the PR is a crawler hitting public dashboard routes, whereas MCP is authenticated before this limiter and its key is shared by all concurrent sessions for the same user, project, or account. A nominal 60 allows only one HTTP request per second across all of those sessions—and less in practice because of the PubSub counting behavior above—so concurrent agents can rate-limit one another even when no session is abusive.

I'd keep MCP at 120 independently of the dashboard default, then tune it from denial telemetry if needed.

fortmarek (commented):
I have two concerns about choosing `60` for these buckets. The route-pattern keying looks good, but the effective request count and the MCP workload differ from the nominal dashboard limit. The existing `CF-Connecting-IP` trust-boundary thread also still needs to be addressed.

fortmarek (commented):
Authenticated dashboard traffic should not share the anonymous IP-only bucket.

server/lib/tuist_web/rate_limit/in_memory.ex
```diff
@@ -49,12 +49,14 @@ defmodule TuistWeb.RateLimit.InMemory do
...
   end
 
-  def rate_limit(%Plug.Conn{} = conn, _opts) do
+  def rate_limit(%Plug.Conn{} = conn, opts) do
     if Environment.tuist_hosted?() do
       scale_ms = to_timeout(minute: 1)
-      limit = 1_000
+      limit = opts[:limit] || Environment.dashboard_rate_limit_bucket_size()
+      route = route_pattern(conn)
+      key = "dashboard:#{conn.method}:#{route}:ip:#{TuistWeb.RemoteIp.get(conn)}"
```
fortmarek: **[P2] Key authenticated dashboard traffic by principal**

`key` always uses the client IP, including after `:browser_app` has loaded the authenticated user. Consequently, everyone behind the same office VPN or NAT shares the 60-request bucket. Because the key uses the route pattern rather than concrete account, project, and resource values, that traffic is also aggregated across every project. One active user can therefore rate-limit unrelated authenticated users.

Could we use the user ID for authenticated requests, falling back to IP for anonymous public-dashboard traffic? A separate, higher IP-level ceiling could remain as defense in depth. This would allow a higher authenticated limit without weakening protection against anonymous crawlers.

fortmarek (approved):
the above notes are non-blocking, but highly recommend to address them before merging.

server/lib/tuist_web/remote_ip.ex
```diff
@@ -6,20 +6,35 @@ defmodule TuistWeb.RemoteIp do
...
-      forwarded_for
-      |> String.split(",")
-      |> Enum.map(&String.trim/1)
-      |> List.first()
-    else
-      conn.remote_ip
-      |> :inet_parse.ntoa()
-      |> to_string()
-    end
+    cloudflare_ip || forwarded_ip || format_ip(conn.remote_ip)
```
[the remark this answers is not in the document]
pepicrft: Restricted `CF-Connecting-IP` to a valid address received from a direct Cloudflare peer or from the final Cloudflare hop behind the private ingress. The ranges come from Cloudflare’s published Internet Protocol version 4 and version 6 lists, and the fallback behavior remains unchanged.

server/lib/tuist/environment.ex
pepicrft: Changed the cluster broadcast to exclude the local listener, so the serving node records only the direct hit. Added a test around the real listener that verifies one request increments the local bucket once.

server/lib/tuist/environment.ex
pepicrft: Restored the [Model Context Protocol](https://modelcontextprotocol.io) defaults to 120 requests per minute and 600 in canary, independently of the dashboard limit.

server/lib/tuist_web/rate_limit/in_memory.ex
pepicrft: Authenticated dashboard requests now use the current user identifier in the route bucket. Anonymous requests continue to use the client Internet Protocol address.

PR #12002, opened 2026-07-22
fix(server): harden Kura deployment reconciliation

pepicrft:
## What changed

This makes Kura deployment scheduling transactional per server, limits each pass to 100 eligible servers, prevents more than one open deployment for a server, contains failures to the affected server, and makes runner-cache reconciliation retryable and idempotent. It also keeps retired-region identity available for teardown and disables warm handoffs in production until a stable account-region binding can own publication atomically.

## Why

A single scheduling collision could abort a whole reconciliation pass, concurrent retries could race to create duplicate open deployments, and an unbounded pass would perform work proportional to every Kura server at once. Partial runner-cache operations could also strand state. Warm handoffs assume stronger endpoint ownership guarantees than the controller currently provides.

## Root cause

The reconciler treated batch scheduling and multi-step provisioning as if they were atomic even though database and cluster operations converge independently. Warm handoffs also tied endpoint publication to a server row instead of a stable account-region identity that survives a move.

## Approach

The eligibility query excludes servers that already have the requested image or an open deployment, orders candidates deterministically, and limits each pass to 100 servers. Each selected server is then locked and scheduled independently, with database uniqueness preserving the concurrency invariant. Reconciliation work remains unique, and multi-step runner-cache operations preserve retryable state.

The existing warm-handoff machinery remains testable but is gated off in production. [#12020](https://github.com/tuist/tuist/issues/12020), assigned to Pepicrft, owns production enablement with these explicit gates:

1. Introduce a stable account-region binding that owns the customer endpoint independently of either source or target server row.
2. Make promotion transfer that binding atomically, then derive endpoint publication from the binding.
3. Exercise forced moves in staging while continuously reading and writing through the customer endpoint, including retries and controller restarts at every phase.
4. Enable one staging account, then a small production allowlist, and expand only after endpoint continuity and rollback alerts remain clean.

Until those gates are implemented and validated, production configuration stays disabled and move requests fail closed.

## Impact

Healthy servers continue reconciling when another server fails. Repeated work converges without duplicate deployments, large fleets roll forward in bounded passes, and unsafe warm handoffs remain unavailable in production.

## Validation

- 69 focused tests in `test/tuist/kura_test.exs` passed
- `mix credo --strict lib/tuist/kura.ex test/tuist/kura_test.exs`
- `mix format` on the changed source and test files
- `git diff --check`

## Review order

Second in the server stack. Review after [#12001](https://github.com/tuist/tuist/pull/12001), then continue with [#12004](https://github.com/tuist/tuist/pull/12004).

### DIFF
--- a/server/config/test.exs
+++ b/server/config/test.exs
@@ -83,6 +83,11 @@ config :tuist, TuistWeb.Endpoint,
   secret_key_base: "pbaHQK0N946e06chs5G1/RUJnkI//2QshGgUvJQkADTV3AiQHV/dXlLdjnaQxtxx",
   server: false
 
+# Production builds keep warm handoffs disabled until the public customer
+# endpoint is owned by a stable account-region binding. Tests enable the
+# transition machinery so its invariants remain covered while that work lands.
+config :tuist, :kura_warm_handoffs_enabled, true
+
 config :tuist,
   api_pipeline_producer_module: Broadway.DummyProducer,
   api_pipeline_producer_options: []
--- a/server/lib/tuist/kura.ex
+++ b/server/lib/tuist/kura.ex
@@ -51,6 +51,7 @@ defmodule Tuist.Kura do
   # to rescue and forces manual intervention. Only terminal servers
   # (`:destroying`/`:destroyed`) are skipped.
   @version_rollout_statuses [:provisioning, :replicating, :active, :failed]
+  @version_rollout_batch_size 100
 
   @doc "Reconciles desired Kura server rows with the observed Kubernetes state."
   def reconcile_orphaned_deployments, do: Reconciler.reconcile()
@@ -155,9 +156,15 @@ defmodule Tuist.Kura do
   end
 
   defp servers_needing_version_query(image_tag) do
-    deployment_exists_query =
+    deployment_for_image_exists_query =
       from(d in Deployment,
         where: parent_as(:server).id == d.kura_server_id and d.image_tag == ^image_tag,
+        select: 1
+      )
+
+    open_deployment_exists_query =
+      from(d in Deployment,
+        where: parent_as(:server).id == d.kura_server_id and d.status in [:pending, :running],
         select: 1
       )
 
@@ -171,7 +178,10 @@ defmodule Tuist.Kura do
       where: s.move_phase == :none,
       where: s.status in @version_rollout_statuses,
       where: s.current_image_tag != ^image_tag,
-      where: not exists(deployment_exists_query)
+      where: not exists(deployment_for_image_exists_query),
+      where: not exists(open_deployment_exists_query),
+      order_by: [asc: s.updated_at, asc: s.id],
+      limit: ^@version_rollout_batch_size
     )
   end
 
@@ -1150,7 +1160,7 @@ defmodule Tuist.Kura do
       Repo.transaction(fn ->
         server
         |> lock_server_or_rollback()
-        |> maybe_insert_scheduled_deployment(region, image_tag)
+        |> insert_scheduled_deployment(region, image_tag)
       end)
     end
   end
@@ -1162,15 +1172,32 @@ defmodule Tuist.Kura do
     end
   end
 
-  defp maybe_insert_scheduled_deployment(server, region, image_tag) do
-    if deployment_for_image_exists?(server.id, image_tag) or open_deployment_exists?(server.id) do
-      nil
-    else
-      case insert_deployment(server, region, image_tag) do
-        {:ok, deployment} -> deployment
-        {:error, reason} -> Repo.rollback(reason)
-      end
+  defp insert_scheduled_deployment(server, region, image_tag) do
+    case insert_deployment(server, region, image_tag) do
+      {:ok, deployment} ->
+        deployment
+
+      {:error, %Ecto.Changeset{} = changeset} ->
+        if open_deployment_conflict?(changeset) do
+          nil
+        else
+          Repo.rollback(changeset)
+        end
+
+      {:error, reason} ->
+        Repo.rollback(reason)
     end
+  end
+
+  defp open_deployment_conflict?(changeset) do
+    Enum.any?(changeset.errors, fn
+      {:kura_server_id, {_message, metadata}} ->
+        metadata[:constraint] == :unique and
+          metadata[:constraint_name] == "kura_deployments_one_open_per_server_index"
+
+      _error ->
+        false
+    end)
   end
 
   defp insert_deployment(%Server{} = server, region, image_tag) do
@@ -1181,14 +1208,6 @@ defmodule Tuist.Kura do
     }
     |> Deployment.create_changeset()
     |> Repo.insert()
-  end
-
-  defp deployment_for_image_exists?(server_id, image_tag) do
-    Repo.exists?(
-      from(d in Deployment,
-        where: d.kura_server_id == ^server_id and d.image_tag == ^image_tag
-      )
-    )
   end
 
   defp open_deployment_exists?(server_id) do
--- a/server/lib/tuist/kura.ex
+++ b/server/lib/tuist/kura.ex
@@ -42,6 +42,7 @@ defmodule Tuist.Kura do
   @public_endpoint_timeout 5_000
   @provisioner_node_ref_format ~r/^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$/
   @provisioner_node_ref_max_length 53
+  @warm_handoffs_enabled Application.compile_env(:tuist, :kura_warm_handoffs_enabled, false)
 
   # A version change is a deliberate fix delivery, so it must reach
   # degraded servers, not only healthy ones. A `:failed` or `:replicating`
@@ -93,7 +94,7 @@ defmodule Tuist.Kura do
   """
   def schedule_runtime_image_deployments do
     case runtime_image_tag() do
-      nil -> {:ok, []}
+      nil -> {:ok, %{scheduled: [], failures: []}}
       image_tag -> schedule_runtime_image_deployments(image_tag)
     end
   end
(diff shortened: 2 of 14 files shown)


fortmarek (commented):
One additional rollout finding.

server/config/test.exs
```diff
@@ -83,6 +83,11 @@ config :tuist, TuistWeb.Endpoint,
   secret_key_base: "pbaHQK0N946e06chs5G1/RUJnkI//2QshGgUvJQkADTV3AiQHV/dXlLdjnaQxtxx",
   server: false
 
+# Production builds keep warm handoffs disabled until the public customer
+# endpoint is owned by a stable account-region binding. Tests enable the
+# transition machinery so its invariants remain covered while that work lands.
+config :tuist, :kura_warm_handoffs_enabled, true
+
```
fortmarek, line 90: [P1] Define the production enablement path for warm handoffs

When is the stable account-region binding expected to be implemented? I can’t find it in this PR stack or a linked follow-up issue. Without a concrete owner and rollout plan, this condition can keep warm handoffs disabled indefinitely. I’d prefer we implement the required binding, exercise the complete handoff on staging, and then promote the same behavior to production rather than merging a test-only feature path.

server/lib/tuist/kura.ex
```diff
@@ -126,16 +127,31 @@ defmodule Tuist.Kura do
...
 
   def schedule_version_deployments(image_tag) when is_binary(image_tag) do
-    deployments =
+    {scheduled, failures} =
       image_tag
       |> servers_needing_version_query()
       |> Repo.all()
-      |> Enum.map(&create_deployment(&1, image_tag))
+      |> Enum.reduce({[], []}, fn server, {scheduled, failures} ->
+        case schedule_version_deployment(server, image_tag) do
```
fortmarek, line 136: [P2] Bound and batch runtime deployment scheduling

This loop loads every server needing the image and calls `schedule_version_deployment/2` for each one. That helper opens a transaction and performs a row-lock query plus up to two `Repo.exists?` queries and an insert, so a fleet-wide rollout creates an unbounded O(N) sequence of database round trips in a single reconciler job. This can monopolize the default Oban queue or exceed the job timeout as the fleet grows. Please limit the initial query to a fixed batch and fetch existing/open deployment state in that query (or in one query keyed by the selected server IDs), while retaining the unique constraint to resolve races.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -315,10 +324,20 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   defp mesh_enabled?(%Regions{provisioner_config: %{mesh: mesh}}) when is_boolean(mesh), do: mesh
   defp mesh_enabled?(_region), do: false
 
+  defp mesh_peers_sync_enabled?(account, %Regions{} = region) do
+    mesh_enabled?(region) and Entitlements.allows?(account, :self_hosted_cache)
```
fortmarek, line 336: [P2] Preload entitlements for manifest reconciliation

`Entitlements.allows?/2` calls `Billing.get_current_active_subscription/1` on hosted deployments. `manifest_revision/2` is evaluated once per server in the reconciler’s observed-server batch, so this changed line adds one subscription query per server; applying a manifest can query it again through `extension_env/2`. Please load the relevant subscriptions once for the batch, build an account-to-plan/entitlement lookup, and pass the derived entitlement into manifest generation.

fortmarek (commented):
Two additional scalability findings. The existing `meshPublicPeerPublished` controller/CRD finding remains unresolved and was not duplicated.

fortmarek (approved):
https://github.com/tuist/tuist/pull/12002#discussion_r3631221433 -> let's figure out whether we even this flag and if so, when can we remove it.

Otherwise, looks good 👍

server/lib/tuist/kura/runner_cache.ex
```diff
@@ -46,45 +47,55 @@ defmodule Tuist.Kura.RunnerCache do
...
+    case runner_cache_cohort() do
+      {:ok, account_ids} ->
+        Enum.each(runner_cache_regions(), &reconcile_region(&1, account_ids))
+
+      {:error, reason} ->
+        # The runner feature flag can control request paths with global or
+        # percentage gates, but those gates are unsafe for durable, per-account
+        # infrastructure. Preserve every existing node and provision nothing
+        # until the flag returns to an explicit actor-only shape.
+        Logger.error("kura.runner_cache: refusing to reconcile an unsafe runner-cache cohort",
```
esnunes, line 76: Failing closed here is the right call, and I like that it preserves existing nodes. One thing I wanted to flag: when the cohort is unsafe, the whole runner-cache loop stops for every account — no provisioning, no teardown — and the only signal is this `Logger.error`. That's a quieter channel than the one this same PR adds for scheduling failures (`reconciler.ex:108` captures to Sentry with a `failure_kind` tag).

It feels reachable, too: `FunWithFlags.UI` is mounted at `/ops/flags`, so a global boolean gate is one click away — and that's the natural thing to do when runners go GA, at which point `FeatureFlags.runners_enabled?/1` starts saying yes for everyone while no cache node ever gets provisioned. Would a `Sentry.capture_message` (or a metric) on this branch be worth adding, so the freeze is loud rather than log-only?

esnunes (approved):
Read through this one carefully — the core of it looks right to me. Isolating each server in its own transaction, taking the row lock before the open-deployment check, and then backing the invariant with a partial unique index is the right layering: the lock orders the common path and the index catches anything that doesn't take it. I also checked the migration's magic integers against both Ecto enums and they line up, the `unique_constraint` name matches the index, and pulling retry history through a single windowed query instead of per-server lookups is a nice touch.

Approving — the two things I left inline are about the manifest side and observability rather than the core logic: the new `meshPublicPeerPublished` field doesn't appear in the chart CRD or the Go controller, so as far as I can tell the apiserver would prune it today, and the unsafe-cohort bail-out is log-only when this same PR routes scheduling failures to Sentry. Neither blocks.

Worth noting the peers-sync / `+nosync` hunks look byte-identical to #11926, which merged a few minutes ago — they should drop out on rebase.

Heads-up on CI: the red check is Trivy flagging pre-existing `pnpm-lock.yaml` CVEs (dompurify, fast-uri), unrelated to this diff; the Seed job was cancelled rather than failed.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -19,8 +19,9 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   alias Tuist.Kura.Server
 
   @namespace "kura"
-  @manifest_revision "2026-07-16-cas-capacity-v1"
+  @manifest_revision "2026-07-19-peer-publication-v1"
```
esnunes, line 22: Small addition to the CRD/controller point above rather than a separate concern: this PR also bumps `@manifest_revision` to `2026-07-19-peer-publication-v1`, so every instance drifts and re-applies — and if `meshPublicPeerPublished` is pruned on the current CRD, that new key is the only thing the re-apply actually delivers. Not costly in itself (the rendered spec should otherwise be identical, so no pod should roll), just worth knowing the fleet-wide re-apply and the field land together.

server/lib/tuist/kura.ex
pepicrft: Addressed in `e37290d433`. The eligibility query now excludes the desired image and open deployments, orders candidates deterministically, and limits each pass to 100 servers. The locked scheduling path no longer runs the two per-server existence queries, while the unique constraint still resolves races. A 101-server regression proves the work is split across two passes.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -335,10 +354,34 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
...
   end
 
+  # Keep the public peer hostname on every move phase because it is also part of
+  # the instance certificate identity. Publication is a separate boolean so a
+  # warming or draining instance does not claim the regional public route. The
+  # controller treats a missing field as the legacy publish-by-host behavior,
+  # which keeps the controller-first rollout backward compatible.
+  defp put_mesh_public_peer_publication(spec, region, server) do
+    if mesh_enabled?(region) do
+      Map.put(spec, "meshPublicPeerPublished", owns_public_endpoints?(server))
```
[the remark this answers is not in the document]
pepicrft: Addressed by the next pull request in the stack, #12004. It adds `meshPublicPeerPublished` to the `KuraInstance` custom resource definition and Go application programming interface, then makes the controller use it when deciding whether to publish the public peer route. #12002 keeps production warm handoffs disabled, so the unsupported path cannot be activated during the interval before #12004 lands. I will keep these two pull requests adjacent in the merge order.

server/lib/tuist/kura/runner_cache.ex
pepicrft: Added in 802c93afad. An unsafe runner-cache cohort now captures `Kura runner-cache reconciliation paused` in Sentry at error level, with a stable `failure_kind` tag and the inspected failure detail. I also added a regression assertion for the global gate case. The focused runner-cache suite passes with 12 tests, and the combined server-stack suite passes with 114 tests.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
pepicrft: Agreed. The revision bump will cause existing instances to reapply. Production warm handoffs remain disabled in #12002, and #12004 immediately adds controller support before any production enablement. For instances whose remaining rendered spec is unchanged, this reapplies the manifest revision and publication field without changing the pod template, so it should not roll pods. I am treating #12002 and #12004 as an adjacent rollout.

PR #12004, opened 2026-07-22
fix(kura): make peer publication cutovers resilient

pepicrft:
## What changed

This makes peer publication tolerate controller convergence in either order. The controller keeps a usable fallback while replacements become ready, avoids stale peer ownership, and publishes peer targets from observed endpoint state. Server-only re-promotions preserve the live Kura controller and Swift registry images instead of rolling compatible infrastructure backward with the server image.

## Why

Kura instances, endpoint discovery, and the peer demultiplexer reconcile independently. A rollout could briefly withdraw a working peer route or restore an older controller image even though the server rollback itself was valid.

## Root cause

The previous cutover assumed resource observation order and server, controller, and registry image versions moved together atomically.

## Approach

Publication now derives from ready endpoint state and retains the previous valid route during replacement. The server deployment workflow resolves controller and registry image tags independently, preserving live compatible images for server-only rollbacks. The legacy account-wide peer Service migration records only state that drives recovery or retirement; the unused cutover timestamp has been removed.

## Migration cleanup

This compatibility path is intentionally one-shot. The Kura controller owner should remove `retireLegacyAccountPublicPeerService` and its legacy annotations, helpers, and tests after every managed cluster reports both of these conditions across two consecutive production releases:

- Zero account-wide peer Services that still own a public peer hostname.
- Zero Services carrying a legacy peer migration phase.

The two observations must be separated by at least ten minutes, which covers two configured Domain Name System cache lifetimes. A rollback remains safe during that window because the repaired fallback Service is retained until the replacement address is observed.

## Impact

Peer routes remain available during ordinary rollout ordering, and server fixes can be re-promoted without crossing a controller protocol boundary.

## Validation

- `go test ./...` passed with Go 1.26.4
- Go formatting completed with `gofmt`
- 123 focused server tests passed before this review update
- Managed staging Helm chart rendered successfully before this review update
- `git diff --check`

## Review order

Third in the server stack. Review after [#12002](https://github.com/tuist/tuist/pull/12002), then continue with [#12005](https://github.com/tuist/tuist/pull/12005).


fortmarek (commented):
The cutover state machine itself is appropriately restart-safe, including `legacyPeerMigrationAnnotation`, but its lifecycle should be bounded. I left two cleanup findings: document when the complete compatibility path will be removed, and avoid retaining state that is not consumed.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -532,6 +605,616 @@ func (r *KuraInstanceReconciler) reconcilePeerDNSEndpoint(ctx context.Context, i
 	return err
 }
 
+// retireLegacyAccountPublicPeerService migrates a host-network region from the
```
fortmarek, line 607: [P2] Please include an explicit cleanup follow-up in the PR description. This is one-shot compatibility machinery for the former ownerless account-wide Service, so after every managed environment reports zero matching legacy Services (and the rollout/TTL soak has elapsed), we should remove `retireLegacyAccountPublicPeerService`, the complete `legacyPeer*` annotation bundle, helpers, and tests. `legacyPeerMigrationAnnotation` is necessary while this state machine exists, but without a removal criterion this large temporary path is likely to become permanent.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -532,6 +605,616 @@ func (r *KuraInstanceReconciler) reconcilePeerDNSEndpoint(ctx context.Context, i
...
+			if now.Before(observedAt.Add(time.Duration(peerDNSRecordTTLSeconds) * time.Second)) {
+				return nil
+			}
+			delete(annotations, legacyPeerFallbackRepublishAnnotation)
+			delete(annotations, legacyPeerFallbackObservedAnnotation)
+		}
+
+		delete(annotations, externalDNSHostnameAnnotation)
+		annotations[legacyPeerMigrationAnnotation] = legacyPeerPhaseCutoverRequested
+		annotations[legacyPeerCutoverRequestedAnnotation] = now.Format(time.RFC3339)
```
fortmarek: [P3] `legacyPeerCutoverRequestedAnnotation` is written here and deleted when restarting the cutover, but it is never read. The migration phase already persists the control state, so please remove this annotation/write unless an external observer consumes it; if it is intentionally diagnostic, document that purpose.

fortmarek (approved):
aligned, let's just not forget the clean-up after the rollout

infra/kura-controller/controllers/kurainstance_controller.go
pepicrft: Addressed in the pull request description. The compatibility path is removed after every managed cluster reports zero legacy account-wide peer Services and zero migration-phase annotations across two consecutive production releases separated by at least ten minutes.

infra/kura-controller/controllers/kurainstance_controller.go
pepicrft: Rebase note: the current commit is `64ec7c4271`.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -532,6 +604,614 @@ func (r *KuraInstanceReconciler) reconcilePeerDNSEndpoint(ctx context.Context, i
...
+		annotations[legacyPeerTargetAddressesAnnotation] = encodePeerAddresses(state.targetAddresses)
+		service.SetAnnotations(annotations)
+		repairLegacyPeerFallback(service, instance)
+		return r.Update(ctx, service)
+
+	case legacyPeerPhaseRepairing:
+		if annotations[externalDNSHostnameAnnotation] != instance.Spec.MeshPublicPeerHost {
+			return nil
+		}
+		if !legacyPeerFallbackRepaired(service, instance) {
```
esnunes, line 693: One thing I wanted to flag about the fallback repair: `repairLegacyPeerFallback` runs before the `ready` gate here, and it sets the selector to `selectorLabels(instance)`, which is instance-scoped. Since `meshPublicPeerHost` is derived from account+region only (`Regions.peer_public_host/2`), a warm handoff gives two CRs — the base one and the `-m` one — the same host, so both match the legacy Service and both enter this phase. As far as I can tell each would then rewrite the selector to its own pods, making `legacyPeerFallbackRepaired` false for the other on every 30s reconcile, so the fallback's endpoints would flap between the two instances.

I might be missing a guard, but this looks latent rather than live: `@warm_handoffs_enabled` defaults to false and `rollout/2` rejects non-`:none` move phases, so no `-m` sibling reaches a cluster today. Given the PR describes re-enabling warm handoffs as the direction, would it be worth electing an owner here the way `peerDemuxDesiredState` already does — oldest by CreationTimestamp, then name — and having non-owners no-op?

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -532,6 +604,614 @@ func (r *KuraInstanceReconciler) reconcilePeerDNSEndpoint(ctx context.Context, i
...
+		delete(annotations, externalDNSHostnameAnnotation)
+		annotations[legacyPeerMigrationAnnotation] = legacyPeerPhaseCutoverRequested
+		service.SetAnnotations(annotations)
+		return r.Update(ctx, service)
+
+	case legacyPeerPhaseCutoverRequested:
+		if targetChanged {
+			return r.restartLegacyPeerCutover(ctx, service, instance, annotations, state.targetAddresses, now)
+		}
+		if !ready {
```
esnunes, line 747: The `!ready` branches in `cutover-requested` and `draining` both call `restartLegacyPeerCutover`, which re-publishes the legacy DNS annotation and resets to `repairing-fallback`. Since `ready` requires the region's demux DaemonSet to be fully rolled at the exact current config hash, and that config is re-rendered whenever any instance in the region changes, an unrelated tenant's reconcile can knock this account's migration back to the start — a fresh ~300s fallback re-observation plus another 600s drain, with the public peer record flipping back and forward each time. In a region with a few accounts churning I'm not sure it reliably converges.

Being conservative here is clearly deliberate and I don't think it's unsafe, so this isn't a blocker. But in `draining` specifically, DNS already points at the target and the repaired fallback is still up — would pausing the deadline while a roll is in flight be strictly safer than tearing the cutover down and rebuilding it? Might also be worth a test that makes the DaemonSet stale mid-`draining`; the current one only does it before the migration starts.

.github/workflows/server-deployment.yml
```diff
@@ -718,6 +730,34 @@ jobs:
...
           echo "Reaching apiserver at $API"
           kubectl --request-timeout=10s get --raw /version >/dev/null
+      - name: Resolve Kura controller image tag
+        run: |
+          set -euo pipefail
+
+          if [ -n "$KURA_CONTROLLER_IMAGE_TAG_INPUT" ]; then
+            controller_tag="$KURA_CONTROLLER_IMAGE_TAG_INPUT"
+            controller_source="explicit"
+          elif [ -z "${{ inputs.image_tag }}" ]; then
```
esnunes: Heads-up, definitely not a blocker: this step reads its own input safely via `$KURA_CONTROLLER_IMAGE_TAG_INPUT`, then drops to `elif [ -z "${{ inputs.image_tag }}" ]`, which expands into the script text before bash sees it, in a job holding cluster credentials. Not a new hole — the kura-runtime block at line 1003 does the same, and dispatching needs repo write — but since the env-var idiom is already right there in this step, passing `image_tag` the same way would close it for free and keep the new code consistent with itself.

esnunes (approved):
Really nice piece of work — the ordering here is the hard part of a cutover and it's right. Repairing the old LoadBalancer into a working Cluster-routed fallback *before* asking DNS to move off it, gating each transition on a real mTLS handshake rather than an inference, persisting the phase on the Service so a controller restart resumes mid-flight, and re-publishing the fallback when the replacement regresses — that's a lot of failure modes closed deliberately. The demux dedup and the DNS-1123 validation also close the nginx config interpolation surface properly.

I've left three inline notes. Two are about convergence rather than safety: sibling instances sharing a peer host during a warm handoff both writing the legacy Service's selector with no owner election, and the `!ready` branches restarting the whole cutover when the region's demux happens to be mid-roll for an unrelated reason. Neither is a correctness hazard for traffic — both fail toward keeping the fallback — but both look like they could fail to settle. The third is a small workflow hygiene point.

Two things I couldn't verify myself, so treating them as open rather than confirmed: I had no Go toolchain available, so everything here comes from reading rather than running, and I haven't traced whether the controller pod can reliably dial the Hetzner LoadBalancer address from inside the cluster — if hairpinning doesn't work there, `peerPublicPathsReady` would stay false forever and the migration would simply never start (safe, but silent apart from an Info log).

On CI: everything is green after the rebase except `Format`, which hit its 20-minute job timeout during the "Check Elixir format" step rather than reporting a formatting violation — worth a re-run to confirm it's just the timeout.

infra/kura-controller/controllers/kurainstance_controller.go
pepicrft: Addressed in `5aafc69b4f`. The migration now derives its owner from the exact route selected by `peerDemuxDesiredState`, and non-owner siblings return before reading or mutating the legacy Service. I added a regression test with an older base instance and a newer `-m` sibling that verifies the sibling cannot rewrite the fallback selector.

infra/kura-controller/controllers/kurainstance_controller.go
pepicrft: Addressed in `5aafc69b4f`. While draining, a target change, failed replacement path, or public name-resolution regression still restarts the cutover. If only the exact demultiplexer rollout readiness becomes stale while the public path remains healthy, the controller now clears the retirement deadline and stays in the draining phase. Once the rollout is ready again, it starts a fresh full drain interval. The new regression test makes the configuration hash stale exactly when the original deadline expires and verifies that the fallback remains unpublished and undeleted.

.github/workflows/server-deployment.yml
pepicrft: Addressed in `5aafc69b4f`. `image_tag` now reaches the shell through `SERVER_IMAGE_TAG_INPUT` instead of being interpolated into the script text. I also switched the existing Kura runtime branch in the same job to use that environment variable.

PR #12005, opened 2026-07-22
fix(server): tear down Kura servers in retired regions

pepicrft:
## What changed

This adds a dedicated reconciliation pass for Kura servers whose regions remain in the catalog only as retirement tombstones. It moves those servers through the existing destruction path in bounded batches.

## Why

Retired-region rows can outlive the node pool and cluster resources they originally targeted. Without an explicit cleanup pass, their Kura instances remain unschedulable and are retried indefinitely.

## Root cause

The region tombstone made old cluster identity resolvable, but no reconciliation step converted stranded servers into destruction work.

## Approach

The reconciler queries non-destroyed servers in retired regions, schedules destruction through the standard server transition, and lets the existing teardown reconciliation remove their cluster resources.

## Impact

Old retired-region workloads converge to deletion without a one-off operator script. Active regions are unaffected.

## Validation

- 64 focused server tests passed
- `mix format --check-formatted`
- `mix credo --strict`
- `git diff --check`

## Review order

Fourth in the server stack. Review after [#12004](https://github.com/tuist/tuist/pull/12004), then continue with [#12006](https://github.com/tuist/tuist/pull/12006).

### DIFF
--- a/server/lib/tuist/kura/reconciler.ex
+++ b/server/lib/tuist/kura/reconciler.ex
@@ -12,14 +12,16 @@ defmodule Tuist.Kura.Reconciler do
   independently-mutated state machine. Each tick:
 
     1. schedule runtime-image drift for active servers,
-    2. finalise destroys after the custom resource disappears,
-    3. apply open deployments (the rollout fast path), and
-    4. project every other present-intent server: observe the backing
+    2. schedule teardown for servers stranded in retired regions,
+    3. finalise destroys after the custom resource disappears,
+    4. drain the source of completed moves,
+    5. apply open deployments (the rollout fast path), and
+    6. project every other present-intent server: observe the backing
        `KuraInstance`, record the observation (`observed_image_tag` /
        `last_observed_at`), and re-derive `status` from
        `(latest deployment intent, observed image, endpoint readiness)`.
 
-  Because step 4 re-derives `status` from observation every tick,
+  Because step 6 re-derives `status` from observation every tick,
   `:failed` is never a sticky terminal sink: a server whose backing
   resource recovers and reports the intended image with a serving
   endpoint heals back to `:active` in place, with no new deployment row
@@ -83,6 +85,7 @@ defmodule Tuist.Kura.Reconciler do
     RunnerCache.reconcile()
 
     schedule_runtime_image_deployments()
+    reconcile_retired_region_servers()
     reconcile_destroying_servers()
     reconcile_moving_out_servers()
     handled = reconcile_deployments()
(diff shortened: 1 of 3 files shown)


fortmarek (approved):

server/test/tuist/kura/reconciler_test.exs
```diff
@@ -616,6 +616,34 @@ defmodule Tuist.Kura.ReconcilerTest do
...
+      |> Repo.update()
+
+    stub(Provisioner, :current_image_tag, fn _server -> {:error, :not_found} end)
+
+    Reconciler.reconcile()
+
+    # A retired region only stays in the catalog so its leftovers can be torn
+    # down. Without this the row keeps its status forever and its pod sits
+    # unschedulable against a node pool that was deleted with the region.
+    assert Repo.reload!(server).status in [:destroying, :destroyed]
```
esnunes: Minor one on this assertion. Because `current_image_tag` is stubbed to `{:error, :not_found}`, `reconcile_destroying_servers` finalises the row to `:destroyed` within the same tick, and `in [:destroying, :destroyed]` accepts either — so the test passes as long as *something* destroyed the server, not specifically the new pass.

Asserting the exact end state (`:destroyed` under that stub) would pin the two-pass sequence; asserting `:destroying` under an `{:ok, tag}` stub would isolate the new pass on its own. The negative test below is a nice control. If you think they're worth the lines, the batch ceiling and the destroy-failure branch are the two paths still uncovered.

server/lib/tuist/kura/reconciler.ex
```diff
@@ -83,6 +83,7 @@ defmodule Tuist.Kura.Reconciler do
     RunnerCache.reconcile()
 
     schedule_runtime_image_deployments()
+    reconcile_retired_region_servers()
```
esnunes, line 88: Tiny docs one: the moduledoc's numbered "Each tick:" list doesn't mention this pass. The move-drain pass is already missing from it so the list had drifted before this PR, but since this is the pass that destroys rows it feels like the one most worth naming there.

esnunes (approved):
Nice, tightly-scoped change — and the integration points are the part I liked most. Sequencing the sweep before `reconcile_destroying_servers` means a swept server is actually torn down in the same tick rather than waiting 30s, and putting it before `reconcile_deployments` means the guard clause at reconciler.ex:264 cancels any still-open deployment instead of rolling it out against a server on its way out. Reusing `destroy_server/1` rather than updating rows directly keeps cache-endpoint cleanup and the broadcast consistent with an operator-initiated destroy, and the query shape matches the neighbouring passes closely enough that it reads as part of the existing loop.

I left two small notes inline — a test assertion that can't tell the new pass apart from the pre-existing destroy pass, and a moduledoc line. Neither is blocking.

One caveat on my side: I reviewed by reading the code paths and didn't execute the suite in my worktree, so I'm taking the PR's test run at its word.

PR #12006, opened 2026-07-22
perf(server): preload Kura manifest entitlements

pepicrft:
## What changed

The Kura reconciler now loads active and trialing subscriptions once, selects the newest subscription for each account with deterministic timestamp ordering, and passes that lookup into manifest generation.

Manifest generation resolves the account plan once and derives all plan-dependent fields from that result. A configured egress floor is applied to entitled plans, while non-entitled plans retain an explicit zero value.

## Why

A reconciliation pass can inspect many accounts. Resolving the plan separately for every observed server made the number of database queries grow with the account count.

## Root cause

Manifest generation called the single-account entitlement path from inside the observed-server loop. The initial preload also depended on database row order and used the configured egress floor without first applying plan entitlement.

## Approach

Load relevant subscriptions in one query before constructing manifests, reduce them to the newest active or trialing subscription per account with an explicit chronological comparison, and resolve each account from the in-memory lookup. Skip billing resolution entirely when the manifest has no plan-dependent fields.

The existing single-account entitlement path remains available to other callers.

## Impact

Kura reconciliation performs a bounded number of billing queries instead of one query per account. Month-boundary ordering is deterministic, and configured egress values cannot grant capacity to a plan that is not entitled to it.

## Validation

- `mix test test/tuist/billing/entitlements_test.exs test/tuist/kura/provisioner/kubernetes_controller_test.exs test/tuist/kura/reconciler_test.exs` with 102 tests passed
- `mix credo --strict` on all six changed Elixir source and test files
- `mix format --check-formatted` on every changed Elixir source and test file
- `git diff --check`

## Review order

Final pull request in the server stack, after [#12005](https://github.com/tuist/tuist/pull/12005).

### DIFF
--- a/server/lib/tuist/billing/entitlements.ex
+++ b/server/lib/tuist/billing/entitlements.ex
@@ -66,7 +66,7 @@ defmodule Tuist.Billing.Entitlements do
     |> Enum.filter(&(&1.status in ["active", "trialing"]))
     |> case do
       [] -> :air
-      active -> active |> Enum.max_by(&{&1.inserted_at, &1.id}) |> Map.fetch!(:plan)
+      active -> active |> latest_subscription() |> Map.fetch!(:plan)
     end
   end
 
@@ -78,4 +78,14 @@ defmodule Tuist.Billing.Entitlements do
   end
 
   defp current_plan(_), do: :air
+
+  defp latest_subscription([subscription | subscriptions]) do
+    Enum.reduce(subscriptions, subscription, fn candidate, latest ->
+      case NaiveDateTime.compare(candidate.inserted_at, latest.inserted_at) do
+        :gt -> candidate
+        :lt -> latest
+        :eq -> if candidate.id > latest.id, do: candidate, else: latest
+      end
+    end)
+  end
 end
--- a/server/lib/tuist/billing/entitlements.ex
+++ b/server/lib/tuist/billing/entitlements.ex
@@ -20,7 +20,28 @@ defmodule Tuist.Billing.Entitlements do
   Returns true when the account's plan grants access to `feature`.
   """
   def allows?(account, feature) do
-    not Environment.tuist_hosted?() or plan_allows?(current_plan(account), feature)
+    MapSet.member?(allowed_features(account, [feature]), feature)
+  end
+
+  @doc """
+  Returns the requested features granted by the account's current plan.
+
+  The plan is resolved once for the complete feature set. When subscriptions
+  are preloaded on the account, the lookup stays in memory so batch callers do
+  not issue one subscription query per account.
+  """
+  def allowed_features(_account, []), do: MapSet.new()
+
+  def allowed_features(account, features) when is_list(features) do
+    if Environment.tuist_hosted?() do
+      plan = current_plan(account)
+
+      features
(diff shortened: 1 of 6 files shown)


server/lib/tuist/billing/entitlements.ex
```diff
@@ -40,6 +61,15 @@ defmodule Tuist.Billing.Entitlements do
 
   defp plan_allows?(_plan, _feature), do: false
 
+  defp current_plan(%Account{subscriptions: subscriptions}) when is_list(subscriptions) do
+    subscriptions
+    |> Enum.filter(&(&1.status in ["active", "trialing"]))
+    |> case do
+      [] -> :air
+      active -> active |> Enum.max_by(&{&1.inserted_at, &1.id}) |> Map.fetch!(:plan)
```
fortmarek: Non-blocking: `Enum.max_by/2` compares `%NaiveDateTime{}` structs using Erlang term ordering rather than chronological ordering. With rows such as Jan 31 and Feb 1, this can select the older plan, so the preloaded path may render stale entitlements after a plan transition. Could we preserve the query ordering (`List.first/1`) or use an explicit `NaiveDateTime.compare/2` comparator, with a month-boundary regression test?

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -649,12 +668,17 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   # is bursty, so non-enterprise tenants run best-effort under the Cilium burst
   # ceiling alone and pack densely. nil (dropped) when the region has no floor or
   # the account isn't entitled.
-  defp egress_guaranteed_mbps(account, %Regions{provisioner_config: %{egress_guaranteed_mbps: mbps}})
+  defp guaranteed_egress_configured?(%Regions{provisioner_config: %{egress_guaranteed_mbps: mbps}})
```
fortmarek: Non-blocking thought: could we model the effective guaranteed egress as `0` for Air and the configured floor for Enterprise? The Kura controller already skips the extended-resource request unless `mbps > 0`, so that may let this predicate and `egress_guaranteed_mbps/2` collapse into one value lookup. We would just want to preserve the current optimization that avoids resolving entitlements in regions with no configured floor; also, `0` is retained in the manifest rather than omitted.

fortmarek (approved):
Looks good overall. I left two non-blocking follow-ups; approving.

server/lib/tuist/billing/entitlements.ex
pepicrft: Addressed in `9c5765b583`. The preload now uses `NaiveDateTime.compare/2` with a deterministic identifier tie-break, and the regression covers an older January 31 plan followed by a newer February 1 plan.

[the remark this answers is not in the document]
pepicrft: Fixed in the current #12006 branch. Subscription selection now uses `NaiveDateTime.compare/2`, with the subscription identifier as a deterministic tie-breaker. The regression test covers an older January 31 Pro subscription and a newer February 1 Enterprise subscription. The combined focused server-stack suite passes with 114 tests.

PR #12007, opened 2026-07-22
fix(kura): bound runtime resource admission

pepicrft:
## What changed

Kura now coordinates memory, temporary-disk, multipart-upload, replication-backlog, snapshot, and file-descriptor admission through bounded runtime controllers.

Transient unreclaimable work shares one fair Tokio byte budget equal to the gap between the soft and hard memory watermarks. Sampled container memory now drives pressure state, cache trimming, and background load shedding only. It does not participate in per-request admission arithmetic.

Mapped-file serving remains a separate try-only optimization for already-resident reclaimable pages, with immediate fallback to streaming. Its capacity can no longer exceed the available watermark gap. File-cache advice is page-aligned and issued through Rustix.

Response permits are retained by the response body, replicated artifacts preserve the staging file-cache policy through persistence, and snapshot encoding reserves its full input, output, and compressor working set before materializing bytes. A staging reservation that is clamped below its desired working set always switches to bounded cache handling, including profiles where the calculated reservation reaches zero.

## Why

A container memory limit did not bound the aggregate working set of concurrent request and background paths. Under pressure, individually valid work could overlap and exhaust memory, temporary storage, descriptors, or replication capacity.

## Root cause

The previous memory controller added sampled `memory.current` to separately tracked reservations. Those values can describe the same bytes, so the sum did not establish a safe invariant. Independent materialization and mapped-file pools could also reserve more than the soft-to-hard headroom in aggregate, especially on small memory profiles.

Some ownership boundaries also ended too early: a handler could release response memory before its body was consumed, replicated staging policy was lost during persistence, clamped staging retained a foreground cache policy, and snapshot admission did not include the complete compression peak.

## Approach

Use owned semaphore permits as the source of truth for admitted transient bytes. Live permits cannot exceed `hard - soft`; waiting acquisition is fair; the permit follows the allocation or transfer that owns the bytes; and growth while already holding a permit is non-blocking.

Pressure is sampled as `memory.current - inactive_file`. It controls optional cache retention, background admission, and outbox pausing without changing the fixed permit capacity. The automatic soft watermark is 60% and the hard watermark remains 85%, leaving a deterministic 25% transient budget on the default profile.

The module split remains in this pull request because the controllers and admission paths cross the former monolithic files. Separating it later would recreate thousands of lines of mechanical movement in a follow-up diff.

## Impact

Kura remains responsive under concurrent pressure, keeps background work subordinate to foreground traffic, and reports explicit capacity failures instead of relying on container termination. The controller bounds admitted work while pressure management covers storage-engine memory, network buffers, and allocator fragmentation.

## Validation

- `cargo test --lib` with 417 tests passed and 1 declared scale test ignored
- `cargo fmt --all -- --check`
- `git diff --check`
- `mise run test-e2e -- spec/e2e/memory_pressure_spec.sh` with three 512-mebibyte containers, repeated concurrent 16-mebibyte uploads and downloads, zero restarts, and zero control-group out-of-memory, kill, or limit events

## Review order

First in the Kura runtime stack: [#12007](https://github.com/tuist/tuist/pull/12007), [#12008](https://github.com/tuist/tuist/pull/12008), [#12009](https://github.com/tuist/tuist/pull/12009).


kura/src/reapi/service.rs
```diff
@@ -0,0 +1,5481 @@
...
+            }
+        }
+
+        let mut response = Response::new(action_result);
+        let response_memory = materialization_budget
+            .into_inner()
+            .expect("action-cache materialization budget lock poisoned")
+            .into_response_guard();
+        if let Some(response_memory) = response_memory {
+            response.extensions_mut().insert(response_memory);
```
fortmarek, line 1263: **[P1] Keep materialization permits in the response body**

`ResponseMemoryGuard` is inserted into the Tonic response extensions here. Those extensions become part of the HTTP response head; Tonic maps the protobuf message into an encoded body, and Hyper separates the head from that body when it sends headers. The head/extensions can therefore be dropped while a slow client still retains the protobuf value and encoded output, returning these permits to the pool too early. Concurrent slow reads can then materialize more data than the pool is intended to allow. The same pattern appears in the snapshot response and `BatchReadBlobs`.

Please use the extension only as a handoff to `GrpcRequestAccountingLayer`: remove the guard there and make the mapped HTTP body own it until completion, cancellation, or drop. A regression test should receive a response without consuming its body, assert that another permit cannot be acquired, then drop the body and assert that the permit is released.

kura/src/store.rs
```diff
@@ -684,7 +910,7 @@ impl Store {
             trunk: None,
         };
         Ok(self
-            .persist_artifact_from_path_with_version(spec, source_path)
+            .persist_artifact_from_path_with_version(spec, source_path, FileCachePolicy::Adaptive)
```
fortmarek: **[P1] Preserve the staged file-cache policy for live replication**

The live internal replication route stages its request through `read_request_to_temp`, which deliberately selects `Bounded` for large or queued uploads and keeps the foreground reservation alive during persistence. This helper replaces that decision with `Adaptive`, so a replicated artifact larger than the 16 MiB working window stops dropping source and destination cache ranges during the segment append while pressure is still sampled as normal. A large peer write can therefore refill the cgroup page cache beyond the working set that was admitted.

Please accept `StagedArtifactPath` here, as the public persistence path already does, and pass `temp.file_cache_policy` from the live route. Bootstrap can explicitly construct an adaptive staged path if that behavior is intentional. Ideally `TempBodyFile::staged()` should return the borrowed staged path so the compiler also keeps its reservation owner alive across the persistence future.

fortmarek (commented):
I found three concrete gaps in the new resource bounds, plus one broader design concern about how the memory controls compose. `mise run test-unit` passes (410 passed, 1 ignored), but the current tests do not exercise the response-head/body lifetime or the full peak working set.

kura/src/reapi/service.rs
```diff
@@ -0,0 +1,5481 @@
...
+            .min(self.state.memory.reapi_materialization_pool_bytes() / 2);
+        if content_budget == 0 {
+            return Err(Status::resource_exhausted(
+                "action-cache snapshot encode declined under memory pressure",
+            ));
+        }
+        let _permit = self
+            .state
+            .memory
+            .try_acquire_reapi_materialization(content_budget)
```
fortmarek: **[P2] Reserve the snapshot encoder's peak live set, not only its input budget**

This permit covers only `content_budget`, but `encode_with_budget` holds the uncompressed body while `compress_snapshot` allocates a compressed `Vec` and then allocates the final `TSNZ` envelope and copies the compressed bytes into it. At the 64 MiB content / 48 MiB wire limits, one encoder can therefore briefly retain roughly 64 + 48 + 48 MiB plus zstd and collection scratch while accounting for only 64 MiB; the loop can also retain the preceding `wire` while evaluating a retry. Two encoders can both acquire half of the 128 MiB pool and substantially exceed it.

Please reserve the calculated peak input + output + bounded scratch, write compressed data directly into the final envelope (with a size-limited writer), and drop each failed attempt before retrying. The resulting lease should be retained or shrunk and handed to the response-body guard rather than dropped and reacquired after the output already exists.

kura/src/memory/mod.rs
```diff
@@ -1,126 +1,107 @@
...
+use pools::MemoryPools;
+use pressure::transition;
+use reservation::{AdmissionClass, FOREGROUND_ADMISSION_TIMEOUT, ForegroundWaiter};
 
 #[derive(Clone)]
 pub struct MemoryController {
     inner: Arc<MemoryControllerInner>,
 }
 
 struct MemoryControllerInner {
```
fortmarek, line 44: **Design concern: this duplicates the system memory controller without establishing one enforceable accounting invariant**

This combines sampled cgroup usage, application-maintained transient reservations, a custom pressure/hysteresis state, independent semaphore pools, and file-cache policies. The composition is difficult to audit: `memory.current` includes Rust heap, RocksDB/C++, mmaps, and page cache, while reservations cover selected operations only; once a leased allocation reaches the next sample, `current_bytes + transient_reserved_bytes` counts it twice, whereas allocations and dirty pages created before that sample are undercounted. The behavior therefore depends on sampling cadence even though the cgroup remains the real enforcement boundary.

We still need Kura-specific cost estimates, but the enforcement mechanics already exist: cgroup v2 provides `memory.high`, `memory.max`, `memory.events`, and PSI ([kernel docs](https://www.kernel.org/doc/html/latest/admin-guide/cgroup-v2.html)); Tokio provides fair, cancellable, byte-weighted RAII permits through [`acquire_many_owned`](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html); and file advice is available safely through [`rustix::fs::fadvise`](https://docs.rs/rustix/latest/rustix/fs/index.html), with the sync/alignment limitations documented by [`posix_fadvise(2)`](https://man7.org/linux/man-pages/man2/posix_fadvise.2.html).

Could we narrow the custom layer to (1) a thin byte budget over Tokio permits, (2) permits owned by the actual allocation/body, and (3) one streaming-file abstraction responsible for sync and `fadvise`, while using cgroup/jemalloc observations for metrics and coarse load shedding? If this state machine remains, I think the PR needs an explicit accounting invariant and a cgroup-limited stress test demonstrating that sampled usage, outstanding reservations, RocksDB allocations, mmaps, and dirty page cache do not create either an admission gap or systematic double-counting.

pepicrft: You were right. Adding sampled `memory.current` to separately tracked reservations did not establish an invariant because both values can describe the same bytes. The independent materialization and mapped-file pools could also exceed the soft-to-hard headroom in aggregate.

I replaced that arithmetic with one fair Tokio byte budget for transient unreclaimable work. Live owned permits can never exceed `hard - soft`; the permit follows the allocation or transfer; waiting acquisition is fair; and growth while holding a permit is non-blocking. Sampled memory no longer enters admission. Pressure now uses `memory.current - inactive_file` only for cache trimming, background gating, outbox pausing, and metrics.

I kept mapped-file serving as a separate try-only sublimit because those pages are already resident and reclaimable, and the path always falls back to streaming. I removed its 64-mebibyte floor, which could exceed the full headroom on small profiles. File advice now goes through the existing `IoController` boundary, expands ranges to complete pages, and uses the safe Rustix wrapper. The synchronous control-group read from the upload guard destructor is gone.

The automatic soft watermark is now 60%, with the hard watermark at 85%, so the default profile has a deterministic 25% transient budget. This bounds admitted work rather than claiming to bound storage-engine memory, network buffers, or allocator fragmentation; sampled pressure manages those separately.

I also rebased the response-stream follow-up and removed its sampled-memory ledger and delayed-release handoff. Response permits still follow the body until the final transport owner drops, but all response sublimits sit inside the same transient budget.

Validation passed with 413 unit tests on this branch, linting, formatting, and a three-node container pressure test using 512-mebibyte limits with repeated concurrent 16-mebibyte uploads and downloads. The containers recorded zero restarts and zero control-group out-of-memory, kill, or limit events.

kura/src/file_cache.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Reserves the maximum source-plus-destination file-cache working set for a
+/// disk-backed foreground upload. Objects larger than one working window keep
+/// both staging and segment-copy cache ranges bounded while this guard is held.
+pub(crate) async fn reserve_foreground_staging(
+    memory: &MemoryController,
+    declared_or_max_bytes: u64,
+) -> Result<ForegroundFileCacheReservation, ForegroundAdmissionTimeout> {
+    let working_set_bytes = declared_or_max_bytes
+        .min(FOREGROUND_STAGING_WINDOW_BYTES)
+        .min(memory.transient_capacity_bytes().saturating_div(2));
```
fortmarek: **[P1] Switch to bounded cache handling when this reservation is clamped**

Capping `working_set_bytes` by half the transient capacity can make the permit smaller than the source-plus-destination working set, but the policy below still returns `Foreground` whenever the upload did not wait and is at most 16 MiB. While it is the only reservation, `should_drop` stays false, so a 16 MiB upload with an 8 MiB transient budget can retain roughly 32 MiB of staging and segment pages while accounting for only 8 MiB. With a valid one-byte soft-to-hard gap, the division yields a zero-byte reservation and concurrent small uploads bypass the semaphore entirely. Please select `Bounded` whenever the desired reservation was clamped, or reject profiles whose headroom cannot support the minimum working window.

fortmarek (approved):
Approving the revised controller direction: replacing sampled-memory admission arithmetic with one fair Tokio byte budget, and using sampled cgroup working-set data only for pressure management, addresses the original architectural concern.

The remaining findings are not waived by this approval and should still be addressed:

- Response materialization permits need to follow the encoded HTTP body through completion or cancellation, rather than remain in response extensions.
- Live replicated persistence needs to preserve the staged artifact's bounded file-cache policy.
- Snapshot encoding needs a reservation for its peak live working set, with the resulting permit handed to the response body.
- A foreground staging reservation that is reduced to fit a small transient pool must switch to bounded cache handling (or the configuration must be rejected), so the admitted permit continues to cover the actual working window.

Validation at this head is clean: 413 unit tests passed with one ignored scale test, plus Clippy, formatting, and `git diff --check`.

kura/src/store.rs
pepicrft: Addressed in `f02b78bd86`. Live replication now passes the staged artifact and its selected file-cache policy through persistence while the temporary-file owner remains alive. Bootstrap constructs a bounded staged path explicitly. The regression observes the bounded segment-copy cache drop for a payload larger than the working window.

PR #12009, opened 2026-07-22
fix(kura): bound response streams under memory pressure

pepicrft:
## What changed

Kura now keeps memory permits attached to encoded response bytes until the final transport owner releases them. This covers accelerated file reads, mapped-file reads, direct key-value downloads, Remote Execution unary and batch responses, ByteStream responses, and ordinary public downloads.

Normal readers use weighted sublimits inside the shared transient budget introduced in [#12007](https://github.com/tuist/tuist/pull/12007). When full-size admission is exhausted, eligible reads use a tightly bounded degraded stream. If neither path has capacity, Kura returns a retryable unavailable response.

The accelerated zero-copy path reserves its bounded transfer window before writing `200 OK`. If admission cannot complete, it writes a retryable `503 Service Unavailable` response without starting the artifact transfer.

Container memory, response-pool, and pressure metrics expose the retained transport working set without feeding sampled usage back into admission arithmetic.

## Why

The runtime foundation bounds allocations made by handlers, but response buffers can outlive the handler that creates them. Their permit must follow the bytes into the transport.

## Root cause

Response memory was released when the handler returned even though encoded chunks could still be queued by the transport. Many slow clients could therefore retain unaccounted buffers. Direct key-value downloads and several unary Remote Execution responses bypassed the initial response guard, while the accelerated path could commit a successful response before reserving transfer capacity.

The first version of this branch attempted to bridge that lifetime with sampled-memory accounting and delayed releases, which recreated the double-counting concern fixed in the foundation.

## Approach

Acquire from the single fair transient budget before opening a reader or returning a materialized response, then attach the owned permit to response bodies and encoded messages. Weighted response semaphores are concurrency sublimits only; they do not create another memory budget. The permit releases immediately when the last transport object drops, while sampled working-set pressure remains responsible for cache trimming and coarse load shedding.

A bounded degraded pool preserves useful read throughput without bypassing transient capacity. Background bootstrap reads do not queue and cannot bypass public waiters. The accelerator reserves its transfer window before sending headers and returns a retryable response if admission is exhausted.

## Impact

Heavy read pressure degrades predictably instead of exhausting the container. Clients receive either a bounded stream or an explicit retry signal, while normal-pressure throughput keeps the larger response path.

## Validation

- `cargo test --lib` with 446 tests passed and 1 declared scale test ignored
- Focused regressions prove direct key-value and unary response reservations follow encoded transport bytes
- A focused accelerator regression exhausts response admission and proves the wire contains `503 Service Unavailable`, never `200 OK`
- `cargo fmt --all -- --check`
- `git diff --check`
- `mise run test-e2e -- spec/e2e/memory_pressure_spec.sh` with three 512-mebibyte containers and zero pressure-test failures or out-of-memory events

## Review order

Final pull request in the Kura runtime stack, after [#12008](https://github.com/tuist/tuist/pull/12008).


kura/src/http.rs
```diff
@@ -146,9 +154,72 @@ pub fn combined_router(state: SharedState) -> Router {
             state.clone(),
             track_http_metrics,
         ))
+        .layer(middleware::map_response(guard_response_stream_transport))
         .with_state(state)
 }
 
+pub(crate) async fn guard_response_stream_transport(mut response: Response) -> Response {
+    let Some(guard) = response.extensions_mut().remove::<ResponseTransportGuard>() else {
```
fortmarek, line 162: [P1] Direct key-value downloads still bypass response accounting. This middleware only acts when a handler has already attached `ResponseTransportGuard`; `get_keyvalue` clones and returns its inline `Vec<u8>` directly. Concurrent slow key-value reads therefore remain outside both full and degraded pools. Please admit before cloning and attach the permit, or route this endpoint through the guarded artifact-serving path.

kura/src/reapi/service.rs
```diff
@@ -127,6 +129,9 @@ pub fn routes(state: SharedState) -> axum::Router {
             admit_grpc_write_decode,
         ))
         .layer(GrpcRequestAccountingLayer { state })
+        .layer(axum::middleware::map_response(
```
fortmarek, line 132: [P1] Several large unary gRPC replies still reach this middleware without a guard. `UpdateActionResult` can echo a 4 MiB result, and `FindMissingBlobs`/`BatchUpdateBlobs` can produce request-sized responses, but none attaches `ResponseTransportGuard`. Their request admission can be released before Tonic finishes encoding and sending. Please reserve from the materialization pool using the encoded size and attach the guard before returning.

kura/src/accelerated_file_serving.rs
```diff
@@ -588,15 +591,26 @@ async fn serve_accelerated(
             // before the body transfer, so large downloads do not inflate the
             // responsiveness signal.
             let time_to_first_byte = request_started_at.elapsed();
-            let bytes = transfer_file(&mut stream, &file, mode, chunk_bytes)?;
-            Ok((stream, bytes, time_to_first_byte))
+            let mut cache_drop = AcceleratedReadCacheDrop::new(chunk_bytes);
```
fortmarek, line 629: [P1] The default plaintext accelerator starts writing `200 OK` and transferring without acquiring any response-stream reservation. Cache advice only drops completed file ranges; it cannot account for or reject socket, pipe, and active page-cache memory retained by slow clients when live headroom is exhausted. Please acquire a bounded accelerator transport reservation before consuming/writing the response, and fall back or return a retryable 503 when admission fails.

fortmarek (approved):
Approved with three findings from the response-memory review. These identify response paths that still sit outside the new admission and transport-lifetime accounting.

kura/src/reapi/service.rs
pepicrft: Addressed in `a826ff77af`. `UpdateActionResult`, `FindMissingBlobs`, and `BatchUpdateBlobs` now reserve from their encoded size and attach the transport guard before returning. The route regression proves an encoded unary response retains the reservation until its final byte owner drops.

kura/src/http.rs
pepicrft: Addressed in `a826ff77af`. Direct key-value downloads now reserve content plus transport materialization before returning inline bytes, attach the transport guard, and return a retryable unavailable response when admission fails. The regression proves the reservation remains until the response transport drops.

PR #12015, opened 2026-07-22
feat(cli): strip release binaries

Loupehope:
### What changed

The CLI bundling task now strips the Tuist executable, ProjectDescription framework, CAS plugin, and CAS proxy before signing and notarization.

The `-rSTx` parameters are based on the [Emerge Tools Strip Binary Symbols (iOS)](https://docs.emergetools.com/docs/strip-binary-symbols). The stripping step runs after runtime-library setup, including install_name_tool changes, and before codesigning so the resulting signatures remain valid. ProjectDescription.dSYM remains bundled for crash symbolication.

### Why it changed
Release artifacts retain unnecessary local, debug, and Swift symbols. Stripping them reduces the distributed size without changing the release build configuration.

| Binary | Original | Stripped | Reduction |
| --- | ---: | ---: | ---: |
| tuist | 232.56 MB | 93.94 MB | 138.62 MB (59.61%) |
| ProjectDescription | 10.99 MB | 2.97 MB | 8.02 MB (72.96%) |
| libtuist_cas_plugin.dylib | 1.27 MB | 1.05 MB | 0.22 MB (17.36%) |
| tuist-cas-proxy | 14.86 MB | 12.94 MB | 1.93 MB (12.96%) |

### DIFF
--- a/mise/tasks/cli/bundle.sh
+++ b/mise/tasks/cli/bundle.sh
@@ -97,6 +97,14 @@ bundle_swift_runtime_libraries() {
     fi
 }
 
+strip_release_binaries() {
+    strip -rSTx \
+        "$BUILD_DIRECTORY/tuist" \
+        "$BUILD_DIRECTORY/ProjectDescription.framework/ProjectDescription" \
+        "$BUILD_DIRECTORY/libtuist_cas_plugin.dylib" \
+        "$BUILD_DIRECTORY/tuist-cas-proxy"
+}
+
 # Builds the cas-plugin (Xcode compilation-cache CAS plugin) as a universal
 # dylib, plus the per-machine proxy binary, both bundled next to `tuist`. The
 # CLI points Xcode's compilation caching at the dylib via
@@ -142,6 +150,9 @@ build_cas_plugin
 
 echo "$(format_subsection "Bundling Swift runtime libraries")"
 bundle_swift_runtime_libraries
+
+echo "$(format_subsection "Stripping release binaries")"
+strip_release_binaries
 
 echo "$(format_section "Copying assets")"
 


Loupehope:
Thanks for tuist! @fortmarek @pepicrft
I hope the pull request will be considered.

pepicrft:
Thanks @Loupehope, and thanks for opening this PR :). Just one minor thing: `strip -rSTx` removes symbols that `xcrun swift` needs from ProjectDescription, causing `tuist generate` to fail on both Apple silicon and Intel. Please leave the ProjectDescription binary unstripped.

Loupehope:
> Thanks @Loupehope, and thanks for opening this PR :). Just one minor thing: `strip -rSTx` removes symbols that `xcrun swift` needs from ProjectDescription, causing `tuist generate` to fail on both Apple silicon and Intel. Please leave the ProjectDescription binary unstripped.

Thank you for raising this. I tried to reproduce the issue using the stripped artifact https://github.com/Loupehope/tuist/releases/download/1.0.0/tuist-strip-probe.zip, which includes a ProjectDescription.framework processed with strip -rSTx.

Running ./run.sh executes tuist install and tuist generate against a minimal project that uses swift-collections as an external Swift Package dependency.

Both the arm64 and x86_64 slices complete successfully. For the x86_64 check on Apple Silicon, only the xcodebuild subprocess runs natively under arm64 because the local Xcode installation cannot load its CoreSimulator framework through Rosetta. The tuist binary and ProjectDescription framework still run as x86_64.

Note: The locally stripped test artifact may require explicit approval in macOS System Settings > Privacy & Security before it can be run, due to Gatekeeper security checks.

Could there be a missing detail in the failing setup, such as a different artifact, a particular manifest configuration, or an additional runtime dependency?

pepicrft:
You’re right, thanks for the thorough reproduction. My test used a mise-installed framework whose symlinks had been flattened. Stripping invalidated its signature, and my attempted re-signing failed. I mistakenly attributed the resulting load failure to missing symbols. I reran this with your artifact and clean signed and unsigned controls, and both architectures work correctly. Please disregard my original concern.

PR #12022, opened 2026-07-22
feat(server): consolidate runner job lifecycle state into Postgres behind kill switches

fortmarek:
## What

Track A of the runner-state consolidation, as one PR: workflow_job lifecycle state gets a Postgres control-plane store (`runner_workflow_jobs`), dark-written alongside today's ClickHouse paths, with the dispatch reads available behind a kill switch and a transactional outbox that can later replace the direct ClickHouse writes. ClickHouse `runner_jobs` becomes analytics-only downstream once the flags flip and bake.

- **Lifecycle table (dark writes).** One `runner_workflow_jobs` row per workflow_job (`workflow_job_id` PK) carrying status (`queued → claimed → running → completed | cancelled`), the full dispatch-candidate metadata (repository, platform, shape, workflow correlation fields, `requested_dispatch_label`), lifecycle timestamps, and the pod/runner binding. Indexed on `(fleet_name, status, enqueued_at)` for the dispatch read. Every mutation is a guarded compare-and-set (`UPDATE … WHERE status = <expected>`, or an upsert whose `ON CONFLICT DO UPDATE` carries the guard), so webhook redeliveries and claim races cannot regress a row.
- **Claim-path transitions.** `Claims.attempt/5` transitions `queued → claimed` inside the same transaction as the claim insert; `mark_running/2` does `claimed → running`; `release/2` and `release_pod_missing/2` move rows back to `queued` atomically with the claim delete, leaving terminal rows alone.
- **Drift comparator (log-only).** `JobStateDriftWorker` (5-min cron) diffs PG status against the ClickHouse argMax state for recently-updated rows and emits telemetry/logs only (`status_mismatch`, `missing_in_clickhouse`, plus a `compared` volume counter, surfaced as `tuist_runners_workflow_job_drift_count`). The comparison is conclusion-aware: CH's `completed` + conclusion pair folds into the Postgres status space first, so a cancelled/completed disagreement between the stores registers as drift. It mutates nothing.
- **Flag-gated Postgres reads.** PG twins of `pick_queued_top_k/5`, `queued_count_by_fleet/1`, and `queued_count_by_fleet_and_account/1` (the dispatchable-demand read #11981 added) behind one kill switch, `:runner_dispatch_postgres_reads`, default OFF. The ClickHouse paths remain the default and the instant, no-deploy rollback.
- **ClickHouse write outbox.** Each applied transition inserts a `runner_workflow_job_transition_events` row in the same transaction, carrying the exact `runner_jobs` insert shape with the transition's own microsecond `updated_at` as the RMT version. Rows carry an `account_id` FK that cascade-deletes with the account, so no payload outlives its tenant. `FlushJobTransitionEventsWorker` (1-min cron) drains it in `FOR UPDATE SKIP LOCKED` batches: CH insert, then delete, per transaction. Behind its own flag `:runner_job_transition_outbox`, default OFF. Direct CH writes stay untouched; dual emission is harmless because the ReplacingMergeTree resolves overlapping rows by version.

## Why / root cause

Runner job state lives in three stores with different consistency models: ClickHouse `runner_jobs` (a ReplacingMergeTree where "current status" is an argMax over transition INSERTs — also the dispatch queue read), Postgres `runner_claims` (the OLTP claim lock, with several release paths), and Kubernetes Pods. No transition spans stores atomically, so every crash window between a CH write and a PG write is a divergence class we've paid for repeatedly: the CH-first ordering contracts documented in `stale_claims_worker.ex` / `list_pods_missing_since`, the anti-list dispatch needs because CH can say `queued` while PG holds a live claim, the `enqueued_at` lookback floor to keep the argMax dedup from tripping ClickHouse memory limits on the hot path, and a family of recovery workers reconciling the stores after the fact.

A single Postgres row per job with CAS transitions makes the claim and its state change one transaction — the CH-first crash-ordering constraint documented in the recovery workers does not apply to the PG row, because claim + state commit or roll back together.

## Why one PR with kill switches

Merge-safety comes from runtime flags, not PR splitting: everything here is dark by default. The dark writes are behavioral no-ops for dispatch (CH stays authoritative), both flags default OFF, and the comparator is read-only. Landing it as one unit keeps the schema, the writers, and the reads consistent with each other, and the rollback story for the risky part (the dispatch read) is flipping a flag off, not reverting a PR chain.

## Placement decisions worth knowing

- **Terminal transitions hook `Jobs.complete_locked`/`record_completed_locked`** (next to `persist_completion!`) rather than only dispatch.ex: that choke point also covers `StaleQueuedJobsWorker` and `OrphanedRunnersWorker` force-completes, which would otherwise leave PG rows non-terminal forever — systematic drift, and stale claimable rows if the read flag flipped.
- **Re-queue transitions live inside `Claims.release/2` / `release_pod_missing/2`**, not in `Jobs.record_queued`: every recovery path (stale claims, orphaned runners, pod reconciliation, `release_safely`) funnels through those deletes, and hooking `record_queued` would let a racing worker clobber a fresh claim's `claimed` status — the CAS guard plus claim-delete atomicity avoids reintroducing the race the CH path tolerates.
- **`cancelled` is a first-class PG status** (CH models it as `completed` + conclusion). The drift comparator folds the CH pair into the same status space before comparing, and the outbox maps it back to the CH shape on replay.
- **Deadlock safety:** every transaction that touches both tables locks `runner_claims` before `runner_workflow_jobs` (attempt, mark_running, both releases, and the completed-webhook path), so there is no lock-order cycle.

## Prior work / adjacent PRs

- A search for an earlier lifecycle-table PR (`gh pr list --author "@me" --search "lifecycle"` and recent branches) found none, so this is not folding in prior commits; the merged seed `runner_job_completions` (migration `20260715120000`) is the model for the idempotency guards.
- #11981 (dispatchable demand) and #11988 (pod↔claim reconciliation) both merged before this branch was cut; their surfaces are accounted for — `queued_count_by_fleet_and_account` has a PG twin behind the same flag, and `release_pod_missing` carries the lifecycle re-queue.

## Out of scope (post-bake follow-ups)

- Deleting the CH dispatch read path (anti-list, lookback floor, head-blocking logic).
- Removing the direct CH writes (the outbox becomes the sole writer only after the drift comparator holds at zero).
- Reconciler/sweep changes (`release_by_pod_name` intentionally leaves the lifecycle row alone, mirroring today's "GitHub re-queues it" semantics).

## Validation

- `mix test test/tuist/runners/ test/tuist/runners_test.exs test/tuist/oban/runtime_config_test.exs` — 562 tests, all passing (includes 26 new context tests, worker tests for the comparator and flusher, claim-transition tests, and flag-gated read tests).
- `mix excellent_migrations.check_safety` — zero warnings from the two new migrations (new-table indexes are safety-assured, matching the seed migration).
- `mix credo` on all changed files — no issues.
- `mise run format` run over the tree.
- `server/data-export.md` updated for both new tables.

## Review follow-ups applied

- Removed the unused `WorkflowJobs.get/1` (unscoped tenant lookup by `workflow_job_id` alone).
- Drift comparator now reads `argMax(conclusion)` alongside `argMax(status)` so PG `cancelled` only matches CH `completed` when the CH conclusion is `cancelled`, with a regression test for a cross-store cancelled/completed disagreement.
- Outbox rows gained an `account_id` FK with cascade delete, so account deletion cannot leave payloads behind for a later replay.


fortmarek:
Superseded by the four-PR stack that splits this change into reviewable parts: #12028 (lifecycle table + dark writes) → #12029 (drift comparator) → #12030 (flag-gated Postgres dispatch reads) → #12031 (CH transition outbox). The stack recomposes byte-identically to this branch's change-set, including the review fixes (scoped-lookup removal, conclusion-aware comparator, outbox account FK) — only the migration timestamps moved, because main claimed 20260722120000 in the meantime.

PR #12024, opened 2026-07-22
feat(kura): health-gated progressive rollout of Kura runtime versions (spec #79)

fortmarek:
Implements [spec #79](https://hive.tuist.dev/specs/79): a new Kura runtime tag no longer reaches the entire managed fleet in a single reconciler tick with only the public `/up` probe as its health authority — it rolls out in ordered, account-grouped waves gated on the health signals Kura already exposes on `/status/rollout`, with an expedited mode preserving today's all-at-once behavior for rollbacks and incident fixes.

## What changed

**Kura runtime (Rust).** `/status/rollout` gains `peer_connection_failure_count` (exported as `kura_peer_connection_failures_total`), incremented when an outbox replication delivery or bootstrap run errors against a peer. This is the spec's naming correction: `fd_timeout_count` counts fd-wait timeouts (fd-pool pressure), not peer failures — the gate now consumes both under their real names.

**KuraInstance controller (Go).** The controller already polled `/status/rollout` per pod each reconcile for primary routing; it now publishes a `status.rolloutHealth` aggregate with explicit per-field semantics (no single worst-of rule reproduces the standalone gate): ready/serving as conjunctions across expected replicas, an all-consistent generation flag, summed counters with per-pod reset clamping, max memory pressure, and the *oldest* per-pod sample timestamp so a frozen report reads as stale. Polling is one pass per reconcile shared by primary selection and the aggregate, now including not-ready pods — a sick standby drags the aggregate down by design. Additive CRD change, applied automatically on deploy.

**Server (Elixir).** New `Tuist.Kura.Rollouts` context, on by default with a `kura_rollout_orchestration_kill_switch` flag as the no-deploy fallback:

- Durable rollout state (`kura_rollouts` + wave assignments + per-server scope + audit events) advanced one reconciler tick at a time; a BEAM death mid-rollout resumes from the same rows next tick.
- Waves: 0 = Tuist-owned canary accounts only, then ~5% / ~25% / remainder of accounts by recent Kura usage ascending. Runner-cache servers follow their account's wave. A superseding rollout orders accounts still on the oldest image first after the canary.
- Gate: converged on target + ready + serving + generation-consistent + no bootstrap in flight + fresh sample + counters within baseline (outbox +10% with a small floor; no growth in fd timeouts or peer failures), evaluated over *every* server updated so far so a slow-burn canary regression stops wave 2. Soak 15m (canary) / 5m; failing ticks reset the clock; 60m wave deadline pauses. Hard signals pause immediately: terminal deployment failure, critical memory pressure, regression to failed after convergence.
- Convergence and soak eligibility are separate: baseline-unhealthy servers are excluded from the comparative soak but must still observe the target image — the gate never stands between a broken fleet and its fix.
- Scheduling is at-most-once per rollout *attempt*, deliberately replacing the documented per-server-per-tag-lifetime invariant: resume cancels open rows, mints fresh deployments for every non-converged server (including previously failed ones), re-captures baselines, and re-enters the gate.
- Same-tag redeploys never touch an active or paused rollout; a tag change supersedes it and cancels its open deployments. Expedite is explicit only (deploy input pinned to the exact tag, or ops-view action) and every operator verb records actor, reason, source/target tags, and whether the target previously completed.
- Servers created mid-rollout inherit their account's wave state (baseline until the wave completes, target after), so a paused-as-suspect version never reaches fresh servers — most important for runner-cache servers, which take traffic immediately.
- Ops surface ships in the first version: `/ops/kura` shows the rollout, wave progress, pause/resume/expedite/abort with reasons, the audit trail, and history.
- Only production paces; canary/staging mint expedited rollouts (today's fan-out). `TUIST_KURA_ROLLOUT_PACING` overrides the default for the staging progressive-mode drills the spec's rollout plan requires.
- Orchestration is on by default (the flag started as an opt-in soak gate and was inverted after the staging drills below): enabling the `kura_rollout_orchestration_kill_switch` flag in /ops/flags falls back instantly to interim pacing (spec phase 2) — bounded batches, Tuist-owned accounts first, next batch only when no deployment for the tag is still open. That kill-switch is the no-deploy rollback path for the machinery itself.

**Observability.** `Tuist.Kura.PromExPlugin` exports rollout state (active/paused/wave/wave age/converged-vs-scoped; `rollout_active` is emitted every poll so absence of the series is meaningful) and the fleet version distribution. A git-synced Grafana dashboard (`tuist-kura-rollout`) joins those with the per-node gate signals; its text panel documents the three alert rules to provision in Grafana Cloud (paused-sustained page, active-too-long warn, metrics-absent-while-recently-active page — the failure mode in-band alerting structurally misses). Slack notifications are best-effort fire-and-forget context via `TUIST_OPS_SLACK_WEBHOOK_URL`.

**Pipeline.** The production cascade gains a `kura-canary-rollout-gate` job: it resolves the same `kura@` tag the canary deploy pinned and polls `GET /api/kura/rollout-status` on canary until that rollout completes; paused/aborted/superseded block promotion. If the kill-switch ever disables orchestration on canary the gate passes with a notice rather than blocking deploys. The hotfix fast-path skips the gate deliberately. `server-deployment.yml` gains the explicit `kura_expedite` input.

## Why this shape

- **Grafana is the paging authority** because the reconciler cannot report its own death: a control plane that dies mid-rollout raises no paused signal, it stops reporting. The deploy job finishes when intent is recorded; convergence is watched by alerting, not CI (blocking the prod deploy job on the production rollout was explicitly rejected in the spec — it would hold a CI job open for an hour, tie rollout lifetime to job timeouts, and turn unrelated merges red while a rollout sits paused).
- **Pause instead of auto-rollback**: rolling back stays an operator decision (re-pin previous tag + expedite, which carries no new risk for a tag that already completed fleet-wide).
- **Baselines instead of absolute thresholds**: the question is "did the new version regress this server", not "is this mesh busy" — absolute thresholds would misfire on naturally busy meshes.

## Validation

- Rust: `cargo check --all-targets`, `cargo clippy --all-targets -- -D warnings`, `cargo fmt --check`, and the rollout-status/metrics unit tests (the route test now covers the new field).
- Go: `go build`, `go vet`, `gofmt`, full controller suite including new aggregate tests (per-field semantics, reset clamping across a simulated pod restart, last-known-sample staleness, unsampled-expected-pod conjunction failure).
- Elixir: `mix compile --warnings-as-errors`, `mix credo` (clean), `mix excellent_migrations.check_safety`, and 300+ tests across the kura domain: 18 new rollout tests (expedited completion, same-tag no-op, supersede + open-deployment cancellation, canary-first scheduling, soak completion, gate reset + deadline pause, critical-pressure hard pause, baseline-unhealthy soak exclusion, resume-as-reattempt, expedite audit, abort, wave sizing, mid-rollout inheritance), plus controller endpoint and ops LiveView tests, with all existing kura/runner/ops suites green.
- Helm values parse and the chart templates past the values layer; the expedite-tag env block mirrors the adjacent runtime-tag block.

### End-to-end on staging (flag on, progressive pacing forced)

The machinery was exercised live on the staging fleet (15 servers across 12 accounts) through a series of real deploys from this branch:

- **Three clean progressive completions** (three consecutive tag flips). Each ran the full path: wave assignment by recent usage (wave 0 = 3 Tuist-owned canary servers, then 1 / 3 / 8 customer-account servers), per-wave scoping as waves opened, observed convergence, the health gate reading the controller's `status.rolloutHealth` aggregate, 15-minute canary soak and 5-minute later-wave soaks, and completion — no pauses, no gate flaps. This meets the spec's drill bar of progressive mode through at least two real releases.
- **Hard pause on a real failure, root-caused from the pause record alone.** The first drill rollout hard-paused with `deployment_failed` within one tick: every KuraInstance apply was rejected with `.spec.meshPublicPeerPublished: field not declared in schema` — main had shipped that spec write ahead of its CRD (#12004, later guarded by #12034), and this branch's redeploys re-applied the older CRD onto CRs already carrying the field. The rollout stopped instead of retry-hammering, fresh servers stayed pinned to the baseline, and the pause event named the failing server, deployment, and apiserver error. Merging main and redeploying superseded the paused rollout (open deployments cancelled) and the successor completed cleanly.
- **Live pause/resume drill.** A running rollout was paused mid-wave-2 convergence (0/3 converged) via `Rollouts.pause/3` with actor and reason recorded. Four minutes paused: convergence marking frozen, wave 3 never scoped, nothing new minted — while already-applied deployments drained to terminal at the Kubernetes level (in-flight work completes; scheduling stops). `Rollouts.resume/3` then re-attempted: all three non-converged wave-2 rollout servers bumped to `attempt=1` with fresh pending deployments and re-captured baselines (verified at the row level), and the rollout converged wave 2, soaked, scheduled wave 3, and completed 15/15.
- **Supersede under contention.** The single-active invariant (partial unique index) held through repeated real supersede chains — tag changes over running rollouts, supersede of a paused rollout, and an accidental fight with a stale-env control plane (below) — without ever double-running a rollout or leaking open deployments.
- **Promotion-gate surface**: `GET /api/kura/rollout-status` tracked every lifecycle transition throughout (the payload the production gate polls).

The drills also exposed that the release runs a reconcile at application boot, so a `bin/tuist eval` Job that starts the full app ran a reconciler with the Job pod's env — which lacks the helm-injected `TUIST_KURA_RUNTIME_IMAGE_TAG` and fell back to a stale `runtime_image_tag` in the staging secrets blob, superseding the live rollout with a stale-tag one (the supersede machinery healed it each time). The reconciler now refuses to run without the helm-injected tag (`Environment.kura_control_plane?/0`), so non-control-plane boots are inert; purging the stale secrets-blob tag remains an ops follow-up. Adversarial review of the branch added further hardening in the same commit: expedited rollouts pause on critical memory pressure (the canary environment runs expedited and gates promotion), fresh servers match their account's mesh tag instead of regressing to the oldest baseline in supersede chains, ring consistency compares a fingerprint of the sorted member identities instead of ring sizes (with a size fallback for older runtimes), ring-skewed baselines are soak-ineligible instead of gating their own fix, and wave assignments are batch-inserted.

## Rollout of this change

Staging already runs this branch with `TUIST_KURA_ROLLOUT_PACING=progressive` (the spec's drill configuration), which is how the validation above was produced. Orchestration is on by default, so merging arms canary and production on their first post-deploy tick — no per-environment enablement step: canary starts minting expedited rollouts (arming the promotion gate), and production's first Kura tag change after the deploy runs the progressive wave machinery. The spec's opt-in flag phase was deliberately collapsed after the staging soak: the flag is now a kill-switch (`kura_rollout_orchestration_kill_switch` in /ops/flags), preserving the instant no-deploy fallback to interim pacing. Remaining step: provision the three Grafana Cloud alert rules (documented on the dashboard's text panel) and route them to the production escalation. The health publication is read-only status.


fortmarek:
screenshots

image
image

pepicrft (approved):
I might be missing a deployment detail here. For the first rollout, `baseline_image_tag` is nil. When the rollout is paused and an account has no running server, this falls back to `default_tag`, so a newly provisioned server gets the paused target. Is there a deployment-time baseline that ensures first-rollout accounts stay on the prior image? If not, should we persist one before creating the rollout?

esnunes:
starting review...

server/lib/tuist/kura/rollouts.ex
```diff
@@ -0,0 +1,1396 @@
...
+  # move), minting yields nil and `mint_missing_deployments/1` retries on
+  # a later tick once that deployment closes.
+  defp mint_deployment(rollout, server) do
+    case adoptable_open_deployment(server, rollout.image_tag) do
+      %Deployment{id: deployment_id} ->
+        deployment_id
+
+      nil ->
+        case Kura.create_deployment(server, rollout.image_tag, rollout_id: rollout.id) do
+          {:ok, deployment} ->
```
esnunes, line 848: I think there's a transaction-nesting problem here that's worth fixing before this lands, because the blast radius is the whole reconcile loop rather than just rollouts.

`advance/1` wraps the tick in `with_locked_rollout` → `Repo.transaction`. Inside it, `scope_server/3` → `mint_deployment/2` → `Kura.create_deployment/3`, which opens its own `Repo.transaction` and calls `Repo.rollback(:deployment_in_progress)` when the server already has an open deployment. Ecto joins the existing transaction rather than opening a savepoint unless `mode: :savepoint` is passed, so that inner rollback marks the connection aborted and the next statement in the outer transaction raises.

Reproduced against a real DB — a server left `:provisioning` with its open install deployment, then `Rollouts.sync()`:

    {:raised, DBConnection.ConnectionError, "transaction rolling back",
     [{Ecto.Adapters.SQL, :raise_sql_call_error, 1, ...},
      {Tuist.Kura.Rollouts, :scope_server, 3, [line: 765]},
      {Tuist.Kura.Rollouts, :advance_expedited, 1, [line: 576]}, ...]}

Two consequences:

1. The `{:error, :deployment_in_progress} -> nil` arm at rollouts.ex:785 is unreachable — the connection is dead before it returns. Same for the `nil` handling in `mint_missing_deployments/1` and `reattempt_non_converged/1`, whose comments explicitly plan for that return.

2. The trigger looks routine rather than exotic: a server created in tick N sits `:provisioning` with an open install deployment on the *baseline* tag (so `adoptable_open_deployment/2` doesn't match the target), and in tick N+1 `scope_servers` picks it up. Since `schedule_runtime_rollout()` runs first in `reconcile/0`, the raise means `reconcile_deployments()` and the observation projection never run — so nothing closes that install deployment, and the next tick does the same thing. That's the whole Kura control plane stopped, not just the rollout. `resume/3` and `expedite/3` hit the same path from the ops LiveView.

The existing tests can't catch it because `create_active_server/1` calls `close_open_deployments/1` in setup.

A few ways out, depending on what you prefer: a non-rollback variant of `create_deployment` that returns `{:error, reason}`; passing `mode: :savepoint` to the inner transaction; or hoisting deployment minting out of `with_locked_rollout` entirely (which would also help with the unbounded-work-under-lock point I raised separately).

server/lib/tuist/kura/rollouts.ex
```diff
@@ -0,0 +1,1396 @@
...
+
+    [
+      {health.memory_pressure_state >= 2, {:critical, :memory_pressure_critical}},
+      {not fresh_sample?(health), {:unhealthy, :sample_stale}},
+      {health.sampled_pods < health.expected_pods, {:unhealthy, :pods_unsampled}},
+      {not health.ready, {:unhealthy, :not_ready}},
+      {not health.serving, {:unhealthy, :not_serving}},
+      {not health.ring_consistent, {:unhealthy, :ring_skew}},
+      {health.backfilling_peers > 0, {:unhealthy, :backfill_in_flight}},
+      {counter_regressed?(health.outbox_messages, outbox_threshold), {:unhealthy, :outbox_regressed}},
```
esnunes, line 1205: The counter half of the gate is the part I'd most like your read on, because I don't think the comparison is well-defined across the restart the rollout itself causes.

`outbox_messages` is a gauge and gets a tolerance band (+10%, floor 50). `fd_timeout_count` and `peer_connection_failures` are process-lifetime cumulative counters and get `current > baseline` with no tolerance — so for those two the check reads as "has this happened at all since the baseline" rather than "is this worse than before".

One case seems unambiguous. On the first rollout onto an image that has this counter, the pre-upgrade pods run a runtime that doesn't emit `peer_connection_failure_count` at all, so `integer_field/2` defaults it to 0 and every scoped server's baseline is 0. The Rust side increments on every errored outbox delivery and every failed backfill pass, and a rollout restarts peer pods by definition — so one error after convergence trips `:peer_connection_failures_regressed`, the soak clock resets every tick, and the wave runs to the 60m deadline. `reattempt_non_converged/1` filters `is_nil(rs.converged_at)`, so resume doesn't re-capture a converged server's baseline; `expedite` skips the comparative gate entirely. Net effect is falling back to the fan-out this is meant to replace.

For subsequent rollouts I genuinely can't tell you which way it fails, and that's the part that bothers me more. If the controller's per-pod cache entry survives the replacement, clamping keeps the aggregate non-decreasing and you get the same latch. But `kurainstance_controller.go:1854` deletes a pod's entry — and its clamping base — whenever a reconcile observes the pod absent, after which accumulation restarts near zero and the check can never fire no matter how badly the new image behaves. For a StatefulSet the name is stable but the object does briefly disappear, so which branch you get depends on whether a reconcile lands in that window.

Each layer is individually right (monotonic counter in Rust; clamp-and-evict in Go is correct for the metric's own purposes) — it's the absolute-vs-baseline comparison on top that assumes a continuity that isn't there.

Would comparing a delta over the soak window — two readings after convergence — work better? That measures the new image's behaviour without depending on counter continuity across the restart. Incidentally the dashboard in this PR already models the metric as `rate(...[5m])`, which is the same instinct.

server/lib/tuist/kura/rollouts.ex
```diff
@@ -0,0 +1,1396 @@
...
+      {:ok, health} when is_map(health) ->
+        baseline = %{
+          outbox_messages: health.outbox_messages,
+          fd_timeout_count: health.fd_timeout_count,
+          peer_connection_failures: health.peer_connection_failures
+        }
+
+        {baseline, baseline_healthy?(health)}
+
+      _ ->
```
esnunes, line 918: I think the two "no health aggregate" fallbacks compose into switching the gate off rather than degrading it.

`KubernetesController.rollout_health/2` returns `{:ok, nil}` when the CR has no `status.rolloutHealth`. `capture_baseline/1` then returns `{%{}, false}`, so the server is scoped `soak_eligible: false`. The ineligible branch of `server_gate_failure/1` only matches `{:ok, %{memory_pressure_state: pressure}}`, so `{:ok, nil}` falls to `_ -> nil` and reports no failure.

The `:health_unavailable` clause that would catch this only runs for soak-*eligible* servers — and a server can only be eligible if its baseline read succeeded. So the two decisions aren't independent: a missing aggregate at scope time guarantees the permissive branch at gate time. If it's missing fleet-wide, `gate_failure/1` returns `nil` for everything, waves complete on convergence alone, and the rollout reports success to /ops/kura and to the CI promotion gate with nothing behind it.

Given this is a brand-new status field and the PR description already recounts an apiserver rejection from a CRD/field skew, I don't think that's far-fetched.

Would it be worth distinguishing "the controller has never published an aggregate for this server" from "this server was already unhealthy at baseline"? The second correctly shouldn't gate its own fix; the first is absence of evidence rather than evidence of health. Even short of blocking, counting ungated servers on the rollout row and surfacing it in /ops/kura would make a blind gate visible instead of green.

server/lib/tuist/kura.ex
```diff
@@ -440,7 +444,10 @@ defmodule Tuist.Kura do
   `attrs` keys: `:account_id`, `:region`, `:image_tag`.
   """
   def create_server(attrs) do
-    attrs = normalize_attrs(attrs)
+    attrs =
+      attrs
+      |> normalize_attrs()
+      |> inherit_rollout_image_tag()
```
esnunes, line 451: The wave-inheritance coverage is asymmetric between creation and retry, and I think it leaves a real gap.

Creation is patched centrally — `Kura.create_server/1` pipes through `inherit_rollout_image_tag/1`, so every caller is covered. Retry is patched at one call site, `RunnerCache.retry/2`, rather than inside `Kura.retry_server/2`. The other two callers of `retry_server/2` are the customer-facing dashboard actions (`cache_live.ex:121` and `account_settings_live.ex:291`), and both pass `socket.assigns.latest_version` — the configured runtime tag, which is the rollout's target.

So while a rollout is paused with the target flagged suspect, a customer hitting "Retry" on a failed cache server re-provisions it straight onto that tag, outside its account's wave — the exact case `provisioning_image_tag/2`'s paused branch exists for, and whose comment calls out retried runner-cache servers by name. It's bounded to servers in `:failed` with `current_image_tag: nil` given `retry_server/2`'s guard, but those are still fresh servers taking the suspect image.

Would moving the `provisioning_image_tag/2` call inside `Kura.retry_server/2` work? It'd mirror what `create_server/1` does and make the `runner_cache.ex` wrapper redundant, so there's one place to reason about rather than a list of call sites to keep in sync.

server/lib/tuist/kura/rollouts.ex
```diff
@@ -0,0 +1,1396 @@
...
+      |> limit(1)
+      |> Repo.one()
+
+    if row do
+      {:deployment_failed,
+       %{server_id: row.server_id, region: row.region, deployment_id: row.deployment_id, error: row.error}}
+    end
+  end
+
+  defp regressed_after_convergence(rollout) do
```
esnunes, line 1008: The comparative gate is carefully excluded for pre-existing sickness via `soak_eligible`, but the two hard signals have no equivalent guard — and I reproduced both ending in a loop no operator verb gets out of.

(a) `@rollout_server_statuses` deliberately includes `:failed` so a fix reaches degraded servers, which is right. But `regressed_after_convergence/1` has no `soak_eligible` check, so a server that was *already* `:failed` before the rollout pauses it with `server_regressed_to_failed` as soon as it converges — attributing pre-existing breakage to the new image.

(b) `baseline_healthy?` marks a server at `memory_pressure_state >= 2` soak-ineligible precisely because the sickness pre-dates the wave, and then the ineligible branch of `server_gate_failure/1` re-raises that same condition as `{:critical, ...}` on convergence.

Both give the same sequence:

    AFTER SCOPE:            :running
    AFTER CONVERGE:         {:paused, "server_regressed_to_failed"}
    AFTER RESUME + TICK:    {:paused, "server_regressed_to_failed"}
    AFTER EXPEDITE + TICK:  {:paused, :expedited, "server_regressed_to_failed"}

`resume` only re-attempts non-converged servers, so the signal is untouched; and `hard_failure/1` / `critical_gate_signal/1` are evaluated in `advance_expedited/1` too, so `expedite` — whose docstring says progressive pacing must never stand between an incident and its fix — re-pauses on the next tick. That leaves `abort` or the kill switch. Since the aggregate takes the max across an instance's pods and Kura nodes do latch at Critical on unreclaimable memory, one known-sick node can hold a fleet-wide fix.

The asymmetry I'd highlight: a server too sick to be trusted as a soak *reference* is still trusted as a fleet-wide *veto*.

The row already has `baseline_captured_at` and the baseline health, so gating both signals on "was this already true at baseline?" seems available — pre-existing cases would join the ungated set, which is where the design already puts them. Separately, should an explicit operator `expedite` be able to acknowledge a standing signal? The actor and reason are already recorded, so it'd be an auditable override rather than a silent one.

infra/grafana-dashboards/tuist-kura-rollout.json
```diff
@@ -0,0 +1,528 @@
...
+            "placement": "bottom",
+            "showLegend": true
+          },
+          "tooltip": { "mode": "multi", "sort": "desc" }
+        },
+        "targets": [
+          {
+            "datasource": "$datasource",
+            "editorMode": "code",
+            "expr": "sum by (pod) (rate(kura_peer_connection_failures_total{cluster=~\"tuist-$env\"}[5m]))",
```
esnunes: The "Peer connection failures" panel is querying `kura_peer_connection_failures_total`, which I don't think is the exported series name.

The counter is registered as `"kura_peer_connection_failures_total"` and `prometheus_client`'s text encoder appends `total` unconditionally, so it scrapes as `kura_peer_connection_failures_total_total`. Every other Kura dashboard in the repo already uses the doubled form — `kura-pod.json` has `kura_http_requests_total_total`, `kura_replication_apply_results_total_total`, `kura_bootstrap_runs_total_total` and a dozen more — so this panel is the outlier rather than the convention.

The reason I'd bump this above a nit: it's the one signal the dashboard exists to explain a pause with, and an empty panel looks exactly like a healthy one. Someone debugging a `peer_connection_failures_regressed` pause would see a flat line and conclude the gate was wrong.

Querying `kura_peer_connection_failures_total_total` matches the other dashboards. Dropping the `_total` from the `registry.register` name would work too, but that changes the exported name for every consumer.

server/lib/tuist/kura/rollouts.ex
```diff
@@ -0,0 +1,1396 @@
...
+    with_locked_rollout(rollout.id, fn
+      %Rollout{status: :running, mode: :expedited} = rollout -> advance_expedited(rollout)
+      %Rollout{status: :running, mode: :progressive} = rollout -> advance_progressive(rollout)
+      _ -> :ok
+    end)
+
+    :ok
+  end
+
+  defp advance_expedited(rollout) do
```
esnunes, line 629: Expedited rollouts look like they have no deadline, and I think that interacts badly with the new promotion gate.

`advance_expedited/1` never calls `open_current_wave/1`, so `wave_started_at` is never stamped and `reset_health_clock_and_check_deadline/1` is never reached — the 60m deadline that protects progressive waves has no counterpart. A server whose deployment stays `:running` indefinitely (image pull backoff, a pod that never goes Ready) leaves the rollout `:running` forever with no pause signal and an empty `pause_reason`.

Canary and staging both run expedited, and the gate waits for the canary rollout to reach `:completed` — so it sees `running`, keeps polling, and exits 1 at its 25-minute deadline. The CI log says "still running" and the rollout record says nothing is wrong, which is a rough place to start debugging a blocked production deploy.

Would stamping a start time on expedited rollouts and applying the same deadline pause work? Then the failure names itself in the rollout record and reaches the gate as `paused` with a reason, rather than as a timeout.

server/lib/tuist/kura/reconciler.ex
```diff
@@ -81,23 +81,44 @@ defmodule Tuist.Kura.Reconciler do
...
-    RunnerCache.reconcile()
+    if Tuist.Environment.kura_control_plane?() do
+      # Version scheduling runs before runner-cache convergence so that on
+      # the first tick after a tag change the rollout record already
+      # exists when a runner-cache node is created or retried — otherwise
+      # `Rollouts.provisioning_image_tag/2` sees no active rollout and
+      # hands the node the target tag outside its account's wave. A
+      # freshly enabled account's node is created one step later in the
+      # same tick and joins its wave on the next one.
+      schedule_runtime_rollout()
```
esnunes: Related to the transaction-nesting issue, but worth its own fix: there's no isolation around the new first step of the tick.

`schedule_runtime_rollout()` runs first in `reconcile/0` and isn't wrapped, so anything raising inside it skips everything after — `RunnerCache.reconcile`, `Lifecycle.reconcile`, the retired-region and destroying/moving-out passes, `reconcile_deployments`, and the observation projection. The projection stopping is the part I'd worry about most, since server status is derived from it.

Two specific raise sources beyond the nesting bug:

- `compute_wave_assignments/1` calls `Usage.recent_request_counts_by_account/2`, which does `ClickHouseRepo.all/1` and raises on failure. That makes ClickHouse availability a hard dependency of the entire Kura control plane, for something that only decides wave *ordering*.
- Several writes are matched as `{:ok, _} = ...` (`scope_server/3`, `create_rollout/2`, `record_event/5`, `mint_missing_deployments/1`), so a constraint or validation error surfaces as a `MatchError` rather than something the caller can skip.

Two fixes that compose: `try/rescue` around the usage read with a fallback to id ordering (wave order degrades, nothing else does), and a rescue around `Rollouts.sync()` itself that logs and lets the rest of the tick proceed. A stalled rollout is visible in /ops/kura; a stalled reconciler is much harder to notice.

server/lib/tuist/kura/rollouts.ex
```diff
@@ -0,0 +1,1396 @@
...
+
+  # The gate is evaluated over every server updated so far in the
+  # rollout — not only the current wave — so a slow-burn regression
+  # surfacing in the canary still stops wave 2 from scheduling.
+  # Soak-ineligible servers (unhealthy at wave-schedule time) skip only
+  # the comparative and absolute soak conditions; the critical safety
+  # signal (memory pressure) still applies to them. Returns nil when all
+  # pass, `{:unhealthy, details}` for a soak-clock reset, or
+  # `{:critical, details}` for an immediate pause.
+  defp gate_failure(rollout) do
```
esnunes, line 1098: The `@scope_batch_size` comment makes exactly the right argument — holding the lock through hundreds of round trips blocks the operator verbs "exactly when someone is most likely to reach for pause". I think the gate loop escapes that reasoning.

`gate_failure/1` does `Repo.all()` over every converged rollout server with no `limit`, then one `Provisioner.rollout_health/1` per row — a live Kubernetes GET each, sequential, no per-tick memoisation. That set only grows, so by the last wave it's the whole fleet. And it's called from both `open_current_wave/1:641` and `evaluate_wave/1:961`, so a tick in the final phase does 2N round trips. All inside `with_locked_rollout`, holding a pooled connection and the same row lock `pause`/`resume`/`abort`/`expedite` need. `mint_missing_deployments/1` is unbounded the same way.

Memoising the health reads for the duration of a tick would remove the double call for free. Beyond that: batch the gate the way scoping is batched, or hoist the reads outside the transaction and take the lock only for the state transition?

Small correction while I'm here — that comment (and a couple of others) says the reconciler ticks every 30s, but the cron is `{"* * * * *", Tuist.Kura.Reconciler}` in `oban/runtime_config.ex`, so 60s. At 50 servers scoped per tick that's double the time to scope a large wave, against the same 60m deadline. Might be worth a look at whether the batch size and the deadline are still sized right for the real cadence.

server/lib/tuist/kura/runner_cache.ex
```diff
@@ -413,6 +413,11 @@ defmodule Tuist.Kura.RunnerCache do
   end
 
   defp retry(%Server{} = server, image_tag) do
+    # A retry re-provisions from scratch, so it inherits the account's
+    # rollout wave state like a fresh server: the baseline tag until the
+    # wave completes, the target after.
+    image_tag = Tuist.Kura.Rollouts.provisioning_image_tag(server.account_id, image_tag)
```
esnunes: The new inheritance in `retry/2` is right, but I think it desynchronises the retry backoff.

`retry/2` now deploys `provisioning_image_tag(server.account_id, image_tag)` — the baseline tag while the account's wave is pending. But the backoff history is still keyed on the unmodified target tag: `retry_failure_histories(servers, image_tag)` filters `where: d.image_tag == ^image_tag` with the original argument.

So: node retried on baseline → deployment fails → the row has `image_tag = baseline` → next tick's history query finds nothing for the target tag → `retry_due?/2` returns `true` for the empty case → retry immediately. The exponential backoff never engages, and a hard-failing runner-cache node gets re-provisioned every tick for the whole rollout.

Resolving the tag once before `nodes_to_retry/3` and passing the resolved value to both the history query and `retry/2` would keep them in sync.

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -1792,6 +1938,85 @@ func (r *KuraInstanceReconciler) primaryPodHealth(ctx context.Context, instance
...
+		if ringFingerprints[i] != ringFingerprints[0] {
+			ringConsistent = false
+			break
+		}
+	}
+	// The conjunctions require every expected pod to have a report: a pod
+	// that exists but cannot be sampled, or a replica that never came up,
+	// reads as unhealthy rather than silently narrowing the aggregate to
+	// the pods that happened to answer.
+	sampledAll := health.SampledPods == expected && int32(len(pods)) >= expected
```
esnunes: Small one in the aggregate, but it has a visible effect on the gate.

`sampledAll := health.SampledPods == expected && int32(len(pods)) >= expected` — `SampledPods` counts pods in the selector-matched list, which can exceed `spec.replicas` during a scale-down while the retired ordinal still exists and still answers `/status/rollout`. I ran the aggregate with 3 healthy pods and `replicas: 2` and got `ready=false serving=false ringConsistent=false` with every pod reporting fine, which the Elixir gate reads as `:not_ready` and which resets the soak clock for the whole drain window.

The second conjunct also looks redundant: `SampledPods <= len(pods)` always holds, so `== expected` already implies `len(pods) >= expected` — which makes me think `>=` was meant on the first term.

Worth lining up across the boundary too: the Elixir side already writes this as `health.sampled_pods >= health.expected_pods` in both `baseline_healthy?` and `gate_checks/2`, so the two halves currently disagree on the comparison.

`health.SampledPods >= expected` (dropping the redundant conjunct) would match — unless an unexpected extra pod is meant to be its own signal, in which case it'd probably be clearer as an explicit condition than folded into ready/serving.

server/test/tuist/kura/rollouts_test.exs
```diff
@@ -0,0 +1,740 @@
...
+  end
+
+  defp healthy_health(overrides \\ %{}) do
+    Map.merge(
+      %{
+        ready: true,
+        serving: true,
+        ring_consistent: true,
+        backfilling_peers: 0,
+        outbox_messages: 10,
```
esnunes, line 46: The lifecycle coverage here is strong — supersede cancelling both minted and adopted deployments, retired-region exclusion, the reclaimed-server drop-out, cancelled-deployment re-mint, the late-joiner unscoped check, and especially the supersede-chain inheritance test, which is the subtlest property in the file and is asserted precisely. Driving soak/deadline by back-dating columns instead of sleeping means none of it is flaky either.

Where I'd add tests is the gate's own arithmetic, which is the novel half of the design and currently has none:

- **The counter checks.** `healthy_health/1` pins `outbox_messages: 10, fd_timeout_count: 0, peer_connection_failures: 0` and nothing overrides them post-convergence, so the threshold arithmetic, `counter_regressed?/2` and `counter_grew?/2` are never exercised. That's how the counter issue I raised separately could ship in either direction — a gate that never fires or one that never passes, and the suite is green for both. An at-threshold/over-threshold pair for outbox and a "counter went down after a restart still passes" case would pin it.
- **One test per `gate_checks/2` condition.** Only `serving: false` and critical pressure are covered; `:sample_stale`, `:pods_unsampled`, `:not_ready`, `:ring_skew` and `:backfill_in_flight` aren't.
- **`rollout_health` returning `{:ok, nil}` and `{:error, _}`.** Only ever stubbed `{:ok, map}`, so the ungated-fleet path is invisible.
- **`regressed_after_convergence/1`** — converge a server, flip it to `:failed`, sync, assert `pause_reason == "server_regressed_to_failed"`. No test at all today, and it's the "new image killed a server that had already taken it" detector.
- **`expedite` from `:paused`.** Only tested from `:running`, but pause-then-expedite is the incident path the feature exists for, and it's the only branch that runs `reattempt_non_converged/1`.

One more that's worth knowing: `create_active_server/1` calls `close_open_deployments/1` in setup, which removes exactly the state that triggers the transaction-nesting crash I flagged. A test that leaves a server `:provisioning` with its install deployment open would have caught it.

server/lib/tuist_web/live/ops_kura_live.ex
```diff
@@ -0,0 +1,62 @@
...
+  def handle_info({:kura_rollouts, :updated}, socket) do
+    {:noreply, load_rollout_state(socket)}
+  end
+
+  @impl true
+  def handle_event("operate", %{"action" => action, "reason" => reason}, socket) do
+    actor = socket.assigns.current_user.email
+
+    socket =
+      case operate(socket.assigns.rollout, action, actor, reason) do
```
esnunes, line 47: Two small crash paths in the ops views — operator-only, so Sentry noise rather than anything exposed. (The authorization looks right: all three sit in `live_session :ops` behind `ensure_authenticated` + the `:ops` read policy, which binds `handle_event` too, and the verbs re-check status under the row lock rather than relying on the hidden buttons.)

1. `load_rollout_state/1` assigns `Rollouts.latest_rollout()`, which is legitimately `nil` on a fresh environment. The template guards with `<%= if @rollout do %>`, but `handle_event("operate", ...)` passes the assign straight through, and `operate/4`'s second clause calls `Rollouts.pause(nil, actor, reason)` against a `%Rollout{}`-matching head — `FunctionClauseError` and a killed LiveView. Same for an event missing `"reason"`, which matches no clause. An `operate(nil, _, _, _)` head returning `{:error, :no_rollout}` plus a catch-all `handle_event` would cover both.

2. `Rollouts.get_rollout/1` is `Repo.get(Rollout, id)` on a `binary_id` key, so `/ops/kura/rollouts/foo` raises `Ecto.Query.CastError` before the `is_nil` check at line 22 can turn it into the intended `NotFoundError`. `ops_database_live.ex:122` already has the `Ecto.UUID.cast/1` guard pattern.

While I'm here: the three new LiveView suites all stub `tuist_operator?` to `true` in setup and never flip it, so nothing asserts a non-operator is rejected from the pages or from the `operate` event that pauses a fleet-wide rollout. The gate is correct today, but a future reorder of the `live_session :ops` `on_mount` list would ship green.

server/lib/tuist/prom_ex.ex
```diff
@@ -80,6 +80,8 @@ defmodule Tuist.PromEx do
         Tuist.License.PromExPlugin,
         Tuist.Kura.PromExPlugin,
         Tuist.Runners.PromExPlugin,
+        Tuist.Kura.PromExPlugin,
```
esnunes: Two small wiring things on the observability side.

1. `Tuist.Kura.PromExPlugin` is now registered twice — it's already at prom_ex.ex:81 and the diff adds it again at :83 next to the new rollouts plugin. PromEx concatenates each plugin's `measurements_mfa` into the poller with no dedup, so that plugin's two polled gauges (which run Postgres *and* ClickHouse queries) execute twice a minute on every server and processor replica, and the duplicate metric definitions get dropped at boot with a `"Metric name already exists. Dropping measure."` warning each. No double-counting, just wasted query load and permanent boot-log noise. Deleting line 83 should do it.

2. I couldn't find `TUIST_OPS_SLACK_WEBHOOK_URL` rendered anywhere under `infra/helm/tuist/**`, and this PR doesn't add a values entry or an external-secret key for it (`server-config-external-secrets.yaml` is untouched). Unless the encrypted secrets blob already carries `ops.slack_webhook_url`, `Notifier.notify/3` would take its `_ -> :ok` branch in every environment and the rollout Slack surface would be inert — silently, since a missing URL is the documented "disabled" state.

Could you confirm which it is? If the blob already provides it, a note near the accessor would save the next person the same check; if not, it needs the env var plumbed through before the notifications do anything.

(The notifier itself looks solid, for what it's worth — `Task.start` is unlinked, `Req.post` returns a tuple rather than raising, and all seven `notify/3` call sites have matching `message/3` clauses, so a webhook failure genuinely can't reach the reconciler.)

server/lib/tuist/kura/rollouts.ex
```diff
@@ -0,0 +1,1396 @@
...
+      end
+
+    [
+      {health.memory_pressure_state >= 2, {:critical, :memory_pressure_critical}},
+      {not fresh_sample?(health), {:unhealthy, :sample_stale}},
+      {health.sampled_pods < health.expected_pods, {:unhealthy, :pods_unsampled}},
+      {not health.ready, {:unhealthy, :not_ready}},
+      {not health.serving, {:unhealthy, :not_serving}},
+      {not health.ring_consistent, {:unhealthy, :ring_skew}},
+      {health.backfilling_peers > 0, {:unhealthy, :backfill_in_flight}},
```
esnunes: I think `{health.backfilling_peers > 0, {:unhealthy, :backfill_in_flight}}` is gating on backfill *happening* rather than backfill *failing to progress*, and those come apart in both directions.

Backfill is resumable — after the initial cycle a restart re-checks manifests rather than refetching from scratch — so an in-flight backfill is expected post-upgrade work, not a regression signal.

**It fires on normal behaviour.** The rollout restarts every pod, and a fresh process re-enters the initial cycle (`Pending` = "in-cycle peers have passes outstanding"), so `backfilling_peers > 0` is the expected state right after convergence. `gate_checks/2` resets the soak clock on it, and the gate runs fleet-wide over every converged server — so the soak can't start until backfill drains everywhere, and one slow mesh holds every wave against the same 60m deadline.

**And it can't see the case it's for.** `backfilling_peers` counts peers with an outstanding pass, not progress. The runtime's own comment on `observe_admission_gate` says it "sits pinned and the initial cycle reads `pending` indefinitely" when memory admission blocks passes — so a fully stalled node and a healthily progressing one look identical through this field.

There's also a tension with `baseline_healthy?`, which treats the same fact as a reason to *exclude* a server from the comparison. Its comment argues backfill "would reset the soak clock, or exhaust the wave deadline, on work that started before the upgrade" — which reads to me as the argument for not gating on it at all.

The good news is `/status/rollout` already publishes what a progress check needs, and the controller drops most of it (it decodes `backfill_backfilling_peers` and `backfill_initial_cycle`, and the aggregate carries only the first):

- `backfill_ring_fullness_percent` — a real progress number, already load-bearing enough that `state.rs:455` uses it for the serving decision
- `backfill_initial_cycle` — `degraded` is explicitly "budget exhausted through real failures", and explicitly non-terminal (retries advance it to `complete`)
- `backfill_budget_exhausted_real_peers` vs `..._capability_peers` — why it isn't progressing, with the capability variant documented as a benign bystander

Two options depending on how much you want to take on:

1. *Cheap, no new state:* drop the `backfilling_peers` condition and gate on `initial_cycle == degraded` or `budget_exhausted_real > 0` instead. Those already mean "backfill is in trouble" rather than "backfill is happening", and `degraded` being non-terminal makes it a natural soak-reset rather than a hard pause. Needs only publishing two more fields on the aggregate.

2. *Proper no-progress check:* publish `ring_fullness_percent` too, keep the last reading per rollout server, and fail only when `backfilling_peers > 0` AND fullness hasn't advanced since the previous sample. `kura_rollout_servers` is already shaped for per-server readings, so it'd sit alongside the baselines.

Either way `backfill_initial_cycle` is already decoded for the `caughtUp` warm-handoff gate, so half the plumbing exists.

esnunes (approved):
This is a lot of careful work and the shape of it is right. Replacing a single-tick fan-out with durable, resumable rollout state; measuring regression against a per-server baseline rather than absolute thresholds; pausing instead of auto-rolling-back; and treating Grafana as the paging authority because a dead control plane can't report its own death — each of those is the non-obvious choice, and each is argued in the code rather than just asserted. The staging drills, especially the one that hard-paused on a real CRD skew and was root-caused from the pause record alone, are genuinely good evidence. The lifecycle test coverage is strong too: supersede with adopted deployments, the late-joiner unscoped check and the supersede-chain inheritance case are the subtle properties and they're asserted precisely.

Approving — nothing here is a structural objection, and almost everything I found sits inside the gate and its failure modes rather than in the design. Flagging the ones I'd prioritise.

The main one is a transaction-nesting bug. `Kura.create_deployment/3` calls `Repo.rollback` inside its own `Repo.transaction`, and the rollout tick calls it from inside `with_locked_rollout`'s transaction — so the inner rollback poisons the outer connection and the next statement raises. I reproduced it against a real DB with a server left `:provisioning` with an open install deployment, which is an ordinary state one tick after a server is created. Because `schedule_runtime_rollout()` runs first in `reconcile/0`, the raise means the observation projection never runs, so nothing closes that deployment and every subsequent tick repeats — the whole Kura control plane stops, not just rollouts. Relatedly, nothing wraps `Rollouts.sync/0`, and `compute_wave_assignments/1` makes ClickHouse a hard dependency of the reconcile loop for something that only decides wave ordering.

Next, the counter half of the gate. `outbox_messages` is a gauge and gets a tolerance band; `fd_timeout_count` and `peer_connection_failures` are process-lifetime counters and get `current > baseline` with none. On the first rollout onto this image the pre-upgrade pods don't emit the field at all, so every baseline is 0 and one transient peer error latches the wave to its 60m deadline. For later rollouts I honestly can't tell you which way it fails — if the controller's per-pod cache entry survives the replacement, clamping latches it; if a reconcile observes the pod absent, the base is evicted and the check goes inert. Each layer is individually right; it's the absolute-vs-baseline comparison across a restart the rollout itself causes that doesn't hold. A delta over the soak window would sidestep all of it — and the dashboard in this PR already models the metric as a rate.

Three more in the same area. If the controller never publishes `status.rolloutHealth`, the baseline fallback and the soak-ineligible fallback compose into a gate that reports no failure for anything — waves complete on convergence alone and report success to the CI gate with no health evidence. Both hard signals fire on conditions that pre-date the rollout with no `soak_eligible` guard, ending in a loop where `resume` doesn't clear the signal and `expedite` re-pauses next tick; I reproduced that for both the already-`:failed` server and the critical-memory-pressure case, and the asymmetry is worth naming — a server too sick to be trusted as a soak reference is still trusted as a fleet-wide veto. And `backfill_in_flight` gates on backfill *happening* rather than *stalling*: since the rollout restarts every pod, `backfilling_peers > 0` is the expected post-convergence state, and because it's a peer count rather than a progress measure it reads identically whether backfill is advancing or fully blocked.

Separately, the dashboard's peer-failure panel queries a name that isn't exported (the doubled `_total` the other Kura dashboards all use), so the one signal it exists to explain a pause with renders permanently empty.

The rest are smaller: wave inheritance patched centrally for creation but only at one call site for retry, so the dashboard retry still hands fresh servers the paused suspect tag; expedited rollouts having no deadline, which hangs the promotion gate on a stalled server; a `==` where the Elixir half uses `>=` in the Go aggregate; the gate's unbounded Kubernetes reads inside the `FOR UPDATE` transaction that the `@scope_batch_size` comment already warns about; the runner-cache retry backoff being defeated during a rollout; a duplicated PromEx plugin registration; and test gaps concentrated on exactly the gate conditions above.

server/lib/tuist/kura/rollouts.ex
fortmarek: Fixed, and thank you for the reproduction — it was the fastest way in. `create_deployment/3` no longer rolls back on the expected error paths: nothing is written before the guards, so it returns `{:error, reason}` and leaves the caller's transaction intact. `mode: :savepoint` turned out not to help, since Ecto applies it to operations rather than to a nested `Repo.transaction`. Added a test that leaves an install deployment open, which reproduced your stack trace before the fix.

server/lib/tuist/kura/reconciler.ex
fortmarek: Fixed. `schedule_runtime_rollout/0` is now contained so a raise there cannot skip the rest of the tick, and the ClickHouse usage read falls back to id ordering — it only decides wave order, so it should not make the analytics store a hard dependency of the control plane. Left the `{:ok, _} = ` writes as-is: with the nesting bug gone they raise only on genuine constraint violations, which the containment now turns into a logged, Sentry-reported tick rather than a stalled reconciler.

server/lib/tuist/kura/rollouts.ex
fortmarek: Took option 1. The gate now keys on `backfill_degraded` and `backfill_budget_exhausted_peers` — backfill in trouble rather than backfill happening — both published as new aggregate fields, and both soak resets rather than hard pauses since `degraded` is non-terminal. `backfilling_peers` stays in the aggregate for the dashboard but no longer gates. You were right that it fired on the rollout's own work and could not see the stall it was meant for.

server/lib/tuist/kura/rollouts.ex
fortmarek: Fixed both halves. The counters now get the same tolerance band as outbox (proportional, floor 25), so the reconnect churn a rollout causes by definition does not read as a regression. And the unambiguous case is fixed at the source: the parser reads an absent counter as `nil` rather than 0, so a runtime that never emitted the field gives no baseline and the comparison is skipped instead of measuring against a zero that was never observed.

server/lib/tuist/kura/rollouts.ex
fortmarek: Fixed. A wave now requires health evidence from at least one soak-eligible server: if the aggregate is missing fleet-wide, every server is ineligible, every check is vacuous, and the rollout would have reported success to /ops/kura and the promotion gate with nothing behind it. Deliberately a fleet-level check rather than per-server, so one sick server still cannot block a wave the others can carry.

server/lib/tuist/kura/rollouts.ex
fortmarek: Fixed both. Soak-ineligible servers no longer raise hard signals at all: their sickness pre-dates the wave, and critical pressure is one of the conditions that makes them ineligible in the first place, so re-raising it was pausing on exactly what the exclusion exists to ignore. The same guard now scopes `regressed_after_convergence/1`. Memory pressure on those servers is still on the Kura dashboards and alerts, which is where pre-existing sickness belongs. The test that asserted the old behaviour now asserts the new one.

server/lib/tuist/kura/rollouts.ex
fortmarek: Fixed. Expedited rollouts stamp `wave_started_at` and get the same deadline pause, so a deployment that never finishes names itself in the rollout record and reaches the promotion gate as `paused` with a reason rather than as a 25-minute CI timeout. Test covers it.

server/lib/tuist/kura.ex
fortmarek: Fixed by moving the resolution into `Kura.retry_server/2`, as you suggested — it mirrors `create_server/1` and covers the dashboard actions, which were the ones passing the configured (suspect) tag. Removed the now-duplicated call at the runner-cache site so there is exactly one resolution point.

server/lib/tuist/kura/runner_cache.ex
fortmarek: Fixed, with one adjustment to the suggested shape: a single resolved tag for the cohort would be wrong, since accounts in the same region can sit in different waves. The history is now keyed per node on the tag that node will actually be deployed with. That keeps the sync you were after and preserves the existing rule that a change in the tag resets the backoff — there is a test asserting a new image retries immediately, which a tag-agnostic history broke.

infra/kura-controller/controllers/kurainstance_controller.go
fortmarek: Fixed — `health.SampledPods >= expected`, dropping the redundant conjunct, which also lines the comparison up with the Elixir side as you noted. Added a test with three pods reporting against `replicas: 2` that fails on the old form.

server/lib/tuist/kura/rollouts.ex
fortmarek: Fixed the double call by memoising health reads per tick (and per operator verb, so a verb never judges on a previous tick's reading), and bounded `mint_missing_deployments/1` like scoping. Left the gate itself reading every converged server rather than batching: unlike scoping, a partial pass would let a wave pass its soak on a subset. Also corrected the cadence comments — the cron is `* * * * *`, thank you.

infra/grafana-dashboards/tuist-kura-rollout.json
fortmarek: Fixed — the panel queries `kura_peer_connection_failures_total_total` now, matching the other Kura dashboards. Agreed on the severity reasoning: an empty panel and a healthy one look identical, and this is the one series the dashboard exists to explain a pause with.

server/lib/tuist_web/live/ops_kura_live.ex
fortmarek: Both fixed. `operate/4` takes a nil rollout and the handler no longer pattern-matches on the params, so a submit missing a field is rejected with a flash instead of killing the view. `get_rollout/1` casts the id first, so a malformed one reads as not-found rather than a cast error — with a test. Thanks for confirming the authorization path while you were in there.

server/lib/tuist/prom_ex.ex
fortmarek: Both fixed. The duplicate registration is gone (the merge with main added it back next to the rollouts plugin). On the webhook: you were right that nothing rendered it, so the notifier was inert. Rather than leave that implicit I wired it through the chart as `kuraRuntime.rollout.opsSlackWebhookSecret`, reading from a Secret since the URL is a credential; unset still means notifications off, which is the documented state.

server/test/tuist/kura/rollouts_test.exs
fortmarek: Added the gate arithmetic tests. The tolerance bands at and past their edges for outbox and the failure counters, a counter that went backwards after a restart, a counter the aggregate never reported, one case per absolute condition in `gate_checks/2`, and backfill in flight NOT holding a wave. You were right that the suite was green for a gate that never fires and one that never passes — writing these is what pinned down the counter semantics.

fortmarek:
You're right, and it's not a deployment detail I had covered — there was no deployment-time baseline, so the first rollout in an environment ran with `baseline_image_tag` nil. Paused, an account with no server has no mesh tag to match either, so both fallbacks were empty and provisioning handed out the suspect target: exactly the case the paused pinning exists for.

Fixed in d29caa0 by persisting one at creation, as you suggested: the baseline is read off the fleet, which is still on the pre-rollout image at that moment. It stays nil only when the fleet has no servers at all, where there is no prior image to fall back to and nothing to protect. Covered by a test for that scenario (paused first rollout, account with no server → baseline, not the target), verified to fail without the change.

PR #12025, opened 2026-07-22
fix(server): self-heal missing Kura cache endpoint mirror row

fortmarek:
## What changed

The Kura reconciler now self-heals a missing `account_cache_endpoints` mirror row for an active public region, instead of only writing it once at activation. A new `Kura.cache_endpoint_mirrored?/1` reports whether the derived `:kura` endpoint row exists for a server's current URL, and the reconciler's `endpoint_in_sync?/1` consults it in the URL-matches branch rather than returning a bare `true`.

## Why it changed (the bug)

A Tuist-managed Kura region can be **Active** (its `kura_servers` row healthy, its public `<account>-<region>.kura.tuist.dev` host serving `/up`) while the CLI never routes to it. `GET /api/cache/endpoints` with the `kura` client feature flag returns only the legacy `cache-<region>.tuist.dev` regionals, so `tuist setup cache` and subsequent builds keep using the legacy cache and the Kura node sits unused. This blocks the trunk-scoped Xcode CAS snapshot feature (#11866), which needs the account routed to its Kura node to be validated at real WAN RTT.

## Root cause

This is a mirror gap, not deploy skew. The observed deployed server already includes the commit that dropped the `:kura_cache` feature flag (#11944), so the read path is ungated and simply queries `AccountCacheEndpoint(technology: :kura)`.

That derived row is written only once, when a server transitions into `:active` (`Kura.activate_server/2` calls `ensure_cache_endpoint/2`). The steady-state reconciler re-derives status every tick, but `converge/2` short-circuits when `converged? and endpoint_in_sync?`. The problem: `endpoint_in_sync?/1` only compared the rendered `public_url` against `kura_servers.url`. It never checked that the mirror row actually exists.

So an active public server whose stored URL matches the rendered host but whose mirror row is absent (activated before the mirror logic existed, or the row pruned out of band during a URL change or handoff) is treated as in sync forever. `ensure_cache_endpoint/2` never re-runs, the row stays missing, and the account is silently pinned to the legacy cache with no self-heal path.

### Regression source

The incomplete convergence check was introduced in #11270 ("scope managed-region public hosts per environment"). That PR added the `converged? and endpoint_in_sync?` gate and `endpoint_in_sync?/1` itself to re-derive the mirror when a managed region's public host is renamed, and in the same change added `prune_superseded_cache_endpoint/3`, a `Repo.delete_all` that removes the old mirror row during such a rename. It paired a row-deletion path with a heal check that keys only on the URL: once the server settles on a URL that matches the render, an absent row is never re-created. This change extends that same `endpoint_in_sync?/1` to also require the row's presence.

Separately, #11944 ("make Kura cache routing CLI-driven and drop the :kura_cache flag") made CLI routing depend solely on the mirror row (no feature-flag gate), which is what turns a missing row into a user-visible routing regression rather than an account that was simply not opted in.

## Why this solution

The reconciler is already the authority that re-derives state every tick, and it already has a precedent for routing a converged server back through activation when its URL drifts. Extending the same in-sync gate to also require the mirror row keeps a single convergence path and reuses the existing idempotent (`on_conflict: :nothing`) insert. A missing row now reports out of sync, the server routes through the endpoint-gated `activate_server`, and the row is rewritten. This means:

- The stranded row backfills automatically within one reconcile tick (~30s) after deploy, with no manual production DB write.
- Any other account in the same state is fixed by the same mechanism.
- Steady-state nodes with the row present stay a per-tick no-op (no extra DB write or broadcast), and private regions (which deliberately never mirror) are always reported as mirrored.

The obvious alternative, a one-off backfill migration for the affected account, would fix the symptom for one account without closing the recurrence, and would still leave the reconciler unable to heal the class of failure.

## Impact

- Accounts with a healthy managed Kura region but a missing mirror row start being surfaced to the CLI, so `tuist setup cache` and builds route to Kura instead of the legacy cache.
- No schema or API changes; no data-export surface change.

## Validation

`mix test test/tuist/kura/reconciler_test.exs test/tuist/kura_test.exs` passes (93 tests), `mix credo` clean on the changed modules. New coverage:

- Reconciler re-activates an active public server whose mirror row was deleted and re-derives the row.
- Reconciler leaves a converged public server with its row present untouched (no re-activation).
- `cache_endpoint_mirrored?/1` returns true for private and unknown regions and reflects row presence for public regions.

Final production validation (the CLI resolving `<account>-<region>.kura.tuist.dev` and a non-404 `get_snapshot`) lands once this deploys and the reconciler heals the row on the running server.

### DIFF
--- a/server/lib/tuist/kura.ex
+++ b/server/lib/tuist/kura.ex
@@ -1011,6 +1011,40 @@ defmodule Tuist.Kura do
     {:ok, server}
   end
 
+  @doc """
+  True iff the server's client-facing cache endpoint is already mirrored into
+  `account_cache_endpoints` (the table the CLI resolves).
+
+  Private regions never mirror — their URL is in-cluster only and reachable
+  solely by runner dispatch — so they are always reported as mirrored.
+
+  Public regions must carry the derived `:kura` endpoint row for their current
+  URL. The mirror is written once, when the server transitions into `:active`
+  (`activate_server/2`); the steady-state reconcile otherwise treats a server
+  whose stored URL matches the rendered host as converged and never re-derives
+  it. A public server missing its row — activated before the mirror existed, or
+  with a row pruned out of band — would therefore stay invisible to the CLI
+  forever. Reporting the absent row as out of sync (see `Reconciler`) routes the
+  server back through activation, which re-runs the idempotent mirror insert and
+  self-heals it on the next tick.
+  """
+  def cache_endpoint_mirrored?(%Server{region: region_id} = server) do
+    case Regions.fetch(region_id) do
+      {:ok, region} -> Regions.private?(region) or cache_endpoint_row_present?(server)
+      {:error, _} -> true
+    end
+  end
+
+  defp cache_endpoint_row_present?(%Server{account_id: account_id, url: url}) when is_binary(url) do
+    Repo.exists?(
+      from(e in AccountCacheEndpoint,
+        where: e.account_id == ^account_id and e.technology == :kura and e.url == ^url
+      )
+    )
+  end
+
+  defp cache_endpoint_row_present?(%Server{}), do: true
+
   defp ensure_cache_endpoint(_account, nil), do: :ok
 
   defp ensure_cache_endpoint(account, url) do
--- a/server/lib/tuist/kura/reconciler.ex
+++ b/server/lib/tuist/kura/reconciler.ex
@@ -482,6 +482,14 @@ defmodule Tuist.Kura.Reconciler do
   # not re-written every tick. A non-binary render (e.g. unknown region) leaves
   # the existing `converged?` behaviour untouched.
   #
+  # A matching URL is necessary but not sufficient: the derived `:kura` endpoint
+  # row must also be present. An active public server whose row is missing (e.g.
+  # activated before the mirror existed, or the row pruned out of band) would
+  # otherwise stay converged and invisible to the CLI forever, since the mirror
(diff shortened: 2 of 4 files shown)


fortmarek:
Closing: this was opened on a mistaken premise.

I verified prod directly (read-only Postgres) after opening this. The `tuist` account's `AccountCacheEndpoint(technology: :kura)` row is **not** missing: it exists and is healthy (`https://tuist-eu-central-1.kura.tuist.dev`, `deactivated_at` null, inserted 2026-07-09), and the EU Central `kura_servers` row is active on 0.19.0 with a matching URL. 7 accounts have `:kura` mirror rows in prod.

The reported symptom (CLI resolving the legacy cache despite an Active region) was **deploy skew**, not a mirror gap: at the time the evidence was captured, the running prod build predated #11944, whose read path gated `kura_cache_endpoints` on the `:kura_cache` flag (off for the account) → empty list → legacy fallback. Prod finished rolling out the #11944 build (`sha-134c825f0d1f`) at ~16:30 UTC on 2026-07-22; with the gate gone and the row already present, the endpoint now returns the Kura URL. The fix was the deploy landing.

The change here addresses a genuinely separate, narrow latent gap — an active public server whose mirror row is *actually* absent would not self-heal via the reconciler's `endpoint_in_sync?` check — but that is not what happened for this account, so the incident/regression framing in the description is incorrect. Not merging on a false premise. If we decide the self-heal hardening is worth having on its own merits, it can be reopened as a standalone change with an honest description.

PR #1486, opened 2020-06-22
Adding fix for tuist bundle issue 1466 caused by trailing whitespace

miscampbell:
Resolves https://github.com/tuist/tuist/issues/1466

### Short description 📝

This fixes an issue with the `tuist bundle` failing when ran inside Terminal

### Solution 📦

After debugging this issue I was able to diagnose it was happening due to a trailing \n being appended when using `let version = try String(contentsOf: versionFilePath.url)` inside BundleService.swift this was returning "1.11.0\n"

Which was then used to lookup the version inside `.tuist/Versions` E.g. `/Users/mmiscampbell/.tuist/Versions/1.11.0\n`

### Implementation 👩‍💻👨‍💻

The fix I have implemented is after verifying we could retrieve a version from the .tuist-version file we clean the String using filter and only allowing characters that are not whitespace.

- [ ] Added Unit test to validate the install does not run when the version file contains whitespace characters

PR #12027, opened 2026-07-22
fix(cli): expand Foundation's #bundle macro to Bundle.module in static targets

fortmarek:
### What changed

Tuist-generated projects now set the `SWIFT_MODULE_RESOURCE_BUNDLE_AVAILABLE` compilation condition on Swift targets whose resources are synthesized into a companion resource bundle (static frameworks/libraries and other products that cannot host resources). The condition is appended to `SWIFT_ACTIVE_COMPILATION_CONDITIONS` in `ResourcesProjectMapper`, next to the existing `PACKAGE_RESOURCE_BUNDLE_NAME` plumbing, and only when the SPM-style Swift `Bundle.module` accessor is synthesized. Dynamic frameworks that host their own resources deliberately do not get the condition, mirroring SwiftPM.

### Why / root cause

Foundation's `#bundle` macro ([BundleMacro.swift](https://github.com/swiftlang/swift-foundation/blob/main/Sources/FoundationMacros/BundleMacro.swift)) chooses its expansion from the compilation conditions of the module being compiled:

- `SWIFT_MODULE_RESOURCE_BUNDLE_AVAILABLE` → `Bundle.module`
- `SWIFT_MODULE_RESOURCE_BUNDLE_UNAVAILABLE` → compile error
- `SWIFT_BUNDLE_LOOKUP_HELPER_AVAILABLE` → `Bundle(for: __BundleLookupHelper.self)` (Xcode sets this only for mergeable libraries)
- none → `Bundle(_dsoHandle: #dsohandle) ?? .main`

SwiftPM sets the first condition for resource-bearing modules ([PackagePIFProjectBuilder+Modules.swift](https://github.com/swiftlang/swift-package-manager/blob/main/Sources/SwiftBuildSupport/PackagePIFProjectBuilder%2BModules.swift)). Tuist set none of them, so `#bundle` in a Tuist-generated static framework fell through to the DSO-handle branch, which resolves to the main bundle for statically linked code. Resources synthesized into the companion `<Project>_<Target>.bundle` are invisible from there, so lookups returned nil at runtime while an explicit `bundle: .module` (which uses Tuist's synthesized accessor) worked.

This also affects `#bundle` used as a caller-side default argument ([SE-0422](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0422-caller-side-default-argument-macro-expression.md)): the macro expands in the calling module's context, so a resource-owning static framework calling such an API got the broken expansion baked in — including into binary-cached frameworks, which is how this surfaced in the wild (implicit `#bundle` failing while explicit `.module` passed, on otherwise identical cached builds).

### Validation

- New fixture `examples/xcode/generated_static_framework_with_bundle_macro`: a static framework owning a resource calls an API from another static framework whose `bundle` parameter defaults to `#bundle`, plus a direct `#bundle` use and an explicit `.module` control, with unit tests asserting the resource URL resolves at runtime. The fixture targets macOS so acceptance runs need no simulator and cache warms build a single slice (the expansion and DSO fallback are platform-independent); the deployment target is pinned so generated targets don't default to an SDK newer than the host macOS.
- New acceptance test `GenerateAcceptanceTestStaticFrameworkWithBundleMacro` (generate + `tuist test`) was run before the fix and failed exactly like the field report — `test_directBundleMacro_resolvesResource` and `test_implicitBundleDefaultArgument_resolvesResource` failed, `test_explicitModule_resolvesResource` passed — and passes after the fix (both parallel clones).
- Four new `ResourcesProjectMapperTests` cases cover: condition set for static targets with Swift sources, appending to existing `SWIFT_ACTIVE_COMPILATION_CONDITIONS` without clobbering, and no condition for dynamic frameworks or Objective-C-only targets. Full suite (75 cases) passes.
- The failing field report consumed the resource-owning framework as a binary from the remote cache, where the expansion is frozen into the artifact at warm time regardless of how consumers build. To cover that path, the fixture gained an `App` + hosted `AppTests` layer and the cache acceptance suite gained `generated_static_framework_with_bundle_macro_from_cached_binaries`: `tuist cache` warms both static frameworks into xcframeworks, then `tuist test App` keeps only `App`/`AppTests` as sources and consumes `StaticFramework` as a cached binary. Run pre-fix it fails with the exact field split (direct `#bundle` and the SE-0422 default argument return nil, explicit `.module` passes); post-fix it passes with all three artifacts warmed fresh. Red → green was validated on both the original iOS shape and the final macOS shape; the cache test runs in ~25s and the generator test in ~19s.
- Stale-artifact invalidation: project mappers run before the EE graph mappers hash the graph (`ManifestGraphLoader`), and `Settings` participate in `TargetContentHasher`, so the added condition changes cache fingerprints and pre-fix artifacts (with the broken expansion baked in) can no longer be served to a fixed CLI.

### Forward-compat note

swift-build [PR #1486](https://github.com/swiftlang/swift-build/pull/1486) (merged 2026-06-29, not yet in any shipping Xcode) will push `SWIFT_MODULE_RESOURCE_BUNDLE_UNAVAILABLE` for all static frameworks. Once that ships, both conditions being set makes the macro diagnose "Both ... cannot be set", so we should pursue an upstream change gating that push on the availability condition not already being present. Tracking this separately; it does not affect currently shipping Xcode versions.


cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
```diff
@@ -275,6 +278,31 @@ public struct ResourcesProjectMapper: ProjectMapping {
...
+            guard !values.contains(condition) else { return target }
+            base["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = .array(values + [condition])
+        case let .string(value):
+            guard !value.contains(condition) else { return target }
+            base["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = .string("\(value) \(condition)")
+        case nil:
+            base["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = .array(["$(inherited)", condition])
+        }
+        var target = target
+        target.settings = target.settings?.with(base: base) ?? Settings(base: base, configurations: [:])
```
pepicrft: This helper only updates `settings.base`. `ConfigGenerator` applies `configuration.settings` afterward, so a configuration-specific `SWIFT_ACTIVE_COMPILATION_CONDITIONS` value without `$(inherited)` removes this generated condition. I reproduced this with a Debug value of `CUSTOM_DEBUG`: the generated target contained only `CUSTOM_DEBUG`, and both `#bundle` resource checks failed while explicit `Bundle.module` passed. Please ensure the generated condition is present after configuration-specific settings are resolved, or append it to every configuration as well.

cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
```diff
@@ -275,6 +278,31 @@ public struct ResourcesProjectMapper: ProjectMapping {
...
+    ///   - https://github.com/swiftlang/swift-package-manager/blob/main/Sources/SwiftBuildSupport/PackagePIFProjectBuilder%2BModules.swift
+    private func addingModuleResourceBundleAvailableCondition(to target: Target) -> Target {
+        let condition = "SWIFT_MODULE_RESOURCE_BUNDLE_AVAILABLE"
+        var base = target.settings?.base ?? SettingsDictionary()
+        switch base["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] {
+        case let .array(values):
+            guard !values.contains(condition) else { return target }
+            base["SWIFT_ACTIVE_COMPILATION_CONDITIONS"] = .array(values + [condition])
+        case let .string(value):
+            guard !value.contains(condition) else { return target }
```
pepicrft: This duplicate check treats any substring as the complete condition. With `NOT_SWIFT_MODULE_RESOURCE_BUNDLE_AVAILABLE`, the helper returned early, the generated target kept only that unrelated condition, and the two `#bundle` checks failed at runtime. Please split the string into compilation-condition tokens and compare exact tokens, matching the array branch.

examples/xcode/generated_static_framework_with_bundle_macro/AppTests/Sources/AppResourcesTests.swift
```diff
@@ -0,0 +1,19 @@
+import StaticFramework
+import XCTest
```
pepicrft: These are new test suites, so they need to use Apple’s [Swift Testing framework](https://developer.apple.com/xcode/swift-testing/) with `import Testing`, `@Test`, and `#expect` rather than Apple’s [XCTest framework](https://developer.apple.com/documentation/xctest). The same applies to `StaticFrameworkTests/Sources/StaticFrameworkResourcesTests.swift`. Please migrate both fixture suites so the new coverage follows the repository’s testing standard.

pepicrft (approved):
This looks good overall. I validated the main behavior and reproduced the two edge cases locally. Approving, but it'd be great to address the inline findings before merging.

examples/xcode/generated_static_framework_with_bundle_macro/AppTests/Sources/AppResourcesTests.swift
fortmarek: this is just a fixture

cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
fortmarek: Fixed in 1dbf933631: the helper now appends the condition to every configuration that defines SWIFT_ACTIVE_COMPILATION_CONDITIONS in addition to the base, with a unit test covering the CUSTOM_DEBUG override shape.

cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
fortmarek: Fixed in 1dbf933631: the string branch now splits the value into whitespace-separated tokens and compares exact tokens like the array branch, with a unit test using NOT_SWIFT_MODULE_RESOURCE_BUNDLE_AVAILABLE.

examples/xcode/generated_static_framework_with_bundle_macro/AppTests/Sources/AppResourcesTests.swift
fortmarek: Migrated both fixture suites to Swift Testing (import Testing, @Test, #expect) in 1dbf933631 and re-ran both acceptance tests green.

PR #12028, opened 2026-07-22
feat(server): add Postgres runner workflow_job lifecycle table with dark writes

fortmarek:
Part 1/4 of the runner-state consolidation (Track A), split out of #12022 into a reviewable stack. Stack: **this PR** → #12029 (drift comparator) → #12030 (flag-gated reads) → #12031 (CH outbox).

## What

One Postgres `runner_workflow_jobs` row per workflow_job, mutated through guarded compare-and-set transitions (`queued → claimed → running → completed | cancelled`), dark-written alongside today's authoritative ClickHouse paths. Nothing reads the table yet — this PR is a behavioral no-op for dispatch.

- Every mutation is `UPDATE … WHERE status = <expected>` (or an upsert whose `ON CONFLICT DO UPDATE` carries the guard), so webhook redeliveries and claim races cannot regress a row. Guard misses are `:noop` + telemetry (`tuist_runners_workflow_job_transition_count`, tagged applied/miss), never an error.
- Webhook pipeline writes: `queued`/`waiting` upserts a queued row (mirroring the `runner_job_completions` resurrection guard), `in_progress` records the runner→job binding, and the terminal transition hooks the completion choke point in `Jobs` (`complete_locked`/`record_completed_locked`, next to `persist_completion!`) — deliberately there rather than dispatch.ex so `StaleQueuedJobsWorker` and `OrphanedRunnersWorker` force-completes are covered too; hooking only the webhook would leave PG rows non-terminal forever on those paths.
- Claim-path transitions commit atomically with the claim mutation: `Claims.attempt/5` does `queued → claimed` in the claim-insert transaction, `mark_running/2` does `claimed → running`, and `release/2` / `release_pod_missing/2` move rows back to `queued` in the same transaction as the claim delete (the CH-first crash-ordering contract the recovery workers document does not apply to this row — claim + state commit or roll back together). Terminal rows never match the re-queue guard.
- The row carries the full dispatch-candidate metadata (repository, shape, workflow correlation fields, `requested_dispatch_label`) so the later flag-gated Postgres dispatch read can return the same candidate map the ClickHouse read does.
- `cancelled` is a first-class status here (ClickHouse models it as `completed` + conclusion); later parts map between the two spaces.
- Lock ordering is cycle-free: every transaction touching both tables locks `runner_claims` before `runner_workflow_jobs`.

## Why / root cause

Runner job state lives in three stores (ClickHouse `runner_jobs` argMax view — also the dispatch queue read, Postgres `runner_claims`, Kubernetes Pods) with no transition spanning stores atomically. Every crash window between a CH write and a PG write is a divergence class we've paid for repeatedly: the CH-first ordering contracts in `stale_claims_worker.ex`, the dispatch anti-list, the `enqueued_at` lookback floor protecting the argMax hot path from ClickHouse memory limits, and a family of recovery workers reconciling after the fact. A single PG row with CAS transitions makes claim + state one transaction; ClickHouse becomes analytics-only downstream once the rest of the stack bakes.

## Validation

- 552 tests green (`test/tuist/runners/`, `test/tuist/runners_test.exs`, `test/tuist/oban/runtime_config_test.exs`), including new coverage for every transition, redelivery idempotency, claim-transaction atomicity/rollback, and terminal-state protection.
- `mix excellent_migrations.check_safety` — no warnings from the new migration (new-table indexes safety-assured, matching the `runner_job_completions` seed).
- `mix compile --warnings-as-errors`, credo clean, `mise run format`.
- `server/data-export.md` documents the new table.


fortmarek:

PR #12029, opened 2026-07-22
feat(server): compare runner job state between Postgres and ClickHouse

fortmarek:
Part 2/4 of the runner-state consolidation stack (Track A), split out of #12022. Stack: #12028 → **this PR** → #12030 (flag-gated reads) → #12031 (CH outbox). Base is #12028's branch; retarget to main once it merges.

## What

A log-only comparator between the Postgres lifecycle rows #12028 introduced and the ClickHouse `runner_jobs` argMax view: `JobStateDriftWorker`, a 5-minute Oban cron that diffs recently-updated PG rows against CH's latest `(status, conclusion)` per job and emits telemetry + logs only (`tuist_runners_workflow_job_drift_count`, tagged `status_mismatch` / `missing_in_clickhouse` / `compared`). It mutates nothing in either store.

Design points:

- **This is the production gate for the read flag.** The next PR puts the Postgres dispatch reads behind a kill switch; that flag only flips once this worker holds at zero drift over a bake window. Shipping the comparator separately means it bakes on production traffic while the read path is still under review.
- ClickHouse's `completed` + conclusion pair is folded into the Postgres status space before comparing (`completed` + conclusion `cancelled` reads as `cancelled`), so a cancelled/completed disagreement between the stores counts as drift rather than equivalence.
- Rows updated in the last minute are excluded — a PG transition commits before its paired CH INSERT lands, and fresh rows would report propagation lag as drift.
- `missing_in_postgres` is deliberately not a kind: jobs enqueued before the lifecycle table shipped exist only in CH, and scanning CH for them would re-run the full-history aggregation the lifecycle table exists to retire. They age out of the window on their own.
- The CH lookup floors on the compared batch's minimum `enqueued_at` so partition pruning applies, same as the dispatch read.

## Validation

- 510 tests green across the runners suite + Oban runtime config, including agree/mismatch/missing/cancelled-equivalence comparator cases and the recency-window read.
- `mix compile --warnings-as-errors`, credo clean, formatted.


fortmarek:

PR #12030, opened 2026-07-22
feat(server): serve runner dispatch queue reads from Postgres behind a kill switch

fortmarek:
Part 3/4 of the runner-state consolidation stack (Track A), split out of #12022. Stack: #12028 → #12029 → **this PR** → #12031 (CH outbox). Base is #12029's branch; retarget as the stack merges.

## What

Postgres implementations of the dispatch queue reads, behind a single FunWithFlags kill switch `:runner_dispatch_postgres_reads`, **default OFF**:

- `Jobs.pick_queued_top_k/5` — the dispatch hot path's candidate selection
- `Jobs.queued_count_by_fleet/1` and `queued_count_by_fleet_and_account/1` — the autoscaler's demand reads (the latter is #11981's dispatchable-demand read)

The Postgres twins return the exact candidate map shape in the same deterministic `(enqueued_at ASC, workflow_job_id ASC)` order, honor the same account/repository/workflow_job exclusions, and floor on the same 7-day `enqueued_at` lookback — so a flag flip cannot resurface jobs the ClickHouse path had already aged out of view, and a flip back is behaviorally seamless.

## Why a flag rather than a cutover

ClickHouse remains the default read path and the instant, no-deploy rollback. The flag only flips once #12029's drift comparator holds at zero over a bake window — that's the whole point of the ordering in this stack. Once flipped and baked, the follow-ups (out of scope here) are deleting the CH read path's anti-list, lookback floor, and head-blocking logic.

Worth noting what the PG read structurally fixes: the CH read needs the claims anti-list because CH can say `queued` while PG holds a live claim; the PG lifecycle row transitions to `claimed` in the claim's own transaction, so the queued scan and the claim state can't disagree. The anti-list still applies while the flag is on (kept for parity during rollout), it just stops being load-bearing.

## Validation

- 548 tests green across the runners suite, including flag-on tests proving `pick_queued_top_k`/`pick_queued` and both queued counts serve from Postgres, plus candidate-shape/ordering/exclusion/floor parity tests on the PG implementations.
- `mix compile --warnings-as-errors`, credo clean, formatted.

### DIFF
--- a/server/lib/tuist/runners/jobs.ex
+++ b/server/lib/tuist/runners/jobs.ex
@@ -83,6 +83,14 @@ defmodule Tuist.Runners.Jobs do
   # still-claimable job is never pruned out of view.
   @queued_lookback_seconds 7 * 86_400
 
+  # Kill switch for reading the dispatch queue from Postgres
+  # (`Tuist.Runners.WorkflowJobs`) instead of ClickHouse. Default OFF:
+  # the ClickHouse paths below stay authoritative, and disabling the
+  # flag is an instant rollback with no deploy. Gate flipping is
+  # driven by `Tuist.Runners.Workers.JobStateDriftWorker` reporting
+  # zero drift between the two stores.
+  @postgres_reads_flag :runner_dispatch_postgres_reads
+
   @doc """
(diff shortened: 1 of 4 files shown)


fortmarek:

PR #12031, opened 2026-07-22
feat(server): make Postgres the control-plane store for runner job lifecycle state

fortmarek:
## What

A runner workflow_job's lifecycle state (`queued → claimed → running → completed`) currently lives in three stores that can disagree: the ClickHouse `runner_jobs` ReplacingMergeTree (which is also what dispatch reads to pick the next job), the Postgres `runner_claims` lock table, and the Kubernetes Pods themselves. Because no transition spans those stores atomically, we keep paying for the gaps — careful CH-before-PG write orderings in the recovery workers, an anti-list so dispatch skips CH rows that Postgres already claimed, a lookback floor to keep the hot-path argMax query under ClickHouse's memory limits, and a growing family of workers whose only job is reconciling the stores after the fact.

This PR moves that state into Postgres as the single control-plane store, effective on deploy:

1. **Lifecycle table.** One `runner_workflow_jobs` row per workflow_job (`workflow_job_id` PK) carrying status, the full dispatch-candidate metadata, lifecycle timestamps, and the pod/runner binding. Every mutation is a guarded compare-and-set (`UPDATE … WHERE status = <expected>`, or an upsert whose `ON CONFLICT DO UPDATE` carries the guard), so webhook redeliveries and claim races cannot regress a row. `Claims.attempt/5` transitions `queued → claimed` inside the claim-insert transaction; `mark_running/2` does `claimed → running`; `release/2` and `release_pod_missing/2` re-queue atomically with the claim delete. Terminal transitions hook the `Jobs` completion choke point so the recovery workers' force-completes are covered, not just the webhook.
2. **All control-plane reads serve from Postgres.** Dispatch queue picks (`pick_queued_top_k`), the autoscaler's queued counts, and the recovery workers' candidate scans (`list_orphaned_running`, `list_stale_queued`) — same candidate shapes, same deterministic ordering, same 7-day `enqueued_at` floor as the ClickHouse reads they replace. The CH read bodies for these paths are deleted. The structural win: the CH read needed the claims anti-list because CH could say `queued` while Postgres held a live claim; the lifecycle row transitions to `claimed` in the claim's own transaction, so the queued scan and the claim state can no longer disagree.
3. **ClickHouse replication via a transactional outbox.** Every applied transition inserts a `runner_workflow_job_transition_events` row in the same Postgres transaction, carrying the exact `runner_jobs` insert shape with the transition's microsecond timestamp as the RMT version. `FlushJobTransitionEventsWorker` (1-min cron) drains it in `FOR UPDATE SKIP LOCKED` batches, CH insert before PG delete, so crashes re-flush rows the RMT dedups. Outbox rows carry `account_id` with a cascade delete so no payload outlives its account. Direct CH writes stay for now — dual emission is harmless (the RMT resolves overlapping rows by version) and keeps the dashboards fresh through the bake.
4. **Deploy-window reconciliation (transitional).** The previous release never writes `runner_workflow_jobs`, so while old and new pods overlap (every roll of this release, and any rollback followed by a roll-forward) the table can miss jobs or hold them in a state the old code has since moved past. `ReconcileWorkflowJobsWorker` (1-min cron) heals both classes each tick:
   - **Adopts ClickHouse-only jobs** — a job whose `queued` webhook the old code handled has no Postgres row and would be invisible to dispatch and every recovery scan. Each adoption runs under the workflow_job ordering lock (the lock every completion writer, old and new, takes), so the completion check and the insert cannot straddle a completion landing between them and adopt a dead job as live; a batched completion prefilter keeps the locked path to candidates that might still be live.
   - **Closes rows the old code completed** — every completion path records `runner_job_completions`; a row still non-terminal alongside one is closed with the completion's conclusion.
   - **Re-queues `claimed` rows no claim backs** — the old code's stale-claim sweep and pod-stopped path delete a claim without touching this table; a `claimed` row is pre-mint by construction, so with its claim gone the job can be re-queued safely (handle-guarded, so a job re-claimed meanwhile is left alone).
   - **Brings `queued` rows a live claim backs to the claim's state** — the old code's `Claims.attempt`/`mark_running` write the claim only.

   Steady state finds nothing on any pass; the three Postgres passes drive from a partial live-set index. Deleted with the direct CH writes.

## Why this fixes the divergence class

The root cause of every reconciliation mechanism above is that a claim mutation and its state transition were two writes to two stores with a crash window between them (see the CH-first ordering contracts documented in `stale_claims_worker.ex` / `list_pods_missing_since`). With the lifecycle row in Postgres, the claim insert/delete and the status transition are one transaction — they commit or roll back together, so the crash window (and the ordering contract that guards it) ceases to exist rather than being reconciled after the fact.

## Rollout

No feature flags — the cutover is the deploy, validated on staging before promoting (the standard canary → acceptance → production cascade applies). Rollback is redeploying the previous release: the old code's ClickHouse paths are fully intact (direct CH writes never stopped), so reverting restores the exact pre-PR behavior; Postgres rows go stale in the meantime and are reconciled by `ReconcileWorkflowJobsWorker` on the next roll-forward — the same passes that cover the overlap window during a normal roll.

## Notes for reviewers

- **`cancelled` is a first-class Postgres status** (ClickHouse models it as `completed` + conclusion). The outbox maps between the two spaces.
- **Deadlock safety:** every transaction touching both tables locks `runner_claims` before `runner_workflow_jobs`.
- **The anti-list and per-poll exclusions are kept** even though the PG read makes them structurally redundant — they're cheap, and deleting them belongs in the cleanup PR once the outbox has proven itself in production.
- **Partial indexes for the permanent sweeps.** `list_orphaned_running` (`running` by `started_at`) and `list_stale_queued` (`queued` by `enqueued_at`) scan one status across every fleet, and rows are kept for the account's lifetime; each gets a partial index, and the reconciler's passes drive from a `(status, workflow_job_id) WHERE status IN (queued, claimed, running)` live-set index — the live set is a tiny fraction of the table.
- **Migrations are numbered `202607221300xx`** because main claimed `20260722120000` mid-flight — a same-version migration is silently masked on any DB that applied ours first.
- Supersedes #12022 and the interim stack #12028/#12029/#12030; the commit sequence preserves the reviewable boundaries (table + dark writes → reads → outbox → recovery scans → cutover → reconciliation).
- **Merged with `main` (2026-08-18) and re-audited against everything that landed in the runner area since the branch was cut.** One upstream change needed a Postgres-side complement — #12395's targeted orphan recovery: at pod stop, `Claims.release_by_pod_name/1` now deletes the claim *without* re-queueing (whether the job runs again is GitHub's call), and enqueues `OrphanedRunnersWorker` in targeted mode to cross-check GitHub and re-queue. On `main` that re-queue lands as a direct ClickHouse write; on this branch dispatch reads Postgres, and `Claims.release/2` (which owns the lifecycle re-queue) returns `:stale_claim` because the claim is already gone — so without a fix the job would stay `running` in Postgres and be invisible to dispatch forever. Added `WorkflowJobs.requeue_by_handle/2` (a CAS guarded on both status and the `claimed_at` handle, so a job re-claimed by another Pod or completed since is left alone) and a Postgres twin of `Jobs.get_orphaned_running/1`; the orphan worker falls through to it on `:stale_claim`. Covered by an end-to-end test that runs the real pod-stop release + targeted recovery against the DB. Everything else on `main` was either analytics-only (#12328 overview cards, #12121 deleted the `p95_concurrent_last_hour` CH read) or already compatible (#12265's re-entrant ordering lock, #12431's stranded-time telemetry — its `fleet_name` is on the Postgres scan too).

## After this rolls into production

**Watch (no action needed on green):**

- **`runner_workflow_job_transition_events` depth** — the outbox drains every minute, so it should sit near-empty. Sustained growth is the one failure that matters before the cleanup: it means the flusher is stalled and ClickHouse is being kept correct only by the direct writes the cleanup deletes. This is the gate, and it needs no new machinery to read — it's a row count. (A full Postgres-vs-ClickHouse status comparison is also a single query against either store if you want belt-and-braces before merging the cleanup; there is no cron running it, because nothing consumes the result but a human making that one decision.)
- The `runners: adopted ClickHouse-only workflow_jobs into Postgres` and `runners: reconciled workflow_job lifecycle rows` log lines — should fire during/shortly after each roll of this release, then go quiet permanently.
- `runner_workflow_job_transition_events` should stay near-empty (flusher lag is ≤1 min); growth means the flusher cron is stuck.

**Then one cleanup PR — prepared as #12050 (draft), gated on the outbox staying drained over a bake window (suggested: one week of production traffic) and on this PR having completed at least one production roll:**

1. Delete the direct ClickHouse writes in `Tuist.Runners.Jobs` (`insert_row!` call sites in enqueue/record_claimed/record_running/record_queued/complete) — the outbox becomes the sole CH writer, and the CH-first ordering comments in `stale_claims_worker.ex` / `orphaned_runners_worker.ex` / `pod_claim_reconciliation_worker.ex` go with them.
2. Delete `ReconcileWorkflowJobsWorker`, `Jobs.list_non_terminal/1`, and the `WorkflowJobs` adopt/close/sync/requeue-unbacked passes — with no CH-first write path left and no code path moving a claim without its lifecycle row, none of those states can arise. The one state this release can still produce on its own (a `claimed` row whose claim `release_by_pod_name/1` deleted because the Pod stopped pre-mint) is closed at the source in the cleanup PR.
3. Delete the anti-list machinery in dispatch (`Claims.workflow_job_ids_for_fleet` exclusion plumbing and the head-blocking logic) — the lifecycle row transitions to `claimed` in the claim's own transaction, so the queued scan can no longer disagree with claim state.

Not part of that cleanup: the outbox + flusher (they become the permanent CH replication path) and the `enqueued_at` lookback floor on the Postgres reads (it aligns dispatch visibility with `StaleQueuedJobsWorker`'s 24h backstop, independent of which store serves the read).

## Validation

- 629 tests green across `test/tuist/runners/`, `test/tuist/runners_test.exs`, and `test/tuist/oban/runtime_config_test.exs`, including: every CAS transition and redelivery/idempotency case, claim-transaction atomicity and rollback, Postgres-served dispatch picks/counts/recovery scans, outbox emission + end-to-end flush into ClickHouse, and the reconciler's adopt-under-lock/skip/completion-guard behavior plus one test per reconciliation pass reproducing the exact old-code interleaving it heals.
- `mix excellent_migrations.check_safety` — zero warnings from the two new migrations.
- `mix compile --warnings-as-errors`, credo clean on all touched files, `mise run format`.
- `server/data-export.md` documents both new tables.


fortmarek:
## Why this is still needed after #12295

Both PRs touch the runner lifecycle, but they close different seams, and neither substitutes for the other.

**#12295 closes the Postgres/Kubernetes seam.** A `runner_sessions` row that never gets `ended_at` set, because `reapRunner` deleted the Pod before reporting it stopped. The consequence is a billing row stuck at the six hour clamp and a host that reads as occupied for that whole window.

**This PR closes the ClickHouse/Postgres seam.** A workflow_job's lifecycle state (`queued -> claimed -> running -> completed`) lives in ClickHouse `runner_jobs`, which is also what dispatch reads to pick the next job, while the claim lock lives in Postgres. No transition spans both stores atomically, which is what the CH-before-PG write orderings, the dispatch anti-list, and the argMax lookback floor all exist to work around.

Ship only #12295 and every job still picks its runner off a ReplacingMergeTree that the claim table can contradict. Ship only this one and sessions still leak at ~22%.

### Verified non-overlap

Checked against #12295's actual diff rather than its description:

- #12295 does not touch `runner_workflow_jobs`, the dispatch reads, the claim transitions, or the transition outbox. Its new reconciler arm operates purely on `runner_sessions`.
- This PR does not touch `runner_sessions`, `Runners.Billing`, or the runners-controller.
- No test subject overlaps.

### One thing this PR makes easier, contrary to the note in #12295

#12295's reviewer notes flag a porting trap: `cancelled` is `status = "completed"` plus `conclusion = "cancelled"` in ClickHouse but a first-class status in `runner_workflow_jobs`, so a literal port of `terminal_completions/1` would silently drop cancelled jobs.

That port is not needed. The transition outbox writes ClickHouse in ClickHouse's own convention (`ch_status/1` maps `cancelled` back to `completed`), so `terminal_completions/1` keeps matching cancelled jobs unchanged, both here and after the cleanup follow-up in #12050 removes the direct ClickHouse writes. The query is also analytics-shaped rather than control-plane: a bounded id list with no ordering requirement, and the session arm only considers sessions at least 15 minutes old, so the sub-minute outbox lag is immaterial. Leaving it on ClickHouse is both correct and less work.

### Interaction to plan for

One mechanical collision: #12295 renames `PodClaimReconciliationWorker` to `PodReconciliationWorker` (file, module, cron entry), and #12050 edits that same file to delete the CH-before-PG choreography. Rename versus modify, resolved by whichever lands second. `prom_ex_plugin.ex` and `oban/runtime_config.ex` are touched by both in different regions.

Suggested order: #12295 first. It is smaller, self-contained, and fixes something actively bleeding (7,092 orphaned rows billing 42,552 clamped hours). This stack then rebases over it, which the two `main` syncs already on this branch show is routine.

Unrelated to this stack, but worth flagging for #12295's own rebase: its diff still references `Jobs.p95_concurrent_last_hour/1` in three places, and `main` deleted that function in #12121.

server/lib/tuist/runners/workflow_jobs.ex
```diff
@@ -0,0 +1,701 @@
...
+  def record_execution(runner_name, executed_workflow_job_id, account_id)
+      when is_binary(runner_name) and runner_name != "" and is_integer(executed_workflow_job_id) and
+             is_integer(account_id) do
+    Repo.update_all(
+      from(j in WorkflowJob, where: j.workflow_job_id == ^executed_workflow_job_id and j.account_id == ^account_id),
+      set: [runner_name: runner_name]
+    )
+
+    Repo.update_all(
+      from(j in WorkflowJob, where: j.runner_name == ^runner_name and j.account_id == ^account_id),
```
esnunes: `WorkflowJobs.record_execution/3`'s second statement is `UPDATE … WHERE runner_name = $1 AND account_id = $2`, and it runs on every `workflow_job.in_progress` delivery via `dispatch.ex:262`. The migration creates six indexes and none leads with `runner_name`, so the plan falls back to the `[:account_id]` index with a filter — a walk over every lifecycle row the account has ever produced.

What makes this different from the sibling tables: both of them got the index for this exact lookup (`create index(:runner_claims, [:runner_name])` and `create index(:runner_sessions, [:runner_name])` in `20260716120000_add_runner_execution_attribution.exs:35,47`), and `runner_claims` would survive without one anyway because claims are deleted on release. `runner_workflow_jobs` is the one table of the three that keeps rows for the account's lifetime, so the scan grows monotonically on a synchronous webhook path.

Would you add `create index(:runner_workflow_jobs, [:account_id, :runner_name])`? And while that line changes — the update has no status filter, so it stamps terminal rows too. `github_runner_name/1` appends 32 bits of randomness, so a repeat within one account is unlikely rather than impossible; restricting to `@live_statuses` would make it moot for free.

server/data-export.md
```diff
@@ -61,12 +61,13 @@ Sensitive authentication data (passwords, tokens) are excluded from exports.
...
+- **Runner volume HEAD** (Postgres `runner_volume_heads`): the canonical pointer to an account's current warm cache set for the Mac runner fleet, so a host whose local cache master lags can converge to the latest. Columns: `id` (PK), `account_id`, `volume_name` (`tuist-cache` reserved), `generation` (monotonic version counter bumped on each promote), `tree_digest` (the sorted cache entry-name hash that fingerprints the warm set — a digest, not cache content), `node_name` (the host that last published the HEAD — infrastructure identifier, observability only), `inserted_at`, and `updated_at`. One row per (account, volume), last-writer-wins on promote. Coordination metadata only — no build content, cache artifacts, or credentials.
```
esnunes: This looks like a merge-resolution casualty rather than an intentional change: the diff removes the whole `Runner cache volume master orphans` (`runner_volume_master_orphans`) bullet and rewrites the `Runner volume HEAD` bullet back to "bumped on each promote … last-writer-wins on promote". Both lines came from `main` and neither relates to this PR — the `main` merge at f5cd7a8b7c seems to have taken the branch's pre-merge copy.

The table is still live on this branch (`20260722120000_create_runner_volume_master_orphans.exs`, `lib/tuist/runners/volume_master_orphans.ex`, `PruneVolumeMasterOrphanWorker`), and `volume_heads.ex` documents fast-forward promote guarded on `base_generation` — so the reverted wording describes semantics the code no longer has.

Since CLAUDE.md makes `data-export.md` the inventory a GDPR/CCPA export is assembled from, a table missing from it is a table that quietly falls out of an export. Re-applying main's two lines verbatim (keeping your two new bullets) should be all it needs — and might be worth a quick pass over the rest of the file for other casualties from the same merge.

server/lib/tuist/runners/workflow_jobs.ex
```diff
@@ -0,0 +1,701 @@
...
+  never touches this table. While old and new pods overlap (a roll,
+  or a rollback and roll-forward), a job the old code completed keeps
+  its Postgres row wherever it was — `queued` churns dispatch, and
+  `claimed`/`running` looks live to the recovery scans — until this
+  closes it with the completion's own conclusion.
+  """
+  def close_completed do
+    stale =
+      Repo.all(
+        from(j in WorkflowJob,
```
esnunes, line 451: `close_completed/0` uses `j.status in ^@live_statuses`, which Ecto compiles to `status = ANY($1)` with the list bound as one parameter. Its two siblings write the status as a literal (`j.status == "claimed"` at :469, `j.status == "queued"` at :497), which Ecto inlines and which Postgres can prove implies `runner_workflow_jobs_live_index`'s predicate.

Postgres can only match `= ANY($1)` against a partial predicate when the parameter has been const-folded — i.e. in a custom plan. Under `plan_cache_mode = auto` a prepared statement can flip to a generic plan, where the parameter survives, the partial index stops being provable, and what's left is a scan of the full `runner_workflow_jobs` joined against the full `runner_job_completions` — neither of which has a retention path.

In practice Postgres will usually keep the custom plan here because the generic estimate is so much more expensive, so I'd call this latent fragility rather than something that bites on day one. But the migration comment at lines 42-45 says the partial index exists precisely so these sweeps don't walk the full history, and this is the one sweep whose SQL shape can defeat it. Writing the three statuses as literals removes the ambiguity for one line.

server/lib/tuist/runners/claims.ex
```diff
@@ -102,6 +103,13 @@ defmodule Tuist.Runners.Claims do
                {:ok, claim} <- insert_claim(workflow_job_id, account.id, fleet_name, pod_name, resources),
                {:ok, limit} <- try_lock_concurrency_limit(account.id, resources.platform),
                :ok <- check_reserved_capacity(account.id, limit, resources) do
+            # Same transaction as the claim insert, so the claim and
+            # the lifecycle row's `queued → claimed` commit or roll
+            # back together. `:noop` (row missing or not queued) never
+            # fails the claim — during dark writes the ClickHouse view
```
esnunes: A few leftovers from the drift-comparator removal in 230e3a7fee.

The main one: the comment above `transition_claimed/3` in `Claims.attempt/5` says a `:noop` "never fails the claim — during dark writes the ClickHouse view stays authoritative and the drift comparator measures the misses." Neither half holds on this branch — `pick_queued_top_k/5` now delegates to Postgres, and the comparator is gone.

That's worth more than a comment fix, because a `:noop` now means the claim landed while the lifecycle row didn't move, and the return value is discarded at both `claims.ex:112` and `claims.ex:311`. As far as I can tell the only thing that would notice is `sync_claimed_from_claims/0`, which is itself slated for deletion. A counter or log on `:noop` would make the miss rate visible for near-zero cost — and it'd be a useful signal to have during the bake.

Much smaller, same origin: `config.exs:200-202` adds `:compared` and `:samples` to the logger metadata allowlist and I couldn't find either emitted anywhere in `lib/` any more (`:kind` is genuinely used by the reconciler). And the `describe` at `jobs_test.exs:1434` is still named "postgres lifecycle dark writes".

server/lib/tuist/runners/claims.ex
```diff
@@ -292,11 +300,16 @@ defmodule Tuist.Runners.Claims do
...
-      )
+    {:ok, _} =
+      Repo.transaction(fn ->
+        {_count, _} =
+          Repo.update_all(
+            from(c in Claim, where: c.workflow_job_id == ^workflow_job_id and c.lifecycle_state == "claimed"),
+            set: [lifecycle_state: "running", runner_name: runner_name]
+          )
+
+        WorkflowJobs.transition_running(workflow_job_id, runner_name)
```
esnunes: `mark_running/2`'s new transaction pairs a claim `UPDATE` guarded only on `workflow_job_id` + `lifecycle_state == "claimed"` with a lifecycle CAS guarded only on `status == "claimed"`. Neither is bound to the caller's claim handle or Pod name.

So if `StaleClaimsWorker` released a slow Pod's claim and another Pod re-claimed the job, a late `mark_running` from the first Pod matches the *new* claim: it stamps the losing Pod's `runner_name` on both stores and moves the row to `running` with `started_at = now`. Since `record_execution/3` matches rows by `runner_name`, the following `in_progress` webhook would then attribute the execution to the wrong job.

The claim-side half of this predates the PR — what's new is extending the same unguarded semantics to the lifecycle row, which is now the store dispatch reads. Threading the `claimed_at` handle through and guarding on it (as `release/2` and `requeue_by_handle/2` already do) would make both writes agree on which claim they're for.

server/lib/tuist/runners/workflow_jobs.ex
```diff
@@ -0,0 +1,701 @@
...
+        completed_at: completed_at
+      ) == :ok
+    end)
+  end
+
+  @doc """
+  Re-queues `claimed` lifecycle rows that no live claim backs. Returns
+  the number of rows transitioned.
+
+  Under this release a claim and its `claimed` row are created and
```
esnunes: `requeue_unbacked_claimed/0`'s docstring says "Under this release a claim and its `claimed` row are created and released in one transaction, so the state cannot arise on its own." But `Claims.release_by_pod_name/1` is unmodified and deletes the claim with no lifecycle transition, on every pod-stopped report from `runner_pods_controller.ex:105` — and this module's own moduledoc at lines 120-127 says as much. The two statements are in the same file and disagree.

The reason it matters: a pre-mint Pod that gets evicted leaves a `claimed` row no sweep can see — not `queued` for dispatch, not `running` for the orphan scans, and no claim for `StaleClaimsWorker`. This pass is the only thing that recovers it today.

Your PR description does say the cleanup PR closes it at the source, so I read this as sequencing rather than a defect. The thing I'd flag is that the docstring is exactly what a future reader will lean on to justify deleting the pass. Worth correcting it, or moving this one pass out of the transitional worker so it survives the cleanup?

esnunes (approved):
This is a genuinely good piece of work, and the framing in the description made a diff this size far easier to review than it had any right to be. The structural argument holds up: once the claim mutation and the status transition are one transaction, the divergence class the recovery workers exist to reconcile stops existing rather than being healed after the fact. The CAS discipline is applied consistently (guard misses as `:noop`, terminal as absorbing, the handle guard on the release paths), and landing the transitional reconciler *with* the cutover rather than after it is the right call — each of its four passes maps to a concrete old-code interleaving with a test that reproduces it.

Approving. What I've left inline is a small set of things worth a look, none of which change the shape of the design.

**Index and plan shape on the new Postgres reads.** `record_execution/3` runs an unindexed UPDATE by `runner_name` on a never-pruned table, on the `in_progress` webhook path — and both sibling tables got exactly that index in `20260716120000_add_runner_execution_attribution.exs` for this same lookup. Separately, `close_completed/0`'s pinned status list compiles to `= ANY($1)`, which a generic plan can't prove implies the live partial index; its two siblings use literals and are fine.

**Two guards worth tightening.** `mark_running/2`'s new lifecycle transition isn't bound to the caller's claim handle, so a late mint from a claim that was reaped in the meantime can promote another Pod's claim generation. And `requeue_unbacked_claimed/0`'s docstring says the state it heals can't arise under this release, which `Claims.release_by_pod_name/1` contradicts — worth correcting, since that docstring is what a future reader will lean on when deleting the pass in the cleanup.

**A little residue from the drift-comparator removal** — a comment in `Claims.attempt/5` that still describes the pre-cutover world, plus two logger metadata keys nothing emits any more.

Separately, and unrelated to the runner work: the `main` merge looks like it took the branch's side on `data-export.md`, deleting the entry for a table that's still live and reverting the volume-HEAD description to semantics the code no longer has. That's the one I'd fix before merging, since that file is the inventory a data export is assembled from.

fortmarek:
Thanks — all six addressed in 46f6d57. Taking them in the order you raised them:

**`record_execution/3` index.** Added `create index(:runner_workflow_jobs, [:account_id, :runner_name])`. You're right that this table is the odd one out: `runner_claims` survives without it because claims are deleted on release, and this one keeps rows for the account's lifetime, so the scan grows monotonically on a synchronous webhook path.

Also took the free win you pointed at: the `runner_name`-keyed update is now bounded to `@live_statuses`. A terminal row has nothing left to attribute, so restricting it makes the 32-bits-of-randomness collision moot rather than merely unlikely.

**`close_completed/0`'s `= ANY($1)`.** Written as literals, matching its two siblings. I agree it reads as latent fragility rather than a day-one problem, but the migration comment does promise those sweeps stay off the full history, and this was the one statement whose shape could quietly break that promise.

**`mark_running/2`'s missing handle.** This was the one worth the most, so it is now `mark_running/3` and both writes are guarded on `claimed_at`. Your read of the consequence is what makes it more than hygiene: `record_execution/3` resolves rows by `runner_name`, so a late mint stamping the losing Pod's name on the winner's row misattributes the following `in_progress` webhook to the wrong job. Covered by a test that reaps a claim, lets another Pod re-claim, then lands the first Pod's mint and asserts neither store moved.

The 26 incidental call sites in tests go through a `mark_running!/2` helper that reads the handle back, so the guard is exercised deliberately in one place rather than accidentally everywhere.

**`requeue_unbacked_claimed/0`'s docstring.** You offered correcting it or moving the pass out of the transitional worker. I went with a third option: making the docstring true. `release_by_pod_name/1` now re-queues the row in the same transaction when the claim is still `claimed`, so the state genuinely cannot arise under this release. A `claimed` claim is pre-mint by construction, so nothing can be running for that job anywhere and no GitHub cross-check is needed — `running` claims still go through `OrphanedRunnersWorker`'s targeted recovery, unchanged. That also means the pass stays purely transitional and remains safe to delete in the cleanup, which was the thing your comment was really protecting.

**Drift-comparator residue.** Comment rewritten, `:compared` and `:samples` dropped from the logger allowlist, the test describe renamed. On the `:noop` being swallowed: agreed it is worth more than a comment, so the claim path now logs it — it means the claim landed while the store dispatch reads did not move, which is worth seeing during the bake. I left `mark_running/3`'s `:noop` unlogged on purpose: with the handle guard above, a miss there is now the designed-for outcome rather than an anomaly.

**`data-export.md`.** Correct, and worse than the two lines you found — the same merge also reverted the `automation_alerts` rolling-window description and the `test_case_runs_recent_per_case` entry, and dropped `test_case_runs_recent_window_per_case` entirely. Five casualties. Rather than re-apply them individually I reset the file to main's copy and re-added only this PR's two bullets, so the diff against main is now exactly `+2` lines. Thanks for prompting the wider pass; I would have fixed two and left three.

One unrelated thing I fixed while here: the orphan worker's recovery-telemetry assertion now matches on `kind` inside the `assert_receive` pattern. `:telemetry_test` attaches a VM-global handler, so a concurrent async test's recovery event could satisfy the bare assertion and fail on the wrong message — it surfaced once during this run.

712 tests green across the runners suite, the Oban config, the runner LiveViews and controllers; format, credo and `excellent_migrations.check_safety` clean. #12050 has been rebased on top and is green too.

server/lib/tuist/runners/workflow_jobs.ex
fortmarek: Added `create index(:runner_workflow_jobs, [:account_id, :runner_name])`, and took the free win too: the `runner_name`-keyed update is now bounded to `@live_statuses`, so a name collision can't restamp terminal rows.

server/data-export.md
fortmarek: Good catch, and it was worse than the two lines you found — the same merge also reverted the `automation_alerts` rolling-window text and the `test_case_runs_recent_per_case` entry, and dropped `test_case_runs_recent_window_per_case` entirely. Reset the file to main's copy and re-added only this PR's two bullets, so the diff against main is now exactly +2 lines.

server/lib/tuist/runners/workflow_jobs.ex
fortmarek: Written as three literals, matching its two siblings, so the partial index predicate stays provable regardless of plan mode.

server/lib/tuist/runners/claims.ex
fortmarek: Comment rewritten, `:compared`/`:samples` dropped from the logger allowlist, describe renamed. The claim-path `:noop` now logs, since it means the claim landed while the store dispatch reads didn't move; I left `mark_running/3`'s unlogged because with the new handle guard a miss there is the designed outcome, not an anomaly.

server/lib/tuist/runners/claims.ex
fortmarek: Now `mark_running/3`, with both the claim UPDATE and the lifecycle CAS guarded on `claimed_at`. Covered by a test that reaps a claim, lets another Pod re-claim, then lands the first Pod's mint and asserts neither store moved.

server/lib/tuist/runners/workflow_jobs.ex
fortmarek: Took a third option: made the docstring true. `release_by_pod_name/1` now re-queues the row in the same transaction when the claim is still `claimed` — pre-mint by construction, so nothing can be running and no GitHub cross-check is needed (`running` claims still go through the targeted orphan recovery). The pass stays purely transitional and safe to delete in the cleanup.

PR #12036, opened 2026-07-23
feat(server): add automation configuration history

pepicrft:
## What changed

Automation settings include a detail page for the current configuration and its chronological history. Each entry records creation or a meaningful edit with the actor, source, changed fields, previous and new values, and a sanitized configuration snapshot.

The history starts with the five newest entries. When older entries exist, **Show more** is centered across the history card and appends the next page without repeating the revisions already shown. The development seed now provides six changes so the control is visible during review.

Automation configuration history with a centered Show more button

## Why

Teams need to understand how an automation changed while investigating unexpected behavior. A long, unbounded list would make the settings page difficult to scan, while hiding earlier changes would weaken the audit trail.

## Approach

Revision writes remain in the same database transaction as automation creation or update. Unchanged updates do not create entries, and encrypted Slack webhook addresses are omitted.

The history query uses a timestamp and stable identifier cursor, with an index that supports the newest-first order. This keeps each request bounded and avoids duplicates when users load older revisions.

## Impact

Existing automations continue to work unchanged. New and edited automations retain their configuration history, and users can progressively inspect older changes when necessary.

## Validation

- `mix format --check-formatted lib/tuist/automations.ex lib/tuist_web/live/project_automation_live.ex test/tuist/automations_test.exs test/tuist_web/live/project_automation_live_test.exs priv/repo/automation_history_seeds.exs`
- `mix assets.build`
- `mix run priv/repo/automation_history_seeds.exs`
- Signed in as the local test user and verified the detail page in headless Google Chrome at 1440 × 1400 pixels. The initial view rendered five revisions and a centered **Show more** button; selecting it appended the sixth creation record and removed the control.
- `mix test test/tuist/automations_test.exs test/tuist_web/live/project_automation_live_test.exs` could not start the local test database sandbox because the configured ClickHouse instance does not support transactions. No test case executed.


server/lib/tuist/automations.ex
```diff
@@ -54,10 +67,41 @@ defmodule Tuist.Automations do
...
+    )
+  end
+
+  defp limit_alert_revisions(query, limit) when is_integer(limit) and limit > 0 do
+    limit(query, ^limit)
+  end
+
+  defp limit_alert_revisions(query, _limit), do: query
+
+  def create_alert(attrs, opts \\ []) do
```
fortmarek, line 97: Revisions are only written through `create_alert/2`, but the default automation that ships with every project never goes through it.

`Projects.create_project/2` and `Projects.create_project!/2` both call `seed_default_alert/1`, which does `%Alert{} |> Alert.changeset(Automations.default_alert_attrs(project_id)) |> Repo.insert()` directly. So on a fresh project, the single seeded automation has no `created` revision at all: clicking into it renders the "No configuration changes recorded yet" empty state, and after the first edit the history starts with an `updated` entry with no origin record before it.

That's the automation most users will open first, and it contradicts the stated guarantee that each entry records creation or a meaningful edit. Routing `seed_default_alert/1` through `create_alert/2` (with `source: "system"`) would cover it.

server/lib/tuist_web/live/project_automation_live.ex
```diff
@@ -0,0 +1,317 @@
...
+        %{"automation_id" => automation_id},
+        _session,
+        %{assigns: %{selected_project: selected_project, current_user: current_user}} = socket
+      ) do
+    if Authorization.authorize(:automation_alert_read, current_user, selected_project) != :ok do
+      raise TuistWeb.Errors.UnauthorizedError,
+            dgettext("dashboard_projects", "You are not authorized to perform this action.")
+    end
+
+    with {:ok, automation} <- Automations.get_alert(automation_id),
```
fortmarek: A malformed `:automation_id` never reaches the `else` branch.

`Automations.get_alert/1` does `Repo.get(Alert, id)` against a UUIDv7 primary key, so a path segment that isn't a UUID raises `Ecto.Query.CastError` inside the `with` rather than returning `{:error, :not_found}`. `GET /<account>/<project>/settings/automations/not-a-uuid` (stale link, typo, crawler) therefore 400s with a reported exception instead of producing the `NotFoundError` this code is written to raise.

Casting first and letting the failure fall through to the existing else branch fixes it, the same way `ops_database_live.ex` does:

```elixir
with {:ok, _} <- Ecto.UUID.cast(automation_id),
     {:ok, automation} <- Automations.get_alert(automation_id),
     true <- automation.project_id == selected_project.id do
```

server/lib/tuist/automations.ex
```diff
@@ -80,34 +124,98 @@ defmodule Tuist.Automations do
...
+  defp alert_revision_actor_id(opts) do
+    case Keyword.get(opts, :actor) do
+      %{id: id} -> id
+      _ -> Keyword.get(opts, :actor_id)
+    end
+  end
+
+  defp alert_revision_changes(nil, _alert), do: %{}
+
+  defp alert_revision_changes(previous_alert, alert) do
```
fortmarek, line 201: Because `revision_value/2` strips `webhook_url_encrypted` before the comparison on the two lines below, a change that only rewrites an action's webhook is seen as no change and no revision row is written at all.

Concretely: a user re-selects the same Slack channel after the Slack app is reinstalled. The action map keeps the same `channel` and `channel_name` but carries a freshly encrypted `webhook_url_encrypted`. Both sides compare equal, `changes` comes back `%{}`, `insert_alert_revision/5` takes the `{:ok, nil}` short-circuit, and the alert is updated with a new delivery target while the audit trail shows nothing happened.

Keeping the encrypted value out of the stored record is clearly intentional and right. Omitting the fact that it changed is a different thing, and this is the field that controls where alerts are delivered. Comparing on presence or a digest, and recording the entry with redacted from/to values, would preserve both properties.

server/lib/tuist/automations.ex
```diff
@@ -80,34 +124,98 @@ defmodule Tuist.Automations do
...
-    result
+      case insert_alert_revision(Repo, alert, updated_alert, "updated", opts) do
+        {:ok, _revision} -> updated_alert
+        {:error, reason} -> Repo.rollback(reason)
+      end
+    end)
+  end
+
+  defp unwrap_alert_transaction({:ok, %{alert: alert}}), do: {:ok, alert}
+  defp unwrap_alert_transaction({:error, _operation, reason, _changes}), do: {:error, reason}
```
fortmarek: This collapses every failing Multi step into `{:error, reason}`, so a failure in the `:revision` step is handed back to callers that assume the changeset belongs to `Alert`.

`Revision.changeset/2` declares `foreign_key_constraint(:actor_id)`. If the acting user's row is removed between authentication and the insert, the transaction returns `{:error, :revision, revision_changeset, _}`, and `AlertsController.create/2` renders it through `RequestParams.format_errors/1` and replies `422 actor_id: does not exist` for a request that has no `actor_id` field. Narrow, but it's worth matching on the `:alert` step explicitly and mapping other steps to a distinct error so the caller contract stays honest.

fortmarek (commented):
Reviewed the automation configuration history feature. Four findings below, the first is the one I'd want addressed before merge. Everything else I checked held up: the keyset cursor is a correct total order over (inserted_at, id) with no duplicate or skip window and the new index serves it via a backward scan; every Noora icon name used exists; `row_navigate` is a real attr and Noora's table CSS explicitly lifts nested buttons above the overlay anchors, so the edit/delete buttons stay clickable; the `[data-part="automation-history"]` CSS block is a top-level rule, so the Show more centering selector does match the DOM; `authenticated_subject` never puts a non-User in `current_user`, so the new `actor:` opt cannot write a project id into the users FK; `RequestParams.normalize/1` stringifies keys recursively, so the webhook strip really does apply on the API path; the `automation_alerts` columns are all NOT NULL, so `revision_value/2` cannot hit nil; all call sites were updated and no Mimic stub depends on the old arity; and the seed script's org/project lookups match what `seeds.exs` creates, so it stays idempotent.

Separately: the PR body notes `mix test` never actually ran. The new LiveView test depends on `automation_alert_fixture/1` routing through `create_alert/2` (it does, so it should pass), but it would be good to get a green run before merging.

server/assets/app/css/pages/project_automations.css
```diff
@@ -403,3 +403,307 @@
     }
   }
 }
+
+#project-automation {
```
fortmarek, line 407: nit: should this be in its own css file?

fortmarek (approved):

server/lib/tuist/automations.ex
pepicrft: Fixed in 14cab8273d. `seed_default_alert/1` now calls `Automations.create_alert/2` with `source: "system"`, so both project-creation paths write a `created` revision in the transaction. Added an assertion in `server/test/tuist/projects_test.exs` for the seeded revision.

server/lib/tuist_web/live/project_automation_live.ex
pepicrft: Fixed in 14cab8273d. `ProjectAutomationLive.mount/3` now casts the identifier before querying, allowing the existing fallback to raise `NotFoundError`. Added a regression test for the malformed route in `server/test/tuist_web/live/project_automation_live_test.exs`.

server/lib/tuist/automations.ex
pepicrft: Fixed in 14cab8273d. Revision values now replace `webhook_url_encrypted` with a deterministic digest of the encrypted value. That preserves redaction while creating a revision when only the delivery target changes. Added coverage in `server/test/tuist/automations_test.exs`.

server/lib/tuist/automations.ex
pepicrft: Fixed in 14cab8273d. Alert validation errors and revision-write failures now have separate result paths, and the controller returns a generic server error for the latter rather than rendering an unrelated `actor_id` validation error. Added rollback coverage for create and update.

server/assets/app/css/pages/project_automations.css
pepicrft: I kept these styles in `project_automations.css`. That file already owns the automation list and detail views, and the history selectors are scoped to `data-part="automation-history"`. A separate file would add an import without creating a reusable styling boundary.

PR #12050, opened 2026-07-23
refactor(server): feed ClickHouse runner_jobs solely through the transition outbox

fortmarek:
The post-bake cleanup for #12031, prepared now so review can see the end state — **do not merge until the gates below hold**. Stacked on #12031's branch; retarget to main once it merges.

## ⚠️ Merge gates

1. **#12031 must be deployed and have completed at least one full production roll first.** The backfill worker this PR deletes needs one deployed cycle to adopt in-flight jobs into Postgres; shipping both in the same deploy would strand any job enqueued by the pre-#12031 release (Postgres reads would never see it).
2. **The outbox must have stayed drained over the bake window** (suggested: one week of production traffic): `runner_workflow_job_transition_events` near-empty on a spot check, i.e. the flusher has been keeping up. That is the thing this PR makes load-bearing — after it, the outbox is ClickHouse's only feed. Worth one Postgres-vs-ClickHouse status comparison before merging as well; both are single queries.
3. The `runners: adopted ClickHouse-only workflow_jobs into Postgres` and `runners: reconciled workflow_job lifecycle rows` log lines have gone permanently quiet (they fire during a roll and should be silent between deploys).

## What

Everything #12031's "after this rolls into production" section promised, all deletions:

- **Direct ClickHouse lifecycle writes in `Tuist.Runners.Jobs` are gone.** ClickHouse `runner_jobs` becomes a pure replica fed by the transition outbox. `enqueue`/`record_claimed`/`record_running` keep only their guards, telemetry, and PubSub broadcasts; `complete/2` reads the Postgres lifecycle row instead of the CH argMax view; `record_queued/1` is deleted outright — `Claims.release/2` and `release_pod_missing/2` already re-queue the lifecycle row atomically with the claim delete, and `WorkflowJobs.requeue/1` now emits the requeued telemetry + broadcast (the transition's one home). The single surviving direct CH write is `set_log_archived_at/2`: `log_archived_at` is analytics-store state with no Postgres twin, and a terminal row emits no further outbox events that could revert the stamp under argMax.
- **`ReconcileWorkflowJobsWorker`, `Jobs.list_non_terminal/1`, and the `WorkflowJobs` adopt/close/sync/requeue-unbacked passes deleted** — with no CH-first write path left and no code path moving a claim without its lifecycle row, none of the states it healed can arise from this release. The one it can — a `claimed` row whose claim `release_by_pod_name/1` deleted because the Pod stopped between claim and JIT mint — is now closed at the source: `release_by_pod_name/1` re-queues pre-mint `claimed` rows in the same transaction as the delete (a `claimed` claim is pre-mint by construction, so nothing can be running for the job and no GitHub cross-check is needed; `running` claims keep going through the orphan worker's targeted recovery). That gap existed on `main` before this work too, in ClickHouse form.
- **The dispatch anti-list deleted** (`Claims.workflow_job_ids_for_fleet/1` and its seed in the poll loop): a queued lifecycle row cannot carry a live claim, because the claim transaction transitions it to `claimed` — the queued scan and claim state are the same store and cannot disagree. The per-poll lost-race exclusion accumulation stays (it's loop-local bookkeeping, not cross-store defence).
- **The orphan worker's re-queue keeps the handle-guarded fallback** #12031 added for #12395's targeted recovery: `Claims.release/2` on success, else `WorkflowJobs.requeue_by_handle/2` for a claim the pod-stopped report already released. With the direct CH `record_queued` gone, that fallback is the only thing that returns a stranded job to dispatch, so it is load-bearing here rather than belt-and-braces.
- **CH-first choreography dropped from the recovery workers** (`StaleClaimsWorker`, `OrphanedRunnersWorker`, `PodClaimReconciliationWorker`): the `safe_record_queued` calls and their crash-window ordering documentation disappear — a release either fully returns the job to the queue or does nothing.

## What deliberately stays

The outbox + flusher (the permanent CH replication path), the `enqueued_at` lookback floor on the Postgres reads (aligned with `StaleQueuedJobsWorker`'s 24h backstop, not a ClickHouse artifact), all the recovery workers (they heal server↔GitHub and server↔Kubernetes divergence, untouched by this work), and the ClickHouse analytics reads powering the customer-facing surfaces.

## Behavior note

The customer-facing dashboards now lag the control plane by up to a flusher tick (≤1 min) instead of being written synchronously with it. Control-plane decisions all read Postgres; the lag is visible only in the dashboards/analytics — the same tradeoff the ReplacingMergeTree's merge lag already imposed on those surfaces.

## Validation

- 684 tests green (after merging #12031's `main` sync and review fixes) across the runners suite, Oban runtime config, the five runner LiveViews, and the runner controllers. Tests reading the CH surfaces drain the outbox first; claim-path seeding goes through real `Claims`/`WorkflowJobs` transitions instead of the deleted CH writes.
- `mix compile --warnings-as-errors`, credo clean, `mix format --check-formatted` clean.


esnunes (approved):
Really enjoyed reading this one — the deletions are argued rather than asserted, and the arguments mostly hold. I checked the dispatch anti-list removal specifically, since that's the one that would bite hardest if wrong: `Jobs.pick_queued_top_k/5` delegates unconditionally to the Postgres read with no flag left that could route it back to ClickHouse, and `Claims.attempt/5` does the claim INSERT and the queued → claimed transition in one transaction. Same store, same transaction, so the scan and the claim genuinely can't disagree. Dropping `workflow_job_ids_for_fleet/1` looks safe to me.

One note, on `server/test/tuist/oban/runtime_config_test.exs:17`:

> Small one: the two `ReconcileWorkflowJobsWorker` assertions went away here but the alias on line 17 stayed behind, so it's now unused. CI runs `mix test --warnings-as-errors` (`.github/workflows/server.yml:288`), and I confirmed locally that this file's 16 tests pass and then the run aborts on `warning: unused alias ReconcileWorkflowJobsWorker`. Dropping line 17 should be all it needs.

Nothing here changes my read of the overall direction, which I think is right.

Issue #6637, opened 2024-08-13
Build documentation fails when the app target has Firebase dependency


cristi-lupu:
### Problem
Build documentation fails when the app target has Firebase dependency.

### How to reproduce
1. Open fixture `multiplatform_app_with_sdk` . It already has the Firebase dependency on the app target.
2. Run `tuist install` and `tuist generate`.
3. Run `Build Documentation`.
Here is the build error:
<img width="1351" alt="image" src="https://github.com/user-attachments/assets/cf2a96d5-4a28-4551-83b6-c0d4976c281f">

### Potential solution

I tried to create a vanilla sample iOS app Xcode project and add the Firebase as an SPM dependency and the build documentation works as expected.

### macOS version

14.6.1

### Tuist version

4.23.0

### Xcode version

15.4.0

hak-ioki:
I've encountered a similar issue after upgrading Firebase (11.8.1) and migrating its integration from Carthage to Swift Package Manager (SPM) in my Tuist-managed project. The documentation build process fails with errors related to missing Firebase headers, specifically `FIRApp.h` from `FirebaseCore`. 

The problem persists. The application builds and runs successfully; the failure occurs during documentation generation. Any updates on resolving this issue would be greatly appreciated.

Blackjacx:
@cristi-lupu and/or @hak-ioki Tuist recently transitioned from GitHub issues to [their own community website](https://community.tuist.dev/). I would highly suggest to repost the issue over there. 

We're also facing a this issue and cannot build out docs anymore since we moved Firebase from Cartage over to SPM.

PR #12055, opened 2026-07-23
fix(cli): pass framework modulemaps to ExtractAPI

Loupehope:
### Summary

This fixes custom module-map handling for framework package targets in generated Xcode projects.

### Root cause

Tuist used MODULEMAP_FILE to describe a custom source module map. Xcode treats that setting as an instruction to copy the map into the framework product. At the same time, ExtractAPI needs a canonical module-map path when it constructs dependency inputs.

That created two competing representations of the same module map: the original source map and a generated framework copy. Depending on the target type and build action, Xcode could discover both maps, use an invalid dependency path, or require the Copy Module Map phase to run again.

### Solution

For framework and static framework targets, Tuist now resolves MODULEMAP_FILE and assigns the original source map to MODULEMAP_PATH, then removes MODULEMAP_FILE.

This follows Swift Build’s model: ProductPlan uses MODULEMAP_PATH as the module’s built map path, and ExtractAPI registers that path for dependent targets. Removing MODULEMAP_FILE prevents Xcode from copying a second map into the framework product.

The result is one canonical module map for compilation and ExtractAPI. No shell script phase, content inspection, or package-specific handling is required.

### Covered cases

GzipSwift/system-zlib: prevents duplicate module-map discovery during ScanDependencies.

GEOSwift/geos: resolves the ExtractAPI and Copy Module Map dependency issue from https://github.com/tuist/tuist/pull/11901.

FirebaseCoreExtension: restores DocC builds for custom module maps, covering https://github.com/tuist/tuist/pull/11135 and https://github.com/tuist/tuist/issues/6637.

Static frameworks: preserves the intended behavior from https://github.com/tuist/tuist/pull/11588. The module map is not copied into the product, while ExtractAPI receives the required source path.

Cache warming: makes https://github.com/tuist/tuist/pull/11532 unnecessary. That PR made the Copy Module Map phase idempotent, but this change removes the phase entirely, so repeated copies and read-only destination failures cannot occur.

### Validation

Enabled the existing generated_multiplatform_app_with_sdk acceptance test from #6637. It installs dependencies, generates the project, and runs iOS docbuild and build.

Added generated_ios_app_with_modulemap_packages, using GzipSwift, GEOSwift, and FirebaseMLModelDownloader, which pulls FirebaseCoreExtension transitively. Its acceptance test runs docbuild and build.

Ran the existing static-framework acceptance test from #11588. Archive succeeds and verifies that the framework product does not contain Modules/module.modulemap.

Ran ModuleMapMapperTests, covering framework and static-framework MODULEMAP_PATH mapping, removal of MODULEMAP_FILE, relative path resolution, and existing dependency module-map behavior.

### Backport

This may need to be backported to supported previous release lines that contain the affected DocC and custom module-map handling.


Loupehope:
Hi! @fortmarek @pepicrft

I've tried to close a large number of cases using MODULEMAP_FILE

PR #12059, opened 2026-07-24
feat(server): never let a stuck Kura deployment block the next version

esnunes:
Implements [Hive spec #81](https://hive.tuist.dev/specs/81). Two mechanisms, one on each side of the control-plane/data-plane seam: the server half (`feat(server)`) makes a release always schedulable, the controller half (`feat(infra)`) makes it reach the replicas that need it.

## The problem

The Kura version scheduler refused to schedule a new runtime image for a server while **any** deployment row for it was open (`:pending`/`:running`). A deployment stays open until the instance converges end to end — the backing `KuraInstance` reports the target image *and* the region's public endpoint answers `/up`.

Several waits on that path are **unbounded by design**: public endpoint not ready (the pod is up but still replicating behind the `/ready` bootstrap gate), DNS not yet propagated, node-port chain not yet observed. Each leaves the row `:running` and retries next tick, forever.

So an instance stuck converging held its deployment open indefinitely and queued every future release behind it. The sharp edge is circular: **the release carrying the fix for a wedge could not reach the wedged instance.**

This is not hypothetical. The `0.19.0` deployment on `kura-tuist-eu-central-1` stayed open for over **14 hours** (2026-07-21 21:59 → 2026-07-22 12:06) while the instance ground through a bootstrap wedge. Any release shipped in that window queued behind it — and the diagnostic gauges built to confirm that very wedge (#12018) were in exactly this position: instrumentation for a stuck instance, blocked by the stuck instance.

### 1 · Newest intent wins — supersede open deployments (server)

When the scheduler holds a newer released tag than an open deployment's `image_tag`, it closes that row as `:superseded` and schedules the new tag in the **same reconciler tick**.

- New terminal status `:superseded` (5), with `pending`/`running` → `superseded` transitions, plus a migration widening the status check constraint to `0..5`.
- **No index change was needed.** The "one open deployment per server" rule is a *partial unique index* keyed on `status IN (0, 1)`, so closing a row to `:superseded` drops it out of the predicate and frees the slot.
- The cross-release `open_deployment_exists` exclusion is dropped from `servers_needing_version_query/1`. The once-per-tag guard (`deployment_for_image_exists`) stays and already excludes any server holding a row for this exact tag — so the only servers newly admitted are those whose open row carries a *different* tag. Serialization within a single release is preserved; only the cross-release block is removed.
- Supersede and the replacement insert share **one server-locked transaction**, so "closed the old row but scheduled nothing" is not reachable.

Safety follows from the architecture as it already is: the Go controller is level-based, so re-targeting a mid-roll instance is a native rolling-update re-target; Kura's rollout-safety rule already requires adjacent versions to interoperate, so a briefly mixed mesh is inside the existing skew budget; and activation semantics are unchanged — supersede changes which image an open rollout converges *toward*, never skipping the convergence gate.

### 2 · Failing replicas roll together (controller)

Supersede puts the new tag on the `KuraInstance` spec, but the StatefulSet's default `RollingUpdate` walk still delivers it one readiness-gated ordinal at a time — and **not at all** below a pod that stays not-Ready. On a degraded instance the fix crawls, gated on a bootstrap that may itself be the bug.

After the StatefulSet is updated, the controller now deletes the instance's not-Ready pods still running a superseded image. The StatefulSet recreates them, and because pod management is `Parallel` the recreations start together.

The predicate carries the safety argument:

- A not-Ready pod serves **no load-balanced traffic** (readiness gates routing), so deleting it costs nothing currently being provided. Its data PVC persists; only in-memory bootstrap progress *on the image being displaced* is lost.
- **Ready pods are never touched** — they keep serving and roll through the ordinary readiness-gated walk.
- A pod already on the desired image is skipped even while bootstrapping, so a pod making progress on the fix isn't thrown back to the start, and a broken fix can't thrash.
- A pod whose kura image can't be read is skipped: deletion rests on *knowing* the pod runs the displaced version.

## Why not the obvious alternatives

- **A TTL on open deployments.** Still serializes the new release behind the wedged one until expiry, and pressures the timeout downward toward misfiring on legitimate long convergences — the 14-hour bootstrap above eventually *succeeded*, and a TTL short enough to be useful would have killed it. Supersede needs no timer: the trigger is the thing that actually matters, a newer release existing.
- **Raising `maxUnavailable`.** Widens the update window for *every* rollout, healthy or not, trading steady-state availability for the degraded case. Deleting only not-Ready pods is scoped to exactly the pods already providing nothing.
- **Destroying stuck servers to ship fixes.** Loses the PVC dataset and costs hours of cold re-bootstrap on a component whose entire value is having the data local, and turns a routine release into an operator intervention.

## Impact

Control-plane intent becomes independent of data-plane convergence: a wedged instance can always receive the fix that unwedges it, and the pods most in need of a fix are the first to get it. Nothing that is currently serving is disturbed.

Superseded rows stay in the deployment history with the displacing tag recorded, so /ops reads the real sequence of intent — scheduled, displaced, rescheduled, converged — instead of losing the attempt. **A server accumulating superseded rows without ever converging is a new, visible signal**: releases are flowing, the instance is not. That is precisely what used to hide inside one silently open deployment.

Observability on both halves:

| Signal | Where |
| --- | --- |
| `tuist_kura_deployments_superseded_total{region, provisioner_node_ref, displacing_image_tag}` | server (new PromEx plugin) |
| `kura_controller_not_ready_pod_deletions_total{namespace, instance}` | kura-controller |
| Structured log per close, with displaced + displacing tags | server |

## Out of scope (deliberately)

- **Failed deployments.** A closed `:failed` row already doesn't block newer releases; whether a failed tag is re-attempted is a separate concern.
- **The warm-handoff deferral.** `move_phase == :none` still defers; that guardrail is untouched.

## Validation

- **Elixir:** 119 tests green across the kura, deployment, prom_ex_plugin, reconciler, and ops-live suites — including new coverage for superseding an older-tag row, leaving a same-tag row alone, the retained history, and the telemetry contract. `mix compile --warnings-as-errors`, `mix credo`, and `mix excellent_migrations.check_safety` all clean.
- **Go:** full kura-controller suite green with six new tests covering the delete case, Ready pods, pods already on the desired image, and the unreadable-image and already-terminating skips (each asserting the counter). `go vet` and `gofmt` clean.
- Reviewed in a clean-context pass before opening: no correctness bugs found; the three actionable findings (a comment that justified the supersede's race-safety with the wrong mechanism, the missing PromEx consumer for the superseded event, and the two untested skip branches) are fixed in this branch.


esnunes:
Closed in favor of https://github.com/tuist/tuist/pull/12099

Issue #9967, opened 2026-03-21
`tuist generate` hangs indefinitely downloading binary artifacts on CI, while `swift package resolve` succeeds in the same environment


marioortizmanero:
### What happened?

`tuist generate` hangs indefinitely on CI during the "Downloading binary artifact" phase of SPM dependency resolution. The process never completes and must be killed by a timeout.

This happens 100% of the time on CI (tested on both GitHub Actions `macos-15` runners and WarpBuild runners), but never locally. The hang occurs regardless of which binary-artifact packages are included. We tested removing Facebook SDK, OneSignal, and other packages individually, and the hang persists as long as any binary artifacts need to be downloaded.

The critical finding: running `swift package resolve` in the `Tuist/` directory **before** `tuist generate` completes successfully (all binary artifacts download in under 2 seconds), and then `tuist generate` succeeds because the `.build` directory is already populated.

This means the issue is specific to how Tuist invokes SPM's binary artifact download, not SPM's CLI itself.

### How do we reproduce it?

`tuist generate` should download binary artifacts and complete, just like `swift package resolve` does in the same environment.

**How do we reproduce it?**

1. Create a Tuist project with several SPM dependencies that use binary targets (xcframeworks). In our case: Facebook iOS SDK (6 xcframeworks), OneSignal iOS SDK (10 xcframeworks), Amplitude (2 xcframeworks), and AnalyticsConnector (1 xcframework). Total of ~19 binary artifact downloads.
2. Run the project on a CI runner (GitHub Actions `macos-15` or any fresh macOS environment with no SPM cache).
3. Run `tuist generate --no-open --path iosApp --verbose`
4. Observe that the command hangs after printing "Downloading binary artifact ..." lines and never progresses.

Workaround:

Running `swift package resolve` in the `Tuist/` subdirectory before `tuist generate` resolves the issue:

```yaml
- name: Pre-resolve SPM dependencies
  run: cd iosApp/Tuist && swift package resolve

- name: Install dependencies
  run: cd iosApp && tuist generate --no-open --verbose .
```
Additional context:

- `curl` can download the binary artifact URLs on the runner with no issue (tested, completes in ~0.2s).
- `swift package resolve` in the same `Tuist/.build` directory downloads all 19 binary artifacts successfully in under 2 seconds.
- The hang is 100% reproducible on CI, never occurs locally (likely because local machines have binary artifacts cached in `~/Library/Caches/org.swift.swiftpm/`).
[...]

PR #12069, opened 2026-07-24
fix(cli): bound and retry SwifterPM binary-artifact downloads

fortmarek:
### Purpose

`tuist install` on a self-hosted macOS runner hung for ~6 hours during Swift Package Manager dependency resolution and was then cancelled at the CI job ceiling. Git resolution for ~40 packages completed in about two minutes; the hang was entirely in the binary-artifact download phase, where roughly ten archives began downloading concurrently and one never completed or errored. Every small request in the same job (git fetches, manifest dumps, version resolution) succeeded, so the network path was healthy: a single transfer wedged on a dead-but-open connection.

Root cause: SwifterPM's `HTTPClient` downloaded binary artifacts and source archives through `URLSession.shared`, whose default configuration has a 7-day `timeoutIntervalForResource` and only an idle per-request timeout, with no retry. A stalled connection (no reset from an egress middlebox or CDN edge) therefore hangs effectively forever, and because artifact downloads run behind a concurrency barrier, one wedged transfer blocks the whole `install`.

### What changed

`swifterpm/Sources/swifterpm/Support.swift`:

- Route all HTTP through a dedicated `URLSession` (instead of `.shared`) whose configuration bounds both the idle gap between received bytes (`timeoutIntervalForRequest` = 60s) and the total transfer (`timeoutIntervalForResource` = 600s). A stall now surfaces as an error in about a minute rather than hanging open-ended.
- Retry transient transport failures (timeouts, dropped or half-open connections) on a fresh connection with linear backoff, up to three attempts. Reconnecting is what actually clears a wedged or half-open stall. HTTP status failures (4xx/5xx) are not retried, since a non-2xx response is deterministic for a given URL.

This is centralized in `HTTPClient`, so it covers binary-artifact downloads, the GitHub tarball-archive path, and the smaller metadata GETs. Net effect: the worst case goes from an open-ended multi-hour hang to a bounded, retried failure of a few minutes, and in practice most transient stalls recover transparently on the retry.

### Alternatives considered

Preferring HTTP/1.1 for artifact downloads would also help: profiling these CDNs from a runner measured HTTP/2 at 2.5x to 14x slower than HTTP/1.1 even on a healthy path, and HTTP/2's single-connection multiplexing is what turns a transient packet loss into a full head-of-line stall. URLSession exposes no public API to disable HTTP/2, so that lever is not addressed here. The available way to get it would be to route downloads through a `curl --http1.1` subprocess (SwifterPM already shells out to git via `SystemProcess`), which can be a follow-up if we want the protocol control and the speedup.

### How to test locally

There is no deterministic way to force the upstream stall, but the change is on the normal resolution path:

- From `swifterpm/`: `swift build --target SwifterPMCore` builds clean, and `swift test --filter SupportTests` passes (22 tests).
- Any SwifterPM resolve that downloads binary artifacts (a package graph with `binaryTarget(url:checksum:)` dependencies) exercises the new session and retry path. A stalled artifact host now fails within the bounded timeouts and retries instead of hanging.

Context: related to the earlier report in #9967 (a different, xcodebuild-path case that was closed as upstream).


pepicrft (approved):

mclaughj:
👋 hey team, we'd love to see this one get backported to the 4.203.x release line, it's the only thing preventing us from adopting SwifterPM at the moment. (great work on that btw, looking really good in our initial benchmarking)

Issue #12038, opened 2026-07-23
Reduce ClickHouse merge memory and ingestion fan-out for rolling test automations


pepicrft:
## Context

Rolling test automations currently maintain five global ClickHouse aggregate tables containing the latest 100, 250, 500, 750, and 1,000 physical run rows for every project and test case. Each table has separate flaky and successful materialized views.

Every `test_case_runs` insert therefore executes ten materialized views and writes up to ten aggregate states, even when no automation reads most of them.

A background merge of `test_case_runs_recent_750_per_case` recently exhausted the production ClickHouse memory limit for roughly 13 hours. The runner autoscaling query reported in [Sentry](https://tuist.sentry.io/issues/135949079/) was a victim of global memory pressure, not the source.

## Impact

During the incident:

- A merge of only 100 to 130 mebibytes of compressed input expanded to 8 to 9 gibibytes in memory.
- The merge failed 1,971 times across three compute replicas.
- 1,370 foreground ClickHouse queries failed while the merge kept the service near its 18-gibibyte memory limit.
- Sentry recorded 220 errors in 14 days whose event text explicitly names the 750-run table, including ingestion buffer failures and unrelated reads. [Sentry query](https://tuist.sentry.io/explore/discover/homepage/?dataset=errors&queryDataset=error-events&query=environment%3Aprod+%22test_case_runs_recent_750_per_case%22&project=4510777094373456&field=issue&field=title&field=count%28%29&sort=-count%28%29&statsPeriod=14d&mode=aggregate&yAxis=count%28%29)

The large merge eventually completed on one compute replica, but the table shape has not changed, so this can recur as new parts accumulate.

## Root cause

The rolling tables use unpartitioned `AggregatingMergeTree` storage with large `groupArraySorted` or `groupArrayLast` states.

During a merge, ClickHouse has to deserialize and combine those arrays. Memory therefore scales with the number of logical keys in the merge, the retained array capacity, and the number of aggregate columns. Compressed source-part size is a poor predictor of peak memory for this data shape.

All five tables are structurally exposed because no partition boundary limits how many project and test-case keys a merge may combine. The 1,000-run table is currently the largest at 15.86 gibibytes uncompressed.

The replacement proposed here creates a fixed number of independently growing merge domains. It reduces the blast radius substantially, but it is not a mathematical upper bound on merge memory because the number of keys inside each partition can continue growing. The 10-times and 100-times load replays are therefore release gates, not optional validation.

The current implementation is in:

[...]

PR #12086, opened 2026-07-26
perf(server): add test case run project lookup

pepicrft:
Part of #12038.

## What changed

This adds a slim ClickHouse table that maps each test case run identifier to its project identifier and insertion timestamp, ordered by run identifier. A materialized view keeps new rows synchronized from `test_case_runs`, and a partition-wise migration backfills existing rows.

The materialized view is created before the backfill starts. Writes that arrive while historical partitions are copied therefore reach the lookup immediately, while `ReplacingMergeTree(inserted_at)` collapses overlap between the live path and the backfill.

## Why

The attachment-retention fix needs to resolve pages of legacy test case run identifiers without joining or repeatedly scanning the full `test_case_runs` table. That table is ordered by project, test case, and run time. Random identifier lookups cannot use that sorting-key prefix and read most of the table even when the identifier bloom filter is present.

## Root cause

The old retention query resolved ownership through a join against `test_case_runs`. Production executions read as many as 29.1 million rows. Replacing that join with 500-identifier lookups against the same source table would reduce peak memory but repeat a near-full scan for every page, producing more total work than the query it replaced.

## Approach

The lookup contains only the three columns needed to preserve replacement semantics and resolve ownership. Ordering it by run identifier lets ClickHouse seek directly to the relevant granules. A separate table is used because projections are not reliable on the source `ReplacingMergeTree` shape used here.

The migration is retry-safe. The table and materialized view use `IF NOT EXISTS`, historical data is copied one source partition at a time, and replayed rows collapse by identifier and insertion timestamp.

## Impact

This pull request does not change application reads or user-visible behavior. It adds a small amount of write and storage overhead for three derived columns. The next stacked pull request must not merge until this migration has completed in production.

## Validation

- Added a destructive ClickHouse integration test covering historical backfill, live materialized-view ingestion, and a repeated migration run.
- The integration test passes against the repository-pinned ClickHouse version.
- `mix compile`: passed.
- `git diff --check`: passed.
- `mix excellent_migrations.check_safety`: no new warning for this ClickHouse migration.


pepicrft:
Closing this because it is no longer required for the ClickHouse incident rollout. #12087 now bounds modern retention reads while preserving historical attachments, so the 2.42-billion-row ownership backfill is not justified on the incident path. Historical cleanup can return as a separate operational project after attribution, retention, and customer-impact review.

PR #12088, opened 2026-07-26
fix(server): enable legacy test attachment retention

pepicrft:
Part of #12038.

This is the final activation step after #12086 and #12087. Keep it as a draft until the lookup migration and dry-run stage have fully deployed and been observed in production.

## What changed

Newly scheduled legacy test attachment sweeps now run with deletion enabled. The implementation and correctness guards remain unchanged from #12087.

## Why

#12087 intentionally ships the global historical sweep in dry-run mode. A separate activation change keeps the destructive transition reviewable and makes the production gate explicit.

## Approach

Merge this pull request only after a complete dry run confirms:

- zero missing test-case-run ownership lookups;
- deleted-project counts are understood;
- page-level rows and bytes remain bounded;
- resolved object and account counts are plausible;
- the sweep reaches the end of its frozen time band without repeated failures.

A dry-run job already in progress keeps `dry_run: true` in its persisted job arguments. Only a newly scheduled sweep after this deployment will delete objects.

The enabled worker still deletes objects before it advances the PostgreSQL cursor. It stops on unresolved ClickHouse ownership, skips and reports already-deleted PostgreSQL projects, compares the locked cursor state before every update, and uses half-open time bands to avoid gaps or overlap.

## Impact

Expired legacy attachment objects become eligible for permanent deletion. ClickHouse attachment metadata is retained. Rolling this change back stops future destructive sweeps, but objects already deleted from storage cannot be recovered.

Modern attachment retention and its retention rules are unaffected by this activation switch.

## Validation

- `mix format` completed successfully.
- `mix compile` completed successfully.
- `mix test test/tuist/storage/workers/schedule_expired_artifacts_worker_test.exs` passed 11 tests against the repository-pinned ClickHouse version.
- The underlying #12087 retention suite passed 33 tests covering exact time boundaries, retained and expired separation, account isolation, object deletion failures, missing ownership, deleted projects, persisted progress, stale retries, and exact-multiple pagination.
- An independent Claude Opus review returned `SHIP` for the underlying implementation and identified the production dry run as the required activation gate.
- `git diff --check` completed successfully.


pepicrft:
Closing this because historical attachment deletion is no longer part of the ClickHouse incident rollout. #12087 preserves those attachments while fixing the recurring broad query. We can revisit deletion separately after safe attribution and retention review.

PR #12115, opened 2026-07-28
feat(cli): add bundle support to AppClip

pepicrft:
Supersedes #11597 (same changes, re-enabled acceptance test, pushed from a non-fork branch so CI runs fully).

## What changed

When declaring a `.bundle` target and making an `.appClip` depend on it, Tuist rejected the graph with a lint error even though App Clips fully support resources and can embed bundles. The root cause was a one-line omission in `GraphLinter.validLinks`: PR #6415 added `.bundle` as a valid dependency for `.appExtension` but the same entry was never added for `.appClip`. The rest of the pipeline (`resourceBundleDependencies`, `supportsResources`, `generateResourceBundle`) already handles AppClips correctly — only the linter was blocking it.

This PR:
- Adds `LintableTarget(platform: .iOS, product: .bundle)` to the `.appClip` entry in `GraphLinter.validLinks`.
- Adds a unit test `lint_appClip_canDependOnBundle` next to the existing `lint_appExtension_canDependOnBundle`.
- Adds a Bundle target to the `generated_ios_app_with_appclip` fixture (identical to the one in `generated_ios_app_with_extensions`).
- Re-enables the `ios_app_with_appclip` acceptance test (disabled in #10128 alongside many others during CI consolidation). The analogous `ios_app_with_extensions` test was kept enabled, so this one should be too. While re-enabling, cleaned up duplicate assertions (AppClip and Framework embedding were each checked twice) and narrowed the build to `BuildCommand "App"` to match the extensions test pattern.

## How to verify

The unit test `GraphLinterTests/lint_appClip_canDependOnBundle` covers the linting rule. The acceptance test `GenerateAcceptanceTestiOSAppWithAppClip/ios_app_with_appclip` covers the end-to-end generation + build + resource embedding.

### DIFF
--- a/cli/Sources/TuistGenerator/Linter/GraphLinter.swift
+++ b/cli/Sources/TuistGenerator/Linter/GraphLinter.swift
@@ -558,6 +558,7 @@ public struct GraphLinter: GraphLinting {
             LintableTarget(platform: .iOS, product: .framework),
             LintableTarget(platform: .iOS, product: .staticFramework),
             LintableTarget(platform: .iOS, product: .appExtension),
+            LintableTarget(platform: .iOS, product: .bundle),
             LintableTarget(platform: .macOS, product: .macro),
         ],
         LintableTarget(platform: .iOS, product: .extensionKitExtension): [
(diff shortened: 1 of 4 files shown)


pepicrft (commented):
Same changes as #11597 (already reviewed and approved there). The additional commit re-enables the `ios_app_with_appclip` acceptance test and cleans up duplicate assertions. CI Linux checks are green; macOS checks are queued waiting for runners.

PR #12116, opened 2026-07-28
feat(cli): strip Linux release binary

Loupehope:
## Summary

Strip unneeded symbols from Linux tuist release binaries before packaging and generating checksums.

The macOS release bundle already strips its Mach-O binaries in [#12015](https://github.com/tuist/tuist/pull/12015). This applies the corresponding ELF-safe operation to both Linux release binaries using `strip --strip-unneeded`.

## Why

The static musl binaries contain debug information and symbols that are unnecessary at runtime. Stripping them substantially reduces the size of the Linux download artifacts without changing runtime behaviour.

## Binary size impact

| Architecture | Binary, original | Binary, stripped | Binary reduction | Gz archive, original | Gz archive, stripped | Gz archive reduction |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| x86_64 | 331.3 MiB | 100.8 MiB | 69.6% | 105.1 MiB | 38.7 MiB | 63.2% |
| aarch64 | 332.0 MiB | 93.4 MiB | 71.9% | 103.1 MiB | 36.6 MiB | 64.5% |


pepicrft:
@Loupehope keep that great work coming :)

PR #12118, opened 2026-07-28
perf(infra): skip cache-volume uploads that cannot win the promote

fortmarek:
Runner cache-volume promotes upload the entire image **before** the server arbitrates the fast-forward, so most uploads are wasted work that blocks VM teardown. This adds a pre-flight generation check so a promote that cannot win never transfers anything.

### The problem

`report_volume_head` in `infra/runner-image/dispatch-poll.sh` did:

1. POST to mint a content-addressed presigned upload URL (keyed by the new inventory digest)
2. `curl -X PUT --upload-file "${CACHE_IMAGE}"` — the WHOLE image, timed into `volume-upload-ms`
3. Only then the server's compare-and-swap (`bump_head`) returns accepted / conflict / error

So a promote that loses the fast-forward race still paid the full upload cost. That upload blocks guest teardown, which blocks VM halt and runner slot reclaim.

Measured across three converged production hosts (`tart_kubelet_cache_volume_*`):

| host | upload mean | promote accepted / rejected |
|---|---|---|
| `2t7nk` | 89.8s (538.8s / 6) | 11 / 16 |
| `8n9bz` | 52.9s (317.6s / 6) | 2 / 13 |
| `c22td` | 44.0s (176.1s / 4) | 6 / 16 |

Fleet-wide: **19 accepted vs 45 rejected, so ~66% of promotes lose**, at a mean upload of ~65s (up to 90s). Roughly two thirds of all that blocked-halt time was spent on work the server was going to reject.

### What changed

The guest now sends the base generation it built on when **minting** the upload URL, and the server 409s there when that base has already been advanced past. The guest records the outcome as a conflict and skips the PUT entirely.

**Server**

- `VolumeHeads.fast_forward_viable?/3` mirrors `bump_head/5`'s acceptance rule exactly: base 0 viable only while the account has no HEAD, base N only while the HEAD is exactly at N.
- `Runners.fast_forward_viable?/2` delegates it, so the controller keeps talking only to the context.
- `volume_head_upload_url` pre-flights before minting and returns the same `stale base generation` 409 shape as the bump.

**Guest** (`dispatch-poll.sh`)

- Reads `base_generation` before the mint and sends it in the mint body.
- Captures the mint's HTTP status explicitly. It was `curl -fsS`, which would have collapsed a 409 into "no upload URL" and mislabeled genuine contention as `error`, the exact conflation the bump call already avoids.
- On 409: writes `conflict`, logs, returns without the PUT.

**Host** — no behavior change needed. `readUploadMillis` already returns -1 when the marker is absent, so a pre-empted promote counts as `rejected` with no upload observation.

### Why this shape

**No new surface.** The mint POST already existed, already carried the digest, and already had the SA-token + server-stamped account-label binding. A separate GET-current-HEAD endpoint would have added a route and a second round trip for the same answer, and the mint is exactly the moment the guest is about to commit to the transfer.

**Optimization, never a decision.** The pre-check is racy by construction: another host can win between the check and the PUT. So it may only ever *skip* doomed work, never grant a promote. Three things enforce that:

- `bump_head`'s compare-and-swap stays the sole authority on what becomes HEAD, and still runs on every upload that happens.
- `fast_forward_viable?` reads as viable for anything it cannot evaluate, so an odd input never suppresses a promote that would have been accepted.
- Every non-409 mint outcome, including transport failure and a 000 status, falls through to the existing upload-then-arbitrate path.

**A missing `base_generation` disables the pre-check** rather than defaulting to 0 the way the bump does. Runner images predating this change send no base, and defaulting would 409 their cold-job promotes for the entire duration of the fleet roll.

### Impact

Rejected promotes should mostly stop transferring anything, which returns the corresponding ~65s of blocked teardown per losing promote to the warm pool. Accepted promotes are unaffected. The pre-check adds no round trip: it rides the mint request that already happened.

The guest half only takes effect once a new runner image ships; until then those runners send no base and keep the old path.

### How to test locally

Server:

```
mix test test/tuist/runners/ test/tuist_web/controllers/runners_controller_test.exs test/tuist/runners_test.exs
```

530 passed. New coverage: `fast_forward_viable?` agreeing with `bump_head` at every base and its fail-safe fallbacks; the 409-before-mint path (the test `flunk`s if an upload URL is minted); the viable path; and the no-base-sent backward-compat path (`flunk`s if the pre-check runs at all).

Host agent:

```
go test ./internal/podagent/ -count=1
```

Full package passes. `TestFinalizeVolumePromoteAccounting` builds an `Entry` and drives the real `finalizeVolume` across three cases: pre-empted (rejected, no upload sample), lost after uploading (rejected, sample), and accepted (accepted, sample). Verified non-vacuous by folding `conflict` into the `default` arm, which fails two of the three subtests. `readUploadMillis` gets its own test including the unparseable case, since it had no coverage before this branch. `go vet` and `gofmt` clean.

`client_model` moves from indirect to direct in `go.mod` — the histogram assertion needs a sample count, which `testutil` cannot express.

Guest script:

```
shellcheck -S error infra/runner-image/dispatch-poll.sh
```

Clean, along with `bash -n`.

### Deploy signal

`tart_kubelet_cache_volume_upload_seconds_count` minus `tart_kubelet_cache_volume_promote_total{result="accepted"}` — the count of uploads that happened and did **not** become HEAD. It should trend toward zero as the image rolls.

Corrected after review: an earlier version of this section read `_count` against `result="rejected"`, which is wrong. `finalizeVolume` records the upload sample from the `volume-upload-ms` marker alone, outside the switch on the promote result, so accepted promotes observe samples too and that ratio would move with the accept rate as much as with the pre-flight.

The difference does not reach zero, and the metric comment says why: an upload followed by a failed bump lands in the `error` bucket, and a job excluded from the promote ratio (unclean halt, account mismatch) can leave a marker with no promote counted at all. Both are small and worth seeing.

### Context

Came out of watching the folded-CAS rollout (#11857 shipped the machinery, #12066 turned it on in staging/canary/production). The CAS itself is healthy: 87% warm materialization, `fill_percent` averaging ~24% with no sample above 50%, so image sizing (`masterCapGib: 28`, `casGib: 16`) is comfortable and is not the issue. The issue was purely the upload-then-arbitrate ordering.

### DIFF
--- a/infra/tart-kubelet/internal/podagent/metrics.go
+++ b/infra/tart-kubelet/internal/podagent/metrics.go
@@ -262,12 +262,19 @@ var cacheVolumeAdmissionDeclinedTotal = prometheus.NewCounter(
 // how long a promoting job holds the host slot before it can be reclaimed — the
 // signal for keeping the volume sized so uploads stay fast (the CAS is folded in,
 // so this now includes CAS bytes). Reported by the guest via the volume-upload-ms
-// status marker and recorded here at finalize.
+// status marker and recorded here at finalize, from the marker ALONE — outside
+// the switch on the promote result, so an accepted promote observes a sample
+// just as a rejected one does.
 //
-// A promote the server pre-empts at mint time (its base was already advanced
-// past) uploads nothing and so contributes NO sample, only a "rejected" promote.
-// The count of this histogram against promote_total{result="rejected"} is
-// therefore how much doomed transfer the pre-flight is still not avoiding.
+// That is what makes _count minus promote_total{result="accepted"} the reading
+// for the mint-time pre-flight: uploads that happened and did NOT become HEAD.
+// A promote the server pre-empts (its base was already advanced past) transfers
+// nothing, so it counts as "rejected" with no sample here, and the difference
+// trends toward zero. It does not reach zero — an upload followed by a failed
+// bump lands in the "error" bucket, and a job excluded from the promote ratio
+// (unclean halt, account mismatch) can leave a marker with no promote counted at
+// all. Reading _count against result="rejected" instead would move with the
+// accept rate rather than with the pre-flight.
 var cacheVolumeUploadSeconds = prometheus.NewHistogram(
 	prometheus.HistogramOpts{
 		Name:    "tart_kubelet_cache_volume_upload_seconds",
--- a/infra/tart-kubelet/internal/podagent/metrics.go
+++ b/infra/tart-kubelet/internal/podagent/metrics.go
@@ -263,6 +263,11 @@ var cacheVolumeAdmissionDeclinedTotal = prometheus.NewCounter(
 // signal for keeping the volume sized so uploads stay fast (the CAS is folded in,
 // so this now includes CAS bytes). Reported by the guest via the volume-upload-ms
 // status marker and recorded here at finalize.
+//
+// A promote the server pre-empts at mint time (its base was already advanced
+// past) uploads nothing and so contributes NO sample, only a "rejected" promote.
+// The count of this histogram against promote_total{result="rejected"} is
+// therefore how much doomed transfer the pre-flight is still not avoiding.
 var cacheVolumeUploadSeconds = prometheus.NewHistogram(
 	prometheus.HistogramOpts{
 		Name:    "tart_kubelet_cache_volume_upload_seconds",
--- a/infra/tart-kubelet/internal/podagent/volume_test.go
+++ b/infra/tart-kubelet/internal/podagent/volume_test.go
@@ -1127,6 +1127,47 @@ func TestReadPromoteResult(t *testing.T) {
 	}
 }
 
+// A promote the server pre-empted at mint time reports a conflict WITHOUT ever
+// uploading, so it must be counted as a rejection while contributing no sample to
+// the upload histogram — the whole point of the pre-flight is that a doomed
+// promote stops holding the slot open for a transfer. That distinction is exactly
+// what the deploy is measured by (upload count falling relative to rejected
+// promotes), so pin it: a conflict with no volume-upload-ms marker leaves the
+// histogram untouched, while an uploaded conflict still records one.
+func TestPreEmptedPromoteRecordsNoUpload(t *testing.T) {
+	statusDir := t.TempDir()
+	if err := os.WriteFile(filepath.Join(statusDir, promoteResultFile), []byte("conflict"), 0o644); err != nil {
+		t.Fatal(err)
+	}
+
+	if got := readPromoteResult(statusDir); got.Result != "conflict" {
+		t.Fatalf("pre-empted promote = %+v; want conflict (cross-host contention, not an error)", got)
+	}
+	if got := readUploadMillis(statusDir); got != -1 {
+		t.Fatalf("upload millis after a pre-empted promote = %d; want -1 (nothing was uploaded)", got)
+	}
+
+	// finalizeVolume's own branches over those two reads: counted as contention,
+	// but gated out of the upload histogram by the negative duration.
+	rejectedBefore := testutil.ToFloat64(cacheVolumePromoteTotal.WithLabelValues("rejected"))
+	RecordVolumePromote("rejected")
+	if readUploadMillis(statusDir) >= 0 {
+		t.Fatal("a pre-empted promote must not reach RecordVolumeUpload")
+	}
+	if got := testutil.ToFloat64(cacheVolumePromoteTotal.WithLabelValues("rejected")); got != rejectedBefore+1 {
+		t.Fatalf("rejected promotes = %v; want %v", got, rejectedBefore+1)
+	}
+
+	// The upload-then-arbitrate path (a promote that lost the race after paying
+	// for the transfer) still reports its duration.
+	if err := os.WriteFile(filepath.Join(statusDir, uploadMillisFile), []byte("4200"), 0o644); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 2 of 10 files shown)


infra/tart-kubelet/internal/podagent/metrics.go
```diff
@@ -263,6 +263,11 @@ var cacheVolumeAdmissionDeclinedTotal = prometheus.NewCounter(
 // signal for keeping the volume sized so uploads stay fast (the CAS is folded in,
 // so this now includes CAS bytes). Reported by the guest via the volume-upload-ms
 // status marker and recorded here at finalize.
+//
```
esnunes, line 268: One thing I wanted to flag about the deploy signal, since it's the thing that tells you whether this worked. The comment here (and the PR's "Deploy signal" section) reads `upload_seconds_count` against `promote_total{result="rejected"}`, but `finalizeVolume` records the upload sample at volume_reconcile.go:566 from the `volume-upload-ms` marker alone, outside the switch on `promote.Result` — so accepted promotes contribute samples too, and always will. That makes the ratio move with the accept rate as well as with the pre-flight.

(The new Go test's header comment repeats the same claim, so it'd be two places to touch.)

Would `upload_seconds_count − promote_total{result="accepted"}` be a tighter reading? That's the count of uploads that happened and did not become HEAD, and it should trend to zero as the image rolls. It still folds in an upload followed by a failed bump (the `error` bucket), which seems small and probably worth seeing anyway. Happy to be wrong if you're planning to read it as a ratio over time rather than an absolute.

infra/tart-kubelet/internal/podagent/volume_test.go
```diff
@@ -1127,6 +1127,47 @@ func TestReadPromoteResult(t *testing.T) {
...
 }
 
+// A promote the server pre-empted at mint time reports a conflict WITHOUT ever
+// uploading, so it must be counted as a rejection while contributing no sample to
+// the upload histogram — the whole point of the pre-flight is that a doomed
+// promote stops holding the slot open for a transfer. That distinction is exactly
+// what the deploy is measured by (upload count falling relative to rejected
+// promotes), so pin it: a conflict with no volume-upload-ms marker leaves the
+// histogram untouched, while an uploaded conflict still records one.
+func TestPreEmptedPromoteRecordsNoUpload(t *testing.T) {
```
esnunes: Small one on this test. The comment says it covers "finalizeVolume's own branches over those two reads", but `finalizeVolume` is never called — the test calls `RecordVolumePromote("rejected")` itself and then asserts that counter incremented, so that assertion is about the test's own call rather than the switch at volume_reconcile.go:544. The `readUploadMillis(statusDir) >= 0` check just above it also repeats the assertion made eight lines earlier.

Practical effect: the mapping you're pinning (`conflict` → `rejected`, and the `ms >= 0` gate keeping a pre-empted promote out of the histogram) would still pass if someone later folded `conflict` into the `default` arm. `TestFinalizeDiscardsWhenFastForwardRejected` already builds an Entry and drives a real `Finalize`, so extending that shape to assert both counters would close it.

The two `readUploadMillis` assertions are genuinely new coverage though — that function had none before this PR — so it's worth keeping either way.

esnunes (approved):
Really like the shape of this one. Reordering an existing round trip instead of adding an endpoint, and keeping the compare-and-swap as the only thing that decides a HEAD, is the version of this optimization that can't go wrong in an interesting way.

I went looking for the one failure mode that would matter — a pre-check that suppresses a promote the bump would have accepted — and I don't think it exists here. `fast_forward_viable?/3`'s two clauses mirror `bump_head/5`'s two clauses exactly, and the HEAD is monotonic: `bump_head` only increments, and I couldn't find any `delete_all` or `Repo.delete` against `VolumeHead` in server/lib. So a base that reads stale can't become viable again before the bump, and the check can only be wrong in the harmless direction. The rollout also looks safe both ways round — old runner sends no base and keeps the old path, old server ignores the field — which matters given the guest half only lands with a new image.

Both comments are non-blocking. The one I'd most want your read on is the deploy signal: the upload histogram counts accepted promotes too, so the ratio it's compared against moves with the accept rate as well as with the pre-flight — which matters because that reading is how you'll decide whether this landed. The other is just a test that names `finalizeVolume` without calling it.

infra/tart-kubelet/internal/podagent/metrics.go
fortmarek: Corrected — you are right, and the `accepted` case is now demonstrated by a test: `RecordVolumeUpload` fires from the marker alone at volume_reconcile.go:566, outside the switch, so accepts observe samples too. Switched both the metric comment and the PR section to `_count − promote_total{result="accepted"}`, and noted the two residues it keeps (upload-then-failed-bump in the `error` bucket, and jobs excluded from the promote ratio by an unclean halt or account mismatch) so nobody later reads a non-zero floor as a regression.

infra/tart-kubelet/internal/podagent/volume_test.go
fortmarek: Rewritten — the test now builds an `Entry` and drives the real `finalizeVolume` rather than asserting its own `RecordVolumePromote` call. Three cases: pre-empted (rejected, no sample), lost after uploading (rejected, sample), accepted (accepted, sample). Confirmed non-vacuous by folding `conflict` into the `default` arm, which fails two of the three subtests — the exact regression you named. Kept the `readUploadMillis` coverage as its own test and added the unparseable case. Needed `client_model` as a direct dep for the histogram sample count, since `testutil` cannot express it.

Issue #12040, opened 2026-07-23
SwiftPM systemLibrary target generates an invalid macOS framework with a top-level Modules directory


RbBtSn0w:
### What happened?

Tuist materializes GRDB's SwiftPM `GRDBSQLite` `systemLibrary` target as a
macOS framework. The generated target's `Copy Module Map` build script creates
a real directory at:

```text
GRDBSQLite.framework/Modules/module.modulemap
```

This is not a valid layout for a versioned macOS framework. Its root should
contain `Versions` plus symlinks to versioned content. As a result, strict code
signature verification consistently fails:

```text
GRDBSQLite.framework: unsealed contents present in the root directory of an embedded framework
```

This also affects real projects using Xcode mergeable libraries. With
`MERGED_BINARY_TYPE=automatic` in Debug, Xcode may copy the generated framework
into `App.app/Contents/ReexportedBinaries`. If the top-level `Modules` directory
is retained, app signing fails with the same error.

Expected behavior: Tuist should either preserve the SwiftPM system-library
semantics without materializing it as a framework, or generate a valid macOS
framework layout, for example with the module map under
`Versions/A/Modules` and a root `Modules` symlink.

### How can we reproduce it?

A minimal reproduction is available here:

https://github.com/RbBtSn0w/tuist-mergeable-system-library-repro

```sh
git clone https://github.com/RbBtSn0w/tuist-mergeable-system-library-repro.git
cd tuist-mergeable-system-library-repro

tuist install
tuist generate --no-open

xcodebuild build \
  -workspace MergeableSystemLibraryRepro.xcworkspace \
  -scheme ReproApp \
  -configuration Debug \
  -destination 'platform=macOS,arch=arm64' \
  -derivedDataPath /tmp/tuist-mergeable-system-library-repro \
  CODE_SIGN_STYLE=Manual \
  CODE_SIGN_IDENTITY=- \
  DEVELOPMENT_TEAM=

codesign --verify --deep --strict --verbose=4 \
  /tmp/tuist-mergeable-system-library-repro/Build/Products/Debug/GRDBSQLite.framework
```

The final command fails consistently.

The generated `GRDBSQLite` target contains a `Copy Module Map` script
equivalent to:

```sh
mkdir -p "$TARGET_BUILD_DIR/$WRAPPER_NAME/Modules"
cp /path/to/GRDB.swift/Sources/GRDBSQLite/module.modulemap \
[...]

PR #12128, opened 2026-07-29
fix(cli): resolve module map paths relative to generated

Loupehope:
### Summary

Fixes a regression introduced by [36e5493](https://github.com/tuist/tuist/pull/12055/changes/36e5493ec2657732123bd14af7e2988f98ec4207) from [@pepicrft](https://github.com/pepicrft) in #12055.

### Root cause

The change preserved the original MODULEMAP_FILE value in MODULEMAP_PATH to avoid machine-specific absolute paths in cache hashes.

However, MODULEMAP_FILE is produced relative to the Swift package source root, while MODULEMAP_PATH is consumed by ExtractAPI from the generated Xcode project. For remote SwiftPM dependencies those directories differ: sources live in checkouts, while generated projects live in tuist-derived/Projects.

Therefore, copying mappedSettingsDictionary[Self.modulemapFileSetting] directly is incorrect. It makes ExtractAPI resolve the module map in the wrong context.

MODULEMAP_FILE also cannot remain in the generated framework target: Xcode would copy the module map into the framework product, reintroducing duplicate module-map discovery and the Copy Module Map dependency-cycle issue addressed by #11588.

### Solution

Tuist now resolves the source module-map setting first, then emits the same map relative to the generated .xcodeproj through MODULEMAP_PATH.

The generated value uses $(PROJECT_DIR), so it resolves correctly for ExtractAPI while remaining independent of the machine-specific checkout path used in cache hashes.

### Validation

Added focused unit coverage for framework and static-framework targets, remote SwiftPM layouts, supported source-root macros, and deterministic output across different absolute checkout roots.

Extended generated_ios_app_with_modulemap_packages with a local path C package that declares a custom module map. The App imports and calls its API, ensuring the package is compiled.

Before merging, it would be useful to run the relevant acceptance tests to further validate real build and DocC behavior for the module-map package fixtures, the SDK fixture, and the static-framework scenario from #11588.


Loupehope:
I also tested the issue https://github.com/tuist/tuist/issues/12040 using the tuist artifact from this pull request

Loupehope:
@pepicrft, when you have a moment, would you mind taking a look at the changes in this PR? I’d really appreciate your feedback :)

pepicrft:
Thanks a lot, @Loupehope. Let's continue the conversation [here](https://github.com/tuist/tuist/pull/12145) where I'm reconciling your changes with some fix that had already landed in `main`.

PR #12129, opened 2026-07-29
fix(kura,cli): absorb module cache bursts without memory flapping

pepicrft:
This makes Kura absorb large module-cache bursts without treating clean artifact page-cache churn as non-reclaimable memory pressure. It also lets command-line clients retry temporary service-unavailable download responses.

## What changed

- Scale the response-stream budget from the configured memory gaps instead of stopping at a fixed 64-mebibyte ceiling.
- Separate non-reclaimable admission pressure from the signal that asks read and write paths to release clean file-backed cache.
- Export the new pressure signal and add it to the Kura Grafana dashboard.
- Retry cache downloads after temporary service-unavailable responses, honor numeric `Retry-After` values, and keep compatibility with older servers that omit the header.
- Restrict cache retries to downloads so uploads are not replayed or buffered.

## Why

A customer generated a concentrated module-cache download burst that exhausted Kura's response admission pool while the container still had memory that could be reclaimed from file-backed cache. The command-line client surfaced the temporary failures directly, amplifying the impact across parallel module downloads.

## Root cause

The response-stream pool had a fixed 64-mebibyte ceiling regardless of the container's available headroom. At the same time, Kura drove its admission state from a conventional working-set estimate. During the incident, the kernel reclassified more than one gibibyte of clean artifact pages between active and inactive cache in roughly 200 milliseconds, which made the state alternate even though the underlying container charge barely changed.

The cache client also did not install the shared retry middleware, so a temporary status `503` became a failed module download.

## Approach

Kura now derives its response-stream ceiling from both the soft-to-hard watermark gap and the hard-watermark-to-runtime-limit gap. Admission pressure is based on anonymous, kernel, shared, dirty, and writeback memory, while clean file-backed cache is excluded. A separate reclaim signal still disables mapped-file serving and releases completed file ranges when the total charge or conventional working set is high.

The client retries download requests only. Newer servers can guide the delay with `Retry-After`; older servers continue to work through the existing exponential delay.

## Impact

Large download bursts can borrow memory from clean file cache while remaining inside the existing transient-memory and runtime reserves. Clean page-cache recency changes no longer constrain healthy requests, but cache reclamation still reacts within the 200-millisecond sampling interval. No server protocol or persisted data format changes, so mixed-version rollouts remain compatible.

### How to test locally

- `cd kura && mise run format -- --check`
- `cd kura && mise exec -- bazel test //:kura_lib_test --test_arg=pressure --test_output=errors`
- `cd kura && mise exec -- bazel test //:kura_lib_test --test_arg=response_streaming_pool_scales_with_memory_headroom --test_output=errors`
- `cd kura && mise exec -- bazel test //:kura_lib_test --test_arg=adaptive_policy_drops_cache_when_reclaim_is_requested --test_output=errors`
- `mise x -- swiftformat cli/Sources/TuistHTTP/RetryMiddleware.swift cli/Tests/TuistCacheTests/Client/CacheClientTests.swift --lint`
- `xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing:TuistHTTPTests/RetryMiddlewareTests -only-testing:TuistCacheTests/CacheClientTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""`
- `jq empty kura/ops/grafana/dashboards/kura-cluster.json`


kura/src/memory/cgroup.rs
```diff
@@ -85,6 +133,36 @@ fn working_set_bytes(current_bytes: u64, inactive_file_bytes: Option<u64>) -> u6
...
+        current_bytes,
+        anon_bytes,
+        file_bytes,
+        kernel_bytes,
+        shmem_bytes,
+        file_dirty_bytes,
+        file_writeback_bytes,
+        inactive_file_bytes,
+    } = components;
+    if let (Some(anon_bytes), Some(kernel_bytes)) = (anon_bytes, kernel_bytes) {
```
esnunes, line 147: One thing I wanted to check on the new pressure sum: on the cgroup v2 primary path it's `anon + kernel + shmem + file_dirty + file_writeback`, and as far as I can tell from the cgroup v2 docs, `sock` (network buffer memory) is a separate `memory.stat` row that `kernel` doesn't fold in — nothing in `kura/src/memory/` reads it. The old `current − inactive_file` signal included socket buffers, so they effectively stop counting toward admission pressure with this change. On a node streaming to many slow clients that could delay the Constrained/Critical transitions while the container is closer to its limit than the signal says. Interestingly the fallback paths (`current − file + …` and the working-set fallback) still include sock implicitly, so older kernels would be more conservative than newer ones. Was leaving `sock` out deliberate? If not, adding `named_value(&stat, "sock")` to the primary sum looks cheap. Adding it shouldn't double count anything as far as I can tell — sock is charged separately from `anon`, `kernel`, and the file rows in `memory.stat`, and the sum is already capped with `.min(current_bytes)` either way.

kura/src/memory/mod.rs
```diff
@@ -136,7 +138,12 @@ impl MemoryController {
         self.inner
             .container_accounting_selected
             .store(true, Ordering::Release);
-        self.observe(sample.working_set_bytes)
+        self.inner.reclaim_file_cache.store(
```
esnunes, line 141: A question about the reclaim condition rather than a bug: `current_bytes >= hard_limit_bytes` (85% of the container limit by default) looks like it becomes quasi-permanently true on a warm serving node — the kernel keeps clean page cache charged until it's forced to reclaim, so once a node's cumulative footprint crosses the hard watermark, `memory.current` doesn't come back down on its own (drop-behind only uncharges ranges Kura itself touches; cold cache stays charged). If that holds, the reclaim signal — mmap-serving denial plus always-on drop-behind — is effectively steady-state on busy nodes rather than reactive. That may well be the intent (cache nodes trading warmth for request capacity), in which case it might be worth saying so explicitly in the docs; if not, the condition might want to subtract reclaimable file cache, or gain the same ~10% recovery hysteresis the pressure state machine has — as written the flag is a raw >= recomputed every 200 ms, and its working-set arm is exactly the quantity this PR documents swinging >1 GiB per sample, so near the soft watermark mmap serving would flip per sample. `kura_container_memory_current_bytes` on the prod dashboard should confirm or refute the steady-state part quickly.

cli/Sources/TuistCache/Client/Client+Cache.swift
```diff
@@ -23,6 +23,7 @@ extension Client {
             serverURL: cacheURL,
             transport: TuistURLSessionTransport(session: session),
             middlewares: HARRecordingMiddlewareFactory.middlewares() + [
+                RetryMiddleware(retryableRequestMethods: ["GET"]),
```
esnunes, line 26: Heads-up on the GET allowlist's reach: `Client.authenticated` is shared by all cache services, so the retry also wraps `LoadCacheCASService` — the CAS hot path that fires thousands of GETs per build through the `.tuistCAS` session. That session's 15 s request timeout is documented as a deliberate fail-fast so a hung backend surfaces in seconds and the circuit breaker skips it for the rest of the build (`URLSession+Tuist.swift:109-117`). `RetryMiddleware` retries thrown transport errors too — including timeouts — so a hung backend on a CAS download now costs up to 4 × 15 s plus backoff before the failure reaches the breaker, per request until it trips. If the goal is retrying Kura's transient 503s, one option is to retry only on retryable status codes (not thrown errors) for this client, or let the CAS services opt out of the middleware. Might deserve a test either way.

kura/src/metrics.rs
```diff
@@ -803,6 +805,11 @@ impl Metrics {
             "Total memory currently charged to the container control group",
             container_memory_current_bytes.clone(),
         );
+        registry.register(
+            "kura_container_memory_pressure_bytes",
+            "Container memory charge excluding clean file-backed cache, used to drive Kura pressure state",
```
esnunes, line 810: Nit, entirely optional: the new gauge's HELP text ends with "used to drive Kura pressure state" — per the convention from the #11834 metrics review, registration strings state what the metric measures and the role/interpretation lives in runbooks and dashboards. "Container memory charge excluding clean file-backed cache" alone would match the neighbouring gauges.

esnunes (approved):
Really nice diagnosis and a clean split of concerns — separating "what constrains admission" from "what triggers cache reclaim" is exactly the right shape for the flapping incident, and the docs/README updates that came with it are thorough. A few things I verified while reviewing: every response-stream permit still double-charges the shared transient budget (memory/mod.rs:574, :602), so removing the 64 MiB cap can't overcommit the container; the new pool arithmetic reproduces the test vectors by hand (4 GiB node → 512 MiB, 256 MiB profile → 19 MiB); and since the pressure-gap arm is headroom/2, downloads can never take more than half the transient budget, so uploads keep the rest. The rollout also looks mixed-version safe — additive metric, additive dashboard query, and the CLI works against servers that don't send Retry-After.

The inline notes carry my main questions: whether leaving `sock` out of the new pressure sum was deliberate, whether the reclaim signal being quasi-steady-state on warm nodes (and unhysteresed near the soft watermark) is the intended trade, and whether the CAS fail-fast path should really inherit retry-on-timeout. Plus a small nit on the new metric's HELP text.

PR #12149, opened 2026-07-30
perf(server): read test-case state from the current-states aggregate

pepicrft:
## Why

Second part of the fix for the recurring ClickHouse slow-query / OOM alerts on the test-analytics path. Part 1 added `test_case_current_states` (pre-aggregated control-plane state) but nothing read it yet. This PR flips the readers over so we actually stop folding the `test_case_states` ledger at query time.

## What this PR does

Points every control-plane state reader in `Tuist.Tests` at `test_case_current_states` via `argMaxIfMerge` instead of `argMaxIf ... GROUP BY` over the raw ledger:

- `resolve_test_case_state/2` (single-key lookup)
- `test_case_states_subquery/1` (feeds both the preloaded resolution and the joined listing)
- `test_case_flaky_flags_chunk_query/2` (ingestion flaky-flag lookup)
- `quarantined_test_case_states_subquery/2`
- the `state_filter_mode/2` probe

## Notes and tradeoffs

- The quarantine subquery is resolved in two levels on purpose. Selecting `argMaxIfMerge(state) AS state` and filtering on it in the same level collides: the alias shadows the source aggregate column, so a `HAVING argMaxIfMerge(state)` re-applies the merge to a finalized `String` and ClickHouse rejects it. The inner query finalizes, the outer filters the plain column. A null state (flaky-only test case) is correctly excluded because `NULL IN (...)` is 0.
- The `FINAL` over `test_cases` on the joined path is intentionally left as-is, and `@test_case_state_join_settings` stays conservative. That `FINAL` cannot spill and remains the dominant cost on that path, so any relaxation is deferred to part 3 and gated on measuring the real per-project peak memory in production.
- Null and default handling is preserved: never-touched, flaky-only, and unmuted-back-to-enabled all resolve identically to the old ledger reads.

## Validation

- Full `tests_test.exs` suite green (256 passed), including `list_test_cases`, the quarantine listing, and the flaky-flag paths.
- Web and API tests through these paths green (`quarantined_tests_live`, `test_case_live`, `test_cases_controller`, 46 passed).
- `mix format` and `mix credo` clean.
- Independently reviewed for rollout safety by two separate agents; the one real risk they surfaced (an over-aggressive join-settings relaxation) was reverted and moved to part 3.

## Rollout

Stacked change, part 2 of 3. Do not merge until part 1 has rolled out and its backfill has completed, so no pod reads the table before it is populated.

### DIFF
--- a/server/lib/tuist/tests.ex
+++ b/server/lib/tuist/tests.ex
@@ -38,6 +38,7 @@ defmodule Tuist.Tests do
   alias Tuist.Tests.Test
   alias Tuist.Tests.TestCase
   alias Tuist.Tests.TestCaseBranchPresence
+  alias Tuist.Tests.TestCaseCurrentState
   alias Tuist.Tests.TestCaseEvent
   alias Tuist.Tests.TestCaseFailure
   alias Tuist.Tests.TestCaseRun
@@ -50,7 +51,6 @@ defmodule Tuist.Tests do
   alias Tuist.Tests.TestCaseRunDashboardCount
   alias Tuist.Tests.TestCaseRunFlakyCorrection
   alias Tuist.Tests.TestCaseRunRepetition
-  alias Tuist.Tests.TestCaseState
   alias Tuist.Tests.TestModuleRun
   alias Tuist.Tests.TestRunDestination
   alias Tuist.Tests.TestRunError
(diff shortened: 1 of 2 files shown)


pepicrft:
Stacked PRs, merge in order:
1. #12148 write path (table + MV + backfill)
2. #12149 read flip (depends on #12148)
3. #12150 join-settings note (depends on #12149)

Before merging #12148, run the aggregate-type probe against production ClickHouse and confirm alert ffeb6l2ax5qtcf fires on this path via system.query_log. Do not merge #12149 until #12148 has rolled out and backfilled.

PR #12152, opened 2026-07-30
feat(kura): cascade action-cache entries on blob eviction

pepicrft:
> Make blob eviction in Kura atomic with the action-cache entries that reference it, so an entry can never outlive its blobs. This closes the strand class that was surfacing as `CAS error: missing object` and failing Xcode builds with the cache enabled on CI.

## Why

An action-cache entry (a Bazel [Remote Execution API](https://github.com/bazelbuild/remote-apis) `ActionResult`) and its output blobs in the [content-addressable storage](https://en.wikipedia.org/wiki/Content-addressable_storage) form one integrity unit: the entry is meaningless once any blob it references leaves the store. The publish path already enforces this (`FindMissingBlobs` then `BatchUpdateBlobs` then `UpdateActionResult` last), but blob eviction is capacity-driven and independent, so `evict_segment` dropped a blob's manifest without touching the entries referencing it. An entry could then outlive its blobs and be served as a hit whose blob fetch fails.

Two serve-side gates normally mask this (per-key `GetActionResult` runs the presence check and returns `NOT_FOUND`, which clients recompute from, and the snapshot serve gates orphan entries). The production incident was those gates not being enough: under sustained memory pressure the snapshot presence gate (background work) was denied for hours and served a snapshot advertising blobs the store had already evicted. clang-compiled Xcode targets fail hard on a missing object mid-build, where Swift recompiles and tolerates it, so this showed up as red CI.

## Root cause

Eviction did not treat `(action-cache entry -> blob)` as an atomic unit. It removed the blob and left the referencing entries behind, relying entirely on lazy serve-time repair to hide the resulting dangling references. When that repair could not run (memory pressure) or had not run yet (the reconcile window), the strand was served.

## Approach

Make eviction honor the same invariant the publish path enforces, at the source, rather than leaning harder on the gates.

- A node-local, derived reverse index: a new `action_cache_blob_refs` RocksDB column family of individual `{blob}\0{entry}` pair rows (empty value), mirroring the existing `segment_artifacts` / `project_artifacts` index shape. Pair rows rather than a set-valued row so two entries referencing one blob write disjoint keys and never lose an update (RocksDB write batches are atomic but have no compare-and-swap).
- The rows are maintained in the same atomic batch as every entry write, re-publish, and delete (`persist_inline_artifact_with_version`, `delete_artifact_metadata`, `delete_namespace_with_version`), rebuilt once at startup from the entries on disk, and consulted by `evict_segment` to cascade-delete referencing entries in the same batch that drops the blob. Each cascade candidate is re-validated against the entry's current `ActionResult`, so a reverse row left stale by a re-publish only deletes itself and never removes an unrelated entry.
- The map is purely derived and never replicated; every node builds its own. The cascade stays inert until the startup backfill completes and is gated by `KURA_ACTION_CACHE_EVICTION_CASCADE_ENABLED`. The serve-side presence gates stay on permanently as the backstop.

Deliberately not done: a publish-side "refuse the entry if its blobs are absent" check. Kura's metadata lane intentionally replicates fresh action-cache entries ahead of the bulk-blob backlog, so a receiving node legitimately sees an entry before its blobs; a local-presence rule would reject valid replicated entries. The narrow publish-then-evict race is left to the serve-side `NOT_FOUND`, which is already an expected, safe client outcome. This direction was cross-checked with an independent design review before implementing.

Tree outputs (`output_directories` leaves) are out of scope for the cascade in this phase (Tuist's action-cache is `output_files` only in practice); the gates remain the backstop for that shape.

## Impact

- Operators: new env var `KURA_ACTION_CACHE_EVICTION_CASCADE_ENABLED` (default `true`), wired through the Helm chart. The cascade also requires the one-time reverse-map backfill to complete before it activates, so it is safe on a fresh start; the flag is a kill switch.
- Rollout: adding a column family is a forward-only on-disk migration (same accepted boundary as the `action_cache_index` column family). A rollback to a binary that predates this column family is not supported once the store has opened with it.
- Observability: new metric `kura_action_cache_cascade_removed_total`. Watch it rise as the cascade takes load, and the serve-side presence-gate hit rate trend toward zero.
- No wire-format or client/plugin contract changes.

## Validation

Run from `kura/`:

- `cargo test --lib store::tests::` — 100 passed (new coverage: write-path rows, re-publish moves rows, delete and namespace-delete drop rows, backfill reconstructs and arms the cascade, eviction cascades a referenced blob, readiness gate leaves entries when inactive, stale reverse row does not take out a live entry).
- `cargo test --lib action_cache_refs::` — 4 passed.
- `cargo test --lib reapi::` — 71 passed; `cargo test --lib replication::` — 47 passed.
- `cargo clippy --all-targets -- -D warnings` — clean.
- `cargo fmt` — clean.
- `helm template ops/helm/kura` — renders `KURA_ACTION_CACHE_EVICTION_CASCADE_ENABLED: "true"`.

### How to test locally

1. Enable the cascade (default on): `KURA_ACTION_CACHE_EVICTION_CASCADE_ENABLED=true`.
2. Publish an action-cache entry and its output blobs, then force eviction of the blob's segment (e.g. via a small `KURA_CAS_CAPACITY_BYTES` and continued writes, or the `evict_segment` path exercised by the new `evicting_a_referenced_blob_cascades_the_action_cache_entry` test).
3. Confirm the entry is removed from the store together with the blob rather than being served as a hit that then 404s the blob fetch.


kura/src/store.rs
```diff
@@ -567,6 +579,14 @@ impl Store {
                     &rocksdb_write_buffer_manager,
                 ),
             ),
+            ColumnFamilyDescriptor::new(
+                ROCKSDB_CF_ACTION_CACHE_BLOB_REFS,
```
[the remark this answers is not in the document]
pepicrft: Good catch, fixed in 9cfeaf04da.

Confirmed the problem: `Store::open` lists a fixed CF set and calls `DB::open_cf_descriptors` with no tolerant open, so a binary that predates `action_cache_blob_refs` rejects a PVC that already contains it and crash-loops on rollback. architecture.md explicitly commits to keeping rollbacks correct, so this needed to go.

Rather than a dedicated column family, the reverse index now lives in the existing `key_value` CF under a `blob_ref/` key prefix (see `action_cache_blob_ref_key`/`action_cache_blob_ref_prefix` in `utils.rs`). Same `{blob}\0{entry}` pair-row semantics, but an older binary just never point-reads the prefixed keys, so it reopens the database fine. I verified `key_value` is never iterated or full-scanned anywhere, and the backfill marker already lives there, so the prefixed rows are inert to every other reader. The prefix can't collide with an inline artifact row either, since those are keyed by a 64-char hex artifact id with no `/`.

kura/src/store.rs
```diff
@@ -2190,6 +2278,108 @@ impl Store {
...
+            {
+                continue;
+            }
+            let Some(entry_bytes) = self.inline_bytes(&entry_id)? else {
+                continue;
+            };
+            let still_references = self
+                .action_cache_entry_blob_ids(&entry_manifest.namespace_id, &entry_bytes)
+                .iter()
+                .any(|id| id == blob_artifact_id);
```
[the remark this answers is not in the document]
pepicrft: The race is real, but I'm intentionally not serializing the validation with `artifact_write_lock_for`, and I've documented why on `stage_action_cache_cascade_for_blob` in 9cfeaf04da.

The blocker is a deadlock: `evict_segment` runs from segment rotation inside `persist_artifact_from_path_with_version`, which is still holding an `artifact_write_lock` stripe when it calls `evict_segments`. The write locks are striped by id, so a blob id and an entry id can hash to the same stripe, and taking the entry lock in the cascade would then self-deadlock on the stripe the caller already holds.

The residual impact is bounded and self-healing rather than corrupting: if a re-publish commits a new version between this read and the batch commit, the fresh entry is removed on a stale validation, but it degrades to a `NOT_FOUND` the client recomputes and republishes, the snapshot reconcile skips any index row whose manifest is gone, and the orphaned reverse row is reclaimed when its blob is later evicted. This is the same read-then-delete-without-writer-lock race that `expire_stale_action_cache_entries` -> `delete_artifact_metadata` already accepts, so closing it properly would need a broader lock-ordering redesign that's out of scope here. Happy to file a follow-up if you'd rather track it.

kura/src/store.rs
```diff
@@ -2190,6 +2270,122 @@ impl Store {
...
+    /// itself and never takes out a live, unrelated entry.
+    ///
+    /// The validation is deliberately not serialized with `artifact_write_lock_for`
+    /// against a concurrent re-publish of the same key. Eviction runs from
+    /// segment rotation *inside* `persist_artifact_from_path_with_version` while
+    /// that path already holds an `artifact_write_lock` stripe, and the striped
+    /// locks can hash-collide a blob id with an entry id, so taking the entry
+    /// lock here could self-deadlock. The residual race (a re-publish commits a
+    /// new version between this read and the batch commit, so the fresh entry is
+    /// removed on a stale validation) is bounded and self-healing: the entry
```
esnunes, line 2296: One thing I wanted to flag from tracing the concurrency here: `stage_action_cache_cascade_for_blob` reads, validates, and stages an entry's deletion without taking `artifact_write_lock_for(entry_id)`, which every entry write serializes on. If a re-publish commits between the cascade's validation read and the eviction batch write, the batch's key-based deletes remove the just-committed newer version.

I traced the fallout and every branch looks fail-safe and self-healing — the fresh entry is lost (a miss; the client recomputes), the newer index row dangles but `action_cache_manifests_bounded`'s stale-row cleanup drops it, and the newer reverse rows are treated as stale by the cascade's own re-validation on a later eviction — so I don't think this needs a lock. But since architecture.md documents the publish-after-scan race explicitly, it might be worth a sentence there (or in this function's doc comment) acknowledging this one too, so a future change doesn't accidentally start relying on the cascade never removing a live newer version.

kura/src/store.rs
```diff
@@ -1064,6 +1078,13 @@ impl Store {
...
             self.bump_action_cache_generation(&manifest.namespace_id);
+            // No blob-refs (reverse index) maintenance here: an action-cache
+            // entry is always persisted inline. `update_action_result` writes it
+            // through the inline path, and a replicated inline entry re-applies
+            // through the same inline path, so its reverse rows are staged there.
+            // Should an entry ever land segment-backed, the startup backfill
+            // reconciles it from the manifests it scans, and the serve-side
+            // presence gates cover it until then.
         }
         if let Some(previous_manifest) = &existing
```
esnunes, line 1093: One clarification after digging here: I first read this comment as "entries are always inline", but that only holds for the current write paths (`update_action_result` caps entries at `MAX_INLINE_REPLICATION_BODY_BYTES` and persists inline). Segment-backed entries remain a supported shape — this very path maintains their `action_cache_index` rows, the serve paths read them via `read_artifact_bytes`, and bootstrap dispatches on the peer manifest's `inline` flag, so a segment-backed entry anywhere in the mesh propagates segment-backed. For that shape the comment's claim doesn't hold: `backfill_action_cache_blob_refs` skips any manifest without inline bytes (`let Some(bytes) = self.inline_bytes(…) else { continue; }`), and the cascade's validation does the same — so segment-backed entries get no reverse rows from any path and are never cascaded; the serve-side gates are their only cover. Worth either tightening the comment to say exactly that, or reading entry bytes through the segment-tolerant reader in the backfill and cascade validation so the shape is actually covered.

kura/src/store.rs
```diff
@@ -2190,6 +2270,122 @@ impl Store {
...
+            IteratorMode::From(prefix.as_bytes(), rocksdb::Direction::Forward),
+        );
+        for item in iter {
+            let (ref_key, _) =
+                item.map_err(|error| format!("failed to iterate blob refs: {error}"))?;
+            if !ref_key.starts_with(prefix.as_bytes()) {
+                break;
+            }
+            let entry_id = std::str::from_utf8(&ref_key[prefix.len()..])
+                .map_err(|error| format!("invalid blob-ref key: {error}"))?
```
esnunes, line 2321: Small test-coverage suggestion: the suite doesn't cover an entry referencing two blobs evicted in the same segment — the common one-action-several-outputs shape. That path exercises the `cascaded_entries` dedupe (second blob's scan skips re-validation), the idempotent double-staged row deletes, and the metric counting the entry once rather than per (blob, entry) pair. I'd expect it to pass as written; a test would just lock those in. Definitely not a blocker.

esnunes (approved):
Really solid change — the design gets the hard parts right: reverse rows committed in the same atomic batch as every entry write and delete, the cascade re-validating each candidate against the entry's current ActionResult so a stale row only deletes itself, the readiness gate keeping an incomplete map from ever driving deletes, and the serve-side gates staying on as the backstop. Factoring the reference derivation into one module shared by the write, delete, backfill, and eviction paths is exactly what keeps a derived index from drifting. I checked all the pre-existing entry-deletion paths (per-key gate, snapshot reconcile, TTL expiry, namespace delete) and each one now maintains the reverse rows. I also re-ran the store and action_cache_refs suites locally at the PR head: 100 + 4 passed.

The inline notes are all low-severity and none block merging: an undocumented (but fail-safe and self-healing, as far as I could trace) race between the cascade and a concurrent re-publish; segment-backed action-cache entries being invisible to both the backfill and the cascade (the comment there currently overstates what the backfill covers); and a missing test for the one-entry-two-evicted-blobs dedupe path.

Two suggestions beyond the notes: it would be good to cover the full Bazel ActionResult surface in this PR — stdout/stderr digests and the tree digest itself are already indexed, but Tree leaves (the file digests inside `output_directories`) aren't walked, and while Tuist's cache is `output_files`-only in practice, Bazel workloads do use `output_directories`, so covering leaves would close the gap for them. And I'd suggest a follow-up to drop `KURA_ACTION_CACHE_EVICTION_CASCADE_ENABLED` and make the cascade unconditional once it proves itself in production (`kura_action_cache_cascade_removed_total` rising, presence-gate hits trending to zero), plus refactoring existing paths that re-derive entry→blob references on the fly to lean on this index where it fits.

kura/src/store.rs
pepicrft: Agreed, done in aa9d834ec4. The doc comment on `stage_action_cache_cascade_for_blob` already spelled this out, and I've now added a sentence to architecture.md right next to the publish-versus-eviction race, including the explicit invariant that the cascade must never be assumed to leave a live newer version in place.

And thanks for tracing every branch, that matches my read exactly: the fresh entry becomes a `NOT_FOUND` the client recomputes and republishes, the dangling index row is dropped by `action_cache_manifests_bounded`'s stale-row cleanup, and the newer reverse rows are re-validated as stale on the next eviction. No lock needed, and it's the same read-then-delete-without-writer-lock shape `expire_stale_action_cache_entries` already accepts.

kura/src/store.rs
pepicrft: Good catch, and you're right it was wrong rather than just imprecise. Fixed in 26811fe22d.

The comment claimed the startup backfill reconciles a segment-backed entry, but `backfill_action_cache_blob_refs` keys on `inline_bytes` and skips a manifest without them, exactly like the write and delete paths. I rewrote the comment to say what actually happens: segment-backed action-cache entries are a structurally supported but not-currently-produced shape (`update_action_result` rejects a body over `MAX_INLINE_REPLICATION_BODY_BYTES`), the reverse index deliberately does not cover them, and the serve-side presence gates are their sole cover. Added the same note at the backfill skip.

I went with tightening the comment rather than reading through the segment-tolerant reader in the backfill and cascade validation: the shape does not arise from any current write path, covering it would mean async segment reads in the sync backfill/cascade paths, and the gates already cover it safely, so it is complexity for a shape that does not occur. Happy to switch to actually covering it if you'd rather, but it seemed like carrying weight for nothing today.

kura/src/store.rs
pepicrft: Added in 26811fe22d: `cascade_removes_an_entry_referencing_two_blobs_in_one_segment_once`. It persists two blobs that co-locate in one segment (asserted), references both from a single entry, evicts the segment, and checks the entry is removed once, both reverse rows are cleared, and the cascade counter reads 1 rather than 2. Passed as written, so it just locks in the `cascaded_entries` dedupe, the idempotent double-staged deletes, and the count-once metric.

PR #12184, opened 2026-07-31
fix(cli): decode devicectl "sameMachine" transport type and skip simulators when listing physical devices

### LINKED ISSUE #12052: `tuist run <preview>`: "Fetching the list of devices failed" — devicectl `transportType: "sameMachine"` (simulators) fails to decode
### What happened?

`tuist run <preview-url>` (and the macOS menu-bar app) **always** fail with *"Fetching the list of devices failed."* It fails before any device can be selected — even for simulator targets — so previews can't be launched at all.

**Root cause:** `DeviceController.findAvailableDevices()` runs `xcrun devicectl list devices --json-output <path>` and decodes it with a single strict `JSONDecoder().decode(DeviceList.self, …)`, mapping **any** decode error to `DeviceControllerError.fetchingDevicesFailed`.

On current Xcode/CoreDevice, `devicectl list devices` includes **simulators**, and every simulator entry has `"transportType": "sameMachine"`. But `DeviceList.Result.Device.ConnectionProperties.TransportType` only defines `localNetwork` and `wired`. `"sameMachine"` is a present-but-unknown enum value, so decoding `TransportType?` throws (Swift doesn't decode an unknown raw value as `nil`), and because the whole list is decoded in one shot the entire fetch fails.

This is **not** covered by #8001/#8010 — that fix handles devices with *missing* properties (made `name`/`udid`/`platform` optional + `compactMap`). The `compactMap` runs *after* decode, so it can't help an enum that throws *during* decode.

### How do we reproduce it?

```sh
# 1. Confirm devicectl returns simulators with the offending value:
xcrun devicectl list devices --json-output - | python3 -m json.tool | grep transportType
# -> "transportType": "sameMachine"   (for every "reality": "simulated" entry)

# 2. Run any preview:
tuist run <any-preview-url>
# -> "Fetching the list of devices failed."
```

Minimal offending device entry (sanitized) from `devicectl`:

```json
{
  "connectionProperties": { "transportType": "sameMachine", "tunnelState": "disconnected" },
  "deviceProperties": { "name": "iPad (10th generation)", "osVersionNumber": "18.2" },
  "hardwareProperties": { "platform": "iOS", "reality": "simulated", "udid": "…" }
}
```

**Environment:** Tuist 4.202.6  · Tuist macOS app 0.25.3 · macOS 26.5.2 (25F84) · Xcode 26.6 (17F113) · devicectl 642.4

### Error log

```sh
✖ Error
  Fetching the list of devices failed.

  Sorry this didn't work. Here's what to try next:
   ▸ If the error is not actionable, let's discuss it in the Troubleshooting & how to
   ▸ If you are very certain it's a bug, file an issue
   ▸ Check out the logs at ~/.local/state/tuist/sessions/<id>/logs.txt
```
The session log shows the preview `GET` returning `200`, immediately followed by the `Noora` "Fetching the list of devices failed." alert. The underlying `DecodingError` (unknown `TransportType` value `"sameMachine"`) is swallowed by the `catch` in `DeviceController.findAvailableDevices()`.

### macOS version

26.5.2 (25F84)

### Tuist version

4.202.6

### Xcode version

26.6 (17F113)

ThomasDutartre:
Resolves https://github.com/tuist/tuist/issues/12052

Running any preview with `tuist run <preview-url>` (or the macOS menu-bar app) fails on Xcode 26 with **"Fetching the list of devices failed."**, before any device can be selected — even for simulator targets.

**Root cause:** `DeviceController.findAvailableDevices()` decodes `xcrun devicectl list devices --json-output` with a single strict `JSONDecoder`, mapping any decode error to `DeviceControllerError.fetchingDevicesFailed`. On Xcode 26, `devicectl` also lists simulators, and every simulator entry carries `"transportType": "sameMachine"` — a value missing from the `TransportType` enum. Decoding this present-but-unknown value throws, and since the whole list is decoded at once, the entire fetch fails.

**Fix:** This applies the recommendation from @dosu in the issue:
- Add the `sameMachine` case to `ConnectionProperties.TransportType`.
- Handle it in the `PhysicalDevice` transport-type mapping (`.sameMachine` → `nil`, since simulators have no physical transport).

In addition, it decodes `hardwareProperties.reality` and skips `"simulated"` entries in `PhysicalDevice.init?`. Without this, simulators (which carry a `udid`, `name` and `platform`) would leak into the physical-device list — duplicated with the simulator-controller output and installed via `devicectl` instead of `simctl`. `reality` is decoded as a raw `String?` on purpose, to avoid reintroducing the same strict-enum fragility for a field whose value set may grow.

### How to test locally

1. On a machine with Xcode 26 and at least one simulator installed, confirm `devicectl` emits the offending value:
   ```sh
   xcrun devicectl list devices --json-output - | grep '"transportType"'
   # -> "transportType" : "sameMachine"   (for every simulator entry)
   ```
2. Run any preview:
   ```sh
   tuist run <any-preview-url>
   ```
   - Before: fails with "Fetching the list of devices failed."
   - After: device/simulator selection proceeds normally.

Automated coverage: `DeviceControllerTests.test_findAvailableDevices_skips_simulators_with_sameMachine_transport_type` feeds a payload with a physical device and a `sameMachine` simulator, and asserts the fetch succeeds and returns only the physical device.

### DIFF
--- a/cli/Sources/TuistAutomation/Device/DeviceController.swift
+++ b/cli/Sources/TuistAutomation/Device/DeviceController.swift
@@ -139,6 +139,7 @@ private struct DeviceList: Codable {
                 enum TransportType: String, Codable {
                     case localNetwork
                     case wired
+                    case sameMachine
                 }
 
                 enum TunnelState: String, Codable {
@@ -167,6 +168,10 @@ private struct DeviceList: Codable {
 
                 let udid: String?
                 let platform: Platform?
+                /// Decoded as a raw string on purpose: `devicectl` may report values
+                /// beyond "physical"/"simulated", and a strict enum would fail the whole
+                /// device list decode (the same fragility that broke on "sameMachine").
+                let reality: String?
             }
         }
     }
@@ -177,6 +182,10 @@ extension PhysicalDevice {
         // Some properties from the `devicectl` can be `nil`.
         // However, when some properties, like the `platform`, are missing, we can't properly work with the device.
         // In those cases, we return `nil` here and filter such devices out before passing them to the caller.
+        // `devicectl` also lists simulators (reality == "simulated"). Those are surfaced
+        // separately through the simulator controller, so we skip them here to avoid
+        // duplicated destinations and attempting a `devicectl` install on a simulator.
+        if device.hardwareProperties.reality == "simulated" { return nil }
         guard let udid = device.hardwareProperties.udid,
               let hardwarePlatform = device.hardwareProperties.platform,
               let name = device.deviceProperties.name
@@ -191,6 +200,7 @@ extension PhysicalDevice {
         let transportType: PhysicalDevice.TransportType? = switch device.connectionProperties.transportType {
         case .localNetwork: .wifi
         case .wired: .usb
+        case .sameMachine: .none
         case .none: .none
         }
 
--- a/cli/Tests/TuistAutomationTests/Device/DeviceControllerTests.swift
+++ b/cli/Tests/TuistAutomationTests/Device/DeviceControllerTests.swift
@@ -110,6 +110,33 @@ final class DeviceControllerTests: TuistUnitTestCase {
         )
     }
 
+    func test_findAvailableDevices_skips_simulators_with_sameMachine_transport_type() async throws {
+        // Given
+        given(commandRunner)
+            .run(arguments: .any, environment: .any, workingDirectory: .any)
+            .willProduce { arguments, _, _ in
+                self.write(text: self.sameMachineDevicesOutputJSON, at: arguments.last!)
+            }
+
+        // When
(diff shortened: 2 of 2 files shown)


ThomasDutartre:
> One thought: the current `TransportType` enum will still throw on any _future_ unknown value Apple might add (beyond `sameMachine`). It might be worth considering a custom `init(from:)` or an `.unknown` fallback case to future-proof it, but that's a separate concern and not a blocker for this PR.

Should i apply this suggestion ? I'm not sure this is a great idea to anticipate future cases like that

icedice:
I would love seeing this merged in since previews are basically not working with Xcode 26 right now...

fortmarek (approved):
Really sorry this took so long to merge and thanks for the PR!

PR #12189, opened 2026-07-31
fix(swifterpm): keep resolution when SwiftPM fails to fetch binary artifacts

pepicrft:
## What changed

`swifterpm` no longer fails a resolution when the `swift package resolve` it shells out to fails while fetching binary artifacts.

- `SystemProcess.outcome` returns a finished process instead of throwing on a non-zero exit, and `SystemProcess.run` becomes a thin wrapper over it, so existing callers behave exactly as before.
- A forwarded run now tees the child's standard error rather than handing it to the terminal, so its diagnostics stream out live and a bounded copy is kept for the caller.
- `PackageResolver` treats a failed resolve whose diagnostics are exclusively SwiftPM binary-artifact fetch failures as successful, provided it produced pins, and prints a `note:` explaining why the command continued.

## Why

Reported in Slack by a customer on 4.202.2 and again on 4.203.1:

```
error: failed downloading 'https://github.com/adjust/adjust_signature_sdk/releases/download/v3.35.2/AdjustSigSdk-iOS-tvOS-Dynamic-3.35.2.xcframework.zip'
which is required by binary target 'AdjustSignature':
/Users/…/.cache/swifterpm/artifacts/https___github_com_adjust_…_xcframework_zip already exists in file system
error: fatalError
```

It reads like network flakiness. It is not. It is deterministic, and it never recovers on its own, which is why upgrading the command line interface changed nothing.

## Root cause

The defect is in the Swift Package Manager, in `Workspace+BinaryArtifacts.swift`. After version solving, it downloads the graph's binary artifacts into a shared cache entry keyed only by the artifact URL. It checks that entry for existence *before* starting the download, then moves the finished file onto the same path afterwards. Two downloads of one URL therefore both miss the cache, both download, and the loser fails moving its copy onto the winner's.

Two ways to trigger it, both reproduced here against the real Adjust archive:

- one artifact URL declared by two binary targets in a single graph, which races inside one process and fails deterministically on a cold cache
- two resolves sharing a cache directory, which races across processes

The failure handler then deletes the cache entry, so the next run starts cold and fails identically. For a graph with a duplicated artifact URL, resolution never succeeds again.

`swifterpm` is exposed because it passes its own cache root to the Swift Package Manager as `--cache-path`. Plain `swift package resolve` hits the same race with no `swifterpm` involved, so this is not a `swifterpm` defect, and it is worth fixing upstream separately.

## Approach

`swift package resolve` does two independent things: it solves versions and writes `Package.resolved`, then it downloads binary artifacts. `swifterpm` only needs the first. It downloads, checksum-verifies and materializes binary artifacts itself immediately afterwards, using its own retrying client and its own cache, so the Swift Package Manager's artifact stage is redundant work that was failing the whole command.

So the fix is not to work around the race but to stop depending on a stage `swifterpm` does not consume. This holds on its own merits: today a cold resolve downloads every binary artifact twice, once by the Swift Package Manager and once by `swifterpm`.

The classification is deliberately narrow. Only the diagnostics for *fetching* an artifact and proving it matches the manifest are tolerated, all of which `swifterpm` repeats itself. The Swift Package Manager's verdicts about an archive's *contents* that `swifterpm` does not repeat, such as refusing an archive whose symlinks escape it, are absent from the list so they stay fatal. If a single unrecognized `error:` line appears, the resolve fails exactly as before. Resolution must also have produced pins, otherwise there is nothing to carry forward.

Alternatives considered and rejected:

- Retrying the failed resolve with `--disable-dependency-cache`. It works, verified, but it costs a second full resolve on the failure path and throws away the shared repositories cache for that run.
- Seeding the Swift Package Manager's artifact cache entry from `swifterpm`'s verified archive. This meant mirroring a private cache-key mangling and writing into another tool's namespace.
- Pre-seeding `workspace-state.json` so the artifact stage is skipped entirely. The Swift Package Manager does skip when the state already lists an artifact, confirmed by testing, but `swifterpm` cannot know a cold graph's transitive binary targets before resolving, so this narrows the window rather than closing it. Worth a follow-up on top of this change.

## Impact

Users whose dependency graph resolves one binary artifact URL through more than one binary target, or who run resolves concurrently against a shared cache, stop being permanently blocked. Nothing is skipped: `swifterpm` still downloads, checksum-verifies and materializes every binary artifact, and a genuine problem with the artifact, such as a bad checksum or an unreachable URL, still fails the command through `swifterpm`'s own restore with its own message.

One behavior change to note: the resolve subprocess' standard error is now a pipe rather than the terminal, so the Swift Package Manager renders plain progress lines instead of an animated bar when `swifterpm` is run directly in a terminal. No output is lost.

The classification depends on the Swift Package Manager's diagnostic text. It fails closed, so a wording change surfaces the error as it does today. It should be deleted once a fixed toolchain is the supported floor.

## Validation

Reproduced against the real archive from the report, with a package declaring the same artifact URL through two binary targets:

- baseline binary fails with the reported error, and fails again on every subsequent run
- patched binary restores both artifacts and exits 0, printing the Swift Package Manager's error followed by the explanatory note
- two concurrent patched invocations sharing one cache directory both succeed
- pre-seeding the Swift Package Manager's cache entry makes the baseline pass, which confirms the mechanism

Negative cases still fail: an unresolvable version requirement, an invalid dependency manifest, and a missing `Package.swift`.

```
swift build --replace-scm-with-registry
swift test --replace-scm-with-registry     # 139 tests in 20 suites passed
swiftformat --lint <changed files>         # no new violations
```

New tests: an offline end to end test that points a binary target at an unroutable URL and asserts the resolution still returns pins, unit coverage for the diagnostic classifier including the manifest error that merely mentions a binary target, and coverage for the teed standard error.

### How to test locally

```bash
cd swifterpm && swift test --replace-scm-with-registry --filter "ResolveTests|SupportTests"
```

To see the original failure, create a package whose manifest declares the same remote `binaryTarget` URL twice, then run `swift package resolve --cache-path <dir>` on a cold cache and observe the "already exists in file system" error. Running the same package through the `swifterpm` built from this branch succeeds.

### DIFF
--- a/swifterpm/Sources/swifterpm/Resolve.swift
+++ b/swifterpm/Sources/swifterpm/Resolve.swift
@@ -65,7 +65,7 @@ enum PackageResolver {
             scmToRegistryTransformation: scmToRegistryTransformation,
             useExistingResolvedFile: useExistingResolvedFile,
             writeResolvedFile: writeResolvedFile,
-            forwardOutput: progress != nil
+            progress: progress
         )
         resolved.originHash = originHash
         resolved.pins = dedupePinsByIdentity(resolved.pins)
@@ -93,8 +93,9 @@ enum PackageResolver {
         scmToRegistryTransformation: SCMToRegistryTransformation,
         useExistingResolvedFile: Bool,
         writeResolvedFile: Bool,
-        forwardOutput: Bool
+        progress: ResolutionProgressReporter?
     ) async throws -> ResolvedPins {
+        let forwardOutput = progress != nil
         let resolvedPath = packageDir.appendingPathComponent("Package.resolved")
         let snapshot =
             (!writeResolvedFile || !useExistingResolvedFile)
@@ -104,7 +105,7 @@ enum PackageResolver {
         }
 
         do {
-            try await SystemProcess.run(
+            let outcome = try await SystemProcess.outcome(
                 "swift",
                 swiftPackageResolveArguments(
                     packageDir: packageDir,
@@ -118,7 +119,21 @@ enum PackageResolver {
                 workingDirectory: packageDir,
                 forwardOutput: forwardOutput
             )
+
+            if !outcome.succeeded, !failedOnlyOnBinaryArtifacts(outcome) {
+                throw resolveFailure(outcome, forwardedOutput: forwardOutput)
+            }
             let resolved = try await ResolvedFile.read(packageDir: packageDir)
+            if !outcome.succeeded, resolved.pins.isEmpty {
+                // Resolution is only proven to have finished by the pins SwiftPM wrote before
+                // it moved on to artifacts. Without them there is nothing to carry forward.
+                throw resolveFailure(outcome, forwardedOutput: forwardOutput)
+            }
+            if !outcome.succeeded {
+                progress?.note(
+                    "swift package resolve failed to fetch binary artifacts; swifterpm restores them itself and continued with the resolved versions"
+                )
+            }
             if !writeResolvedFile, let snapshot {
                 try await restoreResolvedFile(snapshot, at: resolvedPath)
             }
@@ -130,6 +145,63 @@ enum PackageResolver {
             throw error
         }
     }
+
+    /// A forwarded run already printed its diagnostics, so the thrown error carries only the
+    /// exit status to avoid repeating them.
+    private static func resolveFailure(
+        _ outcome: SystemProcess.Outcome,
+        forwardedOutput: Bool
+    ) -> ToolError {
+        ToolError.message(
+            forwardedOutput ? outcome.terminationStatusDescription : outcome.failureMessage
+        )
+    }
+
+    /// `swift package resolve` runs two independent stages: it solves versions and writes
+    /// Package.resolved, then it downloads the graph's binary artifacts. swifterpm only needs
+    /// the first stage, since it downloads, verifies and materializes binary artifacts itself
+    /// right after resolution, so a failure confined to SwiftPM's artifact stage says nothing
+    /// about the resolution swifterpm asked for and must not fail it.
+    ///
+    /// That stage fails for reasons swifterpm's own restore does not share. SwiftPM keys its
+    /// shared artifact cache on the URL alone and checks that entry for existence before
+    /// downloading rather than when moving the finished file into place, so two downloads of
+    /// one URL (two binary targets declaring it, or two resolves sharing a cache directory)
+    /// both miss, both download, and the loser fails moving its copy onto the winner's. The
+    /// failure also deletes the entry, so the next resolve starts cold and fails identically:
+    /// for a graph with a duplicated artifact URL, resolution never succeeds on its own.
+    ///
+    /// Only SwiftPM's own binary-artifact diagnostics count. Anything else in the output,
+    /// including a manifest error that merely mentions a binary target, propagates as before.
+    static func failedOnlyOnBinaryArtifacts(_ outcome: SystemProcess.Outcome) -> Bool {
+        let errors = (outcome.stderrString + "\n" + outcome.stdoutString)
+            .split(separator: "\n")
+            .map { $0.trimmingCharacters(in: .whitespaces) }
+            .filter { $0.hasPrefix("error: ") }
+            .map { String($0.dropFirst("error: ".count)) }
+            // swift-package prints this after the diagnostics that caused it to give up.
+            .filter { $0 != "fatalError" }
+
+        guard !errors.isEmpty else { return false }
+        return errors.allSatisfy { error in
+            binaryArtifactDiagnosticPrefixes.contains { error.hasPrefix($0) }
+                && error.contains("binary target '")
+        }
+    }
+
+    /// Only the diagnostics for fetching an artifact and proving it matches the manifest, all of
+    /// which swifterpm repeats itself: it downloads over its own retrying client, compares the
+    /// checksum against the same manifest value, and rejects an archive that carries no binary
+    /// artifact. SwiftPM's verdicts about an archive's *contents* that swifterpm does not repeat,
+    /// such as refusing an archive whose symlinks escape it, are deliberately absent so they stay
+    /// fatal.
+    private static let binaryArtifactDiagnosticPrefixes = [
+        "failed downloading '",
+        "failed extracting '",
+        "failed validating archive from '",
+        "invalid archive returned from '",
+        "checksum of downloaded artifact of binary target '",
+    ]
 
     private static func swiftPackageResolveArguments(
         packageDir: URL,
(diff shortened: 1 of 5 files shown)


swifterpm/Sources/swifterpm/Resolve.swift
```diff
@@ -131,6 +146,63 @@ enum PackageResolver {
...
+
+    /// Only the diagnostics for fetching an artifact and proving it matches the manifest, all of
+    /// which swifterpm repeats itself: it downloads over its own retrying client, compares the
+    /// checksum against the same manifest value, and rejects an archive that carries no binary
+    /// artifact. SwiftPM's verdicts about an archive's *contents* that swifterpm does not repeat,
+    /// such as refusing an archive whose symlinks escape it, are deliberately absent so they stay
+    /// fatal.
+    private static let binaryArtifactDiagnosticPrefixes = [
+        "failed downloading '",
+        "failed extracting '",
```
esnunes, line 200: One thing I wanted to flag about the tolerated prefix list, and I think it is worth resolving before this merges.

The doc comment just above says SwiftPM's verdicts about an archive's contents, "such as refusing an archive whose symlinks escape it", are deliberately absent so they stay fatal. I do not think that holds, because in SwiftPM that verdict is not its own diagnostic. `validateNoEscapingSymlinks` is called right after `archiver.extract`, and its throw is caught by the same handler that emits `remoteArtifactFailedExtraction` / `localArtifactFailedExtraction`:

    error: failed extracting '<url>' which is required by binary target '<name>': symlink '<e>' points to '<p>' which is outside '<root>'

That matches `"failed extracting '"` and contains `binary target '`, so `failedOnlyOnBinaryArtifacts` returns true and the resolve is tolerated. And on our side `extractBinaryArtifactArchive` shells out to `/usr/bin/unzip -q` with no equivalent check, so nothing re-applies it afterwards. Every other prefix in the list lines up with something swifterpm genuinely re-does; this is the one row where the stated principle and the list disagree.

The cheapest fix looks like dropping `"failed extracting '"` from the list. The collision this PR targets surfaces as `failed downloading '...'`, so removing it does not weaken the fix, and the classifier fails closed anyway. The other direction would be to keep the prefix and add an escaping-symlink check to `extractBinaryArtifactArchive`, which would make the comment true again.

For reference, I read this in swiftlang/swift-package-manager at main: `Sources/Workspace/Workspace+BinaryArtifacts.swift` for the call sites, `Sources/Workspace/Diagnostics.swift` for the message text, and `Sources/Basics/FileSystem/FileSystem+Extensions.swift` for `validateNoEscapingSymlinks`. Worth a quick re-check against the pinned toolchain in case its wording differs.

swifterpm/Sources/swifterpm/Resolve.swift
```diff
@@ -118,7 +119,21 @@ enum PackageResolver {
...
+            if !outcome.succeeded, !failedOnlyOnBinaryArtifacts(outcome) {
+                throw resolveFailure(outcome, forwardedOutput: forwardOutput)
+            }
             let resolved = try await ResolvedFile.read(packageDir: packageDir)
+            if !outcome.succeeded, resolved.pins.isEmpty {
+                // Resolution is only proven to have finished by the pins SwiftPM wrote before
+                // it moved on to artifacts. Without them there is nothing to carry forward.
+                throw resolveFailure(outcome, forwardedOutput: forwardOutput)
+            }
+            if !outcome.succeeded {
```
esnunes, line 132: Two consequences of `forwardOutput = progress != nil`, neither blocking.

The explanation goes out through `progress?.note(...)`, and `progress` is nil exactly when stderr was captured instead of forwarded. So under `--quiet` a tolerated binary-artifact failure is completely silent: SwiftPM's error went into a buffer we then drop, and the note is suppressed. The sharpest case is `swifterpm resolve --quiet --print-only` (or the library API with restore off), which now exits 0 with no output at all, and since restore never runs nothing else fetches the artifact either. There is no way for the caller to tell that apart from a fully successful resolve. Sending the note to stderr regardless of the reporter would cover it.

Separately, the two modes classify from different text. The forwarded branch leaves `Outcome.stdout` empty, so `failedOnlyOnBinaryArtifacts` decides on stderr alone, while the captured branch sees both. That is fine against today's SwiftPM, which puts its resolve diagnostics on stderr. The thing I would flag is that the stdout concatenation reads like coverage the forwarded path does not actually have: if any of these diagnostics ever moved to stdout, the tolerance would quietly stop working in the interactive path while the unit tests, which feed stderr directly, kept passing. Teeing stdout too would make the two paths identical.

swifterpm/Sources/swifterpm/Support.swift
```diff
@@ -86,21 +109,62 @@ enum SystemProcess {
...
+        let outcome = try await outcome(
+            executable,
+            arguments,
+            workingDirectory: workingDirectory,
+            environment: environment,
+            forwardOutput: forwardOutput,
+            outputLimit: outputLimit
+        )
+
+        guard outcome.succeeded else {
```
esnunes, line 121: Heads-up, nothing to change here if you would rather keep it. After this change nothing calls `run(forwardOutput: true)` any more: grepping `Sources/` for `forwardOutput` only turns up `Resolve.swift:98` and `:120`, and both feed `outcome` rather than `run`. That leaves the `forwardOutput ? terminationStatusDescription : failureMessage` ternary inside `run` unreachable, and `resolveFailure` is the same ternary written out again a file away.

Folding it into one place on the type that owns both values, something like `Outcome.error(forwardedOutput:)`, would leave one copy of the rule instead of two that can drift apart.

esnunes (approved):
Really liked this one. Diagnosing it as an upstream race rather than flakiness, and then fixing it by dropping a dependency on a stage we never consumed rather than working around the race, is the stronger of the two available moves, and the write-up of the rejected alternatives makes it easy to review. The classifier is also built to fail closed in the three ways that matter: allSatisfy rather than contains, a non-empty error set required, and pins required, so an unrecognised diagnostic degrades to today's behaviour rather than to a silent skip.

I checked the five prefixes against SwiftPM's `BinaryArtifactsManagerError` and all five are real, current strings. Four of them line up cleanly with something swifterpm genuinely re-does itself. The one I would want to settle before merging is `"failed extracting '"`: that is the diagnostic SwiftPM emits when `validateNoEscapingSymlinks` rejects an archive after extraction, and `extractBinaryArtifactArchive` shells out to `unzip` with no equivalent check, so it is the one row where the stated principle and the list come apart. Details inline.

A few things I checked that came back clean, so you do not have to re-do them: SwiftPM does write `Package.resolved` before the artifact stage even when the pins move, the `--force-resolved-versions` re-check is safe because `writeWorkspaceState` seeds the artifacts first, and the new closure overload of `Subprocess.run` binds with `input: .none`, so stdin behaviour is unchanged.

Everything else is small: the silence under `--quiet`, and a bit of duplication left behind by the run/outcome split.

PR #12202, opened 2026-08-02
fix(cli): remove automatic static XCFramework force-loading

### LINKED ISSUE #12183: Reconsider generic force-loading of static binary-target wrappers
### Why is this needed?

This follows #10947 and its implementation in #10953. While #10947 correctly identified empty wrappers, it may be useful to reconsider the generic `-force_load` behavior added in #10953.

An empty SwiftPM wrapper target is a normal package layout mechanism. It exposes a product in the dependency graph and connects a binary target with its transitive dependencies. It does not imply that a dynamic wrapper must absorb or re export every object in the underlying static archive.

Before #10953, static binaries used normal archive selection semantics. Since the empty wrapper referenced no archive symbols, no members were selected and the wrapper remained empty. That is expected behavior for a static archive. Declaring the SwiftPM product as `.framework` alone does not request that all members be included in a dynamic library.

Force loading can help simple Swift or Objective C libraries, but it is not safe as a generic rule for vendor static XCFrameworks. Such archives may contain private ObjC metadata, duplicate members, or unused objects with unresolved dependencies. These details are harmless under ordinary archive selection but can fail when every member is loaded.

### Reproduction

Repo: https://github.com/Loupehope/tuist-force-load-repro
Pinned version: Tuist 4.204.0-rc.1
Run: `mise install; ./verify.sh`

The repro contains three static XCFramework cases, each exposed through an intentionally empty wrapper target:

1. PrivateObjC: force loading places private ObjC metadata in the wrapper. Importing the binary module autolinks another copy, leaving the same class metadata in two dynamic images.
2. CxxDuplicates: the archive contains the same dependency object twice. Normal selection avoids the duplicate. Force loading loads both and produces a duplicate C++ symbol error.
3. CxxUndefined: an unused debug only object has an unresolved `debug::Probe::Initialize()` reference. Normal selection ignores it. Force loading selects it and the link fails.

### Results

With static products, all three consumers build successfully.

With dynamic products on Tuist 4.204.0-rc.1, Tuist emits `-Wl,-force_load` for the static XCFrameworks. The ObjC case only builds because another copy is autolinked. The C++ cases fail with duplicate and undefined symbol errors.

Converting these fixtures explicitly to dynamic XCFrameworks makes all consumers build. The conversion handles required ObjC metadata, duplicate native members, and the debug only object without manual force load or re export flags.

### Requested change:

Consider removing or disabling by default the automatic `-Wl,-force_load` injection introduced by #10953 for direct static XCFramework dependencies of dynamic targets. If retained, an explicit opt in path may be preferable.

Related PR: https://github.com/tuist/tuist/pull/12202

Loupehope:
### Summary

Resolves https://github.com/tuist/tuist/issues/12183

Removes the automatic -Wl,-force_load,... linker settings added in #10953 for static XCFrameworks directly linked by dynamic targets.

The mapper now operates on the original graph again and no longer mutates dynamic targets with inferred linker flags.

### Root cause

The automatic force-load behavior is unsafe for arbitrary vendor static XCFrameworks.

-force_load pulls every archive member into the final link, including object files that the linker would otherwise leave unselected. This can introduce duplicate C/C++ symbols, duplicated Objective-C metadata, and undefined symbols from previously unreferenced archive members.

The mapper also parsed module maps to infer additional linker flags, making the direct-link behavior broader than the graph dependency alone.

### Solution

Removed the direct dynamic-target → static-XCFramework force-load path, including:

- mapDynamicTargetsLinkedToStaticXCFrameworks;
- inferred linker settings and module-map linker-flag parsing;
- forceLoadPath;
- focused tests that covered only this automatic behavior.

The existing mapper behavior for static XCFrameworks reached through dynamic XCFramework dependencies remains unchanged. In particular, it continues to preserve platform conditions, matching XCFramework slice selection, module-map generation, and framework/header search paths.

Projects that need forced loading can continue to provide explicit linker flags through PackageSettings.targetSettings.


Loupehope:
@fortmarek The purpose of this pull request was to remove the -force-load logic entirely, not to keep it :(

fortmarek:
@Loupehope I'm still working on this, I'll post what I can find in a bit.

fortmarek:
@Loupehope I went with your take in the end, after some back-and-forth. @shgew I'd suggest you take the workaround suggested by @Loupehope. When you force a product type of a given SwiftPM target, you may create invalid combinations that may need extra settings, and I think that's expected. I don't see a way how to provide a good generic solution that wouldn't break use cases where users _don't_ override a product type, which is what _primarily_ needs to work.

shgew:
@fortmarek sounds good. Thanks!

fortmarek:
Thanks @Loupehope for the PR!

Issue #12208, opened 2026-08-03
Registry: a cached archive generation may survive an overwrite


esnunes:
> **Rescoped.** This issue originally covered two things: cross-region read divergence, and object caching serving a superseded archive. The registry bucket has since been changed from globally distributed to single-region (Frankfurt), which resolves the first — Tigris guarantees strict read-after-write consistency within a region, and there is no longer a second replica to fall behind. What remains is the caching question below.

Object caching is [enabled by default on all buckets](https://www.tigrisdata.com/docs/objects/caching/) and is populated by request pattern rather than by bucket configuration — *"objects stored in San Jose but requested frequently from Sydney will result in getting cached in the Sydney region"*. Making the bucket single-region controls where the object lives, not where it is cached.

The documentation does not state whether overwriting an object invalidates its cached copies. If it does not, then after any archive overwrite a consumer routed to a cache node holding the previous generation receives bytes that no longer match the checksum the registry advertises — reported as `invalid registry source archive checksum` on a version that is entirely healthy in storage.

This is not hypothetical: it is what happened to `adyen/adyen-3ds2-ios@2.3.0` before the region change. What is now unknown is whether the caching layer alone can still produce it.

## Open questions

1. Does an overwrite invalidate cached copies of the object, and if not, what is the cache lifetime?
2. Is a purge available for a specific key?
3. Is object caching active for single-region buckets, and can it be disabled for this bucket if the answer to (1) is unfavourable?

These are questions for the storage provider. Answers determine whether anything further is needed here.

## Evidence from the original occurrence

Kept because it is the clearest demonstration of the failure, and because two attempted mitigations are worth not retrying.

The same presigned URL, fetched twice within a second, differing only in a request header:

| | sha256 of served bytes | `Server-Timing` | `X-Tigris-Served-From` |
|---|---|---|---|
| As the registry issues it | `7ffdf974…` | `cache;desc=read` | `iad1` |
| Same URL + `X-Tigris-Consistent: true` | `4f6cc437…` | `server;desc=meta` | `fra` |

The catalog advertised `4f6cc437…`. A full 17,613,886-byte download with the header hashed to exactly that; without it, the same URL returned an archive generation written 26 seconds earlier.

Two mitigations were tried and neither reached the cached copy:

1. **Rewriting the object.** Re-uploaded via a single consistent `put_object`, producing a fresh generation at origin with the correct digest. Reads without the consistency header continued to return the superseded one.
[...]

Issue #12195, opened 2026-08-01
Registry: archive construction is not deterministic, so any resync of a non-passthrough package breaks pins


esnunes:
Rebuilding a registry version is not a safe operation today: archive construction is not reproducible, so any resync of an affected package publishes a new checksum and breaks every `Package.resolved` pinned to it.

This is independent of any incident. It is the mechanism that turns a routine resync into a broken pin.

> **Scope corrected.** This issue originally attributed the problem to the clone path only, on the strength of two builds of one package through the repack path coming out byte-identical. Production has since shown the repack path is *also* non-deterministic. Both paths share `zip_directory/2`, so the proposed fix is unchanged — but the affected population is much larger than "packages with submodules".

## Measurement

| Path | Selected when | Two independent builds |
| --- | --- | --- |
| Zipball passthrough | default | identical (GitHub's bytes) |
| `unzip` → `zip -r -y` | archive contains symlinks | **different** |
| `git clone` → `zip -r -y` | repository has submodules | **different** |

Only passthrough is reproducible, because those bytes are GitHub's rather than ours.

**Clone path**, measured directly: two clones three seconds apart produced different archives. `git clone` writes *checkout* time to the filesystem and `zip -r` records what it finds. The divergence is at second granularity, so a `unzip -l` listing can show identical minutes while the bytes differ.

**Repack path**, observed in production on `Adyen/adyen-3ds2-ios@2.3.0` — no `.gitmodules`, 5 symlink entries in the zipball, so it routes through `unzip` → `zip -r -y`:

```
15:23:33  attempt 1 uploads  etag fd0480f1…-4  17,613,886 bytes  → verification fails
15:23:59  attempt 2 uploads  etag db59e41c…-4  17,613,886 bytes  → verification passes
```

Same size, different bytes, 26 seconds apart, same input. Restoring mtimes from the zipball — which is why an earlier two-build test of a simpler package came out stable — is not sufficient on its own. The remaining variance is most likely entry ordering: `zip -r` walks the directory, and after `resolve_symlinks/1` rewrites entries the traversal order need not be stable between runs.

## Why this is worse than one changed checksum

A verification failure triggers an Oban retry, and **the retry rebuilds from scratch**. Because the build is not reproducible, the retry produces *different* bytes again rather than reproducing the ones that failed to verify. So a transient failure does not converge — each attempt uploads a fresh archive and leaves the previous one behind.

[...]

PR #12209, opened 2026-08-03
fix(server): re-read a registry archive before declaring it unverified

esnunes:
## What changed

`ReleaseWorker` re-reads the stored archive up to six times, with a linear backoff, before concluding that a publish failed verification. Errors that cannot resolve by waiting still fail on the first read.

## Why

A read immediately after an upload can miss. `upload_file/3` goes through the multipart path, and `S3.request(%Upload{})` skips `with_tigris_consistency/1` — the consistency header is only added to single requests — so a completed write is not necessarily visible to the verification read that follows it.

Both failure shapes the verification can produce mean that lag far more often than they mean the archive is wrong:

| Prior object at the key? | What the read returns |
|---|---|
| No | `404` → `stored_archive_unverifiable` |
| Yes | the previous generation's digest → `stored_archive_mismatch` |

Both occurred in production within 47 seconds of each other, on unrelated packages.

## Root cause of the damage this prevents

Failing the job on the first read is worse than the fault it reports. Oban restarts `perform/1` from the top, so the retry re-runs the **build** — and archive construction is not deterministic (#12195). The retry therefore uploads *different* bytes to the same key while the first write may still be settling.

One version has already diverged permanently this way:

```
15:23:33  attempt 1  built A → uploaded → read back the previous generation  ✗ failed
15:23:52  attempt 2  built B → uploaded → read back B                        ✓ catalog := B
```

Three digests for one version. The two overlapping writes settled with storage keeping A and the catalog advertising B, and the divergence is not self-correcting. The other case that day recovered on its own — the difference between them was only whether the rebuild happened to produce equivalent bytes.

So the recovery path was more dangerous than the error it was recovering from. Re-reading costs a few seconds; rebuilding costs a GitHub refetch, a repack, a second upload, and the risk above.

## Choices worth flagging

**Which errors retry.** Only a stale digest and a missing object. A rejected read — permissions, configuration — will not become an accepted one by waiting, so it fails immediately rather than spending the budget. There is a test asserting exactly one read in that case.

**No jitter**, unlike the metadata-lock retry directly above it in the same module. That one exists to spread sibling jobs contending for a shared lock; here each job waits on its own object, so there is no herd.

**A recovered read is logged.** How long a write takes to become visible is otherwise unobservable — the job succeeds either way, so without this the only trace in production is the failures. The log line reports which attempt succeeded, which is what would tell us whether six attempts is the right budget.

**Backoff is configurable** so tests do not sleep. Default 500 ms linear, ~7.5 s total.

## What this does not fix

Publishing is still not safe against a concurrent overwrite. As long as archives live at a mutable key, a rebuild can collide with an in-flight write to the same key, and a stale cached copy of the old generation can outlive the new one (#12208). The structural fix is content-addressed object keys, where a rebuild that produces different bytes writes a different key and the catalog switches between immutable objects. This change removes the trigger; it does not remove the hazard.

## Validation

`mix test test/tuist/registry/swift/release_worker_test.exs` — 17 passed, including two new cases:

- a first read that 404s, then succeeds: asserts the release publishes, the catalog records the uploaded digest, and exactly two reads happened
- a read rejected with a 403: asserts the failure is immediate and the catalog is untouched

The existing mismatch test was updated from 2 storage requests to 7, which is the assertion that a stale digest is now re-read rather than trusted on the first look.

## Related

- #12195 — non-deterministic archive construction, which is what turns a retried upload into two competing objects rather than two identical ones
- #12208 — archive downloads bypass storage consistency; same underlying gap, on the read path, where it is not guarded at all

### DIFF
--- a/server/config/test.exs
+++ b/server/config/test.exs
@@ -74,6 +74,8 @@ config :tuist, Tuist.OpsClickHouseRepo,
     session_timezone: "UTC"
   ]
 
+config :tuist, Tuist.Registry.Swift.ReleaseWorker, verification_backoff_ms: 0
+
 # Configure your database
 #
 # The MIX_TEST_PARTITION environment variable can be used
--- a/server/lib/tuist/registry/swift/release_worker.ex
+++ b/server/lib/tuist/registry/swift/release_worker.ex
@@ -26,6 +26,8 @@ defmodule Tuist.Registry.Swift.ReleaseWorker do
   @metadata_lock_max_attempts 5
   @metadata_lock_backoff_ms 200
   @metadata_lock_snooze_seconds 30
+  @verify_max_attempts 6
+  @verify_backoff_ms 500
   # Symlinks nested inside these code-signed bundles are part of the bundle's
   # sealed layout (e.g. a Mac Catalyst framework's `Versions/Current -> A` and
   # the `Binary`/`Resources` links). Code signing records them as symlinks in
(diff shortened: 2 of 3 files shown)


esnunes:
Closing: no longer needed.

The registry bucket has been changed from globally distributed to single-region (Frankfurt). Tigris guarantees strict read-after-write consistency within a region — eventual consistency only applies across regions — so the verification read can no longer observe a state that precedes the upload it is checking. That includes the multipart path.

Both production failures that motivated this change happened under the global bucket, and are explained by a read being served from a region other than the one written to: one read returned a previous generation of the object, the other returned 404 for an object that had just been uploaded. Neither is reachable in a single-region setup.

The retry therefore has no failure mode left to cover, and adding a code path that can never execute is not worth the maintenance.

PR #12242, opened 2026-08-07
fix(cli): safely deep merge array build settings from SPM manifests and baseSettings

### LINKED ISSUE #12241: Regression in 4.202.0: SPM `enableExperimentalFeature` flags are destructively clobbered by `baseSettings`
## What happened?
We are running into a regression introduced exactly in `4.202.0` when generating projects for external SPM dependencies that heavily rely on experimental compiler flags (such as `apple/swift-collections` >= 1.3.0 which requires `Lifetimes`, `BuiltinModule`, etc. on Xcode 26/Swift 6.2).

In Tuist `4.201.0` and older, generating a project with `swift-collections` compiles successfully. However, starting in `4.202.0`, Tuist drops or fails to serialize the `.enableExperimentalFeature` and `AvailabilityMacro` definitions provided in the librarys `Package.swift` if the project uses a global `baseSettings`.

In the generated `.pbxproj` in 4.202.0, the external targets (`InternalCollectionsUtilities`, `RopeModule`, etc.) end up with a vanilla `OTHER_SWIFT_FLAGS = "$(inherited)"`. This causes the Xcode 26 compiler to evaluate the code as standard Swift, throwing strict memory errors across the library:
* `error: @_lifetime attribute is only valid when experimental feature Lifetimes is enabled`
* `error: a function cannot return a ~Escapable result`

### Root Cause
This was introduced in PR #11574 (*fix(cli): propagate package base settings to macro targets*).
The `PackageInfoMapper.swift` (line ~2139) now merges the `baseSettings` directly on top of the SPM `settingsDictionary` using `uniquingKeysWith: { _, new in new }`.
This causes a destructive overwrite: if the user supplies `OTHER_SWIFT_FLAGS: ["$(inherited)", "-DSTAGING"]` in `baseSettings`, it completely erases the `OTHER_SWIFT_FLAGS` array constructed from the SPM manifest (which held the `enableExperimentalFeature` flags).

## Proposed Solution
When mapping array settings like `OTHER_SWIFT_FLAGS` or `OTHER_LDFLAGS`, `PackageInfoMapper` should inspect if the users base settings array contains `"$(inherited)"`.
* If it **does** contain `$(inherited)`, Tuist should perform a deep merge (concatenating the SPM flags and the user flags).
* If it **does not**, it should clobber the SPM flags (respecting the users intent to overwrite).

### How do we reproduce it?

I have attached a zipped minimal reproducible project (`tuist-repro.zip`). 

[tuist-repro.zip](https://github.com/user-attachments/files/30805312/tuist-repro.zip)

1. Unzip the project, cd tuist-repro and run `tuist install && tuist generate` using Tuist **4.201.0**.
2. Build the `tuist-repro` scheme targeting iOS. **(Build Succeeds)**
3. Clean, and run `tuist install && tuist generate` using Tuist **4.202.0** or newer.
4. Build the `tuist-repro` scheme targeting iOS. **(Build Fails with `~Escapable` errors)**

### Error log

`error: @_lifetime attribute is only valid when experimental feature Lifetimes is enabled`
`error: a function cannot return a ~Escapable result`

### macOS version

26.4.1

### Tuist version

4.202.0

### Xcode version

26.3.0

jonurq-peya:
## What changed
This PR updates the `SettingsDictionary.merge` behavior in `PackageInfoMapper` to perform a "deep merge" on array settings (like `OTHER_SWIFT_FLAGS` and `OTHER_LDFLAGS`) when mapping SwiftPM packages, rather than destructively clobbering them.

When `baseSettings` contains an array setting with `"$(inherited)"`, Tuist now concatenates the user`s array with the array generated from the SwiftPM manifest, ensuring both sets of flags are preserved in the final `.xcconfig`.

## Why
In PR #11574, logic was added to propagate `baseSettings` to SwiftPM targets using `uniquingKeysWith: { _, new in new }`. This unintentionally caused global string arrays (like `-DSTAGING` injected via `OTHER_SWIFT_FLAGS`) to completely wipe out critical compiler flags generated by SwiftPM (such as `.enableExperimentalFeature("Lifetimes")` used heavily by `swift-collections` >= 1.3.0).

Without these experimental flags, generating projects for Swift 6.2 / Xcode 26 results in fatal compilation errors (`error: a function cannot return a ~Escapable result`) for modern Apple libraries.

## Approach
1. Extended `SettingsDictionary` with a `deepMerge` function that respects Xcode`s idiomatic `$(inherited)` token.
2. In `PackageInfoMapper.swift`, replaced the destructive `.merge()` with `.deepMerge()` when applying `baseSettings` to the resolved SPM `settingsDictionary`.
3. Added unit tests to ensure that scalar values (like `MACOSX_DEPLOYMENT_TARGET`) still clobber as intended, while arrays safely concatenate if `$(inherited)` is present.

Resolves #12241


fortmarek:
Thanks for tracking down the regression and putting together the original fix. The diagnosis was correct: starting in 4.202.0, package-wide `baseSettings` were copied onto every generated SwiftPM target with a last-value-wins merge. For a target whose manifest generated:

```
OTHER_SWIFT_FLAGS = ["$(inherited)", "-enable-experimental-feature \"Lifetimes\""]
```

a package setting such as:

```
OTHER_SWIFT_FLAGS = ["$(inherited)", "-DSTAGING"]
```

replaced the entire target value. The generated experimental feature flag disappeared, which led to errors such as `@_lifetime attribute is only valid when experimental feature Lifetimes is enabled` and `a function cannot return a ~Escapable result`.

The original PR addressed the common array case with a single inherited-aware deep merge. We made a few follow-up changes because package-wide and target-specific settings are applied at different layers and therefore need different policies:

1. **Package-wide settings inherit from the project instead of being copied into the target.** `PackageSettings.baseSettings` are already emitted in the generated package project's build settings. A target value containing `$(inherited)` receives those project values through Xcode's normal inheritance. Concatenating the package tokens into the target as well would apply values such as `-DSTAGING` twice, which is especially risky for order-sensitive compiler or linker options. The new `.inheritFromProject` policy keeps the SwiftPM-generated target flags, ensures they inherit, and leaves the package-wide tokens at the project level exactly once. If the package setting omits `$(inherited)`, it still replaces the target value as an explicit override.

2. **Target-specific settings retain their previous behavior.** `PackageSettings.targetSettings` are not project-level values, and their existing contract was to append arrays even when the incoming array did not contain `$(inherited)`. Applying the original deep-merge rule there would have introduced a separate clobbering regression. The explicit `.appendArrays` policy preserves that behavior.

3. **Inherited string values are supported as well as arrays.** `SettingValue` publicly supports both representations. An array-only merge left a valid setting such as `.string("$(inherited) -DSTAGING")` on the destructive fallback path, reproducing the original loss of SwiftPM flags. The inherited-aware policy now recognizes both forms.

The resulting layering is:

```
Project: $(inherited) -DSTAGING
Target:  $(inherited) -enable-experimental-feature "Lifetimes"
```

so the effective build settings contain both sources once, while target-specific overrides still append according to their established semantics.

Validation included SwiftFormat plus the complete `PackageInfoMapperTests` and `SettingsDictionaryMergeTests` suites through Xcode. All 160 selected tests passed with zero failures, including parameterized mapper coverage for both string and array package settings, explicit override behavior, missing target values, and target-specific array appending.

PR #12257, opened 2026-08-10
fix(infra): stop a mis-NAT'd runner VM from silently hanging every cache request

fortmarek:
A customer reported `CAS operation failed: ... "putKeyValue" ... The request timed out.` on Tuist runners, with 8-minute test jobs occasionally running for six hours before someone noticed and cancelled them. This is the infrastructure half of that investigation: the reason the cache became unreachable, and why nothing told us.

### What was actually happening

The runner VM's cache traffic *does* reach the kura node — and is dropped there. A per-instance kura NetworkPolicy admits `http` only from `namespaceSelector: {}` and `ipBlock: 172.16.0.0/22` (the Private Network). A macOS runner VM whose egress is not masqueraded to its host's PN VLAN address arrives from outside that block, so Cilium denies it at ingress. Silently: no RST, so the client never learns the connection failed and every module-cache download and Xcode CAS operation hangs until its own timeout.

Everything else looked healthy, which is why this ran for two weeks. The kura pod had 0 restarts, `kura_http_inflight_requests` = 1, no 5xx, and almost no traffic at all. Every other job on the same pool, hitting the same pod, finished in 146-1283s.

`cilium_drop_count_total{reason="Policy denied", direction="INGRESS"}` on the cache node was the only signal that fired, and it tracks each stuck job almost exactly — zero before it starts, a steady ~2/s SYN-retransmit trickle for its entire life, zero again within a scrape of the cancellation. Verified independently on two incidents:

| | Aug 8 job (03:12-09:12) | Jul 30 job (01:19-07:19) |
|---|---|---|
| before | 0 | 0 |
| during | 1.4-4.2 drops/s for 6h | 1.9-3.5 drops/s for 6h |
| after cancel | 0 | ~0.002 |

Recurring roughly 1 job in 20-40 since 2026-07-25, on a different runner pod each time.

### Changes

**The PN NAT leg no longer disappears** (`infra/macos-host-bootstrap`). It was emitted only once the VLAN device already held an inet address, so while DHCP had not landed the ruleset carried no PN leg at all and a VM booting into that window ran its whole job un-NAT'd. It now keys on the device *existing*: `-> ($PNIF)` is a dynamic interface reference, so pf re-resolves the address as it arrives and emitting the rule early is correct. An addressed VLAN still wins when several exist, so a leftover from a re-attachment cannot shadow the live one. A host with no PN VLAN at all now says so in the daemon log instead of degrading silently, and the general-internet leg carves the PN out of its destination so cache traffic can never be masqueraded onto the public NIC while the PN route is briefly absent.

**The drift loop's terminal Failed state now expires** (`infra/cluster-api-provider-tuist`). It previously lifted only on a new `HostConfigHash`, which is right for a host that *rejected* a config but never fires for the far more common verdict — a host the operator could not reach (`dial tcp ...:22: i/o timeout`). Those stayed terminal indefinitely while remaining Ready and schedulable: still taking jobs, pinned to whatever config was last pushed, so a fleet-wide networking fix could roll out and miss them. Three production hosts are in exactly that state right now. A new `--tartkubelet-terminal-retry-after` (default 30m) re-arms them off `Status.LastUpdateFailureTime`. A genuinely broken config still backs off to one attempt budget per cooldown rather than per reconcile, so the retry cap keeps doing its job.

**An alert on the drop counter**, documented in `alerts.md` and created in Grafana.

### Reviewer notes

- **The new status field had to be registered in the CRD by hand.** `infrastructure.cluster.x-k8s.io_scalewayapplesiliconmachines.yaml` has a structural status schema with no `x-kubernetes-preserve-unknown-fields`, so an unregistered `Status.X` is pruned by the API server, reads back empty every reconcile, and would leave the machine drifting forever. That file is hand-maintained (it carries chart labels beyond raw controller-gen), so it is edited surgically rather than regenerated.
- **The three stuck hosts should self-heal on deploy.** This PR changes `renderVMNATScript`, which is part of `HostConfigHash`, so the desired hash moves away from each host's recorded `failedHostConfigHash` and the *existing* drift exit clears them — the cooldown is the backstop for whichever are still unreachable. No manual patch needed unless you want them unstuck before the deploy.
- **Change 1 is hardening, not a proven single trigger.** The mechanism is confirmed — packets arriving policy-denied — but `cilium_drop_count_total` carries no source address, `kube_pod_info` is not retained for runner pods, and `runner_claims`/`runner_sessions` store only `pod_name`, so I could never map a stuck job back to its Mac mini. If the alert fires again after this ships, the remaining suspect is a tailnet route covering `172.16.0.0/22` pulling cache traffic off the VLAN, which the interface-scoped NAT legs do not cover. That is a different fix.
- **The alert is already live**, not just documented: rule `ffuscyncueo74d`, folder `Alerts`, group `Runners`, receiver `Slack #notifications 2`, `severity: critical`, 10m pending. I matched the existing estate rather than `alerts.md`'s setup instructions — most rules in that document do not actually exist in Grafana, and the folder it names does not either. The metric has no `node` label, hence the `kube_pod_info` join on the Cilium agent pod.
- **Composes with #12119, which landed on `main` while this was open.** That change stops the reconcile tail overwriting the terminal `Failed` phase with `Ready`, which is why the "Runner Machine Stuck Failed" rule could never fire for those same three hosts. The two are halves of one problem: #12119 makes the terminal state *observable*, this makes it *expire*. They interlock rather than overlap — clearing the failure unpins the phase, which is what its new test asserts. `main` is merged in here, and that test's `clearUpdateFailure` call needed the `reason` argument this branch added; both changes were textually disjoint, so the merge was clean and the break only showed at compile time.
- **Not included:** the client-side half. `PutCacheValueService`/`GetCacheValueService` have neither the `.tuistCAS` short timeout nor the circuit breaker their `saveCASArtifact`/`loadCASArtifact` siblings have, which is what turns a cache outage into a six-hour build rather than a cache-less one (3,362 `putKeyValue` timeouts vs 324 `saveCASArtifact` in one job). That is a CLI change and belongs in its own PR.

### How to test locally

```bash
cd infra/macos-host-bootstrap && go test ./...
cd ../cluster-api-provider-tuist && go build ./... && go test ./...
```

The bootstrap tests cover the PN leg keying on device existence, the general-internet leg's PN carve-out, and a new `sh -n` pass over the rendered helper — a syntax error in that heredoc is invisible until pf silently holds no rules on a live host, which reads exactly like the outage this leg exists to prevent. The controller tests cover both terminal-failure exits, including a missing timestamp on CRs that predate the field.

There is no fixture for the runtime behaviour: it needs a Mac mini attached to a Scaleway Private Network. On a host after deploy, `pfctl -a com.apple/tuist.vmnat -s nat` should list the `nat on vlan0 ... to 172.16.0.0/22` leg, and the general leg's destination should read `! 172.16.0.0/22`.


infra/macos-host-bootstrap/bootstrap.go
```diff
@@ -1515,11 +1528,17 @@ if [ -n "$PNCIDR" ]; then
...
         ;;
     esac
   done
   if [ -n "$PNIF" ]; then
     RULES="${RULES}nat on $PNIF from 192.168.64.0/22 to $PNCIDR -> ($PNIF)${NL}"
+  else
+    # No PN VLAN at all: installVMCachePNInterface has not run or the
+    # attachment is gone. Nothing here can translate cache traffic, so say so
+    # in the daemon log rather than leaving a silent cache-less host.
+    echo "tuist-pf-vmnat: no vlan* interface for PN $PNCIDR; VM cache traffic will not be NAT'd" >&2
```
esnunes, line 1541: The new "no vlan* interface" stderr line is a real improvement over silence, and I noticed an interplay that might be worth a follow-up: with the PN carve-out in place, the no-VLAN state becomes invisible to the new Cilium alert. Cache traffic in that state falls to the default route, is excluded from masquerade, and dies at the upstream gateway — so it never reaches the kura node and the policy-drop counter stays at zero, while the build hangs exactly the same way. The stderr line in the per-host daemon log is then the only witness, the host stays Ready and schedulable, and nothing re-converges the VLAN (the drift loop keys on desired config, not live host state).

If that reading is right, a per-host sink with existing precedent could close the gap later: a Node condition (tart-kubelet already has the DiskPressure probe pattern) or a node_exporter textfile gauge, either of which would be alertable per host. Nothing to change in this PR — flagging it so the blind spot is a known one.

esnunes (approved):
Really thorough root-cause work — the writeup alone makes the mechanism reviewable, and the tests are the right ones (the `sh -n` pass over the rendered heredoc and the deepcopy aliasing test both guard failure modes that would otherwise be invisible). I verified independently: both Go suites pass; `pfctl -nf` on macOS accepts the negated destination and a parenthesized translation target on an addressless interface, so the dynamic-interface claim holds; a terminal CR still falls through to the 5-minute requeue, so the cooldown expiry is actually noticed; the deploy workflow applies `crds/` before helm; and the Grafana rule exists and matches alerts.md — it was actively firing while I reviewed, so the alert is already earning its keep pre-deploy.

Three notes, none blocking: a hardening gap in the vmnat helper (a failed pfctl load silently latches stale rules via the snapshot short-circuit), a monitoring blind spot (in the no-VLAN state the new carve-out keeps traffic from ever reaching the kura node, so the new alert can't see exactly that state — the stderr line is the only witness), and the pre-existing Phase="Failed" flap, which the cooldown turns periodic. The first is inline on the diff; the other two touch unchanged lines, so they're below. The change strictly improves current behavior and the three stuck hosts self-heal on deploy as described.

---

**`infra/macos-host-bootstrap/bootstrap.go:1573-1579`** — Heads-up on the vmnat helper's snapshot short-circuit — pre-existing lines, but this PR grows the ruleset variants, so it seemed worth flagging here: the `pfctl -f -` load's exit status isn't checked (the heredoc has no `set -e` — the outer `set -euo pipefail` doesn't reach it), and the `.loaded` snapshot is written unconditionally right after. On a parse-failed load pf keeps the old anchor contents, and the short-circuit's live check (`pfctl -s nat | grep -q nat`) is satisfied by those old rules — so after one failed load, every 60s run exits 0 with stale NAT rules (e.g. without the new PN carve-out), until a bootstrap re-run clears the snapshot.

Writing the snapshot only on a successful load would turn that into a self-healing retry: `printf '%s' "$RULES" | pfctl -a "com.apple/tuist.vmnat" -f - && printf '%s' "$RULES" > /usr/local/etc/tuist-vmnat.loaded`. Definitely not a blocker.

**`infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go:933-934`** — Pre-existing and out of scope for this PR, but adjacent enough to note: Phase="Failed" only survives a single reconcile. The next reconcile (triggered by the status patch itself) skips the drift block but still falls through to `machine.Status.Ready = true; machine.Status.Phase = "Ready"` at the tail — with FailureReason still set. So the phase gauge reports Ready for terminal hosts, the `if Phase == "Failed"` reset in clearUpdateFailure never matches, and the phase-persistence alert machine_metrics.go describes can't ever fire. Might be worth a small follow-up issue (persist Failed while FailureReason is set, or alert on the failure_reason label instead) — with the new cooldown, the Failed blip just recurs every ~35 minutes, which an alert would either miss or flap on.

infra/macos-host-bootstrap/bootstrap.go
fortmarek: Reading confirmed, and closed it in this PR rather than deferring: `node_exporter`'s netdev collector already reports a series per interface, so `count(node_load1) unless count(node_network_transmit_bytes_total{device=~"vlan.*"})` per host is alertable today with no Node condition or textfile gauge needed — rule `afuvzdl0z4mwwe`, alongside its companion, plus a note recording the one residual gap (a VLAN present but unaddressed is still invisible to both). One correction: the carve-out does not create the blind spot, it predates it — `to any` would have masqueraded that traffic to the public source, but an RFC1918 destination still dies at the upstream gateway and never reaches kura either way.

PR #12258, opened 2026-08-10
feat(kura): tier instance memory into a floor and a burst ceiling

fortmarek:
> Raises a managed Kura instance's memory ceiling to 4Gi, which addresses the recurring 5xx alert on `/api/cache/module/{id}`, and introduces the floor/ceiling split the shared bare-metal boxes need to hold more tenants. The mechanism ships here; the per-tier sizing that uses it is switched on in the stacked follow-up, alongside the kubelet change that makes a floor enforceable.

### What changed

- `defaultResources` in the Kura controller drives `requests.memory` from `spec.memoryFloorMib` and `limits.memory` from `spec.memoryCeilingMib`, defaulting to 2Gi/4Gi.
- `Regions.memory_profile/1` maps a billing plan to a profile: `:enterprise` gets 1Gi/4Gi, `:pro` 512Mi/3Gi, and every other plan including `:air` gets 256Mi/768Mi. Deliberately **not** an entitlement. The `Entitlements` API answers yes or no, but the question here is which of three sizes, and every governed instance gets a profile either way, so a feature grant would have no denial branch to express. Defined and tested here, not yet applied to any region (see below).
- The CAPI provider advertises `tuist.dev/memory-ceiling-mib` node capacity as a bounded multiple of allocatable, alongside the existing egress budget.
- Kura exports `kura_memory_protection_{min,low}_bytes`, the reclaim protection the orchestrator actually granted it.
- The controller publishes `requests.memory` to the container as `KURA_MEMORY_FLOOR_BYTES` via the Downward API, and Kura bounds its anonymous-memory admission budget by it.

### Why

**The 4Gi ceiling.** Kura sizes every admission pool from its cgroup limit at startup: soft and hard watermarks at 60% and 85%, and the public response-stream pool is half of what remains above the hard watermark. On 2Gi that is a 154 MiB pool and a 76-slot waiter queue, matching `kura_response_stream_pool_capacity_bytes` = 161,480,704 in production.

During a customer's nightly build fan-out the us-east instance reached a 1,669 MiB working set against the 1,740 MiB hard watermark, tipped into Constrained memory pressure, and shed public module downloads. Constrained disables the elastic borrow, so a stream can no longer take spare transient capacity and the fixed 154 MiB pool becomes the entire budget. Requests over it still queue (up to 5s, retried on every permit release and pressure transition), but the waiter queue holds one slot per stream the pool could serve concurrently, 76 here, so with ~330 clients in flight everything past the 76th waiter is rejected without waiting. `kura_memory_pressure_state` was 1 for exactly the two minutes the 503s occurred and 0 for the rest of the burst. At 4Gi the soft watermark moves to 2,458 MiB, clearing that peak by 47%.

3Gi would also have cleared it, but only by 10%, and the way it fails is worse than that margin suggests. The pressure machine enters Constrained at the soft watermark and leaves it at `0.9 x soft`. At 3Gi those are 1,843 MiB and 1,659 MiB, so the recovery threshold sits *below* a burst already observed: a marginally larger one would trip shedding and then stay shedding for the whole burst instead of recovering between peaks. At 4Gi recovery is 2,212 MiB, clear by 33%.

| ceiling | soft (sheds at) | recovers at | vs the 1,669 MiB peak |
|---|---|---|---|
| 2Gi | 1,229 | 1,106 | both below the peak |
| 3Gi | 1,843 | **1,659** | recovery 10 MiB under it |
| 4Gi | 2,458 | 2,212 | clear by 47% / 33% |

**The profile.** Every instance currently reserves 2Gi regardless of use. Measured over 14 days, 37 of 47 pods peak at 150-290 MB against that reservation while 10 peak near their limit, and no Kura box has exceeded 29% of its physical memory while reserving 65% of it. Density is capped by a number that reflects nothing.

The two halves answer different questions, which is why they are now separate:

- The **ceiling** is what the runtime reacts to, so it sets how large a burst an instance absorbs before shedding. Its pools are semaphores rather than allocations, so headroom costs nothing until a burst arrives.
- The **floor** is the standing reservation the scheduler bin-packs. It is sized from idle footprint, not peak, because over-reserving is what caps tenants per box.

**Who this actually moves.** Every live instance in the governed regions is on an enterprise plan: 10 of them across eu-central, us-east and us-west, with zero on pro or air. The five non-enterprise instances all sit in the private runner-cache pool, which is not governed and keeps the controller's 2Gi/4Gi default. So when the follow-up switches tiering on, the tier that actually lands is the enterprise one, and its floor drops from 2Gi to 1Gi, halving the standing reservation on the governed boxes.

That is a snapshot, not the steady state: non-enterprise instances on the governed boxes are expected, and the smaller profiles are what keep each of them from reserving 2Gi it will not use.

The tiers are set from measured 14-day peak working sets rather than from guesses. The three air instances peak at 149-151 MiB, the two pro ones at 225 MiB and 1,217 MiB. That spread is why pro has its own profile: a single shared tier sized for air would leave a pro tenant shedding constantly, and one sized for pro would hand free accounts several times the memory they use. Each ceiling clears its plan's peak by more than the runtime's `0.9x` recovery hysteresis, so a burst that trips shedding can recover rather than staying shed for its duration. Enterprise is set differently on purpose: those instances expand to fill whatever ceiling they are given, so their usage measures the limit, not demand. Sizing that tier before the tenants arrive is the cheap ordering. Retrofitting it later means resizing live instances on boxes that are already dense, which is the same restart-under-load problem the rest of this stack is arranged to avoid.

That does not by itself raise density. At a 4:1 ceiling-to-floor ratio the ceiling bin-pack overtakes the native `requests.memory` pack: on a 27.4 GiB-allocatable box the floor admits 27 instances and the ceiling budget admits 13, so 13 remains the limit, the same as at a 2Gi floor. The lower floor buys reclaimable headroom and kernel protection sized to real use. Turning that into tenants means raising `MemoryCeilingOversubscription`, which is a separate decision and deliberately not bundled here.

Note that observed usage cannot size the ceiling: for the active pods, observed usage *is* the ceiling, since Kura expands to fill whatever limit it is given. Only the 37 idle pods emit a real demand signal, and it is ~160 MB.

**What the ceiling costs.** Almost everything Kura sizes from its cgroup limit is a semaphore permit count rather than an allocation, so it bounds admitted concurrency and consumes nothing until a burst arrives: the response-stream pool (7.5% of the limit), transient/foreground reservations (25%), mmap serving (<=512 MiB), and the REAPI materialization and response budgets. The genuinely-allocating caches are clamped: the manifest cache is `soft/16` capped at 64 MiB and is already at that cap at 2Gi, while the metadata store's block cache and write-buffer pool are each `limit/32` capped at 128 MiB. So 3Gi to 4Gi adds 64 MiB of steady-state allocation and nothing else, and both caches stop growing exactly at this ceiling. Page cache does grow toward the limit, but it is clean and reclaimable, and the pressure machine excludes it by keying off `pressure_bytes` rather than `memory.current`.

**Ratios, and which constraint binds.** The controller default lands at exactly 2:1 (2Gi/4Gi), which is the point where the `tuist.dev/memory-ceiling-mib` bin-pack stops being slack: below 2:1 the native `requests.memory` pack is tighter and the extended resource never binds, and at 2:1 the two coincide. So raising the default ceiling to 4Gi costs no density against the previous `request == limit` shape.

The per-plan tiers all sit above 2:1 (enterprise 4:1, pro 6:1, air 3:1). Once the follow-up switches them on, the ceiling bin-pack rather than the floor is what decides how many instances fit on a box.

**The metric.** Whether a floor is kernel-enforced or merely a scheduling promise is otherwise unobservable: a request reaches the kernel as cgroup v2 `memory.min`/`memory.low` only under the kubelet's MemoryQoS gate, and no container metric distinguishes the two. That state can also change silently across a kubelet upgrade, so a zero here is alertable rather than discovered during an incident.

**Anon has to fit under the floor.** Kura sized every admission pool from the cgroup limit, transient included, and transient is the budget for response send buffers, upload staging and materialization: anonymous memory. Above the floor that memory is unguaranteed *and* unreclaimable, so node contention resolves it with the OOM killer rather than by shedding. Sizing it from the ceiling therefore traded a retryable `503` for a dead pod, and every tier was over the line, worst on pro at 2.2x its floor.

The floor is invisible inside a container, so the controller now publishes `requests.memory` through the Downward API and Kura bounds the transient pool by it, less the anon it holds outside admission (the metadata store's block cache and write-buffer pool, the manifest cache, and a process baseline). The floor-to-ceiling gap stays available to page cache and mmap-served segments, which the kernel *can* reclaim, so a larger ceiling still buys burst absorption.

Measured demand says the tiers already fit. Peak `kura_memory_transient_reserved_bytes` over 14 days:

| tier | busiest instance | derived budget |
|---|---|---|
| enterprise | 511 MiB | ~630 MiB |
| pro | 77 MiB | ~180 MiB |
| air | 0 | ~105 MiB |

Enterprise is the one to note: 511 MiB against the 512 MiB the 2Gi ceiling gives it today, so it has been saturating the anon pool, which is part of what the ceiling bump fixes. No floor published means no change at all, so self-hosted deployments and a half-rolled fleet keep the ceiling-derived sizing.

### Investigated, deliberately not addressed: admission queue depth

Reading the admission path while writing this up corrected two things about the failure and surfaced one design question worth recording rather than acting on.

Pressure gates exactly one thing, the elastic borrow. Queueing is not pressure-gated: `acquire_response_stream_memory` takes a waiter slot and waits up to 5s, retrying on every pressure transition and on every permit release (the permit's `Drop` notifies waiters, and the notification future is armed before the try, so there is no lost wakeup). Foreground transient reservations are blocked only at Critical, not at the Constrained state this incident actually hit.

So the 503s were `QueueFull` from queue depth, not from queueing being unavailable.

Deepening that queue is the obvious next move and is not worth making yet:

- The depth is one slot per servable stream, a 1x overcommit. Past that, admitted waiters have no path to a permit inside the 5s patience, so `QueueFull` just becomes `Timeout`: the same 503 five seconds later, holding a connection and its send buffer the whole time.
- The binding constraint was capacity, not queue policy. At 4Gi the peak sits well below the soft watermark, so pressure stays Normal and the elastic pool stays available. Pool and queue both scale with the limit anyway, to roughly 307 MiB and 152 slots.
- A fast 503 here is retryable and the CLI already handles it.

The signal that would change this is `QueueFull` recorded while the pool itself still has room, meaning waiters ran out with capacity going unused. The admission metric already separates `queue_full` from `timeout`, so that is visible without new instrumentation.

One note for whoever reaches this next: admission turns are serialized behind a mutex, so only the head waiter attempts acquisition. A head waiter holding out for a large reservation blocks the rest for the remainder of its budget even when smaller requests could be satisfied from what is free. That is deliberate (FIFO fairness, no thundering herd on release) and it is bounded, but it is a more interesting knob than depth if 503s persist after the ceiling bump.

### Deliberately not in this PR

Neither region flag is on. `memory_governed` (per-tier sizing) and `memory_ceiling_bin_packed` (the extended-resource request) both stay off, so every managed instance takes the controller's default 2048/4096 profile. That default is what carries the 4Gi ceiling, so the 503 fix lands in full; only the floor stays where it is.

They are off for different reasons:

- **Ceiling bin-packing** waits on deploy order. The CAPI provider deploys to the management cluster on its own cadence while the server deploys from `main`, so requesting `tuist.dev/memory-ceiling-mib` before a node advertises it would leave every cache pod Pending. The manifest revision bump means pods are recreated onto whatever the CR says rather than drifting there slowly.
- **Per-tier sizing** waits on enforcement. `requests.memory` never reaches the kernel, so a tiered floor is only a scheduling promise: nothing stops a pod that has expanded toward its ceiling from taking memory a quieter neighbour was promised. The kubelet's MemoryQoS gate is what closes that, and it ships in the follow-up with the node reservations it depends on.

The effect is that this PR moves overcommit from none to 2x (2Gi reserved against a 4Gi ceiling) rather than to 3x, and the aggressive tier arrives in the same deploy as its enforcement instead of depending on a tight window between two pipelines. Some overcommit is unavoidable: a burst ceiling above the reservation is overcommit, so there is no version of the 503 fix that keeps request == limit.

### Deploy order

The provider must roll before the server, so the node budget exists before anything can consume it. Nothing in the code enforces that; the flags above are what make getting it wrong harmless rather than an outage. This PR on its own is safe in either order.

### Follow-up

Stacked on this branch: node reservations, eviction margin, MemoryQoS, per-tier sizing, and ceiling bin-packing.

### How to test locally

```
cd infra/kura-controller && go test ./...
cd infra/cluster-api-provider-tuist && go test ./...
cd server && mix test test/tuist/kura/ test/tuist/billing/entitlements_test.exs
cd kura && mise exec -- cargo test --lib metrics
```

`gofmt`, `go vet`, `mix format --check-formatted`, `mix credo --strict`, and `cargo clippy --all-targets -- -D warnings` are clean.

One behaviour change worth noting for review: resolving the plan directly drops the catch-all in `Entitlements.resolve_plan/1` that answered `:air` for anything that is not an `%Account{}`. A malformed account now raises instead of silently taking the standard floor. The only caller relying on that tolerance was a test fixture passing a bare map, which now builds a real `%Account{}`. `memory_plan/1` keeps the one thing the entitlement carried beyond the lookup: off the hosted server there are no subscriptions, so it short-circuits to `:enterprise` rather than resolving every self-hosted account to `:air`.


server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -386,9 +406,17 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
...
       peers_revision_suffix(peer_urls) <>
       mesh_peers_sync_revision_suffix(region, entitlements) <>
-      backfill_revision_suffix(entitlements)
+      backfill_revision_suffix(entitlements) <>
+      memory_revision_suffix(entitlements)
   end
 
+  # Folded in so an account whose plan changes re-applies onto the other profile. Without it the instance would keep the
+  # profile it was created with until some unrelated field happened to change.
+  defp memory_revision_suffix(%{memory: %{floor_mib: floor_mib, ceiling_mib: ceiling_mib}}),
```
esnunes: The revision suffix nicely covers the profile (`+mem<floor>-<ceiling>`), but `memoryCeilingBinPacked` reaches the rendered spec without reaching the revision — it's region config, so `memory_revision_suffix/1` never sees it. Since the reconciler converges on revision equality alone, flipping `memory_ceiling_bin_packed` on a region whose profile didn't also change would alter the rendered manifest with an identical revision, and existing instances would never re-apply — pods wouldn't pick up (or, on rollback, drop) the extended-resource request until some unrelated suffix moved. That's the same class of hole the `mesh_peers_sync_revision_suffix` comment in this file warns about. Worth folding a `"+binpack"` marker into the suffix (or into `memory_revision_suffix/1`) so both flip directions self-apply? The rollback direction seems like the one that could hurt: if the CAPI provider ever has to roll back, turning the flag off is the fix for Pending pods, and it wouldn't propagate.

kura/src/memory/cgroup.rs
```diff
@@ -272,6 +272,35 @@ pub fn container_memory_snapshot() -> Option<ContainerMemorySnapshot> {
...
+/// decide anything; they describe what the kernel will do on Kura's behalf, so
+/// folding them into admission input would confuse two different things.
+pub fn container_memory_protection() -> Option<(u64, u64)> {
+    #[cfg(target_os = "linux")]
+    {
+        // memory.min is absent on cgroup v1 and on a v2 root cgroup; treat an
+        // unreadable file as no protection rather than as missing data, because
+        // "no protection" is exactly what it means for the caller.
+        Some((
+            read_u64_file("/sys/fs/cgroup/memory.min").unwrap_or(0),
```
esnunes: Tiny edge case in the new protection reader: `memory.min`/`memory.low` can hold the cgroup-v2 literal `max` (systemd `MemoryLow=infinity`, an operator override), and `read_u64_file` fails to parse it, so `unwrap_or(0)` would report *zero* protection exactly when protection is maximal — and zero is the alertable state here. The kubelet writes concrete bytes today so this can't fire on the managed fleet, but since `read_required_memory_limit_file` right below already handles the `max` sentinel, reusing it (mapping `max` to the limit or `u64::MAX`) seems like a cheap way to keep the gauge truthful everywhere.

kura/src/app.rs
```diff
@@ -651,6 +651,11 @@ fn spawn_memory_pressure_tasks(state: Arc<AppState>) {
                         sensor_state.memory.observe(snapshot.resident_bytes);
                     }
                 }
+                if let Some((min_bytes, low_bytes)) = crate::memory::container_memory_protection() {
```
esnunes: Definitely not a blocker: the protection read runs in the 200ms pressure-sampling loop — two sysfs reads five times a second for values the kubelet writes once at container creation, feeding a gauge scraped every 15–60s. Periodic sampling is right (a silent kubelet change is what the metric exists to catch), but the cadence feels out of proportion, and it adds a little I/O to the loop the watchdog supervises. The 1Hz metrics loop — or even once a minute — would carry the same signal.

esnunes (approved):
An important point for reflection:

The floor (memory request), translates to the maximum amount of memory we can guarantee Kura will be able to retain without causing an OOM in a busy node, as Kubernetes performs the scheduling of pods based on this value. Therefore, we shouldn't allocate non-reclaimable memory over the floor limit.

It would be great to know how many requests each plan would support based on their floor memory, how many peers (due to replication/bootstrap), etc

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
fortmarek: Confirmed and fixed in 792d4de. `memory_revision_suffix/1` now takes the region and appends `+binpack`, so both flip directions move the revision. You are right that the rollback direction is the one that bites: the combined flip in the stacked PR happens to move the suffix anyway (the profile changes from the controller default at the same time), but turning bin-packing off on its own leaves floor and ceiling identical, so the Pending-pods remedy would silently not propagate. Added a regression test that flips only the flag and asserts the revision moves.

kura/src/memory/cgroup.rs
fortmarek: Fixed in 792d4de — `max` now maps to the container limit (falling back to `u64::MAX` when the limit itself is unreadable) instead of failing to parse into `0`. Agreed this is the worst possible direction to get wrong, since `0` is exactly what the gauge is alerted on. I split the parsing into a pure `protection_bytes/2` helper rather than calling `read_required_memory_limit_file` directly, so it is `#[cfg(any(target_os = "linux", test))]` and unit-tested off Linux like `named_value` and `bracketed_working_set_bytes` in the same module; the test covers `max`, a concrete byte count, and malformed input.

kura/src/app.rs
fortmarek: Fixed in 792d4de. Kept it on the pressure loop but gated behind a 60s deadline (`MEMORY_PROTECTION_SAMPLE_INTERVAL`), so it still catches a silent kubelet change without doing two sysfs reads five times a second inside the loop the watchdog supervises. The deadline starts at zero so the first pass publishes the gauge immediately rather than leaving it unset for a minute.

fortmarek:
Thanks — the reflection point was right, and it turned out to be actionable rather than just a note, so it's implemented in 33d8014 rather than deferred.

### The invariant

Agreed on the reasoning, and Kura was violating it. Every admission pool was sized from the cgroup **limit**, including the transient pool — and transient is precisely the anonymous budget: response send buffers, upload staging, materialization. Anything anonymous above the floor is unguaranteed *and* unreclaimable, so contention resolves it with the OOM killer rather than by shedding. Sizing it from the ceiling traded a retryable `503` for a dead pod. Theoretical anon reach against each floor was 1.4x (enterprise), **2.2x (pro)**, 1.4x (air).

`requests.memory` isn't visible inside a container, so the controller now publishes it as `KURA_MEMORY_FLOOR_BYTES` via the Downward API, and Kura bounds the transient pool by it — minus the anon it holds outside admission (the metadata store's block cache and write-buffer pool, the manifest cache, and a process baseline). The floor-to-ceiling gap stays available to page cache and mmap-served segments, which the kernel *can* reclaim, so a larger ceiling still buys burst absorption. Exactly your "don't allocate non-reclaimable memory over the floor".

No floor published means no change, so self-hosted deployments and mixed-version pods mid-rollout keep the ceiling-derived sizing.

### Capacity per plan

Per instance (each account runs two replicas), derived from the constants rather than estimated:

| plan | floor / ceiling | anon budget | concurrent streams | bound by | degraded slots | replication reserve |
|---|---|---|---|---|---|---|
| enterprise | 1Gi / 4Gi | 608 MiB | 148 | stream pool | 614 | 6 MiB (1% of pool) |
| pro | 512Mi / 3Gi | 160 MiB | 78 | **floor** | 460 | 6 MiB (2%) |
| air | 256Mi / 768Mi | 83 MiB | 25 | stream pool | 115 | 6 MiB (10%) |

Two things worth pulling out:

**Pro is the tier the floor actually binds on.** Enterprise and air are limited by the ceiling-derived stream pool; pro is limited by its floor, which drops it from 110 concurrent streams to 78. That's the invariant doing real work rather than being a no-op, and it's the tier where a floor bump would buy the most if we decide 78 is thin.

**Replication is a flat 6 MiB on every tier** (`MAX_INLINE_REPLICATION_BODY_BYTES` + four chunks), carved out of the stream pool before foreground gets any. Fine in absolute terms, but that's 1% of enterprise's pool and 10% of air's — worth knowing before air's ceiling drops any further.

### What the measurements said

`kura_memory_transient_reserved_bytes`, 14-day peak, is the anon budget actually consumed:

| tier | busiest instance | new budget |
|---|---|---|
| enterprise | **510.9 MiB** | 608 MiB |
| pro (`locket`) | 77 MiB | 160 MiB |
| air | 0 | 83 MiB |

All three clear their observed peak, so the floors hold as they are — I'd expected this to force them back up.

The enterprise number is the interesting one: transient capacity at today's 2Gi ceiling is exactly 512 MiB, and the busiest instance peaked at 510.9. It has been **saturating the anon pool**, which is independent corroboration of the ceiling bump from a different metric than the pressure-state evidence in the description.

PR #12264, opened 2026-08-11
perf(kura,server,cli): let cache nodes authorize without calling the server

fortmarek:
Cache authentication was costing a server round-trip on nearly every request, and the round-trip itself was slow. This makes the hot path avoid the server entirely, and makes the server-side work cheap when it is still needed.

## The problem

Cache nodes can already authorize from a Tuist JWT without calling back, by reading the cache grants embedded in it. In practice that path never fired: over 24 hours production ran 8,963 authenticate hooks and made 8,967 introspection calls, a ratio of one to one.

Two separate causes, and it is worth being precise about which one dominated.

**The credential CI uses is opaque.** `TUIST_TOKEN` resolves to a `tuist_<uuid>_<hash>` string that a cache node has nothing to verify, so it must ask the server about every authorization that misses its local cache. Embedding grants in the token is not possible, and widening the claims embedded in JWTs does not scale either: grants are `account/project` strings, so a large account would need tens of kilobytes of claims on every request. That is why tokens embed only the five most recently used projects today, and why that guess usually misses.

Note that moving customers to account tokens does not help here. Both project and account tokens are the same opaque shape, both verified with Bcrypt, and neither can be verified by a cache node.

**Introspection itself scaled with account size.** Resolving cache grants walked every accessible project and read state back from the database for each one, so authorization got slower as an account grew and the largest accounts were the slowest to authorize.

## What changed

**Kura stops running its policy in Lua.** Authorization moved out of a scripted extension point and into the binary, so nothing serializes behind an interpreter any more. Concurrent evaluations for the same token are coalesced into one in-flight request. This is described in full under [Removing the scripting layer](#removing-the-scripting-layer).

**Resolving cache grants no longer issues a query per project.** Three causes: each of the four grant buckets resolved the accessible account and project lists independently; deciding whether a subject belongs to a project's account read that account and organization back per project even though the project list already had them loaded; and membership itself was two `exists?` queries per check. Users now carry their organization roles in a virtual field resolved once, following the same per-request pattern the schema already uses for `operator_grant`. The field is nil unless a caller resolves it and the checks fall back to querying, so callers that have not opted in behave exactly as before.

**The server can mint a cache token.** `POST /api/cache/token` returns a short-lived token carrying the caller's cache grants, minted for whatever credential they already authenticated with. Cache nodes verify it with the Guardian verifier they are already configured with, so no cache-side change was needed. It takes an optional project handle and narrows the grants to it. An account-wide credential reaches every project its account owns, so an unscoped token hands a cache node everything that credential can reach, and the claim grows with the account: handles average 23 characters and appear in both the read and write buckets, so each project costs about seventy bytes once base64 encoded.

For almost everyone this is irrelevant. Of the 2913 accounts with a project, 94% have one or two, p90 is two, and p99 is five, which is a claim of a few hundred bytes. The tail is thin and long: twelve accounts have ten or more, and the largest has 107, whose unscoped grants come to roughly 7KB before the rest of the token, comfortably past what proxies allow in a header. So the size argument justifies scoping for a handful of accounts rather than for the product, and the reason it is worth doing everywhere is the first one: the token reaches a cache node carrying whatever it grants, and narrowing it to the project being built limits what a leaked one opens.

The scope only ever filters, never adds.

**The module cache takes the exchange too.** Scoping only the CAS and key-value services would have left most cache traffic authorizing the old way, since the split of what reaches nodes is roughly half Xcode CAS over HTTP, a quarter REAPI over gRPC, and a fifth module cache over HTTP. The six module cache services build the same cache client and already hold both handles; they just never passed a project through, so the middleware skipped the exchange. That brings the covered share to roughly seventy percent.

REAPI is deliberately left out. The obvious change there does not work: the proxy is machine-wide and multiplexes every project on the box by REAPI instance behind one shared bearer, so it cannot hold a token scoped to a single project, and its scope is not knowable at startup because the instance arrives per build through the `tuist-instance` plugin option. Its `current()` also runs on every request, so a failed exchange has to be cached rather than retried. Keying acquisition by resolved instance is a change to the proxy's token path rather than an addition to it, and belongs in its own change.

The token is scoped to cache access only. Guardian refuses to resolve a subject from one, so it cannot double as an API credential: its `sub` is a project id, which the general claim path would otherwise read as a user id.

**The CLI exchanges once and reuses.** `CacheTokenStore` holds the exchanged token until shortly before expiry. Concurrent callers on a cold store share one in-flight exchange rather than each starting their own, since a build issues many cache requests at once. A failed exchange yields nil rather than throwing, so a server that cannot mint one, such as an older self-hosted deployment, leaves cache access working exactly as before. Only callers that name the project they are caching for take this path, which is the CAS and key-value services; clients built without a handle keep sending the credential they always did.

## Results

**The broad win is that CI traffic stops reaching introspection at all.** A CI credential is opaque, so a cache node could never verify one itself: every authorization missing the node's local cache cost a round-trip here, and that round-trip is dominated by Bcrypt rather than by any query. Measured across cost factors, the doubling per level confirming the curve: cost 4 is 1.2 ms, cost 8 is 16 ms, cost 10 is 64 ms, and cost 12, the production default, is 223 ms. That is roughly 70% of the 317 ms mean introspection latency. The test suite pins `log_rounds: 1`, so none of it shows up in test timings.

The exchange replaces that with local HMAC verification on the node, and nothing about it scales with account size. 85% of runs over the last 30 days are CI, so this is the path most traffic was already on, and with the module cache clients also exchanging it covers roughly seventy percent of what reaches a cache node. The REAPI quarter still authorizes the old way.

**The N+1 fix is the tail.** Wall-clock for one `cache_grants/2` call, median of 20 runs after warmup:

| projects | before | after |
| --- | --- | --- |
| 2 | ~14 ms | 2.0 ms |
| 12 | 12.9 ms | 2.0 ms |
| 50 | 80.5 ms | 2.1 ms |

Query counts over the same range: 52 to 20, 172 to 20, 628 to 20. Flat in the number of projects. Small-N baselines are noisy, with outliers up to 1.1 s, so read 2 and 12 as roughly 13 ms. Worth reading against the distribution: p90 is two projects, so the 50-project row describes one account rather than a representative one. Still worth having, since that account exists and the cost grew without a bound.

**Removing the hook bottleneck is the third piece.** The single Lua state serialized every hook invocation, so one slow introspection stalled authorization for unrelated requests regardless of their account size or credential type. That ceiling applied to everyone and is independent of the other two.

## How to test locally

```
cd server && mix test test/tuist/cache_test.exs test/tuist_web/controllers/api/cache_controller_test.exs
cd kura && cargo test --lib
xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistCacheTests/CacheTokenStoreTests -only-testing TuistCacheTests/CacheClientAuthenticationMiddlewareTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""
```

## Notes for the reviewer

**The claim shape is asserted twice, not once.** The server test asserts what it puts in the claims and the extension test asserts what it reads back out of a token it signs itself, so nothing fails if the two drift apart: a rename on one side, with that side's test updated, passes both suites and breaks in production. A cross-language test covering this existed on an earlier revision of the branch and was dropped, since running it needs Postgres, ClickHouse, an Elixir compile and a Rust compile in one job to carry a single assertion. Worth knowing the gap is open rather than closed.

**The regeneration commit carries unrelated churn.** `chore(cli): regenerate the OpenAPI client` is split out on purpose. The generator reordered path parameters for `listBundles`, `createBundle` and `getBundle`, which the new endpoint did not touch: 55 call sites pass `account_handle` first and only those three stopped compiling, so the order varies between runs. The call sites are adapted rather than hand-editing generated code. Worth fixing at the source, since a deterministic spec export would stop this recurring on every regeneration.

**The kura suite is flaky, independently of this change.** `retries_introspection_transport_failures_once` fails roughly 1 in 3 runs in isolation. Measured before and after: 4 of 4 full-suite runs failed on the parent commit, versus 1 of 3 passing after. So this change does not introduce it and may marginally reduce it, but do not read a single green run as stable. Those tests appear to have timing and global `set_var` races and deserve a separate look.

**Still on the introspection path:** `cas-plugin` is untouched and authenticates the old way, so REAPI traffic still pays the Bcrypt cost per authorization. Giving it the exchange needs the per-instance token work described above.

## Removing the scripting layer

Kura's cache policy ran as a Lua hook loaded through a general extension point. Nothing outside Tuist ever wrote one, so the generality cost without buying: a pool of interpreters existed only because a Lua state is single-threaded and hooks perform network I/O, every request was marshalled into Lua values and the result marshalled back, and the claim shape was agreed by convention between two languages that could not check each other. The policy is now Rust (`kura/src/auth/`), and `mlua` is gone.

What went with it: the interpreter pool, the HMAC signer nothing called, and the response-headers hook no script implemented. That last one is why the principal is no longer threaded through the transports — it was only ever carried to feed that hook.

Fail-closed becomes structural rather than a flag. The policy answers 503 when it cannot reach the backend, so there is no hook failure left for a flag to interpret, and the knobs that only configured the script are gone with it. Notably the cache TTLs never applied: they were the fallback for a hook that always returned its own, so the effective values were the policy's 60 and 3 seconds, which is what it now returns. Treating the documented 600 as authoritative would have decupled how long a revoked token keeps working.

The hook's tests came over intact and still drive the same outcomes through a stand-in for Tuist's server, which is what says the behaviour survived. They no longer set process environment variables to build an engine, so they no longer share state with the rest of the binary. Two changed: the retry test failed a request by racing a timeout and flaked under load, so it now fails one by dropping the connection; and the gRPC namespace tests scripted a fake policy inline and now mint a real token, since the policy is no longer theirs to define.

Metrics move from `kura_extension_*` to `kura_auth_*`, and the per-client label goes — there is one backend now. No dashboard in this repo reads them.

### Caching decisions

Both stages are cached per credentials and concurrent misses on the same key are coalesced, so a build issuing a hundred requests at once makes one call to the server rather than a hundred. That is [`moka`](https://docs.rs/moka)'s `get_with` rather than something hand-rolled: the cache, the single-flight, the expiry sweep and the capacity check were about 115 lines that a maintained crate already does.

Per-entry expiry carries over through moka's `Expiry` trait, so a grant still lives 60 seconds and a denial 3, decided by the decision rather than by the cache.

It also fixes what capacity did. At the cap with nothing expired, the old path stopped caching entirely and warned once per request — a full cache degraded into no cache plus log spam. moka evicts instead. That behaviour predates this change; it came over from the extension engine.

One deliberate check: moka documents that it recovers when the caller running the init future *panics*, and says nothing about that caller being **cancelled**. These evaluations are request-scoped, so a client that disconnects drops the one every other caller is waiting on, which is what the old `FlightGuard` drop handled on purpose. It does recover, and the test proving it is kept rather than deleted as scaffolding, because the crate does not promise it and the failure it would catch is requests hanging on a disconnected client's key.

The cost is the `coalesced` metric label. `is_fresh()` says whether this call ran the evaluation, but a waiter and a plain cache hit look alike, so both count as `hit` now. The surviving labels still answer whether a request cost a backend call.

Ten new crates — moka plus nine small pure-Rust transitives — with no system requirements, so the release image is unchanged.

### What is deliberately left behind

The hook file, the server's `kura_hook_path`, the controller's `ExtensionScript` handling and the CRD property all stay, and each configuration setting still falls back to the name the extension point used.

A Kura node given no authorization configuration does not authorize at all — it serves the cache to anything that can reach it. The reconciler re-applies a drifted manifest with the instance's **current** image tag, and most server deploys never bump that tag, so removing any of this now would strip the configuration from nodes still running a release that needs it. It comes out in #12297, which is gated on every instance having converged to a release with native authorization.

### Not verified here

The end-to-end suite has not been run; it needs `docker compose build`. Its Kura shard was rewritten, since it previously covered the extension mechanism through a custom script that authorized by namespace and signed responses. It now covers the policy itself and needs no stand-in server, because a token Kura can verify carries the grants that answer the request. What that costs is the 403 the old suite asserted for a token scoped to another project: without a server to ask, a token that cannot settle a request yields 503, which the suite now states as the behaviour it is.


esnunes:
### Self-hosted nodes joined to a managed mesh will reject the exchanged cache token

The node half of self-hosting is preserved here — a self-hosted node has no verifier secret by design, authorizes through tenant-scoped introspection, and the ported tests still cover that (`kura/src/auth/policy_tests.rs:206`, `:217`, `:242`). The problem is on the client half: the CLI decides to exchange from the project handle alone, never from where the cache request is going.

`CacheClientAuthenticationMiddleware` swaps in the cache token whenever `fullHandle` is set (`cli/Sources/TuistCache/Client/CacheClientAuthenticationMiddleware.swift:48`). `Client.authenticated` takes `cacheURL` and `authenticationURL` as separate parameters precisely because the node can be somewhere other than the server, but the middleware never reads `cacheURL`. So an account that authenticates against the managed server and caches to its own node sends that node a token it cannot read.

What the node then does with it:

1. `verify_token` returns `None` — no verifier is configured, which is the intended self-hosted shape (`kura/src/auth/tuist.rs:102`, and the rationale in `server/lib/tuist/kura/self_hosted_clients.ex:5`).
2. It forwards the token to introspection, as designed (`kura/src/auth/policy.rs:157`).
3. `Introspection.token_response/2` calls `Authentication.authenticated_subject/1`, which returns nil: `resource_from_claims` matches `%{"typ" => "cache"}` and refuses (`server/lib/tuist/guardian.ex:42`), and none of the three remaining lookups match a JWT string. The response is `%{active: false}`.
4. The node falls back to the legacy `cache_access` route with the same token (`kura/src/auth/policy.rs:235`). That route sits behind `authenticated_api`, so it answers 401.
5. The node denies with 401 (`kura/src/auth/policy.rs:257`).

The CLI does not recover. `RetryMiddleware` retries 408/429/5xx only, and `CacheTokenStore`'s unavailability suppression covers a failed *exchange* — the exchange succeeded here, the node rejected its result — so the store keeps handing out the same rejected token until it expires.

Worth being explicit that the Guardian refusal in step 3 is correct and should stay. `resource_from_claims` is the shared front door for every credential the server accepts, so resolving a cache token there would make it valid on every API endpoint, and the `sub` collision is the sharper edge: a cache token minted for a CI project token carries a project id as `sub`, which the fallback clause would read as a user id and resolve to an unrelated user in an unrelated account. The collision between that necessary refusal and the introspection fallback is what makes this non-trivial rather than a missing case.

Affected traffic is everything that passes a handle — CAS, key-value and the six module cache services. REAPI is unaffected, since it never exchanges.

Flagging the confidence level: this is read from the code, not reproduced against a self-hosted node.

fortmarek:
> Self-hosted nodes joined to a managed mesh will reject the exchanged cache token

Confirmed and fixed in bfd2f9ce20. I walked each link — the middleware never reads `cacheURL`, `verify_token` returns `None`, `resource_from_claims` refuses `typ: cache`, `authenticated_subject` falls through to nil, and the node retries `cache_access` and gets a 401 — and they all hold.

Fixed on the server rather than the client: tenant-scoped `Introspection.token_response/2` now verifies a cache token itself and answers from the grants it was minted with, scoped to the account like every other response on that path.

Guardian's refusal is untouched, and a test pins it. Your point about `sub` being a project id that the fallback clause would read as a user id is exactly why the answer belongs in introspection rather than in `resource_from_claims` — the question there is only ever what a token may reach in the cache. `principal_kind` is omitted, since the claims do not record what the token was minted for and a node reads its absence as an unnamed subject.

I took this over the client-side options because the CLI cannot distinguish a managed node from a self-hosted one by URL alone — managed nodes also sit on a different host from the server — so gating the exchange would have needed a new per-endpoint signal, with a rollout ordering problem of its own.

Four tests. The one that matters fails without the change and passes with it, which I verified by reverting the fix and re-running.

kura/src/auth/config.rs
```diff
@@ -0,0 +1,231 @@
...
+    pub connect_timeout: Duration,
+    pub request_timeout: Duration,
+    pub verifier: Option<JwtVerifier>,
+    pub introspection: Option<IntrospectionCredentials>,
+    pub cache_max_entries: usize,
+}
+
+impl AuthConfig {
+    pub fn from_env() -> Result<Option<Self>, String> {
+        let enabled = env_truthy(ENABLED).unwrap_or(false) || env_value(TUIST_URL).is_some();
```
esnunes: Small one. `enabled = env_truthy(ENABLED).unwrap_or(false) || env_value(TUIST_URL).is_some()` means an explicit `KURA_AUTH_ENABLED=0` loses to the URL being set.

The `||` mirrors the old rule exactly, so the behaviour is not new. What is new is that this PR documents `KURA_AUTH_ENABLED` as a `No | Disabled` setting in the self-hosting table, and that the managed chart now always sets `KURA_AUTH_TUIST_URL` — so an operator who reads the table and sets `0` to take a node out of the auth path gets authorization anyway, with nothing in the logs saying the setting was overridden.

Failing safe is clearly the right default; it is the documented-knob-that-is-not-a-knob I would tidy — either honour an explicit `false`, or say in the table that the URL alone enables it.

server/lib/tuist/cache.ex
```diff
@@ -25,31 +28,104 @@ defmodule Tuist.Cache do
...
+
+  # Narrows the grants to the one project the caller is about to use. An
+  # account-wide credential reaches every project its account owns, so an
+  # unscoped token hands a cache node everything that credential can reach and
+  # grows with the account at roughly seventy bytes a project. Nearly every
+  # account has one or two, where none of this matters; the largest has enough
+  # for a claim in the thousands of bytes. This bounds both regardless.
+  defp scope_grants(grants, nil), do: grants
+
+  defp scope_grants(grants, scope) do
```
esnunes: `scope_grants/2` matches nil for "unscoped" and treats anything else as a real scope, so `?full_handle=` — an unset shell variable interpolated into a URL, for instance — reaches the second clause: account buckets emptied, `only(handles, "")` matching nothing. The endpoint answers 200 with a well-formed token that grants zero access, and the caller is then denied on every cache request with nothing pointing back here. A handle the caller cannot reach behaves the same way.

Treating a blank string as absent, or rejecting a `full_handle` that filters down to nothing, would turn a confusing runtime failure into an immediate one.

esnunes (approved):
Approving.

This is a genuinely good change and an unusually honest description of one — separating the three wins, saying which dominates, and naming what is left on the table (REAPI, the cross-language claim assertion, the flaky Kura test) makes it much easier to review. The Lua → Rust port reads as faithful: I went through the two side by side and the target resolution, the action rules, write-implies-read, the introspection → cache-access fallback and the "no `scopes` claim means a pre-grants token" rule all carry over with the same ordering and status codes. Deleting 2,513 lines of interpreter plumbing for a policy nobody outside Tuist ever wrote is clearly right. I ran the Kura auth unit tests locally: 57 pass, including the retry test the description flags as flaky.

The scoping story also holds up where it matters — `only/2` filters a list the caller was already granted rather than adding to it, the test pinning that is the right test to have, and the Guardian clause refusing to resolve a subject from a `typ: "cache"` token is correctly placed first among its clauses.

I started reading at 1e2c305, and `fix(server): answer a cache token from the grants it carries` landed while I was in there, so I re-read against bfd2f9c. That commit closes the self-hosted path: `token_response/2` now verifies the token itself and answers from the grants it carries, constrained to the node's account, so a node holding no verifier secret gets a usable answer instead of `active: false`. Leaving the Guardian refusal intact and pinning it with "a cache token still resolves to no API subject" is the right shape for it.

Three small things left — two inline, and one here because it does not sit on a changed line.

**`kura/src/http.rs:1223`** — Now that the response-headers hook is gone, I think the body buffering in `authorize_request` no longer buys anything.

Every `PUT /api/cache/keyvalue` without a `cas_id` query param still has its whole body read via `to_bytes(body, max_keyvalue_bytes)`, copied into a Vec and re-wrapped, inside the auth middleware and before admission control. Its only consumer is `artifact_key` in `http_request_metadata`, and the new policy reads none of it — `request_target` takes tenant and namespace from the path and query, `request_action` takes the operation and method, and the authorization cache key covers route/method/operation/tenant/namespace/producer.

On a memory-sensitive node that is a real allocation on the write path. One caveat if you go for it: the read is also what returns 413 above `max_keyvalue_bytes`, so it is worth checking the handler still enforces that before deleting it.

kura/src/auth/config.rs
fortmarek: Fixed in 8b623c0954, taking your second option rather than the first. The rule stays — a node told which server to authorize against should not stay open — but it is no longer silent: an explicit `false` alongside a set URL now logs that authorization stays on and how to actually turn it off, and both the self-hosting table and the README say the URL enables it on its own. Honouring the `false` was tempting, but the managed chart sets `KURA_AUTH_TUIST_URL` on every instance, so it would put a one-line lever to serve the cache unauthenticated within reach of a values change.

server/lib/tuist/cache.ex
fortmarek: Fixed in 8b623c0954: a blank or whitespace-only scope is now treated as no scope, so `?full_handle=` mints the unscoped token rather than one filtered down to nothing. Two tests, both verified to fail without the change.

I did not take the second half — rejecting a `full_handle` the caller cannot reach, which does behave identically. That needs a declared error response on the operation and a regenerated CLI client, and I would rather do that deliberately than fold it into this PR; happy to open it as a follow-up if you would like it tracked.

PR #12266, opened 2026-08-11
feat(kura): reserve node headroom and make the memory floor kernel-enforced

fortmarek:
> Stacked on #12258. Gives the Linux bare-metal nodes real daemon headroom and a managed-eviction margin, then tiers instance memory floors and makes them kernel-enforced rather than advisory, and turns on the ceiling bin-pack now that the node budget it packs against exists. The tiering and its enforcement are deliberately in one PR so they reach production in one deploy.
>
> **Base branch is `claude/kura-5xx-errors-b353f4`, not `main`.** Review the diff against that.

### What changed

Three settings on the self-join kubelet config, plus the two server-side flags they unblock:

1. **`systemReserved` / `kubeReserved`** (500m + 1Gi each) carve the daemons that live outside any pod (kubelet, containerd, systemd, kernel) out of allocatable.
2. **`evictionHard`** raises `memory.available` to 5%. The map replaces the kubelet's defaults rather than merging with them (an omitted signal is set to 0, not defaulted), so all five default signals are restated and only `memory.available` moves. `imagefs.inodesFree` matters here specifically because the bootstrap points containerd's root at `/data` wherever `/data` is a separate mount, making imagefs a distinct filesystem that the `nodefs` signals on `/` do not cover.
3. **`MemoryQoS`** with `memoryThrottlingFactor: 1.0`.
4. **`memory_ceiling_bin_packed`** flips on, so managed instances request their ceiling as `tuist.dev/memory-ceiling-mib`.
5. **`memory_governed`** flips on, so managed instances are sized per tier. Enterprise drops from the controller's 2Gi default to a 1Gi floor, which is the change that reaches the current fleet: all 10 governed-region instances are on enterprise plans.

### Why

**The nodes have no headroom today.** Capacity 32463340Ki against allocatable 32360940Ki, a margin of exactly 100 MiB on a 31GiB box, the kubelet default, with zero daemon reservation. That is thin enough that the kernel OOM killer usually beats the eviction manager, so memory contention arrives as a `SIGKILL` mid-transfer rather than an evicted pod with an event. For a Kura primary that also means an unplanned role handoff and its read-after-write inconsistency window, plus client-visible connection resets instead of the retryable `503`s the CLI already handles.

**Why per-tier sizing lands here and not in #12258.** Tiering is what opens the gap between what a pod reserves and what it may take: an air instance drops to a 256Mi floor against a 768Mi ceiling, and a pro one to 512Mi against 3Gi. `requests.memory` never reaches the kernel, so until MemoryQoS maps it onto cgroup v2 `memory.min` that floor is only a scheduling promise, and nothing stops a bursting pod from taking memory a quieter neighbour was promised. Landing it in #12258 would leave the fleet overcommitted by as much as six to one per instance, unbounded (bin-packing still off), and unenforced, for however long the server and provider deploy cadences drift apart. Here they are one deploy.

All 10 governed-region instances are on enterprise plans, so what lands in practice is the 1Gi enterprise floor, halving the standing reservation per instance. The pro (512Mi/3Gi) and air (256Mi/768Mi) profiles have no governed instances on them yet, since the only pro and air instances today sit in the ungoverned runner-cache pool, but they are the sizing that makes non-enterprise tenants affordable on a shared box once they land there, and putting them in place ahead of the tenants avoids resizing live instances on an already-dense box. Note that neither raises density on its own, because the ceiling bin-pack binds before the floor pack does at these ratios.

**The order matters.** MemoryQoS derives its protection from pod *requests*, so with allocatable equal to capacity the scheduler can promise pods the whole box and the protection would then squeeze the kubelet itself. Reservations first is what makes the denominator honest. Equally, the floors have to tier before MemoryQoS reads them, which is why both are in this PR: `memory.min` protects clean page cache too, so deriving it from the old uniform 2Gi reservations would turn most of a warm box's reclaimable artifact cache into unreclaimable memory. Small floors are the precondition, not a refinement. (`memoryThrottlingFactor: 1.0` does not help here; it governs `memory.high` throttling, not `memory.min` reclaim protection.)

**`memoryThrottlingFactor` is pinned to 1.0**, which sets `memory.high` to the limit and disables early throttling, keeping only the `memory.min`/`memory.low` protection. At the 0.9 default, `memory.high` lands at `requests + 0.9*(limits-requests)` and triggers on `memory.current`, which counts clean page cache. A warm Kura node holds clean artifact pages at its hard watermark as normal steady state, and its own shedding keys off `pressure_bytes` precisely to exclude them, so the default would have the kernel throttle a healthy cache node into reclaiming the cache it exists to hold, fighting the runtime's own admission control.

**Why enable an alpha feature now rather than later.** Its value is protection under contention, and no box has exceeded 29% of RAM in 14 days, so it earns nothing measurable yet. That is the argument for doing it now, not against: a mechanism that only activates under contention should be introduced before contention exists, so the first time it matters it already has months of production behind it. Waiting until the boxes are dense means introducing a new variable exactly when the system is stressed, and the rollback below is far cheaper on quiet boxes.

`kura_memory_protection_{min,low}_bytes` (added in #12258) is what makes that soak observable rather than assumed.

### Version dependency

The Linux fleet nodes run **v1.34.9**, where MemoryQoS sets `memory.min` from requests for Burstable containers too: hard, unreclaimable protection. That is safe only because the reservations land first: `memory.min` is memory the kernel may not reclaim, so it OOMs rather than reclaims once the protected total approaches capacity, and `sum(requests) <= allocatable` with allocatable properly reserved is what keeps it clear of the box.

**Upgrading to v1.36+ silently disables this.** v1.36 splits protection behind a `memoryReservationPolicy` field defaulting to `None`, and retiers it so Burstable pods get the softer `memory.low`. With this config unchanged the result is no protection and no throttling either, a silent no-op, not a visible failure. The field cannot be set pre-emptively: an unknown key fails KubeletConfiguration's strict decode and the kubelet will not start. The warning sits on `linuxCloudInitOptions.K8sMinor`, which is what a version bump actually edits.

### Rollout and rollback

`desiredKubeletConfigHash` fingerprints the rendered config, so this re-pushes through the drift loop to every already-Ready node and restarts its kubelet. No machine roll, and running pods survive, since the re-push never touches containerd, apt, or the `/data` mounts. It lands on all three Linux fleets at once, so prefer a low-traffic window and not the 05:07 UTC nightly window.

Rollback is not symmetric: turning `MemoryQoS` back off does not reliably clear the `memory.min`/`memory.low` values already written to existing cgroups ([kubernetes/kubernetes#138436](https://github.com/kubernetes/kubernetes/issues/138436)), so a node may need a kubelet restart or reboot to shed them.

### Deploy order

**No longer a runbook step.** As of the gate in #12258, the kura-controller resolves `memoryCeilingBinPacked` against the nodes behind the instance's own `nodeSelector` and omits the extended-resource request until one actually advertises `tuist.dev/memory-ceiling-mib`. Its reconcile loop already runs every 30s, so it converges either way on its own: the request appears once the provider rolls, and disappears if nodes stop advertising, which is what lets a provider rollback un-wedge Pending pods without a server deploy.

That matters because no node advertises the budget right now (`kubectl get nodes -o custom-columns=...` shows `<none>` across all three pools while `tuist.dev/egress-mbps` is populated), so the provider has not yet rolled. Before the gate, merging this ahead of the provider would have left every managed cache pod Pending.

Provider-first is still the intended order, and still worth following so bin-packing takes effect promptly rather than on the next reconcile after the provider lands. It is no longer load-bearing for availability.

Floors tier in this PR rather than in #12258, so there is no intermediate fleet state to wait out or keep short. #12258 can sit in production as long as it needs to.

### How to test locally

```
cd infra/cluster-api-provider-tuist && go test ./...
cd server && mix test test/tuist/kura/
```

`gofmt`, `go vet`, `mix format --check-formatted`, and `mix credo --strict` are clean.

### Suggested follow-up

An alert on `kura_memory_protection_min_bytes == 0` for managed instances. The metric only helps if something watches it, and a silent drop to zero is exactly the v1.36 failure mode above.


server/lib/tuist/kura/regions.ex
```diff
@@ -514,6 +514,17 @@ defmodule Tuist.Kura.Regions do
...
+        # Packing density is what constrains the shared bare-metal boxes, so
+        # their instances are sized per tier rather than taking the controller
+        # default, and their ceilings are bin-packed against the node budget the
+        # CAPI provider advertises. Per-tier sizing rides with the kubelet's
+        # MemoryQoS gate rather than landing ahead of it: a tiered floor sits
+        # far below its ceiling, so it is only a scheduling promise until the
+        # kernel enforces it as memory.min. The private runner-cache pool runs
+        # on Elastic Metal, which the provider does not patch, so it takes the
+        # controller default and stays off the bin-pack.
+        memory_governed: true,
```
esnunes, line 526: The mechanics here look right, and folding the memory profile into the manifest revision means the flip propagates within one reconcile pass of the server deploy — which is what prompted my one real question: the server deploys with the main cascade, while the CAPI provider that advertises `tuist.dev/memory-ceiling-mib` (and pushes the MemoryQoS kubelet config) rolls to the management cluster on its own cadence. If the server lands first, the re-applied manifests have every managed cache pod requesting an extended resource its node doesn't advertise yet, which leaves them Pending — the scenario the removed test documented as the reason these flags stayed off. Is provider-first ordering already guaranteed (or has the provider already rolled with the advertisement)? If it's a runbook step, it might be worth a line in the PR description so the merge doesn't cascade ahead of it.

esnunes (approved):
This is a carefully built change — the AGENTS.md rewrite is genuinely good operator documentation, the three kubelet settings are ordered for the right reasons, and parsing the rendered config in a test is exactly the right guard given that a malformed push stops kubelets on all three fleets at once. The one thing I'd want settled before this cascades is deploy sequencing: the server half re-applies every managed manifest within a reconcile pass of its deploy (the memory revision suffix makes propagation immediate), while the CAPI provider that advertises the node budget and pushes the MemoryQoS kubelet config rolls on its own cadence. Server-first would leave managed cache pods Pending on an extended resource no node advertises yet. If provider-first ordering is already guaranteed, the sequencing question resolves; a line in the PR description would make that explicit.

server/lib/tuist/kura/regions.ex
fortmarek: Good catch, and the answer to "has the provider already rolled" is no: `tuist.dev/memory-ceiling-mib` is `<none>` on all three pools right now, while `tuist.dev/egress-mbps` is populated. So ordering was load-bearing and enforced only by someone remembering it, which for a governed region means every cache pod Pending if the cascade got there first.

Rather than document it, I made it not matter. The kura-controller now resolves `memoryCeilingBinPacked` against the nodes behind the instance's own `nodeSelector` and omits the request until one actually advertises the budget (`ceilingBudgetAdvertised`, in 12258). It reads `allocatable` rather than `capacity` since that is what the scheduler fits against, and the kubelet mirrors externally patched extended-resource capacity into it, which I confirmed against the live egress resource.

The reconcile loop already runs every 30s, so it converges both ways without intervention: the request appears once the provider rolls, and disappears if nodes stop advertising. That second direction also closes the rollback gap you raised on the earlier review, since turning the region flag off is no longer the only way back from Pending pods.

Provider-first is still the intended order and I have kept it in the deploy-order section, now framed as promptness rather than availability. Test covers all four cases: no budget, budget present, budget only on another pool, and the spec flag off short-circuiting before the List.

PR #12270, opened 2026-08-11
fix(cli): exclude test-plan-skipped targets from selective testing

esnunes:
## What changed

`TestService.testActionTargetReferences` now filters out testables marked `isSkipped`, across all four branches it resolves targets from (named test plan, all plans when building, default plan, and a scheme's plain test action targets). The predicate is extracted as `Sequence<TestableTarget>.enabled` and shared with `GraphTraverser.filterIncludedTargets`.

## Why

A test target disabled in a test plan (`"enabled": false` in an `.xctestplan`, or `isSkipped: true` on a `.testableTarget` in a manifest) was still counted as a selective-testing candidate, even though xcodebuild will never run it.

That created a self-perpetuating phantom. Successful-test hashes are only stored after a target actually runs, so a target that can never run can never become a cache hit — it is reported as an un-cached miss on **every** run, forever. Users see a module the dashboard says needs testing, which then silently never runs.

## Root cause

Three code paths answer "which test targets are in this plan", and they disagreed:

| Path | Honors `isSkipped`? |
|---|---|
| `BuildGraphInspector.isIncluded` (what gets built/run) | yes |
| `GraphTraverser.filterIncludedTargets` (test-plan graph filter) | yes |
| `TestService.testActionTargetReferences` (what gets reported) | **no** |

Only the reporting path was out of step. It feeds `initialTestTargets` → `updateTestServiceAnalytics` → `RunMetadataStorage.selectiveTestingCacheItems`, which is what the server stores and the UI renders.

The worst case is when a skipped target is the *only* non-hit in a plan. Every other target prunes as a cache hit, tree-shaking then removes the now-empty scheme, and the run takes the "test action has no tests to run, finishing early" path. For a sharding build that emits a 0-shard matrix, so CI fans out to zero test jobs and the pipeline passes green having executed nothing.

## Behavior beyond analytics — deliberate, not incidental

`testActionTargetReferences` is not analytics-only, so filtering inside it changes runtime behavior for projects with skipped testables. These are intended; each is the same principle applied consistently — a target xcodebuild will never run should not influence the run:

- **Platform inference.** `resolvePlatform` intersects supported platforms across build-action targets plus these test targets, and needs exactly one survivor. A skipped target no longer narrows that intersection, so a project that relied on a disabled target to disambiguate will now get `unspecifiedPlatform` instead of an inferred one. Inferring a platform from a target that never runs was wrong, and the error names the fix (pass `--platform`).
- **Scheme splitting.** Mixed hosted/hostless detection, the workspace target set, and the single-target eligibility check all see the filtered list. Pinned by a new test: when the only hostless testable is skipped, the scheme is no longer "mixed" and runs whole rather than split per target — one invocation instead of two.
- **All-skipped plans.** A plan whose every testable is disabled no longer invokes xcodebuild. Previously `tuist test` surfaced xcodebuild's own error; now it finishes as a no-op.
- **Explicit `--test-targets` naming a plan-disabled target.** Now a silent no-op rather than an xcodebuild error. `-only-testing` can never run a plan-disabled target, so neither outcome runs tests — but the new one is quieter, which is worth revisiting alongside the broader "a run that tested nothing should say so" work.

A known remaining divergence, now documented in the code rather than papered over: `BuildGraphInspector.testableTarget`'s final no-plan fallback still anchors on `testAction.targets.first` without consulting `isSkipped`. Aligning it changes which target is chosen as buildable and is deliberately out of scope here.

## Impact

Projects that disable targets in a test plan will see their reported test-target count drop and their selective-testing hit rate change, since those targets leave the denominator. This is a correction, but it will read as a discontinuity in historical charts.

## Validation

- Four unit tests, one per patched branch. **Mutation-tested**: removing any single branch's filter fails exactly one test, so every branch is individually pinned — including the default-plan branch that plain `tuist test MyScheme` uses.
- A scheme-planning test pinning the splitting behavior described above.
- `TuistKitTests`: 333 XCTest + 340 swift-testing, 0 failures. `TuistCoreTests`: 299 + 49, 0 failures.
- End-to-end against a self-hosted server with selective testing active, using a fixture with one enabled and one plan-disabled unit-test target. At identical cache state, with only the binary differing: **before** — `test_targets ['A','B']`, hits `['A']`, missed `['B']`, and the run finishes early having tested nothing (the phantom miss, reproduced); **after** — `test_targets ['A']`, hits `['A']`, nothing missed, and finishing early is now the truthful outcome.
- `mise run lint` clean on the changed files.

### DIFF
--- a/cli/Sources/TuistKit/Services/TestService.swift
+++ b/cli/Sources/TuistKit/Services/TestService.swift
@@ -1715,16 +1715,17 @@ public struct TestService { // swiftlint:disable:this type_body_length
             scheme.testAction?.testPlans?
                 .first(
                     where: { $0.name == testPlanConfiguration.testPlan }
-                )?.testTargets.enabled.map(\.target) ?? []
+                )?.testTargets.filter { !$0.isSkipped }.map(\.target) ?? []
         } else if action == .build, let testPlans = scheme.testAction?.testPlans {
             // If we are building a scheme that has testplans but none specified then we should return all test targets
-            testPlans.flatMap(\.testTargets).enabled.map(\.target)
+            testPlans.flatMap(\.testTargets).filter { !$0.isSkipped }.map(\.target)
         } else if let defaultTestPlan = scheme.testAction?.testPlans?.first(where: {
             $0.isDefault
         }) {
-            defaultTestPlan.testTargets.enabled.map(\.target)
-        } else if let testActionTargets = scheme.testAction?.targets.enabled.map(\.target),
-                  !testActionTargets.isEmpty
+            defaultTestPlan.testTargets.filter { !$0.isSkipped }.map(\.target)
+        } else if let testActionTargets = scheme.testAction?.targets
+            .filter({ !$0.isSkipped }).map(\.target),
+            !testActionTargets.isEmpty
         {
             testActionTargets
         } else {
--- a/cli/Sources/TuistKit/Services/TestService.swift
+++ b/cli/Sources/TuistKit/Services/TestService.swift
@@ -1705,26 +1705,26 @@ public struct TestService { // swiftlint:disable:this type_body_length
         testPlanConfiguration: TestPlanConfiguration?,
         action: XcodeBuildTestAction
     ) -> [TargetReference] {
-        // Skipped testables are dropped here so this stays in sync with the two places that decide
-        // what actually runs — `BuildGraphInspector.testableTarget` and the test-plan filter in
-        // `GraphTraverser.filterIncludedTargets`. Counting them would report a target as needing
-        // tests that xcodebuild will never run, and since it never runs it never stores a hash, so
-        // it stays an un-cached miss on every subsequent run.
+        // Skipped testables are dropped because xcodebuild never runs them: counting one reports a
+        // target as needing tests that can never record a result, so it stays an un-cached miss on
+        // every subsequent run. On the plan-based paths this matches the test-plan filter in
+        // `GraphTraverser.filterIncludedTargets` and `BuildGraphInspector.isIncluded`. Note the two
+        // still diverge on the no-plan path: `BuildGraphInspector.testableTarget`'s final fallback
+        // anchors on `testAction.targets.first` without consulting `isSkipped`.
         return if let testPlanConfiguration {
             scheme.testAction?.testPlans?
                 .first(
                     where: { $0.name == testPlanConfiguration.testPlan }
-                )?.testTargets.filter { !$0.isSkipped }.map(\.target) ?? []
+                )?.testTargets.enabled.map(\.target) ?? []
         } else if action == .build, let testPlans = scheme.testAction?.testPlans {
             // If we are building a scheme that has testplans but none specified then we should return all test targets
-            testPlans.flatMap(\.testTargets).filter { !$0.isSkipped }.map(\.target)
+            testPlans.flatMap(\.testTargets).enabled.map(\.target)
         } else if let defaultTestPlan = scheme.testAction?.testPlans?.first(where: {
             $0.isDefault
         }) {
-            defaultTestPlan.testTargets.filter { !$0.isSkipped }.map(\.target)
-        } else if let testActionTargets = scheme.testAction?.targets
-            .filter({ !$0.isSkipped }).map(\.target),
-            !testActionTargets.isEmpty
+            defaultTestPlan.testTargets.enabled.map(\.target)
+        } else if let testActionTargets = scheme.testAction?.targets.enabled.map(\.target),
+                  !testActionTargets.isEmpty
         {
             testActionTargets
         } else {
--- a/cli/Sources/TuistKit/Services/TestService.swift
+++ b/cli/Sources/TuistKit/Services/TestService.swift
@@ -1705,20 +1705,26 @@ public struct TestService { // swiftlint:disable:this type_body_length
         testPlanConfiguration: TestPlanConfiguration?,
         action: XcodeBuildTestAction
     ) -> [TargetReference] {
+        // Skipped testables are dropped here so this stays in sync with the two places that decide
+        // what actually runs — `BuildGraphInspector.testableTarget` and the test-plan filter in
+        // `GraphTraverser.filterIncludedTargets`. Counting them would report a target as needing
+        // tests that xcodebuild will never run, and since it never runs it never stores a hash, so
+        // it stays an un-cached miss on every subsequent run.
         return if let testPlanConfiguration {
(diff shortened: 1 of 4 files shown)


cli/Sources/TuistKit/Services/TestService.swift
```diff
@@ -1705,20 +1705,26 @@ public struct TestService { // swiftlint:disable:this type_body_length
         testPlanConfiguration: TestPlanConfiguration?,
         action: XcodeBuildTestAction
     ) -> [TargetReference] {
+        // Skipped testables are dropped here so this stays in sync with the two places that decide
```
esnunes: Small accuracy note on the new comment: it says this stays in sync with `BuildGraphInspector.testableTarget` and `GraphTraverser.filterIncludedTargets`, but `testableTarget`'s no-plan fallback (`BuildGraphInspector.swift:161`) picks `scheme.testAction?.targets.first` without checking `isSkipped`. On a no-plan scheme whose first testable is skipped, the two functions now disagree — this one excludes it, that one anchors the buildable target on it. Might be worth scoping the comment to the plan-based paths, or aligning the fallback while you're here.

cli/Sources/TuistKit/Services/TestService.swift
```diff
@@ -1705,20 +1705,26 @@ public struct TestService { // swiftlint:disable:this type_body_length
...
             // If we are building a scheme that has testplans but none specified then we should return all test targets
-            testPlans.flatMap(\.testTargets).map(\.target)
+            testPlans.flatMap(\.testTargets).filter { !$0.isSkipped }.map(\.target)
         } else if let defaultTestPlan = scheme.testAction?.testPlans?.first(where: {
             $0.isDefault
         }) {
-            defaultTestPlan.testTargets.map(\.target)
-        } else if let testActionTargets = scheme.testAction?.targets.map(\.target),
-                  !testActionTargets.isEmpty
+            defaultTestPlan.testTargets.filter { !$0.isSkipped }.map(\.target)
```
esnunes: Two small gaps in the new tests: (1) the default-plan branch (`TestService.swift:1724`) is the only patched branch without a test — the first test sets `isDefault: true` but passes a named `TestPlanConfiguration`, so it exercises the named-plan branch; reverting just that one filter leaves the suite green, and that branch serves plain `tuist test MyScheme`, the most common shape. (2) The build-action test asserts only the analytics dictionary and skips the `XCTAssertEqual(testedSchemes, ...)` check the other two have, so it wouldn't catch the scheme being silently dropped from the build. Both look like one-liners to close.

cli/Sources/TuistKit/Services/TestService.swift
```diff
@@ -1705,20 +1705,26 @@ public struct TestService { // swiftlint:disable:this type_body_length
...
+        // what actually runs — `BuildGraphInspector.testableTarget` and the test-plan filter in
+        // `GraphTraverser.filterIncludedTargets`. Counting them would report a target as needing
+        // tests that xcodebuild will never run, and since it never runs it never stores a hash, so
+        // it stays an un-cached miss on every subsequent run.
         return if let testPlanConfiguration {
             scheme.testAction?.testPlans?
                 .first(
                     where: { $0.name == testPlanConfiguration.testPlan }
-                )?.testTargets.map(\.target) ?? []
+                )?.testTargets.filter { !$0.isSkipped }.map(\.target) ?? []
```
esnunes: Cleanup thought, definitely not a blocker: `.filter { !$0.isSkipped }` is now inlined four times here, the same predicate lives in `GraphTraverser+Extra.swift:16`, and `BuildGraphInspector.isIncluded` carries a superset check — since the new comment says these must stay in sync, a small `enabled` helper on `Sequence<TestableTarget>` would give the concept one home. Same spirit for the three new ~90-line tests, which duplicate the fixture nearly verbatim; a shared builder would cut ~150 lines and keep the next skip-related test from being a fourth copy.

esnunes (commented):
Nice find on the phantom-miss mechanism, and the PR description made this easy to review — the three-paths table and the revert-verified tests especially. One theme worth recording: `testActionTargetReferences` isn't analytics-only — it also feeds the scheme-run gate and the mixed hosted/hostless scheme splitting, so the filter changes runtime behavior for projects with skipped testables (verified against Xcode 26.3; details below and in the inline notes). Those new behaviors are intended — building or running nothing when everything is disabled is correct — but they're broader than the description frames, so the notes record them for reviewers. The inline notes cover the sync comment's accuracy, two test-coverage gaps, and a possible shared helper for the skip predicate.

---

Two notes on lines outside the diff:

**All-skipped schemes/plans (`TestService.swift:1204`)** — Heads-up for reviewers — no change requested: this also changes runtime behavior for a scheme or plan whose testables are all skipped. It's now dropped from `testSchemeRuns` entirely (the empty-set guard around `TestService.swift:1204`), so xcodebuild isn't invoked at all. Verified against Xcode 26.3: previously `build-for-testing` built every test product (including plan-disabled ones) and plain `test` failed with "Scheme … is not currently configured for the test action"; now `--build-only` builds nothing and plain `test` completes as a skipped run. Building nothing when there's nothing to run is the intended behavior here — flagging it because it's a behavioral change beyond analytics, and it may deserve a line in the PR description.

**Mixed hosted/hostless scheme splitting (`TestService.swift:1457-1538`)** — The mixed hosted/hostless scheme splitting also consumes this list: mixed-ness detection, the workspace target set, and the single-target eligibility check all now see the filtered targets, so which schemes get split and invoked can shift for projects with skipped testables. The new outcomes look more correct to me — a skipped target arguably shouldn't count as "mixed" — but the change is untested; a case with a skipped hostless target would pin it. Definitely not a blocker.

cli/Sources/XcodeGraph/Sources/XcodeGraph/Models/TestableTarget.swift
```diff
@@ -81,3 +81,12 @@ public struct TestableTarget: Equatable, Hashable, Codable, Sendable {
...
     #endif
 }
+
+extension Sequence<TestableTarget> {
+    /// The testables the plan actually runs. A skipped testable is disabled in the test plan, so
+    /// `xcodebuild` never runs it — treating it as part of the run makes it look like work that is
+    /// pending forever, because it can never record a result.
+    public var enabled: [TestableTarget] {
+        filter { !$0.isSkipped }
+    }
```
fortmarek: I find this pattern a bit odd. I'd remove this extension and just write `.filter { !$0.isSkipped }`. I think this sugar is making _less_ readable, not more.

fortmarek (approved):

PR #12275, opened 2026-08-11
fix(app): prevent authentication failures after relaunch

pepicrft:
## What changed

The Tuist iPhone application now uses the shared authentication dependency instances instead of installing fresh credential and refresh-coordination stores around its root view. It also persists a bounded, generic application log history on iPhone and Mac. When an authentication error is shown on iPhone, the alert offers to share a redacted plain-text log file. The permanent diagnostics actions have been removed from the sign-in and profile screens.

## Why

Some users could reopen the application, still see their signed-in profile, and receive “You need to be authenticated to access this resource” when loading Previews. Signing out and back in restored access, but removed the state needed to understand why the stored session had become inconsistent.

## Root cause

The application entry point created separate credential and refresh-coordination store instances around the root view. Those instances shared secure storage but not their in-memory change streams and refresh coordination. After an access credential expired, one context could rotate the credentials while another continued with the previous value. The persisted signed-in presentation state then disagreed with the credentials used for requests.

## Approach

The root view now relies on the shared default stores used by the authentication service and request clients. This keeps credential observation and refresh coordination in one context.

Application logging remains in the existing `TuistLogging` module and reuses the same `SimpleFileLogHandler` used by the command-line application. The iPhone and Mac applications bootstrap that handler alongside the existing console handler, without adding another target or a third-party logging dependency. Application code accesses the store through a protocol-backed task-local value, matching the dependency pattern used elsewhere in Tuist.

The file handler owns synchronized storage shared by every handler copy. Its lock protects writes, size-based truncation, synchronization, and reads, so exporting a report cannot race with an active write. Sensitive authentication values are redacted before each line reaches disk. Each launch is limited to five megabytes, and the application retains at most two launches for three days.

The support export is plain text and combines the retained launch files with application and operating-system context. Logs are prepared and offered only when the application presents an authentication error, keeping diagnostics close to the failure instead of adding permanent interface controls. Each concurrent export receives a unique temporary filename, and the temporary file is removed after sharing or cancellation.

## Impact

Existing signed-in sessions should refresh consistently after the application is terminated and reopened. There is no migration or user action required. If another authentication failure occurs, users can share recent application logs with support directly from the error alert. The application uses up to ten megabytes of local storage for logs and removes launch files older than three days.

## Validation

- Reproduced the reported error in a signed simulator build against the local Tuist server by shortening the access credential lifespan to five seconds, signing in, terminating the application, and reopening it after expiration.
- Repeated the same lifecycle with the fix and confirmed that credentials refreshed and Previews loaded without an authentication alert.
- Passed five focused application logging tests covering persistence, redaction before storage, two-launch retention, task-local overrides, and concurrent exports.
- Passed the focused server authentication controller tests.
- Passed the focused authentication service tests on the Mac application target.
- Built the iPhone simulator application with Xcode and code signing disabled.
- Built the Mac application with Xcode and code signing disabled.
- Checked the implementation with the repository's Swift formatting and lint configuration.
- Confirmed that Pulse and PulseLogHandler are no longer present in the package or project manifests.

### How to test locally

1. Run the local Tuist server and generate the application workspace for the development environment.
2. Build and install the Tuist application in an iPhone simulator with signing enabled.
3. Sign in, let the access credential expire, terminate the application, and reopen it.
4. Confirm that Previews loads without requesting another sign-in.
5. Trigger an authentication error and confirm the alert offers to share a `.txt` file containing recent application logs without credential values.

## Screenshots

| Before | After |
| --- | --- |
| The signed-in application reports that authentication is required after reopening. | The same expired-credential lifecycle refreshes successfully and loads Previews. |
| Authentication error reproduced after reopening the application | Previews loaded after refreshing authentication credentials |

### Log sharing experience

| Authentication error | Plain-text share sheet |
| --- | --- |
| The error alert offers log sharing only when the failure is related to authentication. | The system share sheet receives a timestamped `.txt` file that can be copied or saved. |
| Authentication error alert with Share logs action | System share sheet containing the generated plain-text Tuist log file |


app/Sources/TuistOnboarding/LogInView.swift
```diff
@@ -76,6 +77,16 @@ public struct LogInView: View {
...
+
+                ShareLink(
+                    item: AuthenticationDiagnostics.shared.currentProcessReport(),
+                    subject: Text("Tuist authentication diagnostics")
+                ) {
+                    Text("Share authentication diagnostics")
+                        .font(.footnote)
+                        .foregroundColor(Noora.Colors.surfaceLabelSecondary)
+                }
+                .padding(.top, Noora.Spacing.spacing2)
```
fortmarek: This feels wrong to have in a sign in page. If anywhere, I think it should be in an alert that shows the authentication error, which is where a user would probably decide to surface diagnostics (fwiw, sharing the diagnostics should be probably more generic than just for auth, but we can start with auth only)

app/Sources/TuistProfile/ProfileView.swift
```diff
@@ -63,6 +63,21 @@ public struct ProfileView: View {
...
+                    HStack {
+                        Text("Share authentication diagnostics")
+                            .font(.body)
+                            .foregroundColor(Noora.Colors.surfaceLabelPrimary)
+                        Spacer()
+                        Image(systemName: "square.and.arrow.up")
+                            .font(.title3.weight(.regular))
+                            .foregroundColor(Noora.Colors.surfaceLabelTertiary)
+                    }
+                }
```
fortmarek: similarly as above, wouldn't include a permanent button for this

cli/Sources/TuistServer/Utilities/AuthenticationDiagnostics.swift
```diff
@@ -0,0 +1,228 @@
...
+
+    public init(maximumEventCount: Int = 200) {
+        precondition(maximumEventCount > 0)
+        self.maximumEventCount = maximumEventCount
+    }
+
+    public func record(
+        _ event: AuthenticationDiagnosticEvent,
+        at date: Date = Date()
+    ) {
```
fortmarek: recording authentication events should be imho a fire-and-forget operation. If we do that, we can remove the `unchecked Sendable` and probably even get rid of locks for something more concurrency native (I assume the current code was written this way to ensure `record` doesn't end up being async)

cli/Sources/TuistServer/Utilities/AuthenticationDiagnostics.swift
fortmarek, line 1: I'm not sure how I feel about this. I think instead the iOS app should support gathering logs and sharing any logs, instead of having something so scoped to a particular function of the app, in this case auth.

fortmarek (commented):
The fix looks good! The authentication logging imho needs some work.

app/Sources/TuistOnboarding/LogInView.swift
pepicrft: Done. I removed the permanent controls and moved sharing to the authentication error alert. The alert now prepares a redacted plain-text export and opens the system share sheet.

app/Sources/TuistProfile/ProfileView.swift
pepicrft: Done. The permanent Profile button is gone. Sharing is offered contextually from the authentication error alert.

cli/Sources/TuistServer/Utilities/AuthenticationDiagnostics.swift
pepicrft: Done. I replaced the custom diagnostics store with Pulse and PulseLogHandler. Existing Logger.current calls feed Pulse, which persists asynchronously, so there is no custom lock or unchecked Sendable implementation in our code.

cli/Sources/TuistServer/Utilities/AuthenticationDiagnostics.swift
pepicrft: Done. The implementation is now generic application logging under TuistAppLogging. Pulse captures the application logger output, and the authentication alert is only the first contextual place from which those logs can be shared.

app/Sources/TuistApp/TuistApp.swift
```diff
@@ -62,15 +63,13 @@ import TuistServer
...
         var body: some Scene {
             WindowGroup {
-                ServerCredentialsStore.$current.withValue(
-                    ServerCredentialsStore(backend: .keychain)
-                ) {
-                    CachedValueStore.$current.withValue(CachedValueStore(backend: .inSystemProcess)) {
-                        RootView()
-                    }
-                }
+                RootView()
```
esnunes: I like removing these wrappers, but I couldn't convince myself they were ever in effect, so I'm unsure this is the fix for the reported logouts.

Both used the synchronous `withValue`, so the binding only covered the `RootView()` initializer inside `WindowGroup`'s content builder. `RootView` reads neither task local in `init` — the `@StateObject` thunk is stored unevaluated and SwiftUI calls it on first render, after the scope unwinds — and that thunk is the one place it would matter, because `AuthenticationService.init` → `startCredentialsListener()` captures the task-local context when it creates its `Task` (AuthenticationService.swift:71).

On top of that, the iOS defaults are the same values that were being injected: `ServerCredentialsStore.current` defaults to `.keychain` (ServerCredentialsStore.swift:95 — and `.keychain` is the only backend that compiles for iOS) and `CachedValueStore.current` to `.inSystemProcess` (CachedValueStore.swift:32). So reader and writer already resolved to the same default instance.

I'd keep the deletion regardless, it removes a trap. But you reproduced the lifecycle before and after with a 5s access-token lifetime — what differed between the two runs? If the wrapper wasn't the cause, the symptom would survive this merge.

cli/Sources/TuistServer/Client/ServerAuthenticationController.swift
```diff
@@ -610,9 +634,9 @@ public struct ServerAuthenticationController: ServerAuthenticationControlling {
                     }
                 #endif
                 if refresh {
-                    #if canImport(TuistSupport)
-                        Logger.current.debug("Refreshing access token for \(serverURL)")
-                    #endif
+                    Logger.current.notice(
```
esnunes: These three land in CLI output, since `ServerAuthenticationController` is shared and the CLI's console handler runs at `.info` (`Logger.defaultLoggerHandler` → `StandardLogHandler(label:)`, which defaults to `.info`).

- `.notice` at :637 and :777 print to stdout on every token refresh. :637 also replaced a `.debug` line that was behind `#if canImport(TuistSupport)`.
- the `.warning` at :318 is more than stdout noise: with coloured output, `StandardLogHandler.log` routes `.warning` into `AlertController.current.warning(...)` (StandardLogHandler.swift:49-55), so the concurrent-rotation case the code deliberately absorbs surfaces as a Noora warning banner.

Would `.debug` work for all three? The app's persistent handler is bootstrapped at `.debug`, so the iOS diagnostics still capture them.

cli/Sources/TuistLogging/ApplicationLogStore.swift
```diff
@@ -0,0 +1,176 @@
...
+            configuration.sizeLimit = Self.maximumStoreSize
+            configuration.maxAge = Self.maximumAge
+            configuration.willHandleEvent = { event in
+                existingEventHandler(event).map(self.sanitized)
+            }
+            LoggerStore.shared.configuration = configuration
+
+            LoggingSystem.bootstrap { label in
+                var persistentLogHandler = PersistentLogHandler(label: label, store: .shared)
+                persistentLogHandler.logLevel = .debug
```
esnunes: I think the persistence level needs a second look before this ships. `persistentLogHandler.logLevel = .debug` switches on `VerboseLoggingMiddleware`, which is installed unconditionally on both clients (Client+Server.swift:19,34) and logs complete request/response bodies up to 64 KB at `.debug` (VerboseLoggingMiddleware.swift:38-56). Before this PR the app never bootstrapped swift-log, so those went nowhere.

The case that worries me: the app lists previews through `Client.authenticated` (ListPreviewsService.swift:65), and the server fills each `AppBuild.url` with `Storage.generate_download_url` (previews_controller.ex:796) — an S3 SigV4 presigned link with `X-Amz-Credential`/`X-Amz-Signature`/`X-Amz-Security-Token`, valid an hour (storage.ex:62-73). None of the three patterns in `redacted(_:)` match it: no `bearer` prefix, none of the six `key=value` names, not `eyJ`-shaped. `VerboseLoggingMiddleware` does list those header names as sensitive, but that set only applies to headers, not bodies.

So whoever receives the exported file can download the org's private builds without any Tuist authentication, while the header says "Sensitive authentication values are redacted." Access/refresh tokens are covered, but only by the JWT shape rule.

There's also a practical cost: 64 KB bodies eat the 1,000-message / 10 MB budget quickly and evict the authentication events the export exists to show.

Would `.info` for the persistent handler work? Or excluding the `VerboseLoggingMiddleware` label inside `willHandleEvent`?

cli/Sources/TuistServer/Client/ServerAuthenticationController.swift
```diff
@@ -748,16 +772,39 @@ public struct ServerAuthenticationController: ServerAuthenticationControlling {
...
                     serverURL: serverURL
                 )
+            let accessToken = try? JWT.parse(newTokens.accessToken)
+            let refreshToken = try? JWT.parse(newTokens.refreshToken)
+            Logger.current.notice(
+                "Token refresh succeeded \(expirationDescription(accessToken: accessToken, refreshToken: refreshToken))"
+            )
             return newTokens
         } catch let error as ClientError {
+            Logger.current.error(
```
esnunes: Small ordering thing: this `.error` runs before `ServerErrorClassifier.isTransient(error)` decides to rethrow for the retry layer. `isTransient` covers 408/429/5xx and `NSURLErrorDomain` timeouts and connection drops, so a single dropped packet on an otherwise successful command prints a red "Token refresh failed" to stderr while the command exits 0.

Moving it below the `isTransient` branch, or logging the transient case at `debug`, would keep the error level meaningful.

Package.swift
```diff
@@ -1881,6 +1887,8 @@ let package = Package(
         .package(id: "swiftGen.SwiftGen", exact: "6.6.2"),
         .package(id: "sparkle-project.Sparkle", from: "2.6.4"),
         .package(id: "kean.Nuke", .upToNextMajor(from: "12.8.0")),
+        .package(url: "https://github.com/kean/Pulse.git", from: "5.2.3"),
+        .package(url: "https://github.com/kean/PulseLogHandler.git", from: "5.1.0"),
```
esnunes: Heads-up on the lockfile shape: `Package.resolved` now has 85 `registry` pins and exactly one `remoteSourceControl` pin, `pulseloghandler`. `kean/Pulse` maps to the registry as `kean.pulse`, but `kean/PulseLogHandler` has no registry mirror, so resolution needs a direct github.com clone — including on Linux CI, where the product is excluded by the platform condition anyway.

Since we resolve with `--replace-scm-with-registry`, a resolve without the flag will churn this pin. Is there a registry id for it? Otherwise vendoring the small `PersistentLogHandler` into `TuistLogging` would keep the lockfile uniform.

app/Sources/TuistErrorHandling/ErrorHandling.swift
```diff
@@ -51,16 +90,55 @@ struct HandleErrorsByShowingAlertViewModifier: ViewModifier {
...
+                                title: Text("Error"),
+                                message: Text(currentAlert.message),
+                                dismissButton: .default(Text("Ok")) {
+                                    currentAlert.dismissAction?()
+                                }
+                            )
+                        }
                     }
             )
+            .sheet(item: $logsToShare) { logsToShare in
```
esnunes, line 118: `logsToShare` is never set back to `nil` — not in the `completion` closure and not in an `onDisappear`. `UIActivityViewController` dismisses itself when an activity finishes, and as a child of the sheet's hosting controller that dismissal can tear the sheet down without SwiftUI updating the binding, leaving the item non-nil while SwiftUI still thinks the sheet is presented.

Worth checking on device: auth error → Share logs → complete or cancel → second auth error → Share logs. Setting `logsToShare = nil` in `completion` and in `onDisappear` would cover it either way.

cli/Sources/TuistLogging/ApplicationLogStore.swift
```diff
@@ -0,0 +1,176 @@
...
+        private static let maximumAge: TimeInterval = 3 * 24 * 60 * 60
+        private static let maximumSessionCount = 2
+        private static let maximumMessageCount = 1000
+        @MainActor private static var hasBootstrapped = false
+
+        public init() {}
+
+        @MainActor
+        public func bootstrap() {
+            guard !Self.hasBootstrapped else { return }
```
esnunes, line 37: The assignment does take effect — Pulse reads `sizeLimit`/`maxAge` lazily inside `_sweep()`. The nuance is when the sweep runs: `sweep()` has one caller in the package, `postInitialization()`, which schedules it once 10s after the store is created, and that decision uses the *default* configuration and returns false outright when `manifest.lastSweepDate` is nil (LoggerStore.swift:200-205, :1073-1080).

So the first run after install never trims, there's no periodic sweep, and a long-lived foreground session can grow past 10 MB. If the cap is a guarantee we want to state, an explicit `LoggerStore.shared.sweep()` in `bootstrap()` (or before an export) would make it real.

Package.swift
```diff
@@ -580,6 +580,12 @@ var targets: [Target] = [
             "TuistEnvironment",
             "TuistAlert",
             .product(name: "LoggingOSLog", package: "chrisaljoudi.swift-log-oslog", condition: .when(platforms: [.macOS])),
+            .product(name: "Pulse", package: "Pulse", condition: .when(platforms: [.iOS, .macOS])),
```
esnunes: `TuistLogging` is a foundation module of the CLI (`TuistHTTP` depends on it unconditionally, Package.swift:142), so conditioning Pulse/PulseLogHandler on `[.iOS, .macOS]` links a Core Data-backed logging framework into the distributed `tuist` binary, and `ApplicationLogStore.swift` compiles there too.

Nothing on macOS calls `bootstrap()` today — the macOS `TuistApp` entry point doesn't (TuistApp.swift:11-24) — so the menu-bar app doesn't benefit yet either. Was `.macOS` there for a planned menu-bar adoption, or would `.iOS` be enough for now (here, in Module.swift:612, and in the `#if`)?

app/Sources/TuistErrorHandling/ErrorHandling.swift
```diff
@@ -27,20 +36,50 @@ public final class ErrorHandling: ObservableObject {
...
+        withLogExport: Bool
+    ) {
+        guard withLogExport else {
+            currentAlert = ErrorAlert(message: message)
+            return
+        }
+
+        Task {
+            do {
+                let logExportURL = try await ApplicationLogStore.current.plainTextExport()
```
esnunes: `showAlert` awaits `plainTextExport()` before assigning `currentAlert`, so every authentication error writes up to 1,000 log lines to tmp whether or not the user taps Share logs — and the alert is delayed by that write.

Cleanup covers the "Ok" branch (:103) and share-sheet completion (:120), but not the case where a second error replaces `currentAlert` first, and nothing prunes older `tuist-logs-*.txt` at launch.

Generating the file inside the Share logs action would remove the unprompted write, the delay, and most of the cleanup paths in one go.

app/Sources/TuistErrorHandling/ErrorHandling.swift
```diff
@@ -17,8 +20,14 @@ public final class ErrorHandling: ObservableObject {
     public func handle(error: Error) {
         if let clientError = error as? ClientError {
             if clientError.underlyingError is ClientAuthenticationError {
+                let clientErrorType = String(reflecting: type(of: clientError))
+                let underlyingErrorType = String(reflecting: type(of: clientError.underlyingError))
                 Logger.current.error(
-                    "Client authentication error received. Deleting stored credentials. Error: \(clientError.underlyingError.localizedDescription)"
+                    "Client authentication error received. Deleting stored credentials. error_type=\(clientErrorType) underlying_error_type=\(underlyingErrorType) error=\(clientError.underlyingError.localizedDescription)"
+                )
+                showAlert(
```
esnunes, line 28: Behaviour question rather than a defect: this branch used to log, delete the credentials and return — the user just landed back on the sign-in screen. It now also shows "Error" with the raw `localizedDescription` plus a Share logs button, so an ordinary refresh-token expiry becomes an error dialog before the sign-out.

Deliberate for the diagnostics goal, or a side effect of routing this branch through `showAlert`?

esnunes (approved):
Thanks for the rework — moving the diagnostics into the existing logging pipeline and dropping the permanent buttons is a much better shape than the first round, and the write-up made the change easy to follow.

Two things I'd like to settle before this lands, and neither is about the code being wrong so much as about what it does:

1. I couldn't convince myself the removed task-local wrappers were ever in effect on iOS. They used the synchronous `withValue`, so the binding only covered `RootView()`'s initializer, and the iOS defaults are the same values that were being injected. I'd keep the deletion either way, but if that reading is right, the reported logouts would survive the merge — so I'd love to know what actually differed in your before/after reproduction.

2. Bootstrapping the persistent handler at `.debug` switches on `VerboseLoggingMiddleware`, which logs full request and response bodies up to 64 KB. Those now persist and end up in a file the alert invites users to send to support. The concrete one: the Previews responses carry one-hour presigned S3 download URLs for the org's private app builds, and none of the three redaction patterns match them — while the export header states that sensitive values are redacted.

The rest are smaller: three new log lines land in CLI output because the controller is shared with the CLI, and there are a few resource, retention and redaction details around the export. Details are in the inline notes.

---

One more that sits on an unchanged line, so I couldn't anchor it inline — `app/Sources/TuistAuthentication/AuthenticationService.swift:81`:

One interaction worth knowing about now that these lines persist: `JWT.parse` throws `JWTError.invalidJWT(jwt)` carrying the whole raw string when it doesn't split into three components (JWT.swift:80), and the description interpolates it — "The access token \(token) is invalid…" (JWT.swift:9). This line and ErrorHandling.swift:41 both log `localizedDescription`, so both now reach the Pulse store.

The redaction's JWT rule needs three non-empty segments, so exactly the malformed/truncated tokens that trigger this error are the ones it can't match. On a branch about unexpected logouts, corrupt stored tokens are the likeliest thing to end up in a shared log.

Logging a length/prefix fingerprint instead of the token would close it. That's a change in `JWT.swift` rather than this diff, so a follow-up is fine by me.

app/Sources/TuistApp/TuistApp.swift
pepicrft: You are right. The synchronous task-local scopes do not encompass SwiftUI’s later state-object initialization, and the injected mobile values match the defaults. I cannot substantiate those scopes as the logout cause, so their removal is defensive cleanup rather than the claimed fix.

cli/Sources/TuistServer/Client/ServerAuthenticationController.swift
pepicrft: Addressed in the current branch. The rotation, refresh-start, and refresh-success events now use debug severity, so they remain in the mobile application log store but do not add normal command-line output or an alert banner.

cli/Sources/TuistLogging/ApplicationLogStore.swift
pepicrft: Addressed in the current branch. ApplicationLogStore rejects verbose network request and response messages before persistence. The focused test uses a presigned download link and verifies that the stored file remains empty.

cli/Sources/TuistServer/Client/ServerAuthenticationController.swift
pepicrft: Addressed in the current branch. refreshTokens now classifies transient client failures first, logs them at debug severity, and rethrows them for retry. Error severity is reserved for failures that are converted to an authentication error.

Package.swift
pepicrft: This is obsolete after the later refactor. The branch no longer references Pulse or PulseLogHandler in Package.swift or Package.resolved. Application logging now uses the in-tree SimpleFileLogHandler, so there is no source-control-only lockfile pin.

app/Sources/TuistErrorHandling/ErrorHandling.swift
pepicrft: Addressed in the current branch. Both the activity completion callback and the sheet disappearance handler delete the temporary export and clear logsToShare, so a subsequent share action receives a new item.

cli/Sources/TuistLogging/ApplicationLogStore.swift
pepicrft: Addressed by replacing the store with explicit file-session retention. A launch creates a file capped at 5 megabytes, and startup removes sessions older than three days or beyond the two newest. The focused test covers retaining only the two newest sessions.

Package.swift
pepicrft: This is obsolete after the later refactor. ApplicationLogStore is an in-tree file logger with no Core Data or Pulse dependency. It remains available on Apple platforms for future menu-bar adoption, but only the mobile entry point bootstraps it today.

app/Sources/TuistErrorHandling/ErrorHandling.swift
pepicrft: Addressed in the current branch. The alert is shown immediately, and the export is created only after the user taps Share logs. The completion and disappearance paths remove the resulting temporary file.

app/Sources/TuistErrorHandling/ErrorHandling.swift
pepicrft: This is deliberate. Authentication failures present the contextual share action before credential deletion returns the view to sign-in, which is the only point at which logs are useful for support. The malformed-token description now reports only token length, so the alert and stored log do not expose the raw token.

PR #12276, opened 2026-08-12
feat(server): redesigned marketing home page behind a feature flag

asmitbm:
First slice of the marketing site redesign. Ships the shared foundation (TuistWeb.Marketing.Design plug + per-page FunWithFlags gating, design tokens, theme bootstrap with dark mode, the bundle-new.css esbuild entry) plus the redesigned navbar, footer, and home page. Nothing changes for visitors: legacy pages render until the :new_marketing_home flag is flipped; the new design can be previewed with ?design=new.

Screenshot 2026-08-12 at 09 36 51
Screenshot 2026-08-12 at 09 37 07
Screenshot 2026-08-12 at 09 37 31
Screenshot 2026-08-12 at 09 37 52
image
Screenshot 2026-08-12 at 09 38 41
Screenshot 2026-08-12 at 09 39 04
Screenshot 2026-08-12 at 09 39 15


server/lib/tuist_web/marketing/controllers/marketing_html/new/home.html.heex
```diff
@@ -0,0 +1,1054 @@
+<TuistWeb.Marketing.Layouts.marketing {assigns}>
+  <main id="marketing-home">
+    <div data-part="platform-hero" id="marketing-platform-hero" phx-hook="PlatformBackground">
```
fortmarek, line 3: [P1] These hooks will not mount on this route. `/` is still rendered by `MarketingController`, so there is no LiveView root for `LiveSocket` to join; in dead-view mode `phx-hook` lifecycle callbacks are not run. That leaves the new navbar/mobile menu/theme switchers and the many canvas illustrations inert or blank. Please either render this page through a LiveView or explicitly initialize and tear down these behaviors from the marketing JS entrypoint.

server/lib/tuist_web/marketing/design.ex
```diff
@@ -0,0 +1,98 @@
...
+  defp resolve("old", _page), do: false
+  defp resolve(_, page), do: Tuist.FeatureFlags.new_marketing_page_enabled?(page)
+
+  # A `?design=` param must win within the request that sets the cookie too:
+  # response cookies aren't visible in conn.cookies until the next request.
+  defp override(conn) do
+    case conn.query_params["design"] do
+      value when value in ["new", "old"] -> value
+      "default" -> nil
+      _ -> conn.cookies[@cookie]
```
fortmarek: [P1] This makes the response body depend on `tuist_marketing_design`, while marketing controller responses are still sent with `Cache-Control: public, max-age=60, stale-while-revalidate=86400`. Once the cookie and session agree, this request emits no `Set-Cookie`, so an intermediary can cache preview-specific HTML at the ordinary `/` URL and serve it to other visitors (or cache the old page after rollout). Please make override-driven responses private/non-cacheable, or include the design variant in the cache key.

server/lib/tuist_web/marketing/components/marketing_layout_components/navbar_new.html.heex
```diff
@@ -0,0 +1,541 @@
...
+      <img
+        data-part="logo"
+        data-theme="dark"
+        src={~p"/marketing/images/components/navbar/logo-dark.svg"}
+        alt="Tuist"
+      />
+    </.link>
+    <nav data-part="menus" id="marketing-navbar-menus" aria-label="Main navigation">
+      <div data-part="menu">
+        <div
```
fortmarek: [P1] The new desktop menu triggers are plain `<div>` elements. `NavbarMegaMenu` only attaches mouse/click handlers; it neither makes these elements focusable nor handles Enter/Space. Keyboard users therefore cannot open Features, Developers, or Resources and cannot reach most of the navigation links. Please use real `<button type="button">` triggers (preferred), or add equivalent focus and keyboard semantics.

noora/css/button.css
```diff
@@ -275,21 +279,31 @@
     color: var(--noora-button-neutral-disabled-label);
   }
 
+  & > span {
```
fortmarek: [P2] This selector treats every direct `span` as the new text label, but existing neutral buttons also use direct spans as icon wrappers (for example the open/closed chevrons in the test views). Those icons now receive horizontal label padding, changing their dimensions and alignment across existing Noora consumers. Please mark the generated label explicitly (for example `data-part="label"`) and scope this rule to that element.

fortmarek (commented):
I found four issues in the rollout path and shared components. Details are inline.

server/assets/marketing/css/layouts/components/footer-new.css
```diff
@@ -0,0 +1,181 @@
+/* Redesigned marketing footer (footer_new.html.heex). */
```
fortmarek: let's remove this comment (same goes for all the other css page). The name of the file being different is good enough.

server/assets/marketing/css/new/layouts/components/footer.css
fortmarek, line 1: nit: instead of having `-new.css` and `_new.html.heex`, I'd consider putting the new css files and html files under a `new` or `redesign` directory.

server/assets/marketing/css/new/routes/home-hero.css
```diff
@@ -0,0 +1,398 @@
+/*
+ * Home hero with the blueprint "platform" decoration (ported from the
+ * tuist-demo prototype). The composition is a fixed 1497x496 stage: wireframe
```
fortmarek: `tuist-demo` prototype links to something that's not relevant to the codebase. Let's make sure Claude only posts comments relevant based on the current state of things.

server/lib/tuist_web/marketing/controllers/marketing_html/new/home.html.heex
asmitbm: `phx-hook` does mount on this route. Since LiveView 1.0, `LiveSocket.connect()` calls `joinDeadView()` on pages without a LiveView root: it wraps `document.body` in a root view, joins it dead, and `execNewMounted()` mounts every `[phx-hook]` element (see `phoenix_live_view/assets/js/phoenix_live_view/live_socket.js` → `joinDeadView`, and the `ViewHook.deadHook` handling in `view.js`). The legacy home page already relies on this today — `home.html.heex` has `phx-hook="LogoTransition"` on the controller-rendered / and it animates in production. Verified on this branch too: the navbar mega menu, mobile menu, theme switcher, and all canvas illustrations run on `/?design=new`. The dead-view limitation `(no pushEvent/updated)` doesn't apply since these hooks are purely client-side.

server/lib/tuist_web/marketing/design.ex
asmitbm: Fixed: https://github.com/tuist/tuist/pull/12276/commits/7477b25b3d7eda364cd151cd7af47120c10dfa9a

fortmarek:
image

I'd rename Kura nodes to Cache Nodes

server/lib/tuist_web/marketing/design.ex
```diff
@@ -0,0 +1,116 @@
...
+  # override on a visitor's browser eventually expires on its own.
+  @cookie_max_age 60 * 60 * 24 * 30
+
+  def init(opts), do: opts
+
+  def call(conn, _opts) do
+    conn = conn |> fetch_query_params() |> fetch_cookies()
+
+    conn =
+      case conn.query_params["design"] do
```
fortmarek: **Remove the public URL override**

The `?design=` path gives any anonymous visitor access to the unfinished design and requires a substantial cookie/session/cache-bypass layer to keep that override working safely. It is fine for previews to require authentication, so please remove the query-parameter and cookie override and target authenticated users directly through FunWithFlags. Anonymous traffic can keep following the global per-page flag until rollout.

noora/lib/noora/button.ex
```diff
@@ -98,6 +98,8 @@ defmodule Noora.Button do
     """
   end
 
+  attr(:label, :string, default: nil, doc: "The label of the button")
```
fortmarek: **Keep visible content in the inner block**

Adding `label` creates two competing ways to provide the button body, and making the slot optional also allows `<.neutral_button />` to render an empty control. Please keep `inner_block` required and render “Log in” through it; this also preserves arbitrary content ordering, while `aria-label` remains the explicit API for icon-only buttons. Other components may already use the same `label` pattern, but we should avoid extending it here and clean those up separately.

fortmarek:
https://github.com/user-attachments/assets/9ac071b1-d3fe-4ac2-aa38-6369d846263e

the cursor animation for the footer doesn't seem to be working as intended (Firefox)

server/assets/marketing/js/hooks/cache-chart.js
```diff
@@ -0,0 +1,491 @@
...
+// kicking in. The two halves are the same trace drawn with different
+// colors, split at the divider.
+//
+// Perf notes: the step buffer and clip strings are reused across frames
+// (no per-frame allocations beyond a few numbers), and each half's path is
+// built only from the steps inside it, with coordinates clamped instead of
+// ctx.clip().
+import { onThemeChange } from "../lib/theme.js";
+
+const FALLBACK_COLORS = {
```
fortmarek: **Keep the chart palette in CSS**

These four values duplicate the existing CSS tokens (`--noora-red-500`, `--marketing-tint-red`, `--noora-purple-500`, and `--marketing-tint-purple`). `resize()` already reads their theme-resolved values from computed CSS, so please remove `FALLBACK_COLORS` and keep CSS as the sole source of truth (or define any necessary fallback in CSS). Besides being able to drift, these fallbacks encode only the light-theme palette and would produce the wrong colors if reached in dark mode.

fortmarek (approved):
Left a couple comments, but mostly looks good 👍 Great job! For full disclosure, I mostly skimmed through the PR given its size 😅 and tested it on my machine

PR #12279, opened 2026-08-12
feat(server): redesigned marketing blog index behind a feature flag

asmitbm:
Third slice of the marketing site redesign, on top of the about page PR. Redesigns the blog index with a grid/list view toggle: the preference is written to a cookie by the BlogViewPreference hook and mirrored into the session by the new TuistWeb.Marketing.Preferences plug, so the first server render is already in the visitor's preferred view. Each view gets its own page size, and all filter/page/view state lives in the query string. Also trims blog and case-study excerpts to fit the new cards. Gated behind :new_marketing_blog; the legacy index renders until the flag flips.

Screenshot 2026-08-12 at 10 03 23
Screenshot 2026-08-12 at 10 03 45


server/lib/tuist_web/marketing/live/marketing_blog_live/new/blog.html.heex
```diff
@@ -0,0 +1,191 @@
...
+            data-part="search-toggle"
+            variant="secondary"
+            icon_only={true}
+            size="large"
+            phx-click={JS.push("open_search") |> JS.focus(to: "#blog-search")}
+            aria-label={dgettext("marketing", "Search")}
+          >
+            <.search />
+          </.button>
+          <.button variant="secondary" icon_only={true} size="large" href={~p"/blog/rss.xml"}>
```
fortmarek: Could we add distinct accessible labels such as “RSS feed” and “Atom feed”? Both icon-only links currently have empty accessible names, so assistive technology cannot distinguish them.

fortmarek (approved):

PR #12282, opened 2026-08-12
fix(server): give build_runs a dedicated dedup version so processed builds replace their placeholder

fortmarek:
## What changed

`build_runs` now dedups on a dedicated `updated_at` version column instead of on `inserted_at`.

- New ClickHouse migration swaps the table to `ReplacingMergeTree(updated_at)`, keeping `PARTITION BY toYYYYMM(inserted_at)` and `ORDER BY (project_id, id)`, followed by a second migration that replays the pre-swap table.
- `Tuist.Builds.Build` gains the `updated_at` field. A first write versions the row with its own `inserted_at`; a write that replaces a known row passes that row's version, bumped, and `to_buffer_map/1` takes the later of it and now.
- `ProcessBuildWorker.base_build_attrs/4` passes the version of the row it read, so ordering does not depend on the writing pods' clocks agreeing.
- `Tuist.Builds.get_build/2` resolves duplicates with `ORDER BY updated_at DESC LIMIT 1`.

## Why

About 1% of builds were permanently displayed as `processing` with duration 0 even though `ProcessBuildWorker` had succeeded: ~4,362 rows older than an hour in the last 30 days, against ~383k `success` and ~51k `failure` in the same window. Only 18 rows were `failed_processing`, so these were not processing failures. Sampled stuck builds each had an Oban `job:stop` with `state: success` in the logs, yet ClickHouse held a single `processing` row for them.

## Root cause

A build is written twice: once as a `processing` placeholder when the CLI reports it, and again once the worker has parsed the xcactivitylog. Because `build_runs` was `ReplacingMergeTree(inserted_at)`, `inserted_at` was both the build's user-visible timestamp and the dedup version.

`ProcessBuildWorker.base_build_attrs/4` builds the processed row from the existing row via `Map.from_struct()`, dropping only associations and count fields, so `inserted_at` was carried over; `Build.to_buffer_map/1` then kept that non-nil value. The replacement was therefore inserted with the *same version* as the placeholder it was meant to supersede, so the merge picked an arbitrary survivor and the placeholder could win.

Two rows observed pre-merge in production confirmed both sides of this:

- One build had its `processing` (duration 0) and `success` (duration 56740) rows both carrying `inserted_at = 2026-08-12 07:28:08.504546` — a version tie.
- Another had `processing` at `07:28:14.238616` and `success` at `07:28:16.467776` — distinct versions, deterministic replacement. That is the `{:error, :not_found}` branch of `base_build_attrs/4`, where the worker does not find the existing row and stamps a fresh timestamp.

The tie also defeated the read path: `get_build/2` ordered duplicates by `inserted_at DESC LIMIT 1`, which cannot break a tie, so the detail page and API could return the `processing` row even before any merge.

## Why this solution

The obvious one-line fix — stamp `inserted_at = now()` on the replacement — was rejected. `inserted_at` is the build's user-visible timestamp, so moving it misreports when the build ran, and it is the partition key: a build created at 23:5x on the last day of a month whose replacement lands in the next month would sit in a different partition, where merges never dedup, leaving two rows forever.

Splitting the roles is the durable shape, and it already exists in this codebase: `runner_jobs` is `ReplacingMergeTree(updated_at)` partitioned by `enqueued_at`, precisely so its state transitions can be ordered without disturbing its timestamps.

The engine's version column cannot be changed in place, so the migration follows the established table-swap pattern (`CREATE TABLE ... _new`, `INSERT INTO ... SELECT`, `EXCHANGE TABLES`). Two details beyond that pattern:

- The copy derives the version as `if(status = 'processing', inserted_at, addSeconds(inserted_at, 1))`, so a placeholder always loses to its replacement. Every stuck pair whose two rows have not been merged away yet is repaired in that one pass, with `inserted_at` and therefore the partition untouched.
- Rows written to the old table between the copy's snapshot and the exchange are replayed by a second migration, which reads the old table whole. They can't be selected on time: a processed row carries the placeholder's `inserted_at`, so filtering on it would skip exactly the rows this change is about. The version expression is deterministic, so rows the copy already took are replayed as identical duplicates and collapse on merge. Splitting it out also means a failure there re-runs only a plain `INSERT ... SELECT`.
- The swap guards on `updated_at` already being present. A failure anywhere after `EXCHANGE TABLES` leaves the migration marked failed with the swap done, and without the guard a re-run would rebuild the table from what is by then the swapped-in copy.

The old table is left behind as `build_runs_new` so the swap can be undone, matching the previous `build_runs` swap, which dropped its predecessor in a follow-up migration. **A follow-up migration should drop `build_runs_new` once the new table is verified.**

## Backfill

The migration repairs every stuck pair still holding both physical rows. Builds whose placeholder already won a merge cannot be repaired: the processed status and duration no longer exist anywhere in `build_runs`, and neither can be derived from the side tables (`build_targets` and friends prove the worker finished but do not carry the build's wall-clock duration).

Mass re-enqueueing `ProcessBuildWorker` for those was considered and rejected. Build archives are not retained indefinitely, and a job whose artifact has expired exhausts its attempts and calls `mark_failed_build_processing/4`, relabelling a build that genuinely succeeded as `failed_processing` — a worse answer than the stale `processing`. The affected set is closed: it stops growing once this ships, and an operator can re-enqueue an individual build now that the replacement is guaranteed to win.

## Other tables written twice

Checked the other ReplacingMergeTree tables with a placeholder-then-replacement write pattern:

- `test_runs`, `test_cases` and `shard_runs` re-insert with an explicit `NaiveDateTime.utc_now()` (`Tuist.Tests`), so their versions always advance. No tie. Their version is still `inserted_at`, which means a re-insert crossing a month boundary lands in a different partition and never dedups — rare, and reads pick the newest row anyway, so it is noted rather than fixed here.
- `runner_jobs` already uses `updated_at` as its version.

`Tuist.Builds` was the only place that rebuilt a row from an existing struct and carried the version with it.

Also noted, not changed: the analytics queries in `Tuist.Builds.Analytics` read `build_runs` without `FINAL`, so an unmerged pair is counted twice and its 0-duration placeholder drags averages down. This is transient (merges collapse the pair) and pre-existing.

## Impact

Processed builds now deterministically replace their `processing` placeholder, both at merge time and on the read path. Users stop seeing builds frozen at `processing` with duration 0.

During the rollout, rows written by pods still on the old code get `updated_at` from the column's `DEFAULT inserted_at` rather than failing to insert. Versioning a first write with its own `inserted_at` rather than with `now()` matters here: a new-code placeholder paired with an old-code processed row ties, exactly as it does today, instead of the placeholder deterministically winning. Rolling the processor deployment ahead of the server one side-steps the window entirely.

The replay leaves every partition holding a duplicate of each row until merges collapse them. The build list reads through `FINAL` and is unaffected; the raw `count()` analytics in `Tuist.Builds.Analytics` see the duplicates for that window.

## Validation

- New regression test in `process_build_worker_test.exs` runs the worker against the real `create_build/1` path and asserts the two physical rows share an `inserted_at` but that the processed row's `updated_at` is strictly greater, then forces `OPTIMIZE TABLE build_runs FINAL` and asserts the merged row is `success` with the parsed duration.
- A second worker test stubs `NaiveDateTime.utc_now/0` to put the processing pod's clock an hour behind the pod that wrote the placeholder, and asserts the processed row still carries the higher version. It fails if the worker stops deriving the version from the row it read.
- New `to_buffer_map/1` tests cover the three cases: a first write versioned from `inserted_at`, a supplied version treated as a floor, and a floor ahead of the clock being kept.
- The worker tests were confirmed to fail when the version tie is reintroduced (the first failing exactly as production did, with the placeholder ranked above the processed row).
- Both migrations were replayed against a dropped-and-recreated ClickHouse test database, then re-run against the already-swapped table: the swap guard no-ops and the replay is idempotent.
- The repair and replay were exercised against a local ClickHouse with seeded data: a tied `processing`/`success` pair copied through the version expression leaves `success` as the sole survivor after merge, still in the `202608` partition; and a pair present only in the old table is recovered by the replay while a pair the copy already took collapses back to one row.
- `mix test test/tuist/builds/ test/tuist/builds_test.exs test/tuist_web/controllers/api/builds_controller_test.exs` (233 tests), plus the runs controller, build/Xcode/Gradle LiveViews, MCP and runner-jobs suites.
- `mix format`, `mix credo --strict` (no new findings), `mix excellent_migrations.check_safety` (no findings for the new migration).

Found while investigating Sentry issue TUIST-4G5; the transient Tigris 404 in that report is a separate problem and is not addressed here.


server/lib/tuist/builds/build.ex
```diff
@@ -172,6 +180,7 @@ defmodule Tuist.Builds.Build do
       %DateTime{} = dt -> DateTime.to_naive(dt)
       other -> other
     end)
+    |> Map.put(:updated_at, NaiveDateTime.utc_now())
```
esnunes: The version here is a wall-clock stamp from whichever pod writes the row, so it isn't strictly ordered against the row it replaces. The placeholder comes from a server pod and the replacement from a processor pod, so a processor whose clock sits behind by more than the processing gap would let the placeholder win again. NTP keeps that small and the download-plus-parse is normally seconds, so it's unlikely — but it could be made structural rather than probabilistic.

`base_build_attrs/4` already reads the row being replaced, so the worker knows the version it needs to beat:

```elixir
updated_at =
  Enum.max([
    NaiveDateTime.utc_now(),
    NaiveDateTime.add(existing_build.updated_at, 1, :microsecond)
  ], NaiveDateTime)
```

That needs `to_buffer_map/1` to treat a caller-supplied `updated_at` as a floor instead of always overwriting it, which is more surface than the current shape.

One related thing that might be worth a line in the description: `to_buffer_map/1` calls `utc_now()` for `inserted_at` and again for `updated_at`, so a new-code placeholder lands with `updated_at` a few microseconds above `inserted_at`, while an old-code pod omits the column and gets `DEFAULT inserted_at` carried over from that same placeholder. If I'm reading it right, a new-pod placeholder paired with an old-pod processed row has the placeholder win deterministically during the rollout, where before it was a coin flip. Short window either way, and rolling the processor deployment ahead of the server one would side-step it.

server/priv/ingest_repo/migrations/20260812120000_add_updated_at_version_to_build_runs.exs
```diff
@@ -0,0 +1,154 @@
...
+    IngestRepo.query!("EXCHANGE TABLES build_runs AND build_runs_new")
+
+    tail_copy_floor = DateTime.add(started_at, -@tail_copy_window_minutes, :minute)
+
+    IngestRepo.query!(
+      """
+      INSERT INTO build_runs (#{columns}, updated_at)
+      SELECT #{columns}, #{version_expression()}
+      FROM build_runs_new
+      WHERE inserted_at >= {floor:DateTime64(6)}
```
esnunes: This might be filtering on the wrong column — let me know if I'm missing the reasoning.

`inserted_at` isn't the write time for the rows the tail copy is there to catch, since the worker carries it over from the placeholder. So as I read it, the filter selects on the build's creation time, and a replacement written during the copy for a build that started more than ~10 minutes earlier would fall below the floor and be skipped.

`mark_failed_build_processing/4` is the case I'm most confident about: it only runs after five attempts with backoff, so its `inserted_at` is well outside the window. Delayed or retried `replace_build_run` writes look similar.

If that's right, the skipped row is the processed one — the placeholder is already in the main copy — so the build would sit at `processing` with duration 0, which is the thing this PR is fixing.

One option: drop the `WHERE` and replay the whole old table. The version expression is deterministic, so replayed rows are exact duplicates and the engine collapses them; the cost is one extra scan. If that's too expensive at this size, filtering on part write time via the `_part` virtual column against `system.parts.modification_time` would also work, and over-selecting there is harmless.

server/priv/ingest_repo/migrations/20260812120000_add_updated_at_version_to_build_runs.exs
```diff
@@ -0,0 +1,154 @@
...
+  @disable_migration_lock true
+
+  # Rows written to the old table between the copy's snapshot and the exchange
+  # would otherwise be lost. The tail copy replays a window wide enough to
+  # cover that gap plus ingestion buffers that flush on a delay; re-inserting
+  # rows already copied is harmless because the engine dedups them.
+  @tail_copy_window_minutes 10
+
+  def up do
+    IngestRepo.query!("DROP TABLE IF EXISTS build_runs_new")
```
esnunes, line 66: On the retry path: if something fails after `EXCHANGE TABLES`, the migration is marked failed and a re-run hits this line against the *old* table — which at that point holds the rollback copy and any tail rows not yet replayed. And since `column_names/1` would then read the swapped-in table, `updated_at` ends up declared twice in the `CREATE TABLE` and the re-run aborts anyway.

Two things that would cover it:

- A guard up top: read `system.columns` for `build_runs`, and return `:ok` early if `updated_at` is already present. Cheap, and it also helps if two replicas migrate at once, which `@disable_migration_lock true` otherwise allows.
- Moving the tail copy into its own migration, so a tail failure re-runs only the tail. That part is a plain `INSERT ... SELECT` and safe to repeat, and it removes the need to thread `started_at` through — which pairs with the filter comment above.

Keeping the drop of `build_runs_new` inside this migration set rather than a follow-up works too — it would just want to run after the tail copy, since the tail reads from it, and it trades away the ability to swap back.

esnunes (approved):
The diagnosis and the fix look right to me — splitting `inserted_at` (timestamp + partition key) from `updated_at` (dedup version) is the durable shape, and matching `runner_jobs` rather than inventing something new is the right call. Rejecting the `inserted_at = now()` one-liner on partition-key grounds is well argued.

Three suggestions inline. The two on the migration are the substantive ones: the tail copy's filter column, and what happens if the migration is retried after `EXCHANGE TABLES` has already run.

server/lib/tuist/builds/build.ex
fortmarek: Made it structural: `base_build_attrs/4` now passes the replaced row's version bumped by a microsecond and `to_buffer_map/1` treats a supplied `updated_at` as a floor, so ordering no longer depends on the two pods' clocks; a first write is versioned from its own `inserted_at`, which also stops a new-code placeholder from deterministically outranking an old-code processed row during the rollout, and both points are now in the description. Covered by a test that stubs the clock an hour behind.

server/priv/ingest_repo/migrations/20260812120000_add_updated_at_version_to_build_runs.exs
fortmarek: You read it right, and the case you name is the one that matters: filtering on `inserted_at` skipped exactly the processed rows. Dropped the `WHERE` and moved the replay to its own migration that reads the old table whole; at this table size the extra scan is cheap, and the duplicates collapse deterministically (noted the transient raw-`count()` effect in the description). Verified with seeded data that a pair present only in the old table is recovered and one the copy already took collapses back to a single row.

server/priv/ingest_repo/migrations/20260812120000_add_updated_at_version_to_build_runs.exs
fortmarek: Did both: the swap now returns early if `updated_at` is already on `build_runs`, and the tail moved to `20260812120001` as a re-runnable `INSERT ... SELECT`, which also removed the `started_at` threading. Re-ran both migrations against the already-swapped table to confirm the guard no-ops and the replay is idempotent. Kept the `build_runs_new` drop as a follow-up so the swap stays reversible.

Issue #12245, opened 2026-08-08
Xcode cache: a cache entry whose references are missing fails the build, permanently


esnunes:
## Summary

The Xcode compilation cache can hand the compiler a cache entry whose referenced objects are not obtainable, which **fails the build**. On the clang lane this is fatal and non-recoverable: `ClangCachingMaterializeKey` requests the result, the object cannot be produced, and the build stops. A cache layer must never be able to break a build — a degraded cache should cost warmth, not correctness.

The failure is also **not self-healing**. Once an entry is in this state it stays in this state, so retries fail identically, forever.

> [!IMPORTANT]
> **Updated 2026-08-08 (twice — please read the correction comment).** The mechanism below was written
> before the defect was isolated, and a first update comment then over-claimed that the upstream store
> broadly violates referential integrity. It does not: a value graph IS copied forward when it is
> **loaded** (`FaultInPolicy::FullTree`), and the earlier reproduction was built out of presence checks,
> which fault nothing in. The narrow upstream hazard, our own write-through, and what is actually
> established are all set out in the correction comment. Section 2 below — our write-through recording an
> association before its graph exists — is now the leading candidate, not a secondary one. #12246
> characterizes all three variants.

## Symptom

```
note: cache hit
error: CAS operation failed: missing object '0~uvz9yog…ExQ=='
…
The following build commands failed:
	ClangCachingMaterializeKey 0~sn59hOgK…PQ==
	Testing workspace Tuist with scheme TuistCacheEEAcceptanceTests
** TEST FAILED **
```

Zero targets compile; the build dies before any test runs. The Swift lane degrades gracefully on the same condition (`cache replay failed` warning + recompile) — only clang hard-fails, and it fails the whole build.

## Evidence that it does not recover

Observed on this repo's own `Acceptance Tests` job, across independent runs, days, and separate runner VMs — with a byte-identical action key and missing object every time:

| run | date | key | missing object |
|---|---|---|---|
| [30948078223](https://github.com/tuist/tuist/actions/runs/30948078223) | Aug 4 | `0~sn59…` | `0~uvz9…` |
| [31155658841](https://github.com/tuist/tuist/actions/runs/31155658841) | Aug 7 | `0~sn59…` | `0~uvz9…` |
| [31241912799](https://github.com/tuist/tuist/actions/runs/31241912799) att. 1–3 | Aug 8 | `0~sn59…` | `0~uvz9…` (+ a second object from att. 2 on) |

[...]

PR #12284, opened 2026-08-12
fix(cli): stop recording CAS associations before their value graph lands

fortmarek:
## What changed

Two write-side invariants in `cas-plugin`, so that a present value root actually means a usable value graph:

1. **The plugin records an association only when the value's root is present locally.** `actioncache_get_impl` previously wrote `key -> value` as soon as a resolve answered.
2. **The proxy publishes a value graph's root last, and only when every other node in its closure landed.** `materialize_manifest` previously stored nodes in manifest order, which puts the root first.

Both conditions are observable rather than silent: `deferred_puts` on the plugin's dispose line, and `stats_incomplete_closures` in the proxy, reported per event and in `stats_line`.

## Why

[#12247](https://github.com/tuist/tuist/pull/12247) added a read guard: a local action-cache hit has its root probed before it is served, because serving a dangling association hands the compiler `missing object '0~...'` permanently (a warning plus recompile on the Swift lane, fatal on the clang one). See [#12245](https://github.com/tuist/tuist/issues/12245).

That guard treats a dangling association as something only a size-driven store prune could create. It is not. Both sides of this crate author them during ordinary operation, with no prune involved:

- **The plugin.** Protocol v2 replies to a RESOLVE before materialization finishes, deliberately, because action-cache gets run on swift-build's serial task-setup path and must never wait on a graph fetch. Recording the association at that moment names a graph that may never arrive. Nothing can retract it afterward: the ABI has no delete at the association or object level, the only deletion primitive is the whole-store `llcas_cas_prune_ondisk_data`, and re-putting the key with a different value is refused while the original survives. That made the write-through an author of exactly the state the read guard exists to catch.

- **The proxy.** Skipping a node during materialization is a designed outcome, not an error path: `materialize_manifest` continues past a blob the server did not return or bytes that do not decode, keeping fetch instructions registered so a demand load retries. Storing the root first therefore published graphs already known to be incomplete. The shape is invisible to a reader, because the get path probes the root and nothing deeper.

## Root cause

The read probe assumes an invariant that nothing on the write side upheld. A root was published, and an association recorded, at moments when the closure behind it was known to be incomplete or not yet fetched. The guard then had to catch, at serve time, a condition both writers were free to keep creating.

Reproduced end to end on a real build over an empty store: the store began with nothing, resolves succeeded, materialization did not land, and the build finished holding 9 distinct dangling roots that only the read guard kept out of the compiler.

**Not a closure walk on the read path.** The obvious fix for "root present, child absent" is to verify the whole closure before serving. That costs a load per node, transitively, on the serial task-setup thread, to rediscover what the writer already knew at publish time. The probe stays root-only at about 12ns per served hit (measured, recorded in `tests/unbacked_local_hit.rs::probe_cost`).

**Root-only depth on the write side too, deliberately.** An interior node that never landed is still recorded and still relies on the load path's FETCH_OBJECT. Matching the read guard's depth is what makes the two sides describe one invariant instead of two overlapping partial ones.

**Deferring the put, rather than retrying or blocking.** Blocking the resolve until materialization completes would reintroduce exactly the stall protocol v2 removed.

## Impact

- One extra resolve per key, once. The materialized root persists on disk, so the next build's resolve finds it present and records the association then.
- No cost on the warm snapshot path, where the graph is already materialized when the resolve is answered, so the probe passes and behavior is unchanged.
- The current build loses nothing in either case: a load of a withheld root goes to FETCH_OBJECT, whose instructions stay registered for every skipped node.
- A `deferred_puts` or `stats_incomplete_closures` count that stays high across builds signals that graphs are not landing at all, which is a materialization or auth failure rather than a cache one. Persistently non-zero closures specifically means the remote is serving entries whose blobs it cannot produce, which is kura's cascade to fix rather than this crate's.

## Review follow-ups

Three findings from review, each fixed in the second commit:

- **A root failing on its own was not counted.** The outcome was decided at the root's turn in the materialization loop, which only saw the case where a child had been skipped. If the root's own blob was missing, undecompressible, or undecodable, its skip incremented the counter after the check had already run and the loop ended, so nothing was counted or logged. The root was still correctly absent, so no dangling association could be authored, but the outcome was silent. It is now reported after the pass, from whether the root actually landed, and the two reasons log distinctly: `incomplete closure, root withheld` when a child was skipped, `incomplete closure, root unavailable` when the children landed and the root did not. The second means retrying is pointless until the remote can serve that root. The skip counter now counts other nodes only.
- **`stats_incomplete_closures` was missing from `stats_line`.** The cumulative signal documented as the thing to watch had no reader outside the per-event log, so it could not be seen in the proxy's periodic stats output. Included now.
- **Expected deferrals were labeled as degradation.** The field documentation says a deferred put is not a degradation, yet any nonzero value produced a `degraded:` line, so a cold build would report degradation while behaving exactly as designed. `deferred_puts` now logs on its own line, leaving `degraded:` to unbacked hits and poisoned puts.

## Validation

Five new tests, plus the full crate suite green at 81 tests.

Proxy (`src/proxy.rs`):
- `an_incomplete_closure_withholds_its_root`: a manifest whose child bytes are not a decodable frame leaves the root unpublished and counts the closure.
- `a_complete_closure_publishes_its_root`: the complementary case, so the withholding cannot be unconditional (otherwise no graph would ever become locally resolvable and every build would re-resolve every key).
- `a_root_that_fails_on_its_own_is_still_reported`: the inverse manifest, a sound child with an undecodable root. Verified this fails against the previous in-loop reporting (counted 0, expected 1) and passes with the fix.

Integration (`tests/unbacked_local_hit.rs`), driving the real plugin over a scripted proxy:
- `a_resolve_hit_whose_graph_never_arrived_records_no_association`: the resolved value is still served, but no association is written. Proved by restoring the root, answering the next resolve with a MISS, and observing a second resolve is required rather than a hit from a record naming nothing.
- `a_resolve_hit_whose_graph_is_present_is_recorded`: when the graph is already present at resolve time, the association is recorded and answers later gets with no round trip.

The crate has no cargo fmt or clippy gate in CI (the only Rust lint gate is `codebase-search.yml`, which does not cover this crate) and its base is not rustfmt-clean, so no blanket reformat was run here. Note that `cli.yml` does include `cas-plugin/**` in its trigger paths, so CLI build, test, and acceptance jobs do run on changes here.


esnunes (commented):
This is a good fix, and the description does the rare thing of explaining why the read guard alone couldn't hold: the writer knows whether the closure landed, while the reader would have to rediscover it a load at a time on the serial task-setup thread. Moving the obligation to the write side, at the same root-only depth as the guard, is what turns two overlapping partial invariants into one. I also liked that both outcomes are counted rather than silent, and that `deferred_puts` is deliberately kept out of `degraded:`.

The one I'd most like your read on: `fetch_object` is a second door into the same store, and it stores whatever single node it's asked for. Right after a withhold the compiler loads the root, that path stores it standalone, and the root is present again with the hole intact — so a later probe passes and the association gets written anyway. I tested the two halves I could: the load of the served value does produce a FETCH_OBJECT for the root digest, and `fetch_object` stores a root whose referenced child is absent. It isn't a regression, since that association was written unconditionally before this PR, but I think it means the withhold buys a delay rather than the invariant AGENTS.md states. Related: while the proxy stays alive, a withheld closure is never re-materialized, because the resolve fast path treats a still-registered fetch instruction as presence.

The other two are smaller: a blind spot in the new counter when the root is already on disk, and a wrong denominator in the new log line.

I ran the crate suite locally and got 81 green, and confirmed the two new integration tests actually execute here rather than taking their fixture's early return.

Separately, in case it saves you a look: the red Unit Tests job isn't this change. It fails on `missing object '0~uvz9yogbl...'`, and the run on main from the same morning fails with the byte-identical digest, so it reads as a poisoned runner cache volume rather than anything the branch did.

cas-plugin/AGENTS.md
```diff
@@ -30,7 +30,9 @@ The crate builds two artifacts from shared modules: the `cdylib` (`libtuist_cas_
...
... be incomplete. Ordering it last makes a present root mean a complete closure FOR THIS AUTHOR, which is exactly what the read guard tests: an incomplete closure leaves the root absent, the plugin records no association, and the next build resolves again and re-materializes. The current build loses nothing — its load of the root goes to FETCH_OBJECT, whose instructions stay registered for every skipped node. `stats_incomplete_closures` counts it, on `stats_line` as well as per event, and the outcome is decided AFTER the pass rather than at the root's turn in the loop: the root can also fail on its own terms (its blob missing, undecompressible, or undecodable) and nothing runs after it, so deciding at its turn left the store's most consequential outcome silent. The two reasons log distinctly — `incomplete closure, root withheld` when a child was skipped, `incomplete closure, root unavailable` when the children landed and the root did not, which is worth separating because the second means retrying is pointless until the remote can serve that root. Persistently non-zero means the remote is serving entries whose blobs it cannot produce, which is kura's cascade to fix, not this crate's.
```
esnunes: I wanted to check one thing about this bullet with you, since I may be missing something.

Inside `materialize_manifest` the ordering does exactly what it says: a child miss withholds the root. What I'm less sure about is whether the root *stays* withheld, because `fetch_object` is a second writer into the same store and it has no closure check — it stores whatever single node it's asked for (proxy.rs:1908-1914), and the plugin's load path calls it on every local miss (lib.rs:888-896).

I traced and tested the steps:

- The get returns SUCCESS and hands the compiler the root's id even when the put is deferred (lib.rs:1117).
- Driving the real plugin over the integration harness, the load of that served value produces exactly one FETCH_OBJECT op, and it asks for the root digest.
- Against a real store, calling `fetch_object` for a withheld root whose frame *references* the skipped child returns `Ok(true)`: `llcas_cas_store_object` accepts a node whose referenced child is absent. `llcas_cas_get_objectid` only interns a digest, it doesn't require presence — which is also why storing the root first could produce the nine dangling roots in the first place.

So the sequence I end up with is: root withheld and counted → the compiler loads the root → FETCH_OBJECT stores it standalone, from the inlined bytes the withhold left behind in `pending_objects` → the root is present again with the hole intact. On a later resolve the probe in `actioncache_get_impl` passes, and the association is recorded over a closure still missing that child.

The one step I haven't verified myself is that the compiler always issues that load. This bullet's own "the current build loses nothing — its load of the root goes to FETCH_OBJECT" is exactly that assumption, which is what made me think the two are the same load.

To be clear, none of this is a regression: that association was written unconditionally before this PR, so the change is still strictly an improvement. It's the "a present root means a complete closure" claim I'd narrow — as far as I can tell it holds until the first demand load of the root. A closure check inside `fetch_object` would cost exactly the walk you deliberately avoided, so I suspect the wording is the thing to adjust rather than the code. Does that match what you saw?

cas-plugin/src/proxy.rs
```diff
@@ -1605,7 +1611,43 @@ impl Proxy {
...
+            let root_pending = missing.iter().any(|entry| is_root(entry));
+            let mut root_stored = false;
+            // Counts the OTHER nodes only. A root that fails on its own is not
+            // evidence about its children, and reporting it as one of them would
+            // misdescribe which side of the graph the remote could not produce.
+            let mut skipped = 0usize;
+
+            for entry in ordered {
+                let entry_is_root = is_root(entry);
+                if entry_is_root && skipped > 0 {
```
esnunes, line 1656: A question about the repair loop this describes.

`commit_and_materialize` inserts `resolved[key] = Hit(root)` and registers the pending instructions before materializing (proxy.rs:1494-1520). When materialization then withholds the root, the next resolve for that key takes the fast path at proxy.rs:1353, whose presence test is `load_present(value) || fetchable(value)` — and `fetchable` is true precisely because the withheld root keeps its `pending_objects` instruction. So it returns Hit and never enqueues another materialize job, and `pending_objects` is explicitly kept across invalidation (proxy.rs:712).

The fast path's own comment justifies the generosity with "the materializer is filling it in", which was true before but isn't for a withheld root — there the materializer already ran and gave up.

Since the proxy is long-lived across builds, I think the AGENTS.md line "the next build resolves again and re-materializes" only holds for the resolve half: the plugin does resolve again, but nothing re-fetches the skipped child until the proxy restarts. Worth saying that this is repair latency rather than a safety problem — nothing dangling gets published down this path.

Is the demand path meant to carry the repair here? If so I'd just soften the wording; if not, re-enqueueing materialization when the fast path serves a value whose root isn't actually present would close the loop the comment describes.

cas-plugin/src/proxy.rs
```diff
@@ -1605,7 +1611,43 @@ impl Proxy {
...
+            // registered for every node skipped here.
+            let root_digest = manifest.first().map(|entry| entry.llcas_digest.clone());
+            let is_root = |entry: &ManifestEntry| Some(&entry.llcas_digest) == root_digest.as_ref();
+            let mut ordered: Vec<&ManifestEntry> =
+                missing.iter().copied().filter(|entry| !is_root(entry)).collect();
+            ordered.extend(missing.iter().copied().filter(|entry| is_root(entry)));
+            // Whether the root is ours to publish at all. A root already present
+            // locally is not in `missing`, and its absence from the loop is not a
+            // withheld closure.
+            let root_pending = missing.iter().any(|entry| is_root(entry));
```
esnunes, line 1647: One thing I wanted to flag about the new counter. This one hangs on the demand-path note I left on the AGENTS.md materialization bullet — if that reading is wrong, what follows is only reachable on a store poisoned before this change, which is a lot less interesting.

Both the withhold and the count key off `root_pending`, which is only true when the root is in `missing` — and `missing` filters out anything `is_local` already answers yes for (proxy.rs:1542), with `is_local` falling back to an authoritative on-disk `load_present` (proxy.rs:2009). So if the root is already on disk from an earlier pass and a child in the same manifest is skipped, the closure is incomplete, the root stays present, and `stats_incomplete_closures` stays 0 with neither log line firing.

I reproduced that part by running `materialize_manifest` twice over a temp store — the root alone, then the root plus an undecodable child. The child stayed absent, the root stayed present, and the counter read 0.

If the demand-path reading holds, entries created *after* this PR reach the same state rather than just old ones: build 1 withholds the root and counts it, the demand load stores that root on its own, and a later pass over the same key finds the root already local, skips the child again, and counts nothing. Which would make the counted case the one where nothing bad can happen yet, and the uncounted case the one where the probe passes and the association gets written over the hole.

Would a third reason be worth having — counting when a child is skipped and the root was already local — or is that deliberately left to the prune's side of the story?

cas-plugin/src/proxy.rs
```diff
@@ -1683,6 +1731,37 @@ impl Proxy {
...
+            // root, which left the store's most consequential outcome silent.
+            if root_pending && !root_stored {
+                state.stats_incomplete_closures.fetch_add(1, Ordering::Relaxed);
+                let root_hex = root_digest
+                    .as_deref()
+                    .map(crate::analytics::hex_upper)
+                    .unwrap_or_default();
+                if skipped > 0 {
+                    crate::log_line(&format!(
+                        "incomplete closure, root withheld: root={} skipped={} of {}",
```
esnunes, line 1772: Small one on the new log line: `skipped` only ever counts non-root entries drawn from `missing` (proxy.rs:1661-1672), but it's printed against `manifest.len()`. On a graph where most nodes are already on disk that reads as `skipped=1 of 40` when the pass was only fetching two nodes — which under-reads the failure rate for anyone using this as the per-event half of the trend signal. `missing.len()` would match the numerator.

esnunes (approved):

cas-plugin/AGENTS.md
fortmarek: Your reading matches what I found, and the trace holds: `fetch_object` stores whatever single node it is asked for with no closure check, and a withheld root keeps its `pending_objects` instruction including the inlined bytes, so the demand load puts the root back standalone without even a round trip. The wording was the thing that was wrong, so I have scoped it in both AGENTS.md and the comment: ordering the root last is a property of the MATERIALIZER, not an invariant of the store, and it holds until the first demand load of that root. What it buys is that this crate stops creating the state unconditionally, not that the state is unreachable. I kept the code as is for the reason you name, a closure check in `fetch_object` costs exactly the transitive walk the root-only probe exists to avoid.

cas-plugin/src/proxy.rs
fortmarek: Confirmed, and yes, the demand path is meant to carry the repair, so I have softened the wording rather than adding a re-enqueue. `fetchable` is true for a withheld root precisely because it keeps its instruction, so the fast path returns the cached Hit and no second materialize job is queued, and the fast path's "the materializer is filling it in" is indeed not true in that case. What does still repair is per object and on demand: every skipped node keeps its fetch instructions, so the load that actually needs one fetches it, which is the part the graph-level re-materialization claim was standing in for. AGENTS.md now says repair is per object rather than a re-run, and calls out explicitly that a long-lived proxy will not re-materialize on the next build, as repair latency rather than a safety hole since nothing dangling is published down that path.

cas-plugin/src/proxy.rs
fortmarek: You are right, and a third reason was worth having, so I added it. `missing` filters out whatever `is_local` answers yes for, so a root already on disk never enters the loop, there is nothing to withhold, and the report keyed off the withhold stayed silent. That inverted the signal exactly as you describe: the counted case was the safe one and the uncounted case was the one where the probe passes and the association is written over the hole. It now logs `incomplete closure, root already local` and increments the counter, and since it is the shape the read guard cannot catch, that makes it the one most worth counting rather than the one to leave to the prune's side of the story. Covered by a new test that materializes the root alone, then re-materializes with an undecodable child, asserting the root stays present and the condition is still counted.

cas-plugin/src/proxy.rs
fortmarek: Fixed, both log lines now count against `missing.len()`. You are right that it under-reads the failure rate: the numerator only ever draws from `missing`, so measuring it against the whole graph reads as `skipped=1 of 40` when the pass was only ever responsible for fetching two nodes, which is misleading for exactly the trend use this counter is for.

PR #12293, opened 2026-08-12
fix(infra): grant a warm Pod to runner pools whose queued work is blocked

fortmarek:
## What changed

A runner pool whose only queued work belongs to an account sitting at its
concurrency limit no longer reads as idle to the autoscaler. It now asks for
exactly one warm Pod.

Two coordinated changes:

**Server.** `dispatchable_queued_count/1` in [runners.ex](server/lib/tuist/runners.ex)
already computed how many queued jobs it withheld from the `queued` signal, but
threw that number away into a telemetry event. It now returns
`{dispatchable, withheld}`, and `scaling_signals_for_fleet/1` surfaces the
second value as a `withheld` key. The two `:telemetry.execute` calls collapsed
into one on the shared value, so the metric and the signal cannot drift apart.
The endpoint at [runners_controller.ex](server/lib/tuist_web/controllers/runners_controller.ex)
serializes the whole signals map, so the new key rides along without any
serializer change.

**Controller.** New `Signals.BlockedDemand()` in
[desired.go](infra/runners-controller/internal/scaling/desired.go) returning
`min(1, Withheld)`, folded into `Signals.Load()`. That single site covers both
consumers: the reconciler already fed `sig.Load()` into `PoolDemand.Load`, so
blocked pools now enter tier 1 of `AllocateFleet` with no change to the
allocator's logic. Only its doc comments changed, since they still described
the old `occupied + queued` formula.

Neither the server-side withholding nor the account concurrency limits are
touched. This changes only how a blocked pool is *classified*, not how much
work it is allowed to run.

## Why

On a saturated fleet, capacity-held pools consume the whole allocation and the
discretionary tiers are empty. A pool reporting `load == 0` competes only in
those tiers, so it gets nothing. But a pool with withheld work is not idle: it
has real jobs it cannot serve yet. Without a Pod polling dispatch it cannot
claim the instant headroom frees, while siblings that already hold warm Pods
take that headroom within seconds. The pool stays at zero and does not recover
on its own.

### Root cause

The withholding itself is correct and stays. Sizing on raw queue depth
provisions Pods for jobs dispatch will refuse, and those Pods then idle on
hosts that pools with claimable work are denied. The defect is that the capped
signal is lossy: after capping, "nothing queued" and "queued but currently
unservable" both arrive at the controller as `queued: 0`, and the controller
has no way to tell them apart. The server had the distinguishing number the
whole time and was dropping it.

### Why one Pod rather than the full withheld count

Scaling with `Withheld` would recreate exactly the idle-Pod squatting the
withholding exists to prevent, just one layer further down. One Pod is
sufficient: it is enough to enter the race for headroom the moment it frees,
and it is the smallest grant that changes the pool's classification from idle
to demanding. Hence `min(1, Withheld)` rather than a proportional term.

The alternative considered was special-casing blocked pools inside
`AllocateFleet`. Putting the fix in `Load()` instead means the allocator's
tier semantics stay untouched and every existing caller inherits the behavior
through the one function they already call.

## Rollout safety

The two sides deploy independently, in either order. An older server omits
`withheld`, it decodes to `0`, and `min(1, 0)` is `0`, so `Load()` is
byte-for-byte what it is today. This is pinned as a test case rather than
argued in a comment.

## Impact

Blocked pools recover on their own instead of waiting for an operator. The
surplus grant goes Pending on a full fleet, which is the existing "add a host"
signal, and the reconciler reaps an idle sibling Pod after cooldown. Since the
grant saturates at one Pod per pool, the worst-case added pressure is one Pod
per blocked shape, not one per blocked job.

### How to test locally

Go, from `infra/runners-controller`:

```bash
go test ./... && go vet ./... && gofmt -l .
```

Elixir, from `server`:

```bash
mix test test/tuist/runners_test.exs test/tuist_web/controllers/runners_controller_test.exs
```

## Validation

- `go test ./...` passes across all 7 controller packages; `go vet` and `gofmt` clean.
- `mix test` on both server test files: 67 pass. `mix format` and `mix credo` clean.

The allocator test reproduces the production incident to its numbers: 9 host
slots fully held by one shape, two other shapes each holding a single withheld
job. Both blocked pools come out allocated 1, and the saturated shape keeps its
full 9. A companion case feeds the same fleet with `Withheld: 0`, which is what
the server reported before this change and what an older server still reports
mid-rollout, and pins that both pools get 0, so the starvation cannot silently
return.

Also covered: 12 and 40 withheld jobs still buying exactly one Pod; the cap
holding when capacity is not the constraint; blocked demand not displacing a
sibling's real load; and a pool already at `MaxReplicas` not being lifted.

## Note for the reviewer

Branched before [#12340](https://github.com/tuist/tuist/pull/12340), so the
first CI run failed the Security check on the `nanoid` advisory. That fix is
now merged in from main rather than duplicated here; no dependency change of
my own is in this diff.


esnunes (approved):
Nice change — the write-up made the failure mode easy to follow, and putting the fix in `Load()` rather than special-casing blocked pools inside `AllocateFleet` looks like the right call: the allocator keeps exactly three tiers, and both consumers inherit the behaviour through the one function they already call. Pinning the mid-rollout case (absent field decodes to 0) as a table test rather than arguing it in a comment is a nice touch too.

I walked the allocator arithmetic by hand for each new table case and the expected values line up, including the incident coming out 9/1/1 on 9 slots, and `headroom_jobs/2` ending in `max(0)` means `withheld` can't go negative.

Two small non-blocking notes, both in files this PR doesn't touch, so I couldn't leave them inline:

**`infra/runners-controller/internal/metrics/metrics.go:197`** — Two stragglers of the same formula you already updated in `allocate.go` and `desired.go`: `RecordAllocation`'s doc here still says "`load` is occupied+queued", and `warmDeficit` at metrics.go:323-324 repeats "real load (occupied+queued)". Both now receive a `load` that includes the blocked-demand Pod, and `warm_deficit_replicas` is derived from it. No behaviour change, just the comments.

**`infra/runners-controller/AGENTS.md:37`** — Still describes tier 1 as "real load (`occupied + queued`)", which this change makes incomplete — and the root CLAUDE.md asks for the nearest AGENTS.md to be kept current.

The part I'd care more about is the "Starvation vs saturation" runbook lower down (~223-243). It enumerates the reachable steady states and leans on `queued` counting only dispatchable work — "the second state should be impossible". This PR deliberately creates a third one that isn't listed: `queued == 0` with a single idle Pod, on a pool holding withheld work. Someone following that section during an incident would read that Pod as an anomaly rather than the fix working. A couple of sentences there would probably pay for themselves.

PR #12295, opened 2026-08-12
fix(server): close runner sessions whose Pod is gone

fortmarek:

## The failure

A `runner_sessions` row with `ended_at IS NULL` is not only a billing record. `RunnerSessions.occupied_counts_per_fleet/0` counts it as an occupied host, so a row that never closes withholds a runner from every pool sharing that host for the full six hours of the billing safety clamp.

This started from a report of seventeen open sessions on a nine-mini fleet. Measuring it turned out to be the important step, because it is not an incident — it is a standing defect running at roughly one session in four:

| | 2026-08-14 |
|---|---|
| Sessions started | 1185 |
| Never closed | **264 (22.3%)** |
| linux-2vcpu-8gb | 140 / 664 (21.1%) |
| macos-26-6 | 78 / 327 (23.9%) |
| linux-4vcpu-16gb | 45 / 184 (24.5%) |

The same rate holds on every working day back to at least 2026-07-07, and it was still leaking while this PR was being written. Each leaked row reads to the allocator as an occupied host for six hours, so a busy afternoon can withhold more hosts than the fleet physically has — and each one also bills at the six-hour clamp instead of its real duration.

## Root cause

The pod-stopped ingestion path already existed and was already wired. `PodLifecycleReconciler` watches runner Pods and POSTs `/api/internal/runners/pods/stopped` on a terminal transition, which is what closes the session. It had a hole rather than a gap:

When the Pod was already gone from the apiserver, `Reconcile` cleared its dedup entry and returned without reporting. The comment justified this by pointing at the server-side max-lifetime clamp in `Tuist.Runners.Billing` — which is a sound argument about billing and says nothing about capacity, where six hours is the whole problem.

That branch is where the 22% goes. `PodLifecycleReconciler` races `RunnerPoolReconciler`, which reaps terminal Pods; both wake on the same watch event into separate workqueues with no ordering between them, so whichever drains first decides whether the object still exists when we `Get` it. The leak rate tracks that directly — **0% in hours with under ten sessions, 8-36% once volume passes ~40/hour**.

Three findings pin it down, and two of them contradict what the code comment used to claim:

- **The Pod's termination grace period never applies.** Grace bounds how long the kubelet waits for *live* containers. A Pod that has already reached Succeeded or Failed has nothing left to terminate, so the reap's `Delete` removes it in milliseconds regardless of the 30s default. The old comment blamed idle/Pending Pods instead — but a Pending Pod never claimed and has no session, and all 264 leaked rows carry an `executed_workflow_job_id`, so every one of them ran a real job.
- **It is not a delivery failure.** A full day of controller logs carries zero `report pod stopped; will retry` lines. The report was never attempted, rather than attempted and lost — which is exactly why the `NotFound` branch is the right place to fix it.
- **Controller restarts reach the same branch** (`reported` is in-memory, and the informer only replays objects that still exist) **but are rare by comparison** — the leak shows no spikes at deploy boundaries.

`PodClaimReconciliationWorker`'s own moduledoc had already documented this exact miss for `runner_claims` ("an edge, skipped entirely when the reaper deletes the Pod before the lifecycle reconciler observes it ending") and grown a level-triggered worker in response. Sessions never got the equivalent.

Separately, the leak detector that should have surfaced this was defined and registered but produced no series. `tuist_runners_session_clamped_count` was emitted from inside `p95_concurrent_last_hour/1`, which only runs for fleets the autoscaler happens to be polling, on whichever replica serves that request — and `Tuist.PromEx.StripedPeep.scrape/1` flushes storage on every scrape. The sample had to land in the same window on the same pod to be exported at all. It never did.

## What changed

**1. Close the session before the reap deletes the Pod** (`runnerpool_controller.go`)

`reapRunner` deleted the Pod as its first statement. It now reports first:

```go
func (r *RunnerPoolReconciler) reapRunner(ctx context.Context, pod *corev1.Pod) error {
	r.reportStopped(ctx, pod)
	if err := r.Delete(ctx, pod); err != nil && !apierrors.IsNotFound(err) {
```

One goroutine, ordered — the race is **removed**, not caught afterwards. All six reap call sites funnel through `reapRunner`, and because the Pod is still in hand the report carries the real `finishedAt` rather than an upper bound.

A failed report still proceeds to the delete. Blocking the reap on the Tuist server would leave terminal Pods and their ServiceAccounts accumulating in the namespace, which is the failure the reap exists to prevent and worse than a late close — and the reconciler arm below already covers what this drops, with a better `ended_at` than this path could produce.

An earlier revision of this PR instead added a `reconcileVanished` branch that reported `now()` when the Pod was already gone. That compensated for the race rather than resolving it, and once the ordering is fixed it would only fire for Pods the controller never reaps — the same class the reconciler arm handles, with a worse timestamp. It is removed.

The pod-lifecycle reconciler still reports on the terminal transition; that is the earlier and equally accurate path, and the two are idempotent and MIN-biased server-side, so a double report cannot extend a billed window.

Neither edge covers a Pod the controller never reaps — node loss, eviction, drain, a manual delete — or one that ended during a controller restart. Hence:

**2. A failsafe: session arm on `PodReconciliationWorker`** (1-minute cron)

Level-triggered: it asks the one question none of the edges depend on, does the Pod exist, and corrects the difference.

This started as its own worker and folded into `PodClaimReconciliationWorker`, which is renamed here. Both were issuing the identical cluster read — same selector, same namespace, same cron, same kill switch — twice a minute to answer the same question, and a second reconciler for one fact is how this family of workers grew in the first place. Now it is one apiserver LIST per tick and two corrections off that single observation, with the complete-read and non-empty-result guards protecting both arms instead of being duplicated.

Sessions need their own arm rather than riding the claim's: a claim is released when its job completes while the session stays open through teardown, so the two are not gone at the same moment.

This is explicitly subordinate to the ordering fix. It catches only Pods the controller never reaps — node loss, eviction, drain, a manual delete — plus a report that failed and proceeded to the delete anyway. Controller restarts are *not* in that set: nothing reaps while the controller is down, so the Pod survives to be reported when it returns.

Closing a live runner's session under-bills a customer, so the guards bias toward doing nothing:

1. Any API error aborts the tick — a partial read is indistinguishable from mass absence
2. Zero Pods returned while sessions are open is treated as a bad read, not an empty fleet
3. Sessions younger than 15 minutes are never considered; the row is written at claim-win, before the Pod exists to be listed
4. At most 100 closes per tick, oldest first, with the overflow logged rather than silently trickled

The `runner_pod_reconciliation_paused` kill switch now stops both arms, since they act on the same observation.

**The session arm deliberately has no consecutive-absence confirmation**, unlike the claim arm's `pod_missing_since` clock. An earlier revision of this PR copied that mechanism; it does not pay for itself here. `K8sClient.list_pods/2` issues an unpaginated GET with no `resourceVersion=0`, so it reads from etcd at quorum rather than the watch cache — a live Pod transiently missing from one read is close to a non-event. And the realistic way a live Pod leaves this selector is a label change during a rollout, which is *persistent*: those Pods stay absent on every subsequent tick, so requiring consecutive absence would delay that bad close by five minutes rather than prevent it. The consequences differ too — over-releasing a claim oversubscribes real hosts, while over-closing a session under-bills and dips occupancy the claim side of `occupied_counts_per_fleet/0` largely still covers. Dropping it removes a table column, a partial index, four functions, and makes orphan recovery five minutes faster.

**3. Close at the job's real end, not at the clamp**

The obvious `ended_at` is `LEAST(now, started_at + 6h)` — the instant billing was already clamping the open row to. That is exactly billing-neutral, which fixes capacity and leaves the invoice wrong in both directions: the reporting account's 101 month-to-date orphans would drain at 36,360 minutes against roughly 1,400 minutes of real work, and every new orphan would bill a floor of the absence threshold regardless of whether its job ran for two hours or ninety seconds.

So each tick resolves its batch against `Jobs.terminal_completions/1` first — one ClickHouse query, bounded by the same per-tick cap, returning `completed_at` for jobs whose latest state is terminal:

- **Both predicates in that lookup matter.** The `completed_at` gives the honest end. The *status* check is what stops a stale completion from closing a session that belongs to a live re-claim: a retry inserts a new session row per claim while `runner_jobs` is keyed on `workflow_job_id` alone, so a job that went `completed` and then back to `queued` → `running` must not hand its old completion to the Pod running now.
- **`executed_workflow_job_id` outranks the claim-time id.** A runner can be handed a sibling's job; it is the job GitHub proved ran, and whose completion released the Pod, that dates the session's end.
- **The write is `GREATEST(started_at, LEAST(completed_at, now, started_at + 6h))`.** `LEAST` against `now` preserves the under-bill bias, so a late or bogus `completed_at` can never extend the window. `GREATEST` against `started_at` floors it — GitHub's clock and ours can disagree, and a completion preceding `started_at` would otherwise write an interval the billing query reads as negative time.
- **The fallback is byte-identical to the clamp.** A job that never reached a terminal state closes there, which is the correct conservative answer when there is no evidence of a real end. So does the whole batch if ClickHouse is unavailable, so the capacity fix never waits on the billing refinement.

This matters at the scale of the backlog: 7,092 rows are currently charged 42,552 clamped hours between them. Closing them at their jobs' real completions rather than at the clamp is the difference between correcting the record and rubber-stamping it.

`Jobs.terminal_completions/1` uses the `argMax(col, updated_at) GROUP BY workflow_job_id` pattern the module's moduledoc mandates rather than `FINAL` — same logical view, no per-read part merge, and the bounded id list is served straight off the table's `ORDER BY (workflow_job_id)`.

**4. Export the leak detector**

Moved to a `Polling.build` alongside the other gauges, with the seen-set zero-drain so a fleet that loses its pool and its clamped rows in one step does not leave `last_value` pinned at a leak alert nobody can clear.

No dashboard changes here — `runners.json` is untouched. `tuist_runners_session_clamped_count` and `tuist_runners_recovery_count{kind="orphaned_runner_session"}` are both exported and queryable; panels for them belong in a separate change against the Git Sync'd dashboard.

## Reviewer notes

- **No migration.** After dropping the absence clock this PR adds no schema.
- **Oban worker rename.** `PodClaimReconciliationWorker` becomes `PodReconciliationWorker`, so a job the cron inserted under the old name and did not execute before the roll fails once as an unknown worker. `max_attempts: 1` discards it and the next tick runs under the new name.
- **Why no finalizer.** A finalizer would make the report unmissable, since the object could not leave etcd until we removed it. The measurement says it is not worth the trade: the class it uniquely fixes is the controller-restart tail, which shows no spikes at deploy boundaries, while the reap race it does *not* need to fix is the whole 22%. Against that, macOS runs one VM per mini and a Pod stuck Terminating holds its node slot, so a controller outage would convert a billing leak into a capacity wedge. Worth noting these Pods already carry `tart-kubelet.tuist.dev/vm-cleanup`, and the fleet has already been bitten by it once ([reconciler.go:196](../blob/main/infra/tart-kubelet/internal/podagent/reconciler.go#L196)).
- **What nothing here fixes.** A Pod can still vanish with no edge to catch it — node loss, eviction, drain, manual `kubectl delete`. Only a level-triggered check covers those, which is what the reconciler arm is for.
- **New traffic.** `reapRunner` now POSTs once per reaped Pod, including warm Pods that never claimed. Those return 204 (`{:ok, :no_open_session}`) and it is bounded by pod churn, but it is new load proportional to fleet turnover, and it sits inline in the reap loop.
- **No backfill step, and this is the immediate reason the failsafe ships.** Production holds **7,092 orphaned rows across 5 accounts, oldest 2026-06-01, currently billing 42,552 clamped hours**. The ordering fix cannot touch them — those Pods are already gone. They drain through the reconciler at 100/tick in a little over an hour, at each job's resolved completion where there is one and at the clamp otherwise.
- **The failsafe's steady-state value is unproven, deliberately.** Once the backlog has drained, `tuist_runners_recovery_count{kind="orphaned_runner_session"}` *is* the residual the edges miss. If it sits at zero the session arm is dead code and should be deleted; the claim arm stays regardless, since it predates this work and covers a documented ten-day claim-stranding bug.
- **Out of scope on purpose.** The 6-hour occupancy window in `occupied_counts_per_fleet/0` is unchanged — narrowing it is an allocator policy call, not a bug fix, and with the reaper running the window stops mattering.
- **Interaction with #12031.** That PR collapses the ClickHouse/Postgres split; this one lives in the Postgres/Kubernetes seam it leaves open, so neither substitutes for the other. Once it lands, `terminal_completions/1` becomes a `LEFT JOIN` against `runner_workflow_jobs` instead of a cross-store lookup, and the degrade-to-clamp rescue goes away. One trap for that port: `cancelled` is `status = "completed"` with `conclusion = "cancelled"` in ClickHouse but a first-class status in `runner_workflow_jobs`, so a literal port of the `"completed"` predicate would silently drop cancelled jobs — the exact class most likely to lose its pod-stopped report.

## How to verify it worked

`tuist_runners_session_clamped_count` should trend to zero and stay there. The direct check, once this has been running for a day:

```sql
SELECT fleet_name, count(*) AS sessions,
       count(*) FILTER (WHERE ended_at IS NULL) AS still_open
FROM runner_sessions
WHERE started_at >= now() - interval '1 day' - interval '6 hours'
  AND started_at <  now() - interval '6 hours'
GROUP BY 1;
```

`still_open` should be a handful per fleet rather than 21-25%. Not zero, though: `started_at` is claim-win and the bound is six hours, so a job running near GitHub's own six-hour ceiling crosses it while its Pod is alive and the reconciler is correctly leaving it alone. Judge the trend, not the absolute.

Anything above a handful means the reconciler arm is carrying load the ordering fix was supposed to take, which shows as a sustained `tuist_runners_recovery_count{kind="orphaned_runner_session"}` rate.

### How to test locally

```bash
cd server && MIX_ENV=test mix test test/tuist/runners/ test/tuist/oban/ test/tuist/runners_test.exs
```

```bash
cd infra/runners-controller && go test ./...
```

Coverage added: 3 Go tests on the reap ordering (session closed before the Pod is deleted and with the real `finishedAt`, Pod still reaped when the report fails, Pod still reaped when no sessions client is wired), 28 worker tests across both arms (every guard, claim release and consecutive-absence behaviour, completion-vs-clamp close timestamps, executed-over-claimed preference, single-query batching, ClickHouse-failure degradation, occupancy drop, recovery telemetry, kill switch), 6 `terminal_completions/1` tests including the re-queued-after-completing guard, 12 sessions-query tests covering all four `ended_at` bound cases plus close idempotency and oldest-first ordering, and 3 PromEx tests including the zero-drain. Full local run: 651 Elixir tests and all Go packages passing, with `mix credo --strict` and `mix format` clean.


server/data-export.md
```diff
@@ -61,7 +61,7 @@ Sensitive authentication data (passwords, tokens) are excluded from exports.
...
...to the job that actually ran), `fleet_name`, `pod_name`, `runner_name`, `repository` (denormalized `owner/name` handle from the workflow_job for billing-page scope filters), `workflow_name` (denormalized for the same), `started_at` (claim-win — proxy for Pod creation), `ended_at` (normally set by the runners-controller via `POST /api/internal/runners/pods/stopped` when it observes the Pod's container terminate; NULL while still in flight. When that report is missed, `Tuist.Runners.Workers.OrphanedRunnerSessionsWorker` closes the row once the Pod is confirmed absent from a complete cluster read. It prefers the workflow_job's terminal `completed_at` from ClickHouse `runner_jobs` — the runner's real end — bounded as `GREATEST(started_at, LEAST(completed_at, now, started_at + 6h))`; when the job never reached a terminal state it falls back to `LEAST(now, started_at + 6h)`, the same bound the billing query already clamps an open session to. Either way the estimate never bills more than the open row would have). Drives metered-compute invoicing via `Tuist.Runners.Billing`. Retries (`Jobs.record_queued/1`) produce additional rows so the customer is billed for every Pod they actually held.
```
esnunes: Small one: this line names `Tuist.Runners.Workers.OrphanedRunnerSessionsWorker`, but the module that ships in this PR is `Tuist.Runners.Workers.PodReconciliationWorker` — I think the name moved when the session arm got folded into the claim worker. The `runner_claims` bullet two lines up (line 62) still points at `PodClaimReconciliationWorker`, which this PR deletes, so it'd be worth fixing both in the same pass.

The rest of the added text is genuinely good — spelling out the GREATEST/LEAST bound means a reader can tell an estimated close from a reported one.

server/lib/tuist/runners/prom_ex_plugin.ex
```diff
@@ -311,6 +306,21 @@ defmodule Tuist.Runners.PromExPlugin do
...
+      Polling.build(
+        :tuist_runners_session_clamp_metrics,
+        poll_rate,
+        {__MODULE__, :execute_session_clamp_telemetry_event, []},
+        [
+          last_value(
+            @metric_prefix ++ [:session, :clamped, :count],
+            event_name: Telemetry.event_name_session_clamp(),
+            description:
+              "Open runner sessions per fleet past the six-hour session safety bound — rows the autoscaler has stopped counting as occupied and billing has stopped charging for, i.e. sessions the orphan reaper failed to close.",
```
esnunes: Moving this to a poll is clearly right — the old placement could never have exported reliably.

One thing on the wording: the query behind it counts every open session past six hours with no predicate on whether the Pod exists, so the description's "i.e. sessions the orphan reaper failed to close" is a bit stronger than what the series measures. `started_at` is claim-win, so a job running near GitHub's 6h ceiling crosses the bound while its Pod is alive and the reaper is correctly ignoring it. The old string kept that nuance ("a live claim can still prove current occupancy").

It matters mostly for the verification step in the description — "should trend to zero and stay there" won't hold if a long job can pin it. Maybe keep the HELP text to what it counts and leave the interpretation to the runbook?

server/lib/tuist/runners/runner_sessions.ex
```diff
@@ -288,6 +281,135 @@ defmodule Tuist.Runners.RunnerSessions do
...
+    |> select([s], {s.fleet_name, count(s.id)})
+    |> Repo.all()
+    |> Map.new(fn {fleet_name, count} -> {fleet_name || "", count} end)
+  end
+
+  @doc """
+  Open sessions old enough to be judged against the observed Pod set,
+  oldest first so a capped batch drains the longest-standing leaks.
+
+  `threshold` keeps young sessions out: the row is written at claim-win,
```
esnunes: Might be reading this wrong, so treat it as a question. The doc justifies the threshold with "the row is written at claim-win, before the Pod exists to be listed" — but following dispatch, the Pod seems to already exist and be polling when it claims: `RunnerSessions.open/1` runs at runners.ex:1070, after `stamp_account_label/4` patches an existing Pod, and `tuist.dev/runner=true` is stamped at Pod creation (podtemplate.go:485), not at claim.

If that's right the Pod is listable before the session row is written, and the 15 minutes is really buying read-staleness margin. The window itself seems fine — I'd just tweak the sentence, since the stated reason points whoever tunes this next in the wrong direction. Is there a dispatch mode where the Pod is created after the claim that I've missed?

infra/runners-controller/controllers/runnerpool_controller.go
```diff
@@ -627,6 +652,27 @@ func (r *RunnerPoolReconciler) reapAlivePod(ctx context.Context, pod *corev1.Pod
...
+// could here (the job's real completion) when it does.
+//
+// nil client disables reporting, matching how the pod-lifecycle
+// reconciler is only wired when the sessions URL is configured.
+func (r *RunnerPoolReconciler) reportStopped(ctx context.Context, pod *corev1.Pod) {
+	if r.SessionsClient == nil {
+		return
+	}
+
+	if err := r.SessionsClient.Stopped(ctx, pod.Name, podEndedAt(pod, time.Now())); err != nil {
```
esnunes: Tiny one: this passes `time.Now()` where the rest of the file goes through the injectable `r.now()` (lines 224, 244, 251, 269). The lifecycle reconciler's call into `podEndedAt` uses its own clock, so this is the only caller that can't be pinned — which matters mainly for the fallback branch, the one that fires when a Pod has no terminated container and no deletion timestamp. `r.now()` would keep it consistent and testable.

Extracting `podEndedAt` so both callers agree on what "ended" means was a good call, by the way.

esnunes (approved):
This is a really good piece of work. Diagnosing the leak as an ordering race between two workqueues rather than a delivery failure — and then fixing it by removing the race instead of compensating for it — is the right shape, and the write-up of what each earlier revision got wrong made this much easier to review. The close arithmetic is careful in the direction that matters: LEAST against now keeps the under-bill bias, GREATEST against started_at rules out an inverted interval, and the terminal-status predicate in the ClickHouse lookup is what stops a stale completion reaching a live re-claim.

I checked the things that would have worried me and they hold up: the double-report path is MIN-biased server-side so it can't extend a window, `close_pod_missing/3`'s `is_nil(ended_at)` guard means an accurate late close always wins, the reap paths that touch alive Pods only ever see idle or never-claimed ones, and every runner Pod carries the label the worker selects on from creation. Folding the two reconcilers into one LIST reads as a straight improvement.

Approving — nothing I found blocks this. The inline notes are all small and independent: a stale module name in data-export.md (the added line names a worker that doesn't exist in the tree), a metric description that claims a bit more than its query measures, the stated rationale for the 15-minute session window, and a `time.Now()` that wants to be the reconciler's injectable clock.

server/data-export.md
fortmarek: Fixed both — lines 62 and 64 now name `Tuist.Runners.Workers.PodReconciliationWorker`. Thanks for catching the second one; I'd only looked at the line my own diff touched.

server/lib/tuist/runners/prom_ex_plugin.ex
fortmarek: You're right, and the verification step was the real casualty. The query filters on `ended_at IS NULL AND started_at < now() - 6h` with no Pod predicate, and `started_at` is claim-win, so a job near GitHub's ceiling pins it while the reconciler correctly ignores it. HELP text now describes what it counts and says so explicitly; the PR's "trend to zero and stay there" became "judge the trend, not the absolute".

server/lib/tuist/runners/runner_sessions.ex
fortmarek: You read it correctly and the sentence was wrong. `tuist.dev/runner=true` is on the Pod template at creation (podtemplate.go:485), and `open/1` runs after `stamp_account_label/4` has patched an already-running Pod — so the Pod is listable well before the row exists. No dispatch mode creates the Pod after the claim. Reworded to say what the window actually buys: margin against dispatch and teardown churn, not waiting for the Pod to appear. Same fix applied to the matching line in the worker moduledoc.

infra/runners-controller/controllers/runnerpool_controller.go
fortmarek: Switched to `r.now()`. The fallback branch was the one place it mattered and the one place a test couldn't pin it.

PR #12296, opened 2026-08-12
fix(cli): link cross-project resource bundles to their producer product

fortmarek:
### What changed

Target-produced cross-project resource bundles are now referenced from the consumer project with a `PBXReferenceProxy` pointing at the exact product reference in the producer `.xcodeproj`, plus an explicit `PBXTargetDependency` on the producer target. Previously each one was a standalone `PBXFileReference` with `sourceTree = BUILT_PRODUCTS_DIR`.

The bundle's output path is unchanged. Precompiled bundles keep the existing path reference, since they have no producing target.

`GraphTraverser` gains `resourceBundleTargetDependencies`. The existing `resourceBundleDependencies` reduces a bundle to its target and product names via `dependencyReference(to:from:)`, which is not enough to reference the producer across projects. The new query returns the same bundle set while retaining the producing project and target.

### Why: cross-project resource bundles break under `findImplicitDependencies: false`

A standalone `BUILT_PRODUCTS_DIR` reference only names a file the build products directory is expected to contain. Nothing declares which target produces it, so the build only works because Xcode rediscovers the producer by name-matching the file against target products. Turn that rediscovery off and the bundle has no producer at all.

`findImplicitDependencies: false` is supported Tuist API on `BuildAction`, and lands as `buildImplicitDependencies = "NO"` on the generated scheme (`SchemeDescriptorsGenerator.swift:299`). With an app depending on a resource bundle in another project, and a scheme built with that option:

| | before | after |
| --- | --- | --- |
| build | `error: The file "ResourceBundle.bundle" couldn't be opened because there is no such file` → **FAILED** | **SUCCEEDED** |
| target graph | `Target dependency graph (1 target)`, producer absent | `Target dependency graph (2 targets)`, `➜ Explicit dependency on target 'ResourceBundle' in project 'Resources'` |
| bundle in the app | missing | copied |

That is a reproducible build failure this change fixes.

With implicit dependencies left on (the default), the improvement is narrower but still real: the edge stops depending on name-matching. Xcode's own `target-graph.txt` shows it:

```
before: ➜ Implicit dependency on target 'ResourceBundle' in project 'Resources'
          via file 'ResourceBundle.bundle' in build phase 'Copy Files'
after:  ➜ Explicit dependency on target 'ResourceBundle' in project 'Resources'
```

### On the reported build-description invalidation

This change was originally proposed to fix a one-time Create Build Description replan, on the theory that a plain path becomes an additional CodeSign input only once the file exists, so the first build of a fresh derived data directory plans without the bundles and the next plans with them.

**That replan does not reproduce, and the theory does not hold up.** Across two builds on a fresh derived data directory the build description was reused in both arms (1 `.xcbuilddata` directory, before and after), on iOS simulator, on a real signed iOS device build, and on macOS, and deleting the built bundle between builds changed nothing. `App.app/ResourceBundle.bundle` never appears in the app's `CodeSign` task inputs at all.

The reason is the same mechanism as above: with implicit dependencies on, resolution registers the producer through the Copy Files phase, so the path is a known build output before the file exists and planning never depends on its existence. Since the reporter's builds succeed, they have implicit dependencies on, which is exactly the configuration where the file-existence fallback does not apply.

The `findImplicitDependencies: false` failure is the justification for this change. The replan is not.

### How to test locally

Unit tests:

```
tuist generate tuist TuistGeneratorTests TuistCoreTests --no-open
xcodebuild build-for-testing -workspace Tuist.xcworkspace -scheme Tuist-Workspace -destination "platform=macOS,arch=arm64" CODE_SIGNING_ALLOWED=NO
xcodebuild test-without-building -workspace Tuist.xcworkspace -scheme Tuist-Workspace -destination "platform=macOS,arch=arm64" -only-testing:TuistGeneratorTests/CrossProjectTargetDependencyGeneratorTests -only-testing:TuistCoreTests/GraphTraverserTests CODE_SIGNING_ALLOWED=NO
```

End to end. Create a workspace with an app in one project depending on a `.bundle` target in another, and give the app project a scheme with `findImplicitDependencies: false`:

```swift
schemes: [
    .scheme(
        name: "AppNoImplicit",
        shared: true,
        buildAction: .buildAction(targets: ["App"], findImplicitDependencies: false)
    ),
]
```

Generate and build that scheme. On `main` it fails with the missing-file error above; on this branch it succeeds, and `<DerivedData>/Build/Intermediates.noindex/XCBuildData/*.xcbuilddata/target-graph.txt` shows the `Explicit dependency` line.

### Validation performed

- `GraphTraverserTests` (166) and `CrossProjectTargetDependencyGeneratorTests` (10) pass. The generator suite gains a transitive-bundle case, a same-project no-op case, and the reporter's direct-dependency case.
- The `findImplicitDependencies: false` before/after in the table above.
- Generated pbxproj verified to contain one `PBXReferenceProxy` per bundle whose `PBXContainerItemProxy` has `proxyType = 2` and a `remoteGlobalIDString` equal to the producer target's `productReference`, with both the Resources and Dependencies build files pointing at it and no standalone `PBXFileReference` left behind. The equivalent workspace generated by the released CLI has the standalone reference and zero reference proxies.
- Builds on simulator and on a signed iOS device, with the bundle copied into the app. `codesign -v --deep --strict` passes on the device build, so the proxy does not disturb signing.

### Known gaps, not addressed here

- Cross-project **framework** links have the same weakness and this PR does not fix them. With `findImplicitDependencies: false`, an app reaching a bundle transitively through a cross-project static framework now gets `➜ Explicit dependency on target 'ResourceBundle'` and builds the bundle, but the `Feature.framework` link still fails because that target is likewise absent from the graph. Pre-existing, and worth its own change.
- With `destinations: .macOS` the same fixture produces no bundle reference at all in the consumer project, so a cross-project resource bundle is silently dropped from a macOS app. This reproduces on the released CLI, so it predates this change. `canEmbedPlugins()` / `canDependencyEmbedAsPlugin` in `resourceBundleDependencies` looks like the place to start.

### Reviewer notes

This differs from the originally proposed patch in three ways. Bundles are no longer added to the direct-dependency edge collection, since `resourceBundleTargetDependencies` already covers exactly the set that reaches build phases and the extra clause added proxies to consumers that never copy the bundle. Reference replacement is scoped to the consumer target's own build files and skips any reference that is a native target's product, so a same-named local product cannot be repointed or deleted. The orphan sweep and product lookups are keyed by object identity instead of rescanning `fileReferences` per edge.


cli/Sources/TuistGenerator/Generator/CrossProjectProductReferenceGenerator.swift
```diff
@@ -0,0 +1,161 @@
...
+/// project, and repoints the consumer's build files at it.
+///
+/// The project generator emits a standalone `PBXFileReference` with `sourceTree = BUILT_PRODUCTS_DIR`
+/// for every product a target consumes, which names a file the build products directory is expected to
+/// contain without saying which project and target produce it. A proxy pairs the product reference in
+/// the producer `.xcodeproj` with a `PBXContainerItemProxy`, so the relationship survives into the
+/// generated project instead of having to be rediscovered by matching file names.
+///
+/// Instances accumulate the references they replaced; call `removeOrphanedReferences` once every edge
+/// has been wired to drop the ones nothing points at anymore.
```
esnunes, line 15: Not a change request — more a suggestion for the description, since I think this change has a stronger justification available than the one it's leaning on.

Tuist's build action supports `findImplicitDependencies: false`, which lands as `buildImplicitDependencies = "NO"` on the scheme (`SchemeDescriptorsGenerator.swift:299`). I set that on a scheme generated from main and the cross-project bundle failed the build outright — `Target dependency graph (1 target)`, then `error: lstat(.../ResourceBundle.bundle): No such file or directory` — because nothing declares the producer. With this branch the same scheme gets `➜ Explicit dependency on target 'ResourceBundle' in project 'Resources'` and the bundle builds and copies. (The static framework link still fails in that configuration, but that looks like a separate pre-existing gap, not something this PR claims.)

That's a reproducible bug this change fixes. It doesn't support the replan theory — the reporter's builds evidently succeed, so they have implicit dependencies on — but it stands on its own, and I think it's more convincing to a reader than the name-matching argument alone.

esnunes (approved):
Nice change. I built a CLI from this branch and checked the generated output rather than reading the diff alone, and everything the description claims about the pbxproj holds: one `PBXReferenceProxy` per bundle with `proxyType = 2` pointing at the producer target's `productReference` UUID, an explicit `PBXTargetDependency` on the producer target, both the Resources and Dependencies build files repointed, and no standalone `PBXFileReference` left behind. Three things beyond the description also held up — two consumer targets in one project share a single proxy and each get their own dependency, the pbxproj is byte-identical across three regenerations, and it covers external SPM package bundles too (Tuist emits one `.xcodeproj` per package, so those are cross-project by construction). The workspace builds and `target-graph.txt` shows the explicit edge.

I couldn't reproduce the build-description replan either. Across the three-project transitive shape, a direct dependency, 45 bundles behind 45 static frameworks, an external SPM bundle, generic and concrete simulator destinations, Xcode 26.3 and 26.6, and signing both on and off, every arm reused a single `.xcbuilddata` across three consecutive builds — with the instrument validated by changing `OTHER_SWIFT_FLAGS` to force a second one. With signing on, the app's `CodeSign` command in `manifest.json` has 9 inputs and the bundle isn't among them, before or after it exists.

So my notes are about the story around the change rather than the change itself: the open question left for the reporter looks answerable from the code (and closes that theory rather than opening it), and there's a reproducible justification available that the description doesn't use. I didn't find a correctness problem in the code itself.

---

**On the open question in the description**

On the open question at the end of the description — whether the reporter's consumers get the bundle only in a Resources phase — I think this one can be closed from the code, and it points away from the replan theory rather than toward it.

`copyProductDependencies` unconditionally unions `resourceBundleDependencies` (`cli/Sources/TuistCore/Graph/GraphTraverser.swift:933`), and `generateCopyProductsBuildPhase` is called from `generateLinks` with no early return above it (`cli/Sources/TuistGenerator/Generator/LinkGenerator.swift:120`), for every target. Both phases read the same set, so I don't think Tuist can generate the Resources-only shape at all.

I also built that shape by hand, by stripping the Dependencies phase out of a project generated by the released CLI. Implicit discovery doesn't cover Resources phases: the bundle edge vanishes from `target-graph.txt` and the build fails outright with `error: lstat(.../ResourceBundle.bundle): No such file or directory`. So a reporter in that state wouldn't be seeing a replan, they'd be seeing a red build.

If it's worth one more round with them, what would actually settle it is the two `.xcbuilddata` directory names from their fresh-DerivedData session — the directory name is the build description signature — plus confirmation that both builds used the same scheme, destination and configuration.

cli/Sources/TuistGenerator/Generator/CrossProjectProductReferenceGenerator.swift
fortmarek: Confirmed and adopted — the description now leads with this. I reproduced it with an app depending directly on a cross-project bundle and a scheme built with `findImplicitDependencies: false`: on `main` the build fails with `The file "ResourceBundle.bundle" couldn't be opened because there is no such file` and `Target dependency graph (1 target)`; on this branch it succeeds with `➜ Explicit dependency on target 'ResourceBundle' in project 'Resources'` and the bundle copied. Your static-framework caveat reproduces too, so I listed the cross-project framework link as a known gap this PR does not address. The replan section is now explicitly labelled as not reproducing, with your point that the reporter must have implicit dependencies on.

PR #12297, opened 2026-08-12
refactor(kura): delete the Lua hook and everything that carried it

fortmarek:
Follow-up to #12264, which moved Kura's cache authorization out of a Lua hook and into the binary. That change deliberately left the hook and its plumbing in place. This removes them.

> [!WARNING]
> **Do not merge until every managed Kura instance runs a release with native authorization.** Merging #12264 is not sufficient. See [Why this has to wait](#why-this-has-to-wait) — landing it early would leave managed cache nodes serving unauthenticated.

## What changed

Nothing renders, ships, mounts or stores a policy script any more. The pieces only make sense removed together:

- `kura/ops/helm/kura/hooks/tuist.lua` — deleted
- **Server**: `kura_hook_path`, the script read and its `:persistent_term` cache, and `extensionScript` in the rendered manifest; the `COPY` that baked the file into the server image
- **Controller**: the `ExtensionScript` field, the ConfigMap it rendered, and the volume, mount and env vars that pointed Kura at it
- **CRD**: the `extensionScript` property
- **Kura**: the environment-variable fallbacks, so each setting is read under one name

`@manifest_revision` moves to `2026-08-12-native-authorization-v1`. The reconciler converges on the revision alone, so an `extraEnv` change that left it untouched would never be applied — the trap the surrounding comments already warn about.

Removing the fallbacks also removes the reason `auth/config.rs` read a list of names per setting. It reads one again, and the helper that walked the list is gone with it.

## Why this has to wait

A Kura node given no authorization configuration does not authorize at all — it serves the cache to anything that can reach it. So the question is whether any node can end up with the configuration removed while still running a release that needs it.

It can, through the path that runs most often. `Tuist.Kura.Reconciler` re-applies a drifted manifest via `apply_current_manifest/2`, using the instance's **current** image tag. The Kura image tag comes from `kura_runtime_image_tag` and rolls in health-gated waves, decoupled from server releases, so most server deploys never bump it. On such a deploy `desired == observed == the old tag`, and a server that stopped rendering `extensionScript` would strip the extension config from a StatefulSet still running the old image.

The version-change path is safe — it renders the new tag and the new manifest together — which is exactly why this is easy to miss. The hazard lives in the no-version-change path.

Order:

1. #12264 lands and ships a Kura release with native authorization.
2. `kura_runtime_image_tag` is bumped; the waves converge every instance.
3. This merges.

## Validation

- `mise run test-unit`, `mise run clippy`, `mise run format -- --check` (Bazel, what CI gates on) and the cargo equivalents: 681 tests, no warnings
- `go build ./... && go test ./...` in `infra/kura-controller`, `gofmt` clean
- `mix compile --warnings-as-errors`; `mix test test/tuist/kura/provisioner/kubernetes_controller_test.exs`: 61 tests

The end-to-end suite has not been run here; it needs `docker compose build`.

## Notes for the reviewer

**A stale object per instance.** The controller no longer reconciles the `<name>-extension` ConfigMap, so existing ones are orphaned. They are owned by their KuraInstance, so they are collected when it is, or can be deleted directly. No cleanup code was added rather than leave a permanent one-shot deletion in the reconcile loop.

**The CI validation preview loses a seal it had already lost.** `values-preview-ci.yaml` sealed itself with an inline deny-all hook, which a release with native authorization ignores — so the seal is gone as soon as #12264 ships, regardless of this PR. The inline script is removed rather than left looking like protection. Whether that preview should be sealed, and how, is a separate decision about exposure.


server/lib/tuist/kura/provisioner/kubernetes_controller.ex
```diff
@@ -535,15 +532,12 @@ defmodule Tuist.Kura.Provisioner.KubernetesController do
   # that ever runs Kura. The controller's envFrom on the StatefulSet
   # picks up that Secret automatically. Non-secret knobs such as the
   # introspection client ID are safe to keep in the spec.
-  defp extension_env(%Regions{} = region, entitlements) do
+  defp auth_env(%Regions{} = region, entitlements) do
```
esnunes, line 532: One thing I wanted to flag about what goes away with `KURA_EXTENSION_ENABLED`.

The controller used to inject `KURA_EXTENSION_ENABLED=true` whenever `spec.extensionScript` was set, and Kura's `ENABLED` list read that name. That made "authorize" an explicit instruction: with it set, a missing or blank URL makes `AuthConfig::from_env` return `Err` and the node refuse to start.

After this PR nothing on the managed path sets `KURA_AUTH_ENABLED` (I grepped — it only appears in docs, the standalone Kura chart, and the e2e compose files), so the sole enable signal is the presence of a non-blank `KURA_AUTH_TUIST_URL`. If that value ever renders blank, `from_env` returns `Ok(None)`, `app.rs:130` logs nothing, and the node starts serving the cache with no authorization engine at all.

I don't think it's reachable today: `tuist_base_url/1` falls back to `Environment.app_url()`, which builds a URI string. But `env_var/2` doesn't filter empty values the way `maybe_env_var/2` does, so nothing on this side would stop a blank from being emitted, and the failure is silent rather than loud.

Would it be worth emitting `KURA_AUTH_ENABLED=true` alongside the URL in `auth_env/2`, and the same pair in `kura-instance.yaml`? It restores the fail-closed startup for about two lines, and keeps "unconfigured" distinguishable from "deliberately unauthenticated".

kura/src/auth/config.rs
```diff
@@ -179,53 +132,12 @@ fn env_truthy(keys: &[&'static str]) -> Option<bool> {
...
-
-        let (key, value) = entry.expect("a value");
-        assert_eq!(key, "KURA_EXTENSION_HTTP_CLIENT_TUIST_BASE_URL");
-        assert_eq!(value, "https://old.example");
-    }
-
     #[test]
-    fn treats_a_blank_value_as_unset() {
-        assert!(first_set(TUIST_URL, lookup(&[("KURA_AUTH_TUIST_URL", "   ")])).is_none());
+    fn parses_the_algorithms_the_server_signs_with() {
```
esnunes, line 175: Two of the three deleted tests here were clearly obsolete — they covered name precedence, which no longer exists. `treats_a_blank_value_as_unset` is the one I'd think about keeping: it covers behaviour that survives, and with the fallback lists gone that rule is what decides whether `from_env` returns `Some` or `None`, i.e. whether the node authorizes at all.

I realise `env_value` now reads `std::env::var` directly, so porting it isn't free — a test would have to mutate real process env. Would keeping a small pure helper that takes a lookup closure (the way `first_set` did) be worth it, just so the blank-is-unset rule stays covered?

esnunes (approved):
This is a satisfying deletion to read — five surfaces coming out together, and the PR body does the hard part of explaining why the order matters rather than just what changed.

I swept the tree for every removed identifier (`KURA_EXTENSION_*`, `extensionScript`/`ExtensionScript`, `kura_hook_path`, `hook_script`, `hooks.lua`, `tuist.lua`) and the only survivors are the negative assertions in the two test files, which is where they belong. `manifest/6` has no callers outside tests, and no `AGENTS.md` node still describes the extension surface.

I also traced the merge warning rather than taking it on trust, and it holds: `project_server/2` only reaches `reconcile_manifest_revision/2` on the branch guarded by `when observed == desired`, so a revision-only bump really does re-render against the tag the instance is already running. Two adjacent worries I checked and dismissed — self-hosted nodes never set the Tuist-managed names this stops reading, and the CI preview still authorizes natively because the chart always emits `KURA_AUTH_TUIST_URL`.

The one thing I'd like your read on is what leaves with `KURA_EXTENSION_ENABLED`: the managed path no longer sets any explicit enable flag, so a blank control-plane URL would flip from a hard startup failure to a silent unauthenticated node. I don't think it's reachable today, and the fix looks like about two lines. The rest is small — one Rust test I think is worth keeping.

server/lib/tuist/kura/provisioner/kubernetes_controller.ex
fortmarek: Done in 82f71e248f — `auth_env/2` and `kura-instance.yaml` now emit `KURA_AUTH_ENABLED=true` alongside the URL, with a test pinning the pair so it cannot quietly disappear again.

I confirmed all three parts: `env_var/2` does not filter blanks the way `maybe_env_var/2` does, nothing under `lib/` or `infra/helm/tuist/` set the name after this PR, and a blank URL therefore lands on `Ok(None)` rather than `Err`. Agreed it is not reachable today, but the difference between the two failures is a node that refuses to start and a node that serves the cache to anyone, so the two lines are worth it.

kura/src/auth/config.rs
fortmarek: Restored in 82f71e248f. Rather than a lookup closure I split the trimming into its own function that takes the already-read value, so the rule is reachable from a test without touching process environment and `env_value` is just `configured(std::env::var(key).ok())`. Verified it fails when the emptiness filter is removed.

You were right that this one was not obsolete — with the fallback lists gone it is exactly what decides whether `from_env` answers `Some` or `None`, and I deleted it along with the two that genuinely were.

PR #12306, opened 2026-08-12
fix(infra): stop unschedulable Pods from starving sibling Linux runner pools

fortmarek:
## What changed

Two changes to the Linux runner fleet, both from the same incident:

1. **`infra/runners-controller`** — Pods the scheduler has explicitly rejected (`PodScheduled=False`, reason `Unschedulable`) no longer count against the fleet-wide `spec.provisioning.maxConcurrentPerFleetSelector` budget. They are bounded by the same ceiling *per pool* instead, blocking with a new reason `pool_unschedulable`.
2. **`infra/helm/tuist/values-managed-production.yaml`** — the `linux-2vcpu-8gb` warm floor drops from 30 to 12.

## Why

On 2026-08-12 the `Runner queue age` alert fired for `tuist-tuist-runner-pool-linux-4vcpu-16gb` at 35m and kept climbing. 26 jobs were queued, including `Release Server (amd64)` from `Server Production Deployment` on `main`, so the deploy cascade was stalled.

None of the causes the alert annotation lists applied. `tuist_runners_queue_withheld` was 0, `allocated_replicas` was 38, dispatch was healthy, webhooks were fine. The pool had `spec.replicas: 41` and **zero Pods in the cluster** — not Pending, not booting, never created.

### Root cause

The controller logged the reason every 5 seconds:

```
"Linux provisioning admission left replica gap"
reason=fleet_cap gap=39 creating=0 pendingForPool=0 pendingForFleet=4 cap=4 healthyNodes=2
```

Both bare-metal Linux nodes sat at **96% memory requests against 24-27% real usage**. Every runner Pod sets request == limit == its full advertised shape, so ~39 idle warm pollers reserved memory they never touched. With the `2vcpu-8gb` pool at a warm floor of 30 and a target of 83, four of its Pods could not be placed at all: `FailedScheduling ... 2 Insufficient memory`.

Those four Pods matched `isLinuxProvisioningPod` (alive, idle, poller not running), so they counted toward `pendingForFleet`, which hit `cap: 4`. That cap is shared across every pool with the same `fleetSelector`, so four unplaceable `2vcpu-8gb` Pods held the entire Linux fleet's creation budget and `4vcpu-16gb` was admitted zero Pods on every reconcile.

Nothing broke the loop. `startTimeoutSeconds` cannot reap these Pods either, because `linuxProvisioningStartedAt` only starts the clock once a Pod is bound to a node — deliberately, since an unscheduled Pod may legitimately wait longer than the sandbox-start timeout. So an unschedulable Pod was unbounded in both directions: never reaped, and holding budget the whole time.

### Why this fix

The gate exists to limit concurrent *sandbox starts*. A Pod the scheduler has rejected is starting nothing, so the boundary it enforces is imaginary, and waiting on it frees nothing — it is a deadlock, not a queue.

The obvious wider fix — exclude every Pod without a node — was wrong, and the existing tests caught it: a freshly created Pod also has no node, and dropping those from the count let two pools each admit a full batch (`TestReconcileCapsLinuxKataProvisioningAcrossSiblingPools` went from 4 Pods to 8) and moved the cache-lag reservations out of the budget they exist to protect. Narrowing to an explicit `Unschedulable` condition keeps both invariants: a new Pod carries no `PodScheduled` condition at all and holds its place, while a rejected one does not. Both pre-existing admission tests pass unchanged, which is the signal that the change is scoped correctly.

Bounding unschedulable Pods per pool rather than dropping them entirely keeps a pool that can place nothing from building an unbounded Pending backlog, without letting it take its siblings down.

### Why the warm floor too

The admission fix restores fairness but does not create memory: a 16 GiB Pod still cannot fit where an 8 GiB one already fails. The floor is a slot count, but because request == limit == shape it reserves `floor * 8 GiB` on a ~504 GiB fleet regardless of use. At 30 that is 240 GiB held against a measured ~11 concurrent claims, leaving no 16 GiB hole. 12 sits above observed demand and leaves the heavier shapes room to schedule.

**This number is the main thing worth a second opinion** — it is sized against today's 2-node fleet and current traffic, not a model.

## Impact

`4vcpu-16gb` (and every other non-default Linux shape) can be created again while a sibling has unplaceable Pods. Operationally, `desired >> observed` with `queue_withheld=0` now points at the admission log line, and `pool_unschedulable` distinguishes "this pool cannot place its shape" from "the fleet is busy starting sandboxes" — previously both read as `fleet_cap`.

## Validation

- `go build ./... && go vet ./controllers/...` clean.
- `go test ./controllers/...` passes, including the two pre-existing admission tests that the broader first attempt broke.
- Three new tests: unschedulable siblings excluded from the fleet budget, scheduled siblings still consuming it, and a pool's own unschedulable backlog bounded.
- Root cause confirmed live against production: controller logs, `RunnerPool` specs, `describe node` allocation, `FailedScheduling` events, and node-exporter memory.

Not deployed — the diagnosis was made with read-only cluster access and nothing in production has been changed. The warm floor takes effect on the next chart deploy.

## Follow-up, not in this PR

The real fix for the reservation wall is decoupling the scheduler **request** (measured P95 working set) from the **limit** (advertised shape), so the floor can be a slot count again. That needs eviction headroom and a PriorityClass first — the bare-metal kubelets set no `systemReserved` or eviction thresholds, so overcommit would hit kernel OOM rather than graceful eviction.

### DIFF
--- a/infra/helm/tuist/values-managed-production.yaml
+++ b/infra/helm/tuist/values-managed-production.yaml
@@ -844,15 +844,22 @@ runnersFleetLinux:
   shapes:
     - { vcpus: 1, memoryGb: 2 }
     - { vcpus: 2, memoryGb: 4 }
-    # Default shape: ~50% of P95 warm floor (the legacy pool's value),
-    # since all `tuist-default` + `linux` default-profile (`tuist-linux`)
-    # traffic lands here. The floor is a slot count, so it absorbs the
-    # same concurrent cold starts at 2/8 as it did at 4/16, for less RAM.
+    # Default shape: all `tuist-default` + `linux` default-profile
+    # (`tuist-linux`) traffic lands here, so it carries the fleet's warm
+    # floor. Sized against RAM, not slots: request == limit == the
+    # advertised shape, so the floor reserves `floor * 8 GiB` on a
+    # ~504 GiB fleet whether or not those Pods ever touch it. At 30 it
+    # reserved 240 GiB and left no 16 GiB hole, so `4vcpu-16gb` could
+    # not place a single Pod while running well under a third of the
+    # fleet's real memory. 12 sits above observed concurrent demand and
+    # leaves the heavier shapes room to schedule. Revisit alongside
+    # decoupling the scheduler request from the shape limit, which is
+    # what would let the floor be a slot count again.
     - vcpus: 2
       memoryGb: 8
       default: true
       autoscaling:
-        minWarmPoolFloor: 30
+        minWarmPoolFloor: 12
     - { vcpus: 4, memoryGb: 8 }
     - { vcpus: 4, memoryGb: 16 }
     - { vcpus: 8, memoryGb: 16 }
--- a/infra/runners-controller/AGENTS.md
+++ b/infra/runners-controller/AGENTS.md
@@ -118,17 +118,14 @@ independent workqueues:
   fleet boundary. Excess demand remains a replica gap and is retried
   every five seconds. macOS pools skip this gate.
 
-  Pods the scheduler has explicitly rejected (`PodScheduled=False`,
-  reason `Unschedulable`) are excluded from that fleet-wide count: they
-  boot no sandbox, so the velocity boundary they would enforce is
-  imaginary, and waiting on them frees nothing. They are instead bounded
-  by the same ceiling *per pool*, blocking with reason
-  `pool_unschedulable`. Without the split, a pool whose shape no node
-  can fit holds the whole `FleetSelector`'s budget indefinitely and
-  starves every sibling shape of creations — a shape with queued jobs
-  and healthy dispatch then reconciles at `observed: 0` forever. Note
-  that unschedulable Pods are also never reaped by the start timeout
-  (the clock starts at node binding), so nothing else bounds them.
+  The count deliberately includes Pods with no node. An unbound Pod is
+  one the scheduler may bind at any moment, and nothing re-checks
+  admission between that bind and the sandbox start, so discounting
+  unbound Pods would let a backlog build and then start all at once when
+  capacity returns — the very burst the ceiling exists to prevent.
+  Scheduling gates do not help here either: a gate can be removed but
+  never re-added, so a Pod that turns out to be unschedulable after
+  ungating is back to holding a slot with no way to reclaim it.
 
   Pod creates are visible to the cached client asynchronously. The
   reconciler therefore keeps a 30-second in-process reservation for each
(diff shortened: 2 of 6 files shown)


fortmarek:
Fixed in 6c94fc6186 — the P1 is real and the approach has changed.

**Why the previous version was wrong.** The ceiling is only meaningful if it is checked at the last point before a sandbox starts. Admission gates *creation*, and nothing re-checks between the scheduler binding a Pod and the Pod booting. Discounting unbound Pods therefore moved the accounting off the transition it was guarding: `7 * cap` Pods could pile up and start together the moment memory freed. That is the same burst the gate exists to prevent, and worse than the starvation it was fixing, because it fires exactly when the fleet is already under pressure.

**On scheduling gates.** I looked at them and they do not close this on their own. A gate can be removed but never re-added, so once the controller ungates a Pod and the scheduler rejects it, that Pod holds a slot again with no way to reclaim it. Gates would move the admission point to ungating — which is genuinely better placed — but they still need a way to release a Pod that turns out unschedulable, so they are a refinement on top of the fix below rather than a substitute for it.

**What it does now.** Every provisioning Pod counts again, bound or not, so the ceiling is exact and unchanged from before this PR. The starvation is fixed at its actual cause instead: an unschedulable Pod was holding a slot *forever*, because the only existing reap (`startTimedOut`) measures from node binding and so can never fire on a Pod with no node. A Pod the scheduler has been rejecting for longer than `startTimeoutSeconds`, timed from the `PodScheduled` condition's `LastTransitionTime`, is now reaped and its slot returned.

Reaping does not conjure capacity and is not meant to. The replica gap is untouched, so the pool retries, goes unschedulable again while the shortfall lasts, and gets reaped again — a slow loop bounded by the timeout, during which siblings can take the slot. Starvation goes from unbounded to at most one timeout.

**Tests**, including the transition you asked for:

- `TestProvisioningAdmissionCountsUnschedulableSiblingPodsAgainstFleetCap` — unschedulable siblings still consume the ceiling.
- `TestProvisioningAdmissionHoldsCapAsUnschedulablePodsBecomeSchedulable` — a Pod that was unschedulable and is now bound still counts, so the ceiling holds across exactly the capacity-returns moment.
- `TestUnschedulableTimedOutReleasesSlotOnlyAfterTimeout` — asserts `startTimedOut` does *not* fire on an unbound Pod, which is the gap that made this unbounded, then covers both sides of the new timeout.
- `TestUnschedulableTimedOutIgnoresPodThatLaterScheduled` — a Pod placed after rejection is not reaped out from under itself.

Reaps are split on the existing start-timeout counter as `poller_not_started` and `unschedulable`, so a rising `unschedulable` rate reads as a capacity shortfall for that shape rather than a boot fault. The `pool_unschedulable` admission reason is gone with the per-pool allowance.

The warm floor change is unaffected.

PR #12310, opened 2026-08-13
fix(cli): run the workspace scheme instead of splitting mixed test bundles

fortmarek:
## What changed

Reverts the scheme-selection behaviour introduced in #11225 (and refined in #11821). `tuist test` without an explicit `--scheme` again runs the autogenerated `<Workspace>-Workspace` scheme as a single `xcodebuild` invocation, instead of splitting into one invocation per test target when the workspace mixes app-hosted and host-less unit tests.

Scoped narrowly to scheme selection. Everything else from those PRs stays:

- the `didRunTests` refactor and per-scheme test-target filtering, which skip `xcodebuild` when every requested test target is pruned from a scheme
- per-scheme result bundle paths and reassembly (#11715, #11789)
- `-skip-testing:` passthrough handling for hash storage (#11821)

## Why

The split does not fix the crash it was written for, and it is very expensive.

### It does not work

The community repro is an app-hosted test bundle plus a host-less framework test bundle, where the app links a dynamic SPM framework. Reproduced end to end on Xcode 26.5 (17F42), the failure is:

```
+[XCTestSuite testSuiteForTestConfiguration:]
  -> +[XCTestCase(RuntimeUtilities) _allSubclasses]
  -> objc_copyClassList -> realizeAllClasses()
  -> <class in LocalStatusKit.framework>
  -> swift_getSingletonMetadata -> _swift_relocateClassMetadata -> SIGSEGV
```

The trigger is not scheme aggregation. It is that a **dynamic framework present in `BUILT_PRODUCTS_DIR`** gets loaded into the bare `xctest` runner even when nothing in the host-less bundle's link closure references it. `otool -L` on the test bundle shows only `XCTest`, `libXCTestSwiftSupport` and the module under test.

Proven by deleting the framework from an already-built products directory and re-running `test-without-building` with no rebuild:

| products dir | result |
| --- | --- |
| `Sharing*.framework` present | exit 65, SIGSEGV |
| `Sharing*.framework` deleted | exit 0, passed |

Because `tuist test` passes one `-derivedDataPath` for the whole invocation, every split scheme shares one products directory. On a completely clean DerivedData:

| sequence | result |
| --- | --- |
| `-scheme App` then `-scheme Feature` (what this code shipped) | App passes, **Feature fails, SIGSEGV** |
| `-scheme Feature` then `-scheme App` | both pass |

`defaultSchemes` sorted split schemes by target name, so `AppTests` always preceded `FeatureTests`, and the app's dynamic frameworks always landed in the products directory before the host-less bundle ran. The only order that works is the one the implementation could not produce. The original PR's validation passed because each scheme was exercised against a pristine products directory, which is a condition CI never has.

Scheme-level alternatives were tested too and all still crash or hang: `MacroExpansion` instead of an app `BuildableProductRunnable`, `shouldUseLaunchSchemeArgsEnv="NO"`, and `parallelizable="YES"`. The whole class of scheme-topology workarounds cannot address a products-directory problem.

### It is expensive

The heuristic fired whenever a workspace contained at least one app-hosted and one host-less unit test bundle anywhere, which is the normal shape of a modularised app. It then ran one `xcodebuild test` per test target, sequentially, each with its own destination resolution, xcresult upload, build-run upload and hash store.

On one large project the average number of test targets per `test` command event went from 282.9 on 4.200.5 to 1.2 on 4.203.1, meaning roughly 300 sequential `xcodebuild` invocations replacing one. Reported CI time went from under an hour to over six. That project rolled back twice and could not adopt any release after 4.200.5.

## For anyone hitting the original crash

Reverting this restores the crash for projects that were hitting it, so here are the two remedies that actually work. Both keep a single `xcodebuild` invocation.

**1. Link the offending dependency statically.** This is Tuist's default. If you forced a dynamic product type, drop the override:

```swift
// Tuist/Package.swift
let packageSettings = PackageSettings() // instead of productTypes: ["Sharing": .framework]
```

Validated: with the dependency static, the aggregate `<Workspace>-Workspace` scheme passes. The framework directory still exists but its binary is a static archive, so there is no dynamic image for `xctest` to load and no classes for XCTest to realise.

**2. Give the host-less test bundle a test host.** App-hosted bundles were immune in every run, against the same products directory. Only host-less bundles crash.

The underlying defect is in `XCTestCore`: `objc_copyClassList` forces realisation of every class in every loaded image, including frameworks the test bundle never links, imports or uses. Nothing Tuist emits can change that, and it is worth an Apple feedback report. Nothing in the Xcode 27 beta 1 to 5 release notes addresses it.

## Validation

- Reverted `defaultSchemes`, `containsMixedHostedAndHostlessUnitTests`, `isHostlessUnitTest` and `isCompatibleHostlessTestScheme`; call site returns `buildGraphInspector.workspaceSchemes(...)`.
- Scheme-planning tests updated to assert the restored behaviour: the workspace scheme is used for mixed hosted and host-less tests, explicit `--test-targets` filters the workspace scheme rather than selecting a per-target scheme, and `-skip-testing:` passthrough still stores hashes only for the targets that ran.
- End-to-end reproduction and all counter-experiments were run on Xcode 26.5 (17F42) against a generated fixture, as described above.


fortmarek:
cc @JanC

PR #12316, opened 2026-08-13
fix(cli): remove automatic Swift compatibility dylib runpaths

Loupehope:
### Summary

Removes the automatic Swift back-deployment compatibility dylib runpath handling added in [#11494](https://github.com/tuist/tuist/pull/11494), including its provider, test helper, and focused tests.

The SPM mapper no longer adds a toolchain-derived `LD_RUNPATH_SEARCH_PATHS` entry to every mapped SPM target.

### Root cause

The automatic behavior added the following runpath search path:
```
$(TOOLCHAIN_DIR)/usr/lib/swift-6.2/$(PLATFORM_NAME)
```
Although `xcodebuild -showBuildSettings` may report `XcodeDefault.xctoolchain`, Xcode can resolve the toolchain path used by an individual link task to the mounted optional `Metal Toolchain`. As a result, the final Mach-O can embed a runtime path such as:
```
Load command 20
          cmd LC_RPATH
      cmdsize 168
         path /var/run/com.apple.security.cryptexd/mnt/com.apple.MobileAsset.MetalToolchain-v17.6.109.0.1x0pPb/Metal.xctoolchain/usr/lib/swift-6.2/iphonesimulator (offset 12)
```

That path is transient build-machine state and is not a valid runtime dependency location. The same `Metal Toolchain` resolution issue has also been observed by [tauri-apps/tauri#14921](https://github.com/tauri-apps/tauri/pull/14921).

Replacing `$(TOOLCHAIN_DIR) with $(DEVELOPER_DIR)/Toolchains/XcodeDefault.xctoolchain` does not make the approach safe. Build settings are expanded before the linker writes `LC_RPATH`, so the final binary contains the absolute path of the Xcode installation that produced it, for example:
```
/Applications/Xcode-26.6.0.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/swift-6.2/iphonesimulator
```
This breaks portability when an artifact is run on a machine where Xcode has a different name or location, resulting in:
```
Library not loaded: @rpath/libswiftCompatibilitySpan.dylib
```

Finally, the behavior applied this runpath to every mapped SPM target. `libswiftCompatibilitySpan.dylib` is only needed by dependencies that use Swift Span back-deployment APIs. For example, SwiftNIO started using RawSpan in NIOCore in [apple/swift-nio#3269](https://github.com/apple/swift-nio/pull/3269). Most SPM dependencies do not require this compatibility runtime.

### Solution

Removed the automatic SPM target runpath behavior introduced in #11494. This removes `SwiftBackDeploymentLibrariesProvider`, its testing trait, its focused tests, the `PackageInfoMapper` logic that appended the toolchain-derived `LD_RUNPATH_SEARCH_PATHS` value, and the mapper test that covered only this automatic behavior.

Projects that need `libswiftCompatibilitySpan.dylib` can explicitly embed it in their final application or XCTest target through `ProjectDescription.TargetScript.post`:

```swift
ProjectDescription.TargetScript.post(
    script: """
    set -euo pipefail

    library_name="libswiftCompatibilitySpan.dylib"
    toolchain_path="${DEVELOPER_DIR}/Toolchains/XcodeDefault.xctoolchain"
    source_library_path="${toolchain_path}/usr/lib/swift-6.2/${PLATFORM_NAME}/${library_name}"
    destination_library_path="${TARGET_BUILD_DIR}/${FRAMEWORKS_FOLDER_PATH}/${library_name}"

    if [[ ! -f "$source_library_path" ]]; then
        echo "Error $library_name not found: $source_library_path"
        exit 1
    fi

    mkdir -p "$(dirname "$destination_library_path")"
    "${toolchain_path}/usr/bin/bitcode_strip" "$source_library_path" -r -o "$destination_library_path"
    codesign --force --sign "$EXPANDED_CODE_SIGN_IDENTITY" --verbose "$destination_library_path"
    """,
    name: "Embed Swift Compatibility Span",
    shellPath: "/bin/bash"
)
```
The swift-6.2 directory is intentional. It contains the Span compatibility runtime described in [Supporting Span in packages](https://forums.swift.org/t/supporting-span-in-packages/81667). The toolchain path is used only during the build. The resulting application or XCTest bundle contains the dylib and loads it through its normal relative framework search paths.


Loupehope:
cc @formatek as the author of the fix https://github.com/tuist/tuist/pull/11494

Loupehope:
cc @pepicrft

Loupehope:
Sorry if it feels like I’m always removing things from Tuist 😅 I’m just trying to keep it flexible, so a fix for one case doesn’t end up causing problems for another

pepicrft:
I dug into this further with a minimal generated project and the Swift Build / Swift Package Manager sources. I think we should be careful about changing a default here.

I modelled the pre-Xcode 26.5 binary shape: a dynamic Swift framework that requires `@rpath/libswiftCompatibilitySpan.dylib`. The results were:

- An Objective-C app or XCTest bundle with the default settings does not get a `CopySwiftLibs` task, so the final bundle does not contain the compatibility library.
- Setting `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES=YES` on that final Objective-C wrapper makes Xcode scan the linked framework and copy exactly `libswiftCompatibilitySpan.dylib` into its `Frameworks` directory. It did not copy arbitrary frameworks or other Swift runtime libraries in this reproduction.
- A Swift application with the setting left at Xcode's default still gets the same copy task automatically, and copies the same compatibility library when required.

This is consistent with [Swift Build's task producer](https://github.com/swiftlang/swift-build/blob/main/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/SwiftStandardLibrariesTaskProducer.swift#L40-L114). The source specifically calls out Objective-C XCTest bundles that test Swift frameworks, and it scans embedded and linked frameworks. Its compatibility filter admits `libswiftCompatibilitySpan` when Span needs back-deployment: [source](https://github.com/swiftlang/swift-build/blob/main/Sources/SWBTaskExecution/TaskActions/EmbedSwiftStdLibTaskAction.swift#L168-L175).

Two caveats:

1. The original report does not identify the host application or test bundle as Objective-C. We should not assume that is the original failing shape without a reproducer.
2. I could not produce the live loader failure locally because Xcode 26.6 no longer emits the standalone dependency from fresh Span source, and the available iOS 26.4 simulator already contains the library. The bundle contents and Xcode tasks do reproduce the relevant packaging difference.

The removal of the toolchain-derived runpath is still the right direction. I would avoid replacing it with `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES=YES` as a global Tuist default, though. It is unnecessary for normal Swift app targets, changes the behavior of every source-less wrapper, and the exact set of copied compatibility libraries depends on the toolchain and deployment target.

Could we first add a regression fixture that captures the original target shape and then use Xcode's native embedding task only where the fixture demonstrates it is needed? The fixture should assert no absolute toolchain runpath, and that the final app or test bundle contains `libswiftCompatibilitySpan.dylib` only when its embedded framework requires it. This would be preferable to asking users to maintain a custom copy-and-sign script.

Loupehope:
Thanks for investigating this further. I agree that removing the toolchain-derived runpath is the right direction, and that `ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES=YES` should not become a global Tuist default.

I think this is broader than the Objective-C wrapper scenario. I found two cases that affect Swift targets as well.

1. `libswiftCompatibilitySpan.dylib` is the compatibility library used to back-deploy Swift’s [`Span` / `RawSpan` APIs](https://forums.swift.org/t/supporting-span-in-packages/81667) to OS versions earlier than iOS 26. SwiftNIO started using `RawSpan` in [NIOCore](https://github.com/apple/swift-nio/pull/3269).

   When an app is built for iOS 26 or later, Xcode treats Span as provided by the OS and does not copy `libswiftCompatibilitySpan.dylib` into the app bundle. This follows the `SwiftSpanMinimumDeploymentTarget` logic used by [SwiftBuild](https://github.com/swiftlang/swift-build/blob/e4f6fc77ebe727657dadfedf50462f5a1a626ead/Sources/SWBTaskConstruction/TaskProducers/OtherTaskProducers/SwiftStandardLibrariesTaskProducer.swift#L108). If that app is then launched in a simulator running an earlier iOS version, the library is neither in the OS nor in the app bundle, causing:

   ```text
   dyld: Library not loaded: @rpath/libswiftCompatibilitySpan.dylib
   ```

2. XCTest targets can miss the library when the requirement is transitive. For example, a test target may depend on GRPC, while GRPC depends on NIOCore, which requires `libswiftCompatibilitySpan.dylib`. SwiftBuild does not discover this transitive requirement unless it is linked explicitly, so the compatibility library is not embedded and the same loader error occurs.

This is why the issue can affect regular Swift apps, XCTest bundles, not only Objective-C.

I will try to add regression fixtures for both cases, with and without the explicit shell-script workaround.

Loupehope:
@pepicrft, would it be feasible to add a regression test in CI that runs the fixture on an iOS simulator below iOS 26 and reproduces the missing `libswiftCompatibilitySpan.dylib` loader failure? This would let us validate the runtime failure directly, in addition to asserting the generated build tasks and bundle contents.

I don’t currently see a way to do this, though, because the CI runner images don’t pin or guarantee the availability of a pre-iOS 26 simulator runtime.

Loupehope:
@fortmarek @pepicrft I'd appreciate a code review 🙌

PR #12320, opened 2026-08-13
feat(atlas): presign stored artifacts for the internal Atlas API

fortmarek:
Atlas can already read the rows behind a support escalation — it has read-only access to this database and to ClickHouse — but not the files those rows point at: the `.xcresult` bundle of a failed test run, the `.ips` crash report attached to a test case run, the archive holding a build's `.xcactivitylog`, a preview's IPA. Those live in object storage Atlas has no credentials for, so today they get handed over manually.

This adds `GET /api/internal/atlas/artifacts/:kind/:id` behind `InternalAtlasAuthPlug`, alongside the existing `/atlas/db` and `/atlas/clickhouse` endpoints. It resolves a record to its storage key and signs a short-lived download; only the URL crosses the wire, so a multi-gigabyte bundle never travels through this server.

| `kind` | `id` | resolved through |
| --- | --- | --- |
| `test_run_result_bundle` | test run id, or the command event id of an older `tuist test` run | `Tests.get_test/1`, falling back to `CommandEvents.get_command_event_by_id/1` |
| `test_run_session` | same | same |
| `test_case_run_attachment` | `test_case_run_attachments.id` | attachment → test case run → project; handles the legacy `test-case-runs/` path when `test_run_id` is null |
| `build_archive` | build run id | `Builds.build_storage_key/3` |
| `preview_app_build` | app build id | `AppBuilds.storage_key/1` |

Two things worth a reviewer's attention:

- **The resolver reuses the product's own key builders** rather than restating the path conventions, so a drift in one shows up in both. The command-event and run-scoped builders already produce the same `<account>/<project>/runs/<id>/…` shape, which is why one lookup fallback covers both generations of test run.
- **Each kind is looked up, and its object sized, before it is presigned.** That is what lets the response separate `record_not_found` (404) from `artifact_not_stored` (404) from `storage_unavailable` (502). A row outliving its artifact is the common escalation shape, and it is the distinction that usually resolves the ticket.

URLs default to 15 minutes and are clamped to an hour.

Runner job logs are deliberately not here: they are ClickHouse rows (`Tuist.Runners.JobLogs`), not a stored object, and Atlas's `query_tuist_clickhouse` already reaches them.

The Atlas-side MCP tool that consumes this is tuist/atlas#400. This needs to merge first; until it does, that tool reports the server as unreachable.

### How to test locally

```
mix test test/tuist/atlas/artifacts_test.exs test/tuist_web/controllers/internal/atlas_artifacts_controller_test.exs
```

21 tests cover each kind, the command-event fallback, the legacy attachment path, expiry clamping, and every error branch. `Tuist.Storage` is stubbed, so no bucket is touched.

### DIFF
--- a/server/lib/tuist.ex
+++ b/server/lib/tuist.ex
@@ -53,6 +53,7 @@ defmodule Tuist do
       Cache.Analytics,
       CacheEndpoints,
       Atlas,
+      Atlas.Artifacts,
       AtlasWorkloadIdentity,
       Kubernetes.Client,
       Kura,
--- a/server/lib/tuist/atlas/artifacts.ex
+++ b/server/lib/tuist/atlas/artifacts.ex
@@ -1,0 +1,177 @@
+defmodule Tuist.Atlas.Artifacts do
+  @moduledoc """
+  Resolves a stored artifact to a short-lived presigned download URL for the
+  internal Atlas workload.
+
+  Atlas runs in its own cluster with no credentials for the artifact buckets, so
+  it cannot reach object storage on its own. Support escalations, however, are
+  usually about the artifact rather than the row: the `.xcresult` bundle behind a
+  failed test run, the crash report attached to a test case run, the
+  `.xcactivitylog` behind a slow build. This module maps a record id to the key
+  its artifact lives under and presigns a download for it — the same key builders
+  the product paths use, so a drift in one shows up in both.
+
+  Each kind is looked up before it is presigned, so a caller learns the
+  difference between "no such run" (`:not_found`) and "the run exists but its
+  artifact was never uploaded or has since been pruned" (`:object_not_found`),
+  which is the distinction that usually resolves the escalation.
+  """
+
+  alias Tuist.AppBuilds
+  alias Tuist.Builds
+  alias Tuist.CommandEvents
+  alias Tuist.Projects
+  alias Tuist.Storage
+  alias Tuist.Tests
+
+  @default_expires_in 900
+  @max_expires_in 3600
+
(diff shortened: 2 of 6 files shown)


fortmarek:
Closing: this adds a private endpoint for artifacts that already have download paths, and the access model is wrong.

The paths exist. Attachments are already served with presigned URLs by the token-authenticated API (`GET /api/projects/:account/:project/tests/test-cases/runs/:id/attachments`) and by the MCP tool `list_test_case_run_attachments`; result bundles, session archives and build archives are on the dashboard at `/:account/:project/runs/:id/download` and friends. This PR restated their key conventions behind a second door.

The access model is the more important part. The endpoint authenticates the Atlas workload and then hands over any account's artifacts — no operator grant, no account scoping, no expiry. That sidesteps `ops_access` and the whole ops.tuist.dev grant flow, which exists precisely to govern how staff reach customer data. It is consistent with the blanket read the `/atlas/db` tools already have, but files (screenshots, crash dumps, signed IPAs) are the half of that precedent worth not widening by reflex.

The replacement is to give Atlas a principal these surfaces already accept rather than a new surface: Atlas's MCP proxy already speaks OAuth2 to upstream servers, this server advertises `/.well-known/oauth-protected-resource/mcp` and offers dynamic client registration at `/oauth2/register`, so each Atlas user can authenticate to this server's MCP as their own Tuist user. Anything still missing then gets added as an MCP tool next to `list_test_case_run_attachments`, in this repo, under the normal authorization.

Branch `claude/atlas-artifact-fetch` is left in place if any of the resolver is worth salvaging. Atlas side: tuist/atlas#400, closed for the same reason.

PR #12322, opened 2026-08-13
fix(cli): honor SwiftPM's netrc options when resolving with SwifterPM

fortmarek:
### What

Gives SwifterPM the netrc surface SwiftPM exposes: `--netrc-file <path>` selects the file, `--enable-netrc` / `--disable-netrc` toggle netrc lookups, and `--netrc` disables the OS credential store for registry requests. It also consolidates two copies of the netrc lookup into one type, forwards the configuration to the child solve, and fixes three parser and validation behaviours that disagreed with every other tool reading the same file.

### Why

`tuist install` forwards its passthrough arguments to SwifterPM verbatim, and SwifterPM defined no netrc options at all. A pipeline that authenticated a private registry with `swift package resolve --netrc-file <path>` therefore broke the moment SwifterPM became the default resolver. A customer hit this while upgrading the CLI and worked around it with `SWIFTPM_NETRC_DATA="$(<"$HOME/.netrc")" tuist install`.

The failure was worse than a clean rejection. ArgumentParser discards the unrecognized `--netrc-file` and then tries its value as the `action` positional, so the actual message was:

```
Error: The value '/path/to/.netrc' is invalid for '<action>'
```

which says nothing about netrc. `Unknown option` appears only for the valueless forms:

```
Error: Unknown option '--disable-netrc'
```

### Root cause

No netrc options existed, and `DeprecatedSwiftPMOptions` is deliberately an exact allowlist rather than a blanket "ignore anything unrecognized", so nothing absorbed them.

The environment variable workaround covering more ground than the flag would have is not a coincidence. The version solve is delegated to a child `swift package resolve`, and the child inherits the environment but not our flags, so `SWIFTPM_NETRC_DATA` was the only mechanism that reached both halves. That gap is closed here.

### Solution, and what was rejected

Netrc configuration travels on the resolution and restore requests and is installed as a task local, matching how `cachedDirectoryMaterialization` is already threaded. Plumbing an extra parameter down to the two authorization helpers would have touched far more code for the same result.

A missing `--netrc-file` fails the command up front instead of silently downgrading to unauthenticated requests. Falling back quietly is the tempting option, and it is the one that produces the bug report a week later: an opaque 401 or 404 from a private registry with no hint that the credentials were never read. SwiftPM errors here too, in both of its authorization factories.

`--netrc-optional` is deliberately not tolerated. Swift 6.3.2 rejects it outright, so swallowing it would make SwifterPM more permissive than SwiftPM, which is the failure mode the allowlist exists to prevent in the other direction. The same reasoning now rejects `--netrc-file` combined with `--disable-netrc`, which upstream fails with "'--disable-netrc' and '--netrc-file' are mutually exclusive". That check lives in `Netrc.resolve` rather than in the CLI so embedders building a configuration directly hit it too.

### Why SwifterPM parses netrc itself

The obvious question on a hand-written parser is why it is not a dependency. Every off-the-shelf option was checked:

- `TSCUtility.Netrc` in swift-tools-support-core is marked `@available(*, deprecated, message: "moved to SwiftPM")`, dated 2/2022, and is gated at macOS 10.13.
- `Basics.Netrc` in swift-package-manager is where it moved and is the real reference implementation, but it lives inside SwiftPM. Depending on it pulls llbuild, swift-driver, TSC and swift-collections into a tool whose whole premise is being the lean alternative, and it would have to resolve under Bazel and Buck2 as well as SwiftPM. `Basics` is not among the products SwiftPM exports either.
- `ggruen/SwiftNetrc` is the only standalone package. Last pushed March 2018, `swiftLanguageVersions: [3, 4]`, a `final class` with `open` members that modern Swift rejects, its own `FileManager` I/O, a hard requirement that the file be `0o600`, comment stripping commented out in the source, and `[String: NetrcMachine]` storage that collapses duplicate hosts by accident.
- The Swift CLI has no netrc dump to shell out to. `swift package config` offers only the mirror subcommands and `swift package-registry` only set, unset, login, logout and publish. Printing credentials on a subprocess's stdout would be a poor idea regardless, since SwifterPM captures subprocess output for diagnostics.

So the parser stays. It is a tokenizer over a frozen file format, about ninety lines, and the dependency cost across three build systems exceeds its maintenance cost. It is not a port of SwiftPM's, which strips comments with one regex and matches whole entries with a second, while this one walks characters into a token stream and scans it. Reading them side by side is what produced the corrections below.

### Behaviours corrected against `main`

**A duplicated host resolved to the last entry, not the first.** SwiftPM uses `machines.firstIndex(where: { $0.name == url.host })`, and curl pointed at a file with two `machine 127.0.0.1` entries sends the first one, confirmed against a local server that decodes the header it receives. Appending a corrected entry therefore changed which credential SwifterPM sent while every other tool kept using the original.

**An unquoted `#` truncated the value it appeared in.** `password se#cret` parsed as the password `se`. SwiftPM's comment pattern is `("[^"]*"|\s#.*$)`, which requires whitespace ahead of the `#`, and curl sends `se#cret` whole. A `#` that opens a token still starts a comment, so whole-line and trailing comments are unaffected.

**A missing value took the rest of the file with it.** `login` and `password` accepted whatever token followed, and since tokenizing has already erased line boundaries, a value that never materialises consumed the next entry's `machine` keyword. Given two entries where the first ends `password ""`, the parser returned one machine whose password was the literal `machine`, and the second entry was gone; `login ""` returned nothing at all for the whole file. Values now refuse a token that opens an entry, so the damage stays inside the malformed entry. This is the same silent credential downgrade the missing-file error exists to prevent, reachable by a typo rather than by configuration.

**`--netrc` was swallowed as a deprecated no-op, and it is not one.** SwiftPM's `SecurityOptions.forceNetrc` is undeprecated, and `getRegistryAuthorizationProvider` consumes it as `authorization.keychain = forceNetrc ? .disabled : .enabled` under the comment "Don't use OS credential store if user wants netrc". Ignoring it left no way to prefer a fresh netrc entry over a stale keychain item from an old `swift package-registry login`, which is exactly the situation someone passes it for. It is now a real flag.

**The netrc configuration never reached the child solve.** `swift package resolve` performs the version solve, which fetches availability and manifests from the same registries, and it was invoked without `--netrc-file`, `--netrc` or `--disable-netrc`. Forwarding them covers `assertResolvedFileUpToDate` as well, since it builds on the same argument function. Verified with a PATH shim, the child now receives `--netrc-file <path> --netrc`.

### Divergences kept on purpose

- Registry lookups fall through on a miss; upstream selects at most one provider and stops. Falling through is what lets `--netrc-file` reach registry auth when the keychain holds an unrelated item, which is the point of the change. It does mean a host absent from the chosen source gets credentials here where SwiftPM would send none.
- `--disable-netrc` switches off registry netrc here. Upstream documents the enable/disable pair as having "no effects on registry communications", and `getRegistryAuthorizationProvider` never reads it.
- Host matching folds case on both sides, where SwiftPM compares the raw URL host. netrc machine names are DNS names, so folding is the more correct reading, and it can only widen what matches rather than change which entry wins.
- `GITHUB_TOKEN` / `GH_TOKEN` still act as a fallback for github.com downloads, behind netrc. SwiftPM has no such fallback; this was added deliberately in #11528.
- A `default` entry is accepted anywhere in the file. SwiftPM throws `invalidDefaultMachinePosition` unless it is last, which discards the whole file.
- Downloads do not consult the keychain, which SwiftPM does. Pre-existing, and left for a separate change.

### Impact

`tuist install -- --netrc-file <path>` works again, as do `--enable-netrc`, `--disable-netrc` and `--netrc`, and they now apply to the child solve as well as to SwifterPM's own requests. Nobody needs the `SWIFTPM_NETRC_DATA` wrapper anymore, though it keeps working. Registry precedence is `SWIFTPM_REGISTRY_TOKEN`, then `SWIFTPM_REGISTRY_LOGIN` and `SWIFTPM_REGISTRY_PASSWORD`, then inline netrc data, then the keychain unless `--netrc` was passed, then the netrc file. Downloads ask netrc in source order, then fall back to an ambient GitHub token.

### Validation

Red first, against the parent commit built in a scratch worktree. A test asserting the four options parse fails there:

```
✘ netrcFileOptionIsAccepted: unableToParseValue(... "/tmp/netrc", forKey: action)
✘ netrcToggleOptionsAreAccepted: unknownOption("disable-netrc")
✘ netrcToggleOptionsAreAccepted: unknownOption("enable-netrc")
✘ netrcToggleOptionsAreAccepted: unknownOption("netrc")
```

and the pre-fix binary reproduces both customer-visible errors quoted above. Each parser correction was written test-first and observed failing, one reporting `password: "se"` and another returning `first.example.com` with the password `machine`.

Green on the fix, same binary and same scratch package: `--netrc-file` with an existing file resolves, `--netrc` resolves, a missing `--netrc-file` exits 1 with `did not find netrc file at ...`, and the flag pair exits 1 with the mutually-exclusive message. Child forwarding was verified with a shim on `PATH` in front of a package with a local git dependency, so a real solve runs without touching the network.

The suite is at 162 tests, including registry precedence as a pure function over its three sources, per-origin lookup, the child argument list, and both authorization headers. `SwifterPMCore` builds in Swift 6 language mode, which bazel does not exercise.

### Note on the branch history

Two things this branch changes were regressions it introduced rather than problems on `main`. The registry ordering was already inline netrc data, then keychain, then netrc file at the merge base, and an intermediate commit here moved netrc ahead of the keychain on a misreading of upstream, citing the download provider's composition for the registry one. The same commit made `--netrc-file` suppress `SWIFTPM_NETRC_DATA`. Both are repaired, and the end state matches the merge base plus the corrections listed above. An early commit message also quotes the pre-fix failure as `Unknown option '--netrc-file'`, which a later commit corrects. This description reflects the end state rather than the route.

### DIFF
--- a/swifterpm/Sources/swifterpm/CLI.swift
+++ b/swifterpm/Sources/swifterpm/CLI.swift
@@ -36,6 +36,8 @@ struct CLI {
     var buildPath: CLIPath?
     var configPath: CLIPath?
     var securityPath: CLIPath?
+    var netrcFile: CLIPath?
+    var netrc = true
     var disableSandbox = false
     var enableDependencyCache = false
     var disableDependencyCache = false
@@ -152,6 +154,7 @@ public enum SwifterPMCommandParser {
                         cli: cli, paths: paths, packageDir: package, commandScratchDir: options.scratchDir),
                     registryConfigurationPath: paths.resolve(cli.configPath),
                     defaultRegistryURL: cli.defaultRegistryURL,
+                    netrc: CLIRunner.netrcConfiguration(cli: cli, paths: paths),
                     disableSandbox: cli.disableSandbox,
                     disablePackageInfoCache: cli.disablePackageInfoCache,
                     packageInfoCacheDirectory: paths.resolve(cli.packageInfoCachePath),
@@ -181,6 +184,7 @@ public enum SwifterPMCommandParser {
                 cli: cli, paths: paths, packageDir: package, commandScratchDir: nil),
             registryConfigurationPath: paths.resolve(cli.configPath),
             defaultRegistryURL: cli.defaultRegistryURL,
+            netrc: CLIRunner.netrcConfiguration(cli: cli, paths: paths),
             disableSandbox: cli.disableSandbox,
             forceResolvedVersions: cli.forceResolvedVersions || cli.disableAutomaticResolution
                 || cli.onlyUseVersionsFromResolvedFile,
@@ -223,6 +227,12 @@ public struct SwifterPMCommand: AsyncParsableCommand {
 
     @Option(name: .customLong("security-path"))
     var securityPath: String?
+
+    @Option(name: .customLong("netrc-file"))
+    var netrcFile: String?
+
+    @Flag(inversion: .prefixedEnableDisable)
+    var netrc = true
 
     @Flag(name: .customLong("disable-sandbox"))
     var disableSandbox = false
@@ -305,6 +315,8 @@ public struct SwifterPMCommand: AsyncParsableCommand {
             buildPath: CLIPath.optional(buildPath),
             configPath: CLIPath.optional(configPath),
             securityPath: CLIPath.optional(securityPath),
+            netrcFile: CLIPath.optional(netrcFile),
+            netrc: netrc,
             disableSandbox: disableSandbox,
             enableDependencyCache: enableDependencyCache,
             disableDependencyCache: disableDependencyCache,
@@ -354,10 +366,18 @@ enum CLIParser {
 /// subcommand parse, because `.allUnrecognized` on `commandArguments` only
 /// collects arguments that trail the `action` positional, and `tuist install`
 /// forwards passthrough arguments ahead of it.
+///
+/// `--netrc` joins that set because SwiftPM deprecated it once netrc became the
+/// default, so it only restates what `--enable-netrc` already gives.
+/// `--netrc-file` and `--enable-netrc`/`--disable-netrc` still change what
+/// happens, so they are real options instead. `--netrc-optional` is deliberately
+/// absent: Swift 6.3.2 rejects it outright, so tolerating it here would be more
+/// permissive than the tool we are standing in for.
 public enum DeprecatedSwiftPMOptions {
     static let ignored: Set<String> = [
         "--disable-prefetching",
         "--enable-prefetching",
+        "--netrc",
     ]
 
     public static func strip(_ arguments: [String]) -> [String] {
@@ -488,66 +508,79 @@ struct CLIPathResolver {
 }
 
 enum CLIRunner {
+    static func netrcConfiguration(cli: CLI, paths: CLIPathResolver)
+        -> SwifterPMNetrcConfiguration
+    {
+        SwifterPMNetrcConfiguration(isEnabled: cli.netrc, path: paths.resolve(cli.netrcFile))
+    }
+
     static func run(_ cli: CLI) async throws {
         try await Environment.withCachedDirectoryMaterialization(
             cli.cachedDirectoryMaterialization
         ) {
             let paths = try await CLIPathResolver(chdir: cli.chdir)
-
-            switch cli.command {
-            case .resolve(let options):
-                try ensureWholePackageResolution(
-                    packageName: options.packageName,
-                    version: options.version,
-                    branch: options.branch,
-                    revision: options.revision
-                )
-                try await runResolutionCommand(
-                    cli: cli,
-                    paths: paths,
-                    packageDir: options.packageDir,
-                    cacheDir: options.cacheDir,
-                    preferResolvedFile: true,
-                    write: options.write,
-                    restore: options.restore,
-                    printOnly: options.printOnly
-                )
-            case .update(let options):
-                if !options.packageNames.isEmpty {
-                    throw ToolError.message("package-specific update is not supported yet")
-                }
-                try await runResolutionCommand(
-                    cli: cli,
-                    paths: paths,
-                    packageDir: options.packageDir,
-                    cacheDir: options.cacheDir,
-                    preferResolvedFile: false,
-                    write: options.write,
-                    restore: options.restore,
-                    printOnly: options.printOnly
-                )
-            case .restore(let options):
-                let cache = try await Cache(
-                    root: cliCacheDir(cli: cli, paths: paths, commandCacheDir: options.cacheDir))
-                let package = canonicalPackageDir(
-                    commandPackageDir(
-                        cli: cli, paths: paths, commandPackageDir: options.packageDir))
-                let scratch = commandScratchDir(
-                    cli: cli, paths: paths, packageDir: package, commandScratchDir: options.scratchDir)
-                let registryConfig = try await cliRegistryConfig(
-                    cli: cli, paths: paths, package: package)
-                let resolved = try await ResolvedFile.read(packageDir: package)
-                try await WorkspaceRestorer.restorePackage(
-                    scratchDir: scratch, packageDir: package, cache: cache, registryConfig: registryConfig,
-                    resolved: resolved,
-                    progress: cli.quiet ? nil : RestoreProgressReporter(),
-                    disableSandbox: cli.disableSandbox)
-                try await maybeWritePackageInfoCache(
-                    cli: cli, paths: paths, package: package, scratch: scratch, resolved: resolved)
-                try await WorkspaceRestorer.writeWorkspaceState(
-                    packageDir: package, scratchDir: scratch, resolved: resolved,
-                    disableSandbox: cli.disableSandbox)
+            let netrc = netrcConfiguration(cli: cli, paths: paths)
+            try await Netrc.validate(netrc)
+            try await Environment.withNetrc(netrc) {
+                try await runCommand(cli: cli, paths: paths)
             }
+        }
+    }
+
+    private static func runCommand(cli: CLI, paths: CLIPathResolver) async throws {
+        switch cli.command {
+        case .resolve(let options):
+            try ensureWholePackageResolution(
+                packageName: options.packageName,
+                version: options.version,
+                branch: options.branch,
+                revision: options.revision
+            )
+            try await runResolutionCommand(
+                cli: cli,
+                paths: paths,
+                packageDir: options.packageDir,
+                cacheDir: options.cacheDir,
+                preferResolvedFile: true,
+                write: options.write,
+                restore: options.restore,
+                printOnly: options.printOnly
+            )
+        case .update(let options):
+            if !options.packageNames.isEmpty {
+                throw ToolError.message("package-specific update is not supported yet")
+            }
+            try await runResolutionCommand(
+                cli: cli,
+                paths: paths,
+                packageDir: options.packageDir,
+                cacheDir: options.cacheDir,
+                preferResolvedFile: false,
+                write: options.write,
+                restore: options.restore,
+                printOnly: options.printOnly
+            )
+        case .restore(let options):
+            let cache = try await Cache(
+                root: cliCacheDir(cli: cli, paths: paths, commandCacheDir: options.cacheDir))
+            let package = canonicalPackageDir(
+                commandPackageDir(
+                    cli: cli, paths: paths, commandPackageDir: options.packageDir))
+            let scratch = commandScratchDir(
+                cli: cli, paths: paths, packageDir: package, commandScratchDir: options.scratchDir)
+            let registryConfig = try await cliRegistryConfig(
+                cli: cli, paths: paths, package: package)
+            let resolved = try await ResolvedFile.read(packageDir: package)
+            try await WorkspaceRestorer.restorePackage(
+                scratchDir: scratch, packageDir: package, cache: cache, registryConfig: registryConfig,
+                resolved: resolved,
+                progress: cli.quiet ? nil : RestoreProgressReporter(),
+                disableSandbox: cli.disableSandbox)
+            try await maybeWritePackageInfoCache(
+                cli: cli, paths: paths, package: package, scratch: scratch, resolved: resolved)
+            try await WorkspaceRestorer.writeWorkspaceState(
+                packageDir: package, scratchDir: scratch, resolved: resolved,
+                disableSandbox: cli.disableSandbox)
         }
     }
 
--- a/swifterpm/Sources/swifterpm/CLI.swift
+++ b/swifterpm/Sources/swifterpm/CLI.swift
@@ -38,6 +38,7 @@ struct CLI {
     var securityPath: CLIPath?
     var netrcFile: CLIPath?
     var netrc = true
+    var forceNetrc = false
     var disableSandbox = false
     var enableDependencyCache = false
     var disableDependencyCache = false
@@ -234,6 +235,9 @@ public struct SwifterPMCommand: AsyncParsableCommand {
     @Flag(inversion: .prefixedEnableDisable)
     var netrc = true
 
+    @Flag(name: .customLong("netrc"))
+    var forceNetrc = false
+
     @Flag(name: .customLong("disable-sandbox"))
     var disableSandbox = false
 
@@ -317,6 +321,7 @@ public struct SwifterPMCommand: AsyncParsableCommand {
             securityPath: CLIPath.optional(securityPath),
             netrcFile: CLIPath.optional(netrcFile),
             netrc: netrc,
+            forceNetrc: forceNetrc,
             disableSandbox: disableSandbox,
             enableDependencyCache: enableDependencyCache,
             disableDependencyCache: disableDependencyCache,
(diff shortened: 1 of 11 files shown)


swifterpm/Sources/swifterpm/CLI.swift
```diff
@@ -354,10 +366,18 @@ enum CLIParser {
...
+/// default, so it only restates what `--enable-netrc` already gives.
+/// `--netrc-file` and `--enable-netrc`/`--disable-netrc` still change what
+/// happens, so they are real options instead. `--netrc-optional` is deliberately
+/// absent: Swift 6.3.2 rejects it outright, so tolerating it here would be more
+/// permissive than the tool we are standing in for.
 public enum DeprecatedSwiftPMOptions {
     static let ignored: Set<String> = [
         "--disable-prefetching",
         "--enable-prefetching",
+        "--netrc",
```
esnunes: I went to check the premise for putting `--netrc` on the ignored list, that SwiftPM deprecated it and it only restates `--enable-netrc`, and I do not think that holds. In `SecurityOptions` on both `release/6.3` and `main` it is a separate, undeprecated flag:

```swift
/// Force usage of the netrc file even in cases where it is not allowed.
@Flag(name: .customLong("netrc"), help: "Use netrc file even in cases where other credential stores are preferred.")
public var forceNetrc: Bool = false
```

and `getRegistryAuthorizationProvider` consumes it as `authorization.keychain = options.security.forceNetrc ? .disabled : .enabled`. So it is not a no-op, it flips registry precedence from keychain-first to netrc-first.

The failure mode is quiet. Someone with a stale keychain entry from an old `swift package-registry login`, passing `--netrc` to force the fresh credential in their netrc file, still gets the stale keychain credential from `Registry.swift:140` and a 401, with nothing to indicate the flag was dropped. Given the care taken over `--netrc-optional` for exactly this reason, I suspect you would rather implement it than swallow it. I have left a note on `Registry.swift` about the capability gap this leaves.

While I was in that function I noticed a smaller thing worth a line in the description. `getRegistryAuthorizationProvider` never reads `options.security.netrc`, only `netrcFilePath` and `forceNetrc`, and the upstream doc comment says the enable/disable pair "has no effects on registry communications". Here `--disable-netrc` does switch off registry netrc, since both call sites read the same resolved value. That seems like a reasonable divergence rather than a bug, it just is not in the "kept on purpose" list yet.

swifterpm/Sources/swifterpm/Netrc.swift
```diff
@@ -0,0 +1,212 @@
...
+                    machines.append(machine)
+                }
+            default:
+                tokens.removeFirst()
+            }
+        }
+        return machines
+    }
+
+    private static func parseMachine(name: String, tokens: inout [String]) -> NetrcMachine? {
```
esnunes, line 175: One thing I wanted to flag on the tokenizer, and it is not a regression so definitely not a blocker.

`parseMachine` takes whatever token follows `login` or `password` as the value, and because `tokenize` has already erased line boundaries, a missing value quietly consumes the next entry's `machine` keyword instead. Two inputs reach that state: an empty quoted value, since the tokenizer never emits an empty token, and a value beginning with `#`, which opens a comment under the new rule.

The part that caught my eye is the blast radius. It is not the one malformed entry, it is the rest of the file. I pulled `NetrcParser` out verbatim and ran it:

```
machine first.example.com  login user  password ""
machine second.example.com login u2    password p2
```

comes back as a single machine, `first.example.com` with the password `machine`, and `second.example.com` is gone. `login ""` is worse: the parser returns zero machines for the whole file, so every credential in it is silently dropped and the requests go out unauthenticated.

This all predates the PR, it came across from `RegistryNetrc`. I am raising it here because the PR's own argument for erroring on a missing `--netrc-file` is that credentials should never downgrade silently, and this is the same failure reachable by a typo instead of by configuration.

A cheap guard would be to refuse a value that is itself an entry keyword, which keeps the damage inside the malformed entry:

```swift
case "login":
    login = tokens.popFirstValue()
case "password":
    password = tokens.popFirstValue()
```

where `popFirstValue` returns nil when `first` is `machine` or `default`. Would pair well with a test next to `parserSkipsEntriesWithoutBothCredentials`.

swifterpm/Sources/swifterpm/Netrc.swift
```diff
@@ -0,0 +1,212 @@
...
+
+    init(sources: [NetrcSource]) {
+        self.sources = sources
+    }
+
+    static func resolve(
+        _ configuration: SwifterPMNetrcConfiguration,
+        environment: [String: String]
+    ) async throws -> Netrc {
+        guard configuration.isEnabled else { return .empty }
```
esnunes: One judgement call worth making deliberately rather than by accident: `--netrc-file` combined with `--disable-netrc`.

`Netrc.resolve` returns `.empty` on this `isEnabled` guard before it looks at `configuration.path`, so the pair silently ignores the file. SwiftPM reports it instead, `SwiftCommandState` emits `.mutuallyExclusiveArgumentsError(arguments: ["--disable-netrc", "--netrc-file"])` when `netrcFilePath` is set and `netrc` is false, which fails the command.

This is the mirror image of the `--netrc-optional` argument in the description. There the reasoning was that swallowing something SwiftPM rejects makes us more permissive than the tool we stand in for, which is exactly what happens here. The new `resolveParsesNetrcOptions` test passes both flags together and asserts they parse, so the permissive reading is pinned by a test now.

`resolveIgnoresAMissingNetrcFileWhenDisabledOrUnconfigured` clearly encodes the other intent on purpose, so I may well be arguing against a decision you already made. If so, a line in the divergences section would settle it. If not, a validation in `CLIRunner.netrcConfiguration` throwing a `ToolError` would match upstream.

swifterpm/Sources/swifterpm/Registry.swift
```diff
@@ -129,24 +129,20 @@ enum RegistryAuthorization {
             )
         }
 
-        if let netrcData = nonEmpty(environment["SWIFTPM_NETRC_DATA"]),
-           let credential = RegistryNetrc(content: netrcData).credential(for: url)
-        {
+        // Inline netrc data, then the keychain, then any netrc file. SwiftPM's
```
esnunes: Small accuracy point on this comment. It describes upstream as reading `SWIFTPM_NETRC_DATA` first and then ordering the rest, but `makeRegistryAuthorizationProvider` does not order them, it picks one. A non-empty `SWIFTPM_NETRC_DATA` returns immediately, so the keychain and the netrc file are never consulted, and the function ends with `return providers.first` under the comment "Use at-most one AuthorizationProvider (i.e., no CompositeAuthorizationProvider)". With the keychain enabled, which is the macOS default, upstream never reads the netrc file for registry requests at all.

We fall through on every miss instead. That is a real behavioural difference: if CI sets `SWIFTPM_NETRC_DATA` covering registry A and the build also talks to registry B, SwiftPM sends B nothing while we fall through to the keychain or a `default` entry in the netrc file and send B credentials. It turns a clean "no auth" into a confusing "invalid credentials".

The fall-through predates this PR, and I would argue it is the better behaviour anyway since it is what lets `--netrc-file` reach registry auth at all, which is the whole point of the change. So I would keep the code and reword the comment, so the next person reading it does not take it as a parity claim.

One related capability worth capturing while we are here. There is no way to prefer netrc over the OS credential store on macOS when the keychain holds a credential for the same host. Upstream models exactly that, and it is `--netrc` / `forceNetrc`, the flag currently on the ignored list. So I would read that other note as a small feature gap rather than a parity nit.

Unrelated, but I noticed the missing-file message matches the registry factory's lowercase wording exactly rather than the download factory's capitalised one. Nice.

swifterpm/Sources/swifterpm/Registry.swift
```diff
@@ -129,24 +129,20 @@ enum RegistryAuthorization {
...
 
-        if let home = environment["HOME"] {
-            let netrcPath = URL(fileURLWithPath: home).appendingPathComponent(".netrc")
-            if let data = try? await fileSystem.readFile(at: netrcPath.absolutePath),
-               let content = String(data: data, encoding: .utf8),
-               let credential = RegistryNetrc(content: content).credential(for: url)
-            {
-                return header(for: credential, url: url, registryConfig: registryConfig)
-            }
+        if let credential = Environment.netrc.credential(for: url, from: .file) {
```
esnunes: The registry ordering here, inline netrc data then keychain then netrc file, is the behaviour the last commit is about, and I could not find a test that pins it.

`RegistryKeychain` is private and reads the real keychain, so nothing can get between the two netrc lookups from a test today. `credentialFromAnOriginIgnoresTheOtherSources` covers the origin filtering nicely, but not the order the three sources are consulted in. Reorder these lines and the suite stays green.

The download path already solved this: `HTTPAuthorization.prioritizedHeader` is a pure function over its inputs, so a test can assert precedence directly with no environment and no keychain. The same shape would work here, something taking the three optional credentials and returning which wins, with `header(for:registryConfig:)` reduced to fetching them and calling it.

Not blocking. It just seems a shame for the ordering to be the one part of this change without a regression test, given it is upstream-derived and deliberately counter-intuitive.

esnunes (approved):
This is a really nicely argued change, and the parts of it that are right are right for good reasons. The dependency survey, the observation that SwiftPM's two authorization factories deliberately disagree, and the decision to error on a missing `--netrc-file` rather than fall through to unauthenticated requests all read as considered rather than assumed. Consolidating the two netrc copies into one type was worth doing on its own, and I verified the two parser corrections that land against `main`: first-match on a duplicated host, and keeping `se#cret` whole.

Approving, since none of what follows needs to hold the change up. Two things are worth a follow-up though, and I checked both rather than guessing.

The first is that the new flags do not reach the solver. Details at the bottom, since `Resolve.swift` is not part of this diff and I could not leave that one inline.

The second is `--netrc`. The premise for putting it on the ignored list is that SwiftPM deprecated it and it only restates `--enable-netrc`. In `SecurityOptions` on both `release/6.3` and `main` it is a separate, undeprecated flag, `forceNetrc`, and `getRegistryAuthorizationProvider` uses it to disable the keychain. So stripping it inverts registry precedence silently. See the notes on `CLI.swift` and `Registry.swift`.

The rest are smaller and left inline: a parser edge case where a missing value consumes the rest of the file rather than the one entry, a comment that reads upstream's registry provider as an ordering when it actually selects at most one provider, the `--netrc-file` plus `--disable-netrc` combination that SwiftPM rejects and we accept, and no regression test on the registry precedence.

One note on the description rather than the code. Of the four behaviours listed as brought into line with SwiftPM, two are visible against `main`. The registry ordering was already inline netrc data, then keychain, then netrc file at the base commit `cf7d45f05`, so that section describes a regression introduced and repaired inside this branch. Same for the `SWIFTPM_NETRC_DATA` suppression. Worth knowing so the description is not read as claiming more surface than the diff has.

For what it is worth, `swift build` is clean and all 14 `NetrcTests` pass locally.

---

### The netrc flags do not reach the child `swift package resolve`

Leaving this here rather than inline, because `Resolve.swift` is not part of the diff.

The configuration reaches swifterpm's own clients through the `Environment.netrc` task local, and that half is solid. But the version solve is still delegated to a child `swift package resolve` at `Resolve.swift:107`, and `swiftPackageResolveArguments` (`Resolve.swift:134`) forwards `--config-path`, `--default-registry-url`, `--disable-sandbox` and the SCM transformation flags without forwarding `--netrc-file` or `--disable-netrc`, both of which real SwiftPM accepts.

Solving a graph with a private registry dependency means fetching availability and manifests from that registry inside the child process, so the flag does not cover the requests the original bug report was about. Concretely, `swifterpm resolve --netrc-file /tmp/ci-netrc` against a private registry needing a fresh solve: swifterpm's own downloads authenticate, the child reads `~/.netrc` or nothing and comes back with a 401 or 404, and the user has every reason to believe credentials were configured. `--disable-netrc` has the mirror problem, the child still reads `~/.netrc`.

I think this is also why the customer's `SWIFTPM_NETRC_DATA` workaround worked. The description reads it as bypassing the flag, but environment variables are inherited by the child (`Support.swift:147`) and the flags are not, so it was the only mechanism that covered both halves.

The fix looks mechanical, thread the configuration through and append the flags next to `--config-path`. Worth checking `assertResolvedFileUpToDate` at `Resolve.swift:297` as well, since it spawns the same child.

swifterpm/Sources/swifterpm/CLI.swift
fortmarek: You are right, `forceNetrc` is undeprecated and disables the keychain for registry requests, so `--netrc` is off the ignored list and implemented as a real flag.

swifterpm/Sources/swifterpm/Netrc.swift
fortmarek: Reproduced exactly as you described, `first.example.com` with the password `machine`, and values now refuse a token that opens an entry so the damage stays inside the malformed one.

swifterpm/Sources/swifterpm/Netrc.swift
fortmarek: Not a decision I had made deliberately, so `Netrc.resolve` now throws the upstream message, placed there rather than in the CLI so embedders hit it too.

swifterpm/Sources/swifterpm/Registry.swift
fortmarek: Reworded: upstream selects at most one provider, we fall through, and the comment now says so instead of reading it as an ordering.

swifterpm/Sources/swifterpm/Registry.swift
fortmarek: Extracted `prioritizedCredential` as a pure function over the three sources with the keychain behind a closure, so the ordering has a test that fails if the lines are reordered.

PR #12327, opened 2026-08-13
fix(infra): repack bloated cache masters so the per-master cap is a ceiling again

fortmarek:
Follow-up to #12290. That PR made dispatch place work by *reported* cache-master residency, and the telemetry it added showed the placement logic working correctly with almost nothing to place onto: hosts hold 0.55 masters each. This is why.

## The finding

The runner-cache volume's sizing arithmetic assumes `masterCapGib` is "a sparse per-master ceiling, not the typical size", reasoning from "a real Tuist cache master is ~2.4 GB". Production disagrees.

Measured across the nine production runner hosts:

| host state | free of 80 GiB quota |
|---|---|
| 0 masters, idle | **79.8 GiB** |
| 0 masters, job running | 71.2 GiB (so a live branch is ~8.6 GiB) |
| 1 master | **52.2 GiB** |

Netting the branch off leaves roughly **19 GiB for one master against a 20 GiB cap**. Each master costs its entire ceiling.

Combined with admission reserving `cap x (liveBranches + 1)` = 40 GiB, that leaves room for about one master per host. The fleet-wide master count peaked at 22 on 2026-07-29 and has sat at 5-7 ever since. Warm materializes are 18% over 7d, and account 3's cache HEAD has not advanced in 15 days despite ~1,200 macOS jobs a week, because no host holds its master any more.

## This was predicted

It is not a leak; it is how sparse images behave, and it was measured and written up on 2026-07-18 while validating ASIF and TRIM: freed blocks are not reused within a mounted session, and nothing returns them on delete, on detach, on mounted idle, or via `diskutil image resize`. A long-lived hot master therefore grows toward the cap even though the cache inside is bounded by the CLI's own LRU.

The decision then was to hold, on the explicit condition: *"bloat is bounded by the per-master cap and resolved by LRU watermark eviction (cost = one cold rebuild for that account/host). Revisit if it bites in practice."* This is that revisit, with the production numbers.

## Why repack and not `hdiutil compact`

The same validation measured naive compaction: it reclaimed **4 MB of ~400 MB** on a partially-churned image. APFS inside an image issues no TRIM to the virtual block device, so `hdiutil compact` only reclaims an image that has been emptied outright — precisely the opposite of a hot master. Reaching for the obviously-named tool here would have shipped a no-op.

The mechanism that does work, and the one that validation recommended if this was ever revisited, is a **host-side repack**: `ditto` the live cache into a fresh sparse image. The new file's size is the live content, whatever slack the old one accumulated. `ditto` rather than `cp -R` because it preserves xattrs on symlinks — the whole reason this cache lives in a disk image instead of on the virtio-fs share.

It is purely host-side, so the guest hot path is untouched. The alternative that also reclaims (guest `-mountoptions trim` plus host compact) would have changed the guest mount path on a feature with a fragile production history.

## Shape

`CompactBloatedMaster` runs on the existing reconcile tick, after eviction, and repacks at most one master per tick. A repack copies live cache, so it is seconds of I/O rather than the metadata-only work everything else on that tick does.

The concurrency design is the part worth reviewing:

- The copy runs **outside** the manager lock, against a read-only attach of the master.
- Only the swap takes the lock, and it re-checks that the master is still the same file at the same generation before committing. A promote or converge that lands mid-repack wins and the repacked image is dropped. Losing that race costs one wasted repack, never a rolled-back master.
- The swap restores the original mtime. LRU orders by it, so a repack must not look like a fresh use, or a bloated cold master would outlive one that jobs are actively cloning.

## Making the assumption checkable

Two new metrics, because the thing that went wrong here is that nobody could see a master's physical size:

- `tart_kubelet_cache_volume_compacted_bytes_total`
- `tart_kubelet_cache_volume_largest_master_bytes`

The latter read against the configured cap distinguishes bloat from cache at a glance. It is what would have caught this without a production dig.

`masterEntry` now also carries the volume name — masters are keyed by (account, volume) on disk, and anything acting on one specific master rather than counting or deleting whole directories needs it.

## How to test locally

```bash
cd infra/tart-kubelet && go test ./internal/podagent -run 'TestCompact|TestLargestMaster' -v
```

## Validation run

- 5 new tests: slack is reclaimed while cache content and generation survive verbatim; eviction order is preserved; lean masters are left alone; a concurrent promote beats an in-flight repack; the largest-master gauge tracks the bloated master.
- `go build ./...`, `go vet ./...`, `gofmt -l`, `go test ./...` all clean.
- The darwin repack path itself is not unit-testable off a Mac. The mechanism was validated on macOS 26.3 on 2026-07-18: 703 MB reclaimed on a churned image, xattr-on-symlink fidelity intact, ~1s for 400 MiB of live content.

## What this does not fix

Two related problems found in the same investigation, deliberately left out:

1. **Cold jobs can never publish.** A cold job promotes with `base_generation = 0`, which `establish_first_head` accepts only when the account has no HEAD at all. Otherwise it 409s, and `Finalize` then discards the branch — so the host stays cold and the cache it just built is thrown away. At least 383 of 601 rejected promotes in 7d are this, not contention. Fixing it means keeping a rejected cold branch as an unpublished local master.
2. **The rejection metric conflates two things.** `rejected` is documented as "genuine cross-host contention" but also counts the structural case above, which is why a 95% rejection rate read as a tuning problem for weeks.

Compaction is the prerequisite: neither is worth much while masters cannot survive on disk in the first place.

## Rollout note

This is a tart-kubelet change, so it takes effect as the fleet picks up a new build, not on a server deploy. Effect to watch: `largest_master_bytes` falling well below the cap, then resident masters per host rising above one, then warm materialize share following.


fortmarek:
Closing without merging.

The work here is sound and reviewed, but it targets the wrong link in the chain. The premise was that masters cost their full ceiling because of sparse-image bloat, so reclaiming that slack would let more accounts stay warm per host. Two things came out of digging further:

**1. Masters at the ceiling are what the accounting already assumes.** The survivor formula `gib/cap - (B+1)` treats each master as costing a full cap. A ~19 GiB master against a 20 GiB cap is not over budget. The claim I was reacting to ("a real Tuist cache master is ~2.4 GB, so 80 GiB holds ~30 masters") sits in the same comment but was never the load-bearing one. The real constraint is that `cap 20` with `gib 80` and a cap-based branch reservation leaves room for one master, which is arithmetic and needs no measurement:

```
80 GiB quota − 40 GiB reservation (cap × 2) − 8.6 GiB live branch ≈ 31 GiB → 1 master
```

**2. Capacity is not the binding constraint anyway.** Convergence is the only way a host with no master can obtain one, because a cold job's promote (base generation 0) is rejected whenever a HEAD already exists and its branch is then discarded. Production did 286 cold materializes and 1 convergence in 24h. Room for more masters does not help when almost none are being created.

So the ordering is: find out why convergence fails, then size the volume so a converged master survives, then compaction as a refinement.

#12336 does the first step. The branch is left in place, so the repack and the live-bytes-per-tree metric can be revived as their own PR once convergence is understood and the sizing is re-derived. The review feedback on it (CAS preservation, non-darwin build, admission reservation, stale mtime, detach errors) is all incorporated in the commits here and should be carried over with them.

PR #12328, opened 2026-08-13
fix(server): narrow the runners overview cards to work that actually ran

fortmarek:
The Recent jobs card on the runners overview was showing nothing but skipped jobs, every row reading 0ms, while the Successful and Failed legends both showed 0. The Recent workflows card above it had the same problem, listing Skipped and Cancelled runs with a `None` duration.

### What changed

Both cards queried for completed work without constraining the conclusion. GitHub reports a job gated out by an `if:` condition as completed with conclusion `skipped`, without a runner ever claiming it, and a workflow run rolls up to `skipped` when every job in it was gated out. None of that work has a duration.

Meanwhile the charts and the Successful/Failed legends on both cards already ignored every conclusion outside success/failure. So each card described two different sets of work: rows that no bar and no legend accounted for, sitting in tables that hold only five entries. On an account whose recent activity is mostly gated preview workflows, that is the whole card.

Both cards now narrow to `["success", "failure"]` at the query, so the table, the bars and the legends describe the same runs.

`cancelled` goes with `skipped`, and that part is a deliberate call rather than a consequence: the rollup ranks `cancelled` above `success`, so a run where nine jobs succeeded and the tenth was cancelled reads `cancelled` and carries a real duration, and it is excluded all the same. Including it would put rows on the card that neither legend counts, and the Jobs and Workflows pages already list them behind their status chips. A test pins that mixed shape so the choice is not silently revisited.

Because the filter is total rather than advisory, an empty card no longer implies the account never ran anything. Each card asks, only when it comes back empty, whether the account has anything at all in the same repository/platform scope. If it does, the copy says the recent work was skipped or cancelled, the onboarding link drops, and `View more` stays enabled so the pages listing those rows remain one click away.

Supporting changes in `Tuist.Runners.Jobs`:

- `list_for_account/2` could only pin a single conclusion, so `:conclusion` now also accepts a list. That is a smaller surface than a second, opposite-signed option beside it, and the only other caller, the conclusion filter chip on the Jobs page, passes a binary and is unaffected. `count_for_account/2` shares the filter and keeps its documented promise of matching `list_for_account/2`.
- `list_recent_workflow_runs_for_account/2` gains the same option. Its filter has to sit outside the rollup rather than in the `having` clause, because the run-level conclusion is a column the rollup computes: filtering in the `having` would mean restating that nested `if(countIf(...))` expression.

Filtering in the query rather than after the fetch matters for both: the limit is applied by ClickHouse, so dropping rows afterwards would shrink each card below its intended size, down to empty for an account whose recent completed work was all gated, which is exactly the state that surfaced this.

With the queries as the single gate, the now-redundant in-memory filters in `recent_jobs_chart_data/2` and `recent_workflow_runs_chart_data/2` are removed.

### Impact

Both cards on the runners overview list work that actually ran. The Jobs and Workflows pages, where the Skipped and Cancelled filter chips are deliberate features, are unchanged.

### How to test locally

Open the runners overview for an account whose recent activity includes gated workflows, for example `/tuist/runners?analytics-date-range=last-24-hours`, and confirm both cards list work with real durations instead of 0ms and `None` rows.

```
mix test test/tuist/runners/jobs_test.exs test/tuist_web/live/runners_live_test.exs test/tuist_web/live/runner_jobs_live_test.exs test/tuist_web/live/runner_workflow_live_test.exs
```

- `narrows to any of the conclusions when :conclusion is a list` covers the jobs query, including the mirrored `count_for_account/2`.
- `narrows to the given conclusions, applying the limit after the filter` covers the workflow-run rollup, and pins the limit-after-filter ordering by asking for one row while the two most recent runs are skipped and cancelled.
- `lists only succeeded and failed jobs in the recent jobs card` seeds a skipped job through `Jobs.record_completed/2`, the real webhook path for a gated job with no preceding queued row, plus a cancelled one, and asserts neither reaches the rendered table. `lists only succeeded and failed runs in the recent workflows card` does the same for the card above.
- `drops a run whose jobs mostly succeeded but one was cancelled` pins the deliberate exclusion described above.
- `tells an account whose recent work was all filtered out from one with no jobs` covers the empty state: the filtered copy, no onboarding link, and both `View more` buttons still enabled.
- Every refutation was confirmed to fail without its filter, so none of them pass vacuously. The pre-existing `charts only succeeded and failed runs` now covers the removal of the in-memory chart filters: it fails if either query-level filter is dropped.

### DIFF
--- a/server/lib/tuist_web/live/runners_live.ex
+++ b/server/lib/tuist_web/live/runners_live.ex
@@ -192,8 +192,11 @@ defmodule TuistWeb.RunnersLive do
   end
 
   defp assign_recent_workflow_runs(socket, account_id, repository, platform) do
+    # Same contract as the Recent jobs card above: a run rolls up to
+    # `cancelled` or `skipped` when none of its jobs ran, so it carries
+    # no duration for the table and nothing for the bars or legends.
     opts =
-      [limit: @chart_limit]
+      [conclusion: @duration_conclusions, limit: @chart_limit]
       |> maybe_repository(repository)
       |> maybe_platform(platform)
 
@@ -223,7 +226,6 @@ defmodule TuistWeb.RunnersLive do
   # just don't carry a navigate target.
   defp recent_workflow_runs_chart_data(recent_workflow_runs, account_name) do
     recent_workflow_runs
-    |> Enum.filter(&(&1.conclusion in @duration_conclusions))
     |> Enum.reverse()
     |> Enum.map(fn run ->
       %{
@@ -336,10 +338,11 @@ defmodule TuistWeb.RunnersLive do
   def platforms, do: @platforms
 
   @doc """
-  Conclusion label for the Recent workflow_runs table — the rollup
-  in `list_recent_workflow_runs_for_account/2` may return `success`,
-  `failure`, `cancelled`, or `skipped`. Anything outside that set
-  collapses to "Unknown" so a stray value renders cleanly.
+  Conclusion label for the Recent workflow_runs table. The card asks
+  `list_recent_workflow_runs_for_account/2` for `@duration_conclusions`
+  only, but the rollup can also return `cancelled` or `skipped`, so
+  those keep a label; anything outside the set collapses to "Unknown"
+  so a stray value renders cleanly.
   """
   def conclusion_label("success"), do: dgettext("dashboard_runners", "Success")
   def conclusion_label("failure"), do: dgettext("dashboard_runners", "Failure")
--- a/server/lib/tuist_web/live/runners_live.ex
+++ b/server/lib/tuist_web/live/runners_live.ex
@@ -19,16 +19,14 @@ defmodule TuistWeb.RunnersLive do
 
   @table_limit 5
   @chart_limit 30
-  # Only runs that reached a success/failure conclusion carry a real
+  # Only work that reached a success/failure conclusion carries a real
   # duration. Running, cancelled and skipped runs would draw as
-  # zero-height bars that scatter the real ones, so they are kept out of
-  # the chart trail (and are not counted by the success/failure legends).
-  @chart_conclusions ["success", "failure"]
-  # A job gated out by an `if:` condition is reported as
-  # completed/skipped without a runner ever claiming it. It represents
-  # no work, so it stays out of the card entirely rather than filling
-  # the table with 0ms rows.
-  @excluded_recent_conclusions ["skipped"]
+  # zero-height bars that scatter the real ones, and count for nothing
+  # in the success/failure legends. The Recent jobs card narrows to
+  # these at the query so its table, bars and legends all describe the
+  # same set of jobs; the workflow-run chart drops the rest before
+  # plotting.
+  @duration_conclusions ["success", "failure"]
 
   @impl true
   def mount(_params, _session, %{assigns: %{selected_account: selected_account, current_user: current_user}} = socket) do
@@ -171,14 +169,15 @@ defmodule TuistWeb.RunnersLive do
 
   defp assign_recent_jobs(socket, account_id, repository, platform) do
     # The Recent jobs card mirrors the Recent Test Runs card on the
-    # Tests page — it's a chronicle of finished work, not a live
-    # status board. Filter to completed runs only so the bars carry
-    # a real duration and the success/failure legends count
-    # something other than zero. The chart spans up to
-    # `@chart_limit` so the bar trail conveys a trend, while the
-    # table below shows only the freshest `@table_limit` rows.
+    # Tests page — it's a chronicle of work that ran, not a live
+    # status board. Filter to jobs that completed with a
+    # `@duration_conclusions` conclusion so the bars carry a real
+    # duration and the success/failure legends count something other
+    # than zero. The chart spans up to `@chart_limit` so the bar trail
+    # conveys a trend, while the table below shows only the freshest
+    # `@table_limit` rows.
     opts =
-      [status: "completed", exclude_conclusions: @excluded_recent_conclusions, limit: @chart_limit]
+      [status: "completed", conclusion: @duration_conclusions, limit: @chart_limit]
       |> maybe_repository(repository)
       |> maybe_platform(platform)
 
@@ -224,7 +223,7 @@ defmodule TuistWeb.RunnersLive do
   # just don't carry a navigate target.
   defp recent_workflow_runs_chart_data(recent_workflow_runs, account_name) do
     recent_workflow_runs
-    |> Enum.filter(&(&1.conclusion in @chart_conclusions))
+    |> Enum.filter(&(&1.conclusion in @duration_conclusions))
     |> Enum.reverse()
     |> Enum.map(fn run ->
       %{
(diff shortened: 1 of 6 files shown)


server/lib/tuist_web/live/runners_live.ex
```diff
@@ -188,8 +192,11 @@ defmodule TuistWeb.RunnersLive do
   end
 
   defp assign_recent_workflow_runs(socket, account_id, repository, platform) do
+    # Same contract as the Recent jobs card above: a run rolls up to
+    # `cancelled` or `skipped` when none of its jobs ran, so it carries
+    # no duration for the table and nothing for the bars or legends.
     opts =
-      [limit: @chart_limit]
+      [conclusion: @duration_conclusions, limit: @chart_limit]
```
esnunes, line 215: Really like how this lands — one gate, and the three parts of the card can't drift apart again.

One thing I wanted to flag about the `cancelled` half of the exclusion. The description frames it as "cancelled when it was cancelled before any job ran", but the rollup ladder puts `cancelled` above `success`, so any run with at least one cancelled job and no failure resolves to `cancelled` — including a run where nine jobs succeeded and the tenth got cancelled. Those runs do carry a real `duration_ms`, and they now drop off the card entirely. We set `cancel-in-progress` in 54 of the 69 workflows in this repo, so a run getting cancelled mid-flight is pretty routine.

Same question one level down for jobs: the first commit kept cancelled jobs because they "can carry real runner time", and the second dropped them to make the table agree with the legends. A job cancelled after eight minutes still burned eight minutes of runner time, which feels like something an account would want to see on this card.

I don't think either is obviously wrong — "only success/failure" is a clean rule and the legends genuinely can't count anything else. But if the intent is the narrower "work that never ran", `["success", "failure", "cancelled"]` with a duration guard might match it more closely. Worth deciding deliberately either way, and maybe worth a test with a mixed run (some success, one cancelled) since right now only the all-cancelled shape is covered.

server/lib/tuist_web/live/runners_live.ex
```diff
@@ -166,14 +169,15 @@ defmodule TuistWeb.RunnersLive do
...
+    # Tests page — it's a chronicle of work that ran, not a live
+    # status board. Filter to jobs that completed with a
+    # `@duration_conclusions` conclusion so the bars carry a real
+    # duration and the success/failure legends count something other
+    # than zero. The chart spans up to `@chart_limit` so the bar trail
+    # conveys a trend, while the table below shows only the freshest
+    # `@table_limit` rows.
     opts =
-      [status: "completed", limit: @chart_limit]
+      [status: "completed", conclusion: @duration_conclusions, limit: @chart_limit]
```
esnunes, line 179: Something worth thinking about now that the filter is total rather than advisory: the card can come back with zero rows for an account that has plenty of jobs.

When that happens the empty branch renders `widget_empty_label/0` — "No jobs yet" — with a "Get started" link to the docs, and "View more" is `disabled={Enum.empty?(@recent_jobs)}`. So the account whose last thirty completed jobs were all gated gets told it has no jobs, offered onboarding it doesn't need, and has the one link to the Jobs page (where those jobs are listed, behind the Skipped/Cancelled chips) switched off. It's a bit of a mirror image of the bug being fixed — and the account described in the PR body, "recent activity is mostly gated preview workflows", is exactly the shape that could land there.

I haven't measured how many accounts would actually hit zero, so I can't say how often this fires in practice. But the three `Enum.empty?` call sites in the template (530, 534, 662) were all written when empty could only mean "never ran a job", and that's no longer true.

Cheapest fix is probably a second unfiltered count so the empty state can tell "no jobs" from "no jobs that ran"; even just leaving "View more" enabled and softening the copy would remove the dead end. Happy to leave it for a follow-up if you'd rather keep this PR tight.

server/lib/tuist_web/live/runners_live.ex
```diff
@@ -19,11 +19,14 @@ defmodule TuistWeb.RunnersLive do
...
-  # Only runs that reached a success/failure conclusion carry a real
+  # Only work that reached a success/failure conclusion carries a real
   # duration. Running, cancelled and skipped runs would draw as
-  # zero-height bars that scatter the real ones, so they are kept out of
-  # the chart trail (and are not counted by the success/failure legends).
-  @chart_conclusions ["success", "failure"]
+  # zero-height bars that scatter the real ones, and count for nothing
+  # in the success/failure legends. The Recent jobs card narrows to
+  # these at the query so its table, bars and legends all describe the
+  # same set of jobs; the workflow-run chart drops the rest before
```
esnunes: Small one: the comment above `@duration_conclusions` ends with "...the workflow-run chart drops the rest before plotting". That was accurate when you wrote it — at that point only the jobs card had moved to a query-level filter — but the next commit removed that `Enum.filter` from `recent_workflow_runs_chart_data/2` and moved the workflow card to the query too. Both cards gate at the query now, so the comment sends the reader looking for an in-memory filter that isn't there any more.

Two other leftovers from the same removal if you want them in the same pass: `workflow_run_chart_color("cancelled")` (:245) and `chart_color_for(%{status: "completed", conclusion: "cancelled"})` (:298) are unreachable from these cards now. Fine to leave as defensive clauses — you already set the pattern for it with the `conclusion_label/1` docstring note, so maybe just the same treatment.

esnunes (approved):
This is a good fix, and I think the diagnosis is the valuable part: the card wasn't just noisy, it was describing two different sets of work at once, and the five-row table meant the noise crowded out everything real rather than merely sitting next to it. Moving the gate into the query so the table, the bars and the legends can't drift apart again is the right shape.

Two details I particularly liked. Replacing `:exclude_conclusions` with a polymorphic `:conclusion` in the follow-up commit avoids two opposite-signed options on the same column, which would have needed a precedence rule nobody wants to document. And the limit-after-filter test is a genuine refutation rather than a restatement — asking for `limit: 1` with the two most recent runs skipped and cancelled fails loudly if the ordering ever regresses. Same for the pre-existing chart test now covering the removed in-memory filters.

I've left a few notes inline. The two worth a decision are both about the edges of the new filter rather than the mechanism: runs cancelled mid-flight roll up to `cancelled` even when their jobs ran, so they now drop off the card along with the ones that never started; and a card that returns zero rows falls into an empty state written for an account that has never run a job. Neither is obviously wrong — they're both places where "work that actually ran" and "success or failure" aren't quite the same set, and it's worth being deliberate about which one the card means. The other note is minor.

server/lib/tuist_web/live/runners_live.ex
fortmarek: Right on the rollup ladder — the PR description was wrong and I have corrected it. Decision after checking with Marek: keep success/failure only, so a mostly-succeeded run with one cancelled job stays off the card deliberately (both pages still list it behind their status chips); added a mixed-run test in `jobs_test.exs` pinning exactly that shape instead of only the all-cancelled one.

server/lib/tuist_web/live/runners_live.ex
fortmarek: Fixed in f4875be — each card now asks (only when it comes back empty) whether the account has anything in the same scope, and if so says the recent work was skipped or cancelled, drops the "Get started" link and leaves "View more" enabled so the Jobs/Workflows pages stay reachable.

server/lib/tuist_web/live/runners_live.ex
fortmarek: Both fixed in f4875be — the `@duration_conclusions` comment now says both cards gate at the query, and `workflow_run_chart_color/1` and `chart_color_for/1` carry the same "defensive clause" note as the `conclusion_label/1` docstring.

PR #12330, opened 2026-08-13
feat(mcp): honour operator grants presented in a header

fortmarek:
Tuist operators are not members of the customer accounts they investigate — that is precisely what `ops_access` and the ops.tuist.dev grant flow exist for. But that flow is browser-shaped, and the MCP endpoint is not:

- `operator_grant` is a **virtual** field on `User`, populated only by `load_operator_grant/2` or the LiveView `on_mount`, both of which read the session cookie *and* require an `account_handle` path param.
- The `:mcp` pipeline is `load_authenticated_subject` → `require_authentication` → rate limit. No `fetch_session`, no grant plug. MCP tool calls carry a record id, not an account handle.

So over MCP `current_user.operator_grant` is always `nil`, `Checks.ops_access/2` falls through to its catch-all `false`, and the read policies are left with: public project, project membership, matching project token, or a customer account token. An operator is none of those for a customer's project.

**Concretely, today:** an operator asking the MCP for a customer's failing test run gets "You do not have access to this resource." The tools work only against projects they are personally a member of.

This accepts the grant in an `x-tuist-operator-grant` header, verified by the same `verify/1` the browser flow uses — offline Ed25519, EdDSA-strict, TTL ceiling, `iss`/`aud` pinning, honoured only for the operator named in `sub`, pinned to the one account it names.

**Authorization is unchanged.** `operator_grant_covers?` already compares the grant's `account_id` against `object_account_id(object)`, and MCP tools pass the project as that object — so "refuse anything outside the granted account" is the existing check, not new logic.

Three decisions worth a reviewer's attention:

- **Nothing is stored between requests.** The browser keeps the grant in the session because a person clicks through many pages after justifying access once; a token-authenticated caller has nowhere to keep it. The grant rides every call, which also means it cannot outlive its `exp` in a cookie.
- **A rejected grant fails the request** rather than continuing unprivileged. The browser strips the token and carries on because a person sees the page they landed on; an agent would instead see a bare "you do not have access" and no way to distinguish a rejected grant from a missing one.
- **Google authentication is required, and must be recent.** The browser will not attach a grant unless the session authenticated through Google; an access token carries no such signal. `users.last_google_authenticated_at` is written in `log_in_user/3` — the one funnel every sign-in passes through with `auth_method` in hand — and a grant is honoured only if it is within the grant's own TTL ceiling. That window fits the existing flow: minting a grant at ops.tuist.dev already requires a Google-authenticated session, so an operator who just justified access satisfies it by construction. A stolen grant on a password-authenticated operator token is refused with `stale_google_authentication`.

  The residual gap, stated plainly: this attests that the *person* recently proved Google, not that *this token* was obtained through it. Putting the authentication method on the token is not reachable from this repo — Boruta's `ResourceOwner.extra_claims` feeds only ID tokens, and `TokenGenerator.generate/2` receives `%Boruta.Ecto.Token{sub, client_id, scope}`, so the authorization's session context never arrives. Closing it means changing the upstream dependency.

Companion to #12324 (artifact download URLs on the run tools), which is what makes reaching a customer's `.xcresult` useful once access works. Both bump `serverInfo.version` to 1.15.0 per `lib/tuist/mcp/AGENTS.md`, so whichever merges second needs the next number.

**Known gap on the Atlas side:** tuist/atlas#401 proxies this server's MCP per user, but its proxy sends only static config headers — there is no per-user, per-request hook, so Atlas cannot present a grant yet. This PR makes grants work for any client that can set a header (Claude Code pointed straight at `/mcp` today); Atlas obtaining and forwarding grants is separate work.

### How to test locally

```
mix test test/tuist_web/operator_grant_plugs_test.exs test/tuist_web/operator_grant_test.exs
```

45 tests pass. The new ones cover a grant attached for its operator, that it authorizes the granted account and refuses another, a grant minted for someone else, one presented by a non-operator, a wrong signing key, an unknown account, and no authenticated user.


pepicrft (changes requested):
The grant fallback drops the bearer token’s scope checks. The authenticated account can carry any [Model Context Protocol](https://modelcontextprotocol.io/) scope, while that endpoint only checks that it is authenticated. Once its normal policy check fails, this authorizes the separate operator user from the grant.

As a result, an authorization token without the `mcp` scope, or one restricted to an unrelated read scope, can invoke customer read tools whenever the operator presents a grant. The grant should widen account visibility only, not what the credential can do. Please gate the fallback on an `mcp`-scoped authenticated account and add a test using a non-`mcp` scope.

fortmarek:
Good catch — fixed in 35991a7d28.

`operator_grant_authorizes_read?/4` now requires the authenticated account to carry the `mcp` scope before the grant is consulted, so the grant widens which accounts are visible without widening what the credential may do. Scope presets expand outwards (`mcp` → its members, never the reverse), so holding an unrelated read scope like `project:admin:read` or a member of the `mcp` group cannot satisfy the gate by accident — I verified that against `AccountToken.expand_scopes/1`. A `User` subject still passes, since a session is not a scope-narrowed credential and the grant checks already require it to be the operator named in `sub`.

Two tests added as asked: one with `["project:admin:read"]` and one with `["project:cache:read"]`, both asserting the read is refused while the grant covers the account. `AccountToken.mcp_scope/0` now exposes the constant so the scope name is not restated in a second module.

pepicrft (approved):

PR #12336, opened 2026-08-13
feat(infra): record why a cache-volume convergence did not happen

fortmarek:
Makes it possible to tell why cache-volume convergence is not happening. Right now it is not observable at all, and it is the load-bearing mechanism for the whole feature.

## Why convergence is the load-bearing part

It is the ONLY way a host holding no master for an account can obtain one.

A cold job promotes with `base_generation = 0`. `establish_first_head` accepts base 0 only when the account has no HEAD at all; otherwise it returns 409. `Finalize` then sees `PromotedGeneration <= 0` and discards the branch. So the host is exactly as cold as it started, and the cache it just built is thrown away.

That closes the loop: a host with no master cannot earn one by running jobs. Only convergence, which downloads the published HEAD archive and installs it, can seed one. And until a host holds the master at the current generation, no job there can present the base the fast-forward needs, so the account's HEAD can never advance either.

## Production is in that state

Last 24h, production fleet:

| | |
|---|---|
| Cold materializes | **286** |
| Convergences | **1** |

A cold materialize means the host holds no master, so its local generation is 0 and it should have attempted to converge. Account 3's HEAD has not advanced since 2026-07-29 despite roughly 1,200 macOS jobs a week, which is the same fact seen from the other end.

## Why the cause is currently unknowable

`convergeMaster` has eight ways to decline and every one was a bare `return`: no log, no metric. The host ships no logs to Loki either. So today these are indistinguishable:

- the guest never staged `volume-head.json` inside the 15s wait
- the dispatch was untrusted, so the server deliberately sent no HEAD
- the presigned download failed
- the stored object no longer matches the advertised digest

Those have completely different fixes, and one of them (untrusted) is not a fault at all.

## What this adds

`tart_kubelet_cache_volume_converge_outcome_total{outcome}` on every exit path, with reasons split by who would have to act:

| outcome | points at |
|---|---|
| `converged` | success |
| `head_absent` | the guest, or the wait being too short |
| `no_head` | nothing published yet for the account |
| `no_download_url` | an untrusted (fork) dispatch; not a fault |
| `already_current` | healthy no-op |
| `download_failed`, `digest_mismatch` | object storage |
| `superseded` | a promote landed mid-download |
| `install_failed`, `error` | local disk |

`converged / (everything except already_current)` is the rate at which hosts that **needed** a master actually got one. That single number decides whether the cache-volume feature can work at all, and nothing measures it today.

`already_current` is deliberately separate: it is the healthy case and would otherwise drown the signal on a fleet that is working.

## Incidental

`awaitVolumeHead`'s interval and attempt count move onto the `Reconciler` so its 15-second wait does not make its own test slow. This mirrors the manager's existing injectable mount-check wait.

## How to test locally

```bash
cd infra/tart-kubelet && go test ./internal/podagent -run TestConvergeMaster -v
```

## Validation run

- 5 new tests covering `head_absent`, `no_head`, `no_download_url`, `already_current` and `download_failed`.
- `go build`, `go vet`, `gofmt -l` clean for darwin AND `GOOS=linux`.
- `go test ./...` and `go test -race ./internal/podagent` pass.

## What this deliberately does not do

It does not fix convergence. It makes the failure legible so the fix targets the real cause instead of a guess. Given the 286:1 ratio, expect the answer within one reconcile tick of the fleet picking this up.

Related work parked on `compact-cache-volume-masters` (closed as #12327): master repacking and a live-bytes-per-tree metric. Both are downstream of this. Repacking makes a converged master cheaper to keep, and the live-bytes split decides whether the per-master budgets need re-deriving, but neither matters while hosts are not converging in the first place.


fortmarek:
Closing in favour of shipping the host's logs instead.

The counter was a workaround for `tart-kubelet` having no queryable output at all. Now that log shipping is being picked up separately, a per-outcome counter is the wrong instrument: logs carry which account, which host, which generation and the actual error text, none of which a bounded label set can. A counter would tell us `head_absent` dominates and then stop being useful exactly when the interesting question starts.

**One thing to carry over, because transport alone will not answer this.** Most of `convergeMaster`'s exits are already logged (mkdir failure, download failure, digest mismatch, install failure, success). Three are silent:

1. `head == nil || head.Generation <= 0 || head.DownloadURL == ""` — one `return` covering three very different situations: the guest never staged `volume-head.json` within the 15s wait; the account has published no HEAD yet; or a HEAD exists but arrived without a URL, which is what an untrusted (fork) dispatch looks like. **This is the most likely cause of the 286 cold materializes to 1 convergence ratio, and it currently emits nothing.**
2. `local >= head.Generation` — the healthy no-op, worth logging at debug so it can be told apart from a failure.
3. `installed == false` — a promote moved the master on mid-download.

Splitting (1) into its three cases and logging each, plus a line for (3), is a much smaller change than this PR and is what the log work actually needs to land alongside it. Whoever picks up log shipping should do that in the same change, or the first Loki query will come back empty for the case we care about.

The branch is left in place. It also carries an incidental improvement worth keeping if convenient: `awaitVolumeHead`'s interval and attempt count moved onto the `Reconciler` so its 15-second wait does not make tests slow, mirroring the manager's existing injectable mount-check wait.

Context for the underlying problem is in #12327's closing comment and the discussion there.

PR #12342, opened 2026-08-14
fix(cli): stop the 90s resource timeout from cancelling shard artifact downloads

fortmarek:
Shard jobs fail intermittently while downloading the shard's test products, with an error like:

```
✖ Error
  Received a session error when performing an url request that sends a GET request to url
  'https://t3.storage.dev/.../shards/<plan-id>/bundle.zip?...': The request timed out.
```

The shard never runs, and the plan ends up with zero rows in `shard_runs`.

### Root cause

`URLSession`'s `timeoutIntervalForResource` is a wall-clock cap on the whole transfer, not an inactivity guard. The shared session sets it to 90s (lowered from 300s in #10503 so a starved cache request fails fast enough to fall back to building from source), and `FileClient.download`, which pulls a shard's test products, used that session.

A multi-gigabyte test-products archive cannot meet that cap: a 3GB archive has to sustain roughly 270Mbps to fit inside 90 seconds. URLSession therefore cancels a transfer that is progressing normally, and every retry restarts from byte zero into the same wall. Shards that lose the race die while the rest pass, which is why it reads as random rather than as a hard failure.

Two details reinforce that the cap, and not the network, is the trigger:

- `timeoutIntervalForRequest` (120s), which is the actual stall guard, is larger than the resource cap, so the cap always fires first on a slow but healthy transfer.
- Concurrent shards on the same runner fleet each pull the full archive, so contention is enough to push a transfer that normally squeaks under 90 seconds over the line.

### What changed

Whole-file transfers now go through a dedicated `URLSession.tuistFile` whose `timeoutIntervalForResource` defaults to 600s. `timeoutIntervalForRequest` stays the stall guard, so a genuinely hung transfer still fails after 120 seconds without new data.

The wall-clock cap is only a backstop against a transfer that trickles just enough to keep resetting the inactivity timer. 10 minutes is 6.7x the cap it replaces: a 3GB archive clears it at 40Mbps, where the old cap demanded 270Mbps. A job moving an archive large enough to need longer, and multi-gigabyte monolithic bundles do exist on CLI versions predating the per-shard split, raises `TUIST_HTTP_TIMEOUT_INTERVAL_FOR_RESOURCE`.

`FileClient` (shard artifacts, plugin releases, whole-file uploads) uses the new session. The cache and CAS paths keep the 90s cap and their fail-fast fallback to building from source, which is what #10503 was after. `TUIST_HTTP_TIMEOUT_INTERVAL_FOR_RESOURCE` still overrides both, and the docs now explain that the resource timeout is a wall-clock cap rather than a stall guard.

### Why this over the alternatives

- Raising the shared 90s cap globally would undo #10503: a starved cache request would stall a build for minutes instead of falling back to a local build.
- Retrying harder does not help. The retry already exists around the shard download; each attempt restarts the whole transfer and hits the same deterministic wall.
- Per-shard split artifacts (#11314) shrink each individual GET and make the wall less likely to be hit, but they do not remove it: a large shared artifact still has to fit in the cap.

### Impact

Shard downloads on slower or contended links stop failing at the 90 second mark. Previews were never affected (`RemoteArtifactDownloader` uses `URLSession.shared` with the 7-day default). Users on 4.201.0 and older additionally lack the split artifacts, the download retry, and the timeout environment overrides, all of which landed in 4.202.0, so upgrading remains worthwhile alongside this fix.

Not addressed here: retries restart the download from the beginning. Range-resume would make a slow link survivable rather than merely tolerable, and is worth a follow-up.

### How to test locally

```bash
mise exec -- tuist generate tuist ProjectDescription TuistHTTPTests --no-open
xcodebuild test -workspace Tuist.xcworkspace -scheme TuistUnitTests -only-testing TuistHTTPTests/URLSessionTuistTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""
```

All cases pass, including the two new ones covering the artifact session's default resource timeout and its environment override.

To exercise the real path, run a sharded test job whose test-products archive is large enough that the download takes more than 90 seconds. Before this change it fails with "The request timed out"; after it, the download completes.


cli/Sources/TuistHTTP/URLSession+Tuist.swift
```diff
@@ -148,12 +148,30 @@ private func makeTuistCASURLSession(useEnvironmentProxy: Bool) -> URLSession {
...
+/// A single transfer here can be several gigabytes, such as a shard's test products. The resource
+/// timeout is a wall-clock cap on the whole transfer rather than a stall guard, so the 90s default
+/// cancels transfers that are progressing normally: finishing 3GB within it requires a sustained
+/// ~270Mbps. The inactivity timeout is the stall guard instead, and the wall-clock cap is only a
+/// backstop against a transfer that trickles: 10 minutes clears a 3GB archive at 40Mbps, and a job
+/// moving something large enough to need longer raises `TUIST_HTTP_TIMEOUT_INTERVAL_FOR_RESOURCE`.
+private func makeTuistLargeTransferURLSession(useEnvironmentProxy: Bool) -> URLSession {
+    let configuration = tuistURLSessionConfigurationResolved(useEnvironmentProxy: useEnvironmentProxy)
+    configuration.timeoutIntervalForResource = Double(
+        environmentInt("TUIST_HTTP_TIMEOUT_INTERVAL_FOR_RESOURCE", default: 600)
```
esnunes: One thing I wanted to flag: `TUIST_HTTP_TIMEOUT_INTERVAL_FOR_RESOURCE` is now read by both sessions — here with a default of 600, and at line 78 for `tuistShared` with a default of 90. Raising it is harmless, but a job that *lowers* it to make cache requests fail faster would also put shard downloads back under a short wall-clock cap, which is the intermittent failure this PR is fixing.

It cuts the other way too: the last commit changed the example in that same docs block to `export TUIST_HTTP_TIMEOUT_INTERVAL_FOR_RESOURCE=1800`, so anyone copying it takes the cache session from 90s to 1800s — which is the fail-fast behaviour #10503 added, undone.

Would it be worth giving this session its own variable, or clamping the shared value up to 600 here, so the two knobs can move independently? Not a blocker either way — just that the two sessions want the value to move in opposite directions.

esnunes (approved):
Nice diagnosis — the wall-clock-versus-stall-guard distinction is the kind of thing that reads as a flaky network for months, and turning it into "how much bandwidth does this cap silently require" makes the whole failure obvious. Separating the two sessions is the right shape too: fail-fast belongs where there's a cheap fallback, patience belongs where the only alternative is failing the job. I also like that the reasoning landed in the code comment rather than only in the commit message.

One note inline, not blocking: the resource-timeout environment variable now drives both sessions, so moving it in either direction has a side effect on the other one — and the example in the docs block now exports 1800, which lands on the cache session too.

One more that has no line in the diff to hang it on: might be worth refreshing the invariants in `cli/Sources/TuistHTTP/AGENTS.md` (lines 18-21) while you're here — they still read as one session with one pair of timeouts ("Default URL session uses explicit request/resource timeouts", "the shared session is resolved lazily, reused process-wide"), and there are three now, two of which diverge from the env-configurable base. Small thing, but that's the file an agent reads before touching this module, so a stale invariant tends to get believed.

One last thing for before you merge — the PR description still describes the session as `URLSession.tuistFile`; the second commit renamed it to `tuistLargeTransfer`.

cli/Sources/TuistHTTP/URLSession+Tuist.swift
fortmarek: Good catch on both halves. The large-transfer session now clamps the shared value to its 600s floor (`max(environmentInt(...), largeTransferResourceTimeout)`), so it is honored only upward and a job lowering it for cache fail-fast cannot re-cap shard downloads; I also reverted the docs example to 600, since raising it to 1800 would have taken the cache session from 90s to 1800s for anyone copying it. Kept one variable rather than adding a second knob, with a test covering the clamp.

PR #12345, opened 2026-08-14
fix(swifterpm): fail local packages that hold no manifest

fortmarek:
### What changed

Three fixes in swifterpm's local (`path:`) package handling, all found while diagnosing a
user report of `tuist install` failing with `no Package.swift in <repo root>`.

**1. A local package directory without a manifest no longer adopts an ancestor package.**
`swift package dump-package` resolves the package root by walking up from its working
directory, and passing `--package-path` walks up as well (verified against Swift 6.3). So
dumping a directory that holds no manifest does not fail. It returns the nearest ancestor
package instead, and swifterpm cached and resolved that ancestor under the identity of the
local dependency that was declared. Any checkout that lives inside another Swift package
turned a broken `path:` dependency into a silently wrong dependency graph rather than an
error. Reproduced before the fix: a repo whose parent directory holds an unrelated
`Package.swift`, with the local dependency pointing at a manifest-less directory, resolved
successfully and wrote this into the package-info index:

```
{ "identity": "feature", "kind": "fileSystem", "package_path": "<the repo, not the package>" }
feature-<hash>.json -> name: DecoyAncestorPackage | products: ['Decoy']
```

Plain SwiftPM errors on the same tree, so this was a divergence from SwiftPM, not a shared
limitation. `ManifestLoader` now requires the manifest before the dump, which also saves the
process spawn. Only the directory-exists-without-a-manifest case is decided there. An absent
or unreadable directory cannot adopt an ancestor either way, so those keep failing through
the subprocess and stay classified in `dumpFailure` next to the error they produce.

**2. The failure now says where the declared path resolved to.** swifterpm canonicalizes the
declared path with `realpath` before dumping, so the directory it names is the declared path
only when nothing along the way is a symlink. When the two differ, the old sentence named a
directory nobody declared and read like a typo, while the redirect is the whole diagnosis:

```
before: failed to load the manifest for the local package foo, declared as "<repo>/Frameworks/Foundation/Foo" by <repo>/Tuist: no Package.swift in <repo>
after:  failed to load the manifest for the local package foo, declared as "<repo>/Frameworks/Foundation/Foo", which resolves to <repo>, by <repo>/Tuist: no Package.swift in <repo>
```

The note is omitted when the paths match, so the common case is unchanged.

**3. `PathCanonicalizer.realpath` builds its string inside the buffer borrow.** It previously
built the string from the pointer `realpath` returned after the `&buffer` borrow had ended.
The pointer is only guaranteed for the borrow. I could not make it misbehave (a 320k-iteration
concurrent stress run produced zero wrong results), so this is hygiene, not a fix for an
observed failure.

### Why

The user report was a self-hosted CI runner where a local package path resolved, through a
symlink, to the repository root. That tree is genuinely broken and plain
`swift package resolve` fails on it too, so the report is not a swifterpm bug. But comparing
the two resolvers surfaced the opposite failure: on the same shape of broken tree, swifterpm
can succeed where SwiftPM errors. That is worse than the reported error, because a wrong graph
surfaces much later than a failed install.

The error-message change comes from the same investigation. Reading the reported message, the
declared path and the named directory being different directories was the single fact that
identified the cause, and nothing in the sentence said the two were related.

### Alternatives considered

Passing `--package-path` to the dump instead of checking for the manifest. It does not work:
SwiftPM walks up from `--package-path` too, so the ancestor is still found. An explicit
existence check is the only thing that stops it.

### Impact

A broken local dependency that previously resolved to an ancestor package now fails with a
clear message. Healthy packages are unaffected, including local packages reached through
symlinks, which keep working (covered by the manual check below).

The gate also applies to root resolution, which runs through the same dump and defaults to
the working directory, so a subdirectory of a package no longer dumps the enclosing package.
That invocation did not work before either: the dump walked up, and then `originHash` read
the manifest at the directory the caller actually named and failed a step later. So the
change there is which error you get, not whether it fails:

```
before: Error: Path not found: <package>/Sources/Root/Package.swift
after:  Error: no Package.swift in <package>/Sources/Root
```

`restore` is unchanged, since it already failed on `Package.resolved` before reaching a dump.
The Tuist call sites pass the package path explicitly and never relied on the walk-up.
A test pins the refusal so it stays a stated contract rather than a side effect.

### How to test locally

Unit tests:

```bash
swift test --package-path swifterpm --filter "ManifestTests|SupportTests"
```

Both new tests fail on `main` for the right reasons: the ancestor test reports
`expected the local package manifest to fail to load, loaded ["Ancestor"] instead`, and the
redirect test shows the message without the redirect.

End to end, against the built binary:

```bash
swift build --package-path swifterpm --product swifterpm
```

Then, in a scratch directory, create a package at the top level, a `Tuist/Package.swift`
under it declaring `.package(path: "../Frameworks/Foundation/Foo")`, and make that path a
symlink to a directory holding no manifest. Before this change
`swifterpm --package-path <dir>/Tuist resolve` succeeded with the top-level package bound to
`foo`. After it, the resolve fails and names both the declared path and where it resolves to.

Full swifterpm suite passes locally (145 tests), and `mise run lint` is clean.

### DIFF
--- a/swifterpm/Sources/swifterpm/Manifest.swift
+++ b/swifterpm/Sources/swifterpm/Manifest.swift
@@ -61,6 +61,7 @@ enum ManifestLoader {
     }
 
     static func dumpPackage(packageDir: URL, disableSandbox: Bool) async throws -> Any {
+        try await requireManifest(packageDir: packageDir)
         do {
             let data = try await loadPackageJSON(
                 packageDir: packageDir, disableSandbox: disableSandbox
@@ -72,12 +73,29 @@ enum ManifestLoader {
     }
 
     static func dumpPackageJSON(packageDir: URL, disableSandbox: Bool) async throws -> Data {
+        try await requireManifest(packageDir: packageDir)
         do {
             return try await loadPackageJSON(
                 packageDir: packageDir, disableSandbox: disableSandbox
             )
         } catch {
             throw await dumpFailure(packageDir: packageDir, underlying: error)
+        }
+    }
+
+    /// `swift package dump-package` resolves the package root by walking up from its working
+    /// directory, and `--package-path` walks up as well, so a directory that holds no manifest
+    /// does not fail the dump: it produces the nearest ancestor package instead, which swifterpm
+    /// would then cache and resolve under the identity of the package the caller asked for. A
+    /// checkout that lives inside another Swift package is enough to turn a broken local
+    /// dependency into a silently wrong graph, so refuse the dump before it starts.
+    ///
+    /// Only the directory-exists-without-a-manifest case is decided here. A directory that is
+    /// absent or unreadable cannot adopt an ancestor either way, and letting the subprocess fail
+    /// keeps the diagnosis of those states in `dumpFailure`, next to the error they produce.
+    private static func requireManifest(packageDir: URL) async throws {
+        if case .absent = await manifestPresence(packageDir: packageDir) {
+            throw ToolError.message("no Package.swift in \(packageDir.path)")
         }
     }
 
@@ -221,7 +239,8 @@ enum ManifestFileSystemDependencyGraph {
                 throw ToolError.message(
                     """
                     failed to load the manifest for the local package \
-                    \(item.dependency.identity), declared as "\(item.dependency.path)" by \
+                    \(item.dependency.identity), declared as "\(item.dependency.path)"\
+                    \(redirect(declared: item.dependency.path, canonical: canonicalPath)) by \
                     \(item.parentPackageDir.path): \(error)
                     """
                 )
@@ -239,6 +258,14 @@ enum ManifestFileSystemDependencyGraph {
         }
 
         return result
+    }
+
+    /// The failure above names the directory the dump was attempted in, which is the declared
+    /// path only when nothing along the way is a symlink. When the two differ, that sentence
+    /// names a directory nobody declared and reads like a typo, while the redirect is the whole
+    /// diagnosis: a local dependency whose path lands somewhere other than the package.
+    private static func redirect(declared: String, canonical: URL) -> String {
+        canonical.path == declared ? "" : ", which resolves to \(canonical.path),"
     }
 
     static func packagePathForFileSystemDependency(
--- a/swifterpm/Sources/swifterpm/Support.swift
+++ b/swifterpm/Sources/swifterpm/Support.swift
@@ -507,16 +507,24 @@ enum PathCanonicalizer {
         #if os(Windows)
             url.standardizedFileURL
         #else
+            // The pointer `realpath` returns is only guaranteed for as long as the buffer is
+            // borrowed, so the string has to be built inside the borrow rather than from the
+            // returned pointer afterwards.
             var buffer = [CChar](repeating: 0, count: Int(PATH_MAX))
-            #if canImport(Glibc)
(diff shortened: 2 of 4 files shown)


swifterpm/Sources/swifterpm/Manifest.swift
```diff
@@ -61,6 +61,7 @@ enum ManifestLoader {
     }
 
     static func dumpPackage(packageDir: URL, disableSandbox: Bool) async throws -> Any {
+        try await requireManifest(packageDir: packageDir)
```
esnunes, line 64: Something the gate reaches that I think is worth a line in the description. The walk-up was not only how a broken `path:` dependency found an ancestor, it was also how the root package was found when nobody named one. `packageDir` defaults to `CLIPath(".")` and `loadPackageJSON` runs the dump with `workingDirectory: packageDir` and no `--package-path`, so `swifterpm resolve` from a subdirectory of a package used to resolve the enclosing package. That mirrors SwiftPM's own root-finding, which still behaves that way: `swift package dump-package` run from `Root/Sources/Root` resolves `Root`.

With the gate in place that invocation fails with `no Package.swift in <cwd>`, and since `PackageResolver.resolve` calls `dumpPackage` with no wrapping frame, the message arrives with no attribution.

Low impact in practice, since the Tuist call sites route through builders that pass `--package-path` explicitly, and I would not want the walk-up back. It is really the "No behavior change for healthy packages" line in Impact that I would adjust, because a healthy package invoked from its own subdirectory is exactly what stops working.

esnunes (approved):
Nice piece of work, and the write-up made it easy to follow. The core finding is a good one: `dump-package` walking up from its working directory means swifterpm could succeed where SwiftPM errors, and a wrong graph surfacing later really is worse than a failed install.

A couple of the surrounding assumptions I checked while reading, in case they are useful. The alternative you ruled out holds up: a directory holding only `Package@swift-5.9.swift` is rejected by SwiftPM's own root-finding, so hard-coding `Package.swift` in the probe matches upstream rather than under-approximating it. And the new redirect note does not misfire the way I first suspected: SwiftPM canonicalizes the package root but leaves the dependency leg unresolved, so the note appears when the dependency itself is the symlink and not on every machine whose repo path crosses one. The `realpath` rewrite reads correctly to me too, building the string inside the borrow is the right shape, and shadowing the outer buffer is a nice touch since it makes the array unreachable inside the closure.

I left one note inline, about something the gate reaches beyond local dependencies: running swifterpm from a subdirectory of a healthy package. It is low impact given how Tuist calls it, and it is the Impact line I would adjust rather than the code. Approving, none of this needs to hold the change up.

swifterpm/Sources/swifterpm/Manifest.swift
fortmarek: You are right about the mechanism, and I checked the consequence: that invocation failed before this PR too, because `originHash` reads `Package.swift` at the directory the caller named a step after the dump, so the walk-up produced a dump of one package and then died anyway (`resolve`/`update` from a subdirectory: `Error: Path not found: <pkg>/Sources/Root/Package.swift` before, `Error: no Package.swift in <pkg>/Sources/Root` now; `restore` is unchanged, it fails earlier on `Package.resolved`). Updated the Impact section to say all of that instead of "No behavior change for healthy packages", and added a test pinning the refusal so it is a stated contract rather than a side effect.

PR #12351, opened 2026-08-14
fix(infra): stop macOS fleets pinning a retired Scaleway image

fortmarek:
## What changed

macOS fleets pin a macOS **release family** (`Tahoe`) instead of an exact image (`macos-tahoe-26.3`).

- **Adoption** accepts any pool host in the family, and **refuses** a pin that names a specific image.
- **Release** reinstalls onto the family's newest published image, so a returned host is adoptable again.
- The runner profile dialog stops advertising a point release it never guaranteed.

## Why

Staging lost every runner mini and could not recover on its own. Two mechanisms combined into a one-way ratchet:

- `AdoptFromPool` matched a pool host's image name **exactly**, so a pin naming a retired image could never match again.
- `ReleaseToPool` called `ReinstallServer` with no `os_id`, so a released host was reimaged with the server type's *current* default rather than the image its fleet adopts on.

Scaleway had retired `macos-tahoe-26.3` — `scw apple-silicon os list` no longer returns it. Every host released for any reason came back permanently unadoptable. Staging released two minis and went to zero runner capacity, with `NoAvailableHost` firing against a pool that was not actually empty. Production pinned the same retired image on all three fleets, and only still worked because its hosts were imaged before the retirement.

## Why a family rather than a newer pin

The first pass bumped the pin to `macos-tahoe-26.6.1`. That moves the trap rather than removing it — the next time Scaleway moves the server type default, an exact pin strands the same way. A family is stable across point releases, so the fleet tracks them instead of chasing them.

## Why a versioned pin is refused rather than widened

Normalizing an exact pin down to its family would mean someone who wrote `macos-tahoe-26.5` silently gets any Tahoe host, with no event and nothing in the CR to say so. A pin nobody can honour should fail where it is written. Adoption reports it as its own `InvalidOSPin` condition, deliberately distinct from `NoAvailableHost`: one needs a config edit, the other needs pre-ordered capacity, and conflating them sends operators to buy hosts that could never match.

**Release stays lenient, and that is load-bearing.** All 15 existing Machines carry `macos-tahoe-26.3` in their own specs. They never re-adopt, so strictness costs them nothing, but they do release on replacement — and refusing there would strand each one: host still claimed and billing, finalizer never clearing, no way to shed the Machine to get a correctly-pinned replacement. A drain path has to accept what already exists.

Enforced in the controller, not as a CRD schema constraint. A pattern on `spec.os` would be validated on every write to CRs that already hold exact values, and the `adoptPoolPrefix` field documents what that class of mistake costs here: a schema constraint on a resource CAPI clones wedges MachineSet scale-up with no repair short of break-glass.

## Details worth review

- **Family boundaries hold.** `devos-tahoe-*` images report family `DevOS` and do not satisfy a Tahoe fleet; neither do Sequoia hosts. Matching prefers Scaleway's `Family` field and falls back to the image name only when it is empty.
- **Multi-word families survive.** The version suffix is cut at the first dash *followed by a digit*, not the first dash, so `macos-golden-gate-27.0` and `Golden Gate` both key to `goldengate`.
- **Not Scaleway's own name filter.** That filter is a prefix match — `26.6` also returns `26.6.1` — so it would quietly cross point releases. We list and compare ourselves.
- **A family with nothing published degrades, it does not wedge.** Release falls back to the server type default and raises an `OSPinUnavailable` warning event naming the pin.
- **The UI label names what actually holds still.** macOS jobs run on a guest built from a floating Tahoe base tag, on hosts that now adopt by family, so no layer held `26.3`. Linux keeps its exact release because the runner Dockerfile is genuinely `FROM ubuntu:22.04`. The asymmetry is documented so the version does not get helpfully restored later. Note the Linux fleet values pin `ubuntu_24.04` — that is the host, a different layer from the image jobs execute in.

## Rollout

Ships with the server chart, so it follows the usual canary → acceptance → production cascade. Canary has a macOS fleet, so that stage is a real rehearsal.

On the production deploy:

1. **CRDs apply.** The `os` default becomes `Tahoe`. Defaults are applied at admission, so the 13 stored CRs keep `macos-tahoe-26.3` verbatim.
2. **Operator rolls.** Family matching goes live.
3. **Templates update.** `runners-fleet` and `builders-fleet` are stable-named, so they are patched in place and their MachineDeployment spec does not change — no rollout, no new MachineSet. `macos-fleet` is content-hashed, so its name moves off `887070e8` and CAPI creates a new MachineSet at **zero replicas**; the existing one keeps both Machines.
4. **The 13 running Machines do nothing.** They hold hosts, so `AdoptFromPool` is never called for them, and `InvalidOSPin` cannot fire despite their versioned pins. Their pin is inert for as long as they hold a host.

Steady-state deploy: **no host churn, no capacity change, one extra empty MachineSet.** All MachineDeployments are `OnDelete`.

The first replacement afterwards exercises the new path end to end: release resolves `macos-tahoe-26.3` to its family and reinstalls the host onto current Tahoe; the MachineSet clones `Tahoe` from the updated template; the replacement adopts. Production's pool host is `macos-tahoe-26.5` — family Tahoe — so it becomes adoptable on deploy and the replacement can swap straight over instead of idling for a reinstall. Canary's pool host recovers the same way.

## Risks

- **Rollback stops being clean once a replacement has happened.** Any Machine created after the deploy carries `os: Tahoe`, and a rolled-back operator does exact-name matching, so it would sit in `NoAvailableHost`. Rolling back is free while no new Machine exists; after that those CRs need patching.
- **`InvalidOSPin` should never appear in production.** If it does, something cloned a versioned pin — a hand-written CR or a stale MachineSet — and it is a real signal.
- **Mixed point releases become normal.** After a few replacements the fleet holds `26.3` and `26.6.1` hosts side by side. That is the intent, but host-level differences across point releases are the one thing family pinning does not insulate against. The guest image is unchanged, so jobs should not notice.

Verification after deploy is three reads: `get samm -A` (all still Ready), `get machineset -A` (one new one at zero), `get events --field-selector reason=InvalidOSPin` (empty).

## Validation

- `go test ./...` passes; `mix test test/tuist_web/live/runner_profiles_live_test.exs` passes.
- Chart renders with `os: Tahoe` on both macOS templates; the macos-fleet template hash moves as expected.
- `mix gettext.extract` run; only the `.pot` changed, no `.po` files touched.
- New tests cover family adoption across point releases, cross-family isolation, family-key normalization, refusal of a versioned pin (and that no candidate is renamed when it fires), newest-in-family reinstall, unpublished-family failure, resolution caching, and the retired-pin release fallback.
- Verified red before green: the adoption test under exact-name matching, and the release fallback without the downgrade.


infra/cluster-api-provider-tuist/internal/scaleway/client.go
```diff
@@ -641,6 +724,151 @@ func (c *Client) ReleaseToPool(ctx context.Context, id, zone, poolPrefix string)
...
+// inside the family keeps the host adoptable by the fleet that
+// released it.
+//
+// Deliberately not Scaleway's own `Name` list filter — that is a
+// prefix match ("26.6" also returns "26.6.1"), so it would quietly
+// cross point releases. We list and compare ourselves.
+//
+// A family with nothing published returns ErrOSNotPublished; see that
+// sentinel for why callers downgrade rather than fail.
+func (c *Client) resolveOSID(ctx context.Context, zone, pin string) (string, error) {
```
esnunes: One thing I think is worth fixing before this ships: `resolveOSID` caches the resolved image UUID for the process lifetime (`client.go:818`), and only *misses* are re-listed. When Scaleway later retires the cached image — the exact event this PR is built around — the retirement is invisible: `ErrOSNotPublished` is only produced by the ListOS resolution path, which a cache hit skips. `ReinstallServer` then fails with a generic API error, `releaseHostToPool`'s downgrade never fires (it matches only `ErrOSNotPublished`), and every release in that zone/family retries the same dead UUID until the operator pod restarts — hosts stay claimed and billing, finalizers never clear. That's the outage class this PR fixes, recreated one layer down.

A small fix that keeps the round-trip saving: evict the cache entry when `ReinstallServer` rejects the request and re-resolve once before giving up. Given how cold the teardown path is, I'd also understand just dropping the cache.

infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
```diff
@@ -1508,6 +1545,18 @@ func (r *ScalewayAppleSiliconMachineReconciler) acquireServer(
 		machine.Spec.OS,
 		poolPrefix,
 	)
+	// A versioned pin is a configuration error, not absent capacity.
+	// Reporting it as NoAvailableHost would send an operator to
+	// pre-order hosts that could never match. Requeue slowly: nothing
+	// changes until someone edits the fleet's os.
+	if errors.Is(err, scaleway.ErrOSPinNotFamily) {
```
esnunes, line 1563: The lenient-release / strict-adopt split looks right, but there's a gap when the *same* Machine crosses the boundary: `handleBootstrapFailure` at exhaustion releases the host and resets the Machine to Pending without deleting it, so the next reconcile passes its baked-in `macos-tahoe-26.3` to `AdoptFromPool` and gets refused — the Machine then loops on `InvalidOSPin` every 5 minutes until someone deletes it by hand. The rollout section's "their pin is inert for as long as they hold a host" is true, but this is precisely a path where a legacy Machine sheds its host while surviving. The same applies to any legacy Machine already sitting hostless in the adoption path at deploy time — if staging still has pending Machines from the incident with versioned pins, they'll flip from `NoAvailableHost` to `InvalidOSPin` and won't self-heal under `OnDelete`.

Two small things would close it: mention the manual `kubectl delete machine` step in the rollout/runbook notes, and consider wording the `InvalidOSPin` event for this case — it currently says to fix the *fleet's* os, but for a pre-existing Machine the fleet template is already correct; the fix is deleting the Machine (or patching that CR's own spec.os). Not asking for adoption to widen legacy pins — the refusal rationale is sound — just for the recovery path to be discoverable.

infra/cluster-api-provider-tuist/internal/scaleway/client.go
```diff
@@ -641,6 +724,151 @@ func (c *Client) ReleaseToPool(ctx context.Context, id, zone, poolPrefix string)
...
+			break
+		}
+	}
+	return strings.NewReplacer(" ", "", "-", "", "_", "").Replace(key)
+}
+
+// osFamilyMatches reports whether a server's image belongs to the
+// family a fleet pins. Falls back to deriving the family from the
+// image name when the API leaves Family empty.
+func osFamilyMatches(os *applesilicon.OS, pin string) bool {
```
esnunes, line 764: A question on the one assumption everything rests on: `osFamilyMatches` trusts the API's `Family` field whenever it's non-empty, and falls back to the image name only when it's empty. The vendored SDK documents the field as "the OS family to which this OS belongs, eg. 13 or 14" — numeric. If production actually returns e.g. "26" for Tahoe images, the family key never equals `tahoe`, adoption matches nothing, and the family couldn't even be pinned (digits are refused as versioned pins). The PR text says devos-tahoe-* images report family "DevOS", which suggests the SDK comment is just stale — but all the test fixtures hand-write `Family: "Tahoe"`, so nothing validates against real values. Could you paste real `scw apple-silicon os list` (and a `ListServers` server's embedded os) output to pin down the field's shape on both endpoints?

Either way, a cheap hardening: treat it as a match when *either* the Family-derived key *or* the name-derived key equals the pin key. Then a surprising Family value degrades to name matching instead of a fleet-wide silent mismatch.

infra/cluster-api-provider-tuist/AGENTS.md
```diff
@@ -480,12 +480,25 @@ kubectl scale machinedeployment <fleet-name> --replicas=1
...
-pre-ordered until you decide to release it via the Scaleway console.
+(`<poolPrefix><uuid>`) and triggers a Scaleway OS reinstall onto the
+Machine's own `spec.os`; the host stays alive, returns to
+factory-default state, and becomes eligible for the next adoption
+once Scaleway flips it back to `Delivered + Ready`. The 24h Apple
+licensing floor stays in operator-owned territory — you keep paying
+for capacity you already pre-ordered until you decide to release it
+via the Scaleway console.
+
+`spec.os` must name an image `scw apple-silicon os list` still
```
esnunes: The AGENTS.md paragraph this branch adds still describes its first iteration: "`spec.os` must name an image `scw apple-silicon os list` still returns. Adoption matches a pool host's image name exactly … Repoint the pin in `values.yaml`." The shipped code refuses image-name pins at adoption with `InvalidOSPin`, matches by family, and downgrades retired pins on release — "family" appears nowhere in the runbook. An on-call operator following this checked-in playbook during the next retirement would repoint the value to an exact image name and put the fleet straight into `InvalidOSPin`. Since the runbook is the thing the next incident reaches for, I'd update it to the family model (and mention the manual delete for a legacy Machine stuck on its own versioned pin).

Two smaller drifts in the same direction: the `osFamilyKey` comment says legacy exact pins "start matching current Tahoe hosts without any migration" — true on the release path, but adoption refuses them; and the orphan reclaimer's new comment still reasons about a fleet that "pins that image" when fleets now pin families.

infra/cluster-api-provider-tuist/internal/scaleway/client.go
```diff
@@ -641,6 +724,151 @@ func (c *Client) ReleaseToPool(ctx context.Context, id, zone, poolPrefix string)
...
+		family = os.Name
+	}
+	return osFamilyKey(family) == osFamilyKey(pin)
+}
+
+// compareOSVersions orders dotted numeric versions ("26.6.1" >
+// "26.6" > "26.5"). Segments that aren't numeric fall back to a
+// string comparison so an unexpected version shape still orders
+// deterministically rather than panicking.
+func compareOSVersions(a, b string) int {
```
esnunes, line 787: Might be worth a guard on the "newest in family" selection: the SDK's `OS` has an `IsBeta` field and `resolveOSID` doesn't check it — if Scaleway ever publishes a Tahoe beta with the highest version number, released hosts get silently reimaged onto a pre-release OS (and the choice is cached). Relatedly, `compareOSVersions`' string fallback ranks a suffixed segment above an absent one, so a semver-style "26.7-beta" would order above "26.7". I can't tell from the catalog data in the repo whether Scaleway publishes betas or suffixed versions here, so this may be theoretical — but skipping `IsBeta` entries unless nothing else is published is about one line and removes the sharpest case. Nothing blocking.

infra/cluster-api-provider-tuist/internal/scaleway/client.go
```diff
@@ -641,6 +724,151 @@ func (c *Client) ReleaseToPool(ctx context.Context, id, zone, poolPrefix string)
...
+	if id, ok := c.osIDCache[cacheKey]; ok {
+		c.osIDMu.Unlock()
+		return id, nil
+	}
+	c.osIDMu.Unlock()
+
+	resp, err := c.API.ListOS(&applesilicon.ListOSRequest{
+		Zone: scw.Zone(zone),
+	}, scw.WithContext(ctx), scw.WithAllPages())
+	if err != nil {
```
esnunes, line 872: Non-blocking hardening idea: `ListOSRequest` has a `ServerType` filter ("list of compatible server types") that `resolveOSID` doesn't use, and the cache key is zone/family only. If Scaleway ever publishes a family's newest image for newer hardware only, the resolver would pick it zone-wide, the reinstall would fail on older SKUs, and the incompatible UUID would stay cached for every fleet in the zone. All three fleets are M2-L today so this is dormant, but the values.yaml comment about picking types under restocking pressure suggests mixed types are anticipated. The machine's type is already in the CR — threading it into `ReleaseToPool` → `resolveOSID` and the cache key would close both ends.

esnunes (approved):
This is an unusually well-argued PR — the strict-adopt / lenient-release asymmetry is reasoned in the code where it matters, the refusal happens before anything is half-claimed, resolution happens before the rename so failures leave the host recoverable, and the staging outage exists as a named regression test. The family model itself looks like the right call over a newer pin.

The inline notes cluster around one theme: the new resolution layer inherits the PR's own premise — Scaleway changes the catalog without notice — but caches and error-handles as if the catalog were stable. The one I'd fix before merge is the cache invalidation (a retired cached UUID currently wedges every release in the zone until a controller restart, and the ErrOSNotPublished downgrade can't fire on a cache hit). Close behind: confirming what Scaleway's Family field actually returns in production, since every test fixture hand-writes it. The lenient/strict boundary also has one gap worth a runbook line — a legacy Machine that sheds its host after bootstrap exhaustion re-enters adoption with its baked-in versioned pin and wedges on InvalidOSPin until deleted.

The remaining notes are doc drift (the AGENTS.md runbook still teaches the exact-pin model this PR replaces) and non-blocking hardening (beta images, server-type compatibility).

infra/cluster-api-provider-tuist/internal/scaleway/client.go
fortmarek: Dropped the cache entirely rather than evicting — you're right that a cache hit skipped the only path producing `ErrOSNotPublished`, so a retirement of the cached image would have failed `ReinstallServer` generically, bypassed the downgrade, and retried the dead UUID until restart. Release runs once per host teardown, so the round-trip was never worth that. Test asserts the catalog is re-listed per release and that a retirement mid-flight surfaces as `ErrOSNotPublished`.

infra/cluster-api-provider-tuist/controllers/macos/scalewayapplesiliconmachine_controller.go
fortmarek: Confirmed and documented. `handleBootstrapFailure` leaves the Machine hostless at exhaustion, so a legacy Machine reaches adoption with its baked-in pin and loops with no operator action. The event now says the fleet template may already be correct and names deleting the Machine (MachineSet re-clones) or patching `spec.os`; AGENTS.md gained a runbook section covering the hostless case. Staging has exactly one such Machine right now.

infra/cluster-api-provider-tuist/internal/scaleway/client.go
fortmarek: The SDK comment is stale — real output, both endpoints, fr-par-1:

`os list`: `macos-tahoe-26.6.1 family=Tahoe`, `macos-sequoia-15.6 family=Sequoia`, `macos-ventura-13.6 family=Ventura`, `devos-tahoe-2026.06.30 family=DevOS`, `fedora-asahi-remix family=Linux`.

`ListServers` embedded os: `os.name=macos-tahoe-26.3 os.family=Tahoe os.version=26.3 os.is_beta=false` — populated identically.

Took the hardening anyway: matching now accepts either the family-derived or name-derived key, so a surprising `Family` degrades to name matching. Cross-family exclusion holds either way — `devos-tahoe` keys to `devos` by family and `devostahoe` by name.

infra/cluster-api-provider-tuist/AGENTS.md
fortmarek: Rewritten to the family model, including that image-name pins are refused at adoption, and added a runbook section for a Machine stuck on `InvalidOSPin` after shedding its host. Both smaller drifts fixed: the `osFamilyKey` comment now says legacy pins normalize on the release path while adoption refuses them, and the orphan reclaimer comment talks about families rather than an image.

infra/cluster-api-provider-tuist/internal/scaleway/client.go
fortmarek: Not theoretical — `fedora-asahi-remix` carries `is_beta=true` in fr-par-1 today. Betas are now skipped unless the family publishes nothing else (better a beta than a stranded fleet). Also fixed the ordering you spotted: versions peel the leading integer and a remaining suffix sorts as a pre-release, so `26.7-beta` < `26.7` instead of above it.

infra/cluster-api-provider-tuist/internal/scaleway/client.go
fortmarek: Threaded through. `ReleaseToPool` now takes a `ReleasePin{Family, ServerType}` and passes the type to `ListOS`, so the catalog is scoped to images the host's SKU can boot. The cache is gone in the same push, so the stale-entry half of this disappears with it.

PR #8774, opened 2025-11-27
feat(server): test cases page

fortmarek:
This PR adds two new pages:
- Test Cases page
- Test Case detail page

Screenshot 2025-11-27 at 17 07 14

Screenshot 2025-11-27 at 17 07 23

PR #12352, opened 2026-08-14
feat(server): make test case summary metrics controllable by timeframe

fortmarek:
Users asked what timeframe the metrics on the test case detail page cover, expecting it to be all time, and pointed out that an all-time number is misleading for a test that has been skipped.

They were right, and it was worse than "all time": each widget used a different, undisclosed window.

| Widget | Before | After |
|---|---|---|
| Flakiness rate | Last 30 days | Selected period |
| Test reliability | All time, default branch | Selected period, default branch |
| Test case runs | All time | Selected period |
| Failed test case runs | All time | Selected period |
| Avg. test case run duration | All time | Selected period |

The skipped-test case is the sharpest version of the problem. Reliability is `countIf(status = 'success') / count()`, so a skipped run counts against the denominator but never as a success, and once a test is quarantined it stops accruing runs entirely. The displayed percentage is a frozen snapshot of how the test behaved before it was skipped, and with an all-time window it stays frozen forever.

### What changed

The three per-test-case analytics functions in `lib/tuist/tests/analytics.ex` now accept optional `:start_datetime` / `:end_datetime` bounds, and the detail page drives them from the same date picker the project-level Tests overview already uses (24 hours / 7 days / 30 days / 12 months / custom, defaulting to 30 days). The picker lives in the Summary card header and patches `analytics-date-range` into the query string, matching the pattern on every other analytics page.

Each function keeps its existing default, so the public API endpoint and the MCP `get_test_case` tool are unaffected: reliability and run counts stay unbounded there, flakiness stays at 30 days as its OpenAPI schema documents.

### Two fixes that fell out of this

**The average duration tooltip was wrong.** It claimed the value covered "the last 50 runs". The query has never had a limit, so it was averaging every run ever recorded. It has said this since the page was introduced in #8774. All five tooltips now state the window they actually use.

**A latent crash became reachable.** `test_case_reliability_by_id` returns `nil` when a test case has no runs in range, and the widget passed that straight into `format_percent/1`, which only accepts numbers. While the query was unbounded this needed a test case with zero runs in its entire history, so nobody hit it. With a period selected, picking "Last 24 hours" on a test that had not run that day would 500 the page. The widget already had an empty state and now reaches it. This surfaced as seven pre-existing tests in the file failing once the period bound went in, which is how it was caught.

### Why relative presets pass no upper bound

`DatePicker.date_picker_params/3` truncates "now" down to the second, deliberately, so the values stay compatible with ClickHouse `DateTime` columns. An upper bound built from that value excludes runs ingested during the current second. For a relative preset the end is "now" and no run can be in the future, so the bound adds nothing and only costs correctness. Custom ranges pass both bounds, since there the end is a moment the user actually chose.

### Not included

The other half of the request was making the metrics controllable by number of runs rather than by calendar window. A last-N-runs mode needs a different query shape (the rolling per-case aggregates rather than a time predicate) and a picker that can express it, so it is left as separate work pending confirmation that it is the more useful of the two.

### How to test locally

1. `mise run dev` and open a test case detail page: `/{account}/{project}/tests/test-cases/{test_case_id}`.
2. The Summary card header now has a date range picker, defaulting to Last 30 days. Change it and confirm all five widgets update.
3. Pick a narrow range with no runs in it (or Last 24 hours on a test that has not run today) and confirm the reliability widget renders "No data" instead of erroring.
4. Hover the tooltip on Avg. test case run duration and confirm it no longer claims "the last 50 runs".

Automated:

```
mix test test/tuist_web/live/test_case_live_test.exs test/tuist/tests/analytics_test.exs test/tuist_web/helpers/date_picker_test.exs
```

81 passed. The API controller and MCP suites that call these functions also pass (140 passed), and `mix credo` is clean on the changed modules. `mix gettext.extract` touched only `dashboard_tests.pot`.


server/lib/tuist_web/live/test_case_live.ex
```diff
@@ -357,21 +381,36 @@ defmodule TuistWeb.TestCaseLive do
...
-      {:ok, %{analytics: Analytics.test_case_analytics_by_id(project.id, test_case_id)}}
+      {:ok, %{analytics: Analytics.test_case_analytics_by_id(project.id, test_case_id, opts)}}
     end)
     |> assign_async([:flakiness_rate, :flaky_runs_grouped, :flaky_runs_meta], fn ->
       {flaky_runs_grouped, flaky_runs_meta} = Tests.list_flaky_runs_for_test_case(project.id, test_case_id)
 
       {:ok,
        %{
-         flakiness_rate: Analytics.get_test_case_flakiness_rate(test_case_detail),
+         flakiness_rate: Analytics.get_test_case_flakiness_rate(test_case_detail, opts),
```
esnunes, line 413: One thing I wanted to flag about the flakiness tile. Its `:if` hides the widget when the rate is `0.0`, which was stable while the window was a fixed 30 days — but now that the period is user-controlled, picking "Last 24 hours" on a test with no flaky runs that day drops the Summary row from five tiles to four, and switching back brings it in again. The other four widgets handle this with `empty` / `empty_label` instead (reliability shows "No data"). Might be worth giving flakiness the same empty state, so 0% reads as an answer rather than as a missing widget?

server/test/tuist_web/live/test_case_live_test.exs
```diff
@@ -61,6 +61,100 @@ defmodule TuistWeb.TestCaseLiveTest do
...
+      test_case_id = seed_runs_across_time(project)
+
+      # When
+      {:ok, lv, _html} =
+        live(conn, ~p"/#{account.name}/#{project.name}/tests/test-cases/#{test_case_id}")
+
+      render_async(lv)
+
+      # Then - the run from 40 days ago is excluded
+      assert widget_value(lv, "widget-test-case-runs") =~ "2"
```
esnunes: Nit on the widget assertions: `=~ "2"` matches any rendered value containing a `2`, so `12`, `20` and `25` would pass too. With counts of 1 and 2 the current fixtures don't collide, so these do catch what they were written for — it's just one extra seeded run away from a false pass. Since `widget_value/2` already hands back the rendered element, would it be worth comparing the exact number instead?

server/lib/tuist/tests/analytics.ex
```diff
@@ -1373,6 +1385,20 @@ defmodule Tuist.Tests.Analytics do
...
 
+  defp apply_period_filter(query, opts) do
+    query
+    |> apply_start_datetime_filter(Keyword.get(opts, :start_datetime))
+    |> apply_end_datetime_filter(Keyword.get(opts, :end_datetime))
+  end
+
+  defp apply_start_datetime_filter(query, nil), do: query
+
+  defp apply_start_datetime_filter(query, start_datetime), do: where(query, [tcr], tcr.inserted_at >= ^start_datetime)
```
esnunes: Question on the column the period bounds. `apply_period_filter/2` filters on `inserted_at`, carried over from the flakiness query — but every other user-facing period filter in this module bounds `ran_at` (analytics.ex:65, :81, :247, :265, :361, :473, :627, :673, :706, :735, :777, including `scoped_project_test_case_ids/4` on the same table). The `inserted_at` ones are the time-bucketed projections, where bucketing on ingestion keeps the series from being rewritten by late arrivals — which makes sense there but reads differently for a scalar over a window the user picked.

It matters because xcresult processing is asynchronous: a run that executed three days ago but was processed today counts under "Last 24 hours", and one that executed inside the window but is still queued doesn't — while the table below shows a "Ran at" that contradicts the widget. Was `ran_at` considered here, or is `inserted_at` deliberate?

esnunes (approved):
Really nice change — the problem framing in the description is sharper than the diff alone would suggest, and the two bugs you found on the way (the average-duration tooltip describing a 50-run limit that never existed, and `format_percent/1` being reachable with `nil` once a period is applied) are the kind of thing that only surfaces when someone actually pulls on the thread.

A few things I liked: keeping the default at the function boundary so the API and MCP callers don't move; reusing `DatePicker` and the `analytics-*` param convention rather than inventing a page-local scheme; and re-decoding the query string in the `:test_created` handler so a live update preserves the window the user picked. The reasoning for why relative presets pass no upper bound is right, and I'm glad it's written down.

I left a few notes inline, none blocking. Two I'd most like your read on. First, which column the period should bound: `apply_period_filter/2` uses `inserted_at`, carried over from the flakiness query, but the other user-facing period filters in this module use `ran_at`, and with xcresult processing being asynchronous the two genuinely diverge. Second, the flakiness tile: its `:if` hides it at 0%, which was stable with a fixed 30-day window but now toggles as the user changes the period, while the other four tiles have proper empty states.

One more that lands on an unchanged line, so it didn't fit inline — `server/lib/tuist_web/live/test_case_live.ex:409`:

> Small one: `flakiness_rate` shares this `assign_async` with `flaky_runs_grouped` / `flaky_runs_meta`, and `list_flaky_runs_for_test_case/3` doesn't take a period. So each period change re-issues that grouped query for an identical result, and the Flaky runs tab keeps showing runs from outside the window the Summary card now advertises. Splitting the flaky-runs pair into their own `assign_async` would drop the redundant query; scoping the list too would make the two agree. Any preference, or is the list intentionally all-time?

server/lib/tuist/tests/analytics.ex
fortmarek: Good catch, switched to `ran_at`. You are right that the `inserted_at` here was carried over from the flakiness query rather than chosen, and that the other `inserted_at` bounds are all bucketed series where ingestion-time bucketing is deliberate. The async-processing case you describe is the deciding one: a widget disagreeing with the "Ran at" column beside it is indefensible. Tests now seed run and ingestion timestamps as inverses, so bounding the wrong column flips the expected value instead of passing silently.

server/lib/tuist_web/live/test_case_live.ex
fortmarek: Agreed, and it was a regression my change introduced by making the window user-controlled. Dropped the `:if` and gave the tile the same `empty` / `empty_label` treatment as the other four, keyed on whether the period has any runs at all. So 0% now renders as 0%, and "No data" is reserved for an empty period. Added a test that asserts the tile stays put at 0.0% so it cannot quietly go back to disappearing.

server/test/tuist_web/live/test_case_live_test.exs
fortmarek: Fair, that was a latent false pass waiting on the next fixture change. `widget_value/2` now parses the element and returns its trimmed text, and the three assertions compare with `==` instead of `=~`.

PR #12353, opened 2026-08-14
fix(kura): label a completed artifact stream as ok instead of aborted

esnunes:
## What changed

`InstrumentedArtifactStream` (`kura/src/http.rs`) now knows how many bytes the response promised, and uses that to decide the `result` label it records on drop:

- `instrument_artifact_stream` passes `Some(manifest.size)` — the same value that becomes the response's `Content-Length` — into the wrapper.
- `Drop` records `result="ok"` when the stream yielded at least the expected bytes and `result="aborted"` only when it yielded fewer.
- `Poll::Ready(None)` still records `ok`, a stream error still records `error`, and `record_once` is still single-shot, so a stream that already reported a terminal outcome is not double-counted on drop.
- When the expected size is unknown (`None`), drop stays conservative and records `aborted`. Every production construction site has a manifest, so this only applies to a caller that cannot state a size; recording `ok` there would let an unverifiable completion silently inflate the success count.

## Why

`kura_artifact_egress_bytes_total` and `kura_artifact_egress_completions_total` labelled almost every successful artifact download as `result="aborted"`. The label was therefore useless: a completed download and a real client abort were indistinguishable, so neither an abort rate nor a "bytes actually delivered" figure could be derived from it.

## Root cause

Three things combined:

1. `InstrumentedArtifactStream` recorded `ok` only on `Poll::Ready(None)` and its `Drop` impl recorded `aborted` for every other termination (`kura/src/http.rs`, `record_once` / `poll_next` / `impl Drop`).
2. Artifact responses always carry a `Content-Length` taken from the manifest size (`apply_artifact_response_headers`, `kura/src/http.rs`).
3. In hyper (1.9.0 is what `kura/Cargo.lock` pins; the same code path exists in 1.10.x), once the fixed-length encoder has written the declared byte count, `Conn::write_body` moves `state.writing` off `Writing::Body` to `Writing::KeepAlive`/`Closed` (`hyper-1.9.0/src/proto/h1/conn.rs:704-727`). On the next dispatcher iteration `can_write_body()` is false, so the dispatcher sets `clear_body = true` and drops the body **without polling it again** (`hyper-1.9.0/src/proto/h1/dispatch.rs:364-371`).

So the terminating `Poll::Ready(None)` never happens for a fully successful streamed download: it always lands in `Drop`, and was recorded as `aborted`.

Production telemetry corroborated this qualitatively: the `ok` share matched the accelerated sendfile serving path exactly — that path records its own result directly (`kura/src/accelerated_file_serving.rs`) and never uses this wrapper — while the `aborted` share matched the streaming path, with near-identical average bytes per read on both. A genuine abort epidemic would not have split cleanly along the serving-path boundary, nor left the two paths with the same average transfer size.

## Why this fix over the alternatives

- **Dropping the `result` label entirely.** That would remove the false signal but also the ability to ever measure client aborts, which is the reason the label exists. The information is recoverable, so discarding it is the wrong trade.
- **Removing `Content-Length` from artifact responses** (so hyper uses a chunked encoder and does poll the body to `None`). This would fix the symptom by changing the wire format of every artifact response — clients would lose the size up front, progress reporting and range/resume behaviour would change, and under a rolling update half the fleet would answer with a length and half without. `kura/AGENTS.md` "Rollout Safety" explicitly rules out node-local behaviour that alters response bytes or headers. Rejected.
- **Recording from the HTTP layer instead of the body** (for example comparing bytes written in middleware). The body stream is the only place that knows how much was actually handed to the transport; the middleware completes when the response head is returned, well before the body finishes.

Comparing yielded bytes against the declared length is the smallest change that restores the label's meaning, touches only the metrics decision, and needs no protocol change.

## Impact

**Users:** none. This is metrics-only. Response bytes, headers and status codes are unchanged — the wrapper's `poll_next` still forwards every chunk and error verbatim, and nothing on the header or admission path was touched. Safe under mixed-version rollout: old and new pods emit the same metric families with the same label names, only the value distribution differs.

**Operators (please note):** the meaning of `result="aborted"` changes at rollout. Before this change it was dominated by successful streamed downloads; after it counts only transfers that ended short of their declared `Content-Length`. Expect `result="aborted"` on `kura_artifact_egress_completions_total` / `kura_artifact_egress_bytes_total` to fall sharply and `result="ok"` to rise correspondingly for the streaming path. Any dashboard panel or alert threshold keyed on `result="aborted"` (or on the `ok` ratio) will shift at the moment of rollout and should be re-baselined; an alert tuned to today's high abort rate will stop firing, and one tuned to a low abort rate could fire during the mixed-version window.

## Tests

Added to the existing `kura/src/http.rs` test module, following the neighbouring convention of asserting on `metrics.render()` output:

- `a_stream_dropped_after_the_expected_bytes_records_ok` — yields exactly the expected bytes, then drops without a terminal poll (mirroring what hyper does), expects `ok` and no `aborted`.
- `a_stream_dropped_before_the_expected_bytes_records_aborted` — yields a short read, then drops; expects `aborted` and no `ok`.
- `a_stream_polled_to_completion_records_ok_once` — polling to `None` records `ok`, and the subsequent drop does not add a second completion.
- `a_failing_stream_records_error_once` — a stream error records `error`, and drop does not reclassify it.

Assertions read the counter values out of the rendered exposition rather than just checking that a label is present, so a regression that flips the label is caught.

## Validation

Run from `kura/`:

- `mise run clippy` (rules_rust clippy aspect over `//...`, warnings as errors) — **passed** (`Build completed successfully, 6 total actions`). It caught one type-inference error in the new test helper on the first run, which was fixed before this commit.
- `mise run test-unit` (`bazel test //...`) — **passed**: `667 passed; 0 failed; 1 ignored`, `//:kura_lib_test PASSED`, `//:kura_bin_test PASSED`. The four new tests were confirmed present and passing in the run output.
- `mise run format` — run before committing; produced no further changes.

The end-to-end shellspec suite was not run.

### DIFF
--- a/kura/src/http.rs
+++ b/kura/src/http.rs
@@ -3594,6 +3594,7 @@ where
     InstrumentedArtifactStream::new(
         state.metrics.clone(),
         manifest.producer,
+        Some(manifest.size),
         stream,
         request_guard,
     )
@@ -3603,6 +3604,7 @@ struct InstrumentedArtifactStream<S> {
     inner: S,
     metrics: Metrics,
     producer: ArtifactProducer,
+    expected_bytes: Option<u64>,
     _request_guard: Option<InflightGuard>,
     started_at: Instant,
     yielded_bytes: u64,
@@ -3613,6 +3615,7 @@ impl<S> InstrumentedArtifactStream<S> {
     fn new(
         metrics: Metrics,
         producer: ArtifactProducer,
+        expected_bytes: Option<u64>,
         stream: S,
         request_guard: Option<InflightGuard>,
     ) -> Self {
@@ -3620,6 +3623,7 @@ impl<S> InstrumentedArtifactStream<S> {
             inner: stream,
             metrics,
             producer,
+            expected_bytes,
             _request_guard: request_guard,
             started_at: Instant::now(),
             yielded_bytes: 0,
@@ -3691,7 +3695,17 @@ where
 
 impl<S> Drop for InstrumentedArtifactStream<S> {
     fn drop(&mut self) {
-        self.record_once("aborted");
+        // Artifact responses declare a Content-Length, and hyper drops a
+        // fixed-length body as soon as the encoder has written that many bytes
+        // instead of polling it to `None`, so a fully delivered stream reaches
+        // this path just like a client abort does. Only the byte count tells
+        // them apart. An unknown expectation stays conservative and counts as
+        // aborted.
+        let result = match self.expected_bytes {
+            Some(expected) if self.yielded_bytes >= expected => "ok",
+            _ => "aborted",
+        };
+        self.record_once(result);
     }
 }
 
@@ -7106,6 +7120,7 @@ mod tests {
         let instrumented = InstrumentedArtifactStream::new(
             context.state.metrics.clone(),
             ArtifactProducer::Xcode,
+            Some(64),
             stream,
             Some(context.state.start_http_request(HttpTrafficClass::Public)),
         );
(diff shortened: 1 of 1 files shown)


kura/src/http.rs
```diff
@@ -3594,6 +3594,7 @@ where
     InstrumentedArtifactStream::new(
         state.metrics.clone(),
         manifest.producer,
+        Some(manifest.size),
```
esnunes, line 3597: One thing I wanted to flag on the test side: the four new tests all build `InstrumentedArtifactStream::new` directly with a literal expected size, so they pin the `Drop` classifier but not the wiring that actually changed in production — the `Some(manifest.size)` argument here. If that argument regressed to `None`, or picked up the wrong manifest, every streamed download would go back to `result="aborted"` and the suite would still be green — which is exactly the bug this PR fixes.

Would it be worth one test that goes through `instrument_artifact_stream` (or a GET for a stored artifact through `serve_file_reader`) and asserts `result="ok"`? Alternatively, making `expected_bytes` a plain `u64` would make that regression unrepresentable and get the same protection for free.

kura/src/http.rs
```diff
@@ -3691,7 +3695,17 @@ where
...
 impl<S> Drop for InstrumentedArtifactStream<S> {
     fn drop(&mut self) {
-        self.record_once("aborted");
+        // Artifact responses declare a Content-Length, and hyper drops a
+        // fixed-length body as soon as the encoder has written that many bytes
+        // instead of polling it to `None`, so a fully delivered stream reaches
+        // this path just like a client abort does. Only the byte count tells
+        // them apart. An unknown expectation stays conservative and counts as
+        // aborted.
+        let result = match self.expected_bytes {
```
esnunes, line 3704: One case I think survives the fix, and it's worth knowing before anyone re-baselines an alert: HEAD requests to the artifact routes will still record `result="aborted"`.

Of the artifact routes only `ROUTE_API_CACHE_MODULE` registers an explicit `head()` handler (`kura/src/http.rs:268`) — `ROUTE_API_CACHE_CAS`, `ROUTE_API_CACHE_GRADLE`, `ROUTE_V1_CACHE`, `ROUTE_API_METRO_CACHE` and `ROUTE_API_CACHE_KEYVALUE_ID` are GET-only. axum falls HEAD back to the GET handler (`method_routing.rs:1157-1158`: `call!(req, HEAD, head); call!(req, HEAD, get);`), and neither `get_xcode` nor `serve_file`/`serve_file_reader` looks at the method, so a HEAD builds the full instrumented stream. hyper then gives a HEAD response an `Encoder::length(0)` (`hyper-1.9.0/.../h1/role.rs:774`), so the body is dropped without a single poll and `Drop` sees `yielded_bytes = 0 < manifest.size`. The code doesn't say whether anything actually HEADs those paths in practice, but if it does, that's a steady floor an alert would sit on top of.

Rather than just documenting it, I think the cleaner fix is to handle HEAD properly on those routes: answer from an existence check with the right status and headers, and never build the instrumented stream at all. `head_module` (`kura/src/http.rs:2048`) is already exactly that shape — it calls `store.artifact_exists(..)` and returns `204`/`404` without opening the artifact — so the other artifact routes could follow it, with `apply_artifact_response_headers` if you want the HEAD to mirror the GET's `Content-Length`. That way a probe never touches the egress counter, instead of being counted as a transfer that failed.

To be clear, this isn't a regression — HEAD recorded `aborted` before this PR too, so it's fine as a follow-up rather than something to hold this up for.

kura/src/http.rs
```diff
@@ -3603,6 +3604,7 @@ struct InstrumentedArtifactStream<S> {
     inner: S,
     metrics: Metrics,
     producer: ArtifactProducer,
+    expected_bytes: Option<u64>,
```
esnunes, line 3607: Small simplification thought, take it or leave it: nothing ever passes `None` for `expected_bytes`. `instrument_artifact_stream` is the only production constructor and always passes `Some(manifest.size)`, and all five test constructions pass `Some(..)`. That leaves the `_ => "aborted"` fallback, and the comment sentence explaining it, covering a state no caller can produce.

A plain `u64` would drop the branch and shorten the comment — and as a bonus it would make the untested-wiring point I left on `instrument_artifact_stream` impossible rather than merely uncaught.

esnunes (commented):
Nice find — this is a subtle one, and the write-up made it easy to check.

I traced the root cause in the vendored hyper 1.9.0 source and it holds: `conn.rs:704-727` moves `state.writing` off `Writing::Body` once the fixed-length encoder is at EOF, and `dispatch.rs:364-371` then sets `clear_body = true` and `continue`s before `poll_frame`, so the body really is released without a final poll. A fully delivered response lands in `Drop` and never in `Poll::Ready(None)`.

Comparing yielded bytes against the declared `Content-Length` looks like the right minimal fix: it touches only the metrics decision, needs no protocol change, and keeps the label's meaning — which matches the rollout-safety bar in `kura/AGENTS.md` that the chunked-encoding alternative would have broken. I also went looking for downstream consumers: outside `kura/src/`, the only reference to `artifact_egress` in the repo is one line of `kura/docs/architecture.md`, and no in-repo dashboard or alert keys on the `result` label, so nothing in-tree needs to ship alongside this.

Two things I'd like your read on, neither blocking. The tests pin the `Drop` classifier but not the one production line that changed. And HEAD requests to the GET-only artifact routes still land in `Drop` with zero bytes, so `aborted` keeps a floor that isn't a client abort — not a regression, but I think it's worth handling properly rather than only documenting. Details inline, plus one small simplification note. Nothing here blocks it from my side.

esnunes:
already implemented on https://github.com/tuist/tuist/pull/12551

PR #12356, opened 2026-08-14
feat(server): scope cache hit rate alerts to a git branch

fortmarek:

Cache hit rate alert rules can now target a single git branch, the way bundle size rules already do.

### Why

Teams that only fill the cache on their default branch were getting noise. Pull request branches that touch a low-level module or an external dependency legitimately miss the cache, so folding those runs into the rolling window made the rule fire on regressions that were entirely expected. There was no way to say "I only care about the branch that populates the cache".

### What changed

**Branch scoping.** `Tuist.Alerts.evaluate/1` passes the rule's `git_branch` into both the current and previous window queries for the `:cache_hit_rate` category. The combined hit rate averages two sources, so the filter had to reach both: `CommandEvents.cache_hit_rate_metric_by_count/3` (module cache) and `Builds.Analytics.build_cache_hit_rate_metric_by_count/3` (Xcode cache). A blank or `nil` branch means every branch, so existing rules keep behaving exactly as they do today.

**The Branch field** now renders for cache hit rate rules in both the create and update modals. It stays required for bundle size; for cache hit rate the placeholder reads `e.g. main (optional)`, matching the convention the Scheme and Bundle name fields already use. The live description reads "…of the last 100 CI builds **on branch main** has decreased by 20%…" when a branch is set, and drops the clause when it isn't.

**Slack notifications** name the branch in the header and message body, so a channel carrying several rules stays readable.

### An adjacent bug fixed along the way

`CommandEvents.cache_hit_rate_metric_by_count/3` accepted an `:is_ci` option and never applied it. A rule scoped to the CI environment therefore only honoured that scope on the Xcode cache half, silently averaging local module cache runs into the module cache half. Since that is exactly the configuration this feature request came from, and it is the same code path as the new branch filter, it is fixed here rather than deferred. **Existing CI-scoped cache hit rate rules will now compute a slightly different, correct value.**

### An unrelated UI bug fixed on request

Dropdown pickers inside the notifications modals scrolled horizontally, clipping option labels. Not a regression from this work, but small enough to fold in.

The cause was a CSS leak rather than anything in the design system. `content` and `body` are generic Noora part names, and a dropdown item's markup is also `[data-part="content"] > [data-part="body"]`. The page's modal rule used a loose descendant match, so `min-width: 420px` landed on every dropdown *item* rendered inside the modal. Items overflowed the 250px panel and made it scroll — measured `scrollWidth` 432 against `clientWidth` 250.

The rules now match the modal's own boxes by structure (`.noora-modal > [data-part="positioner"] > [data-part="content"] > [data-part="body"]`), which is how `noora/lib/noora/modal.ex` nests them. Fixing it in Noora instead was tried first (`min-width: 0` on the item), but the page rule outranks it on specificity via the `#project-notifications` id selector, so it did nothing and was reverted — `noora/` is untouched.

Worth a follow-up: the same leaky-selector pattern exists on three other pages — `project_bundle_settings.css` (identical rule), `account.css` (`min-width: 392px`), and `project_settings.css` (padding only). `members.css` already carries a `min-width: unset` workaround for it. Left alone here to keep this PR scoped.

### How to test locally

1. Start the server and open a project's **Project Settings → Notifications**.
2. Click **Add alert rule** and set **Category** to *Cache hit rate*. The **Branch** field appears; the description sentence updates live as you type a branch.
3. Save with a branch (e.g. `main`) and confirm only that branch's runs are compared; save with the field blank and confirm every branch is considered.
4. Open a rule's edit modal and expand the Category, Metric, or Environment dropdown — the options render in full with no horizontal scrollbar.

Automated coverage, all passing:

```bash
mix test test/tuist/alerts test/tuist/slack_test.exs test/tuist_web/live/project_notifications_live_test.exs test/tuist/builds/analytics_test.exs test/tuist/command_events_test.exs test/tuist/cache/analytics_test.exs
```

New cases cover branch filtering at the query level (both sources), rule evaluation with a branch set and blank, the `:is_ci` filter that previously did nothing, the create and update forms, and the Slack payload.

### Notes for the reviewer

- No migration. `alert_rules.git_branch` already exists for bundle size rules; this reuses it. `server/data-export.md` needs no update for the same reason.
- Alert rules are not exposed over the API, so no OpenAPI regeneration.
- `server/mix.lock` is deliberately **not** in this PR. Every `mix deps.get` in this repo re-resolves `phoenix`, `phoenix_live_view`, `hackney`, `h2`, `quic`, and `webtransport` to whatever is newest on Hex — three runs today produced three different sets. That looks like the same class of problem as the `plug` drift fixed earlier by pinning with `override: true`, and deserves its own deliberate change.


server/lib/tuist/alerts.ex
```diff
@@ -125,15 +125,17 @@ defmodule Tuist.Alerts do
   end
 
   def evaluate(%AlertRule{category: :cache_hit_rate} = alert_rule) do
-    current_opts = maybe_add_environment([limit: alert_rule.rolling_window_size, offset: 0], alert_rule.environment)
+    current_opts =
+      [limit: alert_rule.rolling_window_size, offset: 0]
+      |> maybe_add_environment(alert_rule.environment)
+      |> maybe_add_git_branch(alert_rule.git_branch)
```
esnunes, line 131: One thing I wanted to flag, and it's the only finding I'd call blocking-ish.

The combined hit rate runs both source queries with the same `limit`/`offset` and folds them with `average_or_nil/2`, which returns the survivor unchanged when one side is nil (`cache/analytics.ex:186-189`). `evaluate/1` calls that twice — `offset: 0` and `offset: rolling_window_size` — so when a source has rows in one window but not the other, the two numbers being compared don't mean the same thing.

Concretely, with `rolling_window_size: 20` and `git_branch: "main"`, on a branch with exactly 20 module-cache events (~0.40 each) and 200 builds (~0.90 each):

- current  = average_or_nil(0.40, 0.90) = 0.65
- previous = average_or_nil(nil,  0.90) = 0.90
- deviation = 27.8% → alert fires

…with every measured run identical in both windows. The mirror case (module events only in the older window) hides a real regression the same way.

This asymmetry is on main already — but it used to need the whole project to sit near the window boundary, which is rare on an active project. Scoping to one branch divides both source counts independently and by different factors, so I think it moves from a corner case to something teams will actually hit. Which is a shame, because it's the same alert noise the PR is setting out to remove.

Would it be worth making the comparison source-symmetric? Something like: compute each source's current and previous separately, and only let a source into the average when both of its windows produced rows. Happy to be told this is acceptable for now and worth a follow-up instead — but if so it might be worth saying so in the docs, since "scope to your default branch" is exactly the config that makes a low-volume source run out.

server/lib/tuist/command_events.ex
```diff
@@ -740,33 +740,51 @@ defmodule Tuist.CommandEvents do
...
+
+    query =
+      case is_ci do
+        nil -> query
+        true -> where(query, [e], e.is_ci == true)
+        false -> where(query, [e], e.is_ci == false)
+      end
+
+    hit_rates = ClickHouseRepo.all(query)
+
```
esnunes, line 787: Related to the window-composition point, and also not introduced here — but branch scoping is what makes it bite.

Both hit-rate reducers (`command_events.ex:797-810`, `builds/analytics.ex:2494-2508`) use a reversed nearest-rank percentile: `:p90 -> trunc(count * 0.1)`, `:p99 -> trunc(count * 0.01)`. Right idea for a higher-is-better metric. But the index floors to 0 when count < 10 (p90) or count < 100 (p99), so the metric becomes literally `min(values)` — and nothing requires a window to actually contain `limit` rows.

Because the branch filter is applied before LIMIT/OFFSET, short windows go from unusual to routine. A p90 rule with `rolling_window_size: 20` where the previous window has 20 rows (index 2 → 0.90) and the current window has 6 rows including one cold-cache run (index 0 → 0.10) reports an 88.9% decrease off a single outlier — and the "Current: 10%" in the Slack message isn't the branch's p90 in any meaningful sense.

Would returning nil when `length(hit_rates) < limit` be reasonable? `check_decrease_regression/3` already treats nil as `:ok`, so the plumbing is free. Alternatively, skip evaluation when the two windows differ materially in row count.

(Worth knowing either way: p99 has effectively been `min(values)` for any window under 100 rows, including at the default rolling_window_size of 100 where the index lands on the second-lowest value.)

server/test/tuist/alerts/alerts_test.exs
```diff
@@ -783,6 +783,129 @@ defmodule Tuist.AlertsTest do
       assert data.current == 0.7
       assert data.previous == 0.8
     end
+
+    test "filters to builds on the configured branch when git_branch is set" do
```
esnunes, line 787: The new query-level tests are good — they assert exact values, so they'd fail if the filter were removed. One gap a level up though.

The whole `describe "evaluate/1 for cache_hit_rate"` block builds its data with `RunsFixtures.build_fixture` only; there isn't a single `CommandEventsFixtures` call in `alerts_test.exs`. So the module-cache half is nil in every one of those tests and `average_or_nil/2` collapses to the builds value.

Which means: deleting the new `git_branch` filter from `CommandEvents.cache_hit_rate_metric_by_count/3` — or the `:is_ci` filter you just added there — leaves the new tests at :787 and :856 passing.

That's worth closing precisely because it's the shape of the bug this PR fixes. The old `:is_ci` bug survived for the same reason: `alerts_test.exs:649` ("filters to CI builds only when environment is ci") seeds only builds, so it passed before the fix and would pass again if it were reverted.

One case seeding both sources would do it — command events on `main` and on `feature`, builds on both, rule scoped to `main`, asserting the exact combined current/previous. As a bonus, that's also the test that would pin down the source-composition behaviour in my other comment.

server/lib/tuist_web/live/project_notifications_live.ex
```diff
@@ -673,15 +677,30 @@ defmodule TuistWeb.ProjectNotificationsLive do
...
+                  metric_category: metric_category,
+                  rolling_window_size: rolling_window_size,
+                  current_unit: current_unit,
+                  unit: unit,
+                  deviation: deviation
+                )
+              else
+                dgettext(
+                  "dashboard_projects",
+                  "Alert when the <strong>%{metric_category}</strong> of the last <strong>%{rolling_window_size} %{current_unit}</strong> on branch <strong>%{git_branch}</strong> has decreased by <strong>%{deviation}%</strong> compared to the previous <strong>%{rolling_window_size} %{unit}</strong>.",
```
esnunes, line 695: Heads-up rather than a blocker, and the pattern is pre-existing — but the PR adds a binding on exactly this line, so it might be worth a moment.

The description sentence is built by Gettext interpolation and handed to `raw/1` (:717). Gettext's `%{}` bindings are plain string substitution with no HTML awareness — `Gettext.Interpolation.Default` builds the result with `to_string(value)` — and `git_branch` has no format or length validation in the changeset, so a branch name is substituted into the markup verbatim.

The edit modal renders the *persisted* value (`form[:git_branch] || rule.git_branch`, heex:578), and Noora's modal portals its body into the DOM at page load — so a stored value reaches other project admins without anyone opening the modal.

Two things bound the impact, and I don't want to overstate it:
- The page's CSP is `script_src: "'self' 'nonce' 'wasm-unsafe-eval'"` with no `unsafe-inline`, so an injected inline handler won't execute. This is markup injection, not script execution.
- `:project` / `:update` is admin-only, so it's admin-to-admin within an org.

And the class is already there: `scheme` flows into this same `raw/1`, `bundle_name` and `git_branch` into the sibling one for bundle_size. I'm definitely not asking you to fix the whole class in this PR.

But escaping the interpolated bindings before they hit `raw/1` — or rendering the sentence in HEEx so LiveView escapes it — is cheap here, and it means the property doesn't rest on the CSP header staying exactly as it is. Your call whether it belongs in this PR or a follow-up.

server/lib/tuist/alerts.ex
```diff
@@ -174,6 +176,10 @@ defmodule Tuist.Alerts do
   defp maybe_add_scheme(opts, ""), do: opts
   defp maybe_add_scheme(opts, scheme), do: Keyword.put(opts, :scheme, scheme)
 
+  defp maybe_add_git_branch(opts, nil), do: opts
+  defp maybe_add_git_branch(opts, ""), do: opts
+  defp maybe_add_git_branch(opts, git_branch), do: Keyword.put(opts, :git_branch, git_branch)
```
esnunes, line 181: Nothing to change here necessarily, just flagging the failure mode.

`maybe_add_git_branch/2` strips exactly nil and "", and the changeset neither trims nor validates `:git_branch`. So `"master"` on a project whose default branch is `main`, or `" main "` with whitespace from a paste, makes both queries return [] → both windows nil → `check_decrease_regression` returns `:ok`, forever, with no error surfaced in the UI or the logs.

The user's mental model is "cache regressions on main are monitored" and nothing ever fires.

A `String.trim/1` in the changeset would kill the whitespace half cheaply. The typo half really wants a UI signal ("no runs matched this branch yet"), which I'd guess is out of scope for this PR — but it might be worth a follow-up, especially since a legitimately quiet branch looks identical today.

server/lib/tuist/slack.ex
```diff
@@ -525,10 +529,14 @@ defmodule Tuist.Slack do
...
   end
 
-  defp format_alert_message(%Alert{alert_rule: %{category: :cache_hit_rate, metric: metric}} = alert) do
+  defp format_alert_message(
+         %Alert{alert_rule: %{category: :cache_hit_rate, metric: metric, git_branch: git_branch}} = alert
+       ) do
     deviation = calculate_decrease_deviation(alert)
 
-    "*Cache hit rate #{alert_metric_label(metric)} decreased by #{deviation}%*\n" <>
+    branch_suffix = if is_binary(git_branch) and git_branch != "", do: " on `#{git_branch}`", else: ""
```
esnunes, line 537: Minor, and the same shape as the existing `scheme` (:511, :529) and `bundle_name` (:559) interpolations.

`branch_suffix` puts the branch into a code span inside a `mrkdwn` section. A backtick is a legal git ref character, so a branch like ``fix-`quote`-parsing`` breaks the span. An unvalidated `*` closes the surrounding bold early. And Slack parses `<url|label>` in section text, so a crafted branch name could drop an arbitrary hyperlink into an alert going to the whole channel — a wider audience than the admins who can set the field.

Escaping `&`, `<`, `>` before building the mrkdwn (Slack's required escapes) would cover it, or just drop the backticks for a plain suffix.

For the record the new title clause is right: correctly ordered before the generic one, properly guarded on a non-blank binary, and emitted as `plain_text` so nothing is interpreted there.

esnunes (approved):
Nice change — it's well scoped, it follows the conventions the bundle size rules already established, and the blank-branch path genuinely leaves existing rules untouched. I especially appreciated that the PR description calls out its own behaviour change for CI-scoped rules rather than letting it land quietly; that's the kind of thing that's easy to skip.

I traced the filter through `Cache.Analytics.cache_hit_rate_metric_by_count/3` and it does reach both halves, which was the thing most likely to go wrong here. The `:is_ci` fix is real and its blast radius is small — both per-source functions have exactly one production caller, and that caller has exactly one caller. The Slack clause ordering and guard are right, the query-level tests assert exact values so they'd fail if the filter were removed, and I checked the CSS selector rewrite against `noora/lib/noora/modal.ex` — the new `.noora-modal > positioner > content > body` chain matches the real DOM.

The one thing I'd genuinely like your read on isn't in the new lines. The combined hit rate paginates each source independently and folds them with `average_or_nil/2`, which returns the survivor when one side is nil. Because `evaluate/1` calls that twice at different offsets, a branch narrow enough to exhaust one source in the older window makes the two compared numbers mean different things — and that can fire a decrease when nothing changed. It's a property of main, not something you introduced; my concern is that branch scoping is exactly what makes it common, and the noise it produces is the same noise the feature exists to remove. The related percentile note is the same story: p90/p99 collapse to `min(values)` on short windows.

The rest is small: a stale branch surviving a category switch (inert before this PR, an active filter after it), a silent no-op on a typo'd branch, the Gettext-into-`raw/1` interpolation which is pre-existing but gains a binding on the line you're touching, Slack mrkdwn escaping, and a one-line error-clearing asymmetry in the edit form.

One aside, no action needed: while checking the CSS I noticed the `#create-flaky-test-alert-modal` / `[data-part="update-flaky-test-alert-modal"]` block matches nothing in the current templates — grep finds those names only in the stylesheet. So the second selector rewrite in that file can't be verified by loading the page. Not this PR's problem, just worth knowing if it ever looks untested.

---

Two notes below land on lines this PR doesn't touch, so I couldn't attach them inline.

**`server/lib/tuist_web/live/project_notifications_live.ex:255-268` (also :313-327, :377, :413) — a stale branch survives a category switch**

Small one, and the state handling here predates the PR — but the change alters what the stale value does.

`update_create_alert_form_category` resets the metric and the branch error but not `create_alert_form_git_branch` (:255-268); the input is only hidden by the `:if`. `create_alert_rule` then sends `git_branch` for every category (:377) and the changeset casts it for every category. The edit path is the same (:313-327, :413).

So: type `main` under Cache hit rate, switch to Build duration, save → the row stores `git_branch: "main"` on a rule that ignores it. Harmless in itself.

The bit I'd flag is the round trip. If someone later edits that rule back to Cache hit rate, `maybe_add_git_branch/2` now injects the stale `main` into both windows and the rule silently stops considering every other branch — no visible cue, and it may just quietly stop firing. Before this PR that value was inert on a cache_hit_rate rule; the branch feature is what activates it.

Probably a one-liner: drop `git_branch` from the attrs when `category not in [:bundle_size, :cache_hit_rate]`, or clear the assign on category change.

**`server/lib/tuist_web/live/project_notifications_live.ex:313-327` — the edit form keeps a stale branch error**

Tiny one. The create-side category handler clears `create_alert_form_branch_error` (:268), but the edit-side one (:313-327) leaves `:branch_error` alone.

That asymmetry used to be invisible: switching a bundle_size rule to any other category hid the Branch field, and the stale error went with it. Now that cache_hit_rate renders the field, the path opens — fail an update on a bundle_size rule with a blank branch, switch the category to Cache hit rate, and the field comes back still showing "Branch is required." even though it's optional there.

One line in the edit handler to mirror :268.

PR #12369, opened 2026-08-15
fix(cli): preserve generated test plans across generation

### LINKED ISSUE #12298: Generated test plans are deleted on consecutive tuist generate runs
### What happened?

A generated `.xctestplan` exists after the first `tuist generate`, but is deleted during the second consecutive generation.

Expected: repeated generation is idempotent and the generated test plan remains available.

### How do we reproduce it?

1. Use Tuist’s official `examples/xcode/generated_app_with_generated_test_plan` example.
2. Run `tuist generate`.
3. Confirm that `Derived/TestPlans/*.xctestplan` exists.
4. Run `tuist generate` again without changing any files.
5. Observe that the test plan is written and then `Derived/TestPlans` is deleted.

This reproduces without custom scripts or project helpers.

### Error log

```text
Side effect: create file Derived/TestPlans/GeneratedTestPlan.xctestplan
Side effect: delete directory Derived/TestPlans
```

### macOS version

26.4.1

### Tuist version

4.201.0

### Xcode version

27.0 (27A5209h)

StefanBoblic:
## What changed

`DeleteDerivedDirectoryProjectMapper` now preserves the `Derived/TestPlans` directory when cleaning up generated project artifacts.

A regression test was added to verify that generated test plans are not included in the deletion side effects.

## Why

Generated test plans are stored in `Derived/TestPlans`.

`DeleteDerivedDirectoryProjectMapper` cleans up the `Derived` directory while preserving generated artifacts that need to remain available across generations. However, `TestPlans` was not included in the list of preserved directories.

As a result, the first `tuist generate` correctly created the generated `.xctestplan`, but a subsequent generation scheduled the existing `Derived/TestPlans` directory for deletion.

This PR preserves `Constants.DerivedDirectory.testPlans` alongside the other generated directories that should survive the cleanup.

Fixes #12298

## Validation

- Updated `DeleteDerivedDirectoryProjectMapperTests` to cover `Derived/TestPlans`
- Verified that the test fails without the fix because `TestPlans` is included in the deletion side effects
- Verified that the test passes after `TestPlans` is added to the preserved derived directories
- Verified that stale derived artifacts continue to be deleted as expected


StefanBoblic:
@fortmarek Hey, I couldn't find any information on how to request review. Could you please take a look?

PR #12392, opened 2026-08-17
feat(infra): copy the OCI mirror artifacts from a builder

fortmarek:
Adds a one-shot workflow that copies the OCI mirror artifacts from GHCR to the Tuist registry, running on a bare-metal builder.

### Why this is needed

[#12334](https://github.com/tuist/tuist/pull/12334) moved the image builds onto our own registry. It moved the *reads* along with the writes — correctly, since leaving them split would have had every later build cloning a base nobody updates. What it did not do is move the artifacts already published to GHCR.

The result is that `main` currently points `macos-xcode-image.yml` at `<registry>/xcode-xips`, which is empty:

```
$ curl -u ... https://oci.tuist.dev/v2/_catalog
{"repositories":[]}
```

So the next dispatch of it fails at its first step, with the "not in the mirror" error the workflow itself prints. That is a gap I left in the cutover, and this closes it. The runner and xcresult builds have the same dependency once a base image is published.

### Why it runs on a builder

I first tried the copy from a laptop. It failed, and the failure is worth recording because it is not a registry bug:

- 64 MB blob → **succeeds**. Auth, TLS, chunking and the Tigris backend are all fine end to end.
- 512 MB and 1 GB → **fail**, and the registry logs the reason: `client disconnected during blob PUT  contentLength=1073741824  copied=171966347  error="unexpected EOF"`. The connection dies mid-body; there is no server-side error.
- Measured throughput from a laptop to the tailnet-only registry: **~1 MB/s** (64 MB in 61s).

At that rate a 2.3 GB `.xip` needs ~40 minutes on one unbroken connection and a ~50 GB image needs ~14 hours. The uploads are not hitting a size limit, they are running long enough for something in that path to drop them. Things ruled out along the way: nginx config (`client_max_body_size 0`, `proxy_request_buffering off`, 900s timeouts all correct), nginx OOM (zero restarts, 22Mi of a 256Mi limit), and disk (500 GB free, no ephemeral limits).

Builders sit on a datacenter link and are the same host class that pushes the images, so running the copy there both avoids the laptop path and exercises the one the image workflows actually use.

### Behaviour

Copies every tag GHCR holds when none are named, and verifies against the destination rather than trusting per-tag exit codes. One unreachable artifact warns instead of stranding the rest; the summary decides the job's outcome. `oras copy` skips blobs the destination already holds, so re-running is cheap.

### How to test

```
actionlint .github/workflows/oci-mirror-copy.yml
```

Two findings, both the custom-runner-label warnings every builder workflow carries (`bare-metal`, `vm-image-builder`) — same as `macos-xcode-image.yml`.

The real test is a dispatch once this is on `main`, which is also the only way to dispatch it at all since `workflow_dispatch` registers from the default branch:

```
gh workflow run oci-mirror-copy.yml -f repository=xcode-xips
```

Then `macos-xcode-image.yml` becomes runnable, and that is the run that answers the ~50 GB question.

### Notes

- **Delete this once the mirror is populated.** It is a migration tool, not a standing capability.
- There is a `pushtest` repository in the registry (one 64 MB blob) left from the threshold testing above. Harmless, and worth deleting alongside this.
- Fleet image pins are still on GHCR, so nothing is broken today — but until this runs, the image-build workflows on `main` cannot succeed.

### DIFF
--- a/.github/workflows/oci-mirror-copy.yml
+++ b/.github/workflows/oci-mirror-copy.yml
@@ -1,0 +1,111 @@
+name: OCI Mirror Copy
+
+# Copies OCI artifacts from GHCR to the Tuist registry, on a bare-metal
+# builder rather than from a laptop.
+#
+# This exists because moving the image builds to our own registry moved
+# the *reads* with the writes, but not the artifacts already published.
+# `macos-xcode-image.yml` clones its .xip from `<registry>/xcode-xips`,
+# and until the .xips are copied across, every dispatch of it fails at
+# its first step.
+#
+# It runs on the builder for a practical reason: the registry is
+# tailnet-only and an operator laptop reaches it at around 1 MB/s, which
+# is not enough to hold a single connection open for a multi-gigabyte
+# blob — uploads there die mid-body with `client disconnected during blob
+# PUT`. The builders sit on a datacenter link and are the same host class
+# that pushes the images, so a copy here also exercises the path the
+# image workflows actually use.
+#
+# Expected to be deleted once the mirror is populated. `oras copy` skips
+# blobs the destination already has, so re-running is cheap and safe.
+
+on:
+  workflow_dispatch:
+    inputs:
+      repository:
+        description: "Repository to copy, e.g. `xcode-xips`"
+        required: true
+        default: xcode-xips
+      tags:
+        description: "Space-separated tags. Empty copies every tag GHCR holds."
+        required: false
+        default: ""
+
+concurrency:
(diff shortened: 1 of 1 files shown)


fortmarek:
Closing in favour of https://github.com/tuist/tuist/pull/12393, which removes the need for a migration rather than automating one. The .xip mirror should not have moved off GHCR: its only writer is a laptop task, and a laptop cannot sustain a multi-gigabyte upload to the tailnet-only registry. Reverting that half means there is nothing to copy across.

PR #12397, opened 2026-08-17
revert(infra): publish the VM images to GHCR again

fortmarek:
Reverts the image builds to GHCR. The release pipeline has been failing since #12334 merged.

### What broke

Every image job fails at the registry login:

```
failed to validate the credentials for oci.tuist.dev:
Get "https://oci.tuist.dev/v2/": dial tcp: lookup oci.tuist.dev: no such host
```

In [run 32010833358](https://github.com/tuist/tuist/actions/runs/32010833358) that took out `Build runner image (Xcode 26.0.1)`, `Build runner image (Xcode 26.5)` and `Release xcresult processor image`, with the remaining matrix legs cancelled by `fail-fast`. The server deploy itself is unaffected — canary, acceptance tests, backward-compatibility and the production deploy all succeeded — so this is image publishing, not a customer-facing outage.

### Root cause

`oci.tuist.dev` is a CNAME to a MagicDNS name, and that name only resolves through Tailscale's resolver:

```
$ dig @8.8.8.8 oci.tuist.dev
oci.tuist.dev.  300  IN  CNAME  tuist-oci-production.taild6d7bb.ts.net.
;; status: NXDOMAIN
```

The builders are on the tailnet but do not resolve through it, so the name is unresolvable there.

The `oci-registry-login` action asserted the opposite, in a comment: *"builders join at bootstrap, before kubelet, so a job that reaches this step is already on it."* Being on the tailnet and resolving through it are different things, and I conflated them.

Worth being explicit about why the earlier validation missed this: everything I checked by hand ran from an operator laptop, which does resolve through MagicDNS. The 64 MB push that proved auth, TLS, chunking and the Tigris backend all worked — from a laptop. The DNS assumption was the one thing that path could not test, and it is the one that was wrong.

### Why revert rather than fix forward

The fix is not a code change. It needs one of:

- MagicDNS on the builders — host bootstrap change plus a fleet drift roll, and it risks disturbing the DNS those hosts already depend on
- a hosts entry pinning the tailnet address — bootstrap change plus a roll, and brittle if the address changes
- an IP SAN on the certificate so the tailnet address can be dialled directly — chart change plus a deploy, but no DNS dependency at all

Each needs a roll or a deploy, and none should be rushed while releases are failing. Reverting takes about a minute and restores a known-good state.

### Scope

The registry stays deployed and is untouched. Nothing consumed it — the fleet pins never moved off GHCR — so this restores the previous working arrangement rather than changing where the fleet pulls from.

Chunk size returns to the action's default of 3, the largest GHCR accepts. The 64 MB chunking is still the reason to re-land this: ~1,000 requests for a ~50 GB image instead of ~17,000.

### How to test

```
actionlint .github/workflows/macos-xcode-image.yml .github/workflows/runner-image.yml .github/workflows/xcresult-processor-image.yml .github/workflows/server-production-deployment.yml
```

4 / 2 / 2 / 30 findings, all matching the pre-#12334 baselines — pre-existing custom-runner-label and shellcheck notes. Every `tart-push` call site was checked to carry `TART_REGISTRY_*` again, at job level where the job already set it and at step level otherwise; the login action used to export those via `$GITHUB_ENV`, so removing it would have left three push steps with no credentials.

The real test is the next release, or a manual dispatch of `runner-image.yml`.

### Re-landing

The IP SAN option looks best: it removes the DNS dependency entirely rather than working around it, and it is a chart change plus a deploy with no fleet roll. Worth confirming a builder can reach the tailnet address at all before committing to it — that is the assumption I should have tested the first time.

Also still outstanding: `ghcr.io/tuist/pushtest` (512 MB of random bytes from my throughput testing) needs deleting by someone with `delete:packages`.


fortmarek:
Closing in favour of https://github.com/tuist/tuist/pull/12398, which fixes the DNS problem rather than reverting the move. The registry is published as an A record onto its tailnet address, sourced from the Service status so it re-syncs when the address changes — which was the reason the CNAME existed in the first place.

PR #12398, opened 2026-08-17
fix(infra): stop workflows from running inside forks

### LINKED ISSUE #12396: ci: workflows run inside forks and spam contributors with failures
### Why is this needed?

When you fork `tuist/tuist` and enable Actions, every `push: main` and `schedule` workflow starts running in your fork too. The deploy, release and infra ones can't succeed there — 1Password vaults, Tuist's clusters, `ghcr.io/tuist/*` — so they fail and GitHub emails you about each one. `Update Tuist CLI` fails once a day, forever, whether or not you're touching anything.

Example: https://github.com/zunda-pixel/tuist/actions/runs/32007898116/job/95321090383 — `Load mgmt kubeconfig from 1Password` fails because `OP_SERVICE_ACCOUNT_TOKEN` doesn't exist in the fork.

The CI workflows are noise too. They target self-hosted `tuist-*` runners a fork doesn't have, so they sit queued until GitHub fails them after 24 hours — and they'd only be re-validating a `main` that upstream has already built.

### Steps to address the need

Add `if: github.repository == 'tuist/tuist'` to the root jobs of every workflow that can fire inside a fork (`push`, `schedule`, `workflow_run`). Those runs are then reported as skipped, and GitHub doesn't notify on skipped runs.

Pull requests keep working as they do today: a PR from a fork runs in `tuist/tuist`, so the guard is true. Jobs that only run on `pull_request` events don't need the guard for the same reason.

zunda-pixel:
Resolves #12396

Forking tuist/tuist and enabling Actions makes every `push: main` and `schedule` workflow run in the fork too. The deploy, release and infra ones can't succeed there (1Password vaults, Tuist's clusters, `ghcr.io/tuist/*`), so they fail and GitHub emails the fork owner about it — `Update Tuist CLI` alone failed once a day, 12 times in my fork (example: https://github.com/zunda-pixel/tuist/actions/runs/32007898116/job/95321090383). The CI workflows aren't any better: they target self-hosted `tuist-*` runners a fork doesn't have, so they sit queued until GitHub fails them after 24 hours, re-validating a `main` that upstream has already built.

So every workflow that can fire inside a fork — `push`, `schedule` or `workflow_run` — now carries `if: github.repository == 'tuist/tuist'` on its root jobs. Those runs are still created, but every job is skipped, and GitHub doesn't notify on skipped runs.

Pull requests are unaffected: a PR from a fork runs in `tuist/tuist`, so the guard is true and every check behaves exactly as before. Jobs that only run on `pull_request` events are left alone for the same reason.


zunda-pixel:
Hi @esnunes @fortmarek, gentle ping on this PR when you have a chance.

The main review points are the workflow coverage and the placement of the repository guards, especially around root jobs and `always()` conditions. If reviewing all 34 workflow files at once is too large, I’m happy to split the change into smaller PRs. Thanks!

pepicrft (approved):

PR #12400, opened 2026-08-17
fix(infra): publish the cache-volume HEAD digest from the settled image

fortmarek:
## What the logs showed

Since #12376 put `tart-kubelet` host logs into Loki, every runner mini has been logging the same convergence decline for account 3:

```
converge: image digest does not match HEAD; keeping local master
want: 869027cf2e5ce71745749c195ac6871339287362
got:  42a3aac2ddbd2399aedac7bf17995084ae0e61a2
```

117 occurrences in 24h, all 9 hosts, one account, one want/got pair, unchanged for at least four days.

## Root cause: the digest described bytes that had not settled

`capture_cache_state` measured the inventory through the job's own **live read-write mount**, and the image was uploaded after `detach_cache_image`. The digest is both the HEAD's `tree_digest` and the immutable object key, so it is a claim about the bytes the upload sends — and anything writing to the image between the measurement and the detach breaks that claim permanently.

The window is not theoretical. It is the reason `detach_cache_image` polls five times and then forces: processes outlive the runner (a lingering build service, or the size-capped compilation-cache store's own asynchronous flush/prune, which is busiest for the largest caches). Every `~cas/` line carries a file **size**, so one late append is enough. Reproduced on a real sparse image — a 512-byte append between the measurement and the detach moves the digest, and the new `TestLiveMountDigestDivergesFromTheUploadedImage` pins it.

This also explains why the pair never varied across nine hosts and four days: the digests are stable because nothing can move, not because the trigger was deterministic. One bad publish froze the account, and the freeze then sustained itself.

## Why it froze rather than costing one download

Convergence verifies the downloaded object and declines — correctly, since that check is what stops a corrupt master propagating fleet-wide. So no host obtains that generation, and every promote is then a cold one whose base generation 0 the fast-forward rejects because a HEAD exists. The account can neither adopt the HEAD nor replace it, on any host, indefinitely.

Confirmed against production (read-only):

- no `tuist.dev/cache-master-3` Node label anywhere in the fleet — masters exist only for two other accounts, so account 3 is cold on all 9 hosts
- `tart_kubelet_cache_volume_converged_total` is **0** on every host
- `cache_volume_promote_total` shows only `rejected`, with `materialize_total{result="cold"}` and `outcome_total{outcome="discarded"}` climbing

## The fix

**Publish from the settled image.** `capture_settled_inventory` re-attaches the already-detached file **read-only** — the same view the verifying host uses — and measures that. The teardown order becomes: sample the live-mount-only signals (fill %) → detach → measure the settled image → authorize promotion and upload. A branch that cannot be measured is withdrawn from promotion instead of published, so an image too torn to mount read-only now fails there rather than becoming every host's master.

**Make the wedge recoverable.** A single bad publish should not need a human to notice. On a digest mismatch, the host stages the disproved digest as `volume-head-unverifiable` in the status share and the guest relays it as `unverifiable_digest` with both promote requests; `VolumeHeads.bump_head/6` then lets that cold promote retire the lineage at `generation + 1` (never resetting, so local masters elsewhere stay comparable against one monotonic counter). The superseded poisoned object is reclaimed on the existing supersession path.

Three things keep that narrow:

- being unable to **measure** the image is now distinguished from measuring a **different** one. The first is a local fault that says nothing about the object and declines quietly; only the second — reproducible on every host that fetches it — is reported.
- the report retires only the exact digest it names, under a compare-and-swap on `(generation, tree_digest)`, so a HEAD that moved in the meantime is untouched.
- it is the cold-promote escape only. A warm job on a stale base still loses the fast-forward, report or not.

**The digest check is not weakened or skipped anywhere.** The pre-flight had to learn the same rule, or its 409 would turn away the only promote that can unwedge an account before the bump ever saw it.

## Impact

Account 3's cache has been frozen cold fleet-wide for 4+ days; this restores it and closes the class of failure. Note the self-heal needs all three pieces deployed (server, `tart-kubelet`, runner image) — the server change alone does not unwedge it, since the evidence comes from the host. It also piggybacks on a promote, so it fires on the account's next successful cache-changing job; a wedged account runs cold by definition, so those are its ordinary jobs.

If the account should be unwedged before the runner image rolls, its HEAD row can be retired by hand instead — that is a production data change and needs the JIT flow, so it is not part of this PR.

## Validation

- extracted the shipped `cache_inventory` / `capture_settled_inventory` from `dispatch-poll.sh` and replayed the poisoning scenario against a real sparse image: the published digest now equals the host-verified digest of the uploaded bytes (previously diverged)
- new server tests for retirement (accept, reject without a report, reject a report naming another digest, no substitute for a stale warm base, no cross-account effect) and for the pre-flight mirroring it; verified they fail without the change
- new host test that a mismatch stages the report and a measurement failure does not, with the local master untouched either way
- `go test ./...` in `infra/tart-kubelet` green; 565 server runner tests green; `bash -n` + `shellcheck` clean on the guest script (two pre-existing warnings unchanged)


fortmarek:
Fixed in c51ed82cdd — the finding was correct, and it left the wedge intact in its more likely shape.

`bump_head/6` discarded `opts` on the `base_generation > 0` clause and the pre-flight gated the retirement rule on base 0, so a host at generation 5 against a poisoned generation 6 proved the mismatch and was still rejected for a stale base. A stale-warm host is stuck identically to a cold one — it can never reach the poisoned generation, since adopting it is exactly what the verification refuses, so its base never catches up. And with the publisher of that generation evicted (the state account 3 was found in: no `cache-master` label anywhere in the fleet), a cold-only escape leaves every survivor permanently rejected.

Both clauses now apply the ordinary rule first and fall through to the same retirement CAS on `(generation, tree_digest)`, and `fast_forward_viable?` mirrors it for either base.

I reworked the test that asserted the old cold-only framing (`does not let a report substitute for a stale warm base`) — its premise was the bug. What replaced it:

- `lets a stale-warm host retire it too, not just a cold one` — HEAD 2 poisoned, reporter at base 1 → `{:ok, 3}`
- `still rejects a stale warm base whose report names a superseded digest` — this is the invariant the old test was reaching for, and it is the one that keeps the report from becoming a way around the compare-and-swap: retirement only ever applies to the digest the HEAD stands on right now
- the pre-flight test gained a stale-warm case

Both new assertions verified red against the previous commit. 566 server runner tests and `go test ./internal/podagent` green; `git diff --check`, `mix credo --strict`, `bash -n` and `shellcheck` clean (two pre-existing script warnings unchanged).

No new capability for a runner: it could already claim the standing digest was unverifiable from a cold base, so widening to stale-warm changes nothing about the trust boundary — an authenticated runner can degrade its own account's warmth, never another's, and never correctness. The digest check on adoption is still untouched.

PR #12403, opened 2026-08-17
fix(server,cli): build the shard suite inventory from several runs and the branch under test

fortmarek:
A suite-granularity shard plan enumerated a module's suites from a single historical test run, which under sharding is one shard's worth of suites. Everything the plan didn't know about fell to the catch-all shard, so a reported run had three shards finish in ~3.5 minutes each while the fourth ran for 34 minutes. The same report included shards running a different selection than the build phase printed, which is a second, independent problem with how a shard binds to its plan.

### What was wrong

**The suite inventory was one test run.** `latest_branch_suite_units` took `argMax(test_run_id, ran_at)` per branch and module and expanded only that run. A sharded execution uploads a test run per shard as each shard finishes, and the catch-all shard, which carries every suite the previous plan didn't know about, finishes last. On the reported run the uploads spread from 10:56 to 12:52 while the next plan was created seconds after its build, so the inventory saw whatever fraction had been ingested. The suites it missed were exactly the ones already unplanned, so they stayed unplanned run after run.

**The branch it read from was usually the wrong one.** Branch preference is the `git_branch` parameter, then the linked build run's branch, then the project's default branch. The CLI never sent `git_branch`, and the build run is written through an async ingestion buffer, so it is typically still unreadable when the plan is created a second later. Plans silently fell back to the default branch.

**A run narrowed to a handful of tests was read as the module's inventory.** On the reported project the default branch runs only a nightly `tuist test run --test-targets` selection of eight suites, against roughly fifty-four on the branch being built. That run executed what it was given, not what the module holds, but nothing recorded what it had been asked for, so the plan inventoried eight suites and left forty-six to the catch-all shard. The existing `test_suite_count > 0` guard does not catch this, because the module is present in the run, just with a rump of its suites.

**Suites that are already known were being inferred anyway.** A test plan that restricts a target bakes the selection into the built products as `OnlyTestIdentifiers`, and the `.xctestrun` a plan is created from is already read for its module list. Nothing used it, so a restricted module's inventory was guessed from history when it was sitting in the bundle.

**A reference names the CI run, not the build job inside it.** Every build job of a matrix and every re-run create their plans under the same reference, and the server can only answer with the most recent one. The reported CI run had 101 plans under a single reference. A shard that resolved by reference therefore downloaded another job's test products and ran its selection. GitHub Actions compounds this: job outputs are shared across matrix legs, so `needs.build.outputs.matrix` from a matrixed build job holds only the last leg's `shard_plan_id` even in a workflow wired the way the docs show.

### What changed

- **The inventory unions the last five CI runs** per branch and module instead of one. That restores the full set under sharding and also stops a selective-testing run from hiding suites it skipped. Five bounds how long a deleted suite can linger, rather than the whole 30-day timing window. Written as raw ClickHouse because it needs `LIMIT n BY`, which Ecto cannot express, matching the raw suite timing query already in the module.
- **The CLI sends the branch it built from**, as a new `git_branch` parameter on the shard plan endpoint, so the inventory no longer depends on the build run being readable yet. The value is known at build time, whereas the server can only look it up after ingestion. It is read from the path being built, not the invocation directory, since both callers can build a project that lives elsewhere.
- **Runs record the tests the caller asked for** (`test_runs.only_test_identifiers` and `skip_test_identifiers`, new columns). The inventory skips runs limited to specific tests while fuller history exists, and falls back to them when a module has nothing else, because a partial inventory still beats none. Three parties add `-only-testing` and `-skip-testing` to one invocation, so only the caller's own arguments are read, before Tuist appends a shard's selection or the quarantine skips; Tuist's own filters are already recoverable from the shard plan and the quarantine state. A run the caller only excluded tests from still ran everything else and stays evidence, which is why the identifiers are recorded rather than a single flag.
- **A selection the products already carry is used as given.** A test plan that limits a target bakes the selection into the bundle as `OnlyTestIdentifiers`, and the `.xctestrun` a plan is created from is already read for its module list. The CLI now sends those suites and the server plans them verbatim. Modules the bundle limits to nothing still resolve from history, so a partial selection no longer hands every other module to the catch-all shard.
- **Plan binding is not addressed here.** Nothing in a CI environment distinguishes one matrix leg's CLI invocation from another's, and GitHub's own job outputs collide across legs, so a workflow that builds more than one plan per run has to carry the binding itself; `--shard-plan-id` already exists for that and the generated matrices already emit it. A mismatch shows up as a shard running a selection that does not match the one the build phase printed. The build phase now names the plan it created, since that is the value a workflow threads into the flag.
- **The catch-all shard still receives packed work.** With a complete inventory it runs nothing, so reserving it would idle a runner; it is only an aggravator when the inventory is wrong, which the changes above address.

### Schema change

`priv/ingest_repo/migrations/20260817120000_add_caller_test_selection_to_test_runs.exs` adds two defaulted string arrays to `test_runs`. Existing rows read as empty, meaning unrestricted, and the columns are only written by CLI versions carrying this change, so the new rule has no effect on history recorded before the rollout and nothing regresses in the meantime.

### How to test locally

Server:

```sh
mix test test/tuist/shards_test.exs test/tuist_web/controllers/api/shards_controller_test.exs test/tuist_web/controllers/api/tests_controller_test.exs
```

105 pass. `mix credo --strict`, `mix format --check-formatted`, and `mix excellent_migrations.check_safety` are clean. The new cases were each written failing first, and the two that reproduce the report were also run against the previous implementation to confirm the failure: a module whose suites are spread across the shards that ran them with the newest run still being ingested (plans one suite before, all ten after), and a branch with no history of its own inheriting the default branch's narrowed nightly (plans the two smoke suites before, the full set after). A further case covers a run the caller only excluded tests from, which stays part of the inventory.

CLI:

```sh
xcodebuild test -workspace Tuist.xcworkspace -scheme TuistUnitTests -only-testing TuistKitTests/ShardServiceTests -only-testing TuistKitTests/ShardPlanServiceTests -only-testing TuistKitTests/XcodeBuildTestCommandServiceTests -only-testing TuistKitTests/XcodeBuildBuildCommandServiceTests -only-testing TuistKitTests/TestServiceTests -only-testing TuistKitTests/InspectResultBundleServiceTests -only-testing TuistKitTests/InspectTestCommandServiceTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""
```

131 cases pass, covering the branch the plan is created with, the suites read out of a restricted `.xctestrun`, and the identifiers a narrowed `xcodebuild` invocation records.

### Notes for reviewers

- The new API fields were added to `server.yml` by hand and only the swift-openapi-generator step was run, to avoid the ordering churn a full spec export produces. Both request bodies are inlined in their paths rather than `$ref`s to the component schemas, so each property had to be added in both places; only the inlined one is what the client compiles against.
- Over-inclusion in an inventory is safe, which is what makes the union and the declared-suite fallbacks reasonable. Measured on Xcode 26.5: `xcodebuild` ignores `-only-testing` and `-skip-testing` identifiers for suites that do not exist, exiting 0 even when nothing at all matches. The one hard failure is an identifier naming a test target absent from the bundle, which exits 70, and the planner cannot produce one because suite units are constrained to the modules the client's `.xctestrun` declares.
- Existing plans are unaffected. The change is in how a new plan's inventory is resolved, so the first plan created after deploy already sees the full suite set.


pepicrft (approved):
Validated the shard-plan binding behavior and reviewed the focused changes. All 36 checks are passing.

Issue #11782, opened 2026-07-12
Missing CAS object during replay fails the build instead of recompiling


fortmarek:
## What happened

During the production rebench of the snapshot-era cas-plugin (canary.20 against kura 0.15.2), a cold build resolved action-cache entries from a stale snapshot whose blobs had since been evicted by production CAS churn. Two demand loads hit those evicted objects; FETCH_OBJECT correctly reported not-found after exhausting every source (pending instructions, publish cache, snapshot node table, remote batch read) — and the build failed with exit 65:

```
error: CAS error: missing object '0~PrRJCIbHVmuMQ7X9VJNQyyyXAF2aNmUp4xYutmJtjISAGh9GD51YHwmJBDP7_mD1h8MxlKVwM9mdtwecQLEYdQ=='
```

## Why this matters

The eviction window is not closable server-side: any presence gate vouches at a point in time, and a build takes minutes — the per-key path has the same exposure (manifest fetched at minute 0, blobs pulled at minute 8 after eviction). The design assumption was that a genuinely unavailable object degrades to a warning + recompile (Swift's compilation-replay path does exactly that), with clang's hard-fail mitigated by retained fetch instructions and the snapshot fallback. This failure shows a consumer — the error shape suggests swift-driver's own CAS read rather than compilation replay — that treats the miss as fatal.

## What needs doing

1. Attribute the fatal consumer precisely (which task type issues this load, and through which llcas entry point).
2. Decide the mitigation: ideally the plugin degrades a definitive FETCH_OBJECT miss into something the consumer can recover from (e.g. surfacing the miss at action-cache-get level so the key resolves as a miss and the task recomputes, rather than failing a load mid-task) — or, if the consumer is inherently strict, the served association must be invalidated before the consumer commits to replaying it.
3. The staleness that widened the window in this incident is tracked separately (observability shipping in #11780); this issue stands regardless, because eviction can race even a fresh snapshot or a per-key hit.

Reproduced on: canary.20 client, kura 0.15.2, production tuist namespace, 2026-07-12.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12408, opened 2026-08-17
fix(cli): stop serving an action-cache hit the remote cannot back

fortmarek:
Context: [#12245](https://github.com/tuist/tuist/issues/12245), [#12555](https://github.com/tuist/tuist/pull/12555), [#12421](https://github.com/tuist/tuist/pull/12421), [#11782](https://github.com/tuist/tuist/issues/11782).

Rebased onto current main. `OP_BACKED` is now op **6**: #12421 shipped `OP_DRAIN` as op 5 while this branch was open, and the drain op is already called from deployed runner teardown, so it keeps 5.

### The bug

CI's `Unit Tests` job died with

```
error: CAS operation failed: missing object '0~Xcwig...'
```

and zero tests run, then passed on a re-run of the identical commit. It reads as a flake and was retried away repeatedly. It is not a flake: the same key and the same missing digest failed builds a day apart in different workflows, so a single poisoned cache entry was failing builds for as long as it stayed served. The same shape later took down every job on an account's cache-volume master for a full day.

### Which lane this is

The failing jobs run with `TUIST_FEATURE_FLAG_KURA: 1` and the legacy daemon's `Artifact not found` fingerprint appears zero times. This is the kura / cas-plugin lane. Kura is not serving a bad entry here; it was never asked. The bad entry lives in Apple's local action cache, and it arrives by disk-image clone.

### Root cause

The failing task is `ClangCachingMaterializeKey <key>` and the missing digest is **not** that key. It is an interior node of the value graph.

Both guards that exist today probe the **root only**, deliberately, because verifying a whole graph would put a load per node on swift-build's serial task-setup path. They are backed by write-side invariants that make a present root imply a complete closure. Those invariants bind only writers in this crate, and the association at fault had none: the per-account runner cache volume promotes a whole local CAS image between hosts, so a runner inherits associations written by another machine, at another time, possibly by another plugin version.

A local hit then short-circuits before any RESOLVE, so the proxy never receives the closure manifest. `Proxy::fetch_object`'s three instruction sources (this build's resolves, this machine's publications, the instance snapshot) are all empty for an inherited key, so the demand load returns a genuine not-found and clang fails the build. Nothing retracts the association (the ABI has no delete and a re-put with a different value is refused), so it fails on every later run of that key.

**The promote drain gate (#12421) does not close this, and the first draft of this description said it would.** The gate proves the job's *spool* is empty, meaning everything the job put on its publication to-do list reached kura. The associations that poison a master were never spooled: a pre-#12555 demand load wrote its root straight into the store, and an inherited one arrived by image clone. A master promotes clean by that gate and hollow by this one. That is not a defect in the gate, which is exact about the narrower thing it proves; it is why a read-side check is needed at all.

### What changed

`verified_local_get` asks the proxy, through `OP_BACKED`, whether it can produce the key's closure before serving a local hit. Only a definitive remote miss withholds the hit, and it withholds it exactly as an unbacked root already does: by falling through to the remote path, whose miss is the recompile. A cache entry pointing at an object nobody has is semantically a miss, and the correct response is to rebuild it, not to abort the build.

The proxy answers from `backing_lookup`, a dedicated path and **deliberately not `resolve`**. Review of the first version found three P1s in reusing the resolve machinery, and they were the same mistake twice: resolve and this check have different freshness and latency requirements.

- **It never consults `resolved`.** `resolve` serves `resolved` first, and `resolved` keeps every hit this proxy has ever answered, remote-sourced ones included, for as long as it runs. A resolve may trust that because it goes on to fetch the blobs, so a key the remote has since dropped fails materialization and is withheld; the answer checks itself. This check authorises serving a graph that is *already local*, fetches nothing, and would never discover the entry was gone. On a long-lived proxy (a developer's LaunchAgent, a long runner) it therefore vouched for a key from a stale cache hours after the remote stopped backing it, which is the exact failure it exists to prevent. Its `yes` now comes only from the two sources that describe the remote now: a snapshot gated within `SNAPSHOT_BACKING_MAX_AGE`, or the remote itself. A per-key miss no longer falls back to `resolved` either; the remote has just said no, and an older yes does not outrank it.
- **The snapshot state is read once, under one lock** (`backing_source`). Read as three separate acquisitions, a `Fetching -> Ready` transition landing between reads answered "no snapshot" and served the unverified hit, and a full refresh landing between the snapshot read and the age read blessed the old snapshot with the new timestamp. Both races fail toward serving, the direction that reintroduces the bug.
- **The check is bounded at both ends.** The plugin reads the socket for `BACKED_READ_TIMEOUT` (2s) instead of the generic 120s, in front of what used to be a 180s × 3 retried RPC. The proxy's per-key leg is one attempt under `BACKED_REMOTE_DEADLINE` (1.5s, sent as `grpc-timeout` so the server can abandon the work too), deliberately shorter than the client's read, so in the slow case it is the proxy that gives up with an error the plugin reads as `Unknown`, and not the plugin walking away from a handler thread still working for it. Before this, a backend outage stalled every warm local hit on the serial task-setup path for two minutes each.

The verdict carries the remote's value digest, and the plugin compares it against the local association's. When they differ, which happens when a recompile was not byte-reproducible, the verdict does not transfer and the fall-through serves the remote's value instead. That is a substitution, not a recompile.

Per-path counters (`backing checks=/snapshot=/per_key=/unbacked=` on the stats line) make the per-key round trips on the serial path measurable. `per_key` is the number to watch; see cost below.

### What #12555 covers and this covers

#12555 shut the demand-load door for a root that is *not yet local*: it records a withheld root against the digests its closure still owes, and `fetch_object` refuses to produce it until they land. Its own body lists the one shape it does not cover, a root that was already on disk before the pass, because there is nothing to withhold. That is this PR. The two are complementary and touch adjacent code, which is what the rebase was.

### Why this shape, and not the simpler alternatives

Two were seriously considered.

**Verify at write time only.** Four causes have already been closed that way (#12245, #12284, the write-through ordering, #12421), and the failure is still live, because two authors are not reachable from our code: Apple's prune, and remote eviction after the association was written. #11782 states why the second cannot be closed by any presence gate: it vouches at a point in time, and a build takes minutes. An open-ended set of authors means prevention alone does not converge here.

**Verify asynchronously and record a negative override.** Serve the local hit immediately, enqueue the check on a worker, and mark the key dead for the next build if it comes back unbacked. That dissolves all three P1s by construction (nothing on the serial path, so no latency budget and no reason to use a cached answer) and is the better design if a master carries few poisoned keys. Its price is one failed build per poisoned key. The synchronous check was kept because its price is zero failed builds and its measured cost is small (below). The number that would flip that decision is the count of distinct poisoned keys per promoted master, which nobody has measured yet; if it turns out to be hundreds, the asynchronous shape should replace this one.

### Why it is not a blanket veto

Deliberately inert wherever a remote miss is not evidence of a dangling association:

- `upload: false`, where the machine keeps its results off the remote by design, so vetoing would recompile the whole build. Read in the **proxy**, never client-side: the plugin's `upload` flag comes from a compiler option that reaches Swift, while swift-build's clang caching runs against a CAS created with a plugin path and no options, so the clang lane (the one that fails builds) reads as uploading even under an explicit opt-out.
- a durably snapshotless instance (a server without snapshot support, or `TUIST_CAS_PREFETCH=0`), where a round trip per served hit would undo the reason a local hit is worth having. `Fetching` is covered on purpose: the proxy and the build start together on CI, so the first build's gets land mid-fetch, on the cold freshly promoted cache volume where an inherited association is most likely to be hollow.
- no routable instance, no proxy, any transport error, and a store whose generation moved while the remote was being asked (not evidence about the remote, so not a miss).

### Why no protocol version bump

The frame layout is untouched, and the dispatch's existing `bad op` fallback already answers with the status that reads as "cannot tell". A stale launchd proxy left running across a CLI upgrade therefore keeps today's behaviour instead of turning every build cold, which a version bump would have caused. Renumbering to op 6 preserves this: a proxy that predates the op still answers `bad op`.

### Cost

Measured on an M-series Mac, release, against the in-process fake proxy over a real unix socket, so the figure is a transport floor rather than a full accounting of the production proxy's work:

| | before | with the check |
|---|---|---|
| per served local hit | ~49ns | ~14.3µs |
| across ~13.5k hits (warm runner build) | 0.66ms | 193ms |

That is 0.2% of the 96.8s ± 3.9 runner warm build in `AGENTS.md`, about a twentieth of that measurement's own noise band.

**Not measured, and worth measuring before relying on this on runners:** a warm cache volume on CI during the seconds the snapshot is still fetching. Cold CI is cheap there because a cold store has few local hits to check; a warm volume has many, each potentially paying a per-key round trip in exactly the window the design refuses to decline in. The `per_key` counter exists to make that visible. If it is thousands, a cap on per-key lookups during `Fetching` with the overflow treated as `Unknown` is the bounded fallback.

### What this does not cover

- **A blob that disappears after the lookup succeeded.** The check runs once, at the get. If the remote evicts between then and the demand load, the clang lane still dies. That is #11782 and it needs either clang's missing-object to become non-fatal upstream or materializing the full closure before answering the get, which is the cost the whole design avoids.
- **Apple's prune as a fatal author.** An experiment in this branch's test suite (`prune_authorship`, ignored, run on demand) rotated a store six times against a 1 MiB limit under six access patterns. A `load` carried the whole closure forward (root and child both survived), a `contains` probe carried nothing, and no pattern produced root-present-over-missing-interior. The prune produced the root-absent shape, which the existing probe already catches. One Xcode version and a two-node graph, so not conclusive, but on this evidence the fatal authors are inheritance and eviction, not the prune.
- **The cause.** This stops the crash. Removing the CAS store from the cache volume would remove the inheritance class structurally, at roughly 21s on a 97s warm build, and that decision has not been taken on its merits; a purge capability for volumes is being designed separately.

### Also, on the legacy lane

`CASService.load` collapsed every failure into `.error`, including a plain 404, even though `CASLoadResponse.Outcome` has `objectNotFound` and `KeyValueService` has always used its `keyNotFound` equivalent. Same class of bug on the lane that does not go through the plugin; now mapped to `.objectNotFound`. The circuit breaker already classified `.notFound` as a healthy backend, so only the outcome was wrong. Independent of the failure above and would not have prevented it.

### Validation

Rust, after the rebase and the three fixes: 105 lib tests and 22 integration tests pass (1 ignored, the `probe_cost` measurement), plus the crate's other targets.

Red before green on each behavioural fix:

- `a_backing_check_never_vouches_from_the_resolved_cache`: seeds a `resolved` Hit with a registered instruction, a fresh snapshot that lacks the key, and an unreachable remote. Answers `HIT` through the old routing and `ERROR` (Unknown) through `backing_lookup`, and asserts the remote was asked exactly once.
- `a_backing_source_is_classified_from_one_read_of_the_snapshot_state`: fresh Ready answers with that same `Arc`, stale Ready and `Fetching` go per key, `Absent` and unregistered decline.
- `a_proxy_that_never_answers_the_backing_check_is_abandoned_within_the_deadline`: a fake proxy that accepts and sits on the check for 8s. Takes 8.0s under the 120s read (fails the 5s bound) and 2.06s under `BACKED_READ_TIMEOUT`.
- `the_proxy_abandons_its_remote_before_the_client_abandons_the_proxy` pins the deadline ordering.

The original branch's tests (`an_association_the_remote_cannot_back_is_withheld`, `an_association_the_remote_can_back_is_served`, the read-only-project decline, the stale-snapshot age policy) all still pass against the new path.

Not re-run here: the Swift `CASService` change and its test came across the rebase by auto-merge with no conflict; main has not touched those files since the branch was cut. `cargo fmt` and clippy report the same pre-existing findings as main (two clippy errors in untouched `llcas_get_plugin_version`, one `unused_mut`); nothing new is introduced.


pepicrft (approved):
Validated the cache-integrity behavior locally. The currently failing continuous-integration jobs stop before tests start because the runner cache volume has no free space, which is unrelated to this change.

PR #12411, opened 2026-08-17
fix(cli): generate package accessors and map conditional settings

### LINKED ISSUE #12375: Generated projects define SWIFT_PACKAGE without providing the Objective-C resource bundle accessor
### What happened?

An Objective-C target declared in `Package.swift` contains resources and uses SwiftPM's generated `SWIFTPM_MODULE_BUNDLE` accessor:

```objc
NSBundle *ResourceBundle(void) {
#if SWIFT_PACKAGE
    return SWIFTPM_MODULE_BUNDLE;
#else
    return NSBundle.mainBundle;
#endif
}
```

Building the package with `swift build` succeeds. SwiftPM:

- generates `resource_bundle_accessor.h` and `resource_bundle_accessor.m`;
- compiles the generated implementation;
- force-includes the generated header when compiling the Objective-C sources.

After running `tuist generate`, the generated Xcode project:

- defines `SWIFT_PACKAGE=1` for the Clang target;
- creates and embeds the resource bundle;
- does not generate or inject the Objective-C resource bundle accessor.

The source therefore enters the `SWIFT_PACKAGE` branch, but `SWIFTPM_MODULE_BUNDLE` is unavailable and compilation fails.

The generated compile arguments contain `-DSWIFT_PACKAGE=1`, but there is no generated `resource_bundle_accessor.h` or `resource_bundle_accessor.m`.

Expected behavior:

A Tuist-generated project should preserve SwiftPM's resource-accessor behavior for Objective-C package targets with resources, either by generating and injecting an equivalent accessor or by not defining `SWIFT_PACKAGE` when that behavior is unavailable.

### How do we reproduce it?

Sample project: [ObjCResourceBundleRepro.zip](https://github.com/user-attachments/files/31123015/ObjCResourceBundleRepro.zip)

1. Unzip the attached sample project.
2. Run `swift build`. The package builds successfully, including `resource_bundle_accessor.m`.
3. Run `tuist install`.
4. Run `tuist generate --no-open`.
5. Run:

```sh
xcodebuild \
  -workspace Workspace.xcworkspace \
  -scheme ObjCResourceBundleRepro \
  -configuration Debug \
  -sdk macosx \
  CODE_SIGNING_ALLOWED=NO \
  build
```

The generated-project build fails because `SWIFTPM_MODULE_BUNDLE` is undeclared.

### Error log

```text
Sources/ObjCResourceBundleRepro/ObjCResourceBundleRepro.m:5:12: error: use of undeclared identifier 'SWIFTPM_MODULE_BUNDLE'
    5 |     return SWIFTPM_MODULE_BUNDLE;
      |            ^~~~~~~~~~~~~~~~~~~~~
1 error generated.

** BUILD FAILED **

The following build commands failed:
    CompileC .../ObjCResourceBundleRepro.o .../ObjCResourceBundleRepro.m normal arm64 objective-c com.apple.compilers.llvm.clang.1_0.compiler
    Building workspace Workspace with scheme ObjCResourceBundleRepro and configuration Debug
```

### macOS version

26.6.1 (25G76)

### Tuist version

4.203.1

### Xcode version

26.6 (17F113)

### LINKED ISSUE #12377: Platform-conditioned SwiftPM settings duplicate unconditional OTHER_SWIFT_FLAGS
### What happened?

Tuist 4.204.0 duplicates unconditional Swift package settings when the same target also has a platform-conditioned Swift setting.

The minimal package target has these settings:

```swift
let sharedSwiftSettings: [SwiftSetting] = [
    .enableExperimentalFeature(
        "AvailabilityMacro=Repro_v1:iOS 13.0, macOS 13.0, tvOS 13.0, watchOS 6.0, visionOS 1.0"
    ),
    .unsafeFlags(["-index-ignore-system-modules"], .when(platforms: [.macOS])),
]
```

Tuist generates both an unconditional and an SDK-qualified build setting, but the SDK-qualified value repeats the unconditional availability macro:

```text
OTHER_SWIFT_FLAGS = (
    "$(inherited)",
    "-enable-experimental-feature \"AvailabilityMacro=Repro_v1:...\"",
);
"OTHER_SWIFT_FLAGS[sdk=macosx*]" = (
    "$(inherited)",
    "-enable-experimental-feature \"AvailabilityMacro=Repro_v1:...\"",
    "-index-ignore-system-modules",
);
```

Because the SDK-qualified value also includes `$(inherited)`, Xcode passes the availability macro to Swift twice. Swift 6.3 reports a duplicate definition when source code references that macro with `@available`.

I expected the SDK-qualified setting to inherit the unconditional flags and append only the platform-conditioned flag:

```text
"OTHER_SWIFT_FLAGS[sdk=macosx*]" = (
    "$(inherited)",
    "-index-ignore-system-modules",
);
```

### How do we reproduce it?

Sample project:

[tuist-conditional-swift-settings-repro.zip](https://github.com/user-attachments/files/31349010/tuist-conditional-swift-settings-repro.zip)

```sh
unzip tuist-conditional-swift-settings-repro.zip
cd tuist-conditional-swift-settings-repro
mise trust
mise install
mise exec -- tuist install
mise exec -- tuist generate --no-open
xcodebuild build \
  -workspace ConditionalSwiftSettingsRepro.xcworkspace \
  -scheme ReproApp \
  -destination platform=macOS
```

The sample pins Tuist 4.204.0 in `.mise.toml`. The issue can also be reproduced with `bash reproduce.sh`.

### Error log

```text
-define-availability argument:1:1: error: duplicate definition of availability macro 'Repro_v1' for version '0'
Repro_v1:iOS 13.0, macOS 13.0, tvOS 13.0, watchOS 6.0, visionOS 1.0
^
```

The failing compiler command contains the `AvailabilityMacro=Repro_v1` argument twice and `-index-ignore-system-modules` once.

### macOS version

26.5.2 (25F84)

### Tuist version

4.204.0

### Xcode version

Xcode 26.6 (17F113), Apple Swift 6.3.3

pepicrft:
## What changed

Local projects generated from Swift packages now receive the Objective-C bundle accessor when they contain Objective-C sources and resources. Package build settings are also mapped so unconditional Swift flags remain in the base settings instead of being repeated in every platform-specific variant.

Regression tests cover local package resource accessors, separation of unconditional and platform-specific settings, and the existing Mac Catalyst mapping behavior.

## Why

Swift packages at the workspace root could fail to compile when an Objective-C target referenced `Bundle.module`, while unconditional experimental feature flags could be emitted twice after project generation.

Fixes #12375
Fixes #12377

## Root cause

The resource mapper generated Objective-C accessors only for projects marked as external, but root Swift packages are represented as local projects. Separately, the settings mapper included unconditional values while collecting settings for a specific platform, which copied common flags into platform-qualified settings.

## Approach

The resource mapper keeps the existing external-project behavior and recognizes local Swift packages through their package preprocessor definition. The settings mapper now selects either unconditional settings or settings matching the requested platform, while preserving the existing Mac Catalyst compatibility mapping.

## Impact

Objective-C targets in local Swift packages can resolve their generated resource bundles. Unconditional Swift flags are emitted once, avoiding duplicate compiler arguments. Ordinary local Tuist projects and existing external package projects retain their previous behavior.

## Validation

- Ran the complete `ResourcesProjectMapperTests` suite.
- Ran `SettingsMapperTests` and `SettingsMapperConditionalTests`.
- Built the Tuist command-line executable with code signing disabled.
- Generated and built the reproduction attached to #12375 with the locally built Tuist executable.
- Ran `mise run cli:lint`.
- Ran `git diff --check`.

### DIFF
--- a/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
+++ b/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
@@ -14,7 +14,8 @@ import XcodeGraph
 ///      cannot host their own resources).
 ///   2. Writes a `TuistBundle+<Target>.swift` accessor into the target's Derived directory so
 ///      that user code can call `Bundle.module`.
-///   3. For external Obj-C targets, also emits SwiftPM-shaped C bridging files.
+///   3. For Objective-C targets generated from Swift packages, also emits
+///      Swift Package Manager-shaped C bridging files.
 public struct ResourcesProjectMapper: ProjectMapping {
     private let contentHasher: ContentHashing
     private let buildableFolderChecker: BuildableFolderChecking
@@ -120,7 +121,7 @@ public struct ResourcesProjectMapper: ProjectMapping {
     }
 
     private func targetNeedsObjcAccessor(_ target: Target, project: Project) -> Bool {
-        guard case .external = project.type else { return false }
+        guard projectUsesSwiftPackageBuildSettings(project) else { return false }
         let containsObjc = target.sources.contains {
             $0.path.extension == "m" || $0.path.extension == "mm"
         }
@@ -129,6 +130,21 @@ public struct ResourcesProjectMapper: ProjectMapping {
             .isEmpty
         let needsBundle = !target.supportsResources || target.product == .staticFramework
         return containsObjc && hasBundleResources && needsBundle
+    }
+
+    private func projectUsesSwiftPackageBuildSettings(_ project: Project) -> Bool {
+        if case .external = project.type { return true }
+
+        guard let definitions = project.settings.base["GCC_PREPROCESSOR_DEFINITIONS"] else {
+            return false
+        }
+        let values: [String] = switch definitions {
+        case let .array(values):
+            values
+        case let .string(value):
+            value.split(whereSeparator: \.isWhitespace).map(String.init)
+        }
+        return values.contains { $0 == "SWIFT_PACKAGE" || $0.hasPrefix("SWIFT_PACKAGE=") }
     }
 
     /// Files inside buildable folders that should trip bundle synthesis but aren't caught by
--- a/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
+++ b/cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
@@ -82,7 +82,7 @@ public struct ResourcesProjectMapper: ProjectMapping {
             }
         }
 
-        if targetNeedsObjcAccessor(target, project: project) {
+        if targetNeedsObjcAccessor(target) {
             try appendObjcBundleAccessor(
                 to: &modifiedTarget,
                 sideEffects: &sideEffects,
@@ -120,8 +120,8 @@ public struct ResourcesProjectMapper: ProjectMapping {
         return containsSwift || containsSourcesInBuildableFolders
     }
 
-    private func targetNeedsObjcAccessor(_ target: Target, project: Project) -> Bool {
-        guard projectUsesSwiftPackageBuildSettings(project) else { return false }
+    private func targetNeedsObjcAccessor(_ target: Target) -> Bool {
+        guard target.metadata.tags.contains(TargetTags.swiftPackage) else { return false }
         let containsObjc = target.sources.contains {
             $0.path.extension == "m" || $0.path.extension == "mm"
         }
@@ -130,21 +130,6 @@ public struct ResourcesProjectMapper: ProjectMapping {
             .isEmpty
         let needsBundle = !target.supportsResources || target.product == .staticFramework
         return containsObjc && hasBundleResources && needsBundle
-    }
-
-    private func projectUsesSwiftPackageBuildSettings(_ project: Project) -> Bool {
-        if case .external = project.type { return true }
-
-        guard let definitions = project.settings.base["GCC_PREPROCESSOR_DEFINITIONS"] else {
-            return false
-        }
-        let values: [String] = switch definitions {
-        case let .array(values):
-            values
-        case let .string(value):
-            value.split(whereSeparator: \.isWhitespace).map(String.init)
-        }
-        return values.contains { $0 == "SWIFT_PACKAGE" || $0.hasPrefix("SWIFT_PACKAGE=") }
     }
 
     /// Files inside buildable folders that should trip bundle synthesis but aren't caught by
--- a/cli/Sources/TuistLoader/SwiftPackageManager/SettingsMapper.swift
+++ b/cli/Sources/TuistLoader/SwiftPackageManager/SettingsMapper.swift
@@ -46,7 +46,7 @@ struct SettingsMapper {
 
         return try map(
             settings: platformSettings,
-            headerSearchPaths: headerSearchPaths
+            headerSearchPaths: platform == nil ? headerSearchPaths : []
         )
     }
 
--- a/cli/Sources/TuistLoader/SwiftPackageManager/SettingsMapper.swift
+++ b/cli/Sources/TuistLoader/SwiftPackageManager/SettingsMapper.swift
@@ -198,18 +198,19 @@ struct SettingsMapper {
         -> [PackageInfo.Target.TargetBuildSettingDescription.Setting]
     {
         settings.filter { setting in
-            if let platformName, setting.hasConditions {
-                let hasMacCatalystPlatform = setting.condition?.platformNames.contains("maccatalyst") == true
-                let platformNames: [String]
-                if hasMacCatalystPlatform {
-                    platformNames = (setting.condition?.platformNames ?? []) + ["ios"]
-                } else {
-                    platformNames = setting.condition?.platformNames ?? []
-                }
-                return platformNames.contains(platformName)
-            } else {
+            guard let platformName else {
                 return !setting.hasConditions
             }
+            guard setting.hasConditions else { return false }
+
+            let hasMacCatalystPlatform = setting.condition?.platformNames.contains("maccatalyst") == true
+            let platformNames: [String]
+            if hasMacCatalystPlatform {
+                platformNames = (setting.condition?.platformNames ?? []) + ["ios"]
+            } else {
+                platformNames = setting.condition?.platformNames ?? []
+            }
+            return platformNames.contains(platformName)
         }
     }
 }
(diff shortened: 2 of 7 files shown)


cli/Sources/TuistLoader/SwiftPackageManager/SettingsMapper.swift
```diff
@@ -198,18 +198,19 @@ struct SettingsMapper {
...
-                    platformNames = (setting.condition?.platformNames ?? []) + ["ios"]
-                } else {
-                    platformNames = setting.condition?.platformNames ?? []
-                }
-                return platformNames.contains(platformName)
-            } else {
+            guard let platformName else {
                 return !setting.hasConditions
             }
+            guard setting.hasConditions else { return false }
```
esnunes, line 204: One thing I wanted to flag about the platform pass: `settingsDictionary(for:)` still seeds `map(...)` with the injected `headerSearchPaths`, so a platform dictionary is never empty for a target that has a module map.

For a C/Obj-C target that also declares an unconditional `.headerSearchPath("include")` in `cSettings`, the base value now carries that path and the platform value doesn't. They differ, so the `self[key] != newValue` guard in `overlay(with:for:)` stops suppressing the write and `HEADER_SEARCH_PATHS[sdk=...]` gets emitted for all five platforms (nine keys with the simulator variants) where previously there were none - and once `$(inherited)` expands, the injected public-headers path is listed twice for those SDKs.

It shouldn't break a build, but it's the same duplication this PR removes elsewhere. Would it make sense to seed the header search paths only on the unconditional pass?

cli/Sources/TuistLoader/SwiftPackageManager/SettingsMapper.swift
```diff
@@ -198,18 +198,19 @@ struct SettingsMapper {
...
+
+            let hasMacCatalystPlatform = setting.condition?.platformNames.contains("maccatalyst") == true
```
esnunes, line 206: The correctness of this rests on Xcode resolving `$(inherited)` inside an SDK-qualified assignment through the unqualified assignment in the same table - that's what makes it safe for `SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=iphoneos*]` to stop listing `Define1`. The duplication reported in #12377 is good evidence that this is exactly how it resolves (the macro appears twice, `-index-ignore-system-modules` once, and the base assignment is the only possible source of the second copy), so I think the direction is right.

The new tests assert the generated dictionary rather than the flags a target ends up compiling with, and the two deleted cases were the ones encoding the old shape. Did you get a chance to run the #12377 reproduction end to end against the built binary? The validation list mentions the #12375 repro but not this one. An acceptance fixture asserting the resolved flags would also lock this down in a way a dictionary-level assertion can't.

cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
```diff
@@ -131,6 +132,21 @@ public struct ResourcesProjectMapper: ProjectMapping {
         return containsObjc && hasBundleResources && needsBundle
     }
 
+    private func projectUsesSwiftPackageBuildSettings(_ project: Project) -> Bool {
```
esnunes: The `SWIFT_PACKAGE` sniff does identify the generated case correctly - `PackageInfo.projectSettings` puts `GCC_PREPROCESSOR_DEFINITIONS = ($(inherited), SWIFT_PACKAGE=1)` on every package project.

The bit worth knowing about is the line right after it: that dictionary is merged with the user's `PackageSettings.baseSettings` using `uniquingKeysWith: { $1 }` (PackageInfoMapper.swift:2650), so the incoming value wins outright. A project that sets its own `GCC_PREPROCESSOR_DEFINITIONS` in `baseSettings` ends up with no marker, the gate returns false, and #12375 comes back with nothing to point at. It widens in the other direction too - a hand-written Project.swift that happens to define `SWIFT_PACKAGE=1` now takes the SwiftPM bridging path.

Would it be worth keying off something a manifest can't overwrite - a flag on `Project`, or a `ProjectType` case set when the project comes out of `PackageInfoMapper`? Minor related note: `hasPrefix("SWIFT_PACKAGE=")` also matches `SWIFT_PACKAGE=0`.

esnunes (approved):
Thanks for taking both of these on - the description states each root cause precisely, and the two fixes are small and well aimed at them.

I read through the settings pipeline and I think the conditional-settings fix is right. The duplication reported in #12377 is itself the evidence: `$(inherited)` inside an SDK-qualified assignment resolves through the unqualified assignment in the same table, which is what makes it safe for the platform slice to stop carrying the unconditional values. Replacing the two XCTest cases that encoded the old shape with Swift Testing equivalents reads well, and the local-package accessor test sits nicely next to the existing negative case.

Two things I'd like your read on. The platform pass still seeds the injected header search paths, so C/Obj-C package targets that declare an unconditional `.headerSearchPath` now pick up SDK-qualified `HEADER_SEARCH_PATHS` keys they didn't have before, with the injected path listed twice once `$(inherited)` expands. And the `SWIFT_PACKAGE` marker the new accessor gate reads can be replaced wholesale by a user's `PackageSettings.baseSettings`, which would bring #12375 back with nothing to point at.

Nothing here is a blocker from my side - details are in the inline notes.

cli/Sources/TuistLoader/SwiftPackageManager/SettingsMapper.swift
pepicrft: Good catch. I changed `SettingsMapper.settingsDictionary(for:)` so injected module-map header paths are seeded only in the unconditional pass. Platform-qualified dictionaries now contain only header paths declared for that platform, and `SettingsMapperConditionalTests` covers the base plus macOS-conditioned shape. This prevents the injected path from being inherited and listed again.

cli/Sources/TuistLoader/SwiftPackageManager/SettingsMapper.swift
pepicrft: I ran the #12377 reproduction end to end with the branch binary. The generated macOS-qualified `OTHER_SWIFT_FLAGS` contains only `$(inherited)` and `-index-ignore-system-modules`, the availability macro remains only in the unqualified setting, and the `ReproApp` Xcode build succeeds. I did not add a second acceptance fixture because `SettingsMapperConditionalTests` pins the exact generated dictionary while the published reproduction exercises Xcode’s resolved compiler flags.

cli/Sources/TuistGenerator/Mappers/ResourcesProjectMapper.swift
pepicrft: Agreed. I removed the `GCC_PREPROCESSOR_DEFINITIONS` sniff. `PackageInfoMapper` now applies the reserved `tuist:swift-package` metadata tag to every generated package target, and `ResourcesProjectMapper` gates the Objective-C accessor on that marker. Package base settings can no longer disable the accessor, `SWIFT_PACKAGE=0` no longer enables it, and a hand-written project does not opt in accidentally through a compiler definition. Tests cover local and external package targets plus the regular-project negative case.

PR #12414, opened 2026-08-17
feat(infra): reap abandoned tagged Tailscale devices

fortmarek:
Adds a grace-period garbage collector for abandoned tagged Tailscale devices, one per environment.

There is no tracked issue; this started from the billing page reporting 63/50 tagged resources and the question of whether we were really using 63.

## What prompted it

We were not. Auditing the tailnet found 67 tagged devices, 44 of them online and 23 dead:

| Stale for | Tag | Devices |
|---|---|---|
| 0 to 4 days | `tag:tuist-macmini-production` | 17 xcresult-processor VMs |
| 24 days | `tag:tuist-k8s-production` | 1 cluster router |
| 6 to 70 days | `tag:tuist-k8s-canary` / `-staging` | 5 operator egress proxies |

A third of the tailnet was billed for nothing, growing by roughly 10 to 15 devices a week against an included allowance of 50. The 17 processor devices are legible as deploy history: eight distinct ReplicaSet hashes over four days, two devices abandoned per roll.

## Root cause

A tailnet device registration has no garbage collector, and tagged devices have key expiry disabled by default, so a registration outlives whatever created it until something deletes it explicitly. Every producer we run registers a device per Pod:

- xcresult-processor Tart VMs join with `TAILSCALE_HOSTNAME` set to the Pod name, so each roll registers a new device.
- The Tailscale operator's proxy Pods re-register when their state Secret is recreated, stranding the previous device.

`ephemeral=true` is already set on the minted key and does not save us: it only reaps VMs that stay online briefly, and these Pods live for days between releases.

## Why a reaper rather than a stable hostname

Stabilising the Pod-derived hostname (a StatefulSet ordinal, or the Node name, which is 1:1 with the Pod here given the hostPort 9091 constraint) was the obvious fix and does not work. A device is identified by its node key, not its hostname, and a Tart VM booted from an image has no persisted tailscaled state, so every boot mints a new key and therefore a new device whatever it is called.

The tailnet already demonstrates this: there are two `macmini-egress-0` devices carrying the same hostname *and* the same `tag:tuist-k8s-canary`, one online and one dead six days. Renaming would have produced colliding names, not fewer devices. Deleting the dead ones is the only fix available.

## What shipped

`tailscale-device-reaper`, modelled on the existing `released-pv-reaper` in the same chart and following its idioms (dry-run default, re-validate before acting, grace window as recovery window, no PrometheusRule because managed clusters scrape via Alloy).

Two design points are worth a reviewer's attention.

**Per-env, not one sweep.** All envs share a single tailnet, so a single production reaper covering every tag was the simpler option and is what the first commits implemented. We chose per-env instead, for failure isolation (a stalled production reaper no longer stops canary and staging being swept) and to confine each credential to one env's tags, matching the confinement the macOS-fleet join credential already has. The cost is three OAuth clients, three 1Password items and three arming decisions.

That choice has a correctness requirement that is easy to miss: `tailscaleDeviceReaper.tags` is load-bearing, not belt-and-braces. The devices listing returns the whole tailnet however the credential is scoped, so without the tag predicate each env's reaper would walk every other env's devices and die on the first 403 from a write it was never authorised to make. The client scope is the second, independent guard, keeping a bug in the predicate from deleting another env's fleet.

**Two grace windows.** A single flat window cannot serve both classes of device:

- 12h for devices matching `disposable.hostnamePattern` (the processor VMs). Their identity regenerates on every boot, so a dead registration can never be reused. 12h is roughly 24x the Deployment's 1800s `progressDeadlineSeconds`, so a slow rollout cannot have its own in-flight VM reaped.
- 7 days for everything else, sized for the Mac mini hosts. Those join once at bootstrap and keep that identity, so reaping one while its host is merely down would strand it off the tailnet until re-bootstrapped.

At a flat 7 days the release debris alone (10 to 15 a week) would sit on top of the 44 live devices and keep the tailnet over its allowance permanently. The split gets steady state to roughly 46.

`tag:tuist-mgmt` is outside every reaper's scope deliberately: one long-lived device that has never produced a duplicate, and the one deletion that could cost the access needed to undo it.

## Safety

- Two independent scopes must agree. The predicate selects only devices whose tags intersect this env's `tags`, and Tailscale itself refuses a write from a client scoped elsewhere. User devices carry no tags and cannot pass either.
- An absent or vacuous tag list is a hard error, not an empty match. A reaper that sweeps nothing is indistinguishable from a healthy one in every signal available (the Job succeeds, the log reads 0 devices), so the misconfiguration would survive until someone questioned the bill. Validation runs on the parsed list, so a stray comma or surrounding whitespace is caught too.
- `lastSeen` older than the window is the liveness test; a live device reports within minutes. A device missing `lastSeen` is skipped, never assumed old.
- Every candidate is re-fetched and re-validated immediately before deletion, so one that reconnected mid-run is skipped without consuming a cap slot.
- `maxDeletions` caps a run at 25, oldest first, remainder rolling into the next run.
- Deletion is recoverable by construction: our joiners re-authenticate from the OAuth client on next boot.
- The run log reports in-scope and tailnet-wide counts separately, so `0 of 0 in-scope (76 on the tailnet)` reads as a mistyped tag rather than a clean tailnet.

## Impact

Nothing changes for self-hosters: `enabled` defaults false and the template renders nothing when off.

All three managed envs ship with `dryRun: true`, so merging this deletes nothing. Arming is a per-env value flip after reading a run's logs. Once armed the reapers clear 21 of the 23 currently-stale devices on their own schedule (the other two are still inside their windows), taking the tailnet from 67 to roughly 46 and stopping the unbounded growth.

The three OAuth clients and 1Password items this depends on already exist, verified as present in each env vault with the `client-id` / `client-secret` labels ESO resolves by, `tskey-client-` prefixes (OAuth clients, not 90-day auth keys) and three distinct client IDs. Provisioning steps for future envs are in section 5d of `infra/k8s/onboarding.md`.

## How to test locally

Render each env overlay and confirm the tag scope reaches the container:

```bash
helm template tuist infra/helm/tuist -n tuist \
  -f infra/helm/tuist/values-managed-canary.yaml \
  --set server.image.tag=t --set runnersFleet.runnerImageSemver=1.0.0 \
  --set runnersFleetLinux.shapeRunnerImage=i:1 --set registry.image.tag=t \
  --set kuraController.image.tag=t --set cache.image.tag=t \
  --set codebaseSearch.image.tag=t --set ociRegistry.image.tag=t \
  --set xcresultProcessor.image.tag=t --set slack.image.tag=t \
  --show-only templates/tailscale-device-reaper.yaml
```

Confirm the reaper renders nothing when disabled (the self-hoster default):

```bash
helm template tuist infra/helm/tuist -n tuist --set server.image.tag=t \
  --set server.license.key=t --show-only templates/tailscale-device-reaper.yaml
```

The behaviour was verified by extracting `reap.sh` from the rendered ConfigMap and running it against a stub `curl` on PATH serving a fixture built from the live tailnet plus edge cases (untagged-and-ancient, `tag:tuist-mgmt`-and-ancient, never-seen, empty `lastSeen`, one hour inside the window, one hour outside), with the stub recording DELETEs and able to simulate a device reconnecting between scan and delete.

Confirmed across the three env scopes:

- Each env selects only its own devices (production 18, canary 1, staging 2), no env emits a device tagged for another, and the three sets sum to exactly the 21 a single all-tag reaper selected.
- Devices that are online, untagged, `tag:tuist-mgmt`, missing `lastSeen`, carrying an empty `lastSeen`, or one hour inside their window are selected by nobody.
- Empty, unset, comma-only and whitespace-only tag lists all abort non-zero with a clear message.
- A nonexistent tag reports `0 of 0 in-scope` against a 76-device tailnet.
- Fractional-second `lastSeen` values parse (the live API returns some).
- `maxDeletions` stops at its limit and reports the remainder; a mid-run reconnect is skipped without consuming a slot; an API failure exits non-zero.

`helm lint` passes.

### DIFF
--- a/infra/helm/tuist/templates/tailscale-device-reaper.yaml
+++ b/infra/helm/tuist/templates/tailscale-device-reaper.yaml
@@ -66,7 +66,13 @@ Prometheus Operator (they scrape via Alloy into Grafana Cloud). Alert on the
 reaper's log lines in Grafana Cloud instead.
 */ -}}
 {{- $name := include "tuist.componentName" (dict "root" . "component" "tailscale-device-reaper") }}
-{{- if .Values.tailscaleDeviceReaper.externalSecrets.enabled }}
+{{- /*
+Also gated on oauthSecretName being empty: pointing the CronJob at a
+hand-managed Secret is how you opt out of the chart's 1Password sync, so
+rendering the ExternalSecret anyway would keep syncing an item nothing reads and
+fail the sync outright when that item does not exist in the vault.
+*/ -}}
+{{- if and .Values.tailscaleDeviceReaper.externalSecrets.enabled (not .Values.tailscaleDeviceReaper.oauthSecretName) }}
 apiVersion: external-secrets.io/v1
 kind: ExternalSecret
 metadata:
@@ -190,13 +196,33 @@ data:
     # An empty DISPOSABLE_PATTERN disables the short window entirely rather
     # than matching every device, which is what an unguarded test("") would do.
     #
-    # lastSeen is RFC3339; fromdateiso8601 rejects fractional seconds, so strip
-    # them rather than letting one oddly-formatted device abort the whole run.
+    # lastSeen is RFC3339. Parsing it is the fiddliest part of the predicate and
+    # every failure mode here defaults to skipping the device, never to reaping
+    # it early:
+    #
+    #   * fromdateiso8601 rejects fractional seconds, which the API does emit
+    #     (".1Z" appears on live devices), so strip them first.
+    #   * Anything else it cannot parse -- a numeric offset instead of Z, a
+    #     shape we have not seen -- is caught by `try` and skipped. Without the
+    #     catch a single odd timestamp aborts the entire sweep, which is the
+    #     failure the strip above exists to prevent in the first place.
+    #   * Go's zero time marshals to "0001-01-01T00:00:00Z", which parses fine
+    #     but to a large negative epoch, so it is older than every grace window
+    #     and sorts ahead of genuinely stale devices. `tailscale status --json`
+    #     emits exactly that for peers with no recorded lastSeen, so treat any
+    #     non-positive epoch as "no timestamp" and skip it, the same as an
+    #     absent field. Never assume old.
+    #
+    # Binding the parsed value once also keeps the sort key below honest: it
+    # reuses $seen rather than re-deriving it with a second, separately-escaped
+    # copy of the same expression.
     predicate='
       select(any((.tags // [])[]; IN($allowed[])))
       | select(.lastSeen != null and .lastSeen != "")
+      | (.lastSeen | sub("\\.[0-9]+Z$"; "Z") | try fromdateiso8601 catch null) as $seen
+      | select($seen != null and $seen > 0)
       | select(
-          ((.lastSeen | sub("\\.[0-9]+Z$"; "Z")) | fromdateiso8601)
+          $seen
           < (now - (if ($pattern != "" and ((.hostname // "") | test($pattern)))
                     then $disposable else $grace end))
         )'
@@ -213,7 +239,7 @@ data:
     # Oldest first, so when the cap bites it reaps the deadest devices and the
     # next run picks up where this one stopped.
     candidates=$(printf '%s' "$devices" | jq -r "${jq_args[@]}" \
-      ".devices[] | ${predicate} | [(.lastSeen | sub(\"\\\\.[0-9]+Z\$\"; \"Z\") | fromdateiso8601), .id, .hostname] | @tsv" \
+      ".devices[] | ${predicate} | [\$seen, .id, .hostname] | @tsv" \
       | sort -n | cut -f2-)
     count=$(printf '%s\n' "$candidates" | grep -c . || true)
     echo "tailscale-device-reaper: ${count} of ${in_scope} in-scope device(s) past their grace window (${total} on the tailnet)"
@@ -222,6 +248,7 @@ data:
     fi
 
     reaped=0
+    failed=0
     while IFS=$'\t' read -r id hostname; do
       [ -z "$id" ] && continue
       if [ "$reaped" -ge "$MAX_DELETIONS" ]; then
@@ -244,16 +271,37 @@ data:
       fi
       last_seen=$(printf '%s' "$current" | jq -r '.lastSeen')
       tags=$(printf '%s' "$current" | jq -r '(.tags // []) | join(",")')
-      if [ "$DRY_RUN" = "true" ]; then
+      # Anything other than the exact string "false" stays in dry-run. Arming is
+      # a hand-edit of a per-env values file, which is where someone writes
+      # `dryRun: "yes"` -- bare True/yes/on coerce to a boolean and render
+      # "true", but the quoted forms reach the container verbatim. Comparing
+      # against "false" means an unrecognised value fails toward not deleting,
+      # which is the direction every other guard here points.
+      if [ "$DRY_RUN" != "false" ]; then
         echo "tailscale-device-reaper: [dry-run] would reap ${hostname} (${id}, ${tags}, last seen ${last_seen})"
       else
         echo "tailscale-device-reaper: reaping ${hostname} (${id}, ${tags}, last seen ${last_seen})"
-        curl -sS -f -X DELETE -H "Authorization: Bearer ${token}" "${api}/device/${id}" >/dev/null
+        # Non-fatal per device. Under `set -e` an unguarded failure here would
+        # abort the run, and because candidates are ordered oldest-first, one
+        # undeletable device would sit at the head of the queue and block every
+        # sweep behind it forever -- a 403 from a tag scope that has drifted out
+        # of agreement with `tags` does exactly that, and so does a transient
+        # 5xx. Record it, move on, and fail the Job at the end so the run is
+        # still loud without being useless.
+        if ! curl -sS -f -X DELETE -H "Authorization: Bearer ${token}" "${api}/device/${id}" >/dev/null; then
+          echo "tailscale-device-reaper: FAILED to delete ${hostname} (${id}, ${tags}); continuing" >&2
+          failed=$(( failed + 1 ))
+          continue
+        fi
       fi
       reaped=$(( reaped + 1 ))
     done <<< "$candidates"
 
-    echo "tailscale-device-reaper: done (${reaped} device(s) $([ "$DRY_RUN" = "true" ] && echo 'would be reaped' || echo reaped))"
+    echo "tailscale-device-reaper: done (${reaped} device(s) $([ "$DRY_RUN" != "false" ] && echo 'would be reaped' || echo reaped), ${failed} failed)"
+    if [ "$failed" -gt 0 ]; then
+      echo "tailscale-device-reaper: ${failed} device(s) could not be deleted; check the tag scope on this env's OAuth client" >&2
+      exit 1
+    fi
 ---
 apiVersion: batch/v1
 kind: CronJob
--- a/infra/helm/tuist/templates/tailscale-device-reaper.yaml
+++ b/infra/helm/tuist/templates/tailscale-device-reaper.yaml
@@ -30,11 +30,23 @@ and not the next one. Registrations are an ungarbage-collected resource on a
 shared tailnet, so the sweep is the general answer and the hostname fix is a
 worthwhile narrowing on top of it.
 
+One reaper per environment, each sweeping only its own env's devices. All envs
+share a single tailnet, so `tags` is what keeps three reapers from colliding on
+one device list: a run only ever considers devices carrying one of the tags
+named here, and the OAuth client backing it is scoped to those same tags. Both
+halves matter. Without the predicate a reaper would list devices it has no
+authority over and fail on the first 403; without the client scope the predicate
+would be the only thing standing between a bug and another env's fleet.
+
 Why this is safe:
-  * The OAuth client is tag-scoped (see the ExternalSecret below), so Tailscale
-    itself refuses any write against a device outside our infrastructure tags.
-    User devices are unreachable to this credential no matter what the script
-    does. The `.tags` predicate below is the second, independent check.
+  * Two independent scopes have to agree before anything is deleted. The
+    predicate only selects devices whose tags intersect this env's `tags`, and
+    Tailscale itself refuses a write from a client scoped elsewhere. Neither
+    user devices (which carry no tags at all) nor another env's devices can pass
+    both.
+  * An empty `tags` list is a hard error rather than a silent no-op. A reaper
+    that quietly sweeps nothing is worse than one that fails: the bill keeps
+    growing while the CronJob reports success.
   * `lastSeen` older than graceHours is the liveness test. A device that is up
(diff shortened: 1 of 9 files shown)


pepicrft (approved):

infra/helm/tuist/templates/tailscale-device-reaper.yaml
```diff
@@ -0,0 +1,330 @@
...
+        echo "tailscale-device-reaper: skip ${hostname} (${id}, reconnected since scan)"
+        continue
+      fi
+      last_seen=$(printf '%s' "$current" | jq -r '.lastSeen')
+      tags=$(printf '%s' "$current" | jq -r '(.tags // []) | join(",")')
+      if [ "$DRY_RUN" = "true" ]; then
+        echo "tailscale-device-reaper: [dry-run] would reap ${hostname} (${id}, ${tags}, last seen ${last_seen})"
+      else
+        echo "tailscale-device-reaper: reaping ${hostname} (${id}, ${tags}, last seen ${last_seen})"
+        curl -sS -f -X DELETE -H "Authorization: Bearer ${token}" "${api}/device/${id}" >/dev/null
```
esnunes: The `DELETE` on line 251 is the only unguarded call in the loop — the per-device GET (234) and the re-validation jq (239) both end in `|| true`, so a failure there degrades into a skip, but a non-2xx here ends the run under `set -e`. Combined with the oldest-first ordering, that means a device that can't be deleted sits at the head of the queue and blocks every sweep behind it, and the closing `done (N reaped)` line never prints either.

I reproduced it: extracted the rendered `reap.sh`, ran it in `alpine/k8s:1.31.1` against a stubbed API with 31 candidates, made the second return 403 on DELETE — exit 22 after 2 attempts, the other 29 never touched. A transient 5xx from the API costs the same run.

§5d of the onboarding doc already warns that `tags` and the client's tag scope have to be kept in agreement, which is precisely the drift that produces a 403 on write. Would it be worth making the delete non-fatal per device — log it, set a flag, `continue` — and `exit 1` at the end if anything failed? The run still fails loudly, but it also makes progress and reports what it managed.

infra/helm/tuist/templates/tailscale-device-reaper.yaml
```diff
@@ -0,0 +1,330 @@
...
+    #     re-bootstrapped. The long window is that recovery margin.
+    #
+    # An empty DISPOSABLE_PATTERN disables the short window entirely rather
+    # than matching every device, which is what an unguarded test("") would do.
+    #
+    # lastSeen is RFC3339; fromdateiso8601 rejects fractional seconds, so strip
+    # them rather than letting one oddly-formatted device abort the whole run.
+    predicate='
+      select(any((.tags // [])[]; IN($allowed[])))
+      | select(.lastSeen != null and .lastSeen != "")
```
esnunes, line 221: Suggestion rather than a defect — I'm not confident this shape actually comes back from the API in practice, but it looks like a cheap guardrail either way.

`select(.lastSeen != null and .lastSeen != "")` rejects an absent field and an empty string, but not `0001-01-01T00:00:00Z`, which is what Go's zero `time.Time` marshals to. I ran the predicate under jq 1.7.1 inside `alpine/k8s:1.31.1`: such a device is selected, `fromdateiso8601` maps it to `-62135596800` so it's past *every* grace window, and `sort -n` puts it first — ahead of the genuinely stale devices, at the head of the maxDeletions budget. If it can ever occur, that device gets reaped on the first armed run with no grace at all, which is the case the invariant on line 68 says is covered. `select(.lastSeen > "0001-01-02")` before the parse would close it.

Possibly worth hardening the parse while you're in there: only `.NNNZ` fractional seconds are normalised, and the candidate jq on line 216 has no `|| true`, so any other timestamp jq can't parse aborts the whole sweep rather than skipping one device — the failure the line-193 comment says the `sub` exists to prevent.

infra/helm/tuist/templates/tailscale-device-reaper.yaml
```diff
@@ -0,0 +1,330 @@
...
+      fi
+      still=$(printf '%s' "$current" | jq -r "${jq_args[@]}" \
+        "${predicate} | .id" 2>/dev/null || true)
+      if [ "$still" != "$id" ]; then
+        echo "tailscale-device-reaper: skip ${hostname} (${id}, reconnected since scan)"
+        continue
+      fi
+      last_seen=$(printf '%s' "$current" | jq -r '.lastSeen')
+      tags=$(printf '%s' "$current" | jq -r '(.tags // []) | join(",")')
+      if [ "$DRY_RUN" = "true" ]; then
```
esnunes: `if [ "$DRY_RUN" = "true" ]` means only the exact string `true` is dry-run — every other value arms live deletion. The shipped state is safe (values.yaml:59 defaults true, no env file overrides it, and all three render `DRY_RUN=true`), and released-pv-reaper.yaml:132 uses the same comparison, so this is existing convention rather than something you introduced.

The thing that nags is that arming is deliberately a per-env values edit, which is exactly the moment someone writes `dryRun: True` or `dryRun: yes` — Helm's parser keeps both as strings, `| quote` renders them verbatim, and the check falls through to live deletion while the values file reads as disarmed. Inverting to `if [ "$DRY_RUN" != "false" ]` would make the default direction match what the rest of the design does, which is fail toward not deleting.

infra/helm/tuist/values.yaml
```diff
@@ -34,6 +34,88 @@ pvReaper:
...
+  # Ceiling on a single run, so a wrong predicate cannot empty the tailnet
+  # before anyone reads the logs. Candidates are reaped oldest-first and the
+  # remainder rolls into the next run.
+  maxDeletions: 25
+  # curl + jq in one image; same image the PV reaper uses.
+  image: alpine/k8s:1.31.1
+  # Secret holding the tag-scoped OAuth client. Empty falls back to the
+  # ExternalSecret this chart syncs from 1Password; set it to opt out and
+  # manage the Secret by hand.
+  oauthSecretName: ""
```
esnunes, line 100: Tiny mismatch between the comment and the template. The comment says "Empty falls back to the ExternalSecret this chart syncs from 1Password; set it to opt out and manage the Secret by hand" — but the ExternalSecret is gated only on `externalSecrets.enabled` (default true), so setting `oauthSecretName` on its own still syncs a `TAILSCALE_DEVICE_REAPER` item into a Secret nothing reads, and fails if that item isn't in the vault.

Either gate the ExternalSecret on `and .externalSecrets.enabled (not .oauthSecretName)`, or reword the comment to say both values have to change. No strong preference.

esnunes (approved):
This is a really well-argued change, and the reasoning that survives in the template header is the best part of it — the "renaming won't help, identity is the node key" analysis in particular saves the next person from re-deriving it. The layered safety story (an independent tag predicate plus a tag-scoped credential, re-validation immediately before the destructive call, a per-run cap, oldest-first ordering, dry-run as the shipped default) is the right shape for something that deletes infrastructure on a schedule, and the two-tier grace window is the detail that makes it actually work rather than merely be safe.

I verified more than I read. Rendered the chart for all three envs — correct two tags each, dry-run on everywhere — then extracted the rendered `reap.sh` and ran it inside `alpine/k8s:1.31.1` under the same read-only-rootfs / UID 65534 constraints, against a stubbed Tailscale API. The predicate behaved correctly on every fixture: untagged user devices, another env's tag, absent and empty `lastSeen`, fractional seconds, and a device an hour inside its window were all excluded; the loop honoured the cap (25 of 31, oldest first, remainder reported) and skipped a live device. No injection path into the shell loop or the jq program — `@tsv` escapes hostnames, everything reaches jq through `--arg`/`--argjson`, and the here-string (rather than a pipe) correctly keeps `reaped` in the parent shell so `maxDeletions` actually binds.

Approving — nothing here blocks. One thing I'd genuinely like your read on, left inline on the delete: it's the only unguarded call in the loop, so a single non-2xx ends the run under `set -e`, and because candidates are ordered oldest-first, an undeletable device parks at the head of the queue and blocks every sweep behind it. I reproduced that with a stubbed 403 — exit 22 after two attempts, 29 candidates untouched. That's the shape of the `tags`-wider-than-client-scope drift §5d warns about, so it seems worth making the delete non-fatal per device and failing the run at the end instead.

The other three inline notes are smaller: a suggested guardrail on the zero-value timestamp in the `lastSeen` check, `dryRun` failing open on any string but `true` (pre-existing convention, but arming is exactly when someone writes `True`), and a values comment that promises an `oauthSecretName` opt-out the template doesn't implement. Take or leave any of them.

infra/helm/tuist/templates/tailscale-device-reaper.yaml
fortmarek: Agreed, fixed. The DELETE is now guarded: a failure logs `FAILED to delete <host> (<id>, <tags>); continuing`, increments a counter, and `continue`s without consuming a maxDeletions slot, and the run `exit 1`s at the end if the counter is non-zero — loud, but after making progress. Verified with a 403 injected on the second of 19 candidates: the other 18 are deleted, the failure is reported, the `done (18 reaped, 1 failed)` line prints, exit 1.

infra/helm/tuist/templates/tailscale-device-reaper.yaml
fortmarek: Good catch, and it turned out worse than suspected — the shape does occur, and the failure is a total abort rather than a mis-selection. `tailscale status --json` emits `0001-01-01T00:00:00Z` for peers with no recorded lastSeen (I had been filtering it out of my own fixture without registering why). Reproduced all three shapes against the old script under jq 1.8.1: zero time dies on `invalid gmtime representation`, and a numeric offset and a garbage string die on `does not match format` — each killing the sweep before any candidate was considered. Your jq 1.7.1 result (selected and sorted first) is the other half of the same bug; the fix covers both.

`.lastSeen` is now parsed once through `try fromdateiso8601 catch null` and bound, with `select($seen != null and $seen > 0)` rejecting both unparseable values and any non-positive epoch, so a degenerate timestamp is skipped exactly like an absent one rather than being treated as infinitely old. Binding it also let the sort key reuse `$seen` instead of carrying a separately-escaped second copy of the expression. Verified: the three bad shapes are skipped and the sweep still reaps the genuinely stale device alongside them.

infra/helm/tuist/templates/tailscale-device-reaper.yaml
fortmarek: Inverted to `!= "false"`, so an unrecognised value now fails toward not deleting. One correction to the mechanism, since I checked before changing it: bare `True`, `yes` and `on` are coerced by Helm and render `"true"`, so those were already safe. The hole is the quoted forms — `dryRun: "yes"` and `dryRun: "True"` stay strings, render verbatim, and previously armed live deletion while the values file read as disarmed. Narrower path than described, same conclusion. Verified only `false` arms across true/True/yes/TRUE/0/empty/unrecognised.

infra/helm/tuist/values.yaml
fortmarek: Fixed by changing the template rather than the comment, since the comment described the behaviour people would expect. The ExternalSecret is now gated on `and .externalSecrets.enabled (not .oauthSecretName)`, so naming your own Secret opts out of the 1Password sync entirely. Verified: renders by default, disappears when `oauthSecretName` is set, and the CronJob reads the named Secret. Reworded the values comment to say so explicitly.

PR #12416, opened 2026-08-17
perf(kura, server): remove the vendored GeoIP database and its per-request lookup

fortmarek:
Prerequisite for [Hive spec 78, "Regional Kura capacity and demand-driven lifecycle"](https://hive.tuist.dev/specs/78), which cannot ship its standard compute profile until this lands.

## Why

Measured on 2026-08-17 across all 80 live Kura pods: 30-day peak anonymous memory is 145-150 MiB for 66 of the 80. That floor is almost entirely the DB-IP City Lite MMDB, which decompresses to roughly 125 MiB and is heap-loaded via `maxminddb::Reader<Vec<u8>>` in every pod. At the ~303 pods the spec's full rollout needs, that is ~37 GiB of fleet RAM, close to a whole 64 GB machine.

The spec's standard profile is a 256 MiB memory request. With GeoIP loaded, `Config::anon_admission_budget_bytes` leaves only ~109 MiB of in-flight admission budget after the RocksDB block cache, write-buffer pool, manifest cache, and the 96 MiB `PROCESS_ANON_BASELINE_BYTES`. Removing GeoIP roughly doubles usable concurrency at the same request.

The refresh task was the sharper problem: it re-downloaded the database daily per pod and held the compressed Vec, the decompressed Vec, and the old reader simultaneously, roughly 310 MiB on an uncoordinated timer. That is fatal at any small memory limit.

## What changed

There is deliberately **no replacement component**. No ingress, sidecar, or cluster service takes this over, and client-IP geolocation is not reintroduced anywhere.

**Per-request client attribution: deleted outright.** Both call sites were observability only, with nothing routing, authorizing, metering, or billing on the value. Gone: the lookup in `kura/src/http.rs`, the `geo.*` span attributes on `http.request`, and `client_ip_from_headers` (whose only caller was the lookup).

`kura_http_client_requests_total` is removed rather than kept as a label-less counter. Its help text was literally "Public HTTP requests by client country"; without `client_country` it only restates a sum over `kura_http_requests_total`. No dashboard in `infra/grafana-dashboards/` and no alert rule referenced either metric, so nothing downstream needed updating.

**Node self-location: now explicit deployment configuration.** `resolve_node_location` loses its GeoIP argument and the egress-IP probe that only fed it, so it is a synchronous pure function of configuration that cannot reach the network. `KURA_NODE_COUNTRY` / `KURA_NODE_SUBDIVISION` already existed but nothing set them, so this plumbs them through.

The values live on the region catalog rather than the fleet values in `infra/helm`, because `server/lib/tuist/kura/regions.ex` already declares the node pool each region pins to. That makes region to fleet to datacenter a single hop, and one entry covers every environment. `Regions.node_location/1` feeds the provisioner, which renders both variables into the KuraInstance `extraEnv`. `@manifest_revision` moves with it, or the reconciler (which converges on the revision alone) would never re-apply the new env.

Each location was confirmed against the fleet definitions rather than inferred from the region name:

| Region | Fleet / datacenter | Country | Subdivision |
| --- | --- | --- | --- |
| `us-east` | OVH Vint Hill, Virginia | US | US-VA |
| `us-west` | OVH Hillsboro, Oregon | US | US-OR |
| `eu-central` | Dedibox DC5 (production, staging), DC2 (canary) | FR | FR-IDF |
| `ca-east` | OVHcloud BHS, Beauharnois, Quebec | CA | CA-QC |
| `scw-fr-par-runners` | Scaleway Elastic Metal fr-par-1 | FR | FR-IDF |

**The region-name fallback was wrong, and is now narrower.** Verifying it (as the spec asks) turned up a live bug: `eu-central` mapped to `DE`, but that region has been served from Scaleway Dedibox in France since the cutover. The hardcoded table is gone, leaving only the ISO-prefix rule. `us-east` and `us-west` still resolve to `US` and `fr-par` to `FR`, while continent prefixes such as `eu-` now resolve to nothing instead of guessing. A region id names a Tuist region, not a geography, so the fallback should stay silent rather than answer wrongly when a region moves datacenter. Every managed region sets both values explicitly, so this only affects an unconfigured node.

**Also removed:** `kura/src/geoip.rs` and its `mod` declaration, the `AppState.geoip` field, `spawn_geoip_refresh_task`, `KURA_GEOIP_REFRESH_INTERVAL_SECS`, the `record_geoip_refresh` metric, the `flate2` and `maxminddb` dependencies, which leave the crate graph entirely, plus the direct `time` declaration, which does not (`rcgen` and `sentry-types` still pull it in) but was only ever used by `geoip.rs`, and the MMDB stage in both Dockerfiles, which also drops the build-time dependency on download.db-ip.com. The self-hosted chart drops `config.geoip`; self-hosters set `KURA_NODE_COUNTRY` / `KURA_NODE_SUBDIVISION` through `extraEnv`, since no dashboard or alert queries the geography and the chart already supports that surface for optional settings.

## Impact

Self-hosters lose client-country attribution on their Kura metrics and traces, and `KURA_GEOIP_REFRESH_INTERVAL_SECS` becomes inert. Serving-node geography is unchanged in shape: the same `geo.country.iso_code` / `geo.region.iso_code` Resource attributes and the `kura_node_geo_info` metric are still exported, now sourced from configuration. A self-hosted node that sets neither variable and whose region name carries no country prefix exports spans without geography rather than guessing one.

### How to test locally

Run from `kura/`:

```bash
mise exec -- cargo test --lib && mise exec -- cargo clippy --all-targets -- -D warnings
```

Run from `server/`:

```bash
mix test test/tuist/kura/regions_test.exs test/tuist/kura/provisioner/kubernetes_controller_test.exs
```

Confirm the chart emits the location variables only when configured:

```bash
helm template kura kura/ops/helm/kura \
  --set 'extraEnv[0].name=KURA_NODE_COUNTRY' --set 'extraEnv[0].value=FR' \
  --set 'extraEnv[1].name=KURA_NODE_SUBDIVISION' --set 'extraEnv[1].value=FR-IDF' | grep KURA_NODE_
```

## Validation run

- 714 Rust unit tests pass, clippy clean, `cargo fmt --check` clean. One full-suite run had `backfill::pass::tests::flush_triggers_on_time_while_listing_is_still_running` fail on a timing assertion; it passed 3/3 in isolation and on the next full run, and this change does not touch backfill.
- 110 server tests pass, including new coverage that a configured region renders both variables, that an unconfigured one renders neither, that no provisionable region is left without a location, and that a fully configured node resolves without any network call.
- `helm lint` passes; the chart renders both variables when set and neither by default.

## Still to run

Docker was unavailable in the authoring environment, so two checks from the spec remain:

1. Verify the released image no longer contains `/opt/geoip/dbip-city-lite.mmdb` and is correspondingly smaller. The Dockerfiles are clean by inspection, but the image itself has not been built and inspected.
2. After deploying to staging, compare `kura_container_memory_anon_bytes` before and after on the same pod. Expectation is the ~147 MiB floor drops to roughly 25-50 MiB.

That second number feeds directly into whether `PROCESS_ANON_BASELINE_BYTES` (currently 96 MiB) should also come down, which would widen the admission budget further at the same memory request. Per the spec, this PR does **not** change the memory request or limit defaults: the standard-profile canary validates the limit under representative traffic separately.


pepicrft (approved):

kura/Cargo.toml
```diff
@@ -23,7 +22,6 @@ hyper = "1.7.0"
 hyper-util = "0.1.20"
 libc = "0.2.177"
 jsonwebtoken = { version = "10.3.0", features = ["rust_crypto"] }
-maxminddb = "0.27.0"
 memmap2 = "0.9.10"
```
esnunes, line 25: Small leftover: `time` (`kura/Cargo.toml:42`) came in with the original geo-attribution change, and the deleted `kura/src/geoip.rs` was its only consumer — `use time::{Month, OffsetDateTime}`, for the `YYYY-MM` segment of the DB-IP URL. After this deletion there's no reference to the crate left in `kura/src`; every remaining `time::` resolves to `std::time` or `tokio::time`.

Cargo won't flag it, so it compiles clean either way, but since rules_rs resolves the Bazel crate graph from `Cargo.toml`/`Cargo.lock` it still gets fetched and built. Might be worth dropping alongside `flate2` and `maxminddb` so the "leaves the crate graph entirely" claim covers all three.

kura/ops/helm/kura/values.yaml
```diff
@@ -192,17 +192,14 @@ config:
...
-  # controls how often the pod re-downloads the monthly dump in the
-  # background. Set to 0 to pin to the vendored copy for the pod
-  # lifetime.
-  geoip:
-    refreshIntervalSecs: 86400
-
-# Use extraEnv or extraEnvFrom to inject optional secrets such as KURA_SENTRY_DSN.
+
+# Use extraEnv or extraEnvFrom to inject optional secrets such as KURA_SENTRY_DSN,
+# or optional settings such as KURA_NODE_COUNTRY / KURA_NODE_SUBDIVISION, the ISO
```
esnunes, line 197: Heads-up on the description rather than the code. The "What changed" section still says the self-hosted chart trades `config.geoip` for `config.node.country` / `config.node.subdivision`, and "How to test locally" suggests:

    helm template kura kura/ops/helm/kura --set config.node.country=FR --set config.node.subdivision=FR-IDF | grep KURA_NODE_

The second commit removed that knob in favour of `extraEnv`, so that command now matches nothing and reads like a broken chart to anyone who runs it.

The `extraEnv` route is the better design — dropping a dedicated surface for a value nothing queries, and avoiding the nil-pointer template error a wholesale `config:` override would hit on `.Values.config.node.country` — so this is just the body needing to follow the second commit, e.g. `--set extraEnv[0].name=KURA_NODE_COUNTRY,extraEnv[0].value=FR`. The docs table in `self-hosting.md` is already correct.

server/lib/tuist/kura/regions.ex
```diff
@@ -370,6 +387,22 @@ defmodule Tuist.Kura.Regions do
...
+  @doc """
+  Where the region's nodes physically are, as `%{country:, subdivision:}` ISO
+  3166-1 alpha-2 / 3166-2 codes, or `nil` when the region declares no location.
+
+  This is deployment configuration derived from the datacenter each fleet was
+  ordered in, not something a node discovers about itself: Kura stamps the pair
+  onto its OTel Resource verbatim. A region that moves datacenter must be
+  updated here, because nothing downstream can detect the move.
+  """
+  def node_location(%__MODULE__{provisioner_config: %{country: country, subdivision: subdivision}})
```
esnunes, line 399: `node_location/1`'s head requires both `:country` and `:subdivision` to be present in `provisioner_config`, and the `is_binary(country) or is_binary(subdivision)` guard then decides. That's correct for every real region, since `managed_region/1` and `private_region/1` both write the keys unconditionally via `Map.get/2`.

What the two new negative tests exercise, though, is the other shape: `eu_region()` in the controller test and `local-controller` in the regions test both omit the keys entirely, so they land on the catch-all clause and the guard never runs. The path a real unlocated region takes — both keys present and nil, which is what the `hetzner-staging-runners` tombstone does today and what any future region added without a location would do — isn't covered.

One extra assertion would close it, if you think it's worth it:

    assert Regions.node_location(Regions.get("hetzner-staging-runners")) == nil

esnunes (approved):
This is a satisfying deletion to read. Removing a feature is usually messier than this — I went looking for orphans and found none: no reference to the database, the removed metrics, the `client_country` label, the env var, or the chart value survives anywhere in the tree, and the one metric you keep, `kura_node_geo_info`, is only consumed by panels that join on its `region` label.

Two things I particularly liked. Verifying the region-name fallback instead of trusting it is what surfaced the stale `eu-central` → `DE` mapping, and narrowing the fallback to "answer only when the region name is genuinely a country code" is the right response — a region id names a Tuist region, not a geography. And the second commit walking back the dedicated `config.node` chart knob in favour of `extraEnv` is a better landing than the first: it drops a first-class surface for a value nothing queries, and it sidesteps the nil-pointer template error a wholesale `config:` override would hit.

The inline notes are all small — an orphaned `time` dependency, a description that didn't follow the second commit, and a test that covers a slightly different shape than the one production takes. None of them block.

One thing I'd flag rather than ask you to change: the trade this makes is real. Location can no longer be wrong-but-self-correcting; it can only be wrong-and-silent. The "leaves no provisionable region without one" test guards that a location exists, not that it's still true after a fleet moves. The docstring says so plainly, which I think is the right way to carry that cost.

kura/Cargo.toml
fortmarek: Dropped in 4df2818bc1, `geoip.rs` was indeed the only consumer. One correction on the rationale: it does not leave the crate graph, since `rcgen` and `sentry-types` both depend on it, so Bazel fetches and builds it either way. I kept the body's "leaves the crate graph entirely" claim scoped to `flate2` / `maxminddb` and described `time` as just an unused direct declaration.

kura/ops/helm/kura/values.yaml
fortmarek: Body updated: the chart sentence now describes the `extraEnv` route, and the test command uses `--set 'extraEnv[0].name=KURA_NODE_COUNTRY'`. Thanks for catching the test command, I had only queued the prose fix.

server/lib/tuist/kura/regions.ex
fortmarek: Added `assert Regions.node_location(Regions.get("hetzner-staging-runners")) == nil` in 4df2818bc1, with a note that the both-keys-nil shape is the path any future region added without a location takes.

Issue #72, opened 2018-07-22
Set up build phases for extensions


pepibumur:
## Context 🕵️‍♀️
Products of type extension are copied into the app that they are extending through a build phase. The current generation logic doesn't take that into account and thus, defining extensions in xpm is not fully supported. 

## What 🌱
Ensure that all kinds of extensions are supported. That includes, watch, iMessage, stickers and tv extension.

- Add the following linting:
  - Watch extension has the right bundle id.
  - Watch extensions have a dependency with the app that they are extending. The app exists and the platform its platform is compatible with the extension's.

pepibumur:
![](https://github.trello.services/images/mini-trello-icon.png) [Support for extensions](https://trello.com/c/L57rnoEn/6-support-for-extensions)

PR #12417, opened 2026-08-17
feat(server): add demand-driven lifecycle for Kura account-region instances

fortmarek:
Implements the "Demand-driven lifecycle" section of [Hive spec 78](https://hive.tuist.dev/specs/78). No GitHub issue; the spec is the reference.

**Validated end to end on staging** against real infrastructure: provisioned, archived after inactivity, and reprovisioned on renewed activity. Details at the bottom.

## Why

Measured against production: 245 accounts had Air-plan cache demand in the last 90 days but only 169 in the last 30. Paid is flat by comparison (58 at 90 days, 55 at 30). At the spec's 24 GiB Air filesystem quota and 1,105 GiB usable per machine, that Air population is the difference between 6 machines and 4.

Without a lifecycle the fleet grows with signups rather than usage: every account that ever ran one build keeps a quota-enforced directory forever. Production has 1,320 organizations, 1,080 of them created in the last 12 months, and only 270 have run a cacheable build in the last year.

## What changed

Five states on one account-region instance, since a single Kura instance serves all of that account's projects in a region:

```
archived -> provisioning -> active -> drain_pending -> archived
                              ^            |
                              +------------+
```

New modules under `server/lib/tuist/kura/`: `demand.ex` records cache demand and reads it back, `lifecycle.ex` is the identity rule that converges instances with demand, `capacity.ex` is the shared model behind admission control and the Air pressure rule, `account_region_lifecycle.ex` plus its migration holds the demand and archival bookkeeping, `telemetry.ex` and `prom_ex_plugin.ex` carry the metrics, and two workers seed demand and run the inactivity sweep.

### What archival actually costs

Kura is terminal storage for what it holds. Its store is RocksDB plus local append-only segment and blob files replicated between peers, with no object-storage backing and a 404 rather than an origin fetch on a miss. **Archiving an account-region discards that account's cache content in that region**, and the only other copy is a peer instance in another region if the account has one.

What makes that safe is not a second copy, it is that a cache miss is always safe: the client rebuilds what it cannot fetch and re-uploads it. An account with no instance builds correctly and slowly. The same property covers every other gap in the loop, so an instance that is provisioning, recovering, or refused for capacity is an account paying cold build times rather than one that is broken.

### Which plans are archived

| Plan | Archived after |
|---|---|
| Enterprise | never |
| Pro | 90 days |
| Air | 90 days, or 60 under capacity pressure |
| Open source | never provisioned, so never archived |

Enterprise is excluded in code rather than by configuration, so it does not depend on anyone remembering. The population makes the trade one-sided: paid accounts are almost entirely active, so archiving them frees under a sixth of a machine while handing the accounts with the highest support expectations two cold builds on their return. The exclusion is archival-side only, so an Enterprise row archived before this change still cold-returns rather than being stranded.

### Demand

Recorded at the request boundary, cache-endpoint resolution, rather than derived from analytics. That covers all three lanes uniformly, is the same call for a developer machine and a runner build, and is the only signal available when an account has no instance, which is what lets an archived account ask for one back instead of deadlocking. The hot path is one ETS insert keyed by account; the flush resolves plan and service region once per account.

It is a proxy rather than a measure, and it errs in both directions. `tuist setup cache` installs a LaunchAgent with `RunAtLoad`, so an account whose agent is installed but idle refreshes its clock on every login without anyone building. In the other direction the CLI caches a resolved endpoint for an hour, so most requests during a build never reach the server. Both are accepted deliberately: holding an idle account warm is the safe error. The consequence is that the archival population should be read as a floor rather than an estimate when ordering machines.

### Configurable windows, and no feature flag

```
TUIST_KURA_INACTIVE_DAYS               (default 90)
TUIST_KURA_PRESSURE_INACTIVE_DAYS      (default 60)
TUIST_KURA_DEMAND_TRACKING_GRACE_DAYS  (default 7)
TUIST_KURA_ARCHIVAL_SWEEP_CRON         (default @daily)
```

Production keeps the spec's values by omitting them. These exist because the windows were compile-time constants, which meant the archival half could not be exercised in any environment without waiting a quarter: its first real run would have been in production against customer instances. Staging runs 1, 1, 0 and sweeps every five minutes, so the whole cycle is exercised continuously.

An unreadable value raises rather than falling back to the default, so a typo cannot leave an operator believing they had changed a window.

Archival on this branch is **live, with no feature flag**: the sweep is scheduled unconditionally and there is no `:kura_archival` gate in the code. An earlier revision had one, off by default, which meant shipping something that never ran. Two consequences worth naming:

- The spec asked that archival be "independently disableable by plan or by account during an incident". That instant, granular switch is gone. What remains is `keep_warm` to hold a named account-region out of archival, the Enterprise exclusion in code, and `TUIST_KURA_INACTIVE_DAYS` as a blunt fleet-wide lever needing a deploy. If a per-account incident switch is wanted, `keep_warm` is the natural home.
- Production archival begins on its own rather than at a chosen moment: after the backfill runs, rows become archivable once the tracking grace elapses. That grace period, plus "an account-region with no lifecycle row is never archived", are the whole of the protection against archiving on unseeded demand.

### Not archiving the fleet on the first sweep

The backfill unions `kura_usage_events`, `cas_events`, `gradle_cache_events`, and `command_events` filtered to `cacheable_targets_count > 0`. It also seeds every live account-region at the lookback boundary first, because every source is bounded to that window: without it, an instance whose account has been silent for longer than the lookback gets no row at all, and the coldest instances, the ones the lifecycle exists to reclaim, would be the only ones it could never touch.

### Where the free tier runs

Staging has no `us-east` pool, and Air was pinned there, so every Air account on staging resolved to a region whose instances could never schedule -- and the pressure rule is Air-only, so the archival half could not be exercised at all. `TUIST_KURA_AIR_REGION` (default `us-east`) names where the free tier runs; staging sets `ca-east`.

An earlier revision solved this with `TUIST_KURA_SERVICE_REGION_SUBSTITUTIONS`, a general `from=to` map applied to the policy's answer. That asserted one region *is* another, which is false, and nothing checked that a substitution preserved the guarantee the policy made -- so a production typo could have moved accounts across a residency boundary. Air is configurable because it carries no residency guarantee; paid regions are not, because an account restricted to Europe or the USA chose that, and no deployment setting may relocate it.

### Admission belongs to the scheduler

Every cache pod requests its claim's worth of `ephemeral-storage`, so Kubernetes bin-packs instances against each node's disk and declines to place one that does not fit. Nothing in the server decides whether a region has room.

This is not a refinement of the previous model, it is a replacement for it. Kura's data volume is a `scw-local-nvme` PV, which is Rancher local-path: a directory on the node's ephemeral-storage filesystem. The declared `storage_size` is a label, not a limit -- ext4 enforces no quota and XFS project quotas are out of scope -- so until now nothing stopped a region filling its box, and a full box crosses kubelet's eviction line and takes down every tenant on it. That is the July eu-central outage, where four paying tenants using about 12 MB each were evicted because one instance filled the disk.

The request is what the scheduler admits against, so it is the admission control the claim never provided. Three details worth reviewing:

- **Request, no limit.** A limit is enforced against the pod's writable layer, logs and emptyDir, none of which is where the cache lives. It would evict on the wrong signal while leaving real consumption unbounded. Kura bounds its own ring with `KURA_CAS_CAPACITY_BYTES`.
- **No pool to roll first.** Unlike `tuist.dev/memory-ceiling-mib` and `tuist.dev/egress-mbps`, which the pods already request and which stay opt-in because a pod requesting a resource its node does not advertise never schedules, `ephemeral-storage` is built in and every node already advertises it.
- **Replicas handle themselves.** Each replica requests one claim, so co-located replicas reserve their sum without the controller reasoning about placement.

`@manifest_revision` is bumped so live instances pick the request up rather than only new ones.

### Pressure is read, not modelled

The scheduler cannot answer the one capacity question that is not per-pod: whether a region is tight enough that Air's window should shorten from 90 days to 60, so archival makes room ahead of placements starting to fail. That reading now comes from the same numbers the scheduler places against -- the `ephemeral-storage` the region's pods have reserved, against what its Ready nodes make allocatable.

Reservations rather than live usage, deliberately. A freshly provisioned instance holds almost nothing and fills over days, so a region full of new instances would read as empty right up to the moment they all filled at once. What the region has promised is the thing that has to fit.

### What this deleted, and why it is worth deleting

`Capacity.admit/2`, the `capacity_refused` telemetry, the installed-capacity constant, the forecast built from this repo's rows, and the per-plan warm-quota table.

That table is worth calling out: it claimed Air holds 24 GiB and paid plans 50, but the claim is a property of the region, so an Air instance has always reserved exactly what a paid one does. It described an enforcement that does not exist. Giving Air a smaller cache would mean sizing the claim per instance, which is its own change -- and worth doing, since Air is currently the same cost per instance as Pro.

Pulling this thread also turned up how far the old model had drifted. It computed installed capacity as `machines * 1105 GiB`, and no box in production is that size:

| Pool | Nodes | Allocatable | Old model assumed |
|---|---|---|---|
| `kura-dedibox` (eu-central) | 1 | 825 GiB | 1105 GiB |
| `kura-us-east` | 1 | 789 GiB | 1105 GiB |
| `kura-us-west` | 1 | 789 GiB | 1105 GiB |

Combined with counting one quota per instance where the bare-metal regions run two co-located replicas, admission believed a region held roughly 2.8x the instances it can. A second copy of the cluster's own accounting drifted from it by a factor of three, which is the argument for not keeping one.

## CLI changes

**The cache daemon starts when no endpoint is serving yet.** An archived account has no endpoint until the server provisions one back. `tuist cache start` warmed the endpoint with a bare `try await` before constructing the gRPC server, so that transient absence stopped the daemon starting at all: no compilation cache for the whole provisioning window, and a failed step in CI. Warming is now best-effort, gated on `CacheURLStoreError.isTransientAbsence`, which separates "no endpoint yet" from "endpoint is wrong". A build never failed even before this, because `CASService` and `KeyValueService` both catch resolution errors and degrade to building locally; the failure was specific to the daemon start.

**The endpoints response carries `Cache-Control`.** How long a resolved endpoint stays good for is a property of the answer, not a fact about server state the client has to reinterpret, so the server states it directly: `private, max-age=3600` while an instance is serving, `private, max-age=30` while one is being provisioned back. An earlier revision put a `provisioning` boolean in the body and had the client map it to a TTL, which is the same decision made twice, in two places that could disagree.

That collapses the window in which a build stays on a stand-in lane from up to an hour down to the provisioning time itself, which otherwise costs two cold builds rather than one. A server that sends no directive leaves `maxAge` nil and the client on its own hour, so an older server keeps working.

## Design decisions worth reviewing

- **`:archived` is not `:destroyed`.** Destroyed is operator-driven and terminal. The partial uniqueness index only excludes destroyed rows, so an archived row keeps owning `(account, region)` and is the row a cold return reuses. The dashboard gains "Draining" and "Archived (inactive)".
- **Teardown is gated on observation.** Entering drain-pending unpublishes the endpoint and leaves the workload running. Teardown is issued only after the drain window, and the row becomes archived only once the backing resource is observably gone. Stamping `teardown_started_at` before the delete is the point of no return.
- **The lock is the authority.** `activate_server`, `record_observation`, and `fail_server` all reject `:drain_pending` and `:archived` under the row lock, because the sweep runs concurrently with the reconciler and a preloaded status can be stale.
- **Open deployments are cancelled on drain entry**, and the reconciler skips lifecycle states like the teardown states they are. A cancelled deployment is also not counted as evidence the image was delivered, or a cancelled drain would strand the instance on an old image.
- **The drain window reuses the warm-handoff constant**, lifted to `Kura.drain_seconds/0`.
- **Capacity is one model for two callers.** Admission refuses rather than overcommitting a node, and Air's window shortens only while the region's forecast does not fit what is installed. An unconfigured machine count means unknown capacity: provisioning is admitted and pressure archival never runs. Provisioning uses the same window archival does, which stops an archived instance being re-provisioned by the demand that failed to save it.
- **Cold return is a genuine cold start.** The reconciler no longer projects `:replicating` when there is nothing to catch up from. This also changes a first-ever deploy, which is peerless for the same reason.
- **Reclaimed bytes are the enforced warm quota** the instance held, which is what the region gets back.

Paid-plan warm quota is 50 GiB, taken from the region catalog's existing `storage_size` rather than invented, since the spec gives only Air's 24 GiB and XFS project-quota enforcement is out of scope.

## Staging validation

The full cycle, driven on staging against real infrastructure. Demand for one Air account was backdated three days; nothing else was touched.

| Time | Event |
|---|---|
| 15:46:03 | Demand backdated (one account only) |
| 15:50:00 | `draining instance d4757e08… (inactive)` |
| 15:50:00 | Endpoint unpublished, pod left running to drain |
| 15:52:10 | Backing `KuraInstance` gone from the cluster |
| 15:53:00 | `archived …: reclaimed 25769803776 bytes after 180000ms` |
| 15:53:15 | Cache request returns `{"endpoints":[],"provisioning":true}` |
| 15:54:00 | `returned account 48 from archive in ca-east` |
| 15:54:37 | New instance Ready, 2 replicas |
| 15:56:20 | Endpoint published, `/up` 200 in 0.40s |

`25769803776` bytes is exactly 24 GiB, the Air enforced warm quota, so plan resolution and `Capacity.warm_quota_bytes/2` were exercised against a real account rather than a fixture. The 180000 ms drain is the 120-second window plus the reconciler tick that observed teardown.

A second account with recent demand was left untouched throughout, confirming the sweep selects on the demand clock rather than archiving everything eligible-looking.

Also confirmed on staging:

- **Archive cancellation.** An earlier attempt was cancelled when a genuine cache request arrived mid-drain: endpoint republished and instance back in service within one tick.
- **Teardown gated on observation.** The row stayed `drain_pending` after the Kubernetes resource vanished, flipping to `archived` only once the reconciler observed it gone.
- **Cold provisioning from a first request**, twice, including the `provisioning` signal moving false to true to false across the cycle.
- **The activation gate.** An instance reported `Ready` by Kubernetes while its public host still refused connections, and the server correctly withheld the endpoint rather than publishing a URL that would have failed.
- **The staging Air region**, since the Air account landed in `ca-east` rather than the `us-east` staging has no hardware for.

## Impact

Nothing archives on merge: production sets none of the window variables, so the tracking grace alone means no row is archivable until a week after the backfill runs.

Rollout: deploy, then run `BackfillCacheDemandWorker`. Because there is no flag to turn on and capacity is read from the cluster rather than declared, archival begins on its own once the tracking grace elapses -- the grace period is the schedule, not a switch someone throws.

Live on merge: the demand hook, which adds one ETS insert to cache-endpoint resolution; the fallback change, which routes lifecycle-managed accounts with no instance to the Tuist-hosted lane rather than their own custom endpoints; and the CLI daemon change.

Before enabling in production, an alert on region occupancy is worth having. `tuist_kura_capacity_reserved_gibibytes` against `tuist_kura_capacity_allocatable_gibibytes` is the signal that says "order a box", and it should page well before the ratio reaches 1: past that point the scheduler starts leaving instances Pending, and an account with no instance keeps re-resolving on the short `max-age` while it waits. Capacity should be procured ahead of demand, with the scheduler as the backstop that stops a region overfilling in the meantime -- not as the mechanism that manages capacity. Grafana alert rules are not provisioned from this repo, so that has to be created separately.

### How to test locally

Server, from `server/`:

```bash
mix test test/tuist/kura/ test/tuist/accounts_test.exs
```

CLI, from the repository root, after `mise run cli:ee` and `tuist install`:

```bash
mise exec -- tuist generate TuistCAS TuistCASTests TuistKit TuistKitTests --no-open
```

Then run `TuistCASTests/CacheURLStoreTests` and `TuistKitTests/CacheStartCommandServiceTests`.

Validation run:

- Full server suite 6690 passing. The 4 failures are pre-existing locale tests that fail identically on a clean tree, verified by stashing.
- `TuistCASTests/CacheURLStoreTests` 14 passing, `TuistKitTests/CacheStartCommandServiceTests` passing.
- `mix credo`, `mix format --check-formatted`, `MIX_ENV=prod mix compile --warnings-as-errors`, `swiftformat --lint`, and `swiftlint` all clean. The prod-env compile is the gate the release build runs; the dev-env one fails on a pre-existing locale warning.

## Out of scope

XFS project-quota enforcement and the quota-aware local provisioner, the full account-region policy override record and its operations UI, renaming `egress_guaranteed_mbps`, the network hard ceiling, and default-on client routing with its per-account opt-out and regional kill switch. `keep_warm` is represented only far enough for the lifecycle to honour it.

Two things were built and then removed rather than shipped. A gate refusing archival when no fallback lane is configured was dropped: with the CLI daemon fix, no fallback means a cold build, which is what archival exists to trade for. An ops Archive/Unarchive control was reverted in favour of the configurable windows: it needed an audit table, a boundary export, and a way to survive the resolution pass cancelling the very override that had bypassed it, which is a lot of machinery for a debugging aid.

## Review fixes

Four correctness issues found in review, each with a test that fails without the fix:

- **An explicit destroy was undone on the next tick.** The lifecycle row still held recent demand, so the identity rule recreated the instance and the destroy control in the dashboard did nothing. Provisioning now skips an account-region whose most recent destroy is newer than its last cache demand. Demand recorded *after* the destroy still provisions, which is the rule archival already follows: a destroy is a statement about now, not a permanent veto.
- **Provisioning trusted the region on the lifecycle row.** That row records where demand was served; the policy decides where it belongs now. They diverge when an account changes plan or storage region while it has no instance, so a downgraded or reassigned account could be resurrected in a region it had left. Provisioning now re-resolves the account and proceeds only when the current resolution names the region being reconciled.
- **Lifecycle transitions could overwrite a concurrent destroy.** `begin_drain/1`, `cancel_drain/1` and `archive_server/1` wrote through preloaded structs, so a destroy committing between the read and the write was silently replaced -- and a cancelled drain would republish the endpoint of a server being torn down. All three now lock the row and re-read its status, the guard activation already used.
- **The demand flush was N+1.** Each paid account allowing every region cost a query for its current assignment, on a minute-level path whose batch is every account that used the cache. `AccountPolicies.resolve_all/1` loads them in one.

Also in this pass: the TuistCacheEE pointer had moved back to the parent of main's commit, dropping the optional-authentication warning fix from TuistCacheEE #72; it is repinned to main's commit. The touched clean-service test moved to Swift Testing.

## Follow-up: two settings removed

Review flagged both deployment settings as smells, and both were.

`TUIST_KURA_REGION_MACHINES` is gone; capacity is counted from the cluster. `TUIST_KURA_SERVICE_REGION_SUBSTITUTIONS` is gone, replaced by `TUIST_KURA_AIR_REGION`, which names the one real decision behind it. See the two sections above.


pepicrft (approved):
Approved. I left one non-blocking inline note in the app-managed review: archiving should also clear the last observed timestamp.

PR #12420, opened 2026-08-17
feat(server): raise the rolling trigger window cap to 1000 runs

fortmarek:
### What changed

Rolling trigger windows for test automations were capped at 75 runs. The dashboard surfaced that as a `1-75` validation error on the "Last N runs" field, which reads as an arbitrary product limit. It was a storage constraint, and this PR removes it: trigger windows now go to 1000, matching the recovery cap.

- New ClickHouse aggregate `test_case_runs_recent_packed_per_case` plus its materialized view.
- `FlakyTestsMonitor` routes on window size: up to 75 keeps reading the existing 100-entry bucket, above 75 reads the new one.
- `Alert.max_rolling_trigger_window_size/0` goes 75 → 1000. Both dashboard inputs already derive their `max` from these functions, so only the stale `placeholder="75"` needed touching.

### Why the cap existed

The only rolling aggregate still ingesting was the 100-entry bucket, and 25 of its 100 slots were reserved as headroom for the correction rows flaky detection re-inserts. The bucket family that used to serve larger windows (250/500/750/1000) was retired in `20260724140000` because merging the 750-entry table exhausted the production ClickHouse memory limit.

So the blocker was real: no aggregate could serve more than ~100 entries per test case without reintroducing that blowup.

### Root cause

I benchmarked the alternatives against a synthetic 2.4M-run dataset on ClickHouse 26.1, at the monitor's own `max_threads=2, max_memory_usage=1 GiB`.

**The retired shape was unviable to read, not just to merge.** `groupArraySortedMerge(1000)` over a 2000-test-case evaluation range costs 650 MiB and 1.7s on its own. Merge memory was only half the problem.

**Raw `test_case_runs` reads don't fit the hot path either.** The original aggregate migration justified itself with "no primary key prefix fits last-N-per-test-case", which is no longer true — `scheduled_alert?/1` returns false for scoped alerts, so an established rolling alert is always evaluated with `project_id` and `test_case_id IN (...)` fully constrained. But measurement killed it anyway: neither `groupArraySorted` over raw rows nor `ORDER BY … LIMIT N BY test_case_id` terminates early (both read 819k rows for 800k actual), and adding a `ran_at` floor still leaves a ~1–2 granule floor per test case. At 2000 test cases per range that is 30M+ rows.

**The cost is in the entry encoding.** `AggregateFunction(groupArraySorted(N), Tuple(Int64, UInt8))` puts the aggregate on ClickHouse's generic tuple comparator, and because each tuple carries one flag it needs two parallel columns to serve flakiness and reliability. Packing each run into a single `Int64` takes the specialized numeric path and lets one column serve all three monitor types:

```
-toUnixTimestamp64Micro(ran_at) * 4 + is_flaky * 2 + is_success
```

| aggregate | window | memory | time | storage |
|---|---|---|---|---|
| tuple ×2, N=100 | 75 | 73 MiB | 152 ms | 3.9 MiB |
| tuple ×2, N=1000 | 500 | 650 MiB | 1658 ms | 38.2 MiB |
| packed Int64, N=2000 | 1000 | 61 MiB | 154 ms | 18.4 MiB |

The packed aggregate serves the full 1000-run cap for less than the 100-entry bucket costs at a 75-run window.

### Why this encoding is correct

- The timestamp is negated and scaled above the flag bits, so ascending order is latest-first and the low bits never perturb ordering across runs.
- A correction row raises `is_flaky` 0 → 1, so it sorts after the run it corrects and the reader's existing linear de-duplication keeps it.
- Flag and run-key reads go through `reinterpretAsUInt64`. Packed entries are negative, and neither `intDiv` nor a signed shift is consistent across the sign boundary — verified directly in ClickHouse that all four flag combinations of one timestamp share a run key and decode correctly.
- The bucket holds 2000 slots for a 1000-run cap. `Tests.report_test_case_run_multiplicity/3` already enforces and alerts on at most two physical rows per logical run, so a window is exact even when every run in it was corrected, instead of relying on the 75% fractional headroom the old cap assumed.

### Containment for the failure that retired the previous tables

The new table is partitioned on a hash of `project_id` — this bounds a merge to a sixteenth of the table and prunes a `project_id` read to a single partition (verified: 16 parts → 1). `merge_max_block_size` is also pinned low, because the byte-based limiter sizes merge blocks from on-disk bytes and these states compress ~59× (38 MiB of tuple state landed as 665 KiB on disk), so it under-counts the in-memory arena by that factor and cannot contain this on its own.

### Two decisions worth reviewing

**Forward-only, no backfill.** Matching `test_case_runs_validated_on_branch`: a synchronous scan of a multi-billion-row fact table blocks the deploy's migration hook, and past backfills of this table family exhausted production memory repeatedly.

**Large windows require a filled window.** Because of the above, a window over 75 is only evaluated once the aggregate holds `size` distinct runs. Without this, a new 500-run alert would quarantine a test off the 20 runs accrued so far. It is also the correct reading of "the last N runs", and the enterprise shape this enables — recover only after N consecutive clean runs — is not answerable from a partial window. Practically this means a new 300-run alert stays silent for a few days while the window fills. Windows up to 75 keep evaluating partial windows exactly as before, so no alert running today changes behavior.

One consequence to note: the `ArgumentError` in `recent_runs_source/2` is now unreachable, since `parse_rolling_size/1` clamps to 1000 first. I kept it as a backstop and replaced its test with one asserting the reachable behavior (clamping), rather than leave a test for dead code.

### How to test locally

From `server/`:

```bash
mix test test/tuist/automations/ test/tuist_web/live/project_automations_live_test.exs
```

The packed-aggregate tests in `flaky_tests_monitor_test.exs` run against real ClickHouse through the ingestion fixtures and the materialized view, so the packing, the bit-level flag reads, the correction de-duplication, and the filled-window gate are all exercised as ClickHouse actually evaluates them — not through a mocked repo.

To see the cap change in the dashboard, open a project's Automations tab, create an automation with a rolling window, and confirm "Last N runs" now accepts up to 1000.

### Validation

- `mix test test/tuist/automations/ test/tuist_web/live/project_automations_live_test.exs` → 182 passed
- `mix credo` → 13571 mods/funs, no issues
- `mise run format` / `mix format --check-formatted` → clean
- Migration applied against a real ClickHouse and the resulting table shape and partition pruning inspected directly

Full `mix test` was still running when this PR was opened; I will confirm the result on the PR.


server/priv/ingest_repo/migrations/20260817120000_create_test_case_runs_recent_window_per_case_mv.exs
```diff
@@ -0,0 +1,121 @@
...
+
+  def up do
+    drop_views!()
+
+    # excellent_migrations:safety-assured-for-next-line raw_sql_executed
+    IngestRepo.query!("""
+    CREATE TABLE IF NOT EXISTS test_case_runs_recent_window_per_case (
+      project_id Int64,
+      test_case_id UUID,
+      recent_runs AggregateFunction(groupArraySortedDistinct(#{@bucket_size}), Int64),
```
esnunes: The reasoning for `Distinct` is really appealing — making the bucket the window cap rather than a multiple of it is clearly the better invariant. But I think the combinator doesn't have the property the design leans on: the `1000` parameter bounds the *output*, not the state. ClickHouse implements `-Distinct` by keeping a set of every distinct argument it has seen, so the state stored per (project_id, test_case_id) grows with every run that test case has ever had, and never prunes.

I measured it on ClickHouse 26.1.2.11. `groupArraySortedState(10)` is 81 bytes regardless of input size; `groupArraySortedDistinctState(10)` is 801 B over 100 distinct values, 80 KB over 10,000, and 1.6 MB over 200,000. `…MergeState` behaves the same way, so background merges carry the whole set forward too.

Read at this module's own `max_threads=2, max_memory_usage=1 GiB` over a 2000-test-case range, using the DDL from this migration:

| distinct runs / test case | peak memory | duration |
|---|---|---|
| 2,000 | 153 MiB | 218 ms |
| 5,000 | 528 MiB | 683 ms |
| 10,000 | MEMORY_LIMIT_EXCEEDED | — |

For comparison, `groupArraySorted(2000)` over packed Int64 — the encoding in 489d392579 — served the same 1000-run window at 20,000 runs per test case in 139 MiB and 185 ms, and stays flat because its state size doesn't depend on run history.

I checked the run-per-test-case distribution in production before flagging this, and a meaningful tail of test cases already accumulates more runs within a recent bounded window than the volume at which I measured the ceiling being crossed — and this aggregate has no time bound at all, so it only accumulates.

I don't think the benchmark in the description could have surfaced it: 2.4M runs over ~2000 test cases is about 1,200 runs each, right at the bucket size, where the distinct set and the bounded array are nearly the same object. The growth only shows up as history accumulates — which is also why it wouldn't fail on the day it ships.

Would you consider going back to the bounded state and keeping the read-time linear collapse for correction rows? The `Distinct` invariant is nicer, but this looks like it reintroduces exactly the unbounded-arena shape that retired the 750-entry table.

server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
```diff
@@ -449,33 +456,42 @@ defmodule Tuist.Automations.Monitors.FlakyTestsMonitor do
...
+  # keys that reach back no further than the window's oldest run counts exactly
+  # the flagged runs inside it. That is a linear pass over an already sorted
+  # array rather than a set intersection.
+  defp matching_run_count_expr, do: "arrayCount(entry -> entry <= recent_runs[-1], flag_runs)"
+
+  # A rolling window measures the last N runs, so it is only meaningful once N
+  # distinct runs exist to measure. Reporting a rate off whatever has accrued so
+  # far answers a question the user did not ask, and at small run counts it is
+  # noise that auto-quarantine would act on.
+  defp window_filled_expr(size), do: "length(recent_runs) >= #{size}"
```
esnunes, line 513: Following up on this one, because you were right to push on it — there *was* a branch, and it isn't a hardcoded 75.

At `ea3a5d9e3c`, `window_filled_expr/2` had two clauses dispatching on which aggregate served the window:

```elixir
defp window_filled_expr({:tuple, _column}, _size), do: "length(recent_runs) > 0"
defp window_filled_expr({:packed, _column}, size), do: "length(recent_runs) >= #{size}"
```

The tuple bucket served windows up to `@max_tuple_rolling_window_size` and kept evaluating partial windows exactly as before; only the new aggregate required a filled one. So the claim in the description was accurate when it was written.

`89ed6fe73a` ("serve every rolling window from one aggregate") collapsed the two aggregates into one, and the two-clause function collapsed with them into the single `window_filled_expr(size)` now at flaky_tests_monitor.ex:491. Grepping the module, there's no size branch left anywhere — the only size-conditional clause is `recent_runs_source/2` raising above 1000. So the filled-window rule now governs small windows too: a 10-run window on a test case with 3 runs returns nothing where it used to evaluate.

The tests moved in step with that, which is probably the clearest signal: the five-run window at flaky_tests_monitor_test.exs:601 needed three extra runs to keep passing, two other windows dropped from 5 to 1, and the new "does not evaluate a small window the aggregate has not filled yet" case asserts the gate at size 10.

I think the gate itself is the right call — a flakiness rate over 3 runs is noise that auto-quarantine would act on. It's really the description (and the changelog entry that was removed, which said windows of 75 or fewer are unaffected) that stopped matching the code when the routing went away. Worth a sentence saying any rolling automation now stays quiet until its window is full?

server/priv/ingest_repo/migrations/20260818120000_backfill_test_case_runs_recent_window_per_case.exs
```diff
@@ -0,0 +1,254 @@
...
+        """
+        SELECT -toUnixTimestamp64Micro(toDateTime64(min(inserted_at), 6))
+        FROM schema_migrations
+        WHERE version = {version:Int64}
+        """,
+        %{version: @window_view_migration_version}
+      )
+
+    case rows do
+      [[boundary]] when is_integer(boundary) ->
```
esnunes: Small one on `window_view_boundary/0`. The `_ ->` branch that logs "is not recorded; backfilling every run in the source" looks unreachable, and the behaviour in that case is the opposite of the intent.

`schema_migrations.inserted_at` is a non-nullable `DateTime`, and `min()` over an empty match returns the epoch rather than NULL — so the expression evaluates to `0`, `is_integer(0)` matches the first clause, and since every run key is negative, `run_key > {boundary}` filters all of them out. The migration then succeeds having inserted nothing, with no warning.

I ran the migration's own boundary query against the production ClickHouse to check this rather than reasoning about it — for version `20260817120000`, which isn't deployed yet and so is genuinely absent, it returns `boundary = 0` against `0` matching rows. The column really is non-nullable there too.

In an ordered deploy the version is always recorded, so I don't think this fires in practice — it's that the one situation the guard exists for is the one it silently inverts, and an empty backfill is hard to notice afterwards. `toNullable(min(inserted_at))`, or a `count()` check, or just guarding on `boundary < 0`, would make the fallback behave as documented.

esnunes (approved):
Really enjoyed reading this one. Chasing the 75-run cap back to its actual cause rather than treating it as a product limit, and then measuring the alternatives instead of arguing about them, is exactly the right shape for this change — and the ClickHouse tests running the packing and the correction collapse through real ingestion fixtures are what made the rest of it reviewable.

Approving, with one thing I'd genuinely like you to look at before it goes out. The final refactor swapped the bounded aggregate for `groupArraySortedDistinct`, and I don't think that combinator has the property the design leans on: the `1000` bounds the output, not the state, so each per-test-case state accumulates every distinct run key that test case has ever had. Measuring it on 26.1.2.11 at this module's own 1 GiB ceiling and a 2000-test-case range, it reads at 153 MiB with 2,000 runs per test case, 528 MiB at 5,000, and hits MEMORY_LIMIT_EXCEEDED at 10,000 — where the bounded encoding from 489d392579 stays at 139 MiB with 20,000. Details and the reproduction are in the inline note on the migration. The `Distinct` invariant is genuinely nicer to reason about, so it's a shame, but this looks like it reintroduces the unbounded-arena shape that retired the 750-entry table, just on a slower fuse.

Two lighter notes inline: the filled-window gate lost its size branch when `89ed6fe73a` collapsed the two aggregates into one, so it now governs small windows too and the "no alert running today changes behavior" framing needs adjusting (I think the gate itself is right); and the backfill's "version not recorded" fallback can't be reached, so it silently backfills nothing instead — I checked that one against the production `schema_migrations` rather than reasoning about it.

One that has no line to anchor to: the marketing changelog entry was added and then dropped again in `2fd8e350de`, so it nets out of the diff entirely. The repo's review guidance (`.blick/skills/tuist-elixir-review/SKILL.md`, rule 11) asks for a `server/priv/marketing/changelog/*.md` entry when a PR's primary purpose is launching a customer-facing capability, and exempts fixes and internal or flagged rollouts — this one raises a limit customers see and type into, and it's a `feat(server)`, so it looks more like the announce case. The entry that was removed also carried the only customer-facing statement of the quiet-until-filled behaviour, which is the part most likely to generate a support question ("my automation stopped firing"). If it's being held for a broader automations announcement that makes sense — just checking it was deliberate rather than swept up in the cleanup.

server/priv/ingest_repo/migrations/20260817120000_create_test_case_runs_recent_window_per_case_mv.exs
fortmarek: Confirmed and reverted in 16b27c2b0b — you were right, and my benchmark had exactly the blind spot you identified. Reproduced your measurement independently on 26.1.2.11: at a bound of 10, `groupArraySorted` stores 89 B regardless of input, while `groupArraySortedDistinct` stores 809 B / 78 KiB / 1.53 MiB over 100 / 10,000 / 200,000 distinct values, so the parameter bounds the output and not the state. That is the unbounded-arena shape that retired the 750-entry table, with no bound at all this time. Back to `groupArraySorted(2000)` over packed Int64 with the read-time linear collapse, as you suggested. The bucket constant now records why de-duplicating in the aggregate is unavailable, so it is not retried — the 2x headroom is the cheaper trade.

server/lib/tuist/automations/monitors/flaky_tests_monitor.ex
fortmarek: Agreed on all of it, and thanks for tracing the history rather than taking my summary at face value — the two-clause dispatch is exactly what existed and 89ed6fe73a is what collapsed it. Documented at the call site in 16b27c2b0b: `window_filled_expr/1` now states that the rule governs every window size, that sizes at or below the old tuple ceiling used to evaluate whatever had accrued, and that a test case with fewer runs than its window stops being measured and so lets recovery re-arm it. The PR description is still stale on this (and on the encoding), so it needs a rewrite rather than a patch — that is with the author.

server/priv/ingest_repo/migrations/20260818120000_backfill_test_case_runs_recent_window_per_case.exs
fortmarek: Good catch, and confirmed by reproducing it — the boundary query against an absent version returns `0` with 0 matching rows, `is_integer(0)` matches the first clause, and every negative run key is then filtered out, so the backfill succeeds having written nothing. Worth noting `toNullable(min(inserted_at))` does not help, since it wraps the epoch-derived `0` rather than a NULL. Fixed in 16b27c2b0b by selecting `count()` alongside the boundary and gating the clause on `recorded > 0`, so the fallback runs in exactly the case it documents.

PR #12422, opened 2026-08-17
fix(kura): stop recording client hangups as 5xx on artifact transfers

fortmarek:
### What changed

`serve_accelerated` writes the `200 OK` response line before it starts sendfile/splice, so once the body is in flight the status is already committed. Every failure after that point was recorded as `INTERNAL_SERVER_ERROR` and the `io::Error` was dropped without a log line.

Two consequences: a client that walked away mid-download was counted as a server error it never actually received, and a genuine server-side fault was indistinguishable from that routine churn.

`TransferFailure::classify` now splits the branch three ways:

| Error kind | Recorded as | Log |
| --- | --- | --- |
| `BrokenPipe` / `ConnectionReset` / `ConnectionAborted` | 499, `client_aborted` | debug |
| `WouldBlock` / `TimedOut` (the 120s `SO_SNDTIMEO`) | 500, `write_timeout` | warn |
| `UnexpectedEof` (file shorter than its record) | 500, `incomplete` | warn |

499 is nginx's convention for a client-closed request, so it reads the same way in dashboards. Nothing writes it to the wire; the client already received its `200 OK`. Aborts log at debug because they are routine and high volume.

`transfer_splice` also had to stop conflating its two zero-length cases. Zero *out of* the pipe means the socket will take nothing more, which is the peer leaving, and now returns `ConnectionAborted`. Zero *into* the pipe is the input running short and remains the only path reaching `ensure_complete_transfer`. Without this, a hangup on the splice path could be reported as a short file, which is precisely the case the new classification exists to identify.

### Why

This is what has been firing the `Kura - 5xx errors on /api/cache/module/{id}` alert ([rule](https://tuist.grafana.net/alerting/grafana/cftoutryd1jwge/view?orgId=1)).

Investigating the 2026-08-17 firing: the pod was at its Cilium egress cap (`kubernetes.io/egress-bandwidth: 1500M`, rendered from `egress_burst_mbps` in `server/lib/tuist/kura/regions.ex`). Transfers stretched past 32s, 1000 of them inside five minutes against 74 in a comparable window with more traffic, and more clients gave up. Roughly 33 aborted out of about 44k requests on the route, or 0.075%.

The key point is that egress shaping alone cannot produce an error. It only makes transfers slow. Turning slow into an error requires a deadline, and both available deadlines are someone giving up rather than the server breaking:

- The 120s `SO_SNDTIMEO`, which needs 120 consecutive seconds with zero bytes accepted, since any partial progress restarts the clock. Under fq/EDT shaping that is implausible, but it is very plausible when the peer stops reading.
- The client's own read timeout.

So the defect was never the shaping, and it was not the alert threshold either. It was that the metric counted the wrong thing. Memory admission was ruled out as a factor: `ResponseStreamAdmissionPatience::Blocking` waits only 5s and sheds with a real 503 on failure, and zero 503s were recorded, so the accelerated path always got its permit.

### Why this over the obvious alternative

The first instinct was to retarget the alert to `status="503"`, or give the 500 arm a rate-relative threshold instead of `> 0`. Both would have silenced it, and both would have buried the `incomplete` case along with the noise. An artifact on disk disagreeing with the record describing it is a real fault that currently has no other signal anywhere. Fixing the classification at the source keeps that visible and leaves the alert correct as written.

### Impact

No behavioural change for clients. A client that hangs up already received a `200` and a truncated body, and still does. This only changes what the server records and logs.

After deploy, `status="499"` and `result="client_aborted"` carry the routine churn, and a 500 on this route means a genuine server fault again. The alert's `> 0` threshold becomes correct rather than something to tune away. Dashboards that compute success as `status="200"` will see a new `499` label value on this route.

Note that this does not address the underlying capacity question, which is deliberately out of scope: one pod carries a tenant's entire regional load against its shaped cap, and raising `egress_burst_mbps` is a trade against the other tenants bin-packed onto the same NIC.

### How to test locally

```bash
cd kura && cargo test --lib accelerated
```

Also run before merging:

```bash
cd kura && cargo clippy --all-targets -- -D warnings
```

All 14 pass, clippy is clean, and `mise run format` has been applied. Tests were written before the implementation and confirmed failing first.

One caveat for reviewers: `client_that_hangs_up_mid_transfer_is_not_recorded_as_a_5xx` is gated to `#[cfg(target_os = "linux")]`, because sendfile/splice only exist there and off Linux `transfer_file` refuses up front without ever reaching the peer. It could not be executed on macOS, only type-checked by temporarily lifting the gate, so CI is the first real run of it. The three `TransferFailure::classify` unit tests are platform-independent and do run everywhere.


pepicrft (approved):

pepicrft (approved):

kura/src/accelerated_file_serving.rs
```diff
@@ -1210,8 +1220,13 @@ fn transfer_splice(
...
                     return Err(error);
                 }
+                // Zero out of the pipe means the socket will take nothing more,
+                // which is the peer going away rather than the input running
+                // short. Saying so keeps it out of `ensure_complete_transfer`,
+                // whose UnexpectedEof is reserved for a file that disagrees
+                // with the record describing it.
                 if spliced_out == 0 {
-                    break;
+                    return Err(std::io::Error::from(std::io::ErrorKind::ConnectionAborted));
```
esnunes: Small thought on this one, and it cuts the opposite way from my note on the sendfile branch.

The inner loop only runs while `pending > 0`, and `pending` came from the bytes the kernel just moved into the pipe — so the source provably has data, and the destination is a blocking socket with SO_SNDTIMEO armed. Which means splice returning 0 here shouldn't be reachable at all: the kernel should move a byte, park, or fail. Handling it is clearly right, but routing it into `ClientAborted` puts a should-never-happen case into the one bucket designed to be silent at volume — debug! at :664, then re-logged as a generic "public HTTP connection failed" at :114, also debug. With `Error::from(ErrorKind::ConnectionAborted)` carrying no message, an occurrence would be indistinguishable from an ordinary peer RST and invisible at production log levels.

Cheapest fix is probably just `Error::new(ErrorKind::ConnectionAborted, "splice to socket returned 0")` — same `kind()`, so classify is unaffected, but the Display says which one it was. `ensure_complete_transfer` already does exactly this with its byte counts at :1310. Very much take it or leave it.

esnunes (approved):
This is a really nicely argued change — the description does the thing I wish more fix PRs did, which is separate "what would make the alert stop" from "what is actually true", and then fix the second one. Rejecting the retarget-the-alert option because it would bury `incomplete` is the right call: an artifact on disk disagreeing with the record describing it has no other signal anywhere in Kura, and that case deserves to keep the 500 to itself.

Two things I checked and can confirm rather than just take on faith. First, the status really is the lever: `record_http` increments `kura_http_exceptions_total{kind="server_error"}` only inside `if status.is_server_error()` (metrics.rs:1466), so choosing 499 is what moves the alert — the `result` label alone wouldn't have. Second, on the consuming end, `infra/grafana-dashboards/tuist-kura.json` filters its error panel on `status=~"5.."` so 499 drops out cleanly, and the `$status` template variable is built from `label_values(...)` so the new value appears on its own; I couldn't find a panel in that file computing success as `status="200"`, so within that dashboard the new label reads as additive.

The `transfer_splice` change is also doing slightly more than the description claims, in a good way. That `break` sat in the inner `while pending > 0` loop, so a zero splice-out didn't stop the transfer — it fell through and kept splicing more file data into a pipe that still held unsent bytes, quietly truncating the tail and under-counting `sent_total` before reporting a short file. Returning `ConnectionAborted` fixes the control flow as well as the classification.

I've left one note inline on that new `ConnectionAborted`. One more here, since it lands on a line this PR doesn't touch — `kura/src/accelerated_file_serving.rs:1150`, the `if sent == 0 { break; }` in `transfer_sendfile`, which still falls through to `ensure_complete_transfer` and reports `incomplete`. Working through it I think that's correct and the asymmetry is real: sendfile's destination is a blocking socket, so it either moves at least one byte, blocks, or fails — a zero return can only be the input file hitting EOF, which is exactly the short-file case. Splice differs because its second call writes from a pipe that provably holds data, so a zero there can only be the socket refusing. Does that match your reasoning? If so, the new comment at :1223 explains beautifully why splice is special, and a one-line counterpart on the sendfile branch would save the next reader re-deriving the difference. I did exactly that on the way through.

Neither is blocking — approving.

kura/src/accelerated_file_serving.rs
fortmarek: Taken as suggested in a4551a4051 — now `Error::new(ErrorKind::ConnectionAborted, "splice to socket returned 0")`, so `kind()` and the classification are unchanged but the Display separates it from an ordinary peer reset. Extended the comment to record that it is not expected to be reachable, and added the counterpart one-liner on the sendfile branch: your reasoning matches mine exactly, a zero there can only be the input hitting EOF because the destination is a blocking socket, which is why that one correctly reports `incomplete`.

PR #12429, opened 2026-08-18
fix(kura): extend blob lifetimes on the REAPI read paths

fortmarek:
### Purpose

`GetActionResult` presence-gates an action-cache entry (`first_evicted_output`, `kura/src/reapi/service.rs`) so it never serves one whose output blobs are already gone. Nothing then keeps those blobs alive. Between the node telling a client "yes, this key is backed" and that client's later `BatchReadBlobs`, eviction can remove them, and clang fails the build on a missing object rather than recompiling.

### Root cause

The gate answers from metadata alone. `first_evicted_output` checked presence with `artifact_manifest_exists`, and `find_missing_blobs` with `artifact_exists`. Neither routes through `prepare_artifact_for_serving`, which is the only caller of `enqueue_promotion`. So the check happened without ever touching the read path that could refresh a blob's lifetime. Correct for answering "is it there", and silently opted out of the mechanism that keeps it there.

### What the spec requires

Verified against `bazelbuild/remote-apis` `main`, [remote_execution.proto](https://github.com/bazelbuild/remote-apis/blob/main/build/bazel/remote/execution/v2/remote_execution.proto):

- `ActionCache.GetActionResult`: "Implementations SHOULD ensure that any blobs referenced from the ContentAddressableStorage are available at the time of returning the ActionResult **and will be for some period of time afterwards**. The lifetimes of the referenced blobs SHOULD be increased if necessary and applicable."
- `ContentAddressableStorage.FindMissingBlobs`: "Servers SHOULD increase the lifetimes of the referenced blobs if necessary and applicable."

Kura implemented the first half of the first clause and neither lifetime clause.

### What this guarantees, and what it does not

Worth stating up front, because the spec language is stronger than what this delivers. **The copy completes after the response, not before it.** Both RPCs answer as soon as the promotion is queued, and a background worker performs the copy. A vouched blob is therefore not safe at the moment the client is told it is backed; it is safe once the worker reaches it.

So this narrows the window rather than closing it. A crash, a `Critical` pressure window, or a promotion queue that cannot drain fast enough can still drop a blob the node vouched for, and the client still sees a missing object in that case.

Refreshing inline before responding is what would actually close it, and that is deliberately **out of scope here**. Kura pays a per-refresh fsync that made inline refresh untenable in #11719, so closing the gap requires batching the promotion drain's durability first. The reasoning and the path are in the follow-up section at the end.

### How eviction works, and why copy-forward

Blob bodies live in a segment ring with `old` / `current` / `new` bands. `push_new` cascades on rotation and evicts whatever falls off the front of `old`. Nothing consults access time. Two consequences drive the design: a blob's remaining lifetime is entirely a property of which segment holds it, and the eviction unit is a whole segment. So the only ways to keep one blob alive are to pin its entire segment or to move it somewhere younger.

This takes the second route, reusing the background promotion added in #11719. `maybe_refresh_manifest` re-validates under `segment_refresh_lock`, streams the bytes into the active segment, and flips the manifest pointer plus both `segment_artifacts` index rows in one atomic batch. The blob is then genuinely young, exactly as far from eviction as something written this second.

A reservation table (record segments with an expiry the evictor consults) was considered and rejected. Eviction unlinks whole segments, so pinning one small blob pins its entire segment. A reservation only defers, so under sustained pressure it must either be broken, in which case it does not close the gap it exists to close, or honored, which stalls rotation and converts a rare correctness problem into an availability one. It does not survive a restart, it needs new policy in the eviction path for the all-candidates-reserved case, and its expiry is a guess. Copy-forward needs no special case anywhere: after promotion the ordinary positional policy is already correct.

### What changed

**Both read paths extend lifetimes.** `Store::extend_artifact_lifetimes` resolves each vouched-for key and queues the ones in an Old segment. `first_evicted_output` returns the blob keys it confirmed present during the walk it already performs (output files, stdout, stderr, each output directory's tree, and the files that tree lists), so extension costs no extra reads and no second tree decode. An entry that fails the gate refreshes nothing: it is not served and is usually deleted, so promoting its surviving blobs would be pure write amplification.

Wiring lives at the REAPI call sites rather than inside `artifact_manifest_exists`, because the snapshot builder also calls that once per snapshot node and it has to stay cheap.

**Bounds and pressure.** Read-triggered refreshes reuse the single `run_promotion_worker` and the `segment_refresh_lock` mutex that serializes all copy-forward, so no new knob and no new config or Helm surface. Because these RPCs promise the blob stays fetchable, they promote one pressure tier deeper than serve-path promotion, which carries no such promise since its read already succeeded: they continue through `Constrained` and stop at `Critical`, where a read-path write would compound the squeeze. This was an explicit call, since it trades node stability against cache correctness.

**Queue admission and drain order.** `VOUCHED_PROMOTION_RESERVE` carves a slice of the existing bound that only `action_cache` and `find_missing` triggers may occupy, so a serve flood cannot crowd them out. Reserving a slice rather than raising the ceiling keeps total queue memory unchanged and admission O(1).

Reserving admission alone turned out to be insufficient. The queue was a single FIFO, so a refresh admitted into the reserve still waited behind the whole existing backlog, on the order of ten minutes at the serve ceiling. A vouch honored ten minutes late has not closed the window it exists to close. The queue now has two lanes and vouched work drains first, with `pending` as the authoritative membership set and stale deque entries skipped at pop time rather than scanned for at upgrade time. Upgrading an already-queued entry re-queues it in the vouched lane, since leaving it in place makes an upgrade indistinguishable from a drop in wall-clock terms.

**FindMissingBlobs allocation and lookups.** That RPC carries a client-controlled digest count up to the 64 MiB decode ceiling and, unlike the write paths, has no `GrpcWriteAdmission` body in front of it. Collecting a heap-allocated key per present digest added roughly 65 MB of untracked allocation at the ceiling, and the extension pass then resolved every key through `manifest()` a second time. That second pass lands on RocksDB whenever manifest-cache admission is closed, which is exactly the `Constrained` tier this change extended read-triggered refresh to run at.

`Store::artifact_exists_extending_lifetime` now answers presence and queues the copy-forward from one manifest lookup, so the loop retains no per-digest allocation and a present blob costs one lookup rather than two. `Store::segment_ring_is_aging` is hoisted once per request: with nothing aged there is nothing promotable, so the plain existence check runs and keeps its existence-cache short-circuit. `first_evicted_output` takes the same predicate as a parameter and skips collecting when it is false.

### Observability

`kura_segment_refreshes_total`, `kura_segment_refresh_bytes_total`, and `kura_segment_refresh_duration_seconds` gain a `trigger` label (`serve`, `action_cache`, `find_missing`) so the read-time write amplification each path adds is separable. Declined refreshes are counted as `result="pressure_skipped"`.

`kura_promotion_drops_total{trigger}` is new. The drop was previously silent, with no metric and no log, which made the failure unobservable. A nonzero `action_cache` or `find_missing` rate means extension is not keeping up and vouched blobs may be evicted before the client's follow-up read.

The `Segment Refresh Rate` panel in `kura-cluster.json` aggregated the new label away and now splits by it.

### Relationship to the reference implementation

Worth noting for review, because it is how the drain-order defect above was found. `buildbarn/bb-storage` solves the same problem with the same structures: its `OldCurrentNewLocationBlobMap` is the same three-band ring, its `Get` returns `needsRefresh` exactly when a blob is in the old region, and copy-forward is its remedy too. Its documented rationale for the old band ("The number of blocks in the 'old' group should not be too low, as this would cause this storage backend to become a FIFO instead of being LRU-like") is the framing kura's `total_segments / 5` had encoded without naming, and `architecture.md` now names it.

The structural difference is what caused this bug. There, the completeness checker validates an ActionResult by calling `FindMissing` on the CAS, and `FindMissing` refreshes, so lifetime extension for the AC is a free side effect of the presence check. Kura's gate took a metadata shortcut around that path.

One difference remains and is not closed here: that implementation refreshes inline under its refresh lock and returns only once the copy is done, so it has no queueing window at all. See the follow-up below.

### Scope boundaries

- This narrows the population where blobs are evicted **between check and use**. It does not address associations the remote never held; #12408 already ends those by declining to serve them.
- It does not remove the need for the upstream fix in `swiftlang/swift-build` (applying the `enableStrictCASErrors` guard at the `ClangCachingMaterializeKeyTaskAction` site). That remains the only change making a missing object non-fatal regardless of cause.
- The snapshot serve path is deliberately not covered. It returns early before `fetch_keyvalue_proto` and vouches for thousands of entries per response, so it is a different problem and belongs with the separate snapshot coverage investigation. Its blobs are still extended when the client fetches them, since that read goes through the serving path.
- Adjacent and deliberately left alone: the same proto notes a server "SHOULD assume that more recently used entries are more likely to be used again", while kura's snapshot wire view ranks by version, which is write time rather than use time. That ranking work belongs to the snapshot coverage investigation.

### Known remaining gap and follow-up

Extension is best-effort. A crash, a `Critical` pressure window, or a promotion queue that cannot drain fast enough can still drop a blob the node vouched for. `architecture.md` names all three.

Closing the asynchrony gap properly means batching the promotion drain's durability the way the backfill apply path already does. Promotion currently pays a per-refresh fsync twice, on the segment append (`ApplyDurability::Sync`) and on the manifest commit (`write_batch_sync`), which is what made inline refresh untenable in #11719 (an 800KB batch read took 15 seconds). Group-committing a batch of promotions would cut the dominant cost by roughly the batch factor and make inline or bounded-wait refresh for vouched triggers affordable.

That is deliberately not in this PR: it rewrites a durability-critical path, it changes serve-path promotion too, and it wants `kura_promotion_drops_total` data first to show whether the remaining window is actually being hit and at what volume.

There is also a thrash ceiling inherent to the technique. Eviction is driven by write volume and promotion is writes, so `append_reader_to_segment` returns `evicted_segments` that promotion itself caused. When the hot set exceeds the ring this is a feedback loop: promoting A evicts B, reading B promotes it and evicts A. The trigger-split byte counters are what would show it. If it appears, admitting read-triggered promotion against segment ring headroom rather than the memory pressure tier is the fix that matches the actual failure, since thrash is a capacity condition that can occur at `Normal` memory pressure.

### How to test locally

From `kura/`:

```bash
mise exec -- cargo test --lib
```

New tests: `extending_lifetimes_queues_only_the_blobs_sitting_in_old_segments`, `extending_lifetimes_is_a_no_op_while_no_segment_has_aged`, `vouching_upgrades_a_promotion_already_queued_by_serving`, `constrained_pressure_keeps_vouched_refreshes_but_drops_serve_promotions`, `critical_pressure_drops_vouched_refreshes_too`, `a_serve_flood_cannot_crowd_out_vouched_refreshes`, `vouched_refreshes_drain_ahead_of_a_serve_backlog`, `the_presence_check_extends_a_blob_lifetime_in_the_same_lookup`, and `a_served_entry_reports_every_blob_whose_lifetime_must_be_extended` on the REAPI side.

### Validation run

- `cargo test --lib`: 725 passed, 0 failed, 1 ignored.
- `mise run clippy`: clean.
- `mise run format -- --check`: clean.
- Falsifiability check on the original change: each mechanism was neutered in turn (extension enqueue, the pressure gate selector, and the collected presence set) and the four positive tests fail in each case, so they are not trivially green. The fifth is a negative assertion and correctly still passes when everything is a no-op.
- Response bytes and headers are unchanged, so no `spec/e2e` shellspec coverage needed updating, and no config knob was added, so no `ops/` values changed beyond the dashboard label.


kura/src/metrics.rs
```diff
@@ -1236,6 +1238,11 @@ impl Metrics {
             "Background segment promotions that failed (the artifact stays in its Old segment and may be reclaimed with it)",
             promotion_failures.clone(),
         );
+        registry.register(
+            "kura_promotion_drops_total",
```
esnunes, line 1164: Preference, take it or leave it. The `kura_promotion_drops_total` HELP text runs into "a nonzero find_missing/action_cache rate means read-triggered lifetime extension is not keeping up and vouched blobs may be evicted before the client's follow-up read". That's the most useful sentence in the change, but it lands better in the panel description or the runbook where an operator meets it — HELP tends to read best as just what the number counts. Most of the neighbours in this file keep to that ("Segment refreshes by producer, result, and the trigger that drove them"), though a couple don't, so there's precedent both ways.

esnunes (approved):
This is a genuinely nice change to read. The copy-forward-versus-reservation reasoning is argued rather than asserted, the asynchrony gap is stated up front instead of buried, and the tests are falsifiable — neutering each mechanism and checking the positive cases fail is the part most PRs skip. Reusing the existing promotion worker and `segment_refresh_lock` so no new knob, config, or Helm surface appears is the right instinct too.

Approving. One small preference left inline on the new metric's HELP text — take it or leave it.

kura/src/metrics.rs
fortmarek: Taken. HELP trimmed to "Promotions dropped for lack of queue room, by the trigger that queued them", matching the terse neighbours. The guidance was not lost: `architecture.md` already carried it, and the metric now has a Promotion Drops panel (it had none) whose description explains why a `find_missing`/`action_cache` drop matters while a `serve` drop is benign.

PR #12435, opened 2026-08-18
feat(server): show and rank test case durations by a real statistic

fortmarek:
> Reported internally against the Test Cases dashboard; no tracking issue was opened, so there is nothing to resolve here.

## What changed

The Test Cases table now shows the duration distribution of every test case, p50 / p90 / p99 / average side by side, over the listing's own 14-day active window and scoped to the selected environment. Each column sorts on its own statistic.

Test Cases table with p50, p90, p99 and average duration columns, sorted by p99

It replaces a single column reading `test_cases.avg_duration`: a denormalized arithmetic mean of the last 50 runs, written at ingestion in `Tuist.Tests.create_test_cases/4`. Every run is appended regardless of environment, branch, or who ran it, so it has no time bound, no environment split, and no outlier resistance. It both displayed the number and decided the ranking.

Confirmed on production. Test case `477831c4-8070-4d8a-a532-da5b56a3c08d`, project 2289:

```
recent_durations = [978, 1022, 1040, 196101, 2, 2, 2, 4676155]  (ms, n=8)
avg_duration     = 609412 ms (10m09s)
median           = 978 ms
```

A 623x inflation from one 78-minute run, almost certainly a paused debugger, which persists until 50 newer runs evict it. All eight runs are local, five of them on a single feature branch, yet selecting "Local" still showed 10 minutes and selecting "CI" dropped the row entirely. Ordering that project by `avg_duration desc` over 30 days put that row and `test_whenNoEnv_fetchSucceeds()` (38.1s avg vs 1.43s median, 27x) near the top of "slowest". A test case with one recorded sample sorted high on that one sample.

The first row in the screenshot above is a seeded reproduction of exactly that shape.

## Why four columns rather than one

This started as a single column with a statistic picker, first in the table toolbar and then as a dropdown on the column header. Reviewing it against real data made the case for showing all four at once, and the reason is visible in the screenshot:

```
testPausedDebugger    1.0s     23m 23s   1h 12m 28s   9m 45s
testHeavyTail         503ms    6.9s      8.1s         1.9s
testSteadyDuration    2.0s     2.0s      2.0s         2.0s
```

One of those is a test with a stalled run in its history, one has a genuine tail, one is stable, and no single column separates them. The p50 beside the p99 answers "slow always or slow sometimes" in a glance, and the average beside the p50 shows whether the average can be trusted at all. Behind a picker, that comparison costs a reload; behind a hover, it is invisible.

Each column sorting on its own statistic means "slowest by median" and "worst tail" are two orderings of the same table rather than two modes of it. Headers are clickable directly, matching the Runner Workflows table; the Test Cases table's previous "only clickable once already sorted by it" behaviour was the odd one out and made four columns unusable.

The two intermediate designs are in the branch history rather than squashed away, since the reasoning against them is the reasoning for this one. The header-dropdown attempt also added a `header_action` slot to Noora's table, which this reverts: nothing else used it, and design-system API with no consumer is worth less than the lines it costs.

## The test case page

The detail page showed a bare "Avg. test case run duration" card while the Tests overview already offered the four statistics on its equivalent widget. It now uses the same percentile dropdown.

Summary metrics for a single test case, showing the p90 run duration selected from a dropdown

`Tuist.Tests.Analytics.test_case_analytics_by_id/3` returns the percentiles from the same pass over the same rows that already produced the mean, so the whole distribution costs what the average cost: no new table, no second query.

## How it works

A new migration adds `test_case_duration_daily_stats_per_case`, an AggregatingMergeTree keyed on `(project_id, test_case_id, date, is_ci)` carrying `count`, `avgState`, and `quantileState` for p50/p90/p99, plus its materialized view and a partition-by-partition backfill. `Tuist.Tests.list_test_cases/3` left-joins it, aliasing each statistic with `selected_as/2` so Flop can sort on values that live in a different table.

`date` is derived from `ran_at`, matching `test_case_runs_active_daily_stats` and the listing's own active window, so the durations shown cover exactly the runs that decide whether a row appears at all.

## Decisions a reviewer should weigh

**A new table rather than columns on `test_case_run_daily_stats_per_case`.** The plan called for extending that table and adding `is_ci` to its sort key. Doing so would change its aggregation grain: every existing row would silently claim `is_ci = false`, and the flaky and reliability automations that read its count states would then see counts and durations keyed on different grains in one table. A separate table is additive, rolls back with a `DROP`, and keeps the backfill off a table with known OOM history.

**The sort key leads with `test_case_id`, not `date`.** The only read is "every test case in this project over the trailing window". Grouping by `test_case_id` then runs in sorted order, so ClickHouse streams one group at a time rather than holding a quantile state per test case for the length of the scan, which for a large project is the whole active suite at once. Monthly partitioning keeps the window filter cheap without putting `date` in front. The backfill keeps the same conservative shape as `20260609121000_add_reliability_aggregates_to_test_case_run_stats.exs` (newest partition first, one project per chunk, `optimize_aggregation_in_order` / `max_threads = 1` / a generous ceiling, a throttle between chunks, retry on `TABLE_IS_READ_ONLY` and `MEMORY_LIMIT_EXCEEDED`), since quantile states are larger than the count states that died at ~3.74 GiB last time.

**A five-run floor before durations are shown or ranked.** A percentile over a handful of samples would trade one misleading number for another. Below the floor every duration cell reads N/A and the values are `NULL`, so `:asc_nulls_last` / `:desc_nulls_last` keeps the row at the bottom whichever column and direction is chosen. Not knowing how long a test takes is not the same as knowing it is quick, so those rows belong out of both "slowest" and "fastest" rather than at one end. The explanation rides the p50 column alone: all four cells in such a row are empty for the same reason, and four tooltips per row would mount four hooks to say it once.

**Reading the durations is opt-in.** It costs an aggregate read across the project's whole active suite. Callers that neither pass `:with_durations` nor order by one of the duration fields (the public API and the MCP tool, both of which sort by `last_ran_at` and expose `avg_duration` by name) run exactly the query they ran before.

**Four merges, one scan.** All four statistics come off the groups the scan already builds, so reading the distribution costs about what reading one statistic did.

## Reconciling the two averages

The list page's number was "mean of the last 50 runs, no time bound" while the drill-down's was a live mean over the date-picker window: one label, two definitions, two numbers. Both are now live aggregates over actual runs in a stated window. They still differ by window and by environment scope, but both of those are visible controls on their respective pages, and the columns name their statistics.

## Knock-on change

Sorting by values from another table needs Flop alias fields, which need virtual fields on the `TestCase` schema, which broke test call sites passing `Map.from_struct/1` output straight to `insert_all`. All 93 occurrences now go through `TuistTestSupport.Utilities.insertable_attrs/1`, which takes only the schema's real fields. That fixes the pattern generically rather than per-site, and closes the same trap for any future virtual field on any schema.

## Out of scope

Branch filtering is excluded: branch is a cardinality multiplier on the sort key and needs its own design. A duration-based alert monitor depends on this and comes after.

## How to test locally

Point at a project with enough history, or seed one, then:

```
mise run dev
```

Open `/{account}/{project}/tests/test-cases`. The table should show p50, p90, p99 and Avg. columns. Click each header and confirm the ordering changes to match that statistic, and that a test case with a skewed distribution moves between them. Switch the Any/CI/Local dropdown and confirm the durations change, not just which rows appear. Find a test case with fewer than five runs in the last 14 days and confirm every duration cell reads N/A, that the tooltip on p50 gives its actual run count, and that it sorts to the bottom under both "slowest first" and "fastest first". Then open a test case and switch the duration widget between avg, p50, p90 and p99.

Automated:

```
mix test test/tuist/tests_test.exs --only describe:"list_test_cases/2 durations"
```

Eight tests cover outlier resistance (100 ms p50 vs 125s mean on the same eight samples), the heavy-tail shape, the sample floor, the CI/local split, unranked rows sorting last in both directions, per-column ordering, and the opt-in contract.

## Validation run

`mix format`, `mix credo` (no issues), and `mix excellent_migrations.check_safety` (no findings for the new migration) are clean. `.pot` regenerated with `mix gettext.extract`; no `.po` touched. `server/data-export.md` documents the new table, and there is a changelog entry with both screenshots at `server/priv/marketing/changelog/2026.08.18-test-case-duration-percentiles.md`.

473 passed across `tests_test.exs`, `tests/analytics_test.exs`, the Test Cases / Tests / test case detail LiveView tests, and `mcp`, with no failures.

An earlier, wider run over the same suites plus `automations`, `webhooks/dispatcher_test.exs`, and the Flaky / Quarantined LiveView tests returned 635/636. The one failure was `test "claims the shard mapping before writing the run row"` (`server/test/tuist/tests_test.exs:2255`), which passes in isolation and in narrower runs. Its helper `capture_clickhouse_queries/1` attaches a global `:telemetry` handler, so concurrent async tests' inserts land in the captured list and shift the indices it asserts on. Pre-existing and unrelated to this change; being fixed separately.


pepicrft (approved):

server/lib/tuist_web/live/tests_live.ex
```diff
@@ -354,14 +354,24 @@ defmodule TuistWeb.TestsLive do
...
+          %{
+            page: 1,
+            page_size: 5,
+            order_by: [:duration_p50, :id],
+            order_directions: [:desc_nulls_last, :asc]
+          },
+          preload: [:duration_p50_ms]
+        )
+
+      slowest_test_cases = Enum.reject(slowest_test_cases, &is_nil(&1.duration_p50_ms))
```
esnunes, line 374: One thing I wanted to flag about the "Slowest test cases" card. Rejecting the rows whose p50 is nil makes the list empty for any project where no test case cleared the five-run floor in the window — not just for a project with no test cases. When the most-flaky list is empty too (common), the shared empty state at `tests_live.html.heex:1152` fires, and that one says "No test cases yet" with a get-started link to the docs. So a project that has been reporting tests for months but ran its suite four times this fortnight would be told it has none.

Would it be worth giving the slowest section its own empty state ("Not enough runs to rank durations yet"), or gating the shared onboarding state on something that actually means "no test cases"? The rejection itself reads right to me — it's just the branch it lands in that changed meaning.

server/lib/tuist/tests.ex
```diff
@@ -2440,12 +2487,16 @@ defmodule Tuist.Tests do
...
+    # The duration join is added after the count: the count only needs the
+    # filtered `test_cases` rows, and the aggregate scan behind the join would
+    # be pure overhead there.
     total_count = test_cases_count(base_query, flop, query_settings)
 
     case state_filter_mode do
       :joined ->
         base_query
         |> select_resolved_test_case_state()
+        |> select_durations(project_id, duration_fields, is_ci)
```
esnunes, line 2499: Question about the `:joined` branch here. It keeps passing `@test_case_state_join_settings` (512 MiB, grace-hash, external group-by), and the comment on those settings reasons about the state join being the dominant cost and defers relaxing the ceiling until someone measures the real per-project peak. That was written for a query with one join; this now adds the duration subquery — a second aggregation with four quantile merges over the project's whole active suite — under the same ceiling.

What makes me want to ask rather than assume: `state_filter_mode/2` only picks this branch above 10,000 test cases with state rows, so the extra work lands specifically on the biggest projects, on the Test Cases page with a Flaky/Muted filter on. Grace-hash and the external group-by do give it somewhere to spill, so this may well be fine. Did the 512 MiB get re-checked with the duration join in place, or should the deferred measurement note be widened to cover both?

server/test/tuist/tests_test.exs
```diff
@@ -3393,6 +3393,231 @@ defmodule Tuist.TestsTest do
     end
   end
 
+  describe "list_test_cases/2 durations" do
```
esnunes, line 3396: These duration tests are good — outlier resistance, heavy tail, the correction double-count, the floor, the CI/local split, nulls-last both ways, per-column ordering, the opt-in contract. One gap I noticed: they all run on fresh projects, which have no `test_case_current_states` rows, so `state_filter_mode/2` returns `:preloaded` every time and the `:joined` SQL (`test_cases` FINAL + state join + duration subquery, two joins in one statement) is never executed.

I forced that branch locally by inserting 10,001 `test_case_states` rows before the listing call and it returns the right p50, so the SQL is fine as written — it's just unpinned, and it's the branch that only ever runs on the largest projects. Would you be up for one test that seeds past `@max_preloaded_test_case_states` and asserts a duration? The seeding is a single chunked `insert_all` and it ran in under a second here.

server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
```diff
@@ -0,0 +1,256 @@
...
+  defp backfill_boundary do
+    {:ok, %{rows: [[boundary]]}} = IngestRepo.query("SELECT now64(6)")
+    Logger.info("Backfilling test_case_duration_daily_stats_per_case up to #{boundary}")
+    boundary
+  end
+
+  # Newest partition first. The listing only reads a trailing two-week window,
+  # so the feature is correct as soon as the current partition lands; an older
+  # partition that needs several retries delays nothing user-visible.
+  defp backfill_by_partition(boundary) do
```
esnunes, line 146: The backfill is carefully built — boundary read from ClickHouse, newest partition first, one project per chunk, generous ceiling, retries on the two transient errors. Two things about its scope that I wanted to raise, since they're both about the same mismatch: the table is written far wider than it is read.

`backfill_by_partition/1` walks every active partition of `test_case_runs`, and the only reader filters `date >= today - 14`. Anything older than the two most recent months is computed at deploy time (synchronously, in the migrate hook) and then stored with no consumer. Would bounding the backfill to the partitions the active window can reach be safe here, or is there a future reader that wants the history?

Second, the `CREATE TABLE` has no TTL, so the table keeps growing forever at a per (project, test case, day, is_ci) grain with three reservoir states per row — a much wider grain than `test_case_runs_daily_stats`, which is the per-project/day precedent for `quantileState` here. Several ClickHouse tables in this repo carry `TTL … + INTERVAL n DAY`; would one be appropriate on this one?

server/lib/tuist_web/live/test_cases_live.html.heex
```diff
@@ -609,21 +613,67 @@
...
           <:col
             :let={test_case}
-            label={dgettext("dashboard_tests", "Avg. duration")}
-            patch={
-              @test_cases_sort_by == "avg_duration" &&
-                column_sort_patch(assigns, "avg_duration")
-            }
-            sort_order={@test_cases_sort_by == "avg_duration" && @test_cases_sort_order}
+            label={duration_field_label(:duration_p50)}
+            patch={column_sort_patch(assigns, "duration_p50")}
```
esnunes, line 617: Small one: the four duration headers are now directly clickable (`patch={column_sort_patch(assigns, "duration_p50")}` and friends), but "Test Case" (line 569) and "Last ran at" (line 699) keep the old `@test_cases_sort_by == "…" && column_sort_patch(…)` guard, so they're still only clickable once you're already sorted by them. The PR description calls that behaviour "the odd one out", which read to me like it was meant to go — was leaving it on those two intentional? Dropping the guard there too would make the whole header row behave one way. Definitely not a blocker.

esnunes (approved):
This is a really well-argued change, and the parts that are easiest to get wrong are the parts that are most carefully reasoned. `uniqExactState(id)` for the run count rather than `countState()`, with the flaky-correction re-insert spelled out in the moduledoc; the boundary timestamp read from ClickHouse before the view is created, and the explicit choice of which side of that race to land on; `NULL` below the sample floor plus nulls-last in both directions instead of a sentinel — those all read like someone thought about the failure case first. I checked the aggregate types against `test_case_runs` (`duration` is Int32, matching the declared state types) and the backfill's `toYYYYMM(inserted_at)` chunking against the source table's actual partition key; both line up. The ten new duration tests pass locally.

Most of what I have is questions rather than defects. Two are about scope rather than correctness: the `:joined` listing branch picks up a second aggregation while keeping a memory budget whose comment reasons only about the first one, and it's the branch that only runs on the largest projects; and the new table is backfilled across all history and never pruned, while the only reader ever looks at 14 days. One is a real user-visible edge: the "Slowest test cases" rejection can empty that card for a project that simply doesn't run its tests five times a fortnight, and the shared empty state it falls through to says "No test cases yet" with an onboarding link. The rest are small — an untested branch, and two header interaction models in one table.

None of it blocks. Happy to be told the memory and backfill questions are already covered by the deferred-measurement note.

PR #12441, opened 2026-08-18
feat(server): keep the Swift registry mirror covering the catalog under GitHub rate limits, and gate archive-replacing repairs

### LINKED ISSUE #12204: Registry: the sync exceeds its GitHub API quota every hour, and rate-limited jobs are discarded rather than retried
The registry sync consumes more GitHub API quota per hour than its token provides, so the budget is exhausted continuously rather than occasionally. Measured directly against production.

## The arithmetic

```
@swift_registry_sync_cron  {"*/10 * * * *", SyncWorker}   →  6 ticks/hour
swift_registry_sync_limit  default 1_000                  →  1,000 packages/tick
production override        none set
```

**1,000 packages × 6 ticks = 6,000 `list_tags` calls per hour, against a 5,000/hour personal access token.**

That is a floor, not the total. It counts one call per package and ignores pagination — packages with hundreds of tags cost several — and it excludes all `ReleaseWorker` work, where each version needs a contents listing, one or more manifest fetches, and a zipball.

## Confirmed exhausted, not misattributed

Two earlier readings looked contradictory: `/rate_limit` reported thousands remaining while jobs failed with `{:rate_limited, 403}`. Probing the failing endpoint directly settles it:

```
status: 403
x-ratelimit-limit:     5000
x-ratelimit-remaining: 0
x-ratelimit-used:      5000
x-ratelimit-resource:  core
retry-after:           (absent)
```

`limit: 5000` means the token is being honoured — this is not requests going out unauthenticated and hitting the 60/hour per-IP limit, which was the competing hypothesis. It is the primary quota, genuinely spent. The earlier healthy readings were simply taken at a different point in the hourly window.

## Impact

Discards attributable to this go back to at least 2026-07-31 20:50, before any repair work, so this is the steady state rather than something the incident introduced.

The consequence is not just delay. **`maybe_skip_release/5` discards on a rate limit rather than snoozing**, and a discarded Oban job never retries — so every version in flight when the budget runs out is silently dropped until someone re-enqueues it by hand. During the #12191 repair this killed 51 jobs in one pass; the same thing happens to ordinary catalog syncing, where nobody is watching for it.

That is the mechanism by which the mirror quietly falls behind upstream, and it is invisible because a discarded job produces no ongoing signal.

## Options

1. **Lower the batch size.** `syncLimit: 600` → 3,600 calls/hour, ~28% headroom. One line in production values, effective on the next deploy. Full-catalog coverage stretches from roughly 2 hours to 3.3 hours per pass, which is immaterial for a mirror.
2. **Lengthen the interval** to `*/20` — same arithmetic, same trade-off, halves consumption at the current batch size.
3. **Move to a GitHub App installation token** (15,000/hour). The durable answer if the current cadence is wanted, and the only option that leaves real headroom for repairs and bursts.

## Recommended

Take option 1 now to unblock work, and treat option 3 as the actual fix.

Independently of which, **snooze instead of discard on a rate limit**. `retry-after` is absent here, but `x-ratelimit-reset` gives the wake-up time, so the job can reschedule itself rather than being thrown away. Without that, every future exhaustion silently strands whatever was in flight, and the only way to notice is to go looking — which is how this was found.

Context: #12191.

pepicrft:
Resolves <https://github.com/tuist/tuist/issues/12204>

Addresses the four items marked `immediate` in [postmortem 1](https://hive.tuist.dev/postmortems/1), covering the Swift Package Registry incidents of July and August 2026.

## What changed

- **The production catalog pass is capped below the GitHub request budget.** The release default drops from 1,000 packages per pass to 600, and production and staging set it explicitly. A limit that cannot be honoured now fails at boot rather than being silently clamped.
- **Throttled synchronization work is deferred, not dropped.** The shared GitHub client reports throttling as `{:rate_limited, status, retry_after_seconds}`, derived from either the `retry-after` header or the `x-ratelimit-reset` timestamp. Both the catalog and release workers snooze to that reset instead of discarding the job.
- **The rotation cursor advances only over packages a pass actually visited.** A pass that stops early no longer jumps the full batch size past packages it never looked at.
- **Lost coverage is now a signal.** A new metrics plugin counts deferred passes, deferred packages, packages passed over without being read, and deferred releases, each labelled by reason. Three alert rules are documented in `infra/helm/k8s-monitoring/alerts.md`.
- **Synchronization authenticates as a GitHub App installation.** The mirror resolves a short-lived installation token through the existing `Tuist.GitHub.App`, falling back to the personal access token. Enabled in every managed environment. Production's installation budget is 12,500 requests an hour against 5,000 for a user token.
- **A pass in which every package failed is read as one systemic failure**, not as many independent ones. It defers and holds the cursor instead of reporting success.
- **Bulk repairs are gated.** A new `Tuist.Registry.Swift.Repair` module is the only supported way to run a repair that can replace published archive bytes, with a dry run, a checksum-change inventory, a bounded batch, an approval digest, an abort threshold, a per-archive backup, and a rollback.
- **`Purge.purge_package/3` requires a confirmation** naming the package it is about to delete.

## Why

Between roughly 23 July and 4 August 2026 the registry served missing versions, changed checksums, and archives that could not be extracted. Two operational facts from that incident drive this change.

The first is that the synchronization pod can be healthy and not be doing its job. On 11 August it was Ready with zero restarts while logging 2,566 GitHub rate-limit failures and dropping nine consecutive scheduled catalog passes in a little over six hours. Nothing paged. Every existing registry signal answered "is the service up", and none answered "is the catalog current".

The second is that recovery is dangerous. A published version's bytes are immutable from a client's point of view, because [Swift Package Manager](https://www.swift.org/package-manager/) records a trust-on-first-use fingerprint the first time it resolves a version. Rebuilding an existing version is therefore a content change, not a neutral precaution. The July recovery treated it as one and changed 22,113 checksums across 1,643 packages, and because the storage bucket had no object versioning, the previously trusted bytes were unrecoverable.

## Root cause

Three separate defects, one per symptom.

The batch size exceeded what a single pass could spend. Each package costs at least one tag-listing request plus one per extra page of tags, so a 1,000-package pass every ten minutes exhausted the hourly quota. Production never overrode the default, so it ran the largest possible batch continuously.

Throttled work was discarded rather than retried. Detection was correct, but nothing scheduled around it: the rate-limit case was folded into the transient-error path, which returns `{:discard, reason}`. In Oban a discard drops the job permanently, so the arguments were lost. The quota reset the response had just reported went unused. Meanwhile the cursor advanced by the full batch size regardless of how far the pass actually got, so every package it never reached waited an entire catalog rotation for another look, with nothing recording that it had been skipped.

Repairs had no controls. The checksum-change override was a plain job argument. There was no dry run, no bounded batch, no approval step, no inventory of what would change, no abort threshold, and nothing to roll back to.

## Approach

**Deferral over discard.** `{:snooze, seconds}` is the right primitive because it preserves the job's arguments, and because Oban raises `max_attempts` on a snooze, deferring does not consume the retry budget. The delay is sized to the reset the response reported and clamped to a sensible window, so a reset that has already elapsed cannot turn into an immediate replay against a budget that has not recovered.

**Distinguishing what should halt a pass from what should not.** An authentication failure halts it: the credential is not the repository, so skipping package by package would advance the cursor over the entire batch and report a clean run. A forbidden response does not halt it, because GitHub uses that for repositories that are individually unavailable and halting on one would stall the rotation indefinitely. A package whose lock is held by a release worker also does not halt it, since that contention is ordinary and sub-second, and holding the cursor there would let one chronically busy package starve the rest of the catalog. That last case is counted rather than silent, which is the actual lesson from the incident: the problem was never that work was skipped, it was that nobody could tell.

**An installation named by organization, not by number.** The credential change could have taken a numeric installation identifier, but that is an opaque value somebody has to go and look up, and a value nobody can read is a value nobody sets. Naming the organization and resolving it through the existing cached lookup means the setting is self-explanatory in the chart.

**Measuring the credential change rather than staging around it.** The mirror reads roughly ten thousand arbitrary public repositories, so the whole approach rests on an installation token being able to read a repository outside its own installation. GitHub's documentation does not answer that: it describes the scope limit for private data and says nothing about public data. Rather than hedge with a slow rollout, this was verified directly against both the canary and production Apps, across every endpoint the mirror uses — tag listing, repository contents, single-file reads, zipball download, and git clone credentials — against a public repository outside the installation. All succeed, even though `contents` is not among the granted permissions, because public data is served to any valid credential regardless of scope. The same probe measured the budgets: 12,500 requests an hour for the production installation against 5,000 for a user token.

**Failing loudly if that ever stops holding.** GitHub answers a repository a credential cannot see with 404 rather than 401, so a credential regression does not arrive as an authentication error. It arrives as every package in a pass failing independently, which each individually survivable failure would happily absorb into a clean-looking run. Reading a wholly failed pass as one systemic failure closes that, and is worth having regardless of this cutover.

**A repair-path backup rather than waiting for storage-level immutability.** Making ordinary synchronization writes physically unable to replace a published version is the durable fix, but it is a storage-layer change tracked separately. Copying each archive to a backup key before replacing it makes the repair path reversible now, which is what the postmortem asked for, and `restore/4` puts both the bytes and the catalog entry back.

**Fencing the rollback.** Restore cancels outstanding repair jobs for the version before touching anything. A job that snoozed on a rate limit still carries the checksum-change override and would otherwise republish over the rollback once the quota reset, hours later. A rollback a queued job can silently undo is not a rollback.

## Impact

Production behaviour changes in three ways on deploy. The catalog pass covers 600 packages instead of 1,000, so a full rotation takes longer but fits inside the request budget. Throttled passes and releases are retried after the reset instead of being lost, so coverage recovers on its own. Three new alert rules need creating in Grafana from the queries documented in `infra/helm/k8s-monitoring/alerts.md`; until they exist the metrics are emitted but nothing pages on them.

All three managed environments switch to GitHub App authentication, which moves registry synchronization off a shared user budget onto a dedicated one and roughly doubles the headroom in production. If the App cannot issue a token the mirror logs and falls back to the personal access token, which stays configured, so the change is revertible by removing one chart value.

`Purge.purge_package/2` becomes `purge_package/3` and requires a confirmation argument. It is an operator console function with no callers in the codebase, so this affects runbooks rather than code.

Deliberately not included, and still open: deriving the batch size from live remaining quota rather than a fixed number, and making ordinary synchronization writes unable to replace a published version at the storage layer.

## Validation

- `mix test test/tuist/registry/ test/tuist/registry_test.exs` in `server/` — 103 passed, covering the new deferral, cursor, coverage-metric, systemic-failure, repair, restore, and copy-verification behaviour.
- `mix test test/tuist_common/github_test.exs` in `tuist_common/` — 31 passed.
- `mix credo` — no issues. `mix format --check-formatted` — clean.
- `helm template` against the production, staging, canary, and preview overlays the way `.github/workflows/helm.yml` renders them — all four render, with `SWIFT_REGISTRY_SYNC_LIMIT: "600"` on production and staging and `SWIFT_REGISTRY_GITHUB_APP_INSTALLATION: "tuist"` on all three managed environments.
- The GitHub App credential path was probed against the real canary and production Apps before enabling it, confirming that an installation token reads tags, contents, zipballs and git clone credentials for a public repository outside its installation, and measuring the installation budgets at 5,000 and 12,500 requests an hour respectively.
- The branch was reviewed adversarially with a second model at high reasoning effort. Seven findings were raised and all seven are fixed in this branch, the most significant being that the repair approval digest did not cover the abort threshold, that restore did not fence in-flight repair jobs, and that a package whose lock was held was counted as visited.

The full server suite could not be run locally: every database-backed test fails in this worktree because the shared local ClickHouse instance is running without Keeper, which the test sandbox needs for transactions. That reproduces on `main` and is unrelated to this branch.

### How to test locally

The registry writer paths are unit-tested and need no running registry:

```
cd server && mix test test/tuist/registry/
cd tuist_common && mix test test/tuist_common/github_test.exs
```

To check the rendered configuration for each environment:

```
  -f infra/helm/tuist/values-ci.yaml \
  | grep -A1 SWIFT_REGISTRY_SYNC_LIMIT
```

The repair path is an operator console workflow rather than something to exercise locally. Its shape, and the reasoning behind each gate, is documented in `server/lib/tuist/registry/AGENTS.md`.


server/lib/tuist/registry/swift/repair.ex
```diff
@@ -0,0 +1,518 @@
...
+    end
+  end
+
+  defp back_up_archive(%{published_checksum: nil}), do: :ok
+
+  defp back_up_archive(target) do
+    checksum = target.published_checksum
+    destination = backup_archive_key(target, checksum)
+
+    with :ok <- S3.copy_object(archive_key(target), destination),
```
fortmarek, line 418: **[P1] Preserve the first backup across repair retries**

This destination is derived only from the catalog checksum and `copy_object/2` overwrites it unconditionally. If a rebuild replaces the archive but fails before the catalog write, the natural re-plan/retry copies the new bytes over the sole pre-repair backup at the same key. Because the bucket has no versioning, the original becomes unrecoverable. Please make backups immutable—for example, key them by a plan or actual-object digest, or refuse to overwrite an existing destination.

server/lib/tuist/registry/swift/repair.ex
```diff
@@ -0,0 +1,518 @@
...
+  version are cancelled, because a job that snoozed on a rate limit carries
+  `allow_checksum_change: true` and would republish the rebuilt bytes over the
+  restore hours later. And everything the restore needs is read first, so a
+  missing backup or an unreadable catalog fails before the archive is replaced
+  rather than halfway through.
+  """
+  def restore(repository_full_handle, version, checksum, _opts \\ []) do
+    with {:ok, target} <- normalize_target({repository_full_handle, version}),
+         {:ok, cancelled} <- cancel_repair_jobs(target),
+         {:ok, metadata} <- Metadata.get_package(target.scope, target.name),
```
fortmarek, line 189: **[P1] Serialize restore with package metadata writers**

`restore/4` reads and later replaces the entire package metadata object without taking the `{:package, scope, name}` lock used by release publication. If a release worker publishes another version between this read and write, the restore can discard that version; the inverse interleaving can also undo the restored checksum. Please hold the package lock across the metadata read-modify-write.

server/lib/tuist/registry/swift/repair.ex
```diff
@@ -0,0 +1,518 @@
...
+
+  # Covers the identity of every target, the checksum it was planned against,
+  # and the abort threshold, so a digest only matches a plan that was actually
+  # read, that still describes the same content, and whose threshold has not
+  # been widened. Leaving the threshold out made the abort trivially bypassable:
+  # raising `max_checksum_changes` on the returned map kept the digest valid.
+  defp approval_digest(targets, max_checksum_changes) do
+    targets
+    |> Enum.map_join("\n", fn target ->
+      "#{target.scope}/#{target.name}@#{target.version}:#{target.status}:#{target.published_checksum}"
```
fortmarek: **[P1] Bind the repository handle to the approval**

The approval digest omits `repository_full_handle`, but `repair_target/1` executes against `target.repository_full_handle`. Editing only that field therefore preserves the approval while making the worker resolve and publish a different catalog package, and the backup was taken for the original target. Please include the normalized handle in the signed digest (or derive it from signed data at execution time).

server/lib/tuist/registry/swift/repair.ex
```diff
@@ -0,0 +1,518 @@
...
+  # blast radius are both unknown, so the whole plan stops rather than acting on
+  # the ones that happened to read cleanly.
+  defp ensure_inspectable(%{counts: %{uninspectable: 0}}), do: :ok
+
+  defp ensure_inspectable(%{counts: %{uninspectable: count}}), do: {:error, {:uninspectable_targets, count}}
+
+  # Reads the threshold rather than pattern-matching it, so a plan missing the
+  # key falls back to the default instead of matching a catch-all clause and
+  # skipping the gate entirely.
+  defp ensure_within_checksum_change_threshold(%{counts: %{unresolvable: count}} = plan) do
```
fortmarek: **[P2] Recompute checksum-change counts from signed targets**

This safety gate trusts `plan.counts.unresolvable`, but the counts are neither included in the approval digest nor recomputed from the signed targets. The count can be changed to zero without invalidating approval, bypassing the checksum-change threshold. Please derive this value from the target statuses when applying the plan, or include and validate the counts in the digest.

fortmarek (commented):
Four actionable repair-safety findings follow. The earlier GitHub App failure finding is already addressed by the latest all-packages-failed handling.

server/lib/tuist/registry/swift/repair.ex
pepicrft: Good catch, and the sequence is real rather than theoretical. `ReleaseWorker.sync_release/7` runs `upload_source_archive` before `update_metadata_with_release`, with three failure points in between (read-back verification, manifest upload, a contended metadata lock), so a rebuild really can leave new bytes under an unchanged catalog checksum. The obvious re-plan then classifies the version as `:unresolvable` and would have copied those bytes over the only remaining copy of the pre-repair ones, at the same key.

Fixed in 1ea93639c3. `back_up_archive/1` now HEADs the destination first and keeps an existing backup rather than overwriting it. Keeping the first one is also right on its merits, not just safe: it is the state the version was in before any repair touched it, which is what a rollback wants.

I went with refuse-to-overwrite rather than keying by object digest because it keeps `restore/4`'s lookup addressable by the checksum an operator can read straight off the catalog, and immutability is what the concern actually needs.

Covered by `keeps the first backup when one already exists for the published checksum`, which rejects `copy_object/2` outright.

server/lib/tuist/registry/swift/repair.ex
pepicrft: Agreed, and this one was the worst of the four. `ReleaseWorker` takes `{:package, scope, name}` for its catalog write (release_worker.ex:1023) and `restore/4` took it zero times, so the pre-flight read and the write were separated by an archive copy that is not quick for a large package. Writing back the older document would have dropped anything published in that window. That is a stale full-document overwrite, which is the first root cause in the postmortem this branch is addressing, so it had no business reappearing in the rollback path.

Fixed in 1ea93639c3. The catalog write moved into `write_restored_checksum/2`, which takes the same lock with the same bounded retry the release worker uses and re-reads the metadata inside it, so the pre-flight copy is never what gets written back.

I deliberately did not wrap the whole restore in the lock: the archive copy would then block release publication for that package for as long as the copy runs, and could outlive the lock's own time to live. The lock now covers the same sub-second read-modify-write the release worker holds it for.

Two tests: `re-reads the catalog inside the package lock rather than writing back the pre-flight copy` interleaves a second version published during the copy and asserts it survives, and `reports the partial state when the package lock cannot be taken` covers contention.

server/lib/tuist/registry/swift/repair.ex
pepicrft: Correct. `repair_target/1` executes against `target.repository_full_handle`, the digest covered only `scope`, `name`, `version`, `status` and `published_checksum`, and the handle is not derivable back from scope and name because `KeyNormalizer.normalize_scope_name/2` is lossy (dots become underscores). So editing only the handle kept the approval valid while pointing the rebuild at a different catalog package than the one `back_up_archive/1` had just backed up.

Fixed in 1ea93639c3 by signing the handle rather than deriving it, since it cannot be recovered from the signed fields. Covered by `refuses a plan whose repository handle was edited after it was approved`.

server/lib/tuist/registry/swift/repair.ex
pepicrft: Confirmed, and it applied to `ensure_inspectable/1` as well as the threshold gate: both read `plan.counts`, which is a convenience summary for a human reading the plan and is not covered by the digest.

Fixed in 1ea93639c3 by deriving both from the signed targets via `count_status/2` rather than adding `:counts` to the digest. Recomputing from signed data seemed better than signing a value that can disagree with the targets it summarizes, and it means the field stays a display aid with no authority. The log line in `apply_plan/2` now reports the same derived number the gate used, so the two cannot drift.

Two tests, one per gate: `counts checksum-changing targets from the signed targets, not the plan's summary` and `counts uninspectable targets from the signed targets, not the plan's summary`, each zeroing the summary and asserting the gate still fires.

server/lib/tuist/registry/swift/repair.ex
```diff
@@ -0,0 +1,620 @@
...
+  end
+
+  defp repair_target(target) do
+    with :ok <- ensure_published_checksum_unchanged(target),
+         :ok <- back_up_archive(target),
+         {:ok, _job} <-
+           Registry.force_resync_swift_package_version(
+             target.repository_full_handle,
+             target.version,
+             allow_checksum_change: target.status == :unresolvable
```
fortmarek: **[P1] Revalidate the live archive before granting the override**

This decision uses the status captured by the dry run, while the apply-time check refreshes only the catalog checksum. If another worker restores the archive to the published checksum between planning and applying, the catalog checksum remains unchanged, so this stale `:unresolvable` status still grants permission to replace a now-valid published archive. Please re-read the stored checksum/classification immediately before backup and enqueue, and skip or fail when the target has become `:published`.

server/lib/tuist/registry/swift/repair.ex
```diff
@@ -0,0 +1,620 @@
...
+  #
+  # The lock is taken here rather than around the whole restore so it covers
+  # only a sub-second read and write, matching what the release worker holds it
+  # for. Holding it across the archive copy would block release publication for
+  # that package for as long as the copy takes, and could outlive the lock's own
+  # time to live.
+  defp write_restored_checksum(target, checksum) do
+    lock_key = {:package, target.scope, target.name}
+
+    case acquire_metadata_lock(lock_key, @metadata_lock_max_attempts) do
```
fortmarek, line 467: **[P1] Fence restore across the archive and catalog writes**

This lock starts only after the archive copy, and `cancel_repair_jobs/1` is a point-in-time snapshot rather than a barrier. An executing `SyncWorker` can enqueue a `ReleaseWorker` after that snapshot; release workers upload the archive outside this metadata lock. Depending on the interleaving, restore can return success with rebuilt bytes plus the restored checksum, or the later worker can undo the rollback. Please serialize the whole restore with repair enqueueing/release execution using a per-version fence that spans cancellation, archive copy, and catalog update.

server/lib/tuist/registry/swift/repair.ex
```diff
@@ -0,0 +1,620 @@
...
+  # condition this backup exists to avoid in the first place.
+  #
+  # Keeping the first backup is the correct choice rather than merely the safe
+  # one: it is the state the version was in before any repair touched it, which
+  # is what a rollback wants.
+  defp back_up_archive(target) do
+    checksum = target.published_checksum
+    destination = backup_archive_key(target, checksum)
+
+    case S3.head_object(destination) do
```
fortmarek, line 403: **[P1] Make backup creation atomic**

The HEAD followed by COPY is still a check-then-act race. Two concurrent applies can both observe a missing destination; after the first copies the original and starts its rebuild, the delayed second copy can overwrite the only backup with rebuilt bytes. With no bucket versioning, that again makes the original unrecoverable. Please use an atomic conditional destination write if available, or serialize this check-and-copy with a per-version backup lock.

server/lib/tuist/registry/swift/repair.ex
```diff
@@ -0,0 +1,620 @@
...
+    * `:published` — the catalog and object storage agree. A rebuild that
+      produces different bytes breaks every client that already resolved it.
+
+  Only `:unresolvable` targets are allowed to change a checksum, and only
+  through the explicit override the release worker requires.
+  """
+  def plan(targets, opts \\ []) when is_list(targets) do
+    with {:ok, normalized} <- normalize_targets(targets) do
+      inspected = Enum.map(normalized, &inspect_target/1)
+      max_checksum_changes = Keyword.get(opts, :max_checksum_changes, @default_max_checksum_changes)
```
fortmarek: **[P2] Reject non-integer checksum thresholds**

This accepts values such as `"10"` or `nil` and signs them into an otherwise valid plan. Elixir's term ordering makes `count > "10"` and `count > nil` false, so either typo disables the abort gate for every batch. Please fail closed unless the threshold is a non-negative integer, ideally both when constructing and applying a plan.

fortmarek (commented):
Re-review of 1ea93639c3. The repository-handle signature and derived-count fixes are sound; four repair-safety gaps remain.

fortmarek (approved):

server/lib/tuist/registry/swift/repair.ex
pepicrft: Agreed, and this is the sharpest of the four: it is the one failure mode where the module actively causes the harm it exists to prevent. `ensure_published_checksum_unchanged/1` re-read only the catalog half of the classification, so an archive restored to the published checksum in between left the catalog checksum unchanged while the version became healthy, and the stale `:unresolvable` carried `allow_checksum_change: true` into replacing a perfectly good published archive.

Fixed in 8fefc1132f. `ensure_published_checksum_unchanged/1` is replaced by `refreshed_target/1`, which re-runs the full `inspect_target/1` (catalog *and* stored archive) at apply time. The override in `attempt_repair/1` is now derived from that fresh reading, never from the plan's. A target that has become `:published` is skipped as `already_consistent`; one that moved to any other unplanned state fails with `{:target_state_moved, planned, current}` rather than being guessed at.

Covered by `skips a target that became healthy between planning and applying`, which returns `"def"` for the stored checksum at plan time and `"abc"` at apply time, and asserts the target is skipped with `copy_object/2` rejected.

Worth noting this fix is only sound because of the fence in your other comment — see my reply there.

server/lib/tuist/registry/swift/repair.ex
pepicrft: Correct on both halves. `ReleaseWorker.perform/1` (release_worker.ex:94) takes `{:release, scope, name, version}` around the whole of `do_sync_release/5`, which covers the archive upload *and* the catalog write, and `restore/4` took nothing comparable — the `{:package, ...}` lock started only after the copy. And the cancellation really was a snapshot: an Oban insert by an already-executing `SyncWorker` is not blocked by anything we hold.

Fixed in 8fefc1132f, using that existing `{:release, ...}` key rather than a new namespace, since it is already the per-version mutex for release execution and `Lock.try_acquire/2` is an `If-None-Match: *` conditional create rather than a read-then-write check. `with_release_fence/2` now wraps the entire restore — cancellation, backup check, archive copy, and catalog update — and `repair_target/1` takes it too, so repairs and rollbacks for a version are mutually exclusive with each other and with release execution. Lock ordering matches the release worker's (`{:release, ...}` then `{:package, ...}`), so there is no deadlock.

For the insert-after-snapshot window specifically, `do_restore/2` now sweeps a second time before releasing the fence. A release enqueued during the restore is blocked from running by the fence and is cancelled by that second pass; anything inserted after the fence drops is a new repair request rather than a leftover of the one being rolled back.

Two tests: `refuses to restore while a release for the version is in flight` and `sweeps for repair jobs again before releasing the fence`.

server/lib/tuist/registry/swift/repair.ex
pepicrft: Right, HEAD-then-COPY was still check-then-act; the previous change only made the overwrite unlikely rather than impossible.

Fixed in 8fefc1132f by the per-version fence described in your restore comment, rather than by a conditional destination write. Two reasons for that choice: `ExAws.S3.put_object_copy/5` builds its headers from a fixed allowlist and does not thread an arbitrary `If-None-Match` through, so the conditional-write route would mean bypassing the library; and the same fence is needed anyway for the classification re-read and the restore, so one mutex covers all three rather than introducing a second mechanism beside it.

Concretely, `repair_target/1` now runs `refreshed_target/1`, `back_up_archive/1` and the enqueue inside `with_release_fence/2` on `{:release, scope, name, version}`, so two concurrent applies for the same version serialize and the second observes the first's backup. The HEAD-then-keep logic stays as the inner guard.

Covered by `does not repair a version while a release for it is in flight`.

server/lib/tuist/registry/swift/repair.ex
pepicrft: Confirmed — `elixir -e 'IO.inspect({3 > "10", 3 > nil})'` gives `{false, false}`, so either typo disables the gate for every batch while leaving a plan that still reads as though it has one.

Fixed in 8fefc1132f at both ends, as you suggested: `plan/2` refuses to build a plan around an invalid threshold, and `ensure_within_checksum_change_threshold/1` re-validates before comparing.

Validating at apply time turned out to be load-bearing rather than belt-and-braces, which I only found while writing the test for it. The digest interpolates the threshold into a string, so `0` and `"0"` hash identically — a string threshold passes approval completely unchallenged, and only the type check stops it. The test asserts exactly that: it builds a plan with `max_checksum_changes: 0`, confirms it is refused with `{:too_many_checksum_changes, 1, 0}`, then flips the threshold to `"0"` and shows the *unmodified* approval still matches while the plan is now refused with `{:invalid_max_checksum_changes, "0"}`.

PR #12444, opened 2026-08-18
feat(server,handbook): retire dormant accounts, record what the logging policy requires, and correct the handbook

pepicrft:
Three security controls existed only as intent. There was no written policy for logging and monitoring, no written policy for data labelling, and nothing anywhere that said what happens to a user ID nobody has signed into for a year. This writes all three down, builds the technical half of the last one so it does not depend on anyone remembering, and fixes the things that writing them down exposed.

The pattern worth knowing before reading the diff: every time a policy was written, checking it against reality found something wrong. Twice the policy was wrong. Twice the code was.

## What changed

**Account inactivity.** `Accounts.touch_last_sign_in/1` records when an account last authenticated, from session login, application programming interface token authentication, Guardian tokens, and remember-me resumption. We had the column and nothing wrote to it, so there was no way to tell an abandoned account from one in daily use.

`Tuist.Accounts.Dormancy` then applies the thresholds the Access control policy now states: disable at 180 days without authenticating, retire the identity at 365. A nightly job runs it on hosted only, and `sweep/1` returns and logs exactly what it actioned.

One migration, adding a nullable `disabled_at` to `users` with no default, so it records in table metadata without rewriting rows.

**Logging.** A Logging and monitoring policy covering the three activity levels, the minimum detail each record carries, review frequency keyed to system risk and information classification, and a severity model for anomalies. An appendix records where each of those events is actually held, written from what the code does rather than from what the policy hoped.

Two gaps that appendix exposed are closed here. Object storage deletions now emit an audit record, and every request now records its client address.

**Data labelling.** A Data labelling policy on top of the existing three tier scheme, covering how information is identified, labelled, inherited, reclassified and verified.

**Handbook accuracy.** The shared responsibility model described Render and Supabase, neither of which we use, and implied a platform-as-a-service split where the cluster and database sit with the provider. Rewritten against the cluster we actually run. The cache infrastructure page documented a NixOS estate that the Kura nodes replace.

## Why

A control that lives only in someone's head stops happening the first busy week. Writing the policies gives the team something to follow. Automating the account thresholds means 180 and 365 hold without anyone remembering to check.

The ordering matters: recording sign-in activity has to ship and run before acting on it means anything, because until the column is populated we cannot distinguish abandoned from active.

## Approach

**Scope is workforce accounts only.** The sweep matches operator accounts by email domain and never touches customer accounts, which are deprovisioned by their own identity provider over Cross-domain Identity Management. That is faster and theirs to decide.

**The 365 day step scrubs rather than deletes.** `Accounts.delete_user/1` cascades and would take projects and command events with it, destroying the audit trail the control exists to protect. Scrubbing clears the password, linked identities, sessions, email and the legacy command-line token, which is a live credential that is easy to overlook, and leaves a tombstone so historical records keep their foreign keys. The user ID goes away; the record of what that user did does not.

**The checkpoint before the irreversible step is a timestamp, not a state.** This took two passes and the second is the interesting one. The first version asked whether the account was currently disabled. Within a single run, disabling first hands freshly disabled accounts straight to the scrub query, which the tests caught. Across runs it is worse and no ordering fixes it: a run that disables an account and then dies leaves it disabled, so the retry a minute later sees an account that looks like it served its six month checkpoint. A timestamp cannot be produced by a crash, which is why the gate reads `disabled_at`. Accounts with no clock are stamped on first observation rather than backdated, and never scrubbed on missing data.

**Storage logging is scoped to deletions.** Logging every read and write means logging routine artifact traffic at a volume that costs real money and buries the events worth seeing, and customer-initiated access is already attributable through the request record. Deletion is the operation that destroys customer data and the one that most often runs unattended, from retention sweeps and cleanup workers where no request record exists.

**Detection is four conditions, not seven.** The policy originally listed seven, several needing behavioural baselines nobody would maintain. Two of the survivors are restated as reconciliations against an authoritative record rather than anomaly detection: privileged access without a matching elevation, and operator access without a matching grant. Those are precise, cheap here because the elevation and grant records already exist, and cannot drift the way a learned baseline does.

**Retention states thirty days, not twelve months.** Twelve was written as good practice and we do not meet it. The compliance chain still holds because review records are retained for a year, so what was examined outlives the raw logs.

## Impact

No user-visible change. The nightly job is hosted only and, because every existing account has a null `disabled_at` and nothing is scrubbed on a null clock, the first runs can only stamp clocks and disable. Nothing is scrubbable until 180 days after a sweep first stamps it. The irreversible path is dormant on arrival by construction.

The sweep's log line and the storage deletion record are both evidence, so their metadata keys are registered in the Logger config. Without that they are silently dropped and the controls produce nothing.

## Also in here, and not this branch's subject

**A duplicated ClickHouse migration version.** Two migrations landed on main with version `20260817120000`. Ecto refuses to run any migration for a repo whose versions collide, so `mix ecto.migrate` raises before executing anything, which takes down `db:reset` and with it the Seed and Gradle cache acceptance jobs. It blocks main too, not just this branch. The caller test selection migration is renumbered forward rather than the materialized view one, because the view has a dependent backfill behind it and has almost certainly been applied in production already.

## Validation

- Full server suite: **6711 of 6716 passing**. Four failures are localized route tests needing the complete locale set, which pass with `TUIST_DEV_ALL_LOCALES=1` as the test job runs them. The fifth I confirmed on an unmodified `origin/main` checkout, so it is not from this branch.
- The directly affected modules: **466 passing** across `accounts_test`, `authentication_test`, `tuist_web/authentication_test`, `accounts/dormancy_test`, `storage_test`, `observability_context_plug_test` and `oban/runtime_config_test`.
- `mix credo`: no issues across 1672 files. `mix format --check-formatted`: clean.
- `mix excellent_migrations.check_safety`: nothing raised for the new migration.
- `mix ecto.migrate`: completes with zero pending after the version collision fix, where it previously raised.
- `mix gettext.extract`: idempotent, which is the check the Gettext job runs.
- `mise run build` in `handbook/`: clean, which also verifies no dead links from the new pages or sidebar entries.

The dormancy suite covers both thresholds, that a still-active account is never scrubbed, that an account disabled by a previous sweep is not scrubbed by the next one, that an account which has not served the disabled checkpoint is left alone, that a disabled account with no clock gets one instead of being scrubbed, that a scrubbed account is not reprocessed, that a scrubbed user cannot authenticate, and that a run filling its window says so.

### How to test locally

The server suite needs ClickHouse with Keeper, since the sandbox uses transactions:

```
mise run clickhouse:start
mise run db:reset
mix test test/tuist/accounts/dormancy_test.exs test/tuist/storage_test.exs
```

For the handbook, from `handbook/`:

```
mise run dev    # the two new pages are under the Security sidebar
mise run build  # verifies links
```

To exercise the sweep by hand in `iex -S mix`, against accounts you create yourself, since `scrub/1` is irreversible:

```elixir
Tuist.Accounts.Dormancy.sweep(operator_email_domain: "example.com")
# => %{disabled: [...], scrubbed: [...], clock_started: [...], unknown_activity: [...], more_pending: false}
```

`list_disable_candidates/3` and `list_scrub_candidates/3` are public and read-only, so you can see what a sweep would do before running one.


infra/log-review.md
```diff
@@ -0,0 +1,161 @@
...
+
+```logql
+sum by (client_address) (
+  count_over_time(
+    {cluster="tuist-production", namespace="tuist"}
+      |= "Request completed"
+      | logfmt
+      | request_path = "/users/log_in"
+      | method = "POST"
+      | status =~ "4.."
```
fortmarek: **[P1] Query an explicit authentication outcome**

Invalid credentials, rate-limit failures, and successful logins all redirect, so they all produce a 3xx response. This `4..` filter therefore returns no failed password attempts, and changing it to `3..` would conflate failures with successes. Please emit a structured authentication outcome or a dedicated failure event and query that field.

server/lib/tuist/accounts/dormancy.ex
```diff
@@ -0,0 +1,379 @@
...
+  defp as_utc(nil), do: as_utc(Tuist.Time.naive_utc_now())
+
+  defp as_utc(%NaiveDateTime{} = naive) do
+    naive |> DateTime.from_naive!("Etc/UTC") |> DateTime.truncate(:second)
+  end
+
+  defp as_utc(%DateTime{} = datetime), do: DateTime.truncate(datetime, :second)
+
+  defp revoke_sessions(%User{id: id}) do
+    Repo.delete_all(from(t in UserToken, where: t.user_id == ^id))
```
fortmarek, line 365: **[P1] Disconnect sessions that are already live**

Deleting the `UserToken` rows prevents the next session lookup, but an existing LiveView socket already holds its authenticated user and is not forced to re-run authentication. It can therefore keep operating after this sweep disables the account. Please collect the session tokens before deletion and broadcast their `UserToken.live_socket_id/1` disconnects, following the password-reset revocation flow.

handbook/handbook/security/secure-development-and-operations/logging-and-monitoring-policy.md
```diff
@@ -0,0 +1,234 @@
...
+
+Each review shall be recorded with the reviewer, the date, the period covered, the sources examined, and the findings. Records of reviews shall be retained for at least one year and serve as evidence of compliance.
+
+Continuous automated monitoring supplements but does not replace periodic review. Alerts that fire between reviews shall be triaged when they fire.
+
+## 3. Anomaly detection and response
+
+### 3.1 Detection
+
+Automated alerting complements periodic review rather than replacing it. It is configured for conditions where a signal arriving in minutes is materially better than one found at the next review, and where the detection can be stated precisely enough to be acted on rather than dismissed:
```
fortmarek: **[P1] Do not claim these alerts are configured yet**

This PR adds manual review queries, not provisioned alert rules; the new ingestion-loss runbook even says the query is worth promoting to an alert once its baseline is known. Saying alerting “is configured” creates inaccurate compliance evidence. Please either add the actual alert rules and reconciliation automation or describe these as manual periodic controls until that work exists.

handbook/handbook/security/secure-development-and-operations/logging-and-monitoring-policy.md
```diff
@@ -0,0 +1,234 @@
...
+- Creation, rotation, and revocation of tokens, keys, and other credentials
+- Administrative operations performed by Tuist GmbH personnel on customer accounts, including impersonation sessions
+- Export or bulk retrieval of data
+- Errors and exceptions that indicate a security-relevant failure
+
+**Transaction activity.** Individual operations against data and service endpoints:
+
+- API requests, including method, endpoint, response status, and originating account
+- Database schema migrations and administrative queries executed outside the application
+- Object storage read and write operations against buckets holding customer data
```
fortmarek: **[P2] Match this requirement to the implemented storage coverage**

The implementation and Appendix A cover deletions only, while this requires every object-storage read and write. Presigned artifact transfers happen directly against the storage provider, so the server request record does not record the resulting storage access. Please narrow this requirement to deletions, matching the stated approach, or document and enable provider-side read/write audit logging.

server/lib/tuist/accounts.ex
```diff
@@ -1733,6 +1734,48 @@ defmodule Tuist.Accounts do
...
+      # write; re-asserting it in SQL makes them serialize on the row and all
+      # but the first find the condition no longer true.
+      stale_before = NaiveDateTime.add(now, -@last_sign_in_touch_interval_seconds, :second)
+
+      Repo.update_all(
+        from(u in User,
+          where: u.id == ^user.id,
+          where: is_nil(u.last_sign_in_at) or u.last_sign_in_at <= ^stale_before
+        ),
+        set: [last_sign_in_at: now]
```
fortmarek, line 1762: **[P2] Update the data-export documentation for the new collection**

This PR starts populating `last_sign_in_at` for customer users, adds `disabled_at`, and records every request client address, but `server/data-export.md` is unchanged. The repository guardrail requires new personal-data collection to document its storage location, retention, and export behavior. Please update the user-profile and operational-telemetry sections accordingly.

fortmarek (changes requested):
The new controls currently have gaps that make failed-login monitoring ineffective, leave existing LiveView access alive after dormancy revocation, and overstate the logging and alerting coverage. The newly collected personal data also needs to be reflected in the data-export documentation.

infra/log-review.md
pepicrft: You're right, and it's worse than the filter being too narrow. Every outcome redirects: wrong password, rate limited, unconfirmed, and success all return 302, so `4..` matched no failed passwords and `3..` would have conflated them with successes. The 4xx responses I did measure were CSRF and content negotiation, not credentials.

`UserSessionController` now emits an explicit `auth_outcome` on every attempt (`success`, `invalid_credentials`, `rate_limited`, `unconfirmed`) and the runbook counts that field grouped by `client_address`, since a failure carries no account. Covered by `user_session_controller_auth_outcome_test.exs`, in its own sync file because asserting on the record means lowering the global Logger level.

I also corrected the baseline in the runbook. The success and failure split recorded there came from response status, so it was measuring the wrong thing and is now marked unmeasured.

server/lib/tuist/accounts/dormancy.ex
pepicrft: Agreed, and this is the one I'm most glad you caught. Deleting the rows only stops the next lookup; a connected socket holds its user in assigns and never repeats it, so a disabled operator kept working in an open tab until they reloaded.

`revoke_sessions/1` now collects the session tokens before deleting them and broadcasts a disconnect for each, following the password reset flow. One deviation: it publishes straight to `Tuist.PubSub` rather than through `TuistWeb.Endpoint`, because a `Tuist` module may not reference `TuistWeb` and the compiler rejects it. That is what the endpoint does internally, and the payload is the same `%Phoenix.Socket.Broadcast{}` a disconnect listener expects.

Test added in `dormancy_test.exs`: subscribe to the socket topic, run the sweep, assert the disconnect arrives.

handbook/handbook/security/secure-development-and-operations/logging-and-monitoring-policy.md
pepicrft: Fair, and this is exactly the failure mode the policy is supposed to prevent. Saying alerting "is configured" when what exists is a set of review queries makes the document evidence of something untrue.

Rewritten to describe them as what they are: manual reconciliations performed on the periodic review cadence in section 2. Continuous alerting is recorded as the intended end state, with an explicit sentence that nothing in the section should be read as evidence a condition is being detected in real time.

handbook/handbook/security/secure-development-and-operations/logging-and-monitoring-policy.md
pepicrft: Agreed, and the pre-signed point is the part I had not thought through. Transfers go directly against the storage provider, so no server ever sees the resulting access and no amount of application logging would capture it.

Narrowed to deletions, matching the implementation and Appendix A, and the bullet now says why reads and writes are absent, that closing it would need provider-side access logging we have not enabled, and that the authorization decision issuing each pre-signed URL is recorded as a transaction level interface request.

server/lib/tuist/accounts.ex
pepicrft: You're right, that is a guardrail violation and I missed it entirely.

`server/data-export.md` now covers all three. The user section gains `last_sign_in_at` and `disabled_at` with what sets them, the twelve-hour write throttle, and that they live for the life of the account. A new operational telemetry section covers the log records: what each carries, why the client address is there at all (a failed sign-in has no account, so it is the only thing to attribute it to), thirty day retention on the telemetry platform, extractable on request within that window rather than part of the standard export, and that credentials, tokens and bodies are excluded.

fortmarek (approved):

PR #12455, opened 2026-08-18
feat(server): chart every metric on the test case page

fortmarek:
> Follows https://github.com/tuist/tuist/pull/12435, which has merged. Rebased onto `main`, so the diff below is only this change.

## What changed

The test case page's Summary card becomes an Analytics card: the metrics it already reported are now widgets you click to chart across the selected period.

Analytics for a single test case: run outcomes stacked per day, each metric carrying its trend against the previous window, and the test's history beside them

## Why

PR 12435 answers "how long does this test take" with a distribution instead of one mean. It answers it for a point in time, and so did every other number on the page: run count, failure count, flakiness rate, reliability. Each could say where a test stands, none could say how it got there.

That is the question you act on. A test at 93% reliability is a different problem depending on whether it has been at 93% for a month or fell there on Tuesday, and no single figure separates them.

## The card

The layout follows the Tests and Test Cases pages: widgets are the metrics, clicking one draws it below, and the selection rides the URL so a link carries the chart the sender was looking at. Every series loads in one pass, so switching between them costs no round trip.

**Runs** is a stacked bar, one bar per bucket, split by outcome: successful, failed, flaky, quarantined, skipped. This replaces two widgets, since a failure count beside a run count says less than the same failures drawn inside the runs they came from. The bar's height is the run count the widget reports, and the composition is the story: a bar that grows a red band on the same day each week is a different problem from one that has always had it.

**Flakiness rate** and **reliability** are lines on a fixed 0-100% axis. Scaling a rate axis to its data makes a test that sat between 98% and 100% look like a cliff.

**Duration** is the four-series chart from the first commit, now inside the card.

## Decisions a reviewer should weigh

**Segments are mutually exclusive, and that needed a precedence.** A run can be quarantined and flaky and failed at once. Counting it under each would make the stack taller than the run count, which is the one thing a stacked bar promises not to do. Order is quarantined, then flaky, then the run's own status: quarantined runs gate nothing and their failures are expected, and flaky is the more actionable label for a run that did not settle. A test asserts the segments sum to the run count for a bucket holding one run of every kind.

**Skipped is a fifth segment the description did not name.** The four requested categories do not cover a skipped run, and folding those into "successful" would misstate the data while the alternative, leaving them out, would break the sum. A segment that is zero across the whole window is dropped from the chart and the legend, so a test that was never quarantined or skipped shows three colours, not five.

**The flakiness rate counts quarantined flaky runs; the bar files them under Quarantined.** The rate answers "how often did this test fail to settle", which quarantining does not change. The bar answers "what happened to these runs", where quarantined is the more useful label. They are different questions over the same rows, so the widget and the bar can differ by design.

**Reliability's branch fallback is decided once per window, not per bucket.** `test_case_reliability_by_id/4` uses the default branch and falls back to all branches when it has no runs. Applying that per bucket would let each day silently choose its own branch scope, so the series decides once over the whole window and plots the set the widget describes.

**Counts zero-fill, rates and durations do not.** A bucket with no runs really did have zero runs, so the bar is honestly empty. A rate or a duration has no answer there, and 0% would claim the test ran and never flaked. Those carry nil, which ECharts draws as a gap.

**Reading the series is one pass.** Two grouped queries over one test case's rows, both primary-key range reads on `(project_id, test_case_id, ran_at, id)`. They read the same rows as the widgets above them, rather than the per-case aggregate table PR 12435 added, so the chart and the widget cannot state two different numbers for one window.

**What the Summary card took with it.** The metadata row (last status, last run duration, last ran at) is gone at the author's direction; last status is already the badge beside the page title. The recent history timeline moved to a card of its own rather than staying as a 40% column beside the widgets, which is space the charts need.

## Chart tooltips and missing values

ECharts includes null-valued series in an axis tooltip's params, and Noora's tooltip ran that null through the configured formatter, so `formatMilliseconds(null)` printed `0ms`. A gap day showed a break in the line and a tooltip claiming a measurement of zero. `tooltipSeries` now renders a missing value as an em dash. This is generic: any Noora chart with a null in a series had it.

## Validation

`mix format --check-formatted`, `mix credo` (no issues) and `prettier --check assets` are clean. `.pot` regenerated with `mix gettext.extract`; no `.po` touched. No schema or storage change, so `data-export.md` is unaffected.

```
mix test test/tuist/tests/analytics_test.exs test/tuist_web/live/test_case_live_test.exs test/tuist_web/live/test_cases_live_test.exs test/tuist_web/live/tests_live_test.exs test/tuist/marketing/changelog_entry_parser_test.exs
```

117 passed. The analytics tests cover bucketing at all three grains, scoping to one test case in one project, the period bound, empty buckets staying empty for durations and rates while counts zero-fill, the outcome segments summing to the run count, the precedence that keeps a triple-labelled run counted once, the reliability branch scoping and its fallback, and the series agreeing with the widgets over a single bucket. The LiveView tests cover the card opening on the runs bar, switching widgets without a refetch, the fixed rate axis, gaps surviving into the rendered option, an outcome that never happened being dropped from the legend, a duration statistic pulling the card to the duration chart, and the empty state replacing the chart. Two Vitest tests pin the tooltip's handling of a missing value; the full Noora JS suite is 194 passed.

Manual, against locally seeded test cases: a healthy test that degrades nine days in shows the red and yellow bands appear in the bar and the reliability line step down on the same day, the bar tooltip reports segments that sum to the day's runs, a test that stopped running for several days shows a gap in the duration and rate lines and zero-height bars, the widgets switch charts and the URL follows, and light and dark both look right.


server/lib/tuist/tests/analytics.ex
```diff
@@ -1364,6 +1375,220 @@ defmodule Tuist.Tests.Analytics do
...
+          successful:
+            fragment(
+              "countIf(? = 'success' AND NOT ? AND NOT ?)",
+              tcr.status,
+              tcr.is_flaky,
+              tcr.is_quarantined
+            ),
+          failed:
+            fragment(
+              "countIf(? = 'failure' AND NOT ? AND NOT ?)",
```
pepicrft, line 1516: The widget failed count includes every failure, while this chart series removes failures that are flaky or quarantined. Selecting Failed can therefore show a nonzero widget while the red series is zero, so the chart has not graphed the selected metric. For example, a run that is both failed and flaky reports 1 in the widget but only appears in the Flaky segment. Could the selected metric match the mutually exclusive segment, or could the chart include a raw-failed series?

server/lib/tuist_web/live/test_case_live.ex
```diff
@@ -410,26 +582,64 @@ defmodule TuistWeb.TestCaseLive do
         [start_datetime: start_datetime]
       end
 
+    # The window before this one, of the same length, is what every trend on the
+    # card compares against. A relative preset runs up to now, so its length is
+    # measured to now rather than to the bound it does not carry.
+    previous_opts = [
```
pepicrft, line 596: The previous window ends at the same instant that the current window begins. The analytics filter uses greater-than-or-equal for its start and less-than-or-equal for its end, and test case run timestamps have microsecond precision. A run exactly at start_datetime is therefore counted in both windows and skews every trend. Please make one bound exclusive, or subtract the smallest supported timestamp unit from the previous end, and add a boundary test.

pepicrft (approved):
Approved. I left two follow-up comments for the metric-to-chart alignment and the trend-window boundary.

PR #12462, opened 2026-08-19
feat(server): grant prepaid runner credit from paid invoices

fortmarek:
Builds on https://github.com/tuist/tuist/pull/11903, which meters runner usage as normalized compute units with one Stripe Meter per platform. That PR has since merged, so this now targets `main` directly.

## What this adds

Prepaid runner access, as a money-denominated Stripe billing credit grant scoped to the runner metered Prices.

- `Tuist.Billing.CreditGrants`: a thin wrapper over Stripe's credit grant and credit balance summary endpoints. `stripity_stripe` 3.2.0 has no generated module for them, so the requests are hand-built the same way `Tuist.Billing.report_meter_event/5` builds meter events.
- `Tuist.Billing.Invoices`: reads an invoice's lines in full, since they are paginated and the webhook payload carries only the first few.
- `Tuist.Runners.Prepaid`: the policy. Funding ratio, expiry, platform scoping, invoice metadata parsing, grant creation, trial grants, and remaining balance.
- `Tuist.Billing.Workers.CreateRunnerPrepaidGrantWorker` plus an `invoice.paid` handler in the billing webhook controller.
- Remaining prepaid credit and per-grant expiry on the billing page.
- A prepaid panel on the ops account page: current balance, and a form that takes minutes, quotes the money live, and adds the charge to the account's next invoice.

## Why it is money and not minutes

Usage is always reported gross, at the on-demand rate, for every customer. Prepaid is not a discounted second Price and not a pool of minutes decremented inside Tuist: it is a balance of money on the Stripe customer that Stripe draws down as those gross-rated line items are invoiced. Reporting is byte for byte identical for a prepaid customer and a pay-as-you-go one, so nothing in the metering path has to know which kind of customer it is looking at.

The discount therefore lives in how over-funded the grant is, not in a second rate card. At the agreed macOS terms ($0.075 per baseline machine-minute on demand, $0.06 prepaid) a customer paying $X can consume $X / 0.8 of on-demand-priced usage, so the grant is funded at 1.25x what they paid.

Two reasons for money over a minutes ledger:

- it keeps mixed machine types fungible without pretending a minute on one shape is worth a minute on another, since the credit is spent at whatever each machine actually costs
- it avoids running a second credit ledger next to Stripe's, which would have to stay reconciled with every invoice, refund, proration and credit note Stripe issues on its own

## Decisions worth reviewing

**Grants scope to explicit Prices, never to `price_type: "metered"`.** A metered scope would let runner credit pay for every metered Price on the account, remote cache hits and language-model tokens included, which is exactly what a runner-specific balance must not do. Scoping by Price is also what makes per-platform prepaid terms possible at all: with a single global runner Price, prepaid would collapse to one discount percentage across every machine type.

**The prepaid charge rides along on the monthly bill, so the marker is on the invoice line rather than the invoice.** A prepaid charge is created as a pending Stripe invoice item, which Stripe sweeps onto the customer's next ordinary invoice; nothing here has to know when the period closes. That invoice therefore also carries the month's metered usage, so marking the invoice and funding from its `amount_paid` would convert the entire bill into runner credit. Funding from the marked line's own amount takes exactly the prepaid portion, and retires the "a prepaid invoice must not carry unrelated line items" constraint an earlier revision needed. Each marked line becomes its own grant on its own terms, so two top-ups bought in one month at different ratios stay independently priced instead of being averaged.

**The webhook cannot decide whether an invoice is prepaid.** Its payload carries at most the first handful of lines, so a prepaid line further down a busy bill would read as "not prepaid" and the credit would be lost silently. The controller enqueues for every paid invoice and the worker pages the lines endpoint and decides on the full picture. An ordinary invoice costs one cheap no-op.

**Minutes are quoted, never stored.** The ops form asks for minutes because that is how a deal is negotiated, and shows the derived money live as it is typed, but what is charged and granted is money. The quote is against the macOS 6 vCPU / 14 GB baseline, the only machine with an agreed rate, and the page says so plainly: the same balance buys fewer minutes on a larger machine, which is the whole reason the balance is money rather than minutes. The two baseline rates are the single source of truth and the funding ratio is derived from them rather than restated, so the 1.25x cannot drift from the rates it represents.

**Per-platform terms are supported but not used by default.** A grant covers every configured runner Price, because fungibility is the point and Linux has no agreed rate yet. An invoice can narrow it to one platform (`tuist_prepaid_runners: "macos"`) when a deal is struck on one platform's terms alone.

**Where the terms are configured.** The default expiry (365 days) and the two baseline rates the ratio derives from are module attributes next to `Tuist.Billing`'s `@unit_prices`, not deployment config: they are rate-card facts, identical in every environment, and a change to any of them should be a reviewed commit rather than a values file edit. Per-deal terms come from metadata on the Stripe invoice line, because prepay is negotiated per customer and one deal's terms must not bind the next one's.

**A hand-typed ratio is a money multiplier, so it is validated hard.** An out-of-range or unparseable ratio is rejected outright rather than falling back to the default. A `1250` typed for `12500` stops the grant instead of quietly issuing a tenth of the credit. The accepted band is par (no discount) to 20000 basis points (50% off). An empty value is treated as unset, since that is how Stripe stores an absent one.

**Top-ups and expiry need no machinery.** Every paid prepaid invoice creates its own grant, and Stripe applies whichever grants are live at invoice time in priority then expiry order. A top-up is just another grant, and an exhausted or expired one simply stops applying, so usage past it falls through to the on-demand rate it was always reported at. No merging, no re-basing, no expiry sweep.

**Trials share the primitive.** A time-boxed runner trial is the same grant with `category: "promotional"` (Stripe reports promotional credit separately from purchased credit, which keeps free trial usage out of recognised revenue) and a shorter clock. Trial grants get a stronger priority so free credit is consumed before credit the customer paid for. There is no automatic enrolment on top of this; a trial is granted deliberately, per account.

## Correctness and failure handling

**Granting money twice is the failure that matters, so idempotency is three layers deep.** The Oban job is unique on the invoice id for all time, so a redelivered `invoice.paid` cannot enqueue a second run. Each request carries an idempotency key derived from the invoice and the line. And Stripe is checked for an existing grant against the same line before one is created, which also means a retry after a partial failure finishes the remaining lines and leaves the rest alone. Oban prunes completed jobs eventually and Stripe expires idempotency keys after 24 hours, so only the innermost check holds indefinitely.

**No runner Price exists in any environment yet, which this has to survive.** The invoice is paid and the grant is owed, so on `:no_runner_prices_configured` the worker snoozes hourly instead of discarding, and only fails into alerting after roughly three days when a human is clearly needed. Snoozing does not consume an Oban attempt (`Oban.Engines.Basic.snooze_job/3` does `inc: [max_attempts: 1]`), so the grant stays owed rather than being dropped.

**The job carries only the invoice id and re-reads the invoice from Stripe on every attempt**, rather than freezing the webhook's copy into its args. A metadata term corrected after the fact, a mistyped funding ratio or a missing platform scope, is then picked up by the retry instead of being replayed wrong.

**The prepaid marker is checked in the controller**, so the subscription renewals that make up nearly every `invoice.paid` do not each enqueue a job that immediately no-ops. An invoice whose marker is present but malformed still enqueues, so the mistake surfaces as a failing job rather than being read as "not prepaid".

**Guardrail carried over from 11903:** do not blank a runner Price id in `stripe.prices.runners` once it has gone live. `configured_runner_price_ids/0` filters empty ids, so a later plan change would stop recognising the existing runner subscription item, delete it, and lose the cycle's accrued usage. This is now written down in `server/lib/tuist/billing/AGENTS.md`.

## Bug found while building

Grants were being sorted by their `DateTime` expiry with `Enum.sort_by/2`, which falls back to Erlang term ordering on the struct. That compares fields alphabetically (day, hour, microsecond, minute, month, ...), so September 2026 sorted after January 2027 and the billing page would have shown the wrong "soonest expiry". Now sorted on the epoch. Covered by a test.

## Impact

No schema changes and no new Tuist-side storage: grants live entirely in Stripe. `server/data-export.md` records that, so a legal request points at the Stripe customer rather than at a table that does not exist. Runner usage reporting is unchanged, so `runner_sessions` and the metering path are untouched by whether an account is prepaid.

Customer-facing change is additive: a "Prepaid runner credit" card appears on the billing page only for accounts that actually hold credit. Operators get a matching panel on the ops account page, which also warns when an account has no subscription, since Stripe never generates an invoice for one and a pending charge added there would sit unbilled indefinitely.

## Notes for the reviewer

- End-to-end validation against Stripe is not possible until the runner Prices are created, which is the state 11903 deliberately leaves every environment in. Everything below the Stripe boundary is covered by tests; the boundary itself is not exercised against a real account.
- This branch was cut from 11903's branch before that PR was squash-merged, so it carries a merge from `main` resolving the resulting conflicts. Every conflicted file was 11903's own, and none carried edits from this branch, so all of them were resolved to `main`'s version, which additionally holds the three fixes that landed on 11903 after this branch was cut. The gettext template was the only file both sides changed; it was regenerated from source on top of `main`'s version rather than hand-merged, so it adds seven strings and drops none. `server/mix.lock` auto-merged to the older side and was taken from `main` instead.
- The net diff against `main` is exactly the 17 files of prepaid work, so the merge introduced nothing beyond conflict resolution.

## How to test locally

From `server/`:

```
mix test test/tuist/runners/prepaid_test.exs test/tuist/billing/
```

Then the suites this touches, to check for regressions:

```
mix test test/tuist/runners/prepaid_test.exs test/tuist/billing/ test/tuist/billing_test.exs test/tuist_web/controllers/webhooks/ test/tuist_web/live/billing_live_test.exs test/tuist_web/live/ops_account_live_test.exs test/tuist/runners/billing_test.exs
```

221 tests pass across that second set. `mix credo` and `mix format --check-formatted` are clean on every changed file. `mix gettext.extract` updated only `dashboard_account.pot`; no `.po` file is modified.

Driving the ops page locally: open `/ops/accounts/<id>`, type a minute count into the prepaid form, and check the quote updates as you type (10000 minutes should read $600.00 added and $750.00 granted). Submitting creates the pending Stripe invoice item; no credit is granted at that point.

Exercising the full flow against Stripe, once a runner Price exists:

1. Create a runner Meter and Price, and put the Price id in `stripe.prices.runners` under the platform's meter event name.
2. Add prepaid minutes from the ops account page. This creates a pending invoice item carrying `tuist_prepaid_runners` metadata; the account can carry any other usage at the same time.
3. Let the monthly invoice generate, or advance a Stripe test clock. The prepaid item appears as one line alongside the metered usage.
4. Pay the invoice. `invoice.paid` enqueues `CreateRunnerPrepaidGrantWorker`, which reads every line, finds the marked one, and grants 1.25x that line's amount, scoped to the configured runner Prices and expiring in a year. The usage lines fund nothing.
5. The billing page and the ops panel both show the remaining balance and the expiry date.


fortmarek:
The UI I built for now until this gets @asmitbm's special touch

image
image

server/lib/tuist/runners/allowance.ex
```diff
@@ -0,0 +1,321 @@
...
+
+    total_ms = per_day |> Enum.map(&elem(&1, 1)) |> Enum.sum()
+
+    %{
+      period_start: DateTime.to_date(period_start),
+      period_end: DateTime.to_date(period_end),
+      usage_through: DateTime.to_date(usage_end),
+      minutes: div(total_ms, 60_000),
+      free_minutes: free_monthly_minutes(),
+      gross: Prepaid.on_demand_cost_for_milliseconds(total_ms),
```
pepicrft: Linux is currently reporting-only: the deployment values omit its Price and the platform rows intentionally omit it. `total_ms`, however, includes Linux before both values are calculated with `Prepaid.on_demand_cost_for_milliseconds/1`. A Linux-only account past the allowance is therefore shown a macOS-dollar gross and billed amount even though no invoice can carry that usage. I reproduced 60 Linux minutes as $4.50 billed with only a $0.00 macOS row. Calculate the monetary totals from priced-platform usage, or render them as unavailable until Linux has a rate.

server/data-export.md
```diff
@@ -82,6 +82,8 @@ Sensitive authentication data (passwords, tokens) are excluded from exports.
...
...ants scoped to the runner metered Prices, not a minutes ledger in Tuist. Tuist stores no balance of its own — it creates the grant when an invoice carrying a prepaid line is paid and reads the remaining balance back from Stripe on demand, so an export of Tuist's databases contains no prepaid balance and the authoritative record is the Stripe customer. Each grant carries Tuist-set metadata identifying it (`tuist_runner_credit` = `prepaid` | `trial`) and, for prepaid, the invoice and invoice line it was funded from plus the amount paid and funding ratio applied (`tuist_prepaid_invoice_id`, `tuist_prepaid_invoice_line_id`, `tuist_prepaid_paid_cents`, `tuist_prepaid_funding_ratio_bp`). The per-deal terms themselves live as metadata on the Stripe invoice line, which is also where the prepaid charge itself sits: it is created as a pending Stripe invoice item so it rides along on the customer's ordinary monthly invoice rather than arriving as a separate bill, and each marked line funds its own grant. Runner usage is always reported to Stripe gross at the on-demand rate for prepaid and pay-as-you-go customers alike, so `runner_sessions` above is unaffected by whether an account is prepaid.
```
pepicrft: This description is now inconsistent with the implementation: trials are stored by account timestamps and removing runner subscription items, not credit grants, and `bill_prepaid_minutes/3` creates the grant before the pending invoice item is paid. Please describe the actual state and timing, otherwise a data export or transparency response will point to metadata that is never created.

server/lib/tuist_web/live/ops_account_live.ex
```diff
@@ -83,6 +91,104 @@ defmodule TuistWeb.OpsAccountLive do
...
+
+  @impl true
+  def handle_event("set_prepaid_minutes", %{"minutes" => minutes}, socket) do
+    case parse_minutes(minutes) do
+      {:ok, minutes} ->
+        account = Accounts.create_customer_when_absent(socket.assigns.account)
+
+        case Prepaid.set_minutes(account, minutes) do
+          {:ok, _result} ->
+            quoted = Prepaid.quote_minutes(max(minutes, 1))
```
pepicrft: When minutes is zero, `Prepaid.set_minutes/2` clears existing grants and invoices nothing. `max(minutes, 1)` still quotes one minute, so the success message says $0.06 was added to the next invoice. Please use a separate clearing confirmation instead of deriving an invoice amount.

pepicrft (approved):

server/lib/tuist/runners/allowance.ex
fortmarek: Confirmed and fixed: `gross`, `billed`, the daily rows, the projection and the per-platform split now all come from priced-platform usage only (via a new `platforms:` filter on the runner billing queries), while `minutes` still reports every minute the account ran. Your repro is a test: 160 Linux minutes now reports 0.00$ gross and billed instead of macOS-priced money.

server/data-export.md
fortmarek: Rewritten to the actual state and timing: trials are the account timestamps described in the entry above rather than credit grants, the grant is created when the minutes are sold rather than when the invoice is paid, and the metadata list now reflects what is really set (no `trial` kind, and `tuist_prepaid_invoice_id` only on the `invoice.paid` backstop path).

server/lib/tuist_web/live/ops_account_live.ex
fortmarek: Fixed: clearing has its own confirmation ("… no longer holds any prepaid runner minutes") and no longer derives an invoice amount, so `max(minutes, 1)` is gone. Covered by a test asserting the flash carries neither an amount nor "next invoice".

PR #12464, opened 2026-08-19
perf(cli): stop deriving cache hashes from a second framework search paths pass

fortmarek:
Follow-up to #12359, which fixed the warm-generate regression a customer reported after upgrading from
4.193.4 to 4.203.4. They confirmed the fix and shared a session log from the canary so we could look for
further headroom. This PR is what came out of that log.

### What the log showed

Reconstructing the phase timeline from their `4.205.0-canary.11` run (106s wall, ~330 projects, ~1,430
targets, ~700 substituted):

| Phase | Time |
|---|---:|
| Load + early mappers | ~8s |
| `FrameworkSearchPathsGraphMapper` inside `CacheHashingGraphMapper` | 26.4s |
| `TargetsToCacheBinariesGraphMapper` (443 lookups) | 8.7s |
| `FrameworkSearchPathsGraphMapper` post-substitution | 16.7s |
| `CachedModulesDebuggingGraphMapper` | 6.0s |
| `GraphLinter` — an unlogged gap before `Generating workspace` | ~14s |
| 327 projects generated concurrently | ~4s |
| `allProjectDependencies` for the app project alone | ~15s |
| Finish + write | ~7s |

### What changed

**Cut duplicated graph traversal.** `transitiveStaticDependencies` was uncached, so the same subtree was
re-walked once per target by `linkableDependencies` and again per direct dynamic dependency by
`transitiveStaticDependenciesOfDynamicFrameworkDependencies`. The two "static xcframeworks behind a dynamic
xcframework" walks ran per consuming target although hundreds of targets link the same artifacts; they are
now walked and cached per root xcframework. Unioning the per-root walks returns the same set — the walk only
reaches xcframeworks, and a root cannot satisfy the test because roots are the dynamic ones while the test
selects static ones. `embeddableFrameworks` is memoized too. Also dropped a dead `.sorted()` in
`FrameworkSearchPathsGraphMapper` and parallelized the linter's duplicated-product-name check, which walks
every target's transitive closure serially.

**Hash graph dependency references without building an intermediate.** `GraphDependencyReference.hash(into:)`
built a `Synthesized` value on every call. These are hashed millions of times over a large graph, and each
hash copied a path out of the reference to build the intermediate. The property subset is unchanged, so the
conformance stays consistent with `==`, which compares a superset of it.

**Derive framework search paths without materializing references.** `searchablePathDependencies` builds a
`GraphDependencyReference` for every dependency reachable from every target — ~300,000 of them for ~700
targets — and its only consumers reduce each to an artifact path or an SDK search path.
`precompiledSearchPathDependencies` answers that directly. It is an independent derivation, not a
projection, so `PrecompiledSearchPathsDifferentialTests` holds it against the original over generated
graphs. That harness earned its keep: the first candidate diverged on 41 of 334 targets. Its coverage
assertions then caught two blind spots the equality check passed — conflicting platform conditions were
never generated, leaving the condition-filtering untested, and statics behind a dynamic xcframework were
barely reached.

**Time each `GraphLinter` check.** The ~14s above was a silent hole in the session log; identifying it as
the linter took reading the call path by hand. Each check now logs its duration in the same shape the graph
mappers do.

### One change was made and then reverted

`7a8c24da0f` removed `FrameworkSearchPathsGraphMapper` from the cache-hashing normalization pass — the
26.4s line above, whose output is otherwise discarded. Review caught that those settings carry something
the rest of the hash does not: lookup precedence. `-F` paths are searched in order, so when two precompiled
artifacts share a basename the order decides which module resolves, which is why the mapper has a
`conflictingFrameworkSearchPaths` branch at all. `DependenciesContentHasher` cannot stand in for it: it
hashes each dependency by content and combines through `hashes.sorted()`, so it is deliberately
order-independent. The trade also went the dangerous way — before, a relocated artifact produced a
different key and so a miss; after, the same key and so a hit on an artifact compiled under the opposite
precedence.

Reverted in `7c3102e53c`. Re-landing it needs the precedence preserved portably rather than dropped: an
ordered fingerprint of the artifacts whose basenames collide, which discriminates on order without
embedding absolute paths. That should still avoid most of the 26.4s, since it only needs artifact paths and
only where basenames collide. **This PR therefore no longer changes cache keys.**

### Measurements

Synthetic graph matching the reported shape (712 targets, 700 substituted xcframeworks), measured on this
branch against the merge base:

| | before | current |
|---|---:|---:|
| `searchablePathDependencies` over every target, CPU | 17.1s | 9.9s |
| what the mapper now calls instead, `precompiledSearchPathDependencies`, CPU | — | 7.6s |
| `FrameworkSearchPathsGraphMapper.map`, CPU across all threads | 53.3s | 36.8s |
| `FrameworkSearchPathsGraphMapper.map`, wall | 5.62s | **3.79s** |

Resident memory is unchanged (226 MB → 227 MB).

Scaled onto the customer's phase timings, the two framework search path passes (26.4s + 16.7s) would come
down by roughly a third, so ~14s of their 106s run. `CachedModulesDebuggingGraphMapper` and the linter also
benefit and are not separately measured. This is a proxy on a Debug build, not their graph — it needs a
canary before any of it is quoted to them.

### What a reviewer should know

**None of this is measured on a real large project.** The synthetic benchmark is a proxy; the numbers above
should not be quoted to a customer without a canary run.

**Traversal is not the lever the phase timings suggested.** Fanning `searchablePathDependencies` across
every target concurrently accounts for only ~2.2s of that mapper's ~4.6s wall, and it reaches roughly 3x on
a 16-core machine. The rest is per-target search path derivation and the serial rewrite that applies
settings back onto every target. A bitset reachability index that cut serial traversal 29% made the
concurrent mapper's wall time *worse* and is deliberately not here.

### How to test locally

```bash
xcodebuild test -workspace Tuist.xcworkspace -scheme TuistCacheEEAcceptanceTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""
```

16 tests, covering warm followed by generate across macros, catalyst slices, app intents metadata, foreign
build outputs, nested header xcframeworks and cached xctest support. `multiplatform_app_module_cache` in the
canary suite fails without the watchOS 26.5 platform installed, unrelated to this change.

```bash
xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistCoreTests -only-testing TuistGeneratorTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""
```

51 XCTest cases and 382 Swift Testing cases pass, and `mise run lint` is clean.


pepicrft (approved):

cli/Tests/TuistCoreTests/Graph/PrecompiledSearchPathsDifferentialTests.swift
```diff
@@ -0,0 +1,263 @@
...
+                    }
+                }
+                dependencies[node] = edges
+            }
+        }
+
+        // Conflicting condition chains. A single conditioned edge only ever narrows a dependency; it takes two
+        // conditions disagreeing along a path for `combinedCondition` to report `.incompatible`, which is the
+        // case the candidate's admission guard exists for. Left to chance those are vanishingly rare.
+        for (edge, condition) in conditions where generator.chance(60) {
```
esnunes, line 148: The conflicting-condition pass here iterates `conditions`, which is a `[GraphEdge: PlatformCondition]`. Since Swift randomizes hash seeds per process, the iteration order — and so both which edges `chance(60)` picks and which write wins when two edges resolve to the same key — changes between runs. Everything else in the generator iterates arrays, so this looks like the only source of drift, but it lands on the corpus itself.

Two things follow. `withConditionDroppedDependency > 5` at :255 becomes a nondeterministic assertion — the margin looks comfortable enough that I'd call a flake unlikely rather than impossible. The one I'd care more about is reproducibility: if this harness ever does report a divergence, re-running won't give you the same corpus. And :234 prints `seed \(iteration)`, but that's the loop index rather than the seed (`0x2545_F491_4F6C_DD1D &+ UInt64(iteration &* 7919)`), so the message doesn't get you back there either.

Since the whole point is to re-run this against future edits, would it be worth iterating a deterministically ordered snapshot of `conditions` and printing the real seed? Feels like a couple of lines for a much better failure story.

cli/Tests/TuistCoreTests/Graph/GraphTraverserStaticXCFrameworksTests.swift
```diff
@@ -0,0 +1,58 @@
...
+import FileSystemTesting
+import Foundation
+import Path
+import Testing
+import XcodeGraph
+@testable import TuistCore
+
+struct GraphTraverserStaticXCFrameworksTests {
+    @Test(.inTemporaryDirectory)
+    func staticXCFrameworksLinkedByDynamicXCFrameworkDependencies_whenDynamicXCFrameworksOverlapAndChain() throws {
```
esnunes, line 11: The commit message is upfront that this passes against both implementations, and I think that's the right call — I worked through `filterDependencies` and the root-exclusion divergence this shape is aimed at genuinely can't happen for these two callers: every root is a dynamic xcframework, both `test` closures require `.static`, and `skip` is false for xcframeworks either way. The doc comment's argument holds.

What that leaves, though, is that the regression which *can* happen isn't covered. The cache is keyed by root xcframework and shared across every target — that sharing is where the speedup comes from and the only place a bug can hide. This test uses one target, so the warm re-call only re-reads entries that same target populated. A cache accidentally keyed by target, or one leaking the caller's root set instead of the root's own subtree, would pass it unchanged.

A two-target graph would have teeth: `T1 -> DynA -> StaticA`, plus `T2 -> DynA` and `T2 -> DynB -> StaticB`. Query `T1` first so `DynA`'s entry is populated, then assert `T2` gets both statics.

One more thing while you're in there: `staticObjcXCFrameworksLinkedByDynamicXCFrameworkDependencies` has no per-target cache at all, only the shared per-root one, so it's the more exposed of the two — might be worth the same shape of case.

cli/Sources/TuistCore/Graph/GraphDependencyReference.swift
```diff
@@ -357,6 +353,55 @@ public enum GraphDependencyReference: Equatable, Comparable, Hashable {
     #endif
 }
 
+// MARK: - Hashable
+
+extension GraphDependencyReference {
+    /// Hashes the same subset of properties `Synthesized` does, but without building one. These references are
+    /// hashed millions of times over a large graph — once per set insertion and again per collision check — and
+    /// the intermediate value copied a path out of the reference on each of them.
+    public func hash(into hasher: inout Hasher) {
```
esnunes, line 362: I went through all nine cases and this is correct — each one combines exactly what `Synthesized` carried, and every one of those sets is a subset of what `==` compares, so `a == b` still implies equal hashes. (`expectedSignature ?? ""` collapsing nil and "" is fine too, and isn't new — `Synthesized.init` already did it.)

What I'd flag is that nothing holds it there. This type has a bespoke `==` that deliberately diverges from the synthesized one, so `==` is a thing people edit, and the hash is now a second hand-maintained list that has to stay a subset of it. If that breaks there's no compile error — you get duplicate entries in `Set<GraphDependencyReference>` and lookups that miss, which is the class of bug #12359 was chasing.

`GraphDependencyReferenceTests` has two tests today and both are about `Comparable`. Would a small property test be worth adding — over a fixture list, for every pair assert `a == b` implies `a.hashValue == b.hashValue`? Cheap, and it pins the invariant for whoever touches `==` next.

esnunes (approved):
This is a really nicely built PR. The commit-by-commit narrative, the honesty about what the synthetic benchmark does and doesn't prove, and especially the write-up of the reverted commit — including that the machine-dependence you'd treated as a defect was partly load-bearing — make it much easier to review than a diff of this shape usually is.

I went through the equivalence claims rather than taking them on trust, and they hold. The hand-written `hash(into:)` combines a strict subset of what `==` compares in all nine cases. The removed `.sorted()` reaches no output unsorted — every path from those Sets to a build setting or the `.resp` file goes through an explicit sort, including the `conflictingFrameworkSearchPaths` branch. The per-root walk really does union to the batched result, for the reason the doc comment gives: roots are always dynamic xcframeworks and both predicates require static. And `precompiledSearchPathDependencies` matches the reference branch for branch — every union site maps onto one, and every branch the new code skips is provably inert (`.target`/`.packageProduct` references carry no `precompiledPath`, the host subtraction is dead at `shouldExcludeHostAppDependencies: false`, and the synthesized App Clip SDK is `.system`, which has no framework search path). The new concurrency is safe too: `GraphCache` is lock-backed, the cached computations are pure, and the concurrent map preserves order. I also confirmed the revert is complete — `GraphMapperFactory.swift` and its test don't appear in the net diff at all.

So I found no correctness bug, and none of the notes below are blocking — approving. They're all about tests, which I think is the right place to invest here precisely because those equivalence arguments are the load-bearing evidence for the whole refactor. The one I'd most like to see is that the differential harness's corpus isn't reproducible run to run: the conflicting-condition pass iterates a Dictionary, so the same seed gives different graphs, and the failure message prints the loop index rather than the seed. That's a couple of lines to close.

One process note: at the time I looked, Linux unit tests and lint were green but the macOS unit tests and acceptance tests were still running, so the new harness hadn't reported yet.

PR #12483, opened 2026-08-19
fix(infra): deliver cilium values to existing clusters instead of only at onboarding

fortmarek:

`infra/k8s/mgmt/bootstrap/cilium-values.yaml` was consumed only by `mise run k8s:bootstrap-workload`, which runs at cluster onboarding. Editing it had no effect on a running cluster until someone re-ran bootstrap by hand.

That is not hypothetical. [#12401](https://github.com/tuist/tuist/pull/12401) added `drop:sourceContext=pod|ip;destinationContext=pod` to the Hubble drop metric so a policy-denied kura cache node could be attributed to a source. It merged, production never picked it up, and two days later the "Kura Cache Rejecting Runner Traffic" alert fired again with `hubble_drop_total` still carrying no `source` label, so the drops were still unattributable. A merged PR was mistaken for a deployed change.

The node CSI drivers hit the same class of drift in [#11346](https://github.com/tuist/tuist/pull/11346) and got a re-apply workflow for it. Cilium was the last kube-system add-on without one.

### What changed

Adds `.github/workflows/cilium-deployment.yml`, path-triggered on the values file, following the `hccm-deployment.yml` / `csi-deployment.yml` precedent already documented in `infra/AGENTS.md` and sharing their `server-deploy-<env>` concurrency lane.

Two deliberate departures from those siblings:

**It cascades staging then canary then production** (`max-parallel: 1` plus `fail-fast`) rather than fanning out in parallel. HCCM and the CSI drivers are peripheral components whose blast radius is one subsystem. Cilium is the CNI, so each env has to prove itself before the next one starts.

**It explicitly restarts the agent DaemonSet.** This is the part that is easy to get wrong by assumption, so it was verified by templating the chart: `hubble.metrics` and most of that values file render into the `cilium-config` ConfigMap, and the cilium-agent pod template carries no checksum annotation over it. A bare `helm upgrade` therefore rewrites the ConfigMap, leaves the pod template untouched, rolls nothing, and reports success while every agent keeps serving its old in-memory config. Without the rollout this workflow would have reproduced the exact failure it exists to prevent: a green deploy that changed nothing. The rollout is also the step carrying the blast radius, which is why the cascade above exists.

Supporting details:

- The chart version is parsed out of `bootstrap-workload.sh` rather than pinned a second time, so a freshly onboarded cluster and an existing one cannot end up on different Cilium versions. Same single-source reasoning as `k8s:lint-version-drift`.
- `k8sServiceHost` / `k8sServicePort` are per-cluster. Bootstrap reads them from the mgmt Cluster CR; this workflow has no mgmt access, so it derives them from the workload kubeconfig's server URL (which clusterctl minted from that same `controlPlaneEndpoint`) and cross-checks them against `helm get values` before upgrading. A wrong kubeconfig fails the job instead of silently repointing the CNI at a different apiserver.
- The job refuses to run where no `cilium` release exists. Onboarding a new cluster stays bootstrap's job.

### Why this shape rather than GitOps

A Flux or Argo `HelmRelease` would also close the "nothing re-applies the file" gap, and would add continuous drift detection, which this workflow does not have (it reconciles on push, so a hand-edit to the live config survives until the next values change).

It was not chosen here for two reasons. Cilium is the worst first candidate for it: the CNI has to exist before anything in the cluster runs, including the GitOps controller that would manage it. And more importantly it would not have fixed this bug. Flux would apply the ConfigMap, report `Reconciled` / `Ready`, and leave every agent on the old config, which is the same silent no-op now carrying a green checkmark. The explicit rollout is required either way.

### Also in this PR

- `infra/AGENTS.md` now lists Cilium alongside HCCM and CSI in the kube-system add-on deploy section, and records the cascade plus ConfigMap-checksum reasoning so the next person does not have to rediscover it.
- The alert runbook's Hubble attribution paragraph now points at the workflow, since that paragraph explicitly discusses clusters that have not had the value applied.

### Not in this PR

The Grafana rule `ffuscyncueo74d` was separately corrected in the Grafana console (rate window `[10m]` to `[5m]`, plus a rewritten description). That rule is hand-created and not provisioned from this repo, so it cannot be changed by a commit. `alerts.md` had documented `[5m]` since [#12401](https://github.com/tuist/tuist/pull/12401) while the live rule still ran `[10m]`, which is what made the alert page on a controller rollout on 2026-08-19. Worth noting as a follow-up: dashboards are reconciled from the repo via Git Sync but alert rules are not, and we found this drift by inspecting one rule out of many in `alerts.md`.

### How to test locally

The workflow only runs in CI, but the two parsing steps it depends on can be checked from a clone:

```bash
grep -E '^\s+--version [0-9]' infra/mise/tasks/k8s/bootstrap-workload.sh | head -1 | awk '{print $2}'
```

That should print the pinned chart version (`1.18.7` at time of writing), matching `bootstrap-workload.sh`.

The ConfigMap claim that motivates the rollout step can be reproduced against the upstream chart:

```bash
helm template cilium cilium/cilium --version 1.18.7 -n kube-system --set k8sServiceHost=x --set k8sServicePort=6443
```

`hubble-metrics` appears as a key in the `cilium-config` ConfigMap, and the `cilium` DaemonSet pod template carries only `kubectl.kubernetes.io/default-container`, with no checksum annotation over that ConfigMap.

End to end, the first real exercise is a `workflow_dispatch` run against staging, which should upgrade the release, roll the agent DaemonSet, and leave every node Ready. After it runs, `hubble_drop_total` on that cluster should start carrying `source` and `destination` labels.


.github/workflows/cilium-apply.yml
```diff
@@ -0,0 +1,243 @@
...
+    name: Cilium (${{ inputs.environment }})
+    runs-on: tuist-linux
+    environment:
+      name: server-k8s-${{ inputs.environment }}
+    # Hold the same lane as server-deployment.yml's deploy + observability
+    # jobs (and hccm/csi-deployment.yml) so CNI upgrades and server
+    # rollouts can't race for the kube apiserver on the same cluster.
+    concurrency:
+      group: server-deploy-${{ inputs.environment }}
+      cancel-in-progress: false
```
esnunes: One thing I wanted to flag about this comment: I grepped every `group: server-deploy` in `.github/workflows/` and `server-deployment.yml` uses four suffixed groups (`-deploy`, `-observability`, `-tailscale`, `-platform`). No bare `server-deploy-<env>` group exists outside hccm, csi and this file, and GitHub matches concurrency groups by exact string. So this lane serializes Cilium against HCCM and CSI, but not against the server deploy.

For HCCM that was probably harmless, but here the job rolls every cilium-agent in the cluster, which could plausibly make an in-flight server `helm --atomic --timeout 60m` look unhealthy and trigger its rollback. Since the same comment sits in the two siblings, I think the fix (if the mutual exclusion was really intended) is a shared group across all four workflows rather than a change here. Either way, might be worth adjusting the comment so it does not promise a guarantee that is not there.

.github/workflows/cilium-deployment.yml
```diff
@@ -0,0 +1,123 @@
...
+
+  preview:
+    needs: staging
+    # Off the production path: it must not block canary, and a targeted
+    # dispatch has to reach it with staging skipped. `!cancelled()` keeps
+    # the job alive through a skipped dependency, which a bare `needs`
+    # would not.
+    if: >-
+      !cancelled()
+      && needs.staging.result != 'failure'
```
esnunes: `needs.staging.result != 'failure'` admits three of the four possible results: `success`, `skipped` and `cancelled`. The `skipped` case is clearly wanted (it is what makes a targeted dispatch reach canary or production), but `cancelled` slips through with it, and `!cancelled()` does not catch it — that function reports whether the whole run was cancelled, not this dependency.

I think that is reachable rather than theoretical: job-level concurrency cancels a *pending* job when a newer one enters the same group, and this job shares `server-deploy-<env>` with hccm and csi. A csi staging job queueing at the wrong moment would evict cilium staging, and production would then roll the CNI on a change staging never applied — which is the one thing the cascade exists to prevent.

Would `contains(fromJSON('["success","skipped"]'), needs.staging.result)` work for you? It keeps the dispatch path and closes the cancelled case.

.github/workflows/cilium-apply.yml
```diff
@@ -0,0 +1,243 @@
...
+            -o custom-columns=DESIRED:.status.desiredNumberScheduled,READY:.status.numberReady,UNAVAILABLE:.status.numberUnavailable
+          UNAVAILABLE="$(kubectl -n kube-system get daemonset cilium -o jsonpath='{.status.numberUnavailable}')"
+          if [ -n "${UNAVAILABLE:-}" ] && [ "$UNAVAILABLE" != "0" ]; then
+            echo "ERROR: $UNAVAILABLE cilium agents still unavailable after rollout" >&2
+            exit 1
+          fi
+          kubectl get nodes -l kubernetes.io/os=linux --no-headers \
+            | awk '$2 != "Ready" { print; bad=1 } END { exit bad+0 }' \
+            || { echo "ERROR: a linux node is not Ready after the Cilium rollout" >&2; exit 1; }
+          echo "All cilium agents ready and all linux nodes Ready."
```
esnunes: `$2 != "Ready"` is stricter than it looks. A cordoned node prints `Ready,SchedulingDisabled`, so it fails this test while being entirely healthy — and any linux node that was already NotReady before the run for unrelated reasons fails it too.

Since the check is absolute over the whole fleet rather than a before/after comparison, one unrelated sick node fails staging and the cascade then keeps canary and production from ever receiving the change. The `numberUnavailable == 0` assertion just above is the precise signal and already covers the rollout itself.

Two options that keep the intent: match `Ready*` with a prefix test so `SchedulingDisabled` passes, or snapshot the not-Ready linux nodes before the rollout and only fail on ones that became unready during it. This is the same reasoning the step already applies when it excludes the macOS nodes.

.github/workflows/cilium-apply.yml
```diff
@@ -0,0 +1,243 @@
...
+        # Writing the annotation is itself what triggers the rolling
+        # update, since it mutates the pod template. If helm ever strips
+        # the annotation, the next run reads it empty and re-rolls: an
+        # extra restart, never a missed one.
+        env:
+          FORCE_RESTART: ${{ inputs.force_restart }}
+        run: |
+          set -euo pipefail
+          DESIRED="$(kubectl -n kube-system get configmap cilium-config \
+            -o json | jq -S -c '.data' | sha256sum | awk '{print $1}')"
```
esnunes: I templated the chart to check the checksum claim and it holds — but it holds more broadly than the fix covers. 1.18.7 renders three ConfigMaps (`cilium-config`, `cilium-envoy-config`, `hubble-relay-config`) and *none* of the four pod templates carries a checksum annotation, so the same silent no-op applies to the `cilium-envoy` DaemonSet and the `hubble-relay` Deployment. Both are live here: the values file sets `envoy.enabled: true`, `l7Proxy: true`, `loadBalancer.l7.backend: envoy` and `hubble.relay.enabled: true`.

Definitely not a blocker, and shipping the agent half first is the right order given the values that have actually moved so far are all agent-side. But it might be worth either extending the hash-and-patch loop over the other two, or saying in the file that it covers the agent on purpose — otherwise the next person reasonably reads this workflow as full coverage and gets the same surprise one layer up.

infra/AGENTS.md
```diff
@@ -147,7 +147,7 @@ The previous "Tailscale ACL audit log" trail no longer applies — the ACL is no
...
+- **kube-system add-ons** (Cilium, HCCM, node CSI drivers) live in the per-workload bootstrap (`mgmt/bootstrap/`), but bootstrap only runs at cluster onboarding. Edits to their values reach existing clusters via dedicated re-apply workflows so they don't silently drift: `.github/workflows/hccm-deployment.yml` (HCCM), `.github/workflows/csi-deployment.yml` (re-applies `hcloud-csi` with its pool-exclusion affinity and uninstalls the now-obsolete `scaleway-csi` on the Elastic Metal kura fleets), and `.github/workflows/cilium-deployment.yml` (Cilium). All three share the `server-deploy-<env>` concurrency lane. HCCM and CSI fan out over all envs in parallel; **Cilium cascades staging → canary → production** and ends with an explicit `rollout restart` of the agent DaemonSet, because it is the CNI and because most of `cilium-values.yaml` renders into the `cilium-config` ConfigMap, over which the agent pod template holds no checksum annotation — a bare `helm upgrade` there rewrites the ConfigMap, rolls nothing, and reports success while every agent keeps its old in-memory config.
```
esnunes: Small doc drift: this line says the workflow "ends with an explicit `rollout restart` of the agent DaemonSet", which matches the first commit but not where it landed — the roll is now gated on the config-hash annotation, and `rollout restart` is only reached under `force_restart` with an unchanged hash. Worth correcting because "restarts every time" and "restarts when the config actually changed" set quite different expectations for a CNI.

Same sentence gives the cascade as staging → canary → production and leaves out `preview`, which the workflow does reconcile (off the production path).

esnunes (approved):
This is a well-argued change and the central claim holds — I templated cilium/cilium 1.18.7 against the repo's values file and there is no checksum annotation anywhere in the chart, so a plain `helm upgrade` really would rewrite `cilium-config` and roll nothing. Reading the hash off a persisted pod-template annotation rather than diffing within the run is the right call, and the comment explaining why is the kind I wish more workflows had. Using `needs` instead of a matrix for the ordering, and reading the chart version from one place, are both good.

Most of my notes are about the edges rather than the core. The two I would most like your read on: the `server-deploy-<env>` concurrency group does not actually overlap `server-deployment.yml`'s jobs (they all carry suffixes), so the mutual exclusion the comment promises is not there; and `needs.staging.result != 'failure'` admits `cancelled` alongside `skipped`, which would let production roll the CNI on a change staging never applied. There is also a node-readiness gate that fails on a cordoned node, and the same no-checksum problem still sitting under `cilium-envoy` and `hubble-relay`. None of it blocks merging.

One thing I could not settle from the repo: whether any cluster carries hand-set Cilium values beyond `k8sServiceHost`/`Port` — the values file itself suggests bumping `operator.replicas` to 2 for production "via overlay", and no such overlay exists here. This workflow re-applies the file verbatim, so any such value would be silently reset. If that is a real possibility, logging `helm get values` before the upgrade would make it visible in the job log rather than a surprise later.

.github/workflows/cilium-apply.yml
fortmarek: You're right, the comment promised something the group does not deliver: `server-deployment.yml` suffixes its lanes (`-deploy`, `-observability`, `-tailscale`, `-platform`), so this only serializes against hccm and csi. Rewrote the comment to state exactly that and to flag the server `helm --atomic` rollback risk; left the shared-group fix out as it belongs across all four workflows rather than here.

.github/workflows/cilium-deployment.yml
fortmarek: Agreed, and thanks for catching that `!cancelled()` does not cover a cancelled *dependency*. Switched all three gates to `contains(fromJSON('["success","skipped"]'), needs.X.result)` as suggested, which keeps the targeted-dispatch path and closes the eviction case you describe.

.github/workflows/cilium-apply.yml
fortmarek: Both problems confirmed. Took the snapshot option: the run now records not-Ready linux nodes before the rollout and fails only on ones that regressed during it, matching `Ready*` by prefix so `Ready,SchedulingDisabled` passes. Same reasoning the step already used to exclude the macOS nodes, as you note.

.github/workflows/cilium-apply.yml
fortmarek: Confirmed by templating with the real values file, and it reaches one workload further than reported: `cilium-operator` also consumes `cilium-config`, and `cilium-envoy` mounts `cilium-envoy-config` as a volume. Rather than document the limit I generalised the hash-and-patch into a loop over every (ConfigMap, workload) pair, skipping any absent from a given cluster.

infra/AGENTS.md
fortmarek: Both corrected: the line now describes the hash-gated roll rather than an unconditional `rollout restart`, and mentions `preview` hanging off staging outside the production path. Also noted there that the shared lane does not cover `server-deployment.yml`.

PR #12484, opened 2026-08-19
fix(server): fail a sharded test run as soon as a shard reports a failure

fortmarek:
### Context

The acceptance tests on [#12464](https://github.com/tuist/tuist/pull/12464) failed, but the Tuist run report comment showed ⏳ instead of ❌, and the [test run](https://tuist.dev/tuist/tuist/tests/test-runs/7201c124-92c8-41e1-a369-576e29ad3c3e) reported shard 0 as in progress long after its job had finished.

The comment was not misrendering. The run genuinely was `in_progress` in ClickHouse. Two independent bugs kept it there, and both are fixed here.

### Cause 1: a lost shard report held the run open

Shard 0 of that run never reported. Its [job](https://github.com/tuist/tuist/actions/runs/32263978415/job/96110507080) was green, but its last lines were:

```
Failed to upload test results: The test run could not be uploaded due to an unknown server response of 500.
✔ Success  The project tests ran successfully
```

The CLI logs a failed test-run upload as a warning and still exits 0, so the job stayed green. Production `shard_runs` confirms it: only `shard_index: 1` ever wrote rows, `processing` then `failure`.

That 500 was deploy-window skew rather than a live bug. `Ch.Error Code 47 ... Identifier 't0.only_test_identifiers' cannot be resolved` (Sentry TUIST-15V / TUIST-170) fired while the `Test` schema carried `only_test_identifiers` but the ingest migration had not yet reached production, fallout from the duplicated migration version renumbered in #12480. The column is present in production now, so that half is already resolved and this PR does not touch it.

What it exposed is the durable bug. `create_or_update_sharded_test/1` only computed a merged status once `reported_count >= expected_shard_count`, returning `in_progress` for anything less. Shard 1 had already reported a hard `failure`, but with shard 0 missing the run stayed `in_progress` until the six-hour reaper.

A `failure` or `failed_processing` from any shard now decides the merged run immediately. Failure is terminal in this model, since `compute_final_shard_status/1` already ranks it above success, so no later shard can turn a failed run back into a passing one and gating it behind a complete count only delays a verdict that cannot change. Success still requires every shard to have reported, so a run is never called green early. The same rule now applies on the create path, where a first shard arriving with a failure previously produced an `in_progress` run.

### Cause 2: the reporting shard did not count itself

Fixing the above was not enough, and the second cause turned out to be the larger one. Sharded runs were getting stuck at `in_progress` even when every shard reported success. Seven of the eight sharded runs on tuist/tuist in the twelve hours before this PR sat at `in_progress` with both shards terminal in `shard_runs`:

| run | expected shards | shards terminal in `shard_runs` | run status |
|---|---|---|---|
| `f09808a8` | 2 | 2 (both `success`) | `in_progress` |
| `03f03f3e` | 2 | 2 (both `success`) | `in_progress` |
| `2120e536` | 2 | 2 (both `success`) | `in_progress` |
| `7201c124` | 2 | 1 (`failure`) | `in_progress` |
| `8802b505` | 2 | 2 (both `success`) | `success` |

The merge inserted the reporting shard's row and then counted the shards by reading `shard_runs` back. That read does not reliably observe the row the same request just wrote: on `f09808a8` the last shard wrote its `success` row at `15:22:20.417` and the merge wrote `in_progress` at `15:22:20.505`, 88ms later, having counted only one shard. The last shard to report therefore counted itself as still processing and fell one short. Nothing reports after the last shard, so the run stayed `in_progress` until the reaper. `8802b505` came out `success` only by winning the same race.

The reporting shard now contributes its status from memory and the query supplies only the other shards, which removes the dependency on reading back the write. This also means a shard whose row is genuinely absent no longer changes the outcome, since its own status is known in memory.

I have not asserted the storage-level reason the read-back is unreliable. Both repos point at the same host and `shard_runs` is a plain `MergeTree`, so the likely explanation is replica routing or server-side `async_insert` that is not visible in this repo. The fix holds under any of them, which is why it is written as "do not re-read what you just wrote" rather than as a tuning change.

Together the two fixes cover every stuck run observed: the seven fully-reported ones by the in-memory count, and `7201c124` by the failure short-circuit.

### Dashboard

The shard table infers unreported shards from the plan's `shard_count`, so a shard that never reported still gets a row. Those rows read "Pending" and "In Progress", which is honest while the run is live but wrong once the run is over and no further report is coming. They now render "Not reported" when the run itself has reached a terminal status.

### Impact

Sharded runs reach their real status when the last shard reports instead of hanging for up to six hours, so the pull request comment and the dashboard stop showing an hourglass over finished runs. Runs that lose a shard's report surface as failed as soon as another shard fails.

### Known limitation

Two shards reporting close enough together that neither observes the other's row would still leave the run `in_progress`, and that case remains covered only by the six-hour reaper. It was not observed in any of the runs examined here, where shards reported minutes apart, and closing it needs a delayed reconciliation pass rather than a change to this decision.

### Not addressed

The CLI treats a failed test-run upload as a warning and exits 0, which is why cause 1 went unnoticed until the comment looked wrong. Changing acceptance-test exit semantics is a larger call than this fix and is left out deliberately.

### How to test locally

Each of the six new tests was confirmed red on the same assertion before the fix. The one covering cause 2 reproduces the production symptom directly, stubbing the shard-row write so the read back cannot see it and asserting the merged run still completes.

```
cd server
mix test test/tuist/tests_test.exs test/tuist_web/live/test_run_live_test.exs
```

Full validation that was run:

- `mix test` across `tests_test.exs`, `test_run_live_test.exs`, `shards_test.exs`, `shards/analytics_test.exs`, `tests_controller_test.exs`, `xcresult_processing_test.exs`, `process_xcresult_worker_test.exs`, and `vcs_test.exs` with `--warnings-as-errors`: 534 passed.
- `mix format --check-formatted` and `mix credo` on the changed files: clean.
- `mix gettext.extract` for the new "Not reported" string; only `dashboard_tests.pot` is committed, and the unrelated line-number churn in `dashboard_runners.pot` was reverted.


esnunes:
starting review...

server/lib/tuist_web/live/test_run_live.ex
```diff
@@ -1506,12 +1506,17 @@ defmodule TuistWeb.TestRunLive do
 
     reported_indices = MapSet.new(reported_shards, & &1.shard_index)
 
+    # A run that has reached a terminal status is never going to hear from the
+    # shards still missing, so calling them pending reads as work in flight on
+    # a run that is already over.
+    unreported_status = if transient?(run.status), do: "pending", else: "not_reported"
```
esnunes: One thing I wanted to flag about how the two halves of this PR meet here. `merged_shard_status/3` now makes the run terminal as soon as any shard reports `failure`, and `transient?/1` is only `processing`/`in_progress` — so from that moment every shard the server hasn't heard from renders the red "Not reported" badge, loses its "In Progress" duration cell, and the "The shards are in progress…" banner disappears with it (it's gated on `@run.status == "in_progress"`).

A shard that's still executing its tests has no `shard_runs` row yet — `shard_metrics/1` only returns rows that exist, and a shard's first POST is when it finishes and starts uploading. So an in-flight shard lands in exactly that bucket. On a four-shard plan where shard 0 fails at two minutes and shards 1–3 finish at twelve, the dashboard spends ten minutes asserting that three running shards never reported. It doesn't correct itself inside a session either: `assign_shard_rows/2` runs only from `mount/3` and `reload_run_state/1` doesn't re-run it, so the table keeps its mount-time values.

The label reads right for the reaper case it was written for; the early-failure case this PR introduces is the one it can't tell apart. Would a time-based gate work — "Not reported" only once the run is older than the reaper window — or would you rather keep the amber "Pending" while the run is `failure` but still young?

esnunes (approved):
Really enjoyed reading this one — the write-up does the thing I wish more fix PRs did, which is separate the two causes and say plainly which part is asserted and which part isn't. "Do not re-read what you just wrote" is exactly the right shape for cause 2: it holds whatever the storage-level explanation turns out to be, so the fix doesn't depend on a diagnosis you didn't make. The asymmetry in `merged_shard_status/3` reads as correct to me too — `compute_final_shard_status/1` already ranks failure above success, so short-circuiting a failure only skips a wait whose outcome can't change, while success still needs a complete count.

One thing I wanted to flag, and it isn't blocking: because a run can now go terminal while other shards are genuinely still running, the new "Not reported" label — and the loss of the "shards are in progress" banner — can land on shards that are mid-execution, and the shard table doesn't recompute after mount, so it stays wrong for the rest of the session. The label is right for the reaper case it was written for; it just can't tell that case apart from the early-failure case this PR makes common. I left an inline note with the detail.

Tests, formatting, and the `.pot`-only gettext handling all look right to me.

server/lib/tuist_web/live/test_run_live.ex
fortmarek: Good catch, and it applied to more than the label: an unreported shard now stays "Pending" until the run outlives the reaper window (both sides read it from `Tests.stale_run_window_hours/0`), and `reload_run_state/1` now rebuilds the shard rows, which it didn't before, so the table no longer holds its mount-time values for the session. Fixed in 101823960e.

PR #12485, opened 2026-08-19
fix(server): stamp the merged scheme on a sharded run's test case runs

fortmarek:
Test case runs on sharded test runs were persisted with an empty `scheme`, so the dashboard showed no scheme for roughly half the runs of any sharded test case (for example [this test case](https://tuist.dev/tuist/tuist/tests/test-cases/50cfea1e-9bcf-47f7-9e11-fe98e3125b81), where 267 of 576 runs are blank, all of them since 2026-06-18).

### Root cause

A sharded run's `Test` row is created by the first shard's controller upload with `status: "processing"`. That payload carries the git fields but no scheme: the scheme is `parsed_data["test_plan_name"]`, which only exists after a worker has parsed that shard's xcresult.

Whichever shard's worker finished parsing first then took the existing-run branch of `create_or_update_sharded_test`. That branch called `create_test_modules(existing_test, ...)` with the placeholder struct and applied `merge_shard_metadata(attrs)` only further down, when rewriting the `Test` row. So that shard's test case runs were stamped with the placeholder's blank scheme, while the run row picked up the real one a few lines later. Every later shard read a run row that already had the scheme, which is why the corruption covers part of a run rather than all of it.

Production confirms the shape. On run `f57cd43f-aaf5-4028-a0ed-54ddd9fa0016`, shard 0's 281 test case runs carry `TuistAcceptanceTests` and shard 1's 222 are blank, while `git_branch` and `git_commit_sha` are populated on both, and the parent `test_runs` row has the scheme.

### What changed

`create_or_update_sharded_test` now merges the incoming shard metadata into the existing run before the test case runs are built, and reuses that merged struct for the run row rewrite. `merge_shard_metadata/2` still only fills blanks, so a shard reporting a different scheme cannot overwrite an established one.

### Why it matters beyond display

`check_cross_run_flakiness` resolves sibling runs on `(project_id, git_commit_sha, scheme, is_ci)`, and `TestCaseRunByCommit` is keyed the same way. A failure written with a blank scheme could never match its passing sibling in another shard of the same run, so same-commit flaky detection was silently blind to the shard that supplied the scheme. That is fixed by the same change.

Existing rows are not backfilled. The fix is forward only; correcting the historical rows would mean re-inserting into the ClickHouse `test_case_runs` table and its projections, which is worth handling separately.

### How to test locally

```bash
mix test test/tuist/tests_test.exs
```

The new case, `stamps the merged scheme on the test case runs of the shard that supplies it`, reproduces the production sequence: a placeholder shard with no scheme, then a parsed shard that carries one and reports test modules. It fails on `main` with `left: [""]`.

Validation run for this change:

- `mix test test/tuist/tests_test.exs`, 280 passed
- `mix test test/tuist_web/controllers/api/tests_controller_test.exs test/tuist/tests/workers/process_xcresult_worker_test.exs test/tuist/shards_test.exs`, 106 passed
- `mise run format`, clean


esnunes:
starting review...

esnunes (approved):

PR #12487, opened 2026-08-19
fix(server): seed test case durations inside the ClickHouse memory budget

fortmarek:
The production deploy has been blocked since #12435 landed. Its
`pre-upgrade` migration Job dies backfilling
`test_case_duration_daily_stats_per_case`, so helm rolls the release
back and nothing ships:

    Code: 241. DB::Exception: User memory limit exceeded: would use
    8.02 GiB, maximum: 8.00 GiB. OvercommitTracker decision: Query was
    selected to stop by OvercommitTracker. (MEMORY_LIMIT_EXCEEDED)

Two things were wrong.

The backfill walked every partition of `test_case_runs`, back to the
table's first month. Nothing reads that data: `Tuist.Tests.list_test_cases/3`
is the table's only reader, it filters `date >= today - 14`, and it
deliberately exposes no date-window option, so no query can reach a row
older than the trailing window. It now seeds a 30-day window, which
covers the read window with room for the view to fill forward past the
seam long before the read window arrives at it.

The insert also carried `max_memory_usage = 12 GB`, above the 8 GiB
`max_memory_usage_for_user` it shares with the running server. A ceiling
above the shared budget cannot bind, so the query grew into the pool the
live server needs and the overcommit tracker picked it off. It now takes
1 GiB and spills to disk at 256 MiB.

A ceiling alone would not be enough. The same range passes or fails
depending on what else the replica is doing, which is why the identical
retries here cleared nothing across four attempts and 20 minutes: an
identical retry clears a marginal overrun but never a large one. This is
the lesson the sibling backfill's moduledoc left behind after it failed
four production deploys the same way on 2026-08-18. A range that
exhausts its retries is now halved by date and each half attempted in
turn, down to a single day; halves are disjoint in `date`, a grouping
key, so every run lands in exactly one group however the range is cut. A
single day that still does not fit is logged and skipped rather than
raised, so one project's worst day costs samples instead of blocking
every server release.

`up/0` also now drops the table before creating it. The Job retries, and
a failed attempt leaves behind the ranges it had already inserted, which
`IF NOT EXISTS` then duplicated on the next attempt. `run_count` survives
that (`uniqExact` over run ids collapses the copies) but the average and
the quantile reservoirs weighted the re-inserted runs once per attempt.
Production has already been through four such attempts, so it needs the
empty start as much as any future retry does.

Validation: a new integration test seeds runs inside and outside the
window, runs `up/0` twice, and asserts the states equal a direct
aggregation over the window and do not move on the second run. It fails
against the previous migration on both counts (it backfills the
out-of-window test case, and the states shift). Against a local
ClickHouse holding a production-shaped `test_case_runs`, the windowed
insert reproduces a direct aggregation exactly over 2325 groups, and
splitting the same range by date reproduces it again, group for group.

### DIFF
--- a/server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
+++ b/server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
@@ -69,6 +69,24 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv 
   empty table, so the states describe each run exactly as many times as the
   source holds it.
 
+  Nothing below that granularity needs the same treatment. A retry and a split
+  both re-cover ground a failed insert had been working on, which would double
+  those states if a failed `INSERT ... SELECT` committed the part of the result
+  it had produced. It does not: parts are written under temporary names and
+  become visible together when the statement completes. Measured on this exact
+  query shape against a local ClickHouse, a 2.1M-row insert broken off
+  mid-result left no rows behind, for a raised exception and for a genuine
+  `MEMORY_LIMIT_EXCEEDED` alike, and no part of it was visible while it ran.
+  So the unit that can be duplicated is a range that finished, and the table
+  drop is where that is handled.
+
+  The price is that a Job retry restarts the pass instead of resuming it. That
+  is affordable at this size: the window selects a single partition outside the
+  few days a month it straddles two, and a pass over it is a few hundred
+  single-project inserts. Resuming would mean recording which ranges finished,
+  and the only honest place to record it is a table this migration would have
+  to create and clean up for the sake of saving that minute.
+
   ## What the backfill covers
 
   The backfill seeds a trailing `@backfill_window_days` window rather than
@@ -126,7 +144,7 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv 
   @disable_migration_lock true
 
   @backfill_window_days 16
-  @project_throttle_ms 250
+  @attempt_throttle_ms 250
   @range_attempts 2
   @max_memory_usage 1_073_741_824
   @max_bytes_before_external_group_by 268_435_456
@@ -235,7 +253,6 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv 
 
       Enum.each(project_ids, fn project_id ->
         backfill_range(partition, project_id, window_start, window_end, boundary)
-        Process.sleep(@project_throttle_ms)
       end)
     end
   end
@@ -283,6 +300,11 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv 
          boundary,
          attempts \\ @range_attempts
        ) do
+    # Ahead of every attempt rather than between projects: the halves a split
+    # produces are fired at a replica that has just refused the range they came
+    # from, which is when leaving it room matters most.
+    Process.sleep(@attempt_throttle_ms)
+
     insert_range(partition, project_id, from_date, to_date, boundary)
   rescue
     e in Ch.Error ->
--- a/server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
+++ b/server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
@@ -77,9 +77,23 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv 
   and it deliberately exposes no date-window option, so no query can reach a
   row older than that window. Seeding all history therefore aggregated years of
   runs nothing would read, and it was that volume rather than the feature that
-  could not fit in the ClickHouse memory budget. The seeded window is wider
-  than the read window, so the view has filled forward well past the seam long
-  before the read window arrives at it.
+  could not fit in the ClickHouse memory budget.
+
+  The seam sits at the day this runs: the backfill ends there and the view
+  carries everything after it. A reader that day asks for the 14 days before
+  it, and every later day asks for a window the view has already covered more
+  of, so the widest gap the backfill ever has to close is the read window
+  itself. The extra days on top of it are slack, not coverage.
+
+  The window cannot go to zero. `list_test_cases/3` renders a test case below
+  the sample floor as "N/A" and explains it as "has run 0 times in the last 14
+  days", which of an unseeded row is untrue rather than merely incomplete, and
+  the Tests overview drops those rows from "Slowest test cases" entirely. The
+  projects that would sit that way longest are the ones that run their suites
+  least often: sampled against the current data, the four largest projects
+  recover 94-99% of their populated duration cells after a single day of
+  filling forward, while several low-frequency projects recover none of theirs
+  in seven.
 
   ## Surviving the memory budget
 
@@ -111,7 +125,7 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv 
   @disable_ddl_transaction true
   @disable_migration_lock true
 
-  @backfill_window_days 30
+  @backfill_window_days 16
   @project_throttle_ms 250
   @range_attempts 2
   @max_memory_usage 1_073_741_824
--- a/server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
+++ b/server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
@@ -55,10 +55,52 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv 
   effect is a weighting bias toward corrected runs rather than an invented
   number. Removing it needs a per-run deduplicated source, which every other
   aggregate over this table would want too and which does not belong in this
-  migration. `FINAL` in the backfill would not close it either: the backfill is
-  chunked by partition to survive memory limits, and a correction lands in a
-  later `toYYYYMM(inserted_at)` partition than the run it corrects, so the two
-  versions are never read together.
+  migration.
+
+  ## Starting from an empty table
+
+  `up/0` drops the table and the view before creating them. This runs as a helm
+  `pre-upgrade` hook whose Job retries, and a failed attempt leaves behind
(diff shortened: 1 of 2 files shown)


server/test/tuist/ingest_repo/create_test_case_duration_daily_stats_per_case_mv_integration_test.exs
```diff
@@ -0,0 +1,159 @@
...
+      date: date,
+      is_ci: is_ci,
+      run_count: run_count,
+      p50: round(p50),
+      avg: round(avg)
+    }
+  end
+
+  defp run_attrs(project_id, test_case_id, date, duration, is_ci) do
+    ran_at = NaiveDateTime.new!(date, ~T[12:00:00.000000])
```
esnunes: One thing I wanted to flag about the seeded timestamps: `run_attrs/5` stamps `ran_at` (and `inserted_at`) at 12:00 UTC, so the `days_ago = 0` rows land at today 12:00. The backfill only takes `inserted_at < boundary` with `boundary = now64(6)`, so when the test runs before midday UTC those three rows are still in the future and the insert skips them — and the view can't pick them up either, since they were written before it existed. `expected_stats/2` has no `inserted_at` bound and still counts them, so I think that makes the first assertion time-of-day dependent. Stamping `inserted_at` with something always in the past (`NaiveDateTime.utc_now()` for the today rows, or `~T[00:00:01]`) would fix it, or `expected_stats/2` could carry the same boundary filter. Worth trying against a morning run?

server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
```diff
@@ -131,74 +196,134 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv
...
 
+  defp backfill_range(
+         partition,
+         project_id,
+         from_date,
+         to_date,
+         boundary,
+         attempts \\ @range_attempts
+       ) do
+    insert_range(partition, project_id, from_date, to_date, boundary)
```
esnunes, line 308: The "Starting from an empty table" section convinced me, and then made me wonder about the same hazard one level down. Inside a single run, the retry re-runs the identical range and the split re-runs the same dates in two halves — both against a table that may already hold part of the failed attempt's output. `INSERT ... SELECT` commits block by block rather than atomically, and with `optimize_aggregation_in_order = 1` over `(project_id, test_case_id)` — a prefix of the current sort key, and the only ordering available since the projection didn't survive the table rebuild — finished groups can stream out during the scan instead of all landing at the end. If a partial write is reachable (the read-only path looks like the likeliest way), the re-insert double-counts: `run_count` is fine since uniqExact collapses the run ids, but `avgState` and the quantile reservoirs weight those runs twice — the same asymmetry you describe for corrections. Do you read it the same way? If so, a narrow `ALTER TABLE ... DELETE WHERE project_id = ... AND date BETWEEN ...` before each attempt, or staging into a side table and EXCHANGE TABLES at the end, would close it — and would also make the table-wide drop unnecessary. Failing that, it'd be worth the moduledoc not claiming the retry path is covered.

server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
```diff
@@ -69,13 +125,18 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv
...
-  @project_chunk_size 1
-  @chunk_throttle_ms 250
+  @backfill_window_days 16
+  @project_throttle_ms 250
+  @range_attempts 2
+  @max_memory_usage 1_073_741_824
+  @max_bytes_before_external_group_by 268_435_456
 
   def up do
+    drop_objects()
```
esnunes, line 153: A question about the time budget rather than the logic — and I checked the throttle half before asking, so scratch that part: `today - 16` currently lands inside the same month, so `partitions_since/1` selects a single partition and each project is visited once, and for the number of projects the window actually selects the 250 ms adds up to a negligible share of the release budget even in the straddling case.

What I'd still raise is the other half. Dropping first buys idempotence, but it also discards everything the previous attempt inserted, so `backoffLimit: 6` gives seven fresh starts rather than seven chances to finish — the whole window has to land in one uninterrupted pass, inside the release's `--atomic --timeout 60m`. The last production run got through four Job attempts in about twelve minutes, so there's room, but a transient error late in a pass still costs the entire pass rather than just the range it hit. Was resumability something you weighed against the double-counting risk, or did the drop just come out clearly ahead?

server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
```diff
@@ -69,13 +125,18 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv
   @disable_ddl_transaction true
   @disable_migration_lock true
 
-  @project_chunk_size 1
-  @chunk_throttle_ms 250
+  @backfill_window_days 16
```
esnunes, line 146: I went and checked this one, and I think it holds. On `b8dcffd518` and `711a661a` — both of which carry this migration — `canary / Deploy to tuist-canary` succeeded while `production / Deploy to tuist` failed at the `pre-upgrade` hook. So canary has already recorded version 20260818130000 with the previous version of this file, which means none of the new behaviour (window, drop, subdivision, ceilings) will execute there: production is the first and only environment to run the rewrite, and the canary → acceptance → production cascade gives it no cover. It also leaves canary holding a full-history table while production gets 16 days, though that part is harmless — the reader never reaches past `today - 14`, so the extra rows are inert. Would it be worth deleting that row from canary's `schema_migrations` before the deploy, so the new path runs once with something ahead of production?

server/priv/ingest_repo/migrations/20260818130000_create_test_case_duration_daily_stats_per_case_mv.exs
```diff
@@ -131,74 +196,134 @@ defmodule Tuist.IngestRepo.Migrations.CreateTestCaseDurationDailyStatsPerCaseMv
...
       )
 
-      project_ids
-      |> Enum.chunk_every(@project_chunk_size)
-      |> Enum.each(fn chunk ->
-        retry_on_transient_failure(fn -> backfill_chunk(partition_int, chunk, boundary) end)
-        Process.sleep(@chunk_throttle_ms)
+      Enum.each(project_ids, fn project_id ->
+        backfill_range(partition, project_id, window_start, window_end, boundary)
+        Process.sleep(@project_throttle_ms)
```
esnunes: Small one: `Process.sleep(@project_throttle_ms)` sits in the per-project loop only, so once a range subdivides the sub-inserts fire back to back with no pause — which is the case where the replica has just told us it couldn't fit the work. The 5s sleep on the retry branch covers the retry but not the descent. Moving the throttle into `insert_range/5`, or adding one to the split branch, would keep the pacing honest where it matters most.

esnunes (approved):
This is a really good rewrite, and the moduledoc is the best part of it. The seam argument — backfill to `today - 16`, the view from the boundary on, the reader never reaching past `today - 14` — checks out against `lib/tuist/tests.ex:68` and `:2871`, and the ceiling change is plainly right: a `max_memory_usage` above `max_memory_usage_for_user` can't bind at all, so the old 12 GB was never a limit. Bisecting by `date` is safe for the reason you give (it's a grouping key and the halves are disjoint), the recursion terminates cleanly at a one-day range, and the partition/`ran_at` split still visits every run exactly once. Carrying the sibling migration's "retry *and* subdivide" lesson forward is the right read of it. Approving — none of the below blocks it.

The one I'd most like your read on is idempotency one level below the table drop: inside a single run, the retry and the split re-insert over whatever the failed attempt already wrote, and `INSERT ... SELECT` commits per block rather than atomically. `run_count` survives that, `avgState` and the reservoirs wouldn't.

In the test, `run_attrs/5` stamps `inserted_at` at 12:00 UTC, so the `days_ago = 0` rows are in the future for any run before midday and the backfill's boundary filter skips them while the expected-value query counts them — I think that makes the first assertion time-of-day dependent.

Two things I went and checked rather than just asking. Canary has already recorded this migration version: on `b8dcffd518` and `711a661a`, both carrying this migration, `canary / Deploy to tuist-canary` succeeded while `production / Deploy to tuist` failed at the `pre-upgrade` hook — so production is the first and only environment that will run the rewrite. And the throttle is not the constraint: the window currently selects a single partition, and the 250 ms per project is a negligible share of the release budget even in the straddling case. What's left of that concern is that dropping first means each Job retry restarts the whole pass rather than resuming it.

The last small one is that the throttle is skipped on the subdivision path, which is exactly when the replica is under most pressure.

fortmarek:
Thanks — all four looked at, two changed, one I went and measured because I think the premise doesn't hold, one acknowledged in the moduledoc.

**Partial inserts inside a run.** I couldn't reproduce it. `INSERT ... SELECT` into this table writes its parts under temporary names and makes them visible together when the statement completes, so a broken-off insert commits nothing and the retry and the split have nothing to double. Against a local ClickHouse holding a production-shaped `test_case_runs`, on this exact query shape at 2.1M output rows with `max_insert_block_size = 10000` and block squashing off:

- thrown mid-result at 90% of the id space: `Code: 395`, 0 rows left, 0 active parts
- real `MEMORY_LIMIT_EXCEEDED` on a ceiling that trips mid-run: `Code: 241`, 0 rows left
- polling `system.parts` every 0.7s through a ~6s successful insert: 0 parts, 0 rows throughout, then 2.1M at completion

So the unit that can be duplicated is a range that *finished*, which is the Job-attempt case the table drop already covers. I've written the measurement into the moduledoc rather than leaving it as a claim, since it's the kind of thing worth not re-deriving. Worth saying the scope: single MergeTree destination, no MV reading this table. If you know of a shape where it does commit progressively I'd rather add the cleanup than be wrong about this one.

**The 12:00 UTC stamp.** You're right, and it was live — the suite passed yesterday only because I ran it after midday. Reproduced at 08:20 UTC this morning, failing on exactly the `days_ago = 0` rows: the boundary filter skips them, the expected-value query counts them. Runs are now counted back from the current instant, so they can't land in the future. Red before, green after, same hour.

**Throttle on the subdivision path.** Fixed — it now runs ahead of every attempt rather than between projects. Same cost per project, and the pause lands where you point out it matters: the halves were being fired back to back at a replica that had just refused the range they came from.

**Restart rather than resume.** Accepted as the price of the empty start, and now stated as such in the moduledoc. It's a few hundred single-project inserts over one partition, and resuming would need a progress table this migration would have to create and clean up. Thanks also for checking canary — production being the only environment that runs the rewrite matches what I saw on `b8dcffd518`, and it's the reason the drop is scoped the way it is.

PR #12490, opened 2026-08-20
fix(server): authenticate ClickHouse dictionary sources used by backfill migrations

fortmarek:
## What changed

Every ClickHouse backfill migration that builds a dictionary over a local table now names the credentials the repo already connects with, instead of leaving the source anonymous.

- New `Tuist.ClickHouseDictionarySource.local_table/2` renders the `SOURCE(CLICKHOUSE(...))` clause from `repo.config()`, escaping the user and password so neither can break out of its SQL literal. When the repo authenticates with no user (dev, test) it emits the bare clause it emits today, so nothing changes there.
- The five migrations that create such a dictionary use it: `20251118211224`, `20260422120001`, `20260422120002`, `20260609120000`, and `20260721115000`.
- Those `CREATE DICTIONARY` calls pass `log: false` so the password does not reach the query log.

Host and port stay out of the clause on purpose. Omitted, they default to the server's own address and native port, which is what a same-instance lookup wants; the repo's port is the HTTP one the source would not dial anyway.

## Why

A clean disaster-recovery restore on a self-hosted instance got past the projection migrations fixed in #12468 and then died on `20260721115000_add_project_id_to_test_case_events` with:

```
Code: 516. DB::Exception: Authentication failed: password is incorrect, or there is no user with such name
```

### Root cause

A `CLICKHOUSE` dictionary source is resolved inside the ClickHouse server, not by the client that issued the `CREATE DICTIONARY`. The server therefore authenticates it separately from the connection that created it, using the user named in the source. With no user named, it falls back to `default` with an empty password. That succeeds on a server with no password set, which is why this never surfaced on the managed instance or in dev and test, and fails on any install that sets one, including the published self-hosted Compose setup.

`CREATE DICTIONARY` itself does not fail, because dictionaries load lazily. The failure lands on the first `dictGet` inside the backfill mutation, which is why it reads as a migration failure rather than a configuration one.

### Why this shape

The obvious alternative was to drop dictionaries entirely and back these columns with a `Join` engine table plus `joinGet`, which needs no credentials at all. That rewrites the memory and failure profile of five migrations that have already run against the production dataset, to fix an authentication problem. Naming the credentials keeps the reviewed backfill behaviour intact and matches the precedent already in this directory, where the `postgresql()` backfills pass their credentials explicitly.

All five were fixed rather than just the reported one. A fresh self-hosted install runs the whole chain, and the earlier four only escaped the report because the tables they read were empty at that point in the restore, so no `dictGet` ever forced a load.

## Impact

Self-hosted instances with a password-protected ClickHouse can complete a restore-and-upgrade through the full migration chain. Instances that have already run these migrations are unaffected; the migrations are not re-run. No schema or data-shape change, so `server/data-export.md` is unchanged.

## Validation

- A regression guard asserts no migration builds a dictionary whose source names no user. It fails against the pre-fix migrations, naming all five, and passes after. Dev, test and the managed instance all run a passwordless ClickHouse, so no environment we migrate against exercises this authentication; the source text is the only place it is catchable.
- New unit tests for `local_table/2` cover the no-credentials, credentials, empty-password, and quote/backslash-escaping cases.
- Reproduced the reported failure against a local ClickHouse 26.1: a dictionary whose source names a wrong password fails its first `dictGet` with exactly `Code: 516 ... (AUTHENTICATION_FAILED)`, confirming the source authenticates independently of the creating connection.
- Ran the `20260721115000` backfill path end to end against a password-protected ClickHouse user with `Tuist.IngestRepo` configured by URL: the helper emitted `CLICKHOUSE(TABLE 'test_cases' USER '...' PASSWORD '...')`, the dictionary loaded, and the mutation backfilled `project_id` correctly.
- Verified a password containing a single quote and a backslash survives the emitted literal and authenticates.
- `mix credo` reports no issues, `mix format --check-formatted` passes, and the changed files compile without warnings.

## Follow-up

The remaining projection migrations in the restore range were checked for the Error 344 class that #12468 fixed. `cas_outputs` and `command_events` are plain MergeTree, and `test_runs` already inherits `deduplicate_merge_projection_mode = 'rebuild'` from `20260529100000`, so none of them need the same treatment.

### DIFF
--- a/server/lib/tuist/clickhouse_dictionary_source.ex
+++ b/server/lib/tuist/clickhouse_dictionary_source.ex
@@ -14,7 +14,20 @@ defmodule Tuist.ClickHouseDictionarySource do
   Host and port stay out of the clause on purpose. Omitted, they default to the
   server's own address and native port, which is what a same-instance lookup
   wants; the repo's port is the HTTP one the source would not dial anyway.
+
+  Issue the resulting statement with `query_opts/0`, which keeps the credential
+  out of both logs that would otherwise receive it.
   """
+
+  @doc """
+  Query options for a statement rendered with `local_table/2`.
+
+  `log: false` keeps the credential out of the application log. `log_queries: 0`
+  keeps the whole statement out of `system.query_log`: ClickHouse masks
+  dictionary source passwords there on its own, but it does so from the parsed
+  statement, so anything that fails to parse is logged verbatim instead.
+  """
+  def query_opts, do: [log: false, settings: [log_queries: 0]]
 
   def local_table(repo, table) do
     config = repo.config()
--- a/server/lib/tuist/clickhouse_dictionary_source.ex
+++ b/server/lib/tuist/clickhouse_dictionary_source.ex
@@ -1,0 +1,39 @@
+defmodule Tuist.ClickHouseDictionarySource do
+  @moduledoc """
+  Builds the `SOURCE` clause for dictionaries that read a table living in the
+  same ClickHouse instance the repo is connected to.
+
+  A `CLICKHOUSE` dictionary source is resolved by the ClickHouse server, not by
+  the client that issued the `CREATE DICTIONARY`, so the server authenticates it
+  separately. With no user named in the source it falls back to `default` with
+  an empty password, which only succeeds on a server that has no password set.
+  Installs following the published self-hosted Compose file do set one, so every
+  backfill dictionary died there with `AUTHENTICATION_FAILED`. Name the
+  credentials the repo already connects with instead.
+
+  Host and port stay out of the clause on purpose. Omitted, they default to the
+  server's own address and native port, which is what a same-instance lookup
+  wants; the repo's port is the HTTP one the source would not dial anyway.
+  """
+
+  def local_table(repo, table) do
+    config = repo.config()
+
+    "CLICKHOUSE(TABLE #{literal(table)}#{credentials(config[:username], config[:password])})"
+  end
+
+  defp credentials(nil, _password), do: ""
+
+  defp credentials(username, password) do
+    " USER #{literal(username)} PASSWORD #{literal(password || "")}"
+  end
+
+  defp literal(value) do
+    escaped =
+      value
+      |> String.replace("\\", "\\\\")
+      |> String.replace("'", "\\'")
+
+    "'" <> escaped <> "'"
+  end
+end
--- a/server/priv/ingest_repo/migrations/20251118211224_backfill_test_runs_from_command_events.exs
+++ b/server/priv/ingest_repo/migrations/20251118211224_backfill_test_runs_from_command_events.exs
@@ -1,4 +1,5 @@
 defmodule Tuist.IngestRepo.Migrations.BackfillTestRunsFromCommandEvents do
+  alias Tuist.ClickHouseDictionarySource
   alias Tuist.IngestRepo
   alias Tuist.Repo
   use Ecto.Migration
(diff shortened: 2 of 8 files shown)


pepicrft (approved):
Approved.\n\nOne recommended security follow-up: the generated dictionary definition contains the ClickHouse password. Application-side `log: false` does not stop ClickHouse from writing the executed statement to `system.query_log`. The narrowest fix is to add `settings: [log_queries: 0]` to each credential-bearing `CREATE DICTIONARY` request, while keeping `log: false`. Please add an integration test with a passworded ClickHouse user that verifies the password is absent from `system.query_log`.\n\nA dedicated dictionary-source user limited to the lookup tables would further reduce impact if a credential is exposed.

pepicrft:
Some research context on the recommended direction:\n\nClickHouse supports predefined connections, now called named collections, for dictionary sources. The standard production shape is to provision the connection, including its credential, in ClickHouse server configuration and have the dictionary reference the collection. That keeps the password out of application-generated statement text and therefore out of the query log.\n\nFor this change, I would take a layered approach:\n\n1. Add `settings: [log_queries: 0]` to the credential-bearing `CREATE DICTIONARY` requests, while retaining `log: false`. This is the small, immediate mitigation.\n2. Follow up by introducing a server-provisioned named collection for dictionary sources. Do not create that collection through a statement containing the password, because it would recreate the same query-log exposure.\n3. Use a dedicated dictionary user limited to `SELECT` on the lookup tables, reducing the impact of a credential disclosure.\n\nClickHouse documents that predefined connections work with dictionaries: https://presentations.clickhouse.com/2021-release-21.11/index.html\n\nNamed collections can be configured and granted independently: https://presentations.clickhouse.com/2024-release-24.6/index.html\n\nBefore adopting the follow-up, add an integration test against the supported ClickHouse version that verifies neither the password nor rendered connection data reaches `system.query_log`.

fortmarek:
Thanks, both taken on board. I checked the query log claim against ClickHouse before acting on it, and it needs one correction, which changes what the follow-up test should be.

## ClickHouse already masks this

On 26.1.2.11, a well-formed credential-bearing statement reaches `system.query_log` as:

```
CREATE DICTIONARY probe_dict (...) SOURCE(CLICKHOUSE(TABLE 'test_cases' USER 'masking_probe_user' PASSWORD '[HIDDEN]'))
```

`SHOW CREATE DICTIONARY` is masked the same way, and a search of the whole of `system.query_log` for the raw secret returns 0 rows. So the statement is written, but the password is not. This is built-in masking, independent of the application-side `log: false`.

## There is still a real gap, and it is why I took the change

The masking works from the parsed statement, so a statement that fails to *parse* is logged verbatim as `ExceptionBeforeStart`. I reproduced it: a malformed source clause put the literal password into `query_log`, 1 row. Our statements are generated and well-formed, and the only route to an unparseable one is a credential-escaping bug, which is what the quote/backslash test covers. So this is defense in depth against a bug class already under test rather than an open hole, but it is cheap, so `log_queries: 0` is in as of 8fc68d7.

Both options now live behind `Tuist.ClickHouseDictionarySource.query_opts/0` so the reason sits next to the code that renders the credential. Verified end to end through the driver against a passworded user: 0 query log rows for the statement, 0 rows containing the password.

## On the requested integration test

Written as specified it would pass on unfixed code. Asserting "the password is absent from `system.query_log`" is green today without `log_queries: 0`, because ClickHouse masks it, so it would be testing ClickHouse's masking rather than our change. Instead there is a source-level guard, matching the one already in the PR, asserting every credential-bearing statement is issued with `query_opts/0`. That one is red against the previous commit, naming all five migrations, and green after. It also constrains future migrations, which the runtime test would not.

One caveat worth recording: I verified masking on 26.1.2.11. The install that reported this runs 25.5.11.15, which I have not verified directly. That asymmetry is another reason to keep the setting rather than rely on masking alone.

## On named collections

I do not think they fit the case this PR exists for. A named collection has to be provisioned in ClickHouse server configuration before migrations run, and per your own caveat it cannot be self-provisioned without recreating the exposure. That makes it a mandatory operator step on every self-hosted upgrade, which is the exact failure class this PR is fixing: an install that has not provisioned it fails the migration, just with a different error code. The dedicated `SELECT`-only source user has the same shape, and would need a fallback path for instances that lack it.

Both look right for the managed instance, where we control provisioning. Suggest splitting them into a follow-up scoped to that, rather than putting the self-hosted restore path behind them. Happy to open it if you agree with the split.

Also flagging that the two links are release presentation decks rather than documentation, in case you meant to link the named collections docs.

PR #12492, opened 2026-08-20
feat(infra): publish the macOS runner guest log off the VM

fortmarek:

A macOS runner that halts without ever taking its job currently leaves an exit code and nothing that explains it. This carries the guest's own log off the VM so those deaths can be diagnosed.

### The gap

`dispatch-poll.sh`'s EXIT trap reports `0` both for a finished job and for a runner that gave up, so the exit code cannot separate the two. The script's log lives at `/var/log/tuist-runner/poll.log` **inside** the ephemeral guest and dies with it. Tart 2.32.1 has no macOS-guest console capture, and the host's `/var/log/tart-vms/<vm>.log` holds `tart run`'s output rather than the guest's. So nothing carried that trail off the VM.

This is not hypothetical. Investigating a customer's 6m04s runner pickup delay, every one of the 43 pods that produced an `orphaned running row — GH still queued` line in a 24h window terminated `exitCode=0 reason=Completed`. On Linux the equivalent question was answered in a single Loki query, because Linux runners are ordinary containers whose stdout is collected; the macOS half was unanswerable.

### What changed

- **Guest** (`dispatch-poll.sh`): `publish_runner_log` copies `${LOG}` into `${STATUS_SHARE}/runner.log`, called from the EXIT trap after the final message and before the halt. Guarded on the share being mounted, exactly like `report_runner_exit`.
- **Host** (`tart-kubelet` `internal/podagent`): `readRunnerLog` returns a bounded tail (200 lines / 64 KiB, byte-bounded first, partial leading line dropped). Both terminal paths in `reconciler.go` — the `Run != nil` branch and the recovered `Run == nil` branch — log it as `"runner log"` **before** `deleteByKey` removes the share.

`tart-kubelet`'s stdout is `/var/log/tart-kubelet.log`, which the host log shipper already tails, so the trail reaches Loki with **no change to the shipper**.

### Decisions a reviewer should know

**Routed through tart-kubelet, not by having the shipper tail the shares.** The shares are the obvious target and the worse one: they are per-VM and ephemeral, so tailing them needs glob discovery, per-file position churn, and a separate answer for pools where cache volumes are off and no share exists. Going through tart-kubelet costs one bounded read on a path that already had to open the share to recover the exit code, and the shipper stays a single-file tailer. `infra/macos-log-shipper/AGENTS.md` previously described this as impossible without new plumbing; that section is rewritten.

**A copy from the trap, not a `tee`.** The tee was written first and backed out. It is still running when the trap fires, so its buffered tail can land *after* the copy and duplicate it — and macOS `/bin/bash` is 3.2, which does not set `$!` for a process substitution, so there is no portable way to reap it first. Copying after the last line is written is exact. The cost is that a guest killed before its trap runs publishes nothing, which already reaches the cluster distinguishably as `TartRunExited`.

**Emitted unconditionally, not gated on the exit code.** Gating is exactly what makes the runners-controller's existing `abnormalEnd` death-log capture blind here: every runner in question reports `0`, so any exit-code gate drops the whole population worth capturing.

### This does nothing until both halves roll

`dispatch-poll.sh` is baked into the runner image, so it needs a `release-runner-image` run plus a `runnersFleet.runnerImage` pin. `tart-kubelet` rides the CAPI operator image and reaches hosts on an SSH drift roll (its bytes are in `bootstrap.HostConfigHash`). Until both land, macOS strandings still explain nothing.

This is instrumentation only. It does not address the two actual defects found alongside it — the 300s cron recovery latency, and job/runner bookkeeping that assumes GitHub binds a job to the runner we minted it for when it binds by label set.

### How to test locally

```bash
cd infra/tart-kubelet && go build ./... && go vet ./... && go test ./internal/podagent/
bash -n infra/runner-image/dispatch-poll.sh
```

The five tests in `runner_log_test.go` cover absent share, whole-file, tail-not-head, partial-first-line drop, and byte bounding. They were checked against a mutant: removing the partial-line drop fails `TestReadRunnerLogDropsPartialFirstLine`.

Once both halves are rolled, confirm in Loki with `{job="tuist-macos-tart-kubelet"} |= "runner log"`.

### DIFF
--- a/infra/tart-kubelet/internal/podagent/volume_reconcile.go
+++ b/infra/tart-kubelet/internal/podagent/volume_reconcile.go
@@ -82,6 +82,30 @@ const dirtyMarkerFile = "cache-dirty"
 // see runnerTermination.
 const runnerExitFile = "runner-rc"
 
+// runnerLogFile is dispatch-poll.sh's own output, mirrored into the
+// writable status share by the guest so it outlives the VM. The copy
+// inside the guest (/var/log/tuist-runner/poll.log) dies with the VM at
+// teardown, and Tart cannot capture a macOS guest's console, so without
+// this the host has an exit code and nothing that explains it.
+//
+// That gap is not academic: the trap reports 0 both for a finished job
+// and for a runner that halted without ever taking one, so the exit code
+// alone cannot tell those apart. This file is what does.
+//
+// Absent for the same two reasons as runnerExitFile — a guest killed
+// before its trap ran, and hosts with no status share attached at all.
+const runnerLogFile = "runner.log"
+
+// runnerLogTailLines / runnerLogTailBytes bound what publishRunnerLog
+// re-emits. dispatch-poll.sh is quiet by design — the job's own output
+// goes to GitHub server-side, so this log is warm-standby ticks plus the
+// cache teardown trail — but the file is guest-writable, so it is bounded
+// rather than trusted.
+const (
+	runnerLogTailLines = 200
+	runnerLogTailBytes = 64 << 10
+)
+
 // readRunnerExit reads the guest-reported exit code from the status share.
 // The bool is false when the guest reported nothing usable.
 //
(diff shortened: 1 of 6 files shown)


infra/tart-kubelet/internal/podagent/volume_reconcile.go
```diff
@@ -110,6 +134,49 @@ func readRunnerExit(statusDir string) (int32, bool) {
...
+//
+// Tail rather than head: the interesting part of a runner that gave up
+// is what it said last. Bounded twice because the file is guest-written
+// — by bytes first so a single pathological line cannot blow up the log
+// record, then by lines.
+func readRunnerLog(statusDir string) string {
+	if statusDir == "" {
+		return ""
+	}
+	f, err := os.Open(filepath.Join(statusDir, runnerLogFile))
```
pepicrft: Blocking: this status directory is writable by the guest, but `readRunnerLog` follows symbolic links when it opens `runner.log`. An untrusted job can leave `runner.log` as a symbolic link to a host-readable file. The guest-side copy then fails to overwrite that link and ignores the failure; tart-kubelet subsequently reads the link target and emits up to 64 kilobytes of it to Loki.

Please open the file without following symbolic links and require a regular file before reading it. A regression test should cover a guest-created symbolic link.

pepicrft (approved):

infra/tart-kubelet/internal/podagent/volume_reconcile.go
fortmarek: Fixed in 19329b21b7 — `runner.log` is now opened `O_RDONLY|O_NOFOLLOW|O_NONBLOCK` and rejected unless `Mode().IsRegular()`; regression tests cover a guest-planted symlink (asserts the host file content does not come back), a FIFO (asserts the open does not park the reconcile — it did, for the full 5s timeout, before the fix), and a directory. Worth flagging separately: the seven sibling readers on this same guest-writable directory (`runner-rc`, `cache-dirty`, `volume-head.json`, `promote-result`, `upload-millis`, `fill-percent`) all use `os.ReadFile`, which has the same symlink exposure and, more importantly, blocks forever on a FIFO — pre-existing and untouched by this PR, so I have not widened it here.

PR #12497, opened 2026-08-20
fix(server): recover a stranded runner job when its Pod is gone, not when its claim is

fortmarek:
A job dispatched to a runner Pod that takes the JIT config and never runs it shows the customer "Waiting for a runner to pick up this job" until something moves its lifecycle row off `running`. The worst case observed in production waited 6m04s, of which 5m54s was recovery latency: the Pod was known dead 67 seconds after dispatch, and nothing acted on it for another 285 seconds.

Dispatch itself is not the problem (1.7s in that case), and neither is capacity. This is entirely about how long recovery takes once a runner strands a job.

### Why the existing fast path did not fire

#12395 added the right mechanism: when the runners-controller reports a Pod stop, the server releases the stranded claim and schedules a targeted `OrphanedRunnersWorker` run. It recovers a stranded job in about 88 seconds when it fires. Over a 24 hour sample it fired for roughly 9 of 42 strandings, and the other ~80% waited out the full 300 second sweep.

`release_stranded_claim/1` scheduled targeted recovery only for the ids that `Claims.release_by_pod_name/1` actually deleted. That is the wrong key. The claim is the account's **capacity** reservation, and it is released by whichever event frees the Pod's slot first. For a runner that executed a sibling's job (the common shape, since GitHub binds a JIT runner by label set and never to a specific job) that job's `completed` webhook has already released this Pod's claim via `Claims.release_by_executor/2`. So the release returns `[]`, `release_stranded_claim/1` returns `:ok` silently, and the fast path never runs, while the job the Pod was minted for is still sitting at `running` where `pick_queued/2` skips it. `release_by_pod_name`'s own docstring says the return is "empty in the common case".

Recovery now keys on the lifecycle rows bound to the Pod (`Jobs.list_running_for_pod/1`), which is what dispatch actually reads. The Pod stopping is proof the job is not executing on it, whether or not a claim row survived to be released.

### Why the report going missing had to stop mattering

Repairing the POST alone would not be enough, because it can go missing in ways no server change can fix: a dropped request, a controller restart mid-reap, or a Pod removed by node loss, eviction or drain rather than by the reap.

So `OrphanedRunnersWorker`'s sweep now asks the same question level-triggered, off one cluster read per tick: is the Pod bound to this `running` row still there? Age decides which rows are candidates; absence decides what evidence each candidate carries, and it outranks age wherever the read proves it. Those two stay separate deliberately, so a row that crosses the floor before the first successful read does not get filed under age for the rest of its life.

The 5 minute floor was never a safety property. It is a stand-in for evidence the sweep cannot otherwise get without paying a GitHub call per running row per minute, and a cluster read supplies that evidence directly. It also carried up to a full extra cron period of phase misalignment on top of the 300 seconds.

Every guard biases toward doing nothing, matching `PodReconciliationWorker`'s posture:

* No running rows at all means no cluster read, so steady state costs two queries and no apiserver traffic.
* A failed or empty read narrows straight back to the age gate rather than treating every Pod as absent, so a read we cannot trust degrades to the old behaviour rather than to a guess.
* A row without a Pod name is never treated as absent.
* Absence widens what is asked about, never what is acted on: every candidate still goes through the GitHub `queued` cross-check before anything is re-queued.

`@stale_after_seconds` deliberately stays at 300. With the pod-gone arm it only governs "Pod still present but the runner never registered", where cutting it trades directly against cold-boot time and burns a second runner on a genuinely slow VM.

### Pod-gone evidence is threaded through recovery, not just used to pick candidates

The queued branch of recovery skips any job whose claim records an execution (`Claims.executing?/1`). That guard is right for the age-based sweep: a runner handed a sibling's job is genuinely busy, and releasing on the claimed job's GitHub status alone would delete a live reservation mid-job.

A Pod that is gone is not busy, so the guard's premise is false exactly where the new arm has proof. Left in force it would skip the arm's main population: a Pod that claimed J1, took sibling J2, then disappeared with its stop report lost keeps a claim carrying `executed_workflow_job_id`, because J2 never completed to release it. J1 would then wait for `PodReconciliationWorker`'s 10 minute grace plus 5 minute confirmation, defeating the fallback this PR is for.

So recovery now carries what the caller knows about the Pod: `:pod_stopped` when it has been observed gone, `:aged_out` when the row's age is all there is. Only the queued branch reads it, and only to settle this guard. The release stays handle-guarded on `claimed_at`, so a row a replacement Pod has since re-claimed is still untouchable.

Targeted mode carries `:pod_stopped` too. It previously satisfied the guard only as a side effect of the controller's report having deleted the claim first, which made a correctness path depend on one release winning a race it is not guaranteed to win.

### Why this could not be diagnosed from the logs

The premise going in was that the POST was not being made. That turned out to be unsupported, and the reason is worth fixing on its own:

* The server emits no request log line for the endpoint. `TuistWeb.Endpoint` has `plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint], log: false`.
* The controller logged its successful report at `logger.V(1)`, suppressed at production zap level.

A POST that never happened and a POST that arrived and found no claim produced identical logs, so a Loki count of `runners/pods/stopped` is not a count of successful POSTs. Both success paths now log at info, tagged `source=phase-transition` or `source=reap` so it is clear which reconciler won the race, and the server logs and counts every recovery it schedules on a Pod stop.

One case in the sample is still unexplained: the account had no sibling job in the window, so its claim should have survived and the old code should have fired for it. The new logs are what will answer that.

### Impact

Worst case for a stranded job drops from about 6 minutes to about 2 minutes even if the Pod stop report is lost entirely, and to about 80 seconds when it lands.

### Deploy note

The two controller logging changes ride the runners-controller image, which is pinned to the latest `runners-controller@` git tag rather than to `main`, so they need a new tag to reach production. The server side ships on the normal deploy and is independently useful.

### How to test locally

Server:

```
cd server && mix test test/tuist/runners/ test/tuist_web/controllers/runner_pods_controller_test.exs test/tuist_web/controllers/runners_controller_test.exs
```

Controller:

```
cd infra/runners-controller && go build ./... && go vet ./... && go test ./controllers/
```

643 Elixir tests pass with `--warnings-as-errors`, including new coverage for:

* recovery scheduled for a job left running by a stopped Pod whose claim was already released
* the Pod-name binding in both directions, so a stopped Pod cannot drag a replacement's row through recovery
* the pod-gone arm against a failed read, an empty read, a Pod still present, and a blank Pod name
* the busy-guard bypass on the pod-gone arm, and in targeted mode when the claim outlived the release
* the bypass reaching a row that has already aged past the floor, which is what stops absence from being lost when the first successful sweep is delayed
* the age-based sweep still honouring the busy guard when the Pod is present, so the bypass stays scoped to proven absence
* `list_running_for_pod/1` and `list_running_since/1` being exact complements of `list_orphaned_running/1`

Go build, vet, tests and gofmt are clean for the controller.

### Verifying in production after rollout

Loki, datasource `grafanacloud-logs`:

* fast path firing: `{namespace="tuist"} |= "recovering jobs left running by a stopped pod"`
* level-triggered arm firing: `{namespace="tuist"} |= "rescued orphaned running rows"`, which now carries `aged_out` and `pod_gone` counts separately
* strandings overall: `{namespace="tuist"} |= "orphaned running row" |= "GH still queued"`
* the POST itself, once the controller image is tagged: `{job="tuist/manager"} |= "reported pod stopped"`

Success is the fast path plus `pod_gone` accounting for nearly all strandings, `aged_out` falling to roughly zero, and enqueue to start for a stranded job dropping from ~350s to under ~100s.


server/lib/tuist/runners/workers/orphaned_runners_worker.ex
```diff
@@ -231,7 +355,14 @@ defmodule Tuist.Runners.Workers.OrphanedRunnersWorker do
...
     # completion (or the Pod stopping) free the slot.
-    if Claims.executing?(workflow_job_id) do
+    #
+    # Unless the Pod has already stopped, which is what `:pod_stopped`
+    # carries. A gone Pod is not busy, so the guard's premise is false and
+    # holding it would strand the job until the executed job's completion
+    # or `PodReconciliationWorker`'s much longer confirmation. The release
+    # below stays handle-guarded on `claimed_at`, so this cannot reach a
+    # replacement Pod's claim.
+    if evidence == :aged_out and Claims.executing?(workflow_job_id) do
```
pepicrft, line 388: Could we also evaluate Pod absence for rows that have crossed the five-minute age threshold? These rows are sent through `:aged_out`, so this guard still returns `false` when the claim has an `executed_workflow_job_id`, even if the Pod is now absent. I reproduced the path by delaying the first successful sweep past five minutes: a queued GitHub job with an executing sibling claim stays stranded. A test covering an aged-out, absent Pod with an executing claim would prevent this regression.

pepicrft (approved):
Runtime validation completed. Approved with one inline follow-up.

server/lib/tuist/runners/workers/orphaned_runners_worker.ex
fortmarek: Fixed in db50bd271e. You were right that evidence was tied to which list a row came from: age now only decides candidacy, and Pod absence is evaluated for the whole candidate set (aged rows included) where it outranks age. Added your exact case as a test (aged row, absent Pod, executing claim), plus the inverse so the bypass stays scoped to proven absence.

PR #12502, opened 2026-08-20
feat(server): restrict who can accept a bundle size increase

fortmarek:
Accepting a bundle size threshold violation from a GitHub check run was unauthorized. [`handle_check_run/2`](https://github.com/tuist/tuist/blob/main/server/lib/tuist_web/controllers/webhooks/github_controller.ex) looked up the installation and flipped the check run to `success` without ever resolving which project the check run belonged to, and without looking at the `sender`. The only gate was GitHub's own rule that requested actions need repository write access, which is far wider than teams want for a decision that unblocks a merge.

This came out of a customer question: whether the people who can accept an over-threshold report can be restricted.

### What changed

Projects carry a `bundle_size_approval_policy` with two values, edited in Settings > Bundles alongside the thresholds it governs:

- `everyone` (default, the previous behavior): anyone with write access to the repository
- `selected`: only the GitHub usernames on a per-project allowlist

The Approvals section of a project's bundle settings

### Why the allowlist is keyed on GitHub handles

The obvious design is a picker over org members storing `user_ids`. I checked what that would resolve to for a large account currently evaluating this feature: every member signs in through SSO, and fewer than 10% have ever also linked GitHub. A user-keyed list would be inert for most people on it, and the label in the UI would promise an ability the webhook could not deliver.

Handle matching needs no identity resolution at all and works regardless of how someone signs in to Tuist. It also means the policy asks nothing of your members before it works.

An earlier revision of this branch also had `admins` and `report_only` policies. Both are gone. `admins` depended on the same GitHub-to-Tuist link that is usually missing, and needed a warning in the settings page telling admins the policy would deny everybody, which is a sign the option should not have been offered. `report_only` existed for projects wanting to gate on a check of their own, but keeping the default and ignoring the Tuist check achieves that without a policy for it; the docs say so directly. Removing both let authorization stop resolving the sender at all, deleting `Accounts.any_admin_with_github_identity?`, the unlinked-admin warning, and every `github_account_not_linked` path.

### Resolving the project

The webhook payload does not name a project, and a repository can back several Tuist projects, so the repository handle alone is ambiguous. The check run now carries `external_id` with the bundle id, which is exact and is the field GitHub provides for this. Check runs created before this change carry no `external_id` and keep accepting as they did, rather than stranding pull requests that were already open when this deploys. The pre-existing webhook tests pass untouched, which is what confirms that path.

### Denials and attribution

A denied click leaves the check run as `action_required` and re-sends the actions array, so the pull request stays blocked and whoever is allowed still has a button to press.

The `everyone` policy short-circuits before any identity lookup, so the default can never deny a member who has not linked GitHub.

Acceptances are recorded in `bundle_size_approvals` with the GitHub handle and, when resolvable, the Tuist user. That is a Postgres table rather than columns on the bundle row because `bundles` is a ClickHouse MergeTree, where updating a single row is an asynchronous mutation.

### Also in this branch

A fix to the dev code reloader, unrelated to the feature. `config/dev.exs` lists `reloadable_apps: [:tuist, :noora]`, and gathering manifests for the `noora` path dependency went through `Mix.Dep.in_dependency/3`, which pushes it onto Mix's project stack. Loading a project whose module is already defined raises there, and once the stack is dirty every later request raises too. The visible symptom was not an error page: `Plug.Static` runs before the reloader and stops resolving `priv/static` in that state, so the dashboard came back styled on the first request after a boot and unstyled on every request after. The manifest path is derivable from the dependency's build directory, so nothing needs to load the project. Happy to split this onto its own branch if you would rather keep this PR single-purpose.

### Follow-up worth considering

A project gating on its own job has no clean way to read the verdict today. `GET /bundles/:id` returns sizes but no threshold evaluation, so such a job has to scrape our check run conclusion off the GitHub API, which races the Oban worker that posts it, or refetch the baseline and recompute the deviation itself. An endpoint returning the verdict would make those jobs a few lines instead of a reimplementation. Out of scope here.

### Reviewer notes

- `server/data-export.md` covers both new tables.
- The three migrations are additive; the `projects` column is added with a default, which is non-blocking on PostgreSQL 11+.
- Only the `.pot` template changed from `mix gettext.extract`; no `.po` files were touched.
- Two commits remove files committed by accident (`.tmp-visual/`, `noora/minio/`). The final diff is clean; say the word if you want them squashed out.

### How to test locally

1. `mise run dev` and open a project's Settings > Bundles tab. The Approvals card sits under Size thresholds.
2. Switch to Selected GitHub users, add an approver through the modal, and try an invalid username to see the validation message.
3. Automated coverage:

```
cd server && mix test test/tuist/bundles_test.exs test/tuist/bundles/workers/bundle_threshold_worker_test.exs test/tuist_web/controllers/webhooks/github_controller_test.exs test/tuist_web/live/project_bundle_settings_live_test.exs
```

`mix credo`, `mix sobelow`, `mix excellent_migrations.check_safety` and `mise run format` are clean.


esnunes:
reviewing

server/lib/tuist_web/controllers/webhooks/github_controller.ex
```diff
@@ -383,6 +384,104 @@ defmodule TuistWeb.Webhooks.GitHubController do
...
+          title: "Bundle size increase accepted",
+          summary: accepted_summary(sender)
+        }
+      })
+    )
+  end
+
+  # Leaves the check run failing and re-sends the action so whoever is
+  # allowed to accept still has a button to press.
+  defp reject_bundle_size(check_run_params, sender, project, reason) do
```
esnunes: One thing I wanted to flag about the denial path: `reject_bundle_size/4` sends a fresh `output` map, and GitHub replaces the check run's output wholesale rather than merging it. That wipes what `build_check_run_output/2` wrote — the baseline/current/change table, the threshold name and branch, and the "View bundle details" link — and leaves just the refusal sentence.

The accept path replacing the output feels fine to me, since the check run is finished at that point. A denial is different: the check run still needs someone to act on it, and the person who *is* allowed to press Accept now arrives at a button with no report next to it explaining what they'd be accepting.

The payload already carries the existing text at `check_run["output"]["summary"]`, so appending the refusal to it (or re-rendering the size table) would keep both. Might be worth it?

server/lib/tuist_web/controllers/webhooks/github_controller.ex
```diff
@@ -383,6 +384,104 @@ defmodule TuistWeb.Webhooks.GitHubController do
...
+          {:error, reason} ->
+            reject_bundle_size(check_run_params, sender, project, reason)
+        end
+
+      {:error, :not_found} ->
+        reject_bundle_size(check_run_params, sender, nil, :bundle_not_found)
+    end
+  end
+
+  defp project_for_bundle(bundle_id) do
```
esnunes, line 452: `project_for_bundle/1` only needs `bundle.project_id`, but it gets there via `Bundles.get_bundle/1`, which always calls `bundle_artifacts/1` — an unbounded SELECT over the ClickHouse `artifacts` table for that bundle, followed by an in-memory group_by and recursive tree build. There's no way to skip it; `drop_artifacts_preload/1` only strips the Ecto preload, and the tree build runs after it either way.

For a real app that's a lot of rows, paid on every Accept press, inside a webhook GitHub times out at ~10s. A narrow query would cover it:

```elixir
from(b in Bundle, where: b.id == type(^bundle_id, Ecto.UUID), select: b.project_id)
|> ClickHouseRepo.one()
```

Nice side effect: dropping the `type(^id, Ecto.UUID)` cast means a non-UUID `external_id` comes back as nil rather than raising `Ecto.Query.CastError`, which the `else` clause here can't catch today.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
```diff
@@ -383,6 +384,104 @@ defmodule TuistWeb.Webhooks.GitHubController do
...
+  # open when this shipped.
+  #
+  # A check run that does carry one is a different case. Failing to resolve it
+  # means the bundle read came back empty, not that the project has no policy,
+  # so accepting would hand an unverified click the same outcome as an
+  # authorized one. `bundles` lives in ClickHouse, where a read is not
+  # guaranteed to see a recent write, and a bundle can also age out of
+  # retention, so this is reachable without anyone doing anything wrong. It
+  # leaves the check failing and invites another press instead.
+  defp resolve_bundle_size_approval(check_run, check_run_params, sender) do
```
esnunes: Not disagreeing with the decision here — the comment explains it well, and stranding already-open PRs would clearly be worse. What I wanted to flag is the shape of the window rather than the trade-off.

`external_id` is stamped at check-run creation, so this isn't bounded by the rollout: a PR opened before the deploy keeps an unauthorized Accept button for as long as it stays open, which for a stale branch can be months. A project that switches to "Selected GitHub users" the day after deploy has a set of PRs where the setting quietly doesn't apply, with nothing indicating which.

One option that needs no new state: `Projects.projects_by_vcs_repository_full_handle/2` already exists. The repository handle can't tell you *which* project a check run belongs to, but it can answer "does any project backed by this repo restrict approvals?" — enough to fail closed on the ambiguous case only where someone actually asked for restrictions, and leave the default untouched. Failing that, a follow-up to drop the fallback once open PRs have turned over would at least keep it from becoming permanent.

server/lib/tuist/bundles.ex
```diff
@@ -751,6 +755,130 @@ defmodule Tuist.Bundles do
...
+          github_handle: username,
+          github_id: github_id
+        })
+        |> Repo.insert()
+
+      {:error, :no_vcs_connection} = error ->
+        error
+
+      _ ->
+        {:error, :github_user_not_found}
```
esnunes, line 809: The `_ ->` catch-all here folds every non-`:no_vcs_connection` outcome into `{:error, :github_user_not_found}`, which the LiveView shows as "No GitHub user with that username." So a GitHub 5xx, a secondary rate limit, an expired installation token or a Req transport error all tell the admin that the account doesn't exist.

That's a claim about the world when what really happened is that we couldn't ask, and it's the kind of message that makes someone give up on a username that was correct. Could we split the 404 from the rest and surface something like "couldn't reach GitHub, try again" for the others? `Client.get_user_by_username/1` already passes the raw `github_request/2` response through for anything that isn't a well-formed user, so the information is there.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
```diff
@@ -383,6 +384,104 @@ defmodule TuistWeb.Webhooks.GitHubController do
...
+          }
+        ]
+      })
+    )
+  end
+
+  defp accepted_summary(%{handle: nil}), do: "The bundle size increase was manually accepted."
+  defp accepted_summary(%{handle: handle}), do: "The bundle size increase was accepted by @#{handle}."
+
+  defp rejected_summary(_sender, _project, :bundle_not_found) do
```
esnunes: Two small things on the same pair of functions, neither blocking.

`rejected_summary/3` only matches `:bundle_not_found` and `:not_an_approver`, which is complete against what `authorize_bundle_size_approval/2` returns today. If a third denial reason ever shows up, though, it's a FunctionClauseError raised inside the webhook handler: a 500, and the check run left untouched with no button re-sent. Since the whole point of the reject path is that whoever *is* allowed still has something to press, a catch-all clause seems like cheap insurance.

Also, `project` is threaded through `reject_bundle_size/4` into `rejected_summary/3` and both clauses ignore it — looks like a leftover from when the summary named the approvers, before 86c45d3b dropped that. Removing the parameter would make both signatures honest.

server/lib/tuist/bundles.ex
```diff
@@ -751,6 +755,130 @@ defmodule Tuist.Bundles do
...
+  numeric id, which is what authorization compares. A username is not a
+  durable identifier: its owner can change it, and the old one becomes
+  available for someone else to claim, so an allowlist keyed on it can be
+  inherited. Resolving also rejects a username that does not exist, which a
+  format check alone cannot do.
+  """
+  def add_bundle_size_approver(%Project{} = project, handle) do
+    handle = normalize_github_handle(handle)
+
+    case VCS.get_user_by_username(%{username: handle, project: project}) do
```
esnunes, line 795: Ordering nit: `normalize_github_handle/1` only trims, strips a leading `@` and downcases, and the result goes straight to `VCS.get_user_by_username/1`, which interpolates it into `"#{api_url}/users/#{username}"`. The `validate_format` that would reject a non-username lives in `BundleSizeApprover.changeset/2`, which doesn't run until the insert — after the HTTP call.

The blast radius looks small to me: the caller already holds `:project_update`, host and scheme are fixed, the token is the project's own installation token, and anything that doesn't come back with both `"id"` and `"login"` falls to `{:error, :not_found}`. But running the format check before the call costs nothing and skips a pointless round-trip on obviously bad input.

server/lib/tuist_web/live/project_bundle_settings_live.html.heex
```diff
@@ -395,4 +395,155 @@
...
+              "dashboard_projects",
+              "This person can accept a bundle size increase that exceeded a threshold, from the check run on a pull request."
+            )
+          }
+          header_size="large"
+          on_dismiss="close_add_approver_modal"
+        >
+          <:trigger :let={attrs}>
+            <.button
+              label={dgettext("dashboard_projects", "Add approver")}
```
esnunes, line 462: Nit, definitely not a blocker: the "Add approver" trigger isn't gated on `@has_vcs_connection`, while the "Add threshold" trigger two card sections above is. Since `@has_vcs_connection` is already assigned on this socket, adding the same `disabled` attribute would make the two cards read the same way.

Right now on a project with no GitHub connection one button is visibly unavailable and the other opens a modal, takes a username, and only then explains it can't be used. The error path is there and tested — it's just a different affordance than its neighbour.

(Same argument arguably applies to the policy picker, which lets you choose "Selected GitHub users" with no connection and no way to populate the list — though the empty state does explain itself there.)

esnunes (approved):
Really nice piece of work — the framing in the description does a lot of the reviewing for me, and the design decisions land where I'd have wanted them to. Three in particular:

Keying the allowlist on GitHub's numeric account id rather than the username, with the migration comment spelling out why (a released username can be re-claimed, so a handle-keyed list can be inherited). Using `external_id` to identify the check run, which is the field GitHub provides for exactly that and sidesteps the repo-handle ambiguity cleanly. And the clause ordering in `authorize_bundle_size_approval/2`: the catch-all denies under `:selected`, so a sender with no usable id fails closed. That's the easy one to get backwards.

The "stamped but unresolvable" case deserves its own mention. Splitting it from "no external_id at all" and refusing rather than accepting is the right call, and encoding the reasoning in the test comment — a bundle read coming back empty isn't evidence that nobody needs checking — is how it stays that way. The empirical argument for handles over a member picker is convincing too; I hadn't considered how thin the SSO-to-GitHub link coverage would be.

Two things I'd like your read on before this goes in. The denial path replaces the check run's output rather than adding to it, which wipes the size table and the details link — leaving the person who *is* allowed to accept looking at a button with no report next to it. And resolving the project runs the bundle's full artifact tree build inside the webhook just to read `project_id`, which is a lot of ClickHouse work on a request GitHub gives ~10s.

Beyond that: the legacy no-`external_id` fallback is deliberate and I agree with it, but it isn't bounded by the rollout, so it's worth deciding now whether it ever closes. The rest are small — an error path that reports every GitHub failure as "no such username", a missing catch-all clause in the webhook handler's denial summary, an ordering nit on when the handle format gets validated, and a UI affordance that differs from its neighbour. Details inline.

On the code reloader fix: I read it and it's correct — `opts[:build]` plus `.mix/` is where the compiler writes the manifest, so nothing needs to load the project. No objection to it riding along, though if you'd rather keep this single-purpose I'd not argue.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
fortmarek: Fixed — the reject path now carries the existing `output.summary` through and appends the refusal below a separator it owns, so the size table, threshold and details link survive and a repeated press replaces the previous refusal instead of stacking. The title is preserved too, so the check keeps reading as "threshold exceeded" rather than flipping to the refusal.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
fortmarek: Fixed — added `Bundles.get_bundle_project_id/1`, which selects just `project_id`, so the artifact tree is no longer built to reach one integer. Kept the UUID validation in Elixir via `Ecto.UUID.cast/1` rather than dropping the cast, which gets the same nil-instead-of-raise result without depending on how ClickHouse handles an untyped comparison.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
fortmarek: Fixed, taking the option you sketched — an unstamped check run now asks whether any project on the repository restricts approvals. Where none does the answer was "accept" anyway, so the default is untouched; where one does, the press is refused with a message saying to push a new commit. That bounds the window to projects that never asked for a gate.

server/lib/tuist/bundles.ex
fortmarek: Fixed — `Client.get_user_by_username/1` now maps only a 404 to `:not_found` and everything else to `:unavailable`, which surfaces as "Couldn't reach GitHub to check that username. Try again in a moment." rather than claiming the account does not exist.

server/lib/tuist_web/controllers/webhooks/github_controller.ex
fortmarek: Both fixed — added a catch-all `rejected_summary/2` clause so an unhandled reason still produces a check run update with the button re-sent, and dropped the unused `project` parameter from the reject path.

server/lib/tuist/bundles.ex
fortmarek: Fixed — the format check moved ahead of the HTTP call via `BundleSizeApprover.valid_handle?/1`, so obviously bad input returns `:invalid_github_handle` without a round-trip. The changeset keeps its own validation as the backstop.

server/lib/tuist_web/live/project_bundle_settings_live.html.heex
fortmarek: Fixed — the Add approver trigger now carries the same `disabled={!@has_vcs_connection}` as Add threshold. Left the policy picker alone as you suggested, since the empty state already explains itself there.

PR #12504, opened 2026-08-20
fix(docs): restore ProjectDescription publishing

pepicrft:
## What changed

- Restores a dedicated GitHub Actions workflow that builds the ProjectDescription reference documentation.
- Builds the documentation for relevant pull requests, deploys the generated site only from `main`, and provides a manual trigger for the initial backfill.
- Restores the Cloudflare Pages deployment for `tuist-projectdescription`.

## Why

The public ProjectDescription reference stopped updating after the documentation migration. New manifest types, including `TestPlan.generated`, have therefore not appeared on the published site.

## Root cause

The migration in #10083 removed `.github/workflows/docs.yml`, which contained the only ProjectDescription build-and-deploy job. The local `docs:project-description` task remained, but no workflow invoked it or deployed its output.

## Approach

The workflow uses the existing generation task, retains the previous Swift Package Manager cache and Cloudflare Pages target, and keeps deployment restricted to `main`. Its path filters cover the ProjectDescription source, package resolution, generator task, and workflow itself.

## Impact

After this pull request merges, dispatching the workflow will publish the current ProjectDescription reference. Later relevant changes on `main` will keep it current automatically.

## Validation

- `mise run docs:project-description`
  - Succeeded and generated the `TestPlan` reference page with the `.generated` option documentation.
- Parsed `.github/workflows/project-description-docs.yml` with Ruby and `yq`.
- Ran `git diff --check`.

### DIFF
--- a/.github/workflows/project-description-docs.yml
+++ b/.github/workflows/project-description-docs.yml
@@ -53,7 +53,9 @@ jobs:
           cache: "false"
           github_token: ${{ env.MISE_GITHUB_TOKEN }}
       - name: Build ProjectDescription documentation
-        run: mise run --skip-tools docs:project-description
+        run: >-
+          MISE_PROJECT_ROOT="$GITHUB_WORKSPACE" mise exec swift@6.2.4 --
+          bash mise/tasks/docs/project-description.sh
       - name: Save Swift Package Manager cache
         uses: actions/cache/save@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/project-description-docs.yml
```diff
@@ -0,0 +1,76 @@
...
+  cancel-in-progress: true
+
+env:
+  MISE_VERSION: "2026.5.15"
+  MISE_GITHUB_TOKEN: ${{ github.token }}
+
+jobs:
+  build-deploy:
+    name: Build and deploy
+    runs-on: ubuntu-latest
```
fortmarek: let's use our runners

fortmarek (approved):

.github/workflows/project-description-docs.yml
pepicrft: Updated it to use `tuist-linux`.

PR #12505, opened 2026-08-20
feat(server): add public accounts so an account's non-admin dashboards can be read signed-out

fortmarek:
`https://tuist.dev/tuist/runners` should be reachable without signing in. Projects already have a `visibility` flag; accounts did not, so every page under `/:account_handle` sat behind `require_authenticated_user`. This adds the account-level equivalent.

### What changed

`accounts.visibility` (`private` by default) mirrors `projects.visibility`. A new `public_account` check feeds a new `:account_dashboard_read` policy action that allows anonymous reads when the account is public.

The `/:account_handle` router scope is split in two:

| Public-capable scope | Authenticated scope |
| --- | --- |
| `/`, `/projects`, `/runners`, `/runners/workflows`, `/runners/workflows/:owner/:repo/:name`, `/runners/jobs`, `/runners/runs/:run/jobs/:job`, the job log download, `/usage` | `/members`, `/webhooks`, `/cache`, `/billing`, `/runners/profiles`, `/settings`, interactive VNC and shell |

The public scope swaps `require_authenticated_user` for a new `require_authenticated_user_for_private_accounts` plug, modelled on the existing `require_authenticated_user_for_private_projects`, and uses `mount_current_user` instead of `ensure_authenticated`. Each LiveView in that scope re-checks `:account_dashboard_read`, which is what actually gates the LiveSocket connect; the plug only covers the dead render.

The account layout hook, the account sidebar, and the project listings now tolerate a nil user. Signed-out visitors see only the account's public projects, and the create-project form is gated on `:project_create`.

The flag is operator-only, flipped from a new Visibility card on the ops account page. It is deliberately kept out of `Account.update_changeset/2`, because the account update API passes body params straight through and would otherwise let an account admin set it.

### Why a new policy action instead of widening `:runners_read`

The obvious change was to add `allow(:public_account)` to `object :runners, action :read`, since that was the gate the runners LiveViews used. That would have been a security bug. `:runners_read` also gates VNC and interactive shell attachment to a running runner VM, in both `interactive_state/5` and the interactive session controllers. Widening it would have handed anonymous visitors, and any signed-in non-member, a path toward those sessions.

So `:runners_read` stays members-only and the page gate moved to the new `:account_dashboard_read`, which grants exactly what `:runners_read` granted plus public reads. `Tuist.AuthorizationTest` asserts that invariant directly: on a public account, a non-member gets `:account_dashboard_read` but is refused `:runners_read`.

### Two latent bugs fixed here

Both are unreachable today and become reachable the moment a page renders for a nil user, so they are fixed rather than deferred:

- `Checks.billing_access/2` had no non-user clause, so the account layout's `:billing_read` call raised a `FunctionClauseError` for signed-out visitors. Caught by the new tests.
- The `create-project` handler's `with` had no `{:error, :forbidden}` branch, so an unauthorized submit raised a `WithClauseError` instead of an unauthorized error. This was already reachable by a non-admin member.

### Bot protection

The Cloudflare rule that shields the public project dashboards has been updated to cover the new paths. It previously matched only the four public projects under the account (`/tuist/tuist`, `/tuist/gradle-plugin`, `/tuist/ios_app_with_frameworks`, `/tuist/android`), none of which match the pages this pull request opens up: the account root, `/tuist/projects`, `/tuist/runners` and its sub-paths, and `/tuist/usage`. Each of those runs analytics queries on mount, so leaving them uncovered would have put crawler traffic straight onto the database.

The rule's exclusion list also needed a carve-out. It exempts anything ending in `/download` so that preview artifacts stay fetchable by devices installing an app, and the new runner job log route
(`/:account_handle/runners/runs/:workflow_run_id/jobs/:workflow_job_id/logs/download`) ends with `/download` too. It would therefore have been exempted by accident, which is the last endpoint that should be: it performs an account lookup, a ClickHouse job lookup, and an object storage presign per request, keyed on sequential integer identifiers. The exclusion now skips paths under `/tuist/runners/`.

Because visibility is per account, this rule needs extending whenever another account is made public.

### Reviewer notes

- **Public account pages stay `noindex`.** The `browser_app` pipeline applies `disable_robot_indexing` across the whole dashboard. Public projects have the same property today. Making these pages crawlable is a separate change to that pipeline.
- **Job logs are public on a public account**, including the log download endpoint, matching how public repositories expose CI logs elsewhere. Worth weighing before flipping any account whose CI output has not been reviewed.
- **SSO enforcement is unchanged.** The public scope keeps `require_sso_authentication`, identical to the public-project scope. An SSO-enforced organization that was flipped public would bounce anonymous visitors to the provider. Since the flag is operator-only, that combination is not expected to occur, and the same latent behaviour already exists for public projects.
- `server/data-export.md` is updated for the new column, per the repository guardrail.

### How to test locally

Start the server and, from an `iex -S mix phx.server` session, mark an account public:

```elixir
{:ok, account} = Tuist.Accounts.get_account_by_id(<id>)
Tuist.Accounts.update_account_visibility(account, :public)
```

Then, in a private browser window with no session, confirm that `/<handle>/runners`, `/<handle>/runners/jobs`, and `/<handle>/projects` render, that `/<handle>/settings` and `/<handle>/members` redirect to the login page, that the projects list omits private projects and shows no create-project button, and that a running job's page offers no terminal or screen tab. The same toggle is available from `/ops/accounts/<id>` under Visibility.

### Validation

- `mix test` over `test/tuist_web/live`, `test/tuist_web/controllers`, `test/tuist_web/authentication_test.exs`, `test/tuist/authorization_test.exs`, `test/tuist/projects_test.exs`, and `test/tuist/accounts_test.exs`: 2240 passing. The 4 remaining failures are pre-existing locale-compilation failures in the marketing and layout suites, reproduced on a stashed tree.
- `mix format --check-formatted`, `mix credo`, and `mix sobelow` clean (sobelow findings are pre-existing and unrelated).
- `mix excellent_migrations.check_safety` clean for the new migration.

New coverage: `TuistWeb.PublicAccountLiveTest` (signed-out reach, private-project hiding, admin-page redirects, member parity), plus additions to the authentication plug, authorization policy, projects context, job log controller, and ops account tests.


pepicrft (commented):
Review findings.

server/lib/tuist_web/router.ex
```diff
@@ -1052,6 +1052,44 @@ defmodule TuistWeb.Router do
...
+  # gates the LiveSocket connect — the pipeline below only covers the
+  # dead render.
+  scope "/:account_handle", TuistWeb do
+    pipe_through [
+      :open_api,
+      :browser_app,
+      :load_operator_grant,
+      :redirect_to_ops_if_operator,
+      :require_authenticated_user_for_private_accounts,
+      :require_sso_authentication,
```
pepicrft, line 1067: High: this public scope still runs the single sign-on redirect. An anonymous visitor has no authentication method in the session, so `require_sso_authentication/2` redirects any public account whose organization enforces a provider instead of rendering the dashboard. Skip this check for anonymously readable accounts, and add coverage for a public account with enforced single sign-on.

server/lib/tuist_web/router.ex
```diff
@@ -1052,6 +1052,44 @@ defmodule TuistWeb.Router do
 
   get "/download", TuistWeb.DownloadController, :download
 
+  # Dashboards a public account exposes to signed-out visitors. Each
+  # LiveView here re-checks `:account_dashboard_read`, which is what
+  # gates the LiveSocket connect — the pipeline below only covers the
+  # dead render.
+  scope "/:account_handle", TuistWeb do
+    pipe_through [
```
pepicrft, line 1060: Medium: this newly anonymous scope omits `:rate_limit`, unlike the public project dashboard scope. It exposes expensive runner and usage pages plus the log-download endpoint to unbounded signed-out traffic. Add the dashboard rate-limit pipeline here before exposing these routes publicly.

server/lib/tuist/authorization.ex
```diff
@@ -356,6 +356,22 @@ defmodule Tuist.Authorization do
     end
   end
 
+  object :account_dashboard do
```
pepicrft, line 365: Low: this new read action admits anonymous visitors and users only. It has no branch for an authenticated project or an account token with a scoped permission, and no description says that it is intentionally browser-only. Document that omission explicitly, or add the appropriate scoped rules, so the policy contract remains complete.

pepicrft (approved):

server/lib/tuist_web/router.ex
fortmarek: Fixed: `require_sso_authentication/2` now skips the redirect for anonymous visitors to a publicly readable account (signed-in users still fall through to enforcement, since they can see member-level data), with three tests covering public+anonymous, public+signed-in, and private+anonymous.

server/lib/tuist_web/router.ex
fortmarek: Fixed: added `:rate_limit` to the public scope, matching the project dashboard scope, plus tests asserting the limit fires and keys on the remote address for signed-out visitors.

server/lib/tuist/authorization.ex
fortmarek: Documented as a comment above the object rather than a `desc`, because LetMe keeps only the last `desc` as the action's description and a trailing one replaced the ops-access text; no scoped rules added, since machine callers read this data through the API objects that carry their own scopes.

Issue #12363, opened 2026-08-14
Kura: per-pod egress ceiling is bypassed on host-network regions


esnunes:
## Summary

On the bare-metal Kura regions, the per-pod egress ceiling (`kubernetes.io/egress-bandwidth`, enforced by the Cilium bandwidth manager) does **not** apply to the customer read path. The shaper itself works correctly — it is simply not in that path. A cache pod can therefore serve reads at many times its configured ceiling, which is the opposite of what the setting is there for.

The ceiling exists so that "one account's restore burst can't monopolize the shared box NIC" (`server/lib/tuist/kura/regions.ex`). On every region whose gateway is host-network, that protection is currently absent.

## Mechanism

The Cilium bandwidth manager works in two halves:

1. A BPF program on the pod's egress stamps each packet with an Earliest Departure Time derived from the pod's configured rate.
2. An `fq` qdisc on the node's **native device** paces packets to those timestamps.

The stamp alone slows nothing; the pacing happens at the physical NIC.

The customer read path on a host-network region is:

```
kura pod  ->  co-located hostNetwork ingress-nginx  ->  client
```

- The first hop is **node-local**. Packets cross the pod veth into the host namespace and are consumed by an nginx socket. They never reach the native device, so `fq` never paces them.
- The second hop originates in the **host network namespace**. Host-namespace traffic carries no pod identity, so a per-pod annotation cannot apply to it.

## Reproduction

Measured against a Tuist-owned instance (our own account, no customer data) on a bare-metal region node whose pod is annotated `kubernetes.io/egress-bandwidth: 500M`. A 200 MiB object was pushed to the cache and read back over three paths, from two throwaway client pods.

| Path | Description | Throughput |
| --- | --- | --- |
| A | cross-node, direct to the pod IP (pod traffic over the NIC) | 462–479 Mbit/s |
| B | cross-node, through the regional host-network ingress | 903–958 Mbit/s |
| C | same-node, direct to the pod IP (node-local) | 13.5–15.5 Gbit/s |

Concurrency is the decisive part, because the pacer is per-endpoint and a real cap must hold as an aggregate:

| Streams | A aggregate | B aggregate |
| --- | --- | --- |
| 1 | 438–459 Mbit/s | 534–609 Mbit/s |
| 4 | 479 Mbit/s | 958 Mbit/s |
| 8 | 462 Mbit/s | 903 Mbit/s |

[...]

PR #12506, opened 2026-08-20
feat(kura): enforce the per-tenant egress ceiling at an HAProxy regional gateway

esnunes:
Implements @pepicrft's proposal on #12363: replace (eventually) the regional gateway with HAProxy and enforce the per-tenant egress ceiling there, using a shared `bwlim-out` filter keyed by the tenant hostname. This PR delivers step 2 of the proposed sequencing — HAProxy with shared `bwlim-out` on one regional gateway, deployed alongside nginx before any cutover — using the haproxytech Kubernetes ingress controller rather than a bespoke gateway.

## Why

The issue shows that `kubernetes.io/egress-bandwidth` (Cilium bandwidth manager) does not govern the customer read path on any host-network region: the pacer lives on the native device, and the read path (`kura pod → co-located hostNetwork gateway → client`) either never reaches it as pod traffic or originates in the host namespace where no pod identity applies. A tenant's restore burst can therefore take the whole box NIC. The gateway is where the bytes actually leave the box, and HAProxy's shared `bwlim-out` is the one mechanism on the table that is a true per-tenant **aggregate** — one `bytes_out_rate(1s)` bucket per hostname shared by every connection and both protocols — which is the property path A had and path B lacked in the issue's measurements.

## What changed

**Platform chart** (`infra/helm/platform`)
- Four new dependency aliases `kura-<region>-haproxy` (haproxytech `kubernetes-ingress` 1.53.1, controller 3.2.13), all disabled in base values, each serving its own `kura-<region>-haproxy` IngressClass.
- Shared per-region config: the `bwlim-out` frontend snippet (limit in bytes/s mirrors `egress_burst_mbps` in `regions.ex`: 62,500,000 B/s = 500 Mbit/s for eu-central/ca-east, 187,500,000 B/s = 1500 Mbit/s for us-east/us-west), `tune.h2.fe.initial-window-size 4194304` + `tune.h2.fe.max-concurrent-streams 32` (the HAProxy equivalent of the nginx `http2_body_preread_size 4m` upload fix, same 128MB worst-case bound), 3600s client/server timeouts, `--disable-http` (no plaintext tenant plane) and `--disable-quic`.
- The stick-table is declared `peers localinstance` so hitless reloads carry the rate counters over instead of briefly lifting the cap.
- Observability lands with the gateway rather than after it: HAProxy's native prometheus exporter is scraped via the standard pod annotations, and access logs go to stdout (the nginx gateways ship neither today).
- **Pilot**: `kura-ca-east-haproxy` is enabled in the staging and canary values (ca-east is the staging/canary-only region). Production values are untouched.

**kura-controller** (`infra/kura-controller`)
- New spec field `haproxyIngressClassName`. When set, the controller renders a second, HAProxy-flavored Ingress pair (`<name>-haproxy`, `<name>-grpc-haproxy`) alongside the nginx pair, and deletes it when cleared.
- A separate pair rather than extra annotations on the nginx objects, for two hard reasons: haproxytech only accepts IngressClasses carrying its own controller string (`haproxy.org/ingress-controller/<class>`), so the nginx class cannot be shared; and it has no regex path routing, so the gRPC paths must be the unescaped begins-with prefixes (`/build.bazel.remote.execution.v2.`, `/google.bytestream.` as `ImplementationSpecific`, which haproxytech implements as `map_beg` — longest prefix wins over the public `/`).
- The Service gains a `grpc` port (4001 → the same co-hosted `http` container port). haproxytech keys its backends on (service, port name), so `haproxy.org/server-proto: "h2"` scoped to the gRPC ingress creates a dedicated h2c backend for REAPI without dragging plain HTTP onto h2 — HTTP/1.1 to the pod is what keeps Kura's `sendfile` accelerated read path alive.
- The public HAProxy Ingress reuses the existing `<name>-public-tls` cert-manager Secret; both gateways terminate TLS for the host during the side-by-side phase. DNS is unaffected (the per-account `DNSEndpoint` already targets the box IP independently of either gateway).

**Server** (`server/lib/tuist/kura`)
- `regions.ex`: a per-region `haproxy_gateway` flag → `haproxy_ingress_class_name` in the provisioner config; set on **ca-east only**. This is the region-by-region rollout gate.
- `kubernetes_controller.ex`: renders `haproxyIngressClassName` on the CR and folds a `+haproxy-gw` suffix into the manifest revision, so flipping the flag re-applies live instances in both directions (the reconciler converges on the revision alone — see the eu-central CRD-skew and reconciler-deadlock incidents).
- CRD schema updated in `infra/helm/tuist/crds/`; the deploy workflow `kubectl apply`s that directory before rollout, and the field is only rendered for ca-east (staging/canary), so schema-first ordering holds.

## Deliberately out of scope

- **No cutover.** nginx keeps `:443`; HAProxy binds `:8443` (its container default, unprivileged). The pilot is validated at `https://<tenant-host>:8443`. Cutover is a later change: flip `controller.containerPort.https` to 443 and disable the region's nginx alias.
- **The Cilium pod annotation stays** as defense in depth until a cutover decision.
- **The peer plane (`:7443` SNI demux) keeps the same bypass** the issue notes; any fix there is separate.
- **Kernel-level enforcement** (`set-fc-mark` + HTB, Cilium BWM off) remains its own future decision, as proposed.
- As pepicrft framed it: this is a **forwarded-data ceiling, not a physical one** — TLS overhead, retransmits, and already-buffered bytes are outside it.

## Validation

- `go test ./...` in `infra/kura-controller` passes, including new tests for the HAProxy pair (render, skip-when-unset, delete-on-clear) and the `grpc` Service port.
- `mix test test/tuist/kura/regions_test.exs test/tuist/kura/provisioner/kubernetes_controller_test.exs` passes (112 tests), including the new pilot-gating, CR-rendering, and revision-suffix tests.
- `helm template` of the platform chart with the staging and canary overlays renders cleanly; the ca-east DaemonSet comes out hostNetwork on the `kura-ca-east` pool with the expected args, snippets, tolerations, and scrape annotations.

Not yet validated (needs a live cluster, part of the pilot): actual bwlim behavior under concurrency at `:8443`, gRPC/ByteStream end-to-end through haproxytech, and rate-counter fidelity across hitless reloads. Per the issue, acceptance must be measured at the physical interface over short intervals and re-run after gateway/Cilium/node restarts.

Closes nothing yet; tracks #12363.


esnunes:
closed in favor of https://github.com/tuist/tuist/pull/12525

PR #12507, opened 2026-08-20
feat(infra): enforce kura egress ceilings with a pod-side qdisc init container

esnunes:
Implements the approved design for [#12363](https://github.com/tuist/tuist/issues/12363): enforce the per-tenant Kura egress ceiling with a root qdisc installed inside each pod's network namespace by an init container.

## Root cause

On the four bare-metal regions the `kubernetes.io/egress-bandwidth` annotation is not enforced on the customer read path. Cilium's bandwidth manager paces at the `fq` qdisc on the **physical NIC** only. The read path is `kura pod → co-located hostNetwork ingress-nginx → client`: the first hop is node-local (never reaches the NIC) and the second originates in the host namespace (no pod identity). A tenant can serve reads at many multiples of its ceiling.

Reproduced in the lab below: with only the annotation (500M), a node-local single stream measured **2290 Mbit/s** — ~4.6× the ceiling.

## What changed

- **`infra/egress-qdisc-init/`** — new init-container image (alpine + `iproute2-tc`; busybox `tc` lacks htb). The entrypoint installs `htb` (rate = ceil = the annotation's Mbit/s) with an `fq_codel` leaf on the pod's `eth0` and exits non-zero on any failure. Built by `.github/workflows/egress-qdisc-init-image.yml` → `ghcr.io/tuist/egress-qdisc-init`.
- **`infra/kura-controller/`** — `podTemplate` renders the init container exactly when the instance spec carries the `kubernetes.io/egress-bandwidth` annotation (the same condition that renders the annotation; no server change needed). `EGRESS_BURST_MBPS` is parsed from the annotation's `<n>M` value; a malformed value passes through verbatim so the entrypoint rejects it and the pod fails closed. The image arrives via `KURA_EGRESS_QDISC_INIT_IMAGE` (same pattern as `KURA_PEER_DEMUX_IMAGE`); the controller **refuses to reconcile** an annotated instance when the image is unconfigured instead of rendering it unshaped. Init security context: `NET_ADMIN` only, all other capabilities dropped, no privilege escalation, read-only rootfs.
- **`infra/helm/tuist/`** — `kuraController.egressQdiscInitImage` value → controller Deployment env.
- The annotation is **kept**: it still paces the wire-bound peer/replication plane; double-shaping is harmless (the lower limit wins).

## Why pod-side qdisc (vs. the obvious alternatives)

- **Fail-closed**: `tc` failure ⇒ the pod never starts. Gateway/nginx-side limiters fail open (a misconfigured or bypassed gateway serves unshaped).
- **Topology-invariant**: shapes at the first hop, so node-local nginx, `:7443` peers, and NodePort/ClusterIP regions are all covered.
- **No ownership conflict**: Cilium programs the host side (`lxc*`, NIC) and installs nothing inside pod namespaces.
- **Lifecycle by construction**: the qdisc is born and dies with the pod netns — no agent, no reconciler, no cleanup.
- **Zero steady-state footprint**: an init container, not a sidecar; the regions pack many instances per box.

Related: #12506 pilots gateway-side enforcement (HAProxy `bwlim-out`) on ca-east. The two are complementary rather than competing: this change bounds the tenant at pod scope regardless of path or topology and fails closed; a gateway-side limiter is the natural home for the **box-level aggregate** cap this change deliberately does not solve (per-pod ceilings oversubscribe the NIC by design).

## Lab validation (Cilium, kernel-enforced, per the design's checklist)

Environment: single-node k3s reconfigured to **Cilium 1.18.7** (same version as the fleet) with `kubeProxyReplacement`, `bpf.masquerade`, and `bandwidthManager: {enabled, bbr}` — status reports `BandwidthManager: EDT with BPF [BBR]`. Read path reproduced with a hostNetwork nginx (`proxy_buffering off`) proxying to the shaped pod; measurements through that path with the real image built from this branch.

| Check | Result |
|---|---|
| 1. Aggregate enforcement (the decisive test) | 500M ceiling: **478.6 / 478.6 / 478.5 Mbit/s** at 1 / 4 / 8 concurrent streams — flat, no per-stream multiplication |
| 2. Cilium non-interference | CiliumNetworkPolicy add/remove cycle **and** full cilium-agent restart: qdisc survives (byte counters continuous) and still enforces (478.7 Mbit/s post-regen) |
| 3. netkit | Not verified — lab and fleet both use veth today. Flagged per design: re-run before any netkit migration |
| 4. Rate accuracy | 478.6/500 and 1434.9/1500 = **95.7%** of ceiling at both rates (HTTP payload vs. L2 shaping overhead accounts for most of the ~4.3%; within the ~5% budget) |
| 5. Probe latency under saturation | 8 bulk streams saturating the ceiling; tiny-request latency **p50=1 ms, p95=1 ms, max=1 ms** — the fq_codel leaf makes the deferred high-priority class unnecessary |
| 6. Fail-closed | `EGRESS_BURST_MBPS="1.5G"` ⇒ pod stuck `Init:Error`, never serves; log states the rejected value |
| Bypass baseline | Annotation only, no init container: **2290 Mbit/s** through the node-local path against a 500M annotation — the bug this PR fixes |

Also verified: `go test ./...`, `go vet`, `gofmt` clean; shellcheck clean on the entrypoint; `helm template` renders the env exactly when the value is set; Pod Security check — the kura namespaces carry no PSA `enforce` labels on the workload clusters (Talos-defaulted PSA exists only on the mgmt cluster), so `NET_ADMIN` is admissible.

## Rollout (not in this PR)

1. Merge builds the image on main; pin `kuraController.egressQdiscInitImage` **by digest** in the managed values (ca-east first, then eu-central, then us-east/us-west).
2. Rolling restart per region picks up the init container; the 2-replica warm-standby design covers gapless deploys.
3. Acceptance per region at the physical interface under concurrency; re-run after a gateway restart, a Cilium agent restart, and a node reboot.
4. Alerting on NIC egress exceeding the node's advertised `tuist.dev/egress-mbps` budget (can land independently; catches any future silent bypass).

## Deliberately out of scope

Box-level aggregate cap, retransmit wire-truth on the nginx→client leg (depends on `proxy-buffering: off` staying set), bandwidth floors (`tuist.dev/egress-mbps` bin-packing), and per-pod qdisc metrics (unprivileged netlink read; can join kura's metrics endpoint later).


esnunes:
closed in favor of https://github.com/tuist/tuist/pull/12525

PR #12511, opened 2026-08-20
ci(cli): cap acceptance test runner parallelism and drop wall-clock time limits

pepicrft:
Related: https://github.com/tuist/tuist/issues/12512

`BuildAcceptanceTestSwiftPMPrebuiltMacro` has been failing several times a day. Every single failure is the same thing, and it is not an assertion:

```
Test exceeded execution time allowance of 4 minutes
```

Every failing run recorded a duration of exactly `240000`ms, which is the `.timeLimit(.minutes(4))` on the test itself. The test never failed a check about prebuilt macro wiring. It ran out of clock.

### Root cause

The time limit measures wall clock, and the test spends nearly all of that wall clock waiting for a saturated machine rather than doing its own work.

This is not inferred from timings. Xcode captures a spindump when a test exceeds its execution time allowance, and the one attached to run [79b1dfbb](https://tuist.dev/tuist/tuist/tests/test-runs/79b1dfbb-d13b-4814-8113-dcf480feeb5d) shows the machine at the exact moment of the failure:

```
Hardware model:   VirtualMac2,1
Active cpus:      6
Memory size:      14 GB
Total CPU Time:   27.813s      (of a possible 30s across 6 cores = 93% saturated)
```

There were 1,199 processes alive, including **78 concurrent `swift-frontend`** compilers, on six cores. One `tuist test` invocation starts three parallel `xctest` workers, each runs Swift Testing tests in parallel, and each of those tests shells out to its own nested `xcodebuild`. Ancestry from the dump:

```
swift-frontend <- SWBBuildService <- xcodebuild <- xctest[50558] <- xcodebuild[48830] <- tuist[44592]
```

Split by worker at the moment of the timeout:

| Worker | Descendant processes | `swift-frontend` | CPU in the 5s window |
| --- | --- | --- | --- |
| `xctest[50556]` | 18 | 16 | 2.42s |
| **`xctest[50557]` (ours)** | **74** | **0** | **1.77s** |
| `xctest[50558]` | 60 | 37 | 10.65s |

The worker running our test held **zero** compiler processes. It had spawned its `xcodebuild` and was waiting on it while 53 compilers belonging to the other two workers consumed the box. It was starved, not busy.

For calibration, the same test measured locally on an unloaded machine, with the toolchain matched to the runner:

| Phase | Time |
| --- | --- |
| `tuist install --force-resolved-versions` (cold cache) | 13.2s |
| `tuist generate --no-open` | 2.6s |
| `xcodebuild build` Release | 6.8s |
| End to end | **15.6s** |

So roughly 16 seconds of real work was stretching past 240.

### This is not swift-syntax being compiled

Worth stating explicitly, because it is the obvious first guess and it is wrong. The string `swift-syntax` does not appear anywhere in the 13MB spindump, and the worker running our test held no compiler processes at all. The prebuilt macro path is working, nothing regressed into resolving swift-syntax from source, and Swift Package Manager resolution is not the bottleneck. Its `swift-package` process was parked in a runloop wait using 0.5s.

### Why the test, and why now

Nothing about this test changed. The suite got more concurrent, so runs finish sooner while each individual test waits longer:

| Date | Tests | Avg test | Run wall clock | Effective concurrency |
| Aug 11 | 207 | 21.2s | 649s | 6.7x |
| Aug 14 | 208 | 23.9s | 565s | 8.8x |
| Aug 18 | 208 | 22.5s | 412s | 11.4x |
| Aug 20 | 207 | 29.2s | 368s | 16.4x |

Run wall clock is down 43% while per test wall clock is up around 38%. That throughput win quietly ate this test's headroom.

In that Aug 20 run, 20 tests took 180s or more and 6 took 240s or more. The test that failed was not even the slowest one:

| Test | Duration | Result |
| --- | --- | --- |
| `multiplatform_app_module_cache` | 289s | passed |
| `app_with_plugins` | 267s | passed |
| `generated_static_framework_with_bundle_macro_from_cached_binaries` | 265s | passed |
| `generated_ios_app_with_cached_xctest_support` | 263s | passed |
| `multiplatform_app_selective_testing` | 255s | passed |
| `app_with_swiftpm_prebuilt_macro_dependency…` | **240s** | **failed** |

Five tests were slower and all of them passed. The only difference is that this one carried a cap, so it was reporting how busy the machine happened to be rather than anything about the code under test.

### What changed

Two things: one stops the false failures, the other reduces the oversubscription that caused them.

**Capped the test runner parallelism.** Added `-maximum-parallel-testing-workers 2` to the sharded acceptance job, which is Marek's suggestion and bounds the outermost multiplier. Two is a starting point rather than a tuned number: it roughly halves module-level concurrency while leaving plenty of headroom against the job's 60 minute timeout, given the run currently finishes in about six minutes. It should be tuned once step 1 of #12512 makes runner load observable.

Being explicit about what this does not do: it caps runner **processes**, so it bounds how many test modules run side by side. It does not bound Swift Testing's in-process concurrency or the nested builds. `xctest[50558]` above was running three concurrent nested `xcodebuild` processes on its own, so the fan-out is reduced rather than removed. Capping the nested build job count is the bigger lever and is tracked in #12512.

I deliberately left the fork job alone. It runs the whole suite unsharded on one runner with a 30 minute timeout, and slowing it down without first measuring its wall clock risks turning a passing job into a timeout for external contributors.

**Removed `.timeLimit` from the prebuilt macro build test.** It has no hang to guard against, and every other test in that module already runs without a cap, including several that sit in the 250s to 290s range.

**Raised the three limits in `DependenciesAcceptanceTests` from 4 to 15 minutes** rather than deleting them. Those exist for a different reason: they guard against `install` followed by `generate` deadlocking on the Swift Package Manager scratch directory lock, which hangs forever rather than merely running slow. A limit there has real value, it just has to be sized to catch a hang instead of bounding normal runtime. Fifteen minutes still fails well inside the job's 60 minute timeout but cannot trip on co-scheduling. Two of those tests were already running at about 160s against their 240s budget, so they were days away from failing the same way.

I also corrected the comment above one of them, which justified the old number by referring to a "1-minute default" that does not exist anywhere in this project. Tests without the trait have no limit at all, which the 289s pass above demonstrates.

That distinction is the reason the time limits are not all simply deleted. A cap that only bounds slowness is noise; a cap that catches an infinite hang is a real assertion.

### Why not just raise the failing test to 8 minutes

That resets the same countdown. The suite has been getting more concurrent, not less, and per test wall clock roughly doubled in nine days. Any fixed ceiling on a shared machine eventually gets crossed again.

If we later want a genuine guard on that test, the right shape is a timeout on the individual `tuist` subprocess invocations. That catches an actual hang and is immune to whatever else is scheduled alongside it. I did not build that here because nothing in the test's history suggests it hangs.

### Impact

Removes a recurring red build that carried no signal, and defuses three more tests on the same trajectory. No production code is touched and no test coverage is lost, since none of these limits were asserting anything about behaviour.

The worker cap should also make the whole shard less thrash-prone, not just this test, since every acceptance test on that runner was competing for the same six cores. Expect shard wall clock to move; whether it goes up or down depends on how much of the current 368s is useful work versus context switching, which is exactly what step 1 of #12512 would tell us.

It does not fully solve the oversubscription. Bounding the nested builds and deciding on runner sizing are still open in #12512, along with a `.serialized` annotation on `GenerateAcceptanceTestiOSAppWithModuleMapPackages` that was added to mitigate this same contention and turns out to be a no-op.

### How to test locally

The failure only reproduces under load, so the useful local check is that the test still does what it claims in a reasonable time:

```sh
# Fixture lives in examples/xcode, copied to a temp dir by the .withFixture trait
tuist install --path <fixture> --force-resolved-versions   # ~13s
tuist generate --no-open --path <fixture>                  # ~3s
xcodebuild build -workspace <fixture>/App.xcworkspace -scheme App \
  -configuration Release -skipMacroValidation \
  CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""   # ~7s
```

One gotcha if you run that by hand: Swift Package Manager picks the prebuilt keyed to whichever `swift` is first on `PATH`, not the Xcode toolchain that compiles the macro. With swiftly installed, its 6.3.2 shadows Xcode's 6.3.3 and the build fails with `Unable to resolve Swift module dependency to a compatible module: 'SwiftSyntaxMacros' ... compiled with an older version of the compiler`. Put the Xcode toolchain first and it resolves the matching prebuilt and builds.

Verification actually run on this branch:

- `swiftformat --lint` on both changed files: `0/2 files require formatting`
- `swiftlint` on both changed files: only pre-existing `type_name` warnings on untouched types
- `xcrun swiftc -parse` on both changed files: clean
- `.github/workflows/cli.yml` parses as valid YAML after the flag was added
- `-maximum-parallel-testing-workers` confirmed in `man xcodebuild`, and confirmed not to be in `tuist test`'s `notAllowedPassthroughXcodeBuildArguments` list, so it forwards through the `--` terminator

`mise run lint` does not complete in a fresh worktree without `tuist install` first, and it fails on graph loading before reaching anything related to this change.


.github/workflows/cli.yml
```diff
@@ -348,8 +348,13 @@ jobs:
...
           security default-keychain -s $KEYCHAIN_PATH
           security unlock-keychain -p $KEYCHAIN_PASSWORD $KEYCHAIN_PATH
+      # -maximum-parallel-testing-workers caps how many test runner processes xcodebuild spawns.
+      # Acceptance tests each shell out to their own nested xcodebuild, so every extra runner
+      # multiplies into a fresh set of compiler processes. Uncapped on a 6-core runner that reached
+      # 78 concurrent swift-frontend processes, which is what starved tests into their time limits.
+      # See https://github.com/tuist/tuist/issues/12512 for the measurements behind this number.
       - name: Run shard ${{ matrix.shard }}
-        run: tuist test --without-building --platform macOS --shard-reference "$TUIST_ACCEPTANCE_SHARD_REFERENCE" TuistAcceptanceTests -- -retry-tests-on-failure -test-iterations 2
+        run: tuist test --without-building --platform macOS --shard-reference "$TUIST_ACCEPTANCE_SHARD_REFERENCE" TuistAcceptanceTests -- -retry-tests-on-failure -test-iterations 2 -maximum-parallel-testing-workers 2
```
fortmarek, line 357: Hmm, can this be higher? 2 feels too low

fortmarek (approved):

.github/workflows/cli.yml
pepicrft: I think two is the right initial cap. The latest continuous-integration run completed the two acceptance shards successfully in 134 and 120 seconds, so there is no observed throughput concern. The cap only limits the outer test runners; the acceptance tests still launch nested `xcodebuild` work, which is why the prior uncapped run saturated the six-core runner. I would keep two until #12512 provides the runner-load measurements needed to tune it safely.

PR #12525, opened 2026-08-20
feat(infra): enforce kura per-tenant egress floors and the node box cap with a shared per-node HTB tree

esnunes:
One shared HTB tree per node, fed by every kura pod on it, enforcing three things the current egress stack cannot: per-tenant **floors**, a kernel-enforced **box cap**, and a **cross-replica 1×C** ceiling. Kernel-level enforcement for #12363.

## The problem

Kura regions promise each tenant `egress_guaranteed_mbps` (a floor) and `egress_burst_mbps` (a ceiling). Today neither is truly enforced:

- The `kubernetes.io/egress-bandwidth` annotation (Cilium's bandwidth manager) paces at the physical NIC only. The node-local read path — kura pod → co-located hostNetwork gateway → client — never crosses the NIC's pacer, so the revenue path is effectively uncapped. The HAProxy gateway (#12506) bounds it at L7 on the ca-east pilot only.
- The floor exists only as scheduler bin-packing (the `tuist.dev/egress-mbps` extended resource). Nothing at runtime guarantees a tenant its share when a noisy neighbor saturates the box.
- Per-pod mechanisms cannot fix this by construction: a per-pod ceiling gives two replicas 2×C, and isolated per-pod shapers cannot arbitrate *between* tenants, which is what a floor is.

Only a single tree that all tenants on a node share can provide floors (HTB `rate` is work-conserving: each class is guaranteed its rate and borrows idle bandwidth up to its `ceil`), a box-wide cap (the root class), and 1×C across replicas (both replicas feed one class).

### Control plane

The server already renders `egress_burst_mbps` into the `kubernetes.io/egress-bandwidth` pod annotation and `egress_guaranteed_mbps` into `KuraInstance.spec.egressGuaranteedMbps` — both unchanged. On top of that, **kura-controller** now:

1. Allocates each shaped account a stable HTB classid minor (range `0x100`–`0xFFFE`; FNV hash of the account handle + linear probe; persisted as the `kura.tuist.dev/egress-class-id` CR annotation so it never changes for a live account). All instances of an account share one id — it keys the tenant's class on every node — so allocation scans the namespace's KuraInstances and adopts an existing claim before probing.
2. Renders one pod annotation, `tuist.dev/egress-class: {"classid":"1:<hex>","floor_mbps":F,"burst_mbps":B}`. That annotation is the entire control-plane→datapath contract, and it is inert until an agent runs on the pod's node.

### Datapath: a veth trampoline, per node

The new **egress-tree-agent** DaemonSet owns a host veth pair, `kura-egress0/1`, with the HTB tree as `kura-egress0`'s root qdisc: root class `1:1` at the node's advertised `tuist.dev/egress-mbps` budget (the box cap), one class `1:<minor>` per tenant (`rate` = floor, `ceil` = burst clamped to the budget, fq_codel leaf), `default 0` so an unclassified packet transmits via the direct queue and trips an alert counter.

Per shaped pod, a small tcx BPF program (`kura_shaper_out`, ~70 lines in `internal/agent/bpf/redirect.c`) is attached to the pod's host-side veth (`lxc*`) ingress hook, anchored **first** — ahead of Cilium's `cil_from_container`. Per packet it:

1. **Loop guard** — if `tc_index` carries the shaped marker, this is the packet's second pass: clear it and fall through to Cilium.
2. **Sibling bypass** — if the IPv4 destination is a co-located same-account pod (an agent-maintained per-`/32` allowlist in a BPF map), fall through untouched: replica-to-replica sync stays on Cilium's `redirect_peer` fast path at node-local line rate and never consumes the tenant bucket. The allowlist polarity is fail-safe — a stale entry briefly *shapes* sibling traffic, never unshapes tenant traffic.
3. Otherwise **stamp** `skb->priority = 1:<minor>` (HTB classifies by priority natively, so the tree device needs no filters), record the source ifindex in `skb->mark`, arm the loop guard, and redirect into `kura-egress0`.

HTB paces the packet, it crosses the veth, and `kura_shaper_ret` on `kura-egress1` sends it back to the original device's ingress hook via `BPF_F_INGRESS` — a plain re-entry with no skip flags, where the loop guard disarms and `cil_from_container` runs. **Cilium processes every shaped packet exactly once**: policy, identity, service handling, and masquerade apply exactly as for an unshaped packet. Anything on the trampoline without shaping metadata is dropped and counted — nothing legitimate transmits there.

### The agent's reconcile model

Event-driven with a slow backstop, the same split Cilium's bandwidth manager uses. Field-selected pod/node informers (this node's pods, this node's Node object) kick a debounced (1s) converge; only shaped-pod events and egress-capacity changes trigger. A netlink link watch on the trampoline pair covers the one state no informer reports quickly enough — a deleted or downed trampoline device. The periodic sweep (`RECONCILE_INTERVAL`, default 2m) repairs what nothing reports: stripped tcx links, stale pins. A Running pod whose Cilium endpoint has not appeared yet requeues a 2s retry.

Each cycle is idempotent: converge the devices and tree, confirm the return program, resolve pod→`lxc*` through the local Cilium agent's endpoint API (the component that created the device is the authority), attach/verify per-pod programs (first position re-checked; maps synced in place, no reattach for config changes), clean stale pins, export metrics. tcx links are bpffs-pinned, so enforcement survives agent restarts and upgrades; a restart adopts the pinned links with zero churn.

### Failure semantics

Every failure state resolves to **unshaped or a bounded drop — never a blackhole, never a policy bypass**:

- No advertised node budget and no `DEFAULT_NODE_EGRESS_MBPS` fallback: no tree, pods stay unshaped, `skipped_pods` carries the alert signal. The agent never invents a cap.
- An unconfigured or partially configured pod program passes packets to Cilium unshaped.
- Hard ordering: no pod program is attached or kept without a confirmed return program. IPv4 forwarding is disabled on both trampoline ends, so if the return program is ever missing, surfacing packets are *dropped by the stack* rather than forwarded around Cilium — the one deliberate drop state. If the return program stays unattachable for `RETURN_ATTACH_MAX_FAILURES` consecutive attempts (default 3, fast-retried), the agent detaches every pod program: unshaped beats blackholed.
- An absent or failing agent leaves pods on the plain Cilium path (the NIC annotation still paces the wire leg). Deliberate fail-open, covered by alerts rather than a fail-closed pod layer.

### Rollout controls and observability

- `egressTreeAgent.enabled` — ships **disabled**.
- `egressTreeAgent.betaPodPrefix` (`BETA_POD_PREFIX`) — when set, the agent attaches only to pods whose name starts with the prefix, so the first enforcement step can target one account before the fleet. Excluded pods stay unshaped and count in `kura_egress_tree_beta_excluded_pods` (separate from the alerting `skipped_pods`). Sibling allowlists are still computed over all annotated pods, and prefix changes converge within one cycle in both directions.
- Prometheus metrics on `:9469`: per-class bytes/drops/backlog, direct-queue packets (any growth means a broken stamp or foreign redirect), per-pod redirect/guard/bypass counters, return drops, reattach churn, return-attach failures, skipped/excluded pods. State transitions (class added/updated, pod attached/updated/removed, budget changes) log at info level; quiet converged cycles log nothing.

## Why this exact datapath shape

Three constraints, verified on a fleet-shaped Cilium 1.18 lab (TCX attach, BPF host routing, veth, vxlan, bandwidth manager), dictate the design; regressing any of them breaks it:

1. **Legacy `tc` filters on the lxc hook never run.** Under TCX attach, Cilium's program always returns a verdict, and the kernel runs the legacy clsact path only when no tcx program gave one — a counting filter measured zero hits across every traffic leg. Redirecting before Cilium is only possible as a tcx program anchored first. Cilium replaces its own link in place across endpoint regeneration and agent restarts and leaves foreign links alone, so the anchored position is stable (and re-verified every cycle anyway).
2. **`skb->priority` cannot carry the tenant class out of the pod.** The kernel scrubs it on every netns crossing (verified on a plain veth pair with no CNI: tagged inside, zero on the host side). So classification lives host-side, in the per-lxc program, from agent-supplied config — which also means the workload can never spoof its own class (verified: a pod setting `SO_PRIORITY`/`SO_MARK` to another tenant's class still lands 100% in its own class). This is also why there is no in-pod shaping component in this PR: nothing tenant-scoped survives the crossing, and an in-pod shaper would throttle the sibling sync the bypass exists to keep at line rate.
3. **The classic ifb detour voids NetworkPolicy.** ifb re-injects with `tc_skip_classify`, which skips the entire tcx hook; shaped traffic re-enters via the host stack with the trusted host identity. Measured with policy controls: both a deny-all-egress policy on the shaped pod and a destination ingress policy excluding it were silently bypassed. The veth trampoline's `BPF_F_INGRESS` re-entry carries no skip flag, and both policy directions were re-verified enforced through it — including on *bypassed* sibling traffic (the bypass skips the trampoline, never Cilium).

## What this deliberately does not do

- Replace the NIC annotation (it stays as the wire-leg pacing) or the #12506 gateway (L7 observability stays there).
- Touch the gateway→client leg (wire-truth accounting is out of scope).
- Support netkit (the fleet runs `datapath-mode: veth`; the lxc hook is the load-bearing attachment point and must be re-verified before any datapath migration).
- Enforce floors in the scheduler: the per-replica floor double-count in bin-packing is a known separate issue and a precondition for floors going live.

## Changes

- **`infra/egress-tree-agent/`** (new): the DaemonSet agent — BPF datapath (`internal/agent/bpf/redirect.c`; bpf2go bindings generated at build time, in CI and in the Dockerfile builder stage, never committed), tree/device management via `ip`/`tc`, tcx attach via cilium/ebpf with pinned links, informer-driven reconcile, Prometheus metrics. `AGENTS.md` documents the architecture, invariants, and alert set.
- **`infra/kura-controller/`**: classid allocation + `tuist.dev/egress-class` rendering (`kurainstance_controller.go`), with unit tests for allocation determinism/probing/sharing/idempotence and annotation rendering.
- **`infra/helm/tuist/`**: `egressTreeAgent` values + DaemonSet/RBAC template (privileged, hostNetwork, kura-pool nodeSelector/tolerations).
- **`.github/workflows/egress-tree-agent-image.yml`**: vet/test (with BPF generation) + multi-arch image to `ghcr.io/tuist/egress-tree-agent`.
- `go.work` registers the new module; repo-map `AGENTS.md` nodes updated.

## Validation

A 30-scenario test matrix ran twice against this branch **built with its own Dockerfile and deployed as the privileged DaemonSet** on a lab cluster configured like the fleet (Cilium 1.18.7: TCX attach, BPF host routing, veth, vxlan tunnel, bandwidth manager; two nodes with disjoint pod slices; hostNetwork proxies reproducing the gateway/demux router-IP legs; an off-node client). The latest run is at the current head. **29 of 30 scenarios pass**; the one remainder (a real node reboot) is covered by a cold-rebuild simulation (full convergence from zero host state in 2 s) and deferred to observe mode.

<details>
<summary><b>All 30 scenarios evaluated</b> (click to expand)</summary>

### 1 · Routing — one scenario per traffic leg (asserted via class/BPF counters, not just throughput)

| # | Scenario | Result | Evidence |
|---|---|---|---|
| 1.1 | Customer response via gateway (router-IP leg) | ✅ | pod's peer verified as the `cilium_host` router IP; ceiling held at 1/4/8 streams (288–296 of 300 Mbit/s), aggregate capped across concurrent legs |
| 1.2 | Sibling sync forward (co-located replica) | ✅ | 149 Gbit/s via `redirect_peer`, bypass counter +1.7 M pkts, class bytes unchanged |
| 1.3 | Sibling sync reverse (both allowlists) | ✅ | 151 Gbit/s, the other device's bypass counter increments — symmetric by traffic, not just map dump |
| 1.4 | Remote in-cluster peer over vxlan (second node, disjoint slice) | ✅ | capped at 286 of ceil 300 vs 772 Mbit/s link baseline; class counted the bytes; `cilium_vxlan` TX delta ≈ transfer; delivered intact cross-kernel |
| 1.5 | Peer-serving responses via demux (off-node client → hostNetwork proxy → pod) | ✅ | 137 MB pulled by an off-node client, tenant class counted the bytes — the "looks internal" leg is not |
| 1.6 | Outbound off-cluster peer | ✅ | 30 MiB delivered byte-exact to an external host; source = node IP (BPF masquerade intact); class counted 33 MB |
| 1.7 | Control plane (cluster DNS via socket-LB, external TLS) | ✅ | DNS OK, HTTPS 200; DNS stays sub-ms under class saturation (see 4.2) |
| 1.8 | Inbound directions untouched (requests, probes) | ✅ | only `tcx/ingress` attachments on the pod veth; host→pod RTT 16 µs vs 15 µs to an unshaped pod |
| — | Negative-space allowlist assert | ✅ | map dumps contain exactly the co-located same-class pod IPs; never the router IP, a remote peer, or a foreign tenant; unannotated pods carry no attachment |

### 2 · Policy integrity (the trampoline's reason to exist)

| # | Scenario | Result | Evidence |
|---|---|---|---|
| 2.1 | Deny-all-egress on a shaped pod blocks every leg through the trampoline | ✅ | pod + external blocked; unshaped control confirms the policy itself |
| 2.2 | Destination ingress policy excluding the shaped source blocks | ✅ | shaped source blocked, allowed tenant unaffected |
| 2.3 | The rendered kura NetworkPolicy behaves identically with the agent enabled vs fully detached | ✅ | identical 6-row allow/deny matrix (same-instance all-ports · same-account peer port · any-namespace http port) |
| 2.4 | Policy applies to **bypassed** sibling traffic | ✅ | sibling pings blocked while the bypass counter incremented by exactly the ping count — the bypass skips the trampoline, never Cilium |
| 2.5 | Exactly-once processing (loop guard) | ✅ | 200 MiB → 219.6 MB counted (headers, not 2×); redirected == guard-pass |

### 3 · Lifecycle and failure (invariant: unshaped or bounded drop — never blackholed, never policy-bypassed)

| # | Scenario | Result | Evidence |
|---|---|---|---|
| 3.1a | Agent down; existing pods | ✅ | pinned links keep enforcing; restart adopts with 0 reattach churn and logs the standing state once |
| 3.1b | Agent down; new pod starts | ✅ | flows unshaped (40 Gbit/s); shaped ~2 s after the agent returns healthy |
| 3.2 | Return program lost / never attachable (the one deliberate drop state) | ✅ | pin removed live: 4.0 s drop, self-healed; guard: broken trampoline → loud error, zero pod programs, fail-open; escalation: persistently unattachable → 3 logged failures at 2 s retries → all pod programs detached (unshaped, never blackholed), recovery 1 s |
| 3.3 | Trampoline device deleted / HTB tree cleared by hand | ✅ | device deletion: 1.0 s drop, rebuilt in 1 s (netlink watch); tree-only clear: 0 lost packets, rebuilt at the backstop |
| 3.4 | Sibling restarts with a new IP | ✅ | shaped-not-dropped through the staleness window; allowlist converged ~1 s after Ready; bypass resumes at line rate |
| 3.5 | Kura pod restart (new lxc, new endpoint) | ✅ | attached ~2 s after Ready |
| 3.6a | Cilium endpoint regeneration; Cilium agent restart | ✅ | tcx link stays attached and anchored first (Cilium replaces its own program in place) |
| 3.6b | Node reboot | ⚠️ simulated | cold rebuild from zero host state with pods running converges in 2 s; a real reboot is deferred to ca-east observe mode (shared lab box) |
| 3.7 | Annotation change / removal under a live stream | ✅ | clean 288→143 Mbit/s step ~2 s after the change; zero interruption through an agent restart; removal detaches + cleans within one cycle |
| 3.8 | No node budget and no fallback | ✅ | no tree, all programs detached, pods unshaped, `skipped_pods` exported as the alert signal |

### 4 · QoS under load

| # | Scenario | Result | Evidence |
|---|---|---|---|
| 4.1 | Floors under contention | ✅ | three tenants saturating: 243/568/143 Mbit/s vs floors 200/300/100 — every tenant ≥ floor, work-conserving excess, sum ≈ the box cap |
| 4.2 | Probe + DNS latency while the tenant's class is saturated | ✅ | HTTP probe p50 6.4 ms / p95 7.8 ms (idle 0.5/0.7); DNS sub-ms — the fq_codel leaf holds |
| 4.3 | Sibling bypass under simultaneous class saturation | ✅ | 40 Gbit/s replica traffic while the class is pinned at its ceiling |
| 4.4 | Aggregate throughput + CPU | ✅ | 18.6 Gbit/s through the tree at a 20 Gbit root, ~5% total CPU on 24 cores |

### 5 · Adversarial

| # | Scenario | Result | Evidence |
|---|---|---|---|
| 5.1 | Workload sets `SO_PRIORITY` / `SO_MARK` (class spoofing) | ✅ | 50 MiB sent with another tenant's classid: 100% lands in the sender's own class, 0 bytes in the target's |
| 5.2 | Forced unstamped packet into the tree | ✅ | one raw injected frame: `direct_packets` 0→1 **and** `return_dropped` 0→1 — both tripwires fire, nothing leaks into a pod; steady-state direct count across the whole matrix: 0 |

**Deferred to ca-east observe mode** (already in the rollout plan): the real-gateway variant of 1.1 (ingress-nginx/HAProxy buffering; the router-IP path itself is covered), a real node reboot (3.6b), an hour-long stream (3.7 ran 60–90 s across both events), and region-hardware re-runs of 4.1/4.4.

</details>

Also green: `go test` / `go vet` in both modules (workspace mode), the Docker image build (with in-builder BPF generation), and `helm template` across value variants.

**Packet-rate runs** (review follow-up; every load scenario above is a byte-rate test and two costs scale with skb/s, not bytes: the two extra per-CPU backlog trips per shaped packet, and the single `sch_htb` root lock). k01, 1.4 KB UDP datagrams (worst case: the eBPF path is protocol-blind, and real TCP GSO-coalesces into far fewer skb/s at the same byte rate), 5-minute sustained runs, shaped vs fully detached, monitoring CPU, `SUnreclaim` slab, and `/proc/net/softnet_stat`:

| Run | Sender rate | Loss | softnet drops | HTB drops | CPU |
|---|---|---|---|---|---|
| 1 pod, unshaped | 811 kpps / 9.1 Gbit/s | 0.18% | 0 | — | 27.7% |
| 1 pod, shaped | 630 kpps / 7.1 Gbit/s | 0.14% | 0 | 0 | 24.2% |
| 50 pods, unshaped | 3.31 Mpps attempted, ~1.16 Mpps delivered | 64.9% (sink sockets) | 0 | — | 94.8% |
| 50 pods, shaped | 1.43 Mpps, ~1.41 Mpps delivered | 1.74% | 0 | 0 | 56.7% |

The root lock manifests as sender backpressure, not drops: shaped senders were paced to 1.43 Mpps aggregate (vs 3.31 Mpps attempted unshaped), which let the receivers keep up — delivered throughput was *higher* shaped than unshaped under overload. Single-flow PPS ceiling costs ~22%. Zero softnet backlog drops in all four runs, so `netdev_max_backlog` needs no tuning at these rates; since backlog overflow stays invisible to every `kura_egress_tree_*` counter, the alert set now includes `node_softnet_dropped_total` on shaped nodes (node-exporter, already fleet-wide). Slab drift over 5 minutes was indistinguishable between shaped and unshaped runs.

**Controller-driven churn run** (k01, both components rebuilt at this branch's HEAD): the kura-controller managed 12 KuraInstance resources — 8 shaped accounts (`egress-bandwidth: 500M`) and 4 unshaped controls — so pods, `tuist.dev/egress-class` annotations, and classid allocation all came from the real controller path. Ten minutes of churn: 11 workers deleting and recreating their instance in a loop plus one worker flipping an instance between shaped and unshaped via its spec — **184 pod-up + 195 pod-down events + 30 shaped/unshaped flips**.

- **Kernel memory:** no leak — `SUnreclaim` 2,087,156 kB before → flat during → 2,081,380 kB after (Δ −5.8 MB); `Slab` likewise flat.
- **Classid allocation:** all 8 accounts held their exact classids through ~184 reallocation bursts, zero duplicates — the uncached-read allocation fix exercised live.
- **Agent:** every tripwire at 0 after the run (`skipped_pods`, `direct_packets`, `return_dropped`, `sibling_overflow`, `reconcile_errors`, `return_attach_failures`); `attached_pods` = 8 matching the shaped pods; `link_reattach_total` = 124, matching shaped pod creations + flips — one attach each, no flapping; 116 transition-log lines, zero error lines. Controller errors were only optimistic-concurrency conflicts (normal retry under churn).
- **Datapath audit:** all 8 shaped pod devices carried `kura_shaper_out` first in the tcx chain (ahead of `cil_from_container`) with a matching pin; all 4 unshaped devices had no program and no pin; no orphan pins; the tree held exactly the 8 expected classes (1 Mbit floor-clamp, 500 Mbit ceil) under the root; mid-churn samples always showed attached = pins = programs.

## Rollout controls and observability

- `egressTreeAgent.enabled` — ships **disabled**.
- `egressTreeAgent.betaPodPrefix` (`BETA_POD_PREFIX`) — when set, the agent attaches only to pods whose name starts with the prefix, so the first enforcement step can target one account before the fleet. Excluded pods stay unshaped and count in `kura_egress_tree_beta_excluded_pods` (separate from the alerting `skipped_pods`). Sibling allowlists are still computed over all annotated pods, and prefix changes converge within one cycle in both directions.
- A `CiliumClusterwideNetworkPolicy` ships with the DaemonSet (same `nodePools` scope), dropping the metrics port from `world` — the listener binds the box's public IP (hostNetwork bare metal, no host firewall). The `ingressDeny` is paired with a baseline allow-all ingress rule: any host policy selecting a node, deny rules included, switches the host endpoint into ingress enforcement, so a deny-only policy would default-deny the node (verified on staging). In-cluster scrapes pass as cluster identities.
- Prometheus metrics on `:9469`: per-class bytes/drops/backlog, direct-queue packets (any growth means a broken stamp or foreign redirect), per-pod redirect/guard/bypass counters, return drops, reattach churn, return-attach failures, skipped/excluded pods. State transitions (class added/updated, pod attached/updated/removed, budget changes) log at info level; quiet converged cycles log nothing.

## Rollout

1. Controller + annotation land first — inert until an agent consumes them. One deploy-impact note: the annotation lands in the StatefulSet pod template, so the controller half rolls every shaped instance once when it ships.
2. Agent to ca-east in observe mode (`defaultNodeEgressMbps` ≈ line rate, floors informational), optionally gated to one account via `betaPodPrefix`; validate per-tenant counters against expected traffic.
3. Real ceilings and floors on ca-east; acceptance at the physical interface; then the other bare-metal regions.
4. Alert rules for the exported tripwires (direct-queue, return-drops, return-attach failures, reattach churn, skipped pods, floor violations) ship before enforcement is enabled.
5. Precondition for floors: fix the per-replica floor double-count in scheduler bin-packing (separate change).


infra/helm/tuist/values.yaml
```diff
@@ -251,6 +251,53 @@ kuraController:
...
+  # other annotated pod stays on the unshaped Cilium-only path and is counted
+  # in kura_egress_tree_beta_excluded_pods. Empty = shape all annotated pods.
+  betaPodPrefix: ""
+  # Observe-mode fallback for the tree's root ceiling on nodes that do not
+  # advertise tuist.dev/egress-mbps capacity. 0 = no fallback: such nodes get
+  # no tree and their pods stay unshaped on the node-local leg.
+  defaultNodeEgressMbps: 0
+  # Schedule only onto kura cache pools; same shape as the regional
+  # kura-*-ingress-nginx blocks in infra/helm/platform/values-tuist.yaml.
+  nodeSelector: {}
```
pepicrft, line 305: **[Critical] Restrict the DaemonSet to the Kura cache pools**

`nodeSelector` defaults to `{}`, so the template emits no selector. The toleration only permits placement on a tainted Kura node; it does not require it. With the agent enabled, Kubernetes can therefore run this privileged, host-networked DaemonSet on every eligible non-Kura node as well. That expands the host-level traffic-control change beyond the intended Kura machines.

Set an explicit selector for the Kura pools in managed values, or reject an empty selector when the agent is enabled.

.github/workflows/egress-tree-agent-image.yml
```diff
@@ -0,0 +1,85 @@
...
+# workflow only handles pre-release iteration via `:sha-<git-sha>` tags.
+
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - infra/egress-tree-agent/**
+      - .github/workflows/egress-tree-agent-image.yml
+  pull_request:
```
pepicrft, line 17: **[High] Run the test job when a draft becomes ready for review**

This trigger uses GitHub default pull-request activity types, which exclude `ready_for_review`. If the pull request is opened as a draft, the guard on line 33 skips the test. Marking it ready does not rerun this workflow, and this pull request agent test and build are currently skipped. The kernel-level traffic-shaping program can therefore merge without its generation and test validation.

Add `ready_for_review` to the trigger types, or run the test job for draft pull requests as well.

pepicrft (approved):

infra/kura-controller/controllers/kurainstance_controller.go
```diff
@@ -2725,6 +2731,182 @@ func sharedSecretsEnvFrom() []corev1.EnvFromSource {
...
+// duplicate rule (smallest account handle keeps a doubly-claimed id, smallest
+// minor wins within an account) still makes any conceivable duplicate
+// self-heal instead of flapping.
+func (r *KuraInstanceReconciler) reconcileEgressClassID(ctx context.Context, instance *kurav1alpha1.KuraInstance) error {
+	if !instanceNeedsEgressClass(instance) {
+		return nil
+	}
+
+	instances := &kurav1alpha1.KuraInstanceList{}
+	if err := r.List(ctx, instances, client.InNamespace(instance.Namespace)); err != nil {
```
fortmarek: **Stale-cache read can hand two accounts the same minor.**

`Client` here is `mgr.GetClient()` (`cmd/manager/main.go`), so this `List` reads the informer cache. The comment above justifies safety with serial reconciles, but serialization does not close this gap: the cache updates asynchronously after `r.Update`, so reconcile(B) can start before A's freshly written `kura.tuist.dev/egress-class-id` is visible. `used` then misses A's claim and B can probe onto it. Most likely during the initial bulk allocation, when every instance is reconciled back to back.

While duplicated:

- both accounts share one HTB class on any node they co-inhabit;
- worse, the agent's `Desired` groups siblings by minor alone (`infra/egress-tree-agent/internal/agent/desired.go`), so the two tenants become each other's *siblings* and traffic between them takes the unshaped bypass entirely;
- the self-heal then renumbers the loser, which breaks this function's own invariant ("The id must never change for a live account") and rolls that account's StatefulSet.

`PeerDemuxReconciler` in the same binary already takes `mgr.GetAPIReader()` for exactly this class of problem; using it for the allocation scan is the minimal fix.

Cheap defense in depth on the agent side: carry the account handle in the `tuist.dev/egress-class` payload and have `Desired` group by (minor, account), logging a conflict instead of silently merging two tenants into one class and one sibling set.

infra/egress-tree-agent/internal/agent/agent.go
```diff
@@ -0,0 +1,378 @@
...
+			skipped++
+			requeue = true
+			a.Log.Warn("no cilium endpoint for pod; leaving it unshaped",
+				"pod", attachment.Namespace+"/"+attachment.Name)
+			continue
+		}
+		reattached, err := a.Attacher.EnsurePod(device, trampoline.Index, attachment)
+		if err != nil {
+			skipped++
+			a.Log.Error("attaching pod device failed",
```
fortmarek, line 194: **A transient `EnsurePod` error strips working enforcement.**

This path `continue`s without setting `active[device]`, so `CleanupStale(active)` a few lines below unpins and detaches that device. A pod that is currently attached and healthy ends up fully unshaped because of a hiccup in a *config sync* — `linkIsFirst`'s `link.QueryPrograms` failing, a `LoadPinnedMap` error, or a `siblings.Put`.

The last one is reachable by design: the `siblings` map is `max_entries: 16` (`bpf/redirect.c`) and `syncPodMaps` returns the raw `Put` error, so an account with more than 16 co-located pods flaps attach/detach every cycle — and partially populates the map first, since Go map iteration order decides which siblings land before `E2BIG`.

Recording the device in `active` even on error (or tracking "known attached" separately from "converged this cycle") keeps the last known-good program in place, which fits the module's fail-safe philosophy better than tearing it down. The sibling cap separately deserves an explicit truncate-and-warn rather than an error path.

infra/egress-tree-agent/internal/agent/tree.go
```diff
@@ -0,0 +1,259 @@
...
+			"quantum", "60000"); err != nil {
+			return fmt.Errorf("class %s: %w", ClassIDString(minor), err)
+		}
+		if _, err := run(ctx, "tc", "qdisc", "replace", "dev", t.TrampolineDev,
+			"parent", ClassIDString(minor), "fq_codel"); err != nil {
+			return fmt.Errorf("leaf qdisc for %s: %w", ClassIDString(minor), err)
+		}
+	}
+
+	for _, class := range existing {
```
fortmarek, line 136: **Stale-class deletion runs before the programs that stamp them are detached, so a normal operation trips the `direct_packets` alarm.**

Order in `Agent.reconcile` is `EnsureTree` (this deletion) -> attach loop -> `CleanupStale`. When a live pod's annotation is removed (validation scenario 3.7), its tcx program keeps stamping the now-deleted classid for the rest of the cycle; those packets land in HTB's direct queue and bump `kura_egress_tree_direct_packets` — the counter `AGENTS.md` says "must alert" and describes as meaning "a foreign redirect or a broken stamp".

Detaching pods before deleting their classes removes the false positive.

Related, and not a defect: a hand-cleared or otherwise lost qdisc/class produces silent, fully-unshaped egress until the 2 m backstop, since there is no watch on qdisc state the way there is on link state. `direct_packets` does catch it, but that 2 m worst case is worth stating in the alert runbook, or closing by extending the netlink watch to `RTNLGRP_TC`.

fortmarek (commented):
Three findings from a read-through of the datapath, the reconcile loop, and the controller half. The trampoline design itself holds up — I traced the packet through `sch_handle_ingress`/`tcx_run` (mac header pushed, so the `data+12`/`data+30` offsets are right), `htb_classify` matching `skb->priority` against a level-0 class before consulting filters, the same-netns `skb_scrub_packet(xnet=false)` that preserves the stamp, and `dev_forward_skb` on return clearing mark/priority but not `tc_index`. These are about the control plane around it.

fortmarek:
One gap in the validation matrix worth closing before ca-east: it is thorough on bandwidth and silent on packet rate.

Every load scenario in section 4 is a byte-rate test. 4.4 measures 18.6 Gbit/s at a 20 Gbit root with ~5% CPU on 24 cores; 4.2 measures probe latency under class saturation. GSO survives the detour (veth carries `NETIF_F_GSO_SOFTWARE` and the trampoline is MTU 65535), which is good news for cost, but it also means that at 18.6 Gbit/s the tree is only handling roughly 35k skb/s of 64 KB super-packets. The same bandwidth in 1.5 KB packets is closer to 1.5M skb/s, and two costs in this design scale with skb/s rather than with bytes:

**1. Per-CPU backlog pressure triples.** Each shaped packet takes two extra trips through `enqueue_to_backlog`: once when `kura-egress1` receives from the veth, once when `dev_forward_skb` re-injects into the `lxc`. `netdev_max_backlog` defaults to 1000 and overflow is a silent drop counted only in `/proc/net/softnet_stat`. Nothing in the exported metric set would show it, so this failure mode is currently invisible: it would present as packet loss with every `kura_egress_tree_*` counter healthy, including the direct-queue and return-drop tripwires.

**2. `sch_htb` serializes on one root lock.** HTB is not `TCQ_F_NOLOCK`, so every shaped pod on a node now dequeues through a single spinlock, replacing a path that ran parallel across TX queues (`redirect_peer` / NIC multiqueue). Lock contention is a cliff rather than a slope, and it is a per-node property, so the 3-tenant setup in 4.1 does not probe it.

Neither is a design objection. The fixed per-bounce cost looks like single-digit microseconds per skb, which is noise against a bulk transfer and still small against a metadata read once the RocksDB lookup and the RTT are counted. The concern is narrower: the regime where these two go non-linear has no coverage at all, and it is a plausible kura shape, since CAS digest lookups and `FindMissingBlobs` are high-rate small-request patterns.

Three things would close it:

- A small-packet PPS ceiling, shaped versus fully detached, with `/proc/net/softnet_stat` drops watched on both sides.
- A concurrency run with many shaped pods active, to probe root-lock contention rather than the 3 tenants of 4.1.
- Export softnet backlog drops alongside the existing tripwires, or at minimum document the check in the alert set, since none of the current counters can see this one.

All three are cheaper to find in the lab than in observe mode.

fortmarek (approved):
Looks good!

One thing to worry about are the perf implications under load. I'd double check the enforcement of the floor/ceiling doesn't negatively impact that scenario. Worth testing before we ship this.

infra/egress-tree-agent/cmd/agent/main.go
```diff
@@ -0,0 +1,358 @@
...
+		map[string]bool{trampolineDev: true, returnDev: true}, kick)
+
+	registry := prometheus.NewRegistry()
+	registry.MustRegister(collectors.NewGoCollector())
+
+	a := &agent.Agent{
+		NodeName:          nodeName,
+		DefaultNodeMbps:   defaultNodeMbps,
+		BetaPodPrefix:     os.Getenv("BETA_POD_PREFIX"),
+		ReturnDetachAfter: int(returnDetachAfter),
```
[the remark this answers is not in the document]
esnunes: Fixed in 6460142a59: `RETURN_ATTACH_MAX_FAILURES` now parses via `strconv.Atoi`, which returns `int` natively, so the flagged `int64`→`int` conversion no longer exists. The 1..1000 bounds check is unchanged (the value was already guarded before the conversion; the scanner could not connect the guard to the conversion site).

infra/helm/tuist/values.yaml
esnunes: Fixed in ef2ee9464c. One wrinkle: a plain `nodeSelector` is an equality match and a cluster runs several kura pools with different `node.cluster.x-k8s.io/pool` values (kura-dedibox, kura-us-east, ...), so one selector cannot cover them. Scoping is instead a new `egressTreeAgent.nodePools` list rendered as required node affinity (`pool In [<pools>]`), with `kubernetes.io/os: linux` always applied (keeps the linux-only image off the darwin Mac mini nodes) and `nodeSelector` kept for extra constraints. The template now fails to render an enabled agent when both `nodePools` and `nodeSelector` are empty — same pattern as the existing image guard, so an unscoped deployment is impossible.

.github/workflows/egress-tree-agent-image.yml
esnunes: Fixed in 46fc5501fa: added the explicit `types` list including `ready_for_review`, matching the convention of the other draft-skipping workflows (kura.yml, cache.yml, cli.yml, ...). Follow-up fbadc275c4 also splits build from push: the multi-arch image now builds on every PR run so a broken Dockerfile cannot surface only after merge, while the ghcr login and push stay gated on non-`pull_request` events — fork PRs can only produce `pull_request` events, so they never reach the registry.

infra/kura-controller/controllers/kurainstance_controller.go
esnunes: Fixed in eb9dd623a1: the allocation scan now Lists through `mgr.GetAPIReader()` (the `PeerDemuxReconciler` precedent), so a claim written by an earlier serial reconcile can never be missed by a lagging informer cache — the quorum read closes the race entirely. The misleading "serial reconciles" justification is rewritten to describe the actual hazard.

We deliberately skipped the (minor, account) defense-in-depth on the agent side: the two damaging consequences are the classid renumbering and two tenants sharing one HTB class, and the agent-side grouping can prevent neither (renumbering happens at the controller; against a shared class it would only demote "shared" to "loser unshaped"). The residual producer assumptions (single leader, no hand-edited annotations, one namespace per cluster) are the same ones the rest of the controller already stands on.

infra/egress-tree-agent/internal/agent/agent.go
esnunes: Fixed in f6ff0cc443, both halves:

- The `EnsurePod` error path now records the device as active, so `CleanupStale` leaves the last known-good program attached, and it requeues the fast retry it was also missing (a transient error previously waited out the backstop).
- `syncPodMaps` no longer errors on a full `siblings` map: it stops adding entries once `MaxEntries()` is reached. The polarity makes that safe — a sibling missing from the map is shaped, never wrongly unshaped — so overflow degrades gracefully instead of flapping attach/detach. Surfaced via the new `kura_egress_tree_sibling_overflow_total` counter rather than a log (a stuck overflow would log every cycle); the cap stays 16 since real co-located sibling counts are 0–3 today.

infra/egress-tree-agent/internal/agent/tree.go
esnunes: Fixed in 31f02d6c52: stale-class deletion moved out of `EnsureTree` into `Tree.PruneClasses`, called after `CleanupStale`, so both orderings now hold — classes exist before programs stamp them (adds), and programs detach before their class is deleted (removes). A class outliving its pod by one errored cycle is harmless; the reverse was the false `direct_packets` alarm.

On the second paragraph: agreed it is not a defect, and we took the runbook route — AGENTS.md now documents that a hand-cleared root qdisc runs unshaped until the backstop interval rebuilds the tree, with `direct_packets` growth as the signal. Extending the netlink watch to `RTNLGRP_TC` did not seem worth the machinery for a hand-tampering scenario the metric already surfaces.

esnunes:
Closed the packet-rate gap on the k01 lab — 1.4 KB UDP datagrams (worst case: the eBPF programs are protocol-blind, and real TCP GSO-coalesces into far fewer skb/s at the same byte rate), 5-minute sustained runs, shaped vs fully detached, 1 pod and 50 pods (50 tenant classes), with CPU, slab, and `/proc/net/softnet_stat` monitored throughout. Full table is in the PR description ("Packet-rate runs"); the short version:

- **Backlog pressure:** zero softnet drops in all four runs, including 1.43 Mpps sustained across 50 shaped pods (~1.7B packets processed). `netdev_max_backlog` never overflowed at the maximum rate this hardware can generate, so no sysctl tuning shipped.
- **Root lock:** it is real but manifests as sender backpressure, not a cliff — 50 unshaped senders attempted 3.31 Mpps and delivered ~1.16 Mpps at 94.8% CPU (sink sockets dropped the rest), while shaped senders were paced to 1.43 Mpps and delivered ~1.41 Mpps at 56.7% CPU. Delivered throughput under overload was *higher* shaped than unshaped. Single-flow PPS ceiling costs ~22% (811 → 630 kpps).
- **Observability:** agreed the blind spot was real — no `kura_egress_tree_*` counter can see backlog overflow. Rather than duplicating a node-global counter in the agent, the alert set (AGENTS.md, 9a6cdc8b06) now includes `rate(node_softnet_dropped_total)` on shaped nodes; node-exporter's softnet collector already runs fleet-wide.

PR #12530, opened 2026-08-20
fix(infra): serialise registry pushes again until uploads are reliable

fortmarek:
## What changed

Removes `concurrency: "4"` from every registry-bound `tart-push`, returning them to serialised uploads. `chunk-size: "64"` stays everywhere, including the two release-path calls that had been silently falling back to `3`.

## Why

[#12479](https://github.com/tuist/tuist/pull/12479) set `concurrency: "4"` and did exactly what it claimed, taking throughput from 3.2 MB/s to 13.1 MB/s. It also broke pushes of new large images, **including on the release path**.

The registry loses in-progress blob upload sessions under that parallelism. Clients see an opaque 500; the registry logs the real cause:

```
error resolving upload: s3aws: Path not found:
  /docker/registry/v2/repositories/<repo>/_uploads/<uuid>/data
```

Observed against both `tuist-runner` and `tuist-xcresult-processor`, so it is not repo-specific and it does affect `server-production-deployment.yml`. `tart` retried through all 8 attempts and the step timed out. The same underlying failure is what makes `crane copy` unusable against this registry, where it surfaces as `RANGE_INVALID` instead.

## Measurements

Same host, same image, both after the sidecar CPU bump was live:

| concurrency | throughput | uploads |
|---|---|---|
| 4 | 13.1 MB/s | sessions lost within ~9 min |
| 1 | 3.2 MB/s | clean |

## What is not implicated

**The sidecar CPU bump stays.** Serialised throughput measures 3.2 MB/s both before and after it, with the sidecar sitting at 396m of its 2000m limit. Its headroom only ever paid off through parallelism, so it costs nothing to leave in place and it is the right shape once uploads are reliable.

**The chunk size stays.** That is the change that actually cut request volume, roughly 1,000 requests for a 50 GB image instead of 17,000, and it is not what breaks uploads: the base image seed pushed all five tags at chunk size 64 with uploads serialised.

## Honest framing

This is a mitigation, not a fix, and it is undoing a regression I introduced. #12479 changed three things at once on the release path (chunk size, concurrency, and the server's CPU ceiling), which is why it took a controlled run at concurrency 1 to work out which one was responsible.

The registry should tolerate concurrent blob uploads. Until it does, this ceiling stays where it is. An intermediate concurrency of 2 was not tested and might buy some of the throughput back safely.

### DIFF
--- a/.github/workflows/macos-xcode-image.yml
+++ b/.github/workflows/macos-xcode-image.yml
@@ -208,7 +208,13 @@ jobs:
           local-name: macos-tahoe-xcode
           remote-names: ${{ vars.OCI_REGISTRY_HOST }}/macos-tahoe-xcode:${{ steps.build.outputs.tag }}
           chunk-size: "64"
-          concurrency: "4"
+          # Deliberately serialised. Four uploads in flight measured
+          # 13.1 MB/s against 3.2 MB/s here, but the registry loses blob
+          # upload sessions under that parallelism and fails the push
+          # with a 500, logging `error resolving upload: s3aws: Path not
+          # found .../_uploads/<uuid>/data`. Reliability wins until that
+          # is fixed; see the chunk size above, which is the change that
+          # actually cut request volume.
 
       - name: Cleanup
         if: always()
--- a/.github/workflows/runner-image.yml
+++ b/.github/workflows/runner-image.yml
@@ -153,7 +153,13 @@ jobs:
           local-name: tuist-runner
           remote-names: ${{ vars.OCI_REGISTRY_HOST }}/tuist-runner:${{ steps.build.outputs.profile }}-${{ steps.image-tag.outputs.value }}
           chunk-size: "64"
-          concurrency: "4"
+          # Deliberately serialised. Four uploads in flight measured
+          # 13.1 MB/s against 3.2 MB/s here, but the registry loses blob
+          # upload sessions under that parallelism and fails the push
+          # with a 500, logging `error resolving upload: s3aws: Path not
+          # found .../_uploads/<uuid>/data`. Reliability wins until that
+          # is fixed; see the chunk size above, which is the change that
+          # actually cut request volume.
 
       - name: Cleanup
         if: always()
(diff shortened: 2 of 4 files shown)


fortmarek:
## Serialising is necessary but not sufficient

Digging further while this ran, the push half of the registry migration turns out never to have worked at all, and the timeouts on these steps are calibrated for a different registry.

The last green `runner-image` build (2026-08-12, run 31637473933) has a step literally named **"Push image to GHCR"**, with `TART_REGISTRY_HOSTNAME: ghcr.io`. It took **50 minutes**, comfortably inside the `timeout-minutes: 90` that #12093 set for it. The switch to `${{ vars.OCI_REGISTRY_HOST }}` came later, in #12413.

So the 90 minute budget was sized against GHCR, and against this registry it does not fit:

| target | mode | throughput | ~55 GB image |
|---|---|---|---|
| GHCR | chunk 3, serialised | ~18 MB/s effective | ~50 min (measured) |
| Tuist registry | chunk 64, serialised | 3.2 MB/s | ~4.8 h (extrapolated) |
| Tuist registry | chunk 64, concurrency 4 | 13.1 MB/s | ~70 min, but uploads fail |

The gap is per-stream latency, not bandwidth: each 64 MB chunk costs roughly 20 seconds against this registry, which is why four streams multiply almost perfectly and one stream is slow no matter how much CPU the sidecar gets.

**What this PR does and does not do.** Serialising stops the 500s, so pushes stop failing outright. It does not make them complete: at 3.2 MB/s they will now hit the step timeout instead. This is still the right thing to merge, because a timeout with a clean registry log is a better failure than corrupted upload sessions on the release path, but nobody should read it as restoring image publishing.

**The real fix is registry-side.** Either the registry tolerates concurrent blob uploads, or per-chunk latency comes down. Raising `timeout-minutes` to fit a ~5 hour serialised push would technically work for a single dispatch, but the release matrix runs five profiles across two builders with `fail-fast: true`, so that is not a serious option.

Worth testing before anything larger: `concurrency: "2"`. If the session loss needs more than two streams to trigger, that alone might land a push inside the existing budget.

fortmarek:
Superseded by #12535, which moves image publishing back to GHCR entirely rather than tuning how it pushes to the Tuist registry.

The measurements here still stand and are worth keeping for whoever picks the registry back up: concurrency is what loses upload sessions, regardless of chunked or monolithic mode, and serialised monolithic (6.2 MB/s) is twice serialised chunked (3.2 MB/s). Neither is close to GHCR's ~50 minutes for the same image.

PR #12531, opened 2026-08-20
feat(server,cache,cli): enforce the Air free tier across the live cache lanes

fortmarek:

The Air plan's cap of 200 billable runs per month was defined in `Tuist.Billing` and shown in the dashboard's estimated payment, but nothing enforced it on any live request path. This makes it real, tells a blocked account why it was refused, and gives every account on the plan a full allowance on the day it ships.

12 accounts are currently over the cap, about 2k USD per month of list-price usage. 5 of them are lapsed customers whose subscriptions were canceled or fell past due, which drops them to Air.

## Why it was unenforced

`BillingPlug` was the only place that checked the threshold, and it is mounted solely on the legacy cache controller actions, none of which the current CLI calls. Cache traffic runs over two lanes that never consulted the plan: Kura and CAS, which authorize from grants carried in a token, and the module cache, served by the `cache/` app. An account on an old CLI was gated while the same account on a current CLI was not, so enforcement was effectively a function of client version.

## How the limit is enforced

One predicate, `Billing.cache_access_blocked?/1`, is true when the account's effective plan is Air and its monthly billable-run counter has reached 200. A lapsed or past-due subscription resolves to Air, so it is gated on the same terms as an account that never subscribed.

It is applied inside `Tuist.Cache`'s grant resolution, in `accessible_accounts/1` and `accessible_projects/2`. Blocking means the account's handles disappear from the grants. Those two functions feed everything a cache node can ever learn about what a subject may reach:

| Carrier | Reaches | Lifetime |
|---|---|---|
| Cache token from `/api/cache/token` | Kura, Kamal nodes | 30 min |
| `cache_grants` embedded in ordinary auth tokens | Kura, Kamal nodes | 10 min |
| `/api/cache/access` introspection, for opaque CI tokens | Kura, Kamal nodes | cached at most 10 min |

A node then runs its unchanged membership check, does not find the handle, and denies. No node evaluates a plan or holds any billing knowledge. That is why a single filter covers both node types and every credential shape, and why a client that never calls the token endpoint is still blocked.

Two timing properties follow from this. Propagation is bounded at about 30 minutes, typically 10, by whichever carrier the client is holding. And the gate reads `accounts.current_month_remote_cache_hits_count`, which a daily worker recomputes from ClickHouse, so crossing the threshold mid-day does not bite until that recompute. Enforcement lags real usage by up to a day by design.

The legacy `/api/cache/*` endpoints remain gated by `BillingPlug` for older CLIs.

## How a refusal reaches the user

Expressing a block as absence from the grants is indistinguishable from never having had access, so nodes would have answered 403 and sent the caller looking for a permissions problem they do not have. The server now names the accounts it withheld for payment alongside the grants themselves, as `payment_required` on the `/api/cache/access` response and as a `cache_payment_required` claim on both token carriers. It is diagnostic rather than a grant, so it survives project scoping.

| Where the request fails | What comes back |
|---|---|
| Kamal cache node | 402 naming the account and the upgrade, on both its token and introspection paths |
| Kura node | 402 with the same message, on both of its authorization paths |
| CLI cache operations | 12 operations read the 402 and raise an error carrying the server's message |
| `/api/cache/token` | 402 with an upgrade link; the CLI reports it once and stops falling back to the raw credential |
| Dashboard | Usage and estimated payment, unchanged |

## The reset

Zeroing the counter alone would not have held, because it is recomputed daily from the start of the month, so a direct write is undone within a day. `accounts.free_tier_reset_at` moves the start of the counting window instead, and the backfill sets both it and the counter for every account on the Air plan, verified against production as 8,971 accounts and no paid account. Nobody is blocked on day one for usage accrued while the cap was unenforced. The timestamp goes inert on its own at the next month boundary, with no cleanup job.

## Also in here

`subscription_active?` compared status against `"active"` alone while `get_current_active_subscription/1` matches `"active"` and `"trialing"`, so any customer mid-trial would have been answered 402 the moment enforcement went live. Fixed and pinned by a test. Production currently has no trialing subscriptions, so it was dormant.

Blocked accounts are resolved in one query rather than one per account, so the cache authorization paths do not scale a subscription query with the number of accounts a subject can reach. A test pins the query count flat.

`/api/cache/endpoints` deliberately does not enforce. It answers where an account's cache is, not whether the account may use it, clients hold that answer for up to an hour, and its account parameter is optional. A plan that lapses inside that window would never be reported through it.

## Reviewer notes

Regenerating the two OpenAPI specs surfaced pre-existing drift that had to be reconciled for the CLI to compile at all. `cache.yml` had never been regenerated after `downloadCASArtifact` and `saveCASArtifact` were renamed to `downloadXcodeArtifact` and `saveXcodeArtifact`, and after several operations moved from 400 to 422. The server spec had drift in the test run schema which reordered a generated initializer. None of this is caused by the billing work, but the diff carries it.

`Access::PaymentRequired` in Kura sits above `Refused` and below `Read`, deliberately: above `Refused` so the introspection floor still only ever rises, which `via_cache_access` relies on, and below `Read` so it grants no action. A test pins both bounds, since that ordering is the authorization check itself.

## How to test locally

Server, cache app and Kura:

```
cd server && mix test test/tuist/cache_test.exs test/tuist/billing_test.exs test/tuist/accounts_test.exs test/tuist_web/controllers/api/cache_controller_test.exs test/tuist_web/plugs/api/authorization/billing_plug_test.exs
```

```
cd cache && mix test
```

```
cd kura && cargo test --lib auth::
```

505 server tests, 547 cache tests and 96 Kura auth tests pass.

The CLI compiles with the flag combination that avoids the resolver crash currently blocking `tuist install`:

```
swift build --target TuistCache --replace-scm-with-registry --force-resolved-versions
```

`TuistCacheTests` is not declared in `Package.swift`, only in the Tuist module graph, so the test added for `CacheTokenStore` still needs `tuist generate` plus `xcodebuild` to run.


pepicrft (approved):
Follow-up: cache tokens issued without `full_handle` carry `cache_payment_required`, but Kura treats that level as inconclusive and falls back to introspection. Introspection drops the claim, and the following cache-access request presents the cache token as an application credential, yielding 401 rather than the intended 402. That prevents the compilation-cache plugin from surfacing its upgrade guidance.

Please preserve the plan-exhaustion claim through cache-token introspection, including tenant-scoped introspection, and treat it as a terminal refusal when a node verifies the token locally. Add an end-to-end test for an unscoped cache token.

PR #12533, opened 2026-08-20
feat(server): add a per-account Kura disk claim override and apply it to running instances

fortmarek:
Builds on #12516. **Base branch is `kura-plan-sized-disk-reservation`**, so this diff is the work on top; merge #12516 first.

> Adds a per-account override of the plan-derived Kura disk claim, and the path that applies a changed claim to an instance that is already running.

### Why

Sizing an instance from its account's plan is the right default and a weak predictor of any individual account. Within a single plan the spread in what an account's cache actually holds reaches two orders of magnitude in both directions:

- Within Enterprise the spread is ~300x. Six of ten provisioned account-regions hold under 2.4 GiB of their 40 GiB ring, while the heaviest writes 23 GiB/day and gets about 1.7 days of retention before the ring evicts.
- Within Air the spread is ~128x (measured from the legacy CAS lane, since no Air account has a Kura instance yet). The median moves 0.17 GiB/day; the heaviest moves 21.75 GiB/day and would get roughly 1.9 days even on the Enterprise ring.

So there are accounts on both the cheapest and the most expensive plan that no choice of plan constant serves. Before this, a value written into `kura_servers.storage_claim_size` by hand reverted to the plan default the next time an instance's volumes were built, which is the worst possible timing: it reverted exactly when the volumes were rebuilt.

### The override

New `kura_storage_claims` table (one row per account, unique on `account_id`) with `Tuist.Kura.StorageClaim` / `Tuist.Kura.StorageClaims`, modelled on `runner_concurrency_limits` / `Tuist.Runners.Concurrency`. `storage_claim/2` resolves the override ahead of `AccountPolicies.sizing_plan/1`, so all three paths that build an instance's volumes pick it up: `create_server/1`, the cold return, and the warm-handoff target.

Overrides are bounded below by `Regions.minimum_storage_claim/0`, which returns air's claim rather than a number of its own, so there is one floor and it follows the ladder. What stops that floor going lower is fixed overhead rather than cache need: staging and one rotation segment come out of a claim before the ring is sized, and Kura clamps the ring up to five segments, so below the floor `cas_capacity_bytes/1` emits no `KURA_CAS_CAPACITY_BYTES` at all and the runtime sizes its ring from the whole box. The budget-cliff test asserts that against the derivation, so the bound holds if the reserves move again.

### Applying a changed claim to a live instance

Two gaps, not one.

**Server side.** The claim that renders into the manifest is the one carried on the row, so writing only the account-level override would leave every running instance serving the size it was built at. The write re-pins the account's rows that still hold volumes, in storage-governed regions only, comparing against the claim the account resolved to *before* the write, so a row that pins nothing is counted rather than silently skipped.

**Cluster side.** The controller responded to a grown `spec.storageSize` by patching live PVCs up, which the API server rejects: the managed regions run `scw-local-nvme` (`allowVolumeExpansion: false`) and `volumeClaimTemplates` are immutable. That patch is removed, and `staleDataStorageReason` now reports size drift alongside storage-class drift and node-orphaned volumes, so `reconcileStaleDataStorage` drops the PVCs via the `whenDeleted: Delete` retention policy and fresh ones bind at the declared size.

**The trigger is one-directional.** A claim larger than the bound volume recreates, because the ring would otherwise be budgeted past the end of the disk. A claim smaller than the volume is left alone: the volume holds the ring with room to spare and evicts down into it. That direction matters because instances carrying no claim of their own resolve one from their plan at render time, so a plan downgrade moves their rendered claim downward, and recreating there would spend an account's whole cache to reclaim disk the next rebuild reclaims anyway.

### What the operator is told

Raising and lowering have different consequences, so the write returns `:raised` and `:lowered` and the ops flash says which happened rather than reporting a successful save.

`:raised` is an upper bound, deliberately. The controller rebuilds a volume only when the claim exceeds what that volume was built at, and the control plane only knows what the row was previously pinned to. Those agree until an earlier decrease leaves a volume oversized — `50Gi` down to `20Gi` and back to `40Gi` is a raise against the pin while the volume is still `50Gi` — so the flash reads "up to N". The bound errs the safe way: a volume is never smaller than the row's pin, so nothing lands in `:lowered` that the cluster then rebuilds. Naming volumes exactly would mean recording what each was built at, which the control plane does not observe today; that is the follow-up if exact reporting is wanted.

### Concurrency

Resolving the claim and writing the row it goes on happen in one step, inside the transaction, behind a shared lock on the account; the override write takes the exclusive side before it reads anything. Without that, a provision that read the claim before an override committed would insert the stale value onto a row the override's re-pin could not reach, and a pinned claim wins from then on, so nothing would converge it. Both sides take the account before touching `kura_servers`, so they queue rather than deadlock.

### Constraints held

No PVC resize is attempted, no `ephemeral-storage` limit is added, and `KuraInstance.Spec.StorageSize` remains the single value the server sets and everything downstream derives from.

### Reviewer notes

- The size-drift trigger fires on drift from any source, not only an operator override. Since the base stopped backfilling the claim, existing instances carry none and resolve from their plan, so the one-directional trigger is what keeps a downgrade from dropping their caches.
- The provisioner's fallback for an unpinned instance in a governed region resolves the plan's claim rather than the account's effective claim, deliberately: reading an override means reading a table on a path that renders every reconcile tick and is resolvable from an in-memory account (`kubernetes_controller_test` is `ExUnit.Case` with no sandbox for that reason). Every path that builds volumes pins a claim, and setting an override pins the rows that carry none, so it is not a reachable state.
- No deterministic test covers the account lock: the sandbox puts both processes on one connection, so the transactions cannot interleave in-process and such a test would pass without exercising it.
- Two controller tests pinned the removed expansion behaviour and were rewritten rather than deleted.
- `server/data-export.md` documents the new table.
- `server/mix.lock` carries an unrelated Phoenix bump in the local worktree; it is deliberately not part of this PR.

### How to test locally

```bash
cd server && mix test test/tuist/kura test/tuist/kura_test.exs test/tuist_web/live/ops_account_live_test.exs
```

```bash
cd server && mix format --check-formatted && mix credo && mix excellent_migrations.check_safety
```

```bash
cd infra/kura-controller && go build ./... && go vet ./... && go test ./...
```

To exercise the override by hand: run `mise run dev`, sign in as an operator, open `/ops/accounts/<id>`, and use the "Disk claim override" field in the Kura card. The Claim column shows what each instance actually holds (its carried claim, or what it resolves to when it carries none). With a running instance in a storage-governed region (`us-east`, `us-west`, `eu-central`, `ca-east`), raising the claim reports how many instances may rebuild, lowering it reports that caches are kept, and a value under the floor is rejected inline.

Validation for this change: 486 tests across the Kura suites and the ops account view, `mix format`, `mix credo` (no issues), `mix excellent_migrations.check_safety` (no findings in the new migration), and `go build` / `go vet` / `go test` for kura-controller.


fortmarek:
This is how the override looks like in the ops page:
image

pepicrft (approved):

PR #12551, opened 2026-08-21
feat(kura): support HTTP Range on artifact reads so downloads resume

fortmarek:
### What changed

Kura's artifact serving path had no HTTP Range support, so every aborted or timed-out download restarted from byte 0. Both serving planes now resolve a single `bytes=` range from the request and answer it with `206 Partial Content` plus `Content-Range`, and the CLI resumes a download that dies mid-body instead of fetching it again from the start.

**Server (`kura/`)**

- New shared `artifact::range` module resolves a request's `Range` header against the stored size. It is shared deliberately: the sendfile/splice accelerator and the Axum path that serves everything else (TLS, HTTP/2, non-Linux, mapped bodies) must not drift on what a range means, or a resume would behave differently depending on which plane answered.
- `accelerated_file_serving.rs` writes the 206 status line and `Content-Range`, and offsets the `sendfile`/`splice` start by the range.
- `http.rs` does the same on the Axum path, slicing the mmap view or opening a ranged reader.
- `Accept-Ranges: bytes` is advertised on full 200 responses too, not just partial ones. A client only knows resume is on offer if the server said so *before* the transfer died.
- `ensure_complete_transfer` now compares against the range's length rather than the artifact's size, so a satisfied partial response is complete at its own last byte while a file that genuinely runs short still reports `Incomplete`.
- Response-stream admission is sized from the range instead of the artifact, so a resume reserves only the tail it will send. Usage, analytics and read counters are committed by the body once it has delivered its bytes rather than when the response is constructed (see the review fixes below), so an abandoned transfer meters nothing and its resumed follow-up meters only its tail.
- Aborted accelerated transfers now report the bytes already put on the wire instead of 0.

**CLI (`cli/`)**

- New `ArtifactResumeMiddleware` keeps what already arrived when a response body fails partway through and requests the remainder with a `Range` header. Wired into the cache client for the CAS and module-cache download operations.

### Why

During a saturation window on a large monorepo CI fleet, 3,185 transfers on a single pod exceeded 32.8 seconds (the top bucket of `kura_artifact_egress_duration_seconds`), which is where client timeouts fire. Fleet-wide that hour shows 5,810 transfers past the same bucket, against 730,400 admission-shed 503s and 294,044 successful reads. Without resume, a transfer that died at 90% consumed 90% of its bytes for zero value and its retry started over. That inflates offered load at exactly the moment the link is most congested, and is a plausible contributor to the observed congestion-collapse sawtooth (5-minute throughput swinging between 100 and 2833 Mbit/s). Average object size for the heaviest tenant is about 2.83 MB, but the distribution has a long tail and the tail is what times out.

### Decisions worth a reviewer's attention

**A range past the end returns 416, never a widened 200.** This is the one range case that must be refused rather than served whole. A client that appends a full body to a partial file corrupts it silently, with no error anywhere to trace it back to. Multi-range and unparseable range headers *are* served whole, which RFC 9110 permits and which keeps every existing client working unchanged.

**Range is resolved after the access check, on both planes.** A 416 carries `Content-Range: bytes */<size>`, so resolving before authorization would let an unauthorized caller learn an artifact's size.

**The waste metric reuses `record_artifact_egress` rather than adding a family.** The transfer functions now thread the sent count out on the error path, so `kura_artifact_egress_bytes_total` split by `result` (`ok` vs `client_aborted` / `write_timeout` / `incomplete` / `aborted`) measures the restart waste directly. A new metric family would have duplicated a dimension the existing one already has. Usage and analytics stay unrecorded for those bytes: the tenant is not billed for bytes that never landed.

**That metric did not previously mean what the paragraph above claims, which is why the second commit is here.** `InstrumentedArtifactStream` recorded `result="aborted"` from `Drop` whenever it had not seen the terminal `None`, and Hyper stops polling a body once the promised `Content-Length` is on the wire, so that `None` never arrives on a healthy response. The label was tracking which serving path answered, not whether the client got its artifact. Verified against production over 24h:

- `result="ok"` equals `path="accelerated"` to six significant figures every hour.
- Bytes per response are the same for both labels (1.33 MB aborted, 1.35 MB ok). A real mid-transfer abort averages well below the object size.
- During a one-hour incident the aborted:ok ratio was 4.04, *lower* than quiet hours reaching 9.4.
- That hour reconciles exactly: `path="streaming"` 964,959 = `result="aborted"` 236,324 + 730,400 admission-shed 503s, which never construct a stream at all.

Completion is now decided by byte count against the length the response promised, which is the same number the range work already threads through both paths. This is load-bearing for the range work rather than incidental: `client_aborted` exists only on the accelerated path and module traffic is ~91% streaming, so a genuine client hangup on the dominant path was invisible both in the result labels and in the HTTP status, since the Axum path has already sent 200 by then.

**Sizing admission from the range is the point of item 3, not a micro-optimization.** The admission call rejects with `QueueFull` when `requested_bytes` exceeds the pool. On the Axum path an inline artifact previously requested its whole size, so a large artifact's resume would be shed under a budget it should have fit in. It now reserves only the tail it will send.

**The CLI middleware buffers, and that is why it is scoped by operation ID.** OpenAPIRuntime's `next` closure is non-escaping, so a resume cannot be driven lazily from inside the returned stream; the body has to be assembled inside `intercept`. Both target call sites already did `Data(collecting:upTo: .max)` immediately, so there is no regression for them, but applying this to every response the client handles would be a real behavior change. Hence the allowlist.

**A resumed 206 whose `Content-Range` does not start at the offset reached is refused, not appended.** A proxy rewriting the range, or an artifact replaced between the two requests, would otherwise corrupt the assembled file with no signal. A server that ignores the header answers 200 and the download restarts, so the client degrades to exactly today's behavior against a serving path without range support.

**`If-Range` and ETag validators are deliberately omitted.** These keys are content-addressed (CAS digests, module hashes), so a resumed range cannot splice two different bodies in practice. Worth revisiting if a mutable-key route ever lands on this path.

### What this does not fix

The incident that prompted this was dominated by admission shedding, not by slow transfers: 730,400 shed 503s in the hour against 5,810 transfers past the 32.8s bucket and 170 recorded client aborts. Resume reduces retry-driven re-offered load, which feeds the shed, but it does not address the shed directly. The jittered `Retry-After` and 429-instead-of-503 changes noted below target that. Sizing the resume benefit precisely also needs the labelling fix in this PR to have been running for a while, since genuine aborts on the streaming path were not previously measurable.

### Impact

Clients that already send `Range` (curl `-C -`, most HTTP libraries) get resume for free. Clients that do not are unaffected: the only behavior change on an unranged request is the additional `Accept-Ranges` response header. The CLI resumes automatically on the CAS and module-cache download paths.

Rollout safety: the change is additive on the wire and safe under version skew. An old node ignores a `Range` header and answers 200 with the whole artifact, which the new CLI middleware handles as a restart rather than an error.

### Validation

- 724 kura unit tests pass, including new coverage for: range resolution (open-ended, closed, suffix, unsatisfiable, and the shapes that fall back to a full response), the 206 wire format on the accelerated path, the Axum 206 / 416 / `Accept-Ranges` responses, a two-request resume that reassembles the artifact byte-for-byte, and a store-level ranged read on a segment holding more than one artifact, which is the test that would catch the offset arithmetic returning plausible bytes from the wrong place.
- Every new behaviour was confirmed red before its fix, not merely against a changed signature: the labelling tests against the previous `Drop` behaviour, the metering test without the completion guard, and the ordering test with the two middlewares composed the other way round. 728 kura tests pass overall.
- `cargo clippy --all-targets -- -D warnings` is clean; `cargo fmt --check` is clean.
- 63 TuistHTTP tests (9 new) and 72 TuistCache tests pass. `mise run lint` is clean.
- The e2e accelerated-serving spec gains a resume case covering 206, `Content-Range`, byte-for-byte reassembly, `Accept-Ranges` on the full response, and the 416. **Not run locally** (needs the docker-compose stack); it runs in CI.

### Review fixes (third commit)

- **Resume requests bypassed retry handling.** Middlewares nest in array order, so placing `ArtifactResumeMiddleware` after `RetryMiddleware` gave it a `next` without retry in it. A ranged follow-up meeting the documented 503 was abandoned: the CAS lane failed outright (transport retries are disabled there) and the module lane restarted from byte zero, which is the waste the middleware exists to avoid. Resume now sits outside retry. A composition test pins the relationship and was confirmed to fail with the two swapped, since the call site only orders an array and nothing else would notice.
- **Streaming responses metered downloads that were never delivered.** Usage and analytics were booked as soon as `serve_file` returned a successful status, but on the streaming plane that status is set when the stream is constructed, before a byte reaches the client. An aborted full response metered the whole artifact and the successful ranged follow-up metered the tail again. Attribution now rides on the response body as `DownloadAttribution` and is committed from the same completion path that decides the egress result, using bytes actually delivered. This aligns the streaming plane with the accelerated one, which already waited for completion.

  Worth flagging: a resumed download is now metered for its tail alone, not for the artifact. Whether a partial delivery should bill for bytes that did leave the node is a product question this PR does not settle, and it is the one open decision here.
- **`@unchecked Sendable` in a test helper** replaced with `Synchronization.Mutex` behind a checked conformance, per the repository's Swift review rule.

### Coordination

Both changes that touch this PR's admission-shed path have landed and are merged in: 429 instead of 503 for a capacity shed (#12548) and the jittered `Retry-After` (#12549). The one conflict was the accelerated plane's denial branch, which both sides rewrote; the resolution takes main's JSON body together with this branch's per-denial headers, since a 416 has to carry `Content-Range`.

Their 429 makes this PR's middleware-ordering fix load-bearing rather than theoretical. 429 is a retryable status, so with resume correctly outside retry a shed ranged follow-up is retried with the jittered delay; with the ordering the review caught, it would have been abandoned and the partial bytes discarded.

### How to test locally

Start the local mesh and round-trip an artifact with a range:

```bash
cd kura && docker compose up --build -d
```

```bash
curl -sS -X POST "http://localhost:4101/api/cache/cas/range-demo?tenant_id=default" -H "content-type: application/octet-stream" --data-binary @<(head -c 262144 /dev/urandom)
```

```bash
curl -sS -D - -o /dev/null -H "range: bytes=100000-" "http://localhost:4101/api/cache/cas/range-demo?tenant_id=default"
```

Expect `HTTP/1.1 206 Partial Content`, `content-range: bytes 100000-262143/262144`, `content-length: 162144`, and `accept-ranges: bytes`. An unranged GET returns 200 with `accept-ranges: bytes` and no `content-range`; `range: bytes=999999-` returns 416 with `content-range: bytes */262144`.

The unit and e2e suites:

```bash
cd kura && mise run test-unit
```

```bash
cd kura && docker compose build && mise exec -- shellspec spec/e2e/accelerated_serving_spec.sh
```

```bash
xcodebuild test -workspace Tuist.xcworkspace -scheme Tuist-Workspace -only-testing TuistHTTPTests CODE_SIGNING_ALLOWED=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY=""
```


cli/Sources/TuistHTTP/ArtifactResumeMiddleware.swift
```diff
@@ -0,0 +1,153 @@
...
+        let (response, body) = try await next(resumeRequest, nil, baseURL)
+        guard let body else { return nil }
+
+        switch response.status.code {
+        case 206:
+            // The offset is verified rather than trusted. A 206 that starts
+            // anywhere else, from a proxy rewriting the range or the artifact
+            // having been replaced between the two requests, would corrupt the
+            // assembled file silently, so it is refused instead.
+            guard Self.partialContentStart(of: response) == offset else {
```
pepicrft, line 209: This only verifies the range offset. Kura permits a newer write to replace an artifact under the same key, so the resumed request can return a valid `206` that starts at `offset` but belongs to a different representation. Concatenating it to the first response silently corrupts the artifact, including when both versions have the same size.

Please bind the resume to a response validator. Return an [entity tag (`ETag`)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag) on the initial response, send `If-Range` on the follow-up, and restart when it no longer matches. Please add a regression test that replaces the artifact between the two requests.

cli/Sources/TuistHTTP/ArtifactResumeMiddleware.swift
```diff
@@ -0,0 +1,153 @@
...
+
+        let (response, responseBody) = try await next(request, body, baseURL)
+        // Only a whole-artifact success is resumable. Errors carry a short JSON
+        // body the caller needs verbatim, and a server that already answered
+        // 206 was not asked to by this middleware.
+        guard response.status.code == 200, let responseBody else {
+            return (response, responseBody)
+        }
+
+        var collected = Data()
```
pepicrft, line 97: This collects the entire download in `Data` even when the transfer never fails. The returned `HTTPBody(collected)` copies it into the runtime buffer, and both download services then collect that body into another `Data`. A module artifact can be 2 gigabytes, so this creates more than 6 gigabytes of full-size buffers before decompression and can terminate the client for large cache hits.

Please use file-backed resume storage or return the assembled bytes directly from a dedicated download path, so the successful path does not retain multiple full copies of the artifact.

kura/src/http.rs
```diff
@@ -1820,6 +1832,7 @@ async fn get_xcode(
         Some(&id),
         analytics,
         Some(usage),
+        request_range_header(&headers),
```
pepicrft: The cache client contract still declares only a `200` response and has no `Range` request header. The generated client therefore cannot issue a typed range request and treats `206` and `416` as undocumented responses.

Please update `cli/Sources/TuistCache/OpenAPI/cache.yml` with the request header and the partial and unsatisfiable responses, then regenerate the client so its public contract matches this handler.

pepicrft (approved):

esnunes:
reviewing

cli/Sources/TuistHTTP/ArtifactResumeMiddleware.swift
```diff
@@ -0,0 +1,153 @@
...
+        while let stream = pending {
+            pending = nil
+            do {
+                for try await chunk in stream {
+                    collected.append(contentsOf: chunk)
+                }
+            } catch is CancellationError {
+                throw CancellationError()
+            } catch {
+                guard attempt < maximumResumeAttempts, !collected.isEmpty else { throw error }
```
esnunes: The resume decision here keys only on "the body stream threw", never on how much was collected relative to what the response promised. I think that leaves one case worse off than before: if the transport delivers every byte and then errors on close (a reset after the final chunk), `collected.count` already equals the artifact size, so the resume asks for `bytes=<size>-`, the server correctly answers 416, `resume(from:)` hits its `default` and returns nil, and the original error gets rethrown. A download that fully arrived is discarded — and it costs an extra round-trip to find that out.

Would it be worth anchoring on the original response's `Content-Length` before deciding to resume?

    let expected = response.headerFields[.contentLength].flatMap(Int.init)
    if let expected, collected.count >= expected { return (response, HTTPBody(collected)) }

The same anchor would also catch the other direction — a body that ends cleanly but short is currently returned truncated, since nothing compares the final count against the promise.

The existing `truncatedBody` helper already produces the right shape for a test: chunks summing to the full length, then a throw.

kura/src/accelerated_file_serving.rs
```diff
@@ -457,16 +464,41 @@ async fn open_and_authorize(
...
+    // Resolved after access, so an unauthorized caller cannot learn an
+    // artifact's size from a 416's `Content-Range`.
+    let range = match resolve_range(parsed.headers.get("range").map(String::as_str), file.size) {
+        RangeOutcome::Full => ServedRange::full(file.size),
+        RangeOutcome::Partial(range) => range,
+        RangeOutcome::Unsatisfiable => {
+            return ClassifiedRequest::Deny(Denial {
+                header_len: parsed.header_len,
+                route: artifact.route,
+                status: 416,
```
esnunes, line 499: Small asymmetry between the planes, definitely not a blocker: the Axum path records `record_artifact_read(producer, "range_not_satisfiable", 0)` before returning its 416, but the accelerated 416 goes out through `ClassifiedRequest::Deny`, and that branch in `serve_connection` only calls `record_http`. So `kura_artifact_read_total{result="range_not_satisfiable"}` will only ever see 416s from the Axum plane — and the accelerated one is the plane carrying plain-HTTP artifact GETs on Linux, so it's the one where a client sending bad ranges would most likely surface.

Would it be worth adding the same `record_artifact_read` next to the `Denial` construction (or in the Deny branch keyed on `status == 416`) so the counter reads the same on both planes?

esnunes (approved):
Read through the whole branch — the shared range resolver, both serving planes, the store-level ranged read, and the Swift middleware with its wiring. This is a really solid piece of work, and a few of the choices in it are the non-obvious ones: resolving the range after the access check on both planes so a 416's `Content-Range` can't leak an artifact's size to an unauthorized caller, refusing an out-of-range request rather than widening it to a 200, verifying the resumed `Content-Range` offset client-side instead of trusting it, and putting resume outside retry with a composition test that pins the relationship. The store test that reads from partway into a segment holding more than one artifact is exactly the test that catches the failure mode that would otherwise be silent, and going back to correct what the `aborted` label had actually been measuring — with production evidence — rather than building on top of it is the part I'd most want other PRs to copy.

Approving. Two non-blocking notes inline, neither of which needs to hold this up.

The one I'd most like your read on is the CLI middleware: the resume decision keys only on "the body stream threw", never on how much was collected against what the response promised, so a body that delivers every byte and then errors on close triggers a resume that is guaranteed to 416, and the complete download is discarded. The other is smaller — the accelerated plane's 416 doesn't record the `range_not_satisfiable` read result its Axum counterpart does, so that counter only ever sees one plane.

One thing I don't think is mine to call, and you already flagged it: a resumed download now bills only for its tail, so bytes that did leave the node during the abandoned attempt go unbilled to anyone. That reads as the right default to me and it matches what the accelerated plane always did, but it's a revenue-affecting semantics change that probably wants a product sign-off rather than a reviewer's.

cli/Sources/TuistHTTP/ArtifactResumeMiddleware.swift
fortmarek: Fixed in 3a284f46bb. You were right that the offset check was no guard at all: a replacement is served from zero, so its tail starts exactly where the client asked, and equal sizes leave the lengths agreeing too — the comment there claimed to cover replacement and never did. Kura now returns an `ETag` derived from the manifest's `version_ms` and size (the store rejects a write whose version does not advance, so every accepted replacement changes it), both serving planes honour `If-Range`, and the client sends it on every resume, refuses a 206 naming a different representation, and will not resume at all against a server that named nothing. Regression tests cover replacement both ways: honoured (200, restart) and ignored (206 with a different tag, refused), each confirmed red without the check.

cli/Sources/TuistHTTP/ArtifactResumeMiddleware.swift
fortmarek: Fixed in 3a284f46bb, and it was worse than three copies: the transport uses `session.data(for:)`, so the body is already whole in memory before any middleware sees it — four full-size buffers for a 2 GB module. A body the transport already holds cannot fail partway through and costs nothing to re-read, so it is now passed through untouched; that is every response on this transport today, so the copies are gone from the path that runs, and the collecting path remains only for a transport that streams. Flagging the consequence rather than burying it: resume only does anything once a download actually streams, since on this transport a mid-transfer failure throws before a body exists. Happy to follow up with a streaming download path if you want resume live on the CLI.

kura/src/http.rs
fortmarek: Fixed in 3a284f46bb. `cache.yml` is generated (`mix openapi.spec.yaml --spec CacheWeb.API.Spec`), so I declared the `Range` and `If-Range` request headers and the 206/416 responses on the two cache controllers and regenerated the spec and client from them rather than editing the yml by hand. Both download services now refuse a ranged response outright: neither sends a `Range`, so a partial body is not an answer to the request they made and returning it would store a truncated artifact under a key claiming to be whole.

cli/Sources/TuistHTTP/ArtifactResumeMiddleware.swift
fortmarek: Fixed in ba421740c3 — you were right on both directions, and the second was the worse one. Anchoring on `Content-Length` as you suggested: a transfer holding everything it was promised is complete however its stream ended, so it is kept and no ranged request is made; a transfer holding less is truncated whether or not the stream said so, and is resumed on the same attempt budget rather than returned. The clean-but-short case was silently handing back a fragment as the whole artifact, which is the bug I would least have wanted to ship. Both confirmed red first: without the anchor the complete-then-error case throws its download away, and the short case returns four of ten bytes. The promise also travels with the response across a 200 restart, since a server refusing the range may be serving a different artifact than the one the download began with. `truncatedBody` was indeed the right shape — I added a `shortBody` sibling that ends cleanly for the other case.

kura/src/accelerated_file_serving.rs
fortmarek: Fixed in ba421740c3 — counted at the `Denial` construction, where `state` and `artifact.producer` are both in scope, so both planes now report `range_not_satisfiable` identically. Your point about which plane matters is the reason I did not leave it: the accelerated one carries the plain-HTTP artifact GETs on Linux, so it is exactly where bad ranges would show up and exactly where the counter was blind. One caveat worth stating rather than implying: this has no test. It sits in `open_and_authorize`, which needs a live TCP stream, and neither plane's 416 counter was covered before — so it is verified by inspection and the build only.

PR #12552, opened 2026-08-21
fix(server): stop the widget title hover from drawing a full-width border, and round the test history markers

fortmarek:
Three visual defects on the test case page: two dark-mode-only CSS bugs, and one chart that hid data.

### Widget title hover: too wide, and a border that should not be there

The dropdown that titles an analytics widget is the only stretch-aligned child of a column flex container that itself carries `flex-grow: 1` (needed to push the tooltip icon to the right edge). The trigger button therefore spanned the entire header: 378px of hover target for a 154px label.

On top of that, hovering it in dark mode drew a secondary-button border. Noora scopes its dark hover style as:

```
html[data-theme="dark"] .noora-dropdown [data-part="trigger"]:hover
```

The `html[...]` prefix gives that rule specificity (0,4,1), just above the (0,4,0) of the widget's own `box-shadow: unset`. So the widget's reset won in light mode and silently lost in dark, which is why this only ever showed up on the dark dashboard.

The trigger now shrinks to its label with `align-self: flex-start`, and the chrome reset is repeated under `:hover` at (0,5,0) so it wins in both themes. The label still dims on hover, so the affordance is unchanged.

This was not specific to the duration widget: it affected every widget with a dropdown title in dark mode, which includes the build run, shards, module cache, and tests pages.

Overriding the Noora rule locally, rather than changing `noora/css/dropdown.css`, keeps the dark hover treatment intact for every ordinary dropdown in the product. The widget is the one place that deliberately strips a dropdown of its button chrome, so the exception belongs with the widget.

### Test history markers rendered as squares

The timeline markers asked for `border-radius: var(--noora-radius-full)`. That token does not exist anywhere: Noora's radius scale is numeric and ends at `--noora-radius-99`. An undefined custom property makes the whole declaration invalid at computed-value time, so the markers fell back to a 0px radius and drew as squares rather than the intended circles.

Switching to `--noora-radius-99` restores the original intent. A token is not added to Noora for this, since the existing scale already covers it.

### Isolated chart buckets drew nothing at all

The rate and duration line charts set `symbol: "none"`. A bucket whose neighbours are both empty therefore had no line segment to sit on and no marker either, so it rendered as blank space. A test case that runs once a week charted as an empty canvas even though every one of its buckets held data. This was the real defect behind the reported "gaps in the chart".

Two alternatives were considered and rejected:

Zero-filling misreports. 0% reliability means every run failed, and 0ms means the test took no time; neither is what "the test did not run" means. The series deliberately carry `nil` for empty buckets, and the widgets above the chart would disagree with a zero-filled line.

Bars are right for counts and wrong for rates here. The runs chart is already stacked bars and stays that way, because a day with no runs genuinely had zero runs. But flakiness rate sits at 0% for a healthy test, so a bar chart of it renders an empty canvas, and a genuine 0% reliability day would be a zero-height bar indistinguishable from no data.

So the buckets that stand alone now carry a symbol of their own, computed in `chart_points/2`, while the rest keep a size of zero so a test case that runs every day still reads as a line rather than a row of dots.

This is what Buildkite Test Engine does on the equivalent page: bars for the execution count, and lines with a marker on every bucket for reliability and for the duration percentiles. Trunk takes the other route and does not chart a per-test rate over time at all, showing the window's composition as a Main-vs-PR proportion bar instead, with a stacked daily bar chart as its only time series. That remains a possible follow-up, but it is a redesign rather than a fix.

### Validation

Both CSS fixes were verified in a browser against the compiled bundle in dark and light mode, measuring the trigger box and the computed border radius directly rather than eyeballing.

The chart change is covered by three tests in `test_case_live_test.exs` that read the real serialized chart option out of the rendered DOM: a lone day carries a symbol, a run of consecutive days does not, and the series still declares a symbol at all (without which the sizes would be inert). Two existing tests were updated for the new point shape. The full file passes with `--warnings-as-errors`, `mix format --check-formatted` is clean, `mix credo --strict` reports nothing new, and `npx prettier@3.4.2 --check assets` passes.

### Follow-up noted separately

The radius token turned out not to be the only one of its kind: 12 more `--noora-*` names are referenced across the app CSS but defined nowhere, so those declarations are silent no-ops. Tracked separately rather than widening this change.

### How to test locally

1. `cd server && mix esbuild app`, then open a test case page in dark mode (`/{account}/{project}/tests/test-cases/{id}`).
2. Hover the title of any analytics widget. The highlight should cover only the label and chevron, not the full header width, and no border should appear. Switch to light mode and confirm the hover is unchanged from before.
3. Check the "Test history" panel on the right: the event markers should be circles.
4. Open a test case that runs infrequently and select the reliability, flakiness, or duration widget. Days that stand alone should show a dot rather than nothing; a test that runs daily should still read as a plain line.


fortmarek:
This is how the new UI looks like: 
image

pepicrft (approved):

